PR #84, opened 2021-12-07
Add ECR Github Action

ellioty:


Voxelot:
Now that the repos are public, we no longer require the SSH configs in the build stage.

.github/workflows/push_to_ecr.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Setup SSH agent
+        uses: webfactory/ssh-agent@v0.5.3
+        with:
+          ssh-private-key: |
+            ${{ secrets.PRIVATE_KEY_FUEL_ASM }}
+            ${{ secrets.PRIVATE_KEY_FUEL_TX }}
+            ${{ secrets.PRIVATE_KEY_FUEL_VM }}
+            ${{ secrets.PRIVATE_KEY_FUEL_STORAGE }}
+            ${{ secrets.PRIVATE_KEY_FUEL_TYPES }}
+
```
adlerjohn, line 32: Specifically
```suggestion
```

Voxelot (approved):

.github/workflows/push_to_ecr.yml
```diff
@@ -0,0 +1,51 @@
...
+jobs:
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Extract branch name
```
Voxelot, line 23: this works, although you may want to take a look at the docker-meta action that @rfuelsh uses as it doesn't require any custom string parsing etc.

ellioty: I think that's a valid approach, however, that would subsequently make a dependency on the GCR image -- not sure if that makes for a better approach. This action would work independently of GCR

Voxelot: I don't think the meta action is tied at all to GHCR, the registry is configurable. The main reason I'm asking about this is that it would be nice if there was consistency in how docker images are tagged and pushed if possible.

ellioty: Just updated to use Docker meta

rfuelsh (approved):

.github/workflows/push_to_ecr.yml
rfuelsh: @ellioty - this is great work! 

I see that it works -> https://us-west-2.console.aws.amazon.com/ecr/repositories/private/756756437019/fuel-core?region=us-west-2 --> Last push was January 23, 2022, 14:46:21

I enabled scanning on this ecr repo so we can look at vulnerabilities here: https://us-west-2.console.aws.amazon.com/ecr/repositories/private/756756437019/fuel-core/image/sha256:2def3ca8c7daa66b0aeaf938e171373667b4f4e463d5f939e6fe2d4c0b0badce/scan-results/?region=us-west-2 cc @Voxelot 

@Voxelot @ellioty - any logic why we have this ECR repo in us-west-2 vs. us-east-1? It doesn't matter but I can adjust EKS deployments to us-west-2

ellioty: @rfuelsh just updated it to be `us-east-1` -- let's unify everything

Voxelot: Isn't us-east-1 where all the AWS outages have happened since AWS internally centralizes everything there?

.github/workflows/build_and_publish_ecr.yml
```diff
@@ -0,0 +1,115 @@
...
+      logLevel:
+        description: "Log level"
+        required: true
+        default: "warning"
+
+env:
+  CARGO_TERM_COLOR: always
+  DOCKER_BUILDKIT: 1
+  RUSTFLAGS: -D warnings
+  REGISTRY: 756756437019.dkr.ecr.us-east-1.amazonaws.com/fuel-core
```
Voxelot, line 15: Can we put this in a real env var in github rather than in a yaml file?

Voxelot (approved):

.github/workflows/build_and_publish_ecr.yml
ellioty: Let's use Github Secrets, that would be the best practice

ellioty:
Closing this PR in favor for this: https://github.com/FuelLabs/fuel-deployment/pull/4

PR #99, opened 2021-12-29
[txpool] Initial impl of Transaction Pool

rakita:
* [txpool] Some checks on inclusion
* [txpool] Initial dependency insert
* [txpool] dep insertion needs testing
* [txpool] wip testing
* [txpool] some testing data
* [txpool] Initial subscriber. Some tidying up
* [txpool] cleanup
* [txpool] integration for fuel-core
* [txpool] find dependent. fmt
* [txpool] inclue utxo_id in dependency
* [txpool] Initial test
* [txpool] inclusion tests. Some fixes
* [txpool] fmt,clippy,cargo-sort
* [txpool] tx removal. subs tested
* [txpool] small cleanups and refactoring


Cargo.toml
```diff
@@ -1,6 +1,9 @@
 [workspace]
 # Use the new resolver to prevent dev-deps and build-deps from enabling debugging or test features in production.
 members = [
+  "txpool",
+  "interfaces",
+  "block-producer",
```
Voxelot, line 6: We should keep the block producer independent of this PR unless there's some traits we require from it.

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,507 @@
...
+    is_spend_by: Option<TxId>,
+    /// how depth are we inside UTXO dependency
+    depth: usize,
+}
+
+#[derive(Debug, Clone)]
+pub struct ContractState {
+    //is Utxo spend as other Tx input
+    used_by: HashSet<TxId>,
+    /// how depth are we inside UTXO dependency
```
leviathanbeak, line 30: nit: `how depth` -> `how deep` ?

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,507 @@
...
+    contracts: HashMap<ContractId, ContractState>,
+    /// max depth of dependency.
+    max_depth: usize,
+}
+
+#[derive(Debug, Clone)]
+pub struct CoinState {
+    //is Utxo spend as other Tx input
+    is_spend_by: Option<TxId>,
+    /// how depth are we inside UTXO dependency
```
leviathanbeak, line 22: nit: `how depth` -> `how deep`

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,507 @@
...
+    coins: HashMap<UtxoId, CoinState>,
+    /// Contract-> Tx mapping.
+    contracts: HashMap<ContractId, ContractState>,
+    /// max depth of dependency.
+    max_depth: usize,
+}
+
+#[derive(Debug, Clone)]
+pub struct CoinState {
+    //is Utxo spend as other Tx input
```
leviathanbeak, line 20: add 3 `///`

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,507 @@
...
+    is_spend_by: Option<TxId>,
+    /// how depth are we inside UTXO dependency
+    depth: usize,
+}
+
+#[derive(Debug, Clone)]
+pub struct ContractState {
+    //is Utxo spend as other Tx input
```
leviathanbeak, line 28: add 3 `///`

rakita:
> Looks like CI breaking

cargo patch fails to load fuel-tx, i am blocked by fuel-tx not being published upstream.

fuel-core/src/database.rs
```diff
@@ -171,46 +176,4 @@ impl InterpreterStorage for Database {
...
-        InterpreterError::Io(std::io::Error::new(std::io::ErrorKind::Other, e))
-    }
-}
-
-impl From<KvStoreError> for InterpreterError {
-    fn from(e: KvStoreError) -> Self {
-        InterpreterError::Io(std::io::Error::new(std::io::ErrorKind::Other, e))
-    }
-}
+pub use interfaces::db::KvStoreError;
```
Voxelot, line 179: Should this be at the top of the file with all the other imports?

fuel-core/src/tx_pool.rs
```diff
@@ -47,20 +50,32 @@ impl From<crate::state::Error> for Error {
     }
 }
 
+impl TxPoolDB for Database {}
+
 /// Holds submitted transactions and attempts to propose blocks
 pub struct TxPool {
     executor: Executor,
     db: Database,
+    fuel_txpool: Arc<dyn TxPoolTrait>,
```
Voxelot, line 59: Does this need to be internally arc'd if the `TxPool` struct is already arc'd? Maybe we could get away with a box instead?

interfaces/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+[dependencies]
+anyhow = "1.0"
+async-trait = "0.1"
+fuel-asm = "0.1"
+fuel-storage = "0.1"
+fuel-tx = { version = "0.2", features = ["serde-types"] }
+fuel-types = { version = "0.1", features = ["serde-types"] }
+fuel-vm = "0.2"
+futures = "0.3"
+hashbrown = "0.12"
```
Voxelot, line 18: According to the docs:
> Since Rust 1.36, this is now the HashMap implementation for the Rust standard library.

So we probably don't need to manually include this from crates.io anymore?

interfaces/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+parking_lot = "0.11"
+thiserror = "1.0"
+tokio = { version = "1.14", features = ["full"] }
```
Voxelot, line 21: Is tokio needed / can this be moved to a dev dep if so?

txpool/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "fuel-txpool"
```
Voxelot, line 2: we'll need to add these new crates to the tag publish tag validation scripts in CI, and ensure these version match fuel-core and fuel-client.

txpool/README.md
```diff
@@ -0,0 +1,60 @@
...
+Graph will be needed for relationship between dependable transactions so that if one of them get included/removed/replaced we need to know what to do with their dependent childrens.
+We will need map of all outputs in same structure so that we can fast search them. That would be something like `HashMap<UtxoId, CoinState>` and `HashMap<ContractId,ContractState>` for both Coin and Contract state.
+
+### Block inclusion Algorithm:
+It is most straightforward one: take PriceSort and iterate over transaction. DependencyGraph inclusion garantee us that every transaction in that sorted array can be included, but only after execution that trasansaction we would be sure that is can go inside block.
+
+### refs:
+* Some information on OpenEthereum and restrictions on tx inclusion: https://openethereum.github.io/Transactions-Queue
+* Possible out of scope, but it is relatively easy to add ethereum's eip1559 that adds Fee market change with basefee. More on eip1559: https://hackmd.io/@timbeiko/1559-resources
+Erigon did good docs on impl of eip1559 txpool: https://github.com/ledgerwatch/erigon/wiki/Transaction-Pool-Design
```
Voxelot, line 60: Should we have a section about future work that may be required? Like managing pending txs that aren't currently valid and waiting for other data, or byte prices independent of gas price?

txpool/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Dependency graph
+
+Few reasonings on decision and restrains made by txpool
+
+Visualization of dependencies between three transactions:
+![UTXO dependency diagram](../../docs/diagrams/fuel_v2_client_design_UTXO_dependency.jpg)
+
+Problem1: T2 arriving before T1.
+Solution: Have restriction that broadcast of hashes need to contains ancestor hashes and they all need to be sorted by gasPby_hashrice
```
Voxelot, line 38:
```suggestion
Solution: Have restriction that broadcast of hashes need to contain ancestor hashes and they all need to be sorted by gasPrice.
```

txpool/src/config.rs
```diff
@@ -0,0 +1,16 @@
+#[derive(Debug, Clone)]
+pub struct Config {
+    /// Maximum number of transaction inside pool
+    pub max_tx: usize,
+    /// max depth of connected UTXO excluding contracts
+    pub max_depth: usize,
```
Voxelot, line 6: should we also have a max size in terms of bytes? That way nodes don't get DOS'd by tx with all their fields filled out to the max?

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,483 @@
+use crate::{types::*, Error};
+use fuel_tx::{Input, Output, UtxoId};
+use hashbrown::{HashMap, HashSet};
+//use interfaces::db::DB;
```
Voxelot, line 4:
```suggestion
```

txpool/src/txpool.rs
```diff
@@ -0,0 +1,336 @@
...
+            .sort
+            .iter()
+            .rev()
+            .map(|(_, tx)| tx.clone())
+            .collect()
+    }
+
+    // When block is updated we need to receive all spend outputs and remove them from txpool
+    pub fn block_update(&self /*spend_outputs: [Input], added_outputs: [AddedOutputs]*/) {
+        todo!()
```
Voxelot, line 93: Any reason not to implement this now?

interfaces/Cargo.toml
rakita: I am using `tokio::sync::RwLock` inside `TxPoolService`

interfaces/Cargo.toml
rakita: Removed

fuel-core/src/tx_pool.rs
rakita: Probably not needed, will see to change it.

txpool/README.md
rakita: Good idea, will be added

txpool/src/config.rs
rakita: Totaly agree, added it in `future work` readme section.

txpool/src/txpool.rs
rakita: I wanted to leave it after block inclusion and potential reorgs are flushed out, but when i think about it now we could just use list of inserted Transaction and go with that.
This could be good first task for someone to introduce himself with working's on txpool, having only one person per code module is not the best way. wdyt?

txpool/Cargo.toml
rakita: added this two lines to .github/workflows/ci.yml :
```
./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} txpool/Cargo.toml
./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} interfaces/Cargo.toml
```
and bumbed txpool,interface versions so that they are same as core,client.

interfaces/Cargo.toml
Voxelot: Isn't that a different crate though? I don't think we need tokio to define our interfaces do we?

relayer/Cargo.toml
```diff
@@ -0,0 +1,43 @@
+[package]
+name = "fuel-relayer"
```
Voxelot, line 2: Shouldn't this be a separate branch / PR?

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,482 @@
...
+        {
+            let i_owner = owner;
+            let i_amount = amount;
+            let i_color = color;
+            match output {
+                Output::Coin { to, amount, color } => {
+                    if to != i_owner {
+                        return Err(Error::NotInsertedIOWrongOwner.into());
+                    }
+                    if amount < i_amount {
```
Voxelot, line 121: Is there any reason why we'd not enforce the amount to be the same? Seems unlikely people would burn their own tokens like this, but seems like it would be better to ensure the amount in == amount out.

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,482 @@
...
+                        return Err(Error::NotInsertedIOWrongAmount.into());
+                    }
+                    if color != i_color {
+                        return Err(Error::NotInsertedIOWrongAssetIt.into());
```
Voxelot, line 125: What does `AssetIt` stand for? Should this be `AssetId`?

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,482 @@
...
+                    }
+                }
+                Output::Contract { .. } => return Err(Error::NotInsertedIOConractOutput.into()),
```
Voxelot, line 128:
```suggestion
                Output::Contract { .. } => return Err(Error::NotInsertedIOContractOutput.into()),
```

interfaces/src/txpool.rs
```diff
@@ -0,0 +1,127 @@
...
+    #[error("Transaction is not inserted. UTXO input contract is not existing: {0:?}")]
+    NotInsertedInputContractNotExisting(ContractId),
+    #[error("Transaction is not inserted. UTXO is not existing: {0:?}")]
+    NotInsertedInputTxNotExisting(TxId),
+    #[error(
+        "Transaction is not inserted. UTXO requires Contract input {0:?} that is priced lower"
+    )]
+    NotInsertedContractPricedLower(ContractId),
+    #[error("Transaction is not inserted. Input output missamatch. Coin owner is different from expected input")]
+    NotInsertedIOWrongOwner,
```
Voxelot, line 111: Capitalizing `IO` isn't following UpperCamelCase:

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn

https://rust-lang.github.io/api-guidelines/naming.html

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,482 @@
...
+                        }
+                    }
+                    // else do nothing, everything is variable and can be only check on execution
+                }
+                Output::ContractCreated { .. } => {
+                    return Err(Error::NotInsertedIOConractOutput.into())
+                }
+            };
+        } else {
+            panic!("Use it only for coin output check");
```
Voxelot, line 162: Why panic instead of just returning a normal `Error`?

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,482 @@
...
+                        max_depth = core::cmp::max(state.depth + 1, max_depth);
+                        if max_depth > self.max_depth {
+                            return Err(Error::NotInsertedMaxDepth.into());
+                        }
+                        // output is present but is it spend by other tx?
+                        if let Some(ref spend_by) = state.is_spend_by {
+                            // get tx that is spending this output
+                            let txpool_tx = txs
+                                .get(spend_by)
+                                .expect("Tx should be always present in txpool");
```
Voxelot, line 196: Can we use a result instead of a panic here?

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,482 @@
...
+                            // compare if tx has better price
+                            if txpool_tx.gas_price() > tx.gas_price() {
+                                return Err(Error::NotInsertedCollision(*spend_by, *utxo_id).into());
+                            } else {
+                                if state.depth == 0 {
```
Voxelot, line 201: Shouldn't we check if the tx depth if it's already in the db / included in a block before comparing gas prices?

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,482 @@
...
+                    }
+
+                    // yey we got our coin
+                }
+                Input::Contract { contract_id, .. } => {
+                    // Does contract exist. We dont need to do any check here other then if contract_id exist or not.
+                    if let Some(state) = self.contracts.get(contract_id) {
+                        // check if contract is created after this transaction.
+                        if state.gas_price > tx.gas_price() {
+                            return Err(Error::NotInsertedContractPricedLower(*contract_id).into());
```
Voxelot, line 276: Seems like we should have a test case to ensure this error is triggered correctly

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,482 @@
...
+                Output::Coin { .. }
+                | Output::Withdrawal { .. }
+                | Output::Change { .. }
+                | Output::Variable { .. } => None,
+                Output::Contract { input_index, .. } => {
+                    if let Input::Contract { contract_id, .. } = tx.inputs()[*input_index as usize]
+                    {
+                        Some(contract_id)
+                    } else {
+                        panic!("InputIndex for contract should be always correct");
```
Voxelot, line 404: This should be an error and not a panic

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,482 @@
...
+                let state = self.contracts.get(contract_id).cloned();
+                if let Some(state) = state {
+                    // call removal on every tx;
+                    for rem_tx in state.used_by.iter() {
+                        let rem_tx = txs.get(rem_tx).expect("Tx should be present in txs");
+                        removed_tx
+                            .extend(self.recursively_remove_all_dependencies(txs, rem_tx.clone()));
+                    }
+                } else {
+                    panic!("Contract state should be present when removing child");
```
Voxelot, line 420: ahhh more panics

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,482 @@
...
+                }
+            }
+        }
+
+        // now we are sure that transaction can be included. remove all collided transactions
+        let mut removed_tx = Vec::new();
+        for collided in collided.into_iter() {
+            let collided = txs
+                .get(&collided)
+                .expect("Collided should be present in txs");
```
Voxelot, line 314: Do we really need to panic here?

Voxelot:
I think this is a great start and a lot of thinking has gone into this, but given how complex this new system is with all the validation and corner cases I feel like we could easily use 2-3x the number of tests. While it'd be nice to deliver quickly, I value testing and safety even more.

txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,482 @@
...
+        txs: &'a HashMap<TxId, ArcTx>,
+        db: &dyn TxPoolDB,
+        tx: &'a ArcTx,
+    ) -> anyhow::Result<Vec<ArcTx>> {
+        let mut collided: Vec<TxId> = Vec::new();
+        // iterate over all inputs and check for colision
+        let mut max_depth = 0;
+        let mut db_coins: HashMap<UtxoId, CoinState> = HashMap::new();
+        let mut db_contracts: HashMap<ContractId, ContractState> = HashMap::new();
+        for input in tx.inputs() {
```
Voxelot, line 180: This function has a lot of deeply nested complexity, while that may be unavoidable in situations with lots of inherent complexity, I think there's probably some room to refactor this into smaller units.

interfaces/Cargo.toml
rakita: I think it is same crate, but i can minimize features included.

relayer/Cargo.toml
rakita: removed

txpool/src/containers/dependency.rs
rakita: I was strict and used == first, but i relaxed restriction few commits back. will return too stricter option.

interfaces/src/txpool.rs
rakita: sure

txpool/src/containers/dependency.rs
rakita: If this uses function named `check_if_coin_input_can_spend_output` is used for checking `coin_contract` than this means somebody is missusing it

txpool/src/containers/dependency.rs
rakita: As said for spend_by: `Tx should be always present in txpool`,

txpool/src/containers/dependency.rs
rakita: Comaprison does apply only for tx in txpool. This comparison is between transaction that collided on same output `spend_by` contains transaction in txpool that currently spends that output.

txpool/src/containers/dependency.rs
rakita: `Collided should be present in txpool` and we already checked that tx is present when comparing `gas_price`.

txpool/src/containers/dependency.rs
rakita: "InputIndex for contract should be always correct" this is not expected behaviour.

txpool/src/containers/dependency.rs
rakita: "Contract state should be present when removing child" this is not expected behaviour

txpool/src/containers/dependency.rs
rakita: There was a bug here it should be `if tx.gas_price() > state.gas_price { `, good catch. Added tests to check for it.

txpool/src/containers/dependency.rs
rakita: Will consider it.

rakita:
> I think this is a great start and a lot of thinking has gone into this, but given how complex this new system is with all the validation and corner cases I feel like we could easily use 2-3x the number of tests. While it'd be nice to deliver quickly, I value testing and safety even more.

I would even say even 10x number of tests to cover all the cases, more tests the better. Proper way to test it and be sure it works as expected is to create fuzzer of dependent transactions and randomly insert and remove them, and check potential errors and periodically check order and simulate block creation and block inclusion. And when we have client parts finished with all functionality and merged into one, we should start testing on test network and simulated stress tests will cover a lot of cases. We are startup and not big company and  should always weight on speed and scope of delivery, if we dont deliver quickly we will stop existing or miss out on benefits of being early.

I will add few more tests.

.github/workflows/ci.yml
```diff
@@ -108,6 +108,8 @@ jobs:
           cargo install toml-cli
           ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuel-client/Cargo.toml
           ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuel-core/Cargo.toml
+          ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} txpool/Cargo.toml
+          ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} interfaces/Cargo.toml
```
adlerjohn, line 112: Nit, but can you sort these? I'll help future us not deal with an unsorted mess 😂

Cargo.toml
```diff
@@ -8,6 +8,8 @@ members = [
   "fuel-indexer/indexer",
   "fuel-indexer/lib",
   "fuel-indexer/schema",
+  "interfaces",
+  "txpool",
```
adlerjohn, line 12: Ditto, sort pls

fuel-core/Cargo.toml
```diff
@@ -28,12 +29,14 @@ env_logger = "0.9"
 fuel-asm = { version = "0.1", features = ["serde-types"] }
 fuel-storage = { version = "0.1" }
 fuel-tx = { version = "0.2", features = ["serde-types"] }
+fuel-txpool = { path = "../txpool", version = "0.2" }
```
adlerjohn, line 32: Should use an exact version for local dependencies, to avoid headaches if we ever mess up semver. And makes find-and-replace easier.

fuel-core/Cargo.toml
```diff
@@ -28,12 +29,14 @@ env_logger = "0.9"
...
 fuel-types = { version = "0.1", features = ["serde-types"] }
 fuel-vm = { version = "0.2", features = ["serde-types"] }
 futures = "0.3"
 graphql-parser = "0.3.0"
 hex = { version = "0.4", features = ["serde"] }
 hyper = "0.14"
+interfaces = { path = "../interfaces", package = "fuel-interfaces", version = "0.2" }
```
adlerjohn, line 39: Should use an exact version for local dependencies, to avoid headaches if we ever mess up semver. And makes find-and-replace easier.

Can you maybe rename this crate to...`fuel-interfaces`?

txpool/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+repository = "https://github.com/FuelLabs/fuel-core"
+description = "Transaction pool"
+
+[dependencies]
+anyhow = "1.0"
+async-trait = "0.1"
+fuel-tx = { version = "0.2", features = ["serde-types"] }
+fuel-types = { version = "0.1", features = ["serde-types"] }
+futures = "0.3"
+interfaces = { path = "../interfaces", package = "fuel-interfaces", version = "0.2" }
```
adlerjohn, line 15: Ditto

txpool/src/config.rs
```diff
@@ -0,0 +1,16 @@
...
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            max_tx: 4064,
+            max_depth: 10,
```
adlerjohn, line 13: Hmmm. Did we not say this should be `1`? Or is this something else?

adlerjohn (commented):
Left a few nits, and second what Brandon said: nice work, needs some more tests, then **L**et's **G**et **T**his **M**erged.

txpool/src/config.rs
rakita: We talked if it is easier to do it with just depth 1, but it was more easier to do it with depth counting (logic is few lines).

fuel-core/Cargo.toml
rakita: It seem little bit strange, but i can see benefit, will be changed.

rakita:
Should be ready for merge. Added tests to cover all error cases and most usual usages.

fuel-core-interfaces/src/txpool.rs
```diff
@@ -0,0 +1,127 @@
...
+    + Sync
+{
+    fn transaction(&self, tx_hash: TxId) -> Option<Arc<Transaction>> {
+        Storage::<Bytes32, Transaction>::get(self, &tx_hash)
+            .ok()
+            .flatten()
+            .map(|t| Arc::new(t.as_ref().clone()))
+    }
+
+    fn contract_exist(&self, contract_id: ContractId) -> bool {
```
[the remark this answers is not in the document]
rakita: In this context this is fine, in both cases with or without result i will behave in similar way.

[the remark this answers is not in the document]
adlerjohn: Hmm I guess it's true from the perspective of the mempool that if it can't find a contract, then it doesn't really matter _why_ it can't find it. You'd just discard the transaction. Still seems like a cleaner API to have it be a `Result` in case you want to be able to handle certain types of failures and e.g. retry a few times if it's a network or other transient failure. 🤷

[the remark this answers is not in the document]
Voxelot: Generally speaking we'll likely want to be able to distinguish between erroneous data that someone has shared with us vs a failure to include in the tx pool due to local issues (i.e. db corruption). Otherwise building things like spam protection mechanisms will be much harder.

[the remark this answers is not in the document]
rakita: This is very insignificant, in terms if database call fails, that is usually point of no return and as point of no return, the least of the problem of user is transaction not included, client will stop syncing.
But as this seems as last comment on pull request, and it is not big code change, i made a commit, and in John words "**L**et's **G**et **T**his **M**erged".

adlerjohn (commented):
lgtm, will defer to @Voxelot for final review!

fuel-txpool/src/subscribers.rs
```diff
@@ -0,0 +1,49 @@
+use std::sync::Arc;
+
+use crate::{types::ArcTx, Error};
+use async_trait::async_trait;
+use fuel_core_interfaces::txpool::Subscriber;
+use parking_lot::RwLock;
```
Voxelot, line 6: [nit] Should we be using an async friendly RwLock instead?

fuel-txpool/src/service.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+}
+
+#[cfg(any(test))]
+pub mod tests {
+    use super::*;
+    use fuel_core_interfaces::db::helpers::DummyDB;
+    use std::str::FromStr;
+
+    //use crate::interface::tests::DummyDB;
```
Voxelot, line 133:
```suggestion
```

fuel-txpool/src/service.rs
```diff
@@ -0,0 +1,192 @@
...
+        }
+        let mut list: Vec<ArcTx> = seen.into_iter().map(|(_, tx)| tx).collect();
+        // sort from high to low price
+        list.sort_by_key(|tx| Reverse(tx.gas_price()));
+
+        list
+    }
+
+    /// Iterete over `hashes` and return all hashes that we dont have.
+    async fn filter_by_negative(&self, tx_ids: &[TxId]) -> Vec<TxId> {
```
Voxelot, line 87: this api is untested

fuel-txpool/src/service.rs
```diff
@@ -0,0 +1,192 @@
...
+                    self.subs.inserted(tx).await;
+                }
+                Err(_) => {}
+            }
+        }
+        res
+    }
+
+    /// find all tx by its hash
+    async fn find(&self, hashes: &[TxId]) -> Vec<Option<ArcTx>> {
```
Voxelot, line 59: looks like this api may be untested

fuel-core-interfaces/src/db.rs
```diff
@@ -0,0 +1,545 @@
...
+        pub fn dummy_tx(txhash: TxId) -> Transaction {
+            // One transfer tx1 depends on db
+            // One dependent transfer tx2 on tx1
+            // One higher gas_price transfer tx3 from tx1
+            // one higher gas_price transfer tx4 then tx2
+            // tx5 that depends on tx1 contract
+            // tx6 same as tx1 but without coin output
+
+            let db1 = TxId::from_str(
+                "0x00000000000000000000000000000000000000000000000000000000000000000",
```
Voxelot, line 99: [nit] Could we use consts for these instead? The tests are pretty hard to read when trying to cross-reference with these tx id strings here.

fuel-txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,635 @@
...
+                Output::ContractCreated { contract_id } => Some(*contract_id),
+            };
+            let utxo_id = UtxoId::new(tx.id(), index as u8);
+            if let Some(contract_id) = is_contract {
+                // insert contract
+                self.contracts.insert(
+                    contract_id,
+                    ContractState {
+                        depth: max_depth,
+                        used_by: HashSet::new(),
```
Voxelot, line 381: Wouldn't this overwrite all the used_by tx ID set on line 307?:
```
                        // insert into contract
                        db_contracts
                            .entry(*contract_id)
                            .or_insert(ContractState {
                                used_by: HashSet::new(),
                                depth: 0,
                                origin: None, //there is no need for owner if contract is in db
                                gas_price: GasPrice::MAX,
                            })
                            .used_by
                            .insert(tx.id());
```

fuel-txpool/src/txpool.rs
```diff
@@ -0,0 +1,473 @@
...
+        let mut txpool = TxPool::new(config);
+
+        let out = txpool.insert(tx1, &db).await;
+        assert!(out.is_ok(), "Tx1 should be OK, get err:{:?}", out);
+        let out = txpool.insert(tx2, &db).await;
+        assert!(out.is_ok(), "Tx2 dependent should be OK, get err:{:?}", out);
+    }
+
+    #[tokio::test]
+    async fn missing_dep_faulty_tx1_tx2() {
```
Voxelot, line 157: We should do better with our test names to clarify which requirements are being tested and what the expected behavior is. Referencing indices into helper data (`tx1`, `tx2`, etc) isn't a great way to clarify intent, and is harder to refactor if the test data changes over time.

Something like "insert_with_missing_input_coin_tx_fails" would be a lot more clear.

Voxelot:
After my last round of comments is addressed I'll be ready to merge this in. Sorry for such a long cycle on this one, maybe in the future we should try to break something like this up into smaller PRs.

rakita:
> After my last round of comments is addressed I'll be ready to merge this in. Sorry for such a long cycle on this one, maybe in the future we should try to break something like this up into smaller PRs.

It is fine, it paid off to have longed cycle, you have found two bugs by just reviewing it.

Not sure if this could be devided into smaller chunk as it covers core mechanisms that txpool is going to function by, i did best to somehow contain it and leave some work for later.

I am currently diving into relayer, and will address/make changes probably later today or maybe tomorrow morning.

fuel-txpool/src/subscribers.rs
rakita: For subscribers, parking_lot is better fit it is faster in this use case, bcs only `read` lock is used and it is not blocking.

fuel-txpool/src/txpool.rs
rakita: These are data specific tests, if data changes it is like rewriting new test. But name of this particualr test is not the greatest, renamed it to: "fails_to_insert_tx2_with_missing_utxo_dependency_on_faulty_tx1"

fuel-txpool/src/containers/dependency.rs
rakita: Yep, this is a bug, if output is contract it should do nothing, When i wrote this one month ago i didnt fully grasp all outputs, for withdrawal output it is same story.

fuel-txpool/src/service.rs
rakita: test made

fuel-txpool/src/service.rs
rakita: test made

fuel-core-interfaces/src/db.rs
rakita: Done the refactoring, and used `lazy_static` to make them static, it is a lot nicer now.

fuel-txpool/src/subscribers.rs
Voxelot: Ah right, and we'd only hold the lock when adding subscribers during initialization.

fuel-core-interfaces/src/txpool.rs
```diff
@@ -0,0 +1,122 @@
...
+
+use fuel_tx::{ContractId, UtxoId};
+use thiserror::Error;
+
+use crate::db::{Error as DbStateError, KvStoreError};
+use fuel_storage::Storage;
+use fuel_tx::Bytes32;
+use fuel_vm::prelude::Contract;
+
+pub trait TxPoolDB:
```
Voxelot, line 13: [**minor** nit]
```suggestion
pub trait TxPoolDb:
```

fuel-txpool/src/containers/dependency.rs
```diff
@@ -0,0 +1,632 @@
...
+                        utxo_id,
+                        CoinState {
+                            is_spend_by: None,
+                            depth: max_depth,
+                        },
+                    );
+                }
+                Output::ContractCreated { contract_id } => {
+                    // insert contract
+                    self.contracts.insert(
```
Voxelot, line 375: What should we do if two txs that don't collide contain the same output contract id?

fuel-txpool/src/txpool.rs
Voxelot: Ultimately what I'm getting at is we should rely more on conveying the requirements being tested, than the particular data being used. This is an improvement though but some of the other tests could probably use similar improvements.

fuel-txpool/src/containers/dependency.rs
adlerjohn: We should probably reject the later one unless it does RBF (iirc, pays more than some multiplier more fees)

rakita: Very unlikely to happen, we are talking here about sha3 colision: https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/identifiers.md#contract-id

fuel-core-interfaces/src/txpool.rs
rakita: Done.

fuel-txpool/src/txpool.rs
rakita: What test do you have in mind? btw to speed up this, be free to push commit dirrectly to branch i wouldn't mind, it is esentially change about preferences of test naming.

fuel-txpool/src/containers/dependency.rs
adlerjohn: @Voxelot are you referring to the _contract ID_, or the ID of the TXO of the contract? The latter would require a collision in the contract ID in the first place.

rakita: Wait, witness in contractid is not a signature, but a bytecode? That means someone can construct same contractid, and this collision check should be needed?

adlerjohn: The only time witness data is not a signature in a way that's interpreted by the consensus rules is for contract creation, one of the witnesses will be contract bytecode (but the consensus rules check that the contract ID of _that_ bytecode must match up with the contract ID in the contract created output).

Someone could try to create a contract with the same contract ID, but there must be a check that prevents this (cannot have two contract UTXOs with the same contract ID).

rakita: ContractId collision check added.

Voxelot (approved):

PR #100, opened 2022-01-02
Query.coins

AlicanC:
- Replace `Query.coinsByOwner(owner, ...)` with `Query.coins(filter: { owner, color }, ...)`
- Return only unspent coins

### DIFF
--- a/fuel-client/tests/coin.rs
+++ b/fuel-client/tests/coin.rs
@@ -5,6 +5,7 @@ use fuel_core::{
 };
 use fuel_gql_client::client::{FuelClient, PageDirection, PaginationRequest};
 use fuel_storage::Storage;
+use fuel_tx::Color;
 use fuel_vm::prelude::{Address, Bytes32, Word};
 
 #[tokio::test]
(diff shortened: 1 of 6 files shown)


Voxelot:
@AlicanC We should also have integration tests within the fuel-gql-client for this test suite to catch issues before they reach the sdk.

fuel-client/tests/coin.rs
```diff
@@ -89,3 +91,115 @@ async fn first_5_coins() {
...
+            PaginationRequest {
+                cursor: None,
+                results: 10,
+                direction: PageDirection::Forward,
+            },
+        )
+        .await
+        .unwrap();
+    assert!(!coins.results.is_empty());
+    assert_eq!(coins.results.len(), 5)
```
Voxelot, line 204: It's not clear to me if the test is expecting these to be unspent or only spent coins

AlicanC: The naming was wrong on that one, it's expecting unspent. I fixed that also added new assertions: https://github.com/FuelLabs/fuel-core/pull/100/commits/0dc03d1c727b99a7a648e4ea03db05f2f4050a97

Voxelot (approved):

PR #114, opened 2022-01-05
UtxoId changes

rakita:
changes for UtxoId: https://github.com/FuelLabs/fuel-tx/pull/53

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1390,8 +1390,8 @@ dependencies = [
  "fuel-asm",
  "fuel-gql-client 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "fuel-pest",
- "fuel-tx",
- "fuel-vm",
+ "fuel-tx 0.1.0",
+ "fuel-vm 0.1.1",
  "futures",
  "hex",
  "prettydiff",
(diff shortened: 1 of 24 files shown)


adlerjohn:
Is this good to go?

rakita:
It is still not good to go, i want to check some more small things, and want to Brandon to look at it so that he is familiar with it, it is more impactful than fuel-vm.

Cargo.lock
```diff
@@ -1510,9 +1510,9 @@ dependencies = [
  "env_logger",
  "fuel-asm",
  "fuel-storage",
- "fuel-tx",
+ "fuel-tx 0.1.0",
```
Voxelot, line 1513: it seems like we're getting mismatched versions of fuel-tx somehow

rakita: Probably some library that uses `fuel-tx` or `fuel-vm` is not updated, will check.

rakita: fueltx_01

It is used by example project `simple-wasm` that imports `fuels-abigen-macro`->`forc`->`fuel-gql-client`. 
To to do the update, we need to merge `fuel-gql-client` and update version to `forc`/`fuel-abigen-macro`

adlerjohn: Given how messy this is, and how the example project is more of an integration thing, should it simply be removed from the workspace?

rakita: Seems resonable to be removed or moved to somewhere else if still needed (not sure where).

adlerjohn (approved):
Changes to the lockfile can be clobbered in #140.

Cargo.lock
adlerjohn: #140

PR #124, opened 2022-01-09
`coinsByOwner(owner, color)` to `coins(filter:{owner,color})`

AlicanC:
This PR renames `coinsByOwner` to `coins` and packs the `owner` and `color` fields into a `filter` arg.

Normally, a filter would be optional, and all of its fields would also be optional, but there is no implementation to query without an owner so both `filter` and `filter.owner` are non-null.

(For example, if we were to pack `transactions` and `transactionsByOwner` together the `filter` arg would be optional.)

### DIFF
--- a/fuel-client/assets/schema.sdl
+++ b/fuel-client/assets/schema.sdl
@@ -71,6 +71,10 @@ type CoinEdge {
 	"""
 	cursor: String!
 }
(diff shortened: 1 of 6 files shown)


adlerjohn:
Noting that separately from this PR, there should be some renaming done from color to asset ID.

Issue #121, opened 2022-01-06
Coin Selection GraphQL API


Voxelot:
The SDK should use a special API to select which utxos to use as inputs, instead of directly picking utxos from the full set. This is important for ensuring consistent behavior across SDK's and the ability to implement efficient dust mitigation strategies such as [random-improve](https://iohk.io/en/blog/posts/2018/07/03/self-organisation-in-coin-selection/).

For inputs, the API will take:

`owner`: The `Address` of the utxo owner
`spend_query`: `Vec<(AssetId, Amount)>` The total amount of each asset type to spend
`max_inputs`: `u8` The max number of utxos that can be used

Success:
Return a list of utxos <= max_inputs that will satisfy the amounts to be spent.

Errors:
- If the spend amount is higher than the available balance of the owner.
- If reaching the spend amount requires the list of utxos to be greater than `max_inputs`

Voxelot:
@digorithm this is done, and will be available in the next release. Integration testing could start using the published docker image from master branch.

PR #131, opened 2022-01-14
Add `Query.coinsToSpend` to GQL schema

AlicanC:
Resovles #121 

- [x] Implement Largest-First
- [x] Implement Random-Improve
- [x] Add `Query.coinsToSpend` to GQL schema
- [x] Implement fuel-client query
- [x] Add fuel-client tests
- [x] Cleanup code (handle TODOs, optimize, etc.)


fuel-core/src/schema/coin.rs
```diff
@@ -167,4 +178,216 @@ impl CoinQuery {
...
+                    if collected_amount < *amount {
+                        return Err(CoinLookupError::NotEnoughCoins);
+                    }
+                }
+
+                Ok(coins)
+            };
+
+        // An implementation of the method described on: https://iohk.io/en/blog/posts/2018/07/03/self-organisation-in-coin-selection/
+        let random_improve =
```
Voxelot, line 280: These should be extracted into smaller modules/functions instead of defined in-line.

AlicanC: Addressed in this commit: https://github.com/FuelLabs/fuel-core/pull/131/commits/6d0ebefdf204333c87027d102f3f0688f16ba95c

Voxelot:
Some tests that might be good to add:

 - coin selection will fallback to largest first when random-improve fails
 - largest first actually uses the largest coins first

fuel-client/src/client.rs
```diff
@@ -264,6 +267,31 @@ impl FuelClient {
...
+        let spend_query: Vec<SpendQueryElementInput> = spend_query
+            .iter()
+            .map(|(color, amount)| -> Result<_, ConversionError> {
+                Ok(SpendQueryElementInput {
+                    color: color.parse()?,
+                    amount: (*amount).into(),
+                })
+            })
+            .try_collect()?;
+        let max_inputs: Option<i32> = max_inputs.map(|max_inputs| max_inputs as i32);
```
melekes, line 288: should we check if max_inputs is > `u32::MAX`? or the limit enforced somewhere else? otherwise a potential data loss is possible (usize on 64bit machine -> i32).

fuel-core/src/schema/coin.rs
```diff
@@ -168,4 +178,46 @@ impl CoinQuery {
...
+        #[graphql(desc = "The max number of utxos that can be used")] max_inputs: Option<i32>,
+    ) -> async_graphql::Result<Vec<Coin>> {
+        let owner: Address = owner.0.into();
+        let spend_query: Vec<(FuelTxColor, u64)> = spend_query
+            .iter()
+            .map(|e| (FuelTxColor::from(e.color.0), e.amount.0))
+            .collect();
+        let max_inputs: u8 = max_inputs
+            .map(u8::try_from)
+            .map(Result::unwrap)
```
melekes, line 198: why not return an error?

fuel-core/src/database/coin.rs
```diff
@@ -83,4 +84,36 @@ impl Database {
...
+            res.map(|(key, _)| {
+                UtxoId::new(
+                    unsafe { Bytes32::from_slice_unchecked(&key[32..64]) },
+                    key[64],
+                )
+            })
+        })
+        .filter_ok(move |id| {
+            Storage::<UtxoId, Coin>::get(self, id)
+                .unwrap()
```
Voxelot, line 113: These unwraps are going to be ok in practice currently, but should probably propagate an error in case something else changes the assumptions being made here in the future.

e.g. something like this (untested)

```
.filter_map(|res| {
    match res {
        Ok(id) => {
            Storage::<UtxoId, Coin>::get(self, id)
                  .and_then(|maybe_coin| maybe_coin.ok_or(NotFoundError))
                  .and_then(|coin| if coin.color == color { Some(coin) } else { None })
                  .transpose()
        },
        e => Some(e)
    }
})
```

AlicanC:
@melekes, I've fixed both of the issues you mentioned.

@Voxelot
>coin selection will fallback to largest first when random-improve fails

I couldn't find a way to test this, but there is a test that has a probability to fail if we break the fallback. If you think it's worth it I can find a way to manipulate the RNG to test this. I also couldn't test dust management for that reason. Tests only verify if the queried amount is met.

I also took note of your comment on `Database::owned_coins_by_color`. Since you've said it's ok for now, I will create a follow up issue to decide what index to add. Adding an index will both cleanup the part you mention and also `random_improve` and `largest_first`. (In SQL speak, I'm thinking of `ON (owner, color) WHERE status = UNSPENT` ATM.)

fuel-core/src/coin_query.rs
```diff
@@ -0,0 +1,446 @@
...
+                        .ok_or(KvStoreError::NotFound)?
+                        .map(|coin| (id, coin.into_owned()))
+                })
+                .filter_ok(|(_, coin)| coin.status == CoinStatus::Unspent)
+                .try_collect()?;
+
+            Ok(coins)
+        })
+        .try_collect()
+        .unwrap();
```
Voxelot, line 153:
```suggestion
        .try_collect()?;
```

fuel-core/src/coin_query.rs
```diff
@@ -0,0 +1,446 @@
+use std::cmp::Reverse;
+
```
Voxelot, line 2:
```suggestion
```

Voxelot (approved):
Thanks for the additional tests and updates. I'll file separate issues if I think of new ways we should test this.

PR #139, opened 2022-01-18
Transaction Status

### LINKED ISSUE #125: Add transaction status to fuel-gql-client
The execution result of a transaction is available on the raw GQL API, but needs to be exposed via the Rust client.

Voxelot:
closes #125

Updates error handling of transactions so that revert's / panic's are marked as failure instead of success. 

Provides more useful info about the transaction status via fuel-client.


adlerjohn (approved):
utACK

PR #142, opened 2022-01-18
Integrate with client

tjsharp1:
Add a new sample indexer module, take log events and send them through the system.
Begginings of log handler,


fuel-indexer/indexer/src/bin/service/main.rs
```diff
@@ -0,0 +1,77 @@
...
+    local: bool,
+    #[structopt(parse(from_os_str), help = "Indexer service config file")]
+    config: PathBuf,
+    #[structopt(short, long, parse(from_os_str), help = "Indexer service config file")]
+    test_manifest: Option<PathBuf>,
+}
+
+#[tokio::main]
+pub async fn main() -> Result<()> {
+    env_logger::init();
```
Voxelot, line 26: We prefer using tracing instead of env_logger: https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/logging.md

Voxelot:
@tjsharp1 are we still waiting on making indexing transactional at the database level for this PR?

adlerjohn (commented):
Merge conflict!

fuel-indexer/indexer/src/api.rs
```diff
@@ -0,0 +1,127 @@
...
+use async_std::sync::{Arc, RwLock};
+use fuel_indexer_schema::db::{
+    graphql::{GraphqlError, GraphqlQueryBuilder},
+    tables::Schema,
+};
+use serde::Deserialize;
+use thiserror::Error;
+use tokio_postgres::{connect, types::Type, NoTls};
+use tracing::error;
+use warp::{http::StatusCode, reply::with_status, Filter, Reply};
```
Voxelot, line 11: Ideally, we could use the same server stack across both fuel-core and fuel-indexer. We're currently using axum in other places.

Voxelot: This would make it really easy to combine API handlers from both projects into a single webserver as well.

fuel-indexer/indexer/src/api.rs
```diff
@@ -0,0 +1,127 @@
...
+        },
+        Err(e) => Ok(with_status(
+            format!("The graph {} was not found ({:?})", name, e),
+            StatusCode::NOT_FOUND,
+        )
+        .into_response()),
+    }
+}
+
+pub struct GraphQlAPI;
```
Voxelot, line 66: Nit: rust naming conventions says we shouldn't uppercase all letters in an acronym. 
https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case

```suggestion
pub struct GraphQlApi;
```

fuel-indexer/indexer/src/bin/service/main.rs
```diff
@@ -0,0 +1,86 @@
+use anyhow::Result;
```
Voxelot, line 1: I think it'd be cleaner to just call this whole file `service.rs` and eliminate the container directory. Then we also wouldn't need the custom `[[bin]]` targets in the cargo.toml.

fuel-indexer/indexer/src/bin/service/main.rs
```diff
@@ -0,0 +1,86 @@
...
+        let mut bytes = Vec::<u8>::new();
+        file.read_to_end(&mut bytes).await?;
+        service.add_indexer(manifest, &schema, bytes, false)?;
+    }
+
+    let service_handle = tokio::spawn(service.run());
+
+    let (first, second) = join!(api_handle, service_handle);
+
+    info!("Exiting.... {first:?} {second:?}");
```
Voxelot, line 84: Users probably only care about the result of these futures if there's an error to report:

```suggestion
    if let Err(e) = first {
        error!("{:?}", e)
    }
    if let Err(e) = second {
        error!("{:?}", e)
    }
```

fuel-indexer/indexer/src/database.rs
```diff
@@ -86,56 +73,73 @@ impl SchemaManager {
...
+    }
+
+    fn upsert_query(
+        &self,
+        table: &str,
+        columns: &[String],
+        inserts: Vec<String>,
+        updates: Vec<String>,
+    ) -> String {
         format!(
```
Voxelot, line 131: Do we have sanitization concerns when constructing queries this way?

fuel-indexer/indexer/src/executor.rs
```diff
@@ -98,31 +115,39 @@ impl IndexExecutor {
...
-
-                unsafe {
-                    // Safety: the alloc call for wasm_mem has succeeded for bytes.len()
-                    //         so we have this block of memory for copying. The fun.call() below
-                    //         will release it.
-                    memory.data_unchecked_mut()[range].copy_from_slice(&bytes);
+                if let Ok(db) = self.db.lock() {
+                    db.start_transaction()?;
+                } else {
+                    return Err(IndexerError::LockPoisoned);
```
Voxelot, line 135: Given that a poisoned lock is likely an unrecoverable error and each subsequent request to this database lock for the lifetime of this program would fail, this is one of the few scenarios where I think an `.expect()` is appropriate vs an error result.

fuel-indexer/indexer/src/manifest.rs
```diff
@@ -5,6 +5,7 @@ pub struct Manifest {
     pub namespace: String,
     pub graphql_schema: String,
     pub wasm_module: String,
+    pub start_block: Option<u64>,
     pub handlers: Vec<Handler>,
     pub test_events: Vec<Event>,
 }
```
Voxelot, line 11: Shouldn't we also subscribe these handlers to a particular contract id?

fuel-indexer/indexer/src/service.rs
```diff
@@ -0,0 +1,197 @@
...
+                    break;
+                }
+            }
+        }
+    }
+
+    pub async fn run(self) {
+        let IndexerService { handles, .. } = self;
+        let mut futs = FuturesUnordered::from_iter(handles.into_values());
+        //loop {
```
Voxelot, line 190:
```suggestion
```

fuel-indexer/indexer/src/service.rs
```diff
@@ -0,0 +1,197 @@
...
+        while let Some(fut) = futs.next().await {
+            info!("Retired a future {fut:?}");
+        }
+        // TODO: ...
+        //}
```
Voxelot, line 195:
```suggestion
```

fuel-indexer/indexer/src/database.rs
tjsharp1: Yeah, this one sort of has to be done like this. These are the tables generated by the user's graphql, so we don't have the luxury of auto-generated rust types here.

fuel-indexer/indexer/src/manifest.rs
tjsharp1: Yeah, definitely want to do that. I was thinking I'd do it once we get event handling. Right now I'm sending in tests that have contract id set to 0.

Voxelot (approved):

PR #145, opened 2022-01-19
initial discovery

leviathanbeak:
current overview of p2p service:
- only committed work on Discovery
- gossipsub and peer_info incoming


fuel-p2p/Cargo.toml
```diff
@@ -13,4 +13,6 @@ log = "0.4.8"
 sha2 = "0.9"
 
 [dev-dependencies]
+env_logger = "0.9"
 rand = "0.8.4"
+tokio = {version = "1.8", features = ["full"]}
```
adlerjohn, line 18:
```suggestion
tokio = { version = "1.8", features = ["full"] }
```

fuel-p2p/src/behavior.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// message propagation for p2p
+    gossipsub: Gossipsub,
+
+    /// node discovery
+    discovery: DiscoveryBehaviour,
+
+    //peer info let's do it later
+    //identify: Identify,
+    #[behaviour(ignore)]
+    events: VecDeque<FuelBehaviourEvent>,
```
melekes, line 43: q: should this be unbounded or a reasonable limit should be enforced? (i.e. how many messages each peer can send us)

fuel-p2p/src/discovery.rs
```diff
@@ -0,0 +1,498 @@
...
+
+    /// Kademlia with MemoryStore
+    kademlia: Kademlia<MemoryStore>,
+
+    /// If enabled, the Stream that will fire after the delay expires,
+    /// starting new random walk
+    next_kad_random_walk: Option<Delay>,
+
+    /// The Duration for the next random walk, after the current one ends
+    duration_to_next_kad: Duration,
```
melekes, line 70: did you mean timeout or this is a limit for each random walk? if the latter, then why not simply name it `kad_random_walk_duration`?

fuel-p2p/src/discovery.rs
```diff
@@ -0,0 +1,498 @@
...
+
+    /// Track the count of peers connected
+    connected_peers_count: u64,
+
+    /// Maximum amount of allowed peers
+    max_peers_connected: u64,
```
melekes, line 76: can this change? if not, shouldn't this be a part of p2p config?

fuel-p2p/src/discovery.rs
```diff
@@ -0,0 +1,498 @@
...
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>();
+
+        {
+            let mut list_to_filter = Vec::new();
+
+            list_to_filter.extend(self.kademlia.addresses_of_peer(peer_id));
```
melekes, line 235: can't we just extend `list` here?

fuel-p2p/src/discovery.rs
leviathanbeak: it's not a limit, it is what the name suggests: duration to next kademlia, once the current kademlia walk ends, the duration of the incoming one

fuel-p2p/src/discovery.rs
leviathanbeak: it will be passed through the config, currently config is used only for creation of the behaviours, rather then referencing configs in each iteration - we pass the values to the behaviours themselves

fuel-p2p/src/discovery.rs
leviathanbeak: no, I need to filter that list first - thanks for reminding me tho 👍

fuel-p2p/src/discovery.rs
```diff
@@ -0,0 +1,498 @@
...
+    Disconnected(PeerId),
+}
+
+/// NetworkBehavior for discovery of nodes
+pub struct DiscoveryBehaviour {
+    /// Store the local peer id
+    local_peer_id: PeerId,
+
+    /// Store Peer IDs of currently connected peers
+    connected_peers: HashSet<PeerId>,
```
melekes, line 48: why don't just have `connected_peer_addresses`? less memory, less chance for mistakes (when you update `connected_peer_addresses`, but forgot to update `connected_peers`).

leviathanbeak: good point, tho I updated it a bit, so they now serve a different purpose, `peer_addresses` will hold *any* peers, while the HasheSet will hold IDs of *only* connected peers.

Having said that, it's a work in progress, depending on the use-cases when other behaviours are defined, they might get updated or removed completely.

fuel-p2p/Cargo.toml
leviathanbeak: ah so my toml formatter removes that extra space between - but I see others are not removing it - so I need to comply :)

PR #156, opened 2022-02-03
Slack Notifications for Worfklow Job Failures

rfuelsh:
For all jobs in the ci.yml, slack notifications will be sent if the job fails at any point.

For the build job, a slack alert will only be sent if the branch is master upon which the workflow is running.

This was tested and the results are in #notify-build of our slack.


.github/workflows/ci.yml
```diff
@@ -62,6 +62,19 @@ jobs:
       with:
         command: test
         args: --verbose --all-targets
+    
+    - name: Notify if Job Fails
+      uses: ravsamhq/notify-slack-action@v1
+      if: always() && github.ref == 'refs/heads/master'
```
adlerjohn, line 68: Is the `always() &&` here actually needed?
```suggestion
      if: github.ref == 'refs/heads/master'
```

.github/workflows/ci.yml
```diff
@@ -92,6 +105,19 @@ jobs:
         with:
           publish-delay: 30000
           registry-token: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+      
+      - name: Notify if Job Fails
+        uses: ravsamhq/notify-slack-action@v1
+        if: always()
```
adlerjohn, line 111: Ditto
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -138,7 +164,20 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           cache-from: type=gha
           cache-to: type=gha,mode=max  
-  
+      
+      - name: Notify if Job Fails
+        uses: ravsamhq/notify-slack-action@v1
+        if: always()
```
adlerjohn, line 170: Ditto
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -180,3 +219,16 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
+      
+      - name: Notify if Job Fails
+        uses: ravsamhq/notify-slack-action@v1
+        if: always()
```
adlerjohn, line 225: Ditto
```suggestion
```

.github/workflows/ci.yml
rfuelsh: @adlerjohn - the always() is needed is there is only failures in the prior steps so the slack notification step is run in the job (if the past steps of a job successful - it does not trigger a slack notification) 

[
](https://docs.github.com/en/actions/learn-github-actions/expressions#:~:text=succeeded%0A%20%20%20%20if%3A%20%24%7B%7B%20success()%20%7D%7D-,always,the%20task%20from%20running.%20For%20example%2C%20if%20getting%20sources%20failed.,-Example)

I tested the recommendation you suggested but it did not trigger a notify after a prior step failed (on the build job and for my branch) as the always() was missing

The alternative is to try "failure()" is we do not want to use "always()" - which I can try

https://docs.github.com/en/actions/learn-github-actions/expressions#:~:text=if%3A%20%24%7B%7B%20cancelled()%20%7D%7D-,failure,dependent%20jobs%2C%20failure()%20returns%20true%20if%20any%20ancestor%20job%20fails.,-Example

adlerjohn: I see. Okay.

adlerjohn (commented):
Looks okay to me I guess, but will defer to @Voxelot

Voxelot (approved):
lgtm, although I think we should really try to use action templates before copying these around to all the other repos. That way we have one place to control all the notifications.

PR #173, opened 2022-02-12
Add GraphQL Account type

AlicanC:
This PR basically tries to do two things:

- Add `Account` type
- Change fields that represent an Account (like `Block.producer`) from `HexString256` to `Account`

---

The Account type can be extended with fields like `balances` and `transactions` so this:
```gql
query {
  balances(address: $address) { ... }
  transactions(address: $address) { ... }
}
```
can become this:
```gql
query {
  account(address: $address) {
    balances { ... }
    transactions { ... }
  }
}
```

---

The change from `HexString256` to `Account` is so we can pack requirements of a UI into one query:
```ts
// Before
let block = await q(`query {
  block(height: ${height}) {
    producer
  }
}`);

let producer = await q(`query {
  account(id: ${block.producer}) {
    address
    producedBlocks { count }
  }
}`);

console.log(`${height} produced by ${producer.address} who produced ${producer.producedBlocks.count} so far!`)

// After
let block = await q(`query {
  block(height: ${height}) {
    producer {
      address
      producedBlocks { count }
    }
  }
}`);

let producer = block.producer;

console.log(`${height} produced by ${producer.address} who produced ${producer.producedBlocks.count} so far!`)
```

---

Given the green light, I will:
- [x] Change `HexString256` to `Account` in all fields necessary
- [x] Add `Account.transactions`
- [x] Add `fuel-client` implementation and tests

### DIFF
--- a/fuel-client/assets/schema.sdl
+++ b/fuel-client/assets/schema.sdl
@@ -1,9 +1,12 @@
+type Account {
+	address: HexString256!
+}
 type Block {
 	id: HexString256!
 	height: U64!
 	transactions: [Transaction!]!
 	time: DateTime!
-	producer: HexString256!
+	producer: Account!
 }
 type BlockConnection {
 	"""
@@ -188,6 +191,7 @@ type Query {
 	coin(utxoId: HexStringUtxoId!): Coin
 	coins(after: String, before: String, first: Int, last: Int, filter: CoinFilterInput!): CoinConnection!
 	coinsToSpend(owner: HexString256!, spendQuery: [SpendQueryElementInput!]!, maxInputs: Int): [Coin!]!
+	account(address: HexString256!): Account
 }
 type Receipt {
 	id: HexString256
--- a/fuel-client/assets/schema.sdl
+++ b/fuel-client/assets/schema.sdl
@@ -1,5 +1,6 @@
 type Account {
 	address: HexString256!
+	transactions(first: Int, after: String, last: Int, before: String): TransactionConnection!
 }
 type Block {
 	id: HexString256!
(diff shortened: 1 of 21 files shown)


AlicanC:
@Voxelot what do you think, should I proceed? Any comments?

Voxelot:
Yes this is a much cleaner approach. Eventually, I would like to avoid exposing any of the graphql schema types directly via fuel-gql-client, and only use the proper corresponding fuel-types or other types that are more rust friendly.

AlicanC:
@Voxelot ready for review.

One note is that I did not use `Account.transactions` in fuel-gql-client since `Query.transactionsByOwner` already does the trick. (Also not planning to use it on fuels-ts, but only on Explorer/Wallet.) For that reason I wrote the test for it in `fuel-core/src/schema/account.rs` and not `fuel-client/src/client/schema/account.rs`.

Another note is that there's a lot of code duplication between `Account.transactions` and `Query.transactionsByOwner`. I believe a huge amount of it would be solved if we abstracted our "gql args -> db iterator -> gql connection" logic. I need to see more connections before I feel comfortable doing that so I left it.

adlerjohn:
Bikeshed time: instead of `Account`, which @Voxelot has pointed out might be confused with the notion of an Ethereum account, we could use a different word like `Statement` (like a bank statement)?

AlicanC:
@rakita the problem mentioned in the OP still exists.

Looks like Sway now has `Identity` which we could use instead of `Account`: https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/identity.sw

Voxelot:
closing since this is falling really out of date and we don't really need this abstraction

PR #177, opened 2022-02-17
Receipt saving bug

Voxelot:
This addresses a bug where receipts aren't saved if the transaction panics. They were being saved to the wrong database transaction which gets dropped if the tx panics, and should only be used for state changes from the script execution.


vnepveu (approved):
Makes sense! Testing it locally.

vnepveu:
Fix the block typo, and testing it locally is a success! thanks a lot

Voxelot:
ah thanks for the fix - I was just waking up

PR #186, opened 2022-02-23
Malleable Blocks Part 2

Voxelot:
Followup PR to #180 to address some feedback there.

Also found some issues with how block id's were handled, so I changed the execution flow a bit to only generate a block id after the MST is finalized.

### DIFF
--- a/fuel-core/src/executor.rs
+++ b/fuel-core/src/executor.rs
@@ -3,7 +3,7 @@ use crate::{
     database::{transaction::TransactionIndex, Database, KvStoreError},
     model::{
         coin::{Coin, CoinStatus},
-        fuel_block::{BlockHeight, FuelBlockFull, FuelBlockLight},
+        fuel_block::{BlockHeight, FuelBlock, FuelBlockDb},
     },
     service::Config,
     tx_pool::TransactionStatus,
@@ -12,6 +12,7 @@ use fuel_asm::Word;
 use fuel_storage::Storage;
 use fuel_tx::crypto::Hasher;
 use fuel_tx::{Address, Bytes32, Color, Input, Output, Receipt, Transaction, UtxoId};
+use fuel_types::bytes::{SerializableVec, SizedBytes};
 use fuel_types::ContractId;
 use fuel_vm::prelude::{Backtrace, Interpreter};
 use fuel_vm::{consts::REG_SP, prelude::Backtrace as FuelBacktrace};
@@ -39,11 +40,7 @@ pub struct Executor {
 }
 
 impl Executor {
-    pub async fn execute(
-        &self,
-        block: &mut FuelBlockFull,
-        mode: ExecutionMode,
-    ) -> Result<(), Error> {
+    pub async fn execute(&self, block: &mut FuelBlock, mode: ExecutionMode) -> Result<(), Error> {
         let block_id = block.id();
         let mut block_db_transaction = self.database.transaction();
 
@@ -103,20 +100,23 @@ impl Executor {
 
             // update block commitment
             let tx_fee = self.total_fee_paid(tx, vm_result.receipts())?;
-            // TODO: use SMT instead of this manual approach
+            // TODO: use merkle-sum-tree instead of this manual hashing approach
             commitment.sum = commitment
                 .sum
                 .checked_add(tx_fee)
                 .ok_or(Error::FeeOverflow)?;
-            commitment.root = Hasher::hash(
-                &commitment
-                    .root
-                    .as_ref()
-                    .iter()
-                    .chain(tx_id.as_ref().iter())
-                    .copied()
-                    .collect_vec(),
-            );
+
+            // include the canonical serialization of the malleated tx into the commitment,
+            // including all witness data.
+            //
+            // TODO: reference the bytes directly from VM memory to save serialization. This isn't
+            //       possible atm because the change output values are set on the tx instance in the vm
+            //       and not also on the in-memory representation of the tx.
+            let tx_bytes = vm_result.tx().clone().to_bytes();
+            commitment.root = Hasher::default()
+                .chain(commitment.root.as_ref().iter())
+                .chain(tx_bytes.iter())
+                .digest();
 
             match mode {
                 ExecutionMode::Validation => {
@@ -210,10 +210,10 @@ impl Executor {
         }
 
         // insert block into database
-        Storage::<Bytes32, FuelBlockLight>::insert(
+        Storage::<Bytes32, FuelBlockDb>::insert(
             block_db_transaction.deref_mut(),
             &block_id,
-            &block.as_light(),
+            &block.to_db_block(),
         )?;
         block_db_transaction.commit()?;
         Ok(())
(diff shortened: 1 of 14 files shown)


fuel-core/src/executor.rs
```diff
@@ -103,20 +105,23 @@ impl Executor {
 
             // update block commitment
             let tx_fee = self.total_fee_paid(tx, vm_result.receipts())?;
-            // TODO: use SMT instead of this manual approach
+            // TODO: use merkle-sum-tree instead of this manual hashing approach
```
adlerjohn, line 108: This should just be a plain binary Merkle tree. I will update the block header specs soon for this.

Voxelot: Ah ok, well that should be an easy change.

fuel-core/src/model/fuel_block.rs
```diff
@@ -69,72 +67,84 @@ impl BlockHeight {
...
-}
-
 #[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct FuelBlockHeaders {
-    pub fuel_height: BlockHeight,
+pub struct FuelBlockHeader {
+    /// Fuel block height.
+    pub height: BlockHeight,
+    /// Ethereum block number.
+    pub number: BlockHeight,
```
rakita, line 75: This is not something we know at moment of block creation, and we cant predict at what exact eth block is fuel block going to be commited. But it makes sense to be included as a field as it is going to be updated when fuel blocks get commited over eth. I would put this as `Option` or maybe zeroed value, and it needs to be removed from `id` hash calculation.

fuel-core/src/model/fuel_block.rs
```diff
@@ -69,72 +67,84 @@ impl BlockHeight {
...
+    /// Block header hash of the previous block.
+    pub prev_hash: Bytes32,
```
rakita, line 77: usualy called `parent` or `parent_hash`, but still `prev_hash` is fine name

fuel-core/src/model/fuel_block.rs
```diff
@@ -69,72 +67,84 @@ impl BlockHeight {
...
+    /// Merkle root of all previous block header hashes.
+    pub prev_root: Bytes32,
```
rakita, line 79: Not exactly sure why is this needed. Shouldn't `prev_hash` be enought to connect the chain?

fuel-core/src/executor.rs
```diff
@@ -169,51 +169,72 @@ impl Executor {
...
             }
             ExecutionMode::Validation => {
-                if block.headers.transactions_commitment != commitment {
+                if block.header.transactions_root != txs_root {
                     return Err(Error::InvalidBlockCommitment);
                 }
             }
         }
+        // set the coinbase amount
+        block.header.coinbase = coinbase;
```
rakita, line 216: This probably still needs to be defines, but i think Nick mentioned long ago that fuel is going to be deflationary so in that sense `coinbase` needs to be increase by some fixed amount to reward validator and we will need to check under `ExecutionMode::Validation` if that amount is okay or not.

Voxelot: Good catch

fuel-core/src/model/fuel_block.rs
Voxelot: It seems hard to avoid non-deterministic behavior between validators if we don't commit to the ethereum global state as of this block. The block producer should be able to pick an ethereum block to claim that they are including all the ethereum events up to that point.

cc @adlerjohn

fuel-core/src/executor.rs
Voxelot: Although it may still be useful to capture the total fees of the block, even if it doesn't all go directly to the validator and some get burned.

fuel-core/src/model/fuel_block.rs
adlerjohn: The hash of the previous block header is sufficient, but if it's a MMR of all previous block header hashes then you can do log-sized proofs to past block headers rather than linear.

fuel-core/src/executor.rs
adlerjohn: I've been thinking of having a third coinbase tx type that's 0-input, 1-output. That way there's an explicit TXO for the coinbase rather than some implicit state mutation like with Ethereum.

fuel-core/src/model/fuel_block.rs
rakita: I see, this is not something needed for client as we need to have all blocks in db for rebroadcasting, but for v2 contract?  But for even v2, wouldn't it be enought to have map of index->hash and there wouldn't be need MMR hashing and checking. Just Wondering, yes we have log-sized proof but where is it needed to have that reduction?

fuel-core/src/executor.rs
rakita: > I've been thinking of having a third coinbase tx type that's 0-input, 1-output. That way there's an explicit TXO for the coinbase rather than some implicit state mutation like with Ethereum.

This is similar problem with token deposits, we need to make token out of thin air in fuel :). Maybe having additional Input that will cover those two cases?

adlerjohn: There's an issue to track deposit IDs here: https://github.com/FuelLabs/fuel-specs/issues/106

fuel-core/src/model/fuel_block.rs
adlerjohn: Storing every block header hash would mean a storage write to a new slot for every block. If we use an MMR, then we only need to keep track of the tip's header hash, which is much cheaper.

fuel-core/src/model/fuel_block.rs
Voxelot: I was thinking this would be L1 block number used for synchronizing deposits, not the L1 block that this block ended up being published in.

adlerjohn: That's a possibility, yes!

adlerjohn: I'm okay with leaving a field in the header for Ethereum block, but there should be a description here that there's some caveats, and linking to https://github.com/FuelLabs/fuel-specs/issues/220

fuel-core/src/model/fuel_block.rs
Voxelot: I like parent hash as well

fuel-core/src/executor.rs
Voxelot: I've removed the coinbase field and will just `debug` log the fee amounts for now.

fuel-core/src/model/fuel_block.rs
Voxelot: Which of our MT implementations supports MMR?

adlerjohn: The binary Merkle tree is actually an MMR 🤷

fuel-core/src/model/fuel_block.rs
Voxelot: updated the description of this field, lmk if that works for you.

adlerjohn: Good enough!

adlerjohn (commented):
No further comments from me, will defer to @rakita

fuel-core/src/model/fuel_block.rs
Voxelot: I think I'll defer this to a separate task

Voxelot: https://github.com/FuelLabs/fuel-core/issues/192

PR #191, opened 2022-03-01
Restructure Schema

### LINKED ISSUE #161: Use `U64` custom scalar instead of Word/i32 in GraphQL Schema
Since JSON / Javascript numbers don't support precise u64, even though the graphql schema defined on the server side uses `Word` for fields like gas_limit or gas_price, the fuel-gql-client parses them as `i32` which for certain fields could be a lossy encoding. 

There is already a custom scalar defined called `U64` in the schema which will preserve the numeric precision, but there are still many places that should be updated to use it in the schema.

AlicanC:
- Rename types: `HexString256` -> `Bytes32`, `HexStringUtxoId` -> `UtxoId`
- Add new scalars: `Address`, `BlockId`, `Color`, `Salt`, `TransactionId`
- Add new object: `Contract`. Currently this only has an `id` field since we can't fetch the rest yet.
- Refine fields with `HexString256` types: `WithdrawalOutput.to: HexString256` -> `...: Address`, etc.
- Rename and retype id fields: `SuccessStatus.block_id: HexString256` -> `SuccessStatus.block: Block`, etc.
- Reorder connection args so `first` and `after` comes first. They map nicely to traditional limit/offset and more widely used.
- Resolves #161

Note: This PR doesn't add `Account`. Leaving that on #173.

### DIFF
--- a/fuel-client/assets/schema.sdl
+++ b/fuel-client/assets/schema.sdl
@@ -96,7 +96,7 @@ enum CoinStatus {
 }
 type Contract {
 	id: ContractId!
-	bytes: HexString!
+	bytecode: HexString!
 }
 type ContractCreated {
 	contract: Contract!
--- a/fuel-client/assets/schema.sdl
+++ b/fuel-client/assets/schema.sdl
@@ -1,9 +1,11 @@
+scalar Address
+scalar AssetId
 type Block {
-	id: HexString256!
+	id: BlockId!
 	height: U64!
 	transactions: [Transaction!]!
 	time: DateTime!
-	producer: HexString256!
+	producer: Address!
 }
 type BlockConnection {
 	"""
@@ -28,6 +30,8 @@ type BlockEdge {
 	"""
 	cursor: String!
 }
+scalar BlockId
+scalar Bytes32
 type ChainInfo {
 	name: String!
 	latestBlock: Block!
@@ -35,15 +39,15 @@ type ChainInfo {
 	peerCount: Int!
 }
 type ChangeOutput {
-	to: HexString256!
-	amount: Int!
-	assetId: HexString256!
+	to: Address!
+	amount: U64!
+	assetId: AssetId!
 }
 type Coin {
-	utxoId: HexStringUtxoId!
-	owner: HexString256!
+	utxoId: UtxoId!
+	owner: Address!
 	amount: U64!
-	assetId: HexString256!
+	assetId: AssetId!
 	maturity: U64!
 	status: CoinStatus!
 	blockCreated: U64!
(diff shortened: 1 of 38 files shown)


fuel-core/src/model.rs
```diff
@@ -1,4 +1,5 @@
 pub mod coin;
+pub mod contract;
```
Voxelot, line 2: This isn't used anywhere and should be removed.

fuel-core/src/model/contract.rs
```diff
@@ -0,0 +1,10 @@
+use super::Hash;
+use fuel_tx::ContractId;
+
+#[derive(Debug, Copy, Clone)]
+pub struct Contract {
```
Voxelot, line 5: This is unused, and I think these fields like utxo_id will be fetched on the fly from the grapqhl resolver in other ways.

Voxelot:
This could be deferred to a different PR, but eventually, I'd like to have all the fuel-gql-client schema types to be `pub(crate)` and only expose Rust API consumer friendly mirrored versions from `fuel-gql-client/src/client/types`. For example, enums and unions have strange constraints via the graphql libraries, and aren't very Rust friendly.

AlicanC:
> [...], I'd like to have all the fuel-gql-client schema types be `pub(crate)` and only expose Rust API consumer friendly mirrored versions from `fuel-gql-client/src/client/types`. For example, enums and unions have strange constraints via the graphql libraries, and aren't very Rust friendly.

Agreed and yes, I will leave this for later so I can focus on SwaySwap issues that came up.

I was going to mention this since I realized I misinformed @digorithm on how I didn't make any changes to fuel-gql-client so fuels-rs shouldn't be affected. I'm actually not sure since some APIs do return schema types and that could affect fuels-rs.

I believe the changes required will be a bunch of `obj.block_id` to `obj.block.id` changes anyways, but I have one thing to verify and then I'll ask for reviews.

fuel-core/src/model/contract.rs
AlicanC: Seems like I badly resolved a merged conflict. Will take care of it.

AlicanC:
Rebased for AssetId and ready for review.

adlerjohn:
Confirmed that as of [842356f](https://github.com/FuelLabs/fuel-core/pull/191/commits/842356f403b4f6291521c10be062be0498de4177) there's no instance of `color` in the code.

fuel-client/assets/schema.sdl
```diff
@@ -75,29 +79,34 @@ input CoinFilterInput {
...
+	amount: U64!
+	assetId: AssetId!
 }
 enum CoinStatus {
 	UNSPENT
 	SPENT
 }
+type Contract {
+	id: ContractId!
+	bytes: HexString!
```
Voxelot, line 99: I think bytecode would be more clear than just bytes?

fuel-client/src/client/schema/contract.rs
```diff
@@ -0,0 +1,43 @@
...
+pub struct ContractByIdQuery {
+    #[arguments(id = &args.id)]
+    pub contract: Option<Contract>,
+}
+
+#[derive(cynic::QueryFragment, Debug)]
+#[cynic(schema_path = "./assets/schema.sdl")]
+pub struct Contract {
+    pub id: ContractId,
+}
```
Voxelot, line 23: [nit] also include the bytecode, otherwise, this is no different than the ContractIdFragment type.

```suggestion
    pub bytecode: HexString,
}
```

Voxelot (approved):

fuel-client/src/client/schema/contract.rs
AlicanC: I left `bytecode` out intentionally because I'm try to avoid making methods like `client.contract(id)` too fat, which of course made both fragments the same.

ATM the `Contract` fragment is basically scaffolding for when we add stuff like `created_block_id`, `last_touched_block_id`, `utxo_id`, etc.

PR #197, opened 2022-03-04
Bump to v0.4.0

adlerjohn:
Depends on
- [x] #191
- [x] #198
- [x] #199


AlicanC:
> Do we want to wait until #191 is in before bumping?

Would be nice since it's also breaking. I don't have anything else to add to that PR, but not sure where we're at with the reviews.

Voxelot:
I found a bug in rocksdb, and in order to prevent it from happening again through CI I had to make a pretty major structural change to the integ tests in #199.

Voxelot (approved):

Voxelot:
Oh now that Roy's PR is merged well also need to update the version on the helm chart.

PR #199, opened 2022-03-05
Fix RocksDb issue, ensure rocksdb runs in CI

Voxelot:
After adding a new column to the database, I forgot to increment the column number constant which was causing errors at runtime.

To avoid this from happening again, I made it so that the tests run twice, once with default features enabled (rocksdb) and again with default features disabled (in-memory). 

However, due to the way our integ tests in fuel-gql-client were setup, it was causing a bug in the v2 feature resolver when attempting to use `cargo test --no-default-features`. Specifically, it seemed like the feature resolver was unifying dev-deps in the fuel-gql-client with the fuel-core crate in the workspace, causing fuel-core to always be compiled with default features enabled when using `--no-default-features` at the workspace level. In order to fix this and make it possible to run tests without default features enabled on fuel-core, I had to move all the integ tests into their own crate (with normal deps only). This is better in the long run and is more similar to how projects like diesel-rs are structured.


Cargo.toml
```diff
@@ -12,6 +12,7 @@ members = [
   "fuel-txpool",
   "fuel-p2p",
   "xtask",
+  "fuel-tests",
```
adlerjohn, line 15: Sort alphabetically?

fuel-client/Cargo.toml
```diff
@@ -37,22 +32,13 @@ surf = "2.2"
...
 
 [features]
-default = []
 test-helpers = []
 
-[[test]]
-name = "integration_tests"
-path = "tests/lib.rs"
-harness = true
+
```
adlerjohn, line 44: Nit: remove extra newline

fuel-tests/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "fuel-tests"
+version = "0.1.0"
```
adlerjohn, line 3: Since it's not being published
```suggestion
version = "0.0.0"
```

fuel-tests/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+edition = "2021"
+license = "BUSL-1.1"
+publish = false
+autotests = false
+autobenches = false
```
adlerjohn, line 8: Can you add the comment to these two fields if it's still relevant?

Voxelot: done

fuel-core/src/database.rs
```diff
@@ -149,11 +197,33 @@ impl AsRef<Database> for Database {
     }
 }
 
-/// Construct an in-memory database
+/// Construct an ephemeral database
+/// uses rocksdb when rocksdb features are enabled
+/// uses in-memory when rocksdb features are disabled
```
adlerjohn, line 202: Hmmmmm. Is this what we want? Even if the `rocksdb` feature is enabled, should it use in-memory if `--db-type in-memory` is set from the command line?

Voxelot: ah good catch, I need to use `::in_memory()` instead

Voxelot: Should be fixed now, but I changed it in the `fuel_core::service::FuelService::new_node` method.

adlerjohn: Should the comment be updated?

Voxelot: No this isn't on the cli args / config side, this is the default constructor for the db. This way we can use one method to test both in-memory and rocksdb without needing to put feature flags in every test.

adlerjohn (approved):
LGTM, though I would prefer a second set of eyes (maybe @rakita or @AlicanC)

AlicanC (approved):
I upgraded fuels-ts, block-explorer and swayswap-demo from 0.3.2 to the head of this branch and all look fine.

PR #200, opened 2022-03-05
Gossipsub implementation for p2p

leviathanbeak:
initial gossipsub implementation

### DIFF
--- a/fuel-p2p/src/behavior.rs
+++ b/fuel-p2p/src/behavior.rs
@@ -14,6 +14,7 @@ use std::{
     collections::VecDeque,
     task::{Context, Poll},
 };
+use tracing::debug;
 
 // todo: define which events outside world is intersted in
 #[derive(Debug)]
--- a/fuel-p2p/src/behavior.rs
+++ b/fuel-p2p/src/behavior.rs
@@ -1,9 +1,15 @@
 use crate::{
     config::P2PConfig,
     discovery::{DiscoveryBehaviour, DiscoveryConfig, DiscoveryEvent},
+    gossipsub,
     peer_info::{PeerInfo, PeerInfoBehaviour, PeerInfoEvent},
+    service::GossipTopic,
 };
 use libp2p::{
+    gossipsub::{
+        error::{PublishError, SubscriptionError},
+        Gossipsub, GossipsubEvent, MessageId, TopicHash,
+    },
     identity::Keypair,
     swarm::{
         NetworkBehaviour, NetworkBehaviourAction, NetworkBehaviourEventProcess, PollParameters,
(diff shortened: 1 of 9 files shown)


fuel-p2p/src/behavior.rs
```diff
@@ -134,3 +168,20 @@ impl NetworkBehaviourEventProcess<PeerInfoEvent> for FuelBehaviour {
         }
     }
 }
+
+impl NetworkBehaviourEventProcess<GossipsubEvent> for FuelBehaviour {
```
adlerjohn, line 172: Can we use American English? Defer to a future PR to avoid polluting this one.

fuel-p2p/src/gossipsub/builder.rs
```diff
@@ -0,0 +1,65 @@
...
+
+pub fn build_gossipsub(local_key: &Keypair) -> Gossipsub {
+    let gossip_message_id =
+        move |message: &GossipsubMessage| MessageId::from(&Sha256::digest(&message.data)[..20]);
+
+    let fast_gossip_message_id = move |message: &RawGossipsubMessage| {
+        FastMessageId::from(&Sha256::digest(&message.data)[..8])
+    };
+
+    let gossipsub_config = GossipsubConfigBuilder::default()
```
adlerjohn, line 19: These are a lot of options 😂

1. Is there some sort of rationale for the parameter choice?
2. Is there a better way to do this?
3. If there's no functional ordering requirement, can these to ordered alphabetically?

fuel-p2p/src/behavior.rs
leviathanbeak: nope since that's what we use from libp2p, not my choice 🇬🇧

fuel-p2p/src/gossipsub/builder.rs
leviathanbeak: in short, we probably won't need many of those, so they will be removed in final code (they are set by default) 

those that we need, we can order them nicely, that was completely arbitrary

fuel-p2p/src/behavior.rs
adlerjohn: 🤷

leviathanbeak: I mean, we could name our own FuelBehavior, but it would be weird since it would still need to implement libp2p's Behaviour

fuel-p2p/src/gossipsub/builder.rs
Voxelot: We should have some kind of network configuration object we can pass into this `build_gossipsub` method. Then we can put all these defaults there and override them using cli args or service config later.

leviathanbeak: yes they will be part of the general p2p config, I will update it

Voxelot (approved):

PR #207, opened 2022-03-10
Relayer skeleton impl

rakita:


adlerjohn:
Looks like still in progress, so converting to draft.

fuel-core-interfaces/src/block_importer.rs
```diff
@@ -0,0 +1,11 @@
+/// Currently just placeholder for new block included and new block created events.
+#[derive(Clone, Debug)]
+pub enum NewBlockEvent {
+    /// send this to eth
+    NewBlockCreated(u64),
+    NewBlockIncluded {
+        height: u64,
+        /// height where we are finalizing stake and token deposits.
+        eth_height: u64,
```
adlerjohn, line 9: If this isn't specific to Ethereum, but is rather generic to some L1, then it shouldn't use `eth` in the name. Is there a better name for some sort of generic data availability layer? Maybe `da_height`?

fuel-core-interfaces/src/db.rs
```diff
@@ -577,4 +606,159 @@ pub mod helpers {
             Ok(self.contract.get(&contract_id).is_some())
         }
     }
+
+    /* RELAYER needs
+    Storage<Bytes32, DepositCoin, Error = KvStoreError> // token deposit
+    Storage<Address, u64,Error = KvStoreError> // validator set
+    Storage<u64, HashMap<Address, u64>,Error = KvStoreError> // validator set diff
+    */
```
adlerjohn, line 614: Is this a todo, or just a note?

fuel-relayer/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+description = "Fuel Relayer"
+
+[dependencies]
+anyhow = "1.0"
+async-trait = "0.1"
+bytes = "1.1"
+env_logger = "0.9"
+ethers-core = { git = "https://github.com/rakita/ethers-rs.git", branch = "master", package = "ethers-core", default-features = false }
+ethers-middleware = { git = "https://github.com/rakita/ethers-rs.git", branch = "master", package = "ethers-middleware", default-features = false }
+ethers-providers = { git = "https://github.com/rakita/ethers-rs.git", branch = "master", package = "ethers-providers", default-features = false, features = [
```
adlerjohn, line 16: To publish this crate to crates.io, we need all dependencies to be releases on crates.io. Could be saved for a future PR if you set `publish = false` so we don't try to publish this.

fuel-relayer/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+[package]
+name = "fuel-relayer"
+version = "0.1.0"
```
adlerjohn, line 3: Set version number to the latest release, unless we don't want to publish this yet, at this point set to `0.0.0` and `publish = false`.

fuel-core-interfaces/src/db.rs
rakita: Just a note, but it does look ugly. Will remove it.

fuel-relayer/Cargo.toml
rakita: Yea, will add `publish = false`. Ethers does not have that up to data version on crates (last version is from december). Will ping them to update it.

fuel-core-interfaces/src/block_importer.rs
rakita: `da_height` seems okay.
This is going to be changes when we start deciding on message format between modules.

rakita: @Voxelot just to note here, i slighty changed idea of `da_height` to represent eth block in past that will enable all token/validator deposits.

fuel-core-interfaces/src/relayer.rs
```diff
@@ -0,0 +1,143 @@
...
+        for (ref address,ref stake) in changes {
+            let _ = Storage::<Address,u64>::insert(self,address,stake);
+        }
+    }
+
+    /// get validator set for current eth height
+    async fn current_validator_set(&self) -> HashMap<Address,u64>;
+
+    /// set last finalized fuel block. In usual case this will be
+    async fn set_current_validator_set_eth_height(&self, block: u64);
```
Voxelot, line 67: Is this suppose to be the eth height of the last fuel-block that's been finalized? Or is it the eth height of the current finalized validator set? This name and the comment seem confusing to me.

fuel-core-interfaces/src/relayer.rs
```diff
@@ -0,0 +1,143 @@
...
+    async fn get_block_height(&self) -> u64;
+
+
+    /// set newest finalized eth block
+    async fn set_eth_finalized_block(&self, block: u64);
+
+    /// assume it is allways set sa initialization of database.
+    async fn get_eth_finalized_block(&self) -> u64;
+
+    /// set last finalized fuel block. In usual case this will be
```
Voxelot, line 88: Should we clarify that this is the fuel block finalized on the da layer as opposed to the latest block finalized by bft consensus?

fuel-core-interfaces/src/relayer.rs
```diff
@@ -0,0 +1,143 @@
...
+            from_fuel_block: u64,
+            to_fuel_block: Option<u64>,
+    ) -> Vec<(u64,HashMap<Address, u64>)>;
+
+    /// current best block number
+    async fn get_block_height(&self) -> u64;
+
+
+    /// set newest finalized eth block
+    async fn set_eth_finalized_block(&self, block: u64);
```
Voxelot, line 83: we should rename usages of eth to da, since we may target other evm compatible chains like celestia.

fuel-core-interfaces/src/relayer.rs
```diff
@@ -0,0 +1,143 @@
...
+use fuel_types::{Address, AssetId, Bytes32, Word};
+use serde::{Deserialize, Serialize};
+use tokio::sync::oneshot;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DepositCoin {
+    pub owner: Address,
+    pub amount: Word,
+    pub asset_id: AssetId,
+    pub eth_block_deposited: u64,
```
Voxelot, line 14:
```suggestion
    pub base_chain_block_deposited: u64,
```

fuel-core-interfaces/src/relayer.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub fuel_block_spend: u64,
```
Voxelot, line 15:
```suggestion
    pub fuel_block_spend: Option<u64>,
```

fuel-core-interfaces/src/relayer.rs
```diff
@@ -0,0 +1,143 @@
...
+    #[error("Temp ProviderError")]
+    ProviderError,
+    #[error("Validator Set not returned, waiting for eth client sync")]
+    ValidatorSetEthClientSyncing,
+    #[error("Asked for unknown eth block")]
+    InitialSyncAskedForUnknownBlock,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum EthSyncedState {
```
Voxelot, line 126: "BaseChainSyncState" or "DaSyncState"

fuel-core/src/database.rs
```diff
@@ -248,3 +260,113 @@ impl InterpreterStorage for Database {
         Ok(block.headers.producer)
     }
 }
+
+#[async_trait]
+impl RelayerDb for Database {
+    /// get validator set for current eth height
+    async fn current_validator_set(&self) -> HashMap<Address, u64> {
```
Voxelot, line 267: Shouldn't we return the database error if one occurs? We shouldn't swallow the error without at least logging it.

fuel-core/src/database.rs
```diff
@@ -248,3 +260,113 @@ impl InterpreterStorage for Database {
...
+                Self(Address::try_from(i.as_ref()).unwrap())
+            }
+        }
+        let mut out = HashMap::new();
+        for diff in self.iter_all::<WrapAddress, u64>(columns::VALIDATOR_SET, None, None, None) {
+            match diff {
+                Ok((address, stake)) => {
+                    out.insert(address.0, stake);
+                }
+                Err(_) => return out,
```
Voxelot, line 280: Shouldn't we return the actual database error instead of swallowing and returning a default response? At the very least the db error should be logged for troubleshooting purposes.

fuel-core/src/database.rs
```diff
@@ -248,3 +260,113 @@ impl InterpreterStorage for Database {
...
+    ) -> Vec<(u64, HashMap<Address, u64>)> {
+        let to_fuel_block = if let Some(to_fuel_block) = to_fuel_block {
+            if from_fuel_block > to_fuel_block {
+                return Vec::new();
+            }
+            to_fuel_block
+        } else {
+            u64::MAX
+        };
+        struct Wrapu64BE(pub u64);
```
Voxelot, line 300:
```suggestion
        struct WrapU64Be(pub u64);
```

fuel-core/src/database.rs
```diff
@@ -248,3 +260,113 @@ impl InterpreterStorage for Database {
...
+                    out.push((block, diff))
+                }
+                Err(_) => return out,
+            }
+        }
+        out
+    }
+
+    /// current best block number
+    async fn get_block_height(&self) -> u64 {
```
Voxelot, line 331: All these methods should return `Result<.., Error>`

fuel-core/src/database/metadata.rs
```diff
@@ -3,8 +3,11 @@ use crate::database::Database;
 use crate::model::fuel_block::BlockHeight;
 use crate::state::Error;
 
-const CHAIN_NAME_KEY: &[u8] = b"chain_name";
-const START_HEIGHT_KEY: &[u8] = b"chain_height";
+pub(crate) const CHAIN_NAME_KEY: &[u8] = b"chain_name";
+pub(crate) const CHAIN_HEIGHT_KEY: &[u8] = b"chain_height";
+pub(crate) const ETH_FINALIZED_BLOCK_NUMBER: &[u8] = b"eth_finalized_block_number";
+pub(crate) const ETH_FINALIZED_FUEL_BLOCK_NUMBER: &[u8] = b"eth_finalized_fuel_block_number";
+pub(crate) const CURRET_VALIDATOR_SET_BLOCK: &[u8] = b"current_validator_set";
```
Voxelot, line 10:
```suggestion
pub(crate) const CURRENT_VALIDATOR_SET_BLOCK: &[u8] = b"current_validator_set";
```

fuel-core-interfaces/src/relayer.rs
rakita: Agree

fuel-core-interfaces/src/relayer.rs
rakita: Yea this is not nice naming, i refactored it

fuel-core-interfaces/src/relayer.rs
rakita: i renamed it to `deposited_da_height` so it is in same naming as others feilds.

fuel-core-interfaces/src/relayer.rs
rakita: I assuming that `fuel_block_spend` is zero if it is not spend, and with that we are "saving" some space. I  am okay with leaving it as it is now or adding Option if you like it more.

fuel-core-interfaces/src/relayer.rs
rakita: DaSyncState seems nice. There is some inconsisntency with state, we dont use it per say inside relayer and those things will get flushed out when we know who needs it.

fuel-core/src/database.rs
rakita: I handled potential error with panic so it is easier to debug, this kind of db error are not something we can recover from.

fuel-core-interfaces/src/relayer.rs
Voxelot: True, the option would add an extra byte. However, I think deposits could be pruned eventually so maybe we can afford to be less conservative and have the types be more expressive.

fuel-core/src/model/fuel_block.rs
```diff
@@ -22,7 +22,7 @@ use std::{
     Into,
     From,
 )]
-pub struct BlockHeight(u32);
+pub struct BlockHeight(pub u32);
```
Voxelot, line 25: we shouldn't need to expose this since we have all the derive_more::{From, Into} setup.

fuel-relayer/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+repository = "https://github.com/FuelLabs/fuel-core"
+description = "Fuel Relayer"
+publish = false
+
+[dependencies]
+anyhow = "1.0"
+async-trait = "0.1"
+bytes = "1.1"
+env_logger = "0.9"
+ethers-core = { git = "https://github.com/rakita/ethers-rs.git", branch = "master", package = "ethers-core", default-features = false }
```
Voxelot, line 15: Can we not use the versions available from crates.io?

fuel-core/src/database.rs
adlerjohn: Is it possible that crashing the node leaves the database in a corrupt state?

fuel-relayer/Cargo.toml
adlerjohn: Ref: https://github.com/FuelLabs/fuel-core/pull/207#discussion_r826727630

rakita: I added few things in ethers that is in current master, when it is published we will switch to cargo version

fuel-core-interfaces/src/relayer.rs
rakita: Changed it to Option

fuel-relayer/Cargo.toml
Voxelot: wow, for some reason I didn't notice John's comment right below saying the same thing. Sorry, for the dupe.

fuel-core-interfaces/src/db.rs
```diff
@@ -120,17 +122,33 @@ pub mod helpers {
...
     }
 
-    impl DummyDB {
+    #[derive(Clone, Debug)]
+    pub struct Data {
+        pub block_height: u64,
+        pub current_validator_set: HashMap<Address, u64>,
+        pub current_validator_set_block: u64,
+        pub validator_set_diff: BTreeMap<u64, HashMap<Address, u64>>,
+        pub eth_finalized_block: u64,
```
Voxelot, line 147: [nit]

```suggestion
        pub da_finalized_block: u64,
```

fuel-core/src/database.rs
Voxelot: looks like this still needs to be addressed

fuel-core-interfaces/src/relayer.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// What needs to be done is to have validator set state and diff as separate database values.
+    async fn insert_validators_diff(&mut self, da_height: u64, stakes: &HashMap<Address, u64>) {
+        let _ = Storage::<u64,HashMap<Address,u64>>::insert(self, &da_height,stakes);
+    }
+
+    /// get stakes difference between fuel blocks. Return vector of changed (some blocks are not going to have any change)
+    async fn get_validator_diffs(
+            &self,
+            from_fuel_block: u64,
+            to_fuel_block: Option<u64>,
```
Voxelot, line 61: given that we insert diffs using da_height, would it be more accurate to call these `from_da_block` and `to_da_block`?

fuel-core/src/database.rs
```diff
@@ -248,3 +260,132 @@ impl InterpreterStorage for Database {
...
+        ) {
+            match diff {
+                Ok((key, diff)) => {
+                    let block = key.0;
+                    if block >= to_fuel_block {
+                        return out;
+                    }
+                    out.push((block, diff))
+                }
+                Err(err) => panic!("get_validator_diffs unexpected error:{:?}", err),
```
Voxelot, line 322: I'm not really a fan of seeing so many panics, but if it's truly unrecoverable, we should set `panic = "abort"` in our main cargo.toml or configure some kind of panic handler on the service where we'd spawn the relayer task. Otherwise, the relayer task may panic but the rest of the node would remain operational in a silently degraded state. 

https://doc.rust-lang.org/cargo/reference/config.html#profilenamepanic

rakita: `panic = "abort"` is good idea, in general at least at beggining i realy would like for us to break spectacularly if something goes wrong or unexpected and in the most painful but fastest way to get to the core of the problem. After this first we can relax a litlle bit and have more error frendly way of handing these kind of situations.

rakita: and in the end i would expect us to have `fuel-core` inside somekind of deamon as if something fails and stops the app it will start again from beggining.

fuel-core/src/database/validator_set_diffs.rs
```diff
@@ -0,0 +1,34 @@
+use crate::database::{columns, Database, KvStoreError};
+use fuel_storage::Storage;
+use fuel_types::Address;
+use std::{borrow::Cow, collections::HashMap};
+
+impl Storage<u64, HashMap<Address, u64>> for Database {
```
Voxelot, line 6: Should we maybe use a new-type similar to BlockHeight for this? That way it's more clear from a type-system perspective whether this is the fuel block height vs da block height?

fuel-relayer/README.md
```diff
@@ -0,0 +1,116 @@
...
+We are adding three colums:
+* Validator set diff: mapping of fuel_block and hashmap of diffs for finalized new stakes.
+* Validator set: mapping of validator address to its current stakes.
+* Deposits: mapping of deposit_nonce to token deposits.
+
+## Block Commitments
+
+When new fuel block is made it needs to be commited to ethereum network.
+
+There is timeframe of few eth blocks where current leader can broadcast block to ethereum network. In case that he do it in time he will receive additional reward.
```
Voxelot, line 112:
```suggestion
There is timeframe of few eth blocks where current leader can broadcast block to ethereum network.
```

there was no reward discussed for this in the specs

fuel-relayer/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+In case that leader does not commit fuel block next leader will have ability to do it for him as bundle block commit, in that case bonus reward will go to second leader that has spend additional gas to push this change to ethereum network.
```
Voxelot, line 114:
```suggestion
In case that leader does not commit fuel block next leader will have ability to do it for him as bundle block commit.
```

again, these reward schemes you came up with were not agreed on or discussed in the specs.

fuel-relayer/README.md
```diff
@@ -0,0 +1,116 @@
...
+* "transaction_hash":Some(0xcf7baaa6e6cd2d863807a3b8d3168f8d2699ad043e3ac249e82887fda6c4ef55),
+* "transaction_index":Some(130),
+* "log_index":Some(172),
+* "transaction_log_index":"None",
+* "log_type":"None",
+* "removed":"Some(false)"
+
+### Database:
+
+We are adding three colums:
```
Voxelot, line 103: The doc won't make sense to say it will be adding new columns after it gets merged

```suggestion
The relayer utlizes the following columns:
```

fuel-relayer/src/config.rs
```diff
@@ -0,0 +1,69 @@
+use std::{str::FromStr, time::Duration};
+
+use ethers_core::types::H160;
+
+#[derive(Clone, Debug)]
+pub struct Config {
+    /// number of blocks between eth blocks where deposits/validator set get finalized.
+    /// Finalization is done on eth height measurement.
+    pub eth_finality_slider: u64,
```
Voxelot, line 9: I think `finality_period` might be a more clear description?

```suggestion
    pub eth_finality_period: u64,
```

fuel-relayer/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+pub mod config;
+pub mod log;
```
Voxelot, line 2: Seems like these eth logs aren't needed by anything else other than the relayer?

Voxelot, line 2: Seems like these eth logs aren't needed by anything else other than the relayer?

```suggestion
pub(crate) mod log;
```

fuel-relayer/src/pending_events.rs
```diff
@@ -0,0 +1,375 @@
...
+
+use fuel_core_interfaces::relayer::RelayerDb;
+use fuel_tx::{Address, AssetId, Bytes32};
+use fuel_types::Word;
+use tracing::info;
+
+use crate::log::EthEventLog;
+
+pub struct PendingEvents {
+    /// Pendning stakes/assets/withdrawals. Before they are finalized
```
Voxelot, line 11:
```suggestion
    /// Pending stakes/assets/withdrawals. Before they are finalized
```

fuel-relayer/src/relayer.rs
```diff
@@ -0,0 +1,726 @@
...
+};
+use fuel_core_interfaces::{
+    block_importer::NewBlockEvent,
+    relayer::{RelayerDb, RelayerError, RelayerEvent, RelayerStatus},
+    signer::Signer,
+};
+
+const REPORT_PROGRESS_EVERY_N_BLOCKS: u64 = 500;
+const PROVIDER_INTERVAL: u64 = 1000;
+///
```
Voxelot, line 30: ?
```suggestion

```

fuel-relayer/src/relayer.rs
```diff
@@ -0,0 +1,726 @@
...
+pub struct Relayer {
+    /// Pendning stakes/assets/withdrawals. Before they are finalized
```
Voxelot, line 32:
```suggestion
    /// Pending stakes/assets/withdrawals. Before they are finalized
```

fuel-relayer/src/validator_set.rs
```diff
@@ -0,0 +1,69 @@
+use std::collections::HashMap;
+
+use fuel_core_interfaces::relayer::RelayerDb;
+use fuel_tx::Address;
+
+pub struct CurrentValidatorSet {
+    /// Current validator set
+    pub set: HashMap<Address, u64>,
+    /// current fuel block
```
Voxelot, line 9:
```suggestion
    /// Current DA block
```

fuel-relayer/src/validator_set.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// it supports only going up
+    pub async fn bump_validators_to_da_height(&mut self, da_height: u64, db: &mut dyn RelayerDb) {
+        match self.da_height.cmp(&da_height) {
+            std::cmp::Ordering::Less => {}
+            std::cmp::Ordering::Equal => {
+                // unusual but do nothing
+                return;
+            }
+            std::cmp::Ordering::Greater => {
+                // if curent block is greater then new included block there is some problem.
```
Voxelot, line 51: Should we `warn!` node operators to this effect?

fuel-relayer/src/validator_set.rs
```diff
@@ -0,0 +1,69 @@
+use std::collections::HashMap;
+
+use fuel_core_interfaces::relayer::RelayerDb;
+use fuel_tx::Address;
+
+pub struct CurrentValidatorSet {
```
Voxelot, line 6: Are there any tests for this?

fuel-relayer/test/run.rs
```diff
@@ -0,0 +1,33 @@
...
+        eth_v2_contract_deployment: 14_380_000,
+        initial_sync_step: 100,
+        eth_initial_sync_refresh: std::time::Duration::from_secs(10),
+    };
+
+    let db = Box::new(Mutex::new(DummyDb::filled()));
+    let (_broadcast_tx, broadcast_rx) = broadcast::channel(100);
+    let signer = Box::new(DummySigner {});
+    let _service = Service::new(&config, db, broadcast_rx, signer).await?;
+    //let sender = service.sender().clone();
```
Voxelot, line 29:
```suggestion
```

fuel-relayer/test/run.rs
```diff
@@ -0,0 +1,33 @@
...
+
+    tokio::time::sleep(std::time::Duration::from_secs(1000)).await;
+    Ok(())
```
Voxelot, line 32: Are there any assertions this test should make? Or is it simply to verify the service can start?

fuel-core/src/database.rs
```diff
@@ -248,3 +260,132 @@ impl InterpreterStorage for Database {
...
+                Err(_) => return out,
+            }
+        }
+        out
+    }
+
+    async fn get_validator_diffs(
+        &self,
+        from_fuel_block: u64,
+        to_fuel_block: Option<u64>,
```
Voxelot, line 288: These say *_fuel_block here, while the `CurrentValidatorSet` which actually calls this api refers to these params as da_height.

fuel-core/src/database.rs
```diff
@@ -248,3 +260,132 @@ impl InterpreterStorage for Database {
...
+    ) -> Vec<(u64, HashMap<Address, u64>)> {
+        let to_fuel_block = if let Some(to_fuel_block) = to_fuel_block {
+            if from_fuel_block > to_fuel_block {
+                return Vec::new();
+            }
+            to_fuel_block
+        } else {
+            u64::MAX
+        };
+        struct WrapU64Be(pub u64);
```
Voxelot, line 298: If we used a new-type for the da_height, we could just impl Into/TryFrom Vec<u8> and avoid all these inline structs.

fuel-relayer/src/pending_events.rs
```diff
@@ -0,0 +1,375 @@
...
+    pub fn clear(&mut self) {
+        self.pending.clear()
+    }
+
+    pub fn push_back(&mut self, pending: PendingDiff) {
+        self.pending.push_back(pending)
+    }
+
+    /// Bundle all removed events to apply them in same time when all of them are flushed.
+    /// Done
```
Voxelot, line 92:
```suggestion
```

fuel-relayer/src/pending_events.rs
```diff
@@ -0,0 +1,375 @@
...
+            }
+        } else {
+            // if there are not pending block for removal just add it.
+            self.bundled_removed_eth_events
+                .push((eth_block, vec![event]));
+        }
+    }
+
+    /// Handle eth log events
+    /// Done
```
Voxelot, line 114:
```suggestion
```

fuel-relayer/src/pending_events.rs
```diff
@@ -0,0 +1,375 @@
...
+                }
+            }
+        }
+        // apply new event to pending queue
+        self.append_eth_events(&event, eth_block).await;
+    }
+
+    /// At begining we will ignore all event until event for new fuel block commit commes
+    /// after that syncronization can start.
+    /// Done
```
Voxelot, line 139:
```suggestion
```

fuel-relayer/src/pending_events.rs
```diff
@@ -0,0 +1,375 @@
...
+                *last_diff.stake_diff.entry(withdrawer).or_insert(0) -= withdrawal as i64;
+            }
+            EthEventLog::FuelBlockCommited { .. } => {
+                // TODO do nothing? or maybe update some state as BlockCommitSeen, BlockCommitFinalized, etc.
+            }
+        }
+    }
+
+    /// Used in two places. On initial sync and when new fuel blocks is
+    /// Done
```
Voxelot, line 179:
```suggestion
```

fuel-relayer/src/pending_events.rs
```diff
@@ -0,0 +1,375 @@
...
+
+pub struct PendingEvents {
+    /// Pendning stakes/assets/withdrawals. Before they are finalized
+    /// it contains every fuel block and its span
+    pending: VecDeque<PendingDiff>,
+    /// This is litlle bit hacky but because we relate validator staking with fuel commit block and not on eth block
+    /// we need to be sure that we are taking proper order of those transactions
+    /// Revert are reported as list of reverted logs in order of Block2Log1,Block2Log2,Block1Log1,Block2Log2.
+    /// I checked this with infura endpoint.
+    bundled_removed_eth_events: Vec<(u64, Vec<EthEventLog>)>,
```
Voxelot, line 18: Why not use a btree map instead?

fuel-relayer/src/pending_events.rs
```diff
@@ -0,0 +1,375 @@
...
+            }
+            EthEventLog::FuelBlockCommited { .. } => {
+                // TODO do nothing? or maybe update some state as BlockCommitSeen, BlockCommitFinalized, etc.
+            }
+        }
+    }
+
+    /// Used in two places. On initial sync and when new fuel blocks is
+    /// Done
+    pub async fn flush_validator_diffs(
```
Voxelot, line 180: this affects more than just the validator diffs, it seems like this is actually handling all the different kinds of diffs.

```suggestion
    pub async fn commit_diffs(
```

fuel-relayer/src/pending_events.rs
```diff
@@ -0,0 +1,375 @@
...
+                // we are okay to cast it, we dont expect that big of number to exist.
+                *value = ((*value as i64) + diff) as u64;
+                stake_diff.insert(*address, *value);
+            }
+            // push new value for changed validators to database
+            db.insert_validators_diff(diffs.da_height, &stake_diff)
+                .await;
+            db.set_finalized_da_height(diffs.da_height).await;
+
+            // push fanalized deposit to db
```
Voxelot, line 203:
```suggestion
            // push finalized deposit to db
```

fuel-relayer/src/relayer.rs
```diff
@@ -0,0 +1,726 @@
...
+    ) -> Result<T::Output, RelayerError> {
+        loop {
+            tokio::select! {
+                biased;
+                inner_fuel_event = self.receiver.recv() => {
+                    tracing::info!("Received event in stop handle:{:?}",inner_fuel_event);
+                    match inner_fuel_event {
+                        Some(RelayerEvent::Stop) | None =>{
+                            self.status = RelayerStatus::Stop;
+                            return Err(RelayerError::Stoped);
```
Voxelot, line 90:
```suggestion
                            return Err(RelayerError::Stopped);
```

fuel-relayer/src/relayer.rs
```diff
@@ -0,0 +1,726 @@
...
+
+    async fn stop_handle<T: Future>(
+        &mut self,
+        handle: impl Fn() -> T,
+    ) -> Result<T::Output, RelayerError> {
+        loop {
+            tokio::select! {
+                biased;
+                inner_fuel_event = self.receiver.recv() => {
+                    tracing::info!("Received event in stop handle:{:?}",inner_fuel_event);
```
Voxelot, line 86:
```suggestion
                    tracing::info!("Received event in stop handle:{:?}", inner_fuel_event);
```

fuel-relayer/src/relayer.rs
```diff
@@ -0,0 +1,726 @@
...
+                            return;
+                        }
+                        self.handle_inner_fuel_event(inner_fuel_event.unwrap()).await;
+                    }
+
+                    new_block = self.new_block_event.recv() => {
+                        match new_block {
+                            Err(e) => {
+                                error!("Unexpected error happened in relayer new block event receiver:{}",e);
+                                return;
```
Voxelot, line 334: should we also set `self.status = RelayerStatus::Stop;` here as well?

fuel-relayer/src/relayer.rs
```diff
@@ -0,0 +1,726 @@
...
+                Err(err) => {
+                    if self.status == RelayerStatus::Stop {
+                        return;
+                    }
+                    error!("Error happened while doing initial sync:{:?}", err);
+                    continue;
+                }
+            };
+            info!("Initial syncing finished. Continue to active sync.");
+            loop {
```
Voxelot, line 319: does this loop need to be nested?

fuel-relayer/src/pending_events.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    /// Bundle all removed events to apply them in same time when all of them are flushed.
+    /// Done
+    fn bundle_removed_events(&mut self, event: EthEventLog, eth_block: u64) {
+        // agregate all removed events before reverting them.
+        // check if we have pending block for removal
+        if let Some((last_eth_block, list)) = self.bundled_removed_eth_events.last_mut() {
+            // check if last pending block is same as log event that we received.
+            if *last_eth_block == eth_block {
+                // just push it
```
vlopes11, line 99: Maybe the comments are more verbose than needed

fuel-relayer/src/relayer.rs
```diff
@@ -0,0 +1,726 @@
...
+const REPORT_PROGRESS_EVERY_N_BLOCKS: u64 = 500;
+const PROVIDER_INTERVAL: u64 = 1000;
+///
+pub struct Relayer {
+    /// Pendning stakes/assets/withdrawals. Before they are finalized
+    pending: PendingEvents,
+    /// Current validator set
+    current_validator_set: CurrentValidatorSet,
+    /// db connector to apply stake and token deposit
+    db: Box<Mutex<dyn RelayerDb>>,
```
vlopes11, line 37: I think its better to make the `Mutex` protection of the DB encapsulated there instead of forcing the consumers to do it themselves. Many dbs uses that approach. Ex: [sled](https://docs.rs/sled/0.34.7/src/sled/db.rs.html#9-14)

vlopes11 (commented):
some questions/nits

fuel-relayer/src/relayer.rs
Voxelot: It's also what we do inside of fuel-core, the Database is Send + Sync and handles any of the locking concerns internally if needed.

fuel-relayer/src/validator_set.rs
rakita: This is covering old logic it is going to be fully rewritten.

fuel-relayer/test/run.rs
rakita: This is just unitily function/file so i could test if relayer does connect and syncs on ganache or eth client.

fuel-core/src/database.rs
rakita: Yep leftover from old logic, will be updated in new pull request.

fuel-relayer/README.md
rakita: @Voxelot  specs do not cover a lot, as already demonstraited with new requirements of how validator key rotation was introduced. 
Are you saying you do not agree with this or are you proposing something else?

fuel-relayer/src/config.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// It does not need to be aqurate and can be set in past before contracts are deployed.
+    pub eth_v2_contract_deployment: u64,
+    /// number of blocks that will be asked in one step, for initial sync
+    pub initial_sync_step: usize,
+    /// how long do we wait between calling eth client if it finished syncing
+    pub eth_initial_sync_refresh: Duration,
+}
+
+impl Config {
+    pub fn new() -> Self {
```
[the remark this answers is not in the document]
rakita: It is nice to have but it is not needed, i find myself sometimes using `Type::new()` and sometimes as `Type::default()` that is why i always have `new` fn inside

fuel-relayer/src/log.rs
```diff
@@ -0,0 +1,372 @@
...
+    },
+    FuelBlockCommited {
+        block_root: Bytes32,
+        height: Word,
+        da_height: u64,
+    },
+}
+
+lazy_static::lazy_static! {
+    pub static ref ETH_ASSET_DEPOSIT : H256 = H256::from_str("0x0000000000000000000000000000000000000000000000000000000000000001").unwrap();
```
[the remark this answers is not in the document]
rakita: good point

fuel-relayer/src/pending_events.rs
rakita: hm, my first thought, why to use it if it is not needed. But maybe there is something here, i will need to see if there is benefits there.

fuel-core/src/database.rs
```diff
@@ -248,3 +260,132 @@ impl InterpreterStorage for Database {
         Ok(block.headers.producer)
     }
 }
+
+#[async_trait]
+impl RelayerDb for Database {
+    async fn get_validators(&self) -> HashMap<Address, u64> {
+        struct WrapAddress(pub Address);
```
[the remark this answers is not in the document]
rakita: It is strange but it is needed for `iter_all` that requires cast from bytes to type so i needed wrapper so that i can specify endian least.

fuel-core/src/database.rs
```diff
@@ -248,3 +260,132 @@ impl InterpreterStorage for Database {
...
+        impl From<Vec<u8>> for WrapAddress {
+            fn from(i: Vec<u8>) -> Self {
+                Self(Address::try_from(i.as_ref()).unwrap())
```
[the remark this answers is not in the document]
rakita: It is table with identical keys, so we are okey to just unwrap. Check out this comment: https://github.com/FuelLabs/fuel-core/pull/207#discussion_r830876650

fuel-core/src/database.rs
rakita: yea it seems i made change in another branch

fuel-core/src/database.rs
rakita: Problem is that it requires `From` and not `TryFrom` so i will still need wrapper type. Maybe changing `iter_all` to take `TryFrom` is a way to do it

fuel-core/src/database/validator_set_diffs.rs
rakita: I didnt have it inside fuel-interfaces and i defaulted to u64, but yeah changing to BlockHeight is better

fuel-relayer/src/config.rs
rakita: I like slider better, but lets change it to period as it represent time needed before finalization

fuel-relayer/src/pending_events.rs
rakita: hah, this was my thinking so that i dont forget that this is only thing needed :D will remove it

fuel-relayer/src/relayer.rs
rakita: Agree, i will make RelayerDb trait require `Send+Sync` as some functions in `Storage` are mut and because of that RelayerDb needs to have some mut functions.

fuel-relayer/src/relayer.rs
```diff
@@ -0,0 +1,726 @@
...
+
+    /// create provider that we use for communication with ethereum.
+    pub async fn provider(uri: &str) -> Result<Provider<Ws>, Error> {
+        let ws = Ws::connect(uri).await?;
+        let provider =
+            Provider::new(ws).interval(std::time::Duration::from_millis(PROVIDER_INTERVAL));
+        Ok(provider)
+    }
+
+    async fn stop_handle<T: Future>(
```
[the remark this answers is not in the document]
rakita: Talked in person yesterday, it is currently a litlle bit ugly but using `spawn` seems like it is not a solution

fuel-relayer/src/relayer.rs
rakita: Idea (not fully realized) is to have continue with a loop label so that we can jump if connection with eth client breaks but yea at current setup nesting is not needed.

fuel-relayer/src/config.rs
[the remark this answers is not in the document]
Voxelot: I think we should prefer the default trait as it is more idiomatic, and allows you to compose structs with default support.

fuel-relayer/Cargo.toml
rakita: line added

fuel-core-interfaces/src/relayer.rs
rakita: Yeah

fuel-core/src/database.rs
rakita: I have just read this again, new type seems like a way to go :)

fuel-core-interfaces/src/db.rs
```diff
@@ -110,7 +112,7 @@ pub mod helpers {
     //const DB_TX1_HASH: TxId = 0x0000.into();
 
     use core::str::FromStr;
-    use std::sync::Arc;
+    use std::{collections::BTreeMap, sync::Arc};
```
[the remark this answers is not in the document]
rakita: There is a need for them to be sorted

fuel-core-interfaces/src/db.rs
```diff
@@ -577,4 +606,136 @@ pub mod helpers {
             Ok(self.contract.get(&contract_id).is_some())
         }
     }
+
+    // token deposit. Used by relayer.
+    impl Storage<Bytes32, DepositCoin> for DummyDb {
+        type Error = crate::db::KvStoreError;
```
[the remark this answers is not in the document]
rakita: It is a Dummy implementation of `Database` and in that sense has the same error type.

fuel-core-interfaces/src/db.rs
```diff
@@ -577,4 +606,136 @@ pub mod helpers {
...
+        fn insert(
+            &mut self,
+            key: &Bytes32,
+            value: &DepositCoin,
+        ) -> Result<Option<DepositCoin>, Self::Error> {
+            Ok(self.deposit_coin.insert(*key, value.clone()))
+        }
+
+        fn remove(&mut self, _key: &Bytes32) -> Result<Option<DepositCoin>, Self::Error> {
+            unreachable!()
```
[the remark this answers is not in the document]
rakita: It is dumm db implementation for tests in that sense there are a lot of unused functions with `unreachable!()`, when functionality can be added.

fuel-core-interfaces/src/signer.rs
```diff
@@ -0,0 +1,28 @@
+use async_trait::async_trait;
+use fuel_types::Bytes32;
+use thiserror::Error;
+
+#[async_trait]
+pub trait Signer {
+    async fn sign(&self, hash: &Bytes32) -> Result<Bytes32, SignerError>;
```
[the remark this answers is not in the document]
rakita: This is just dummy Signer, will be removed in next iteration as ethers-rs has it own key handling.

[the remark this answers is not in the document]
rakita: This is not going to be used, I left a comment that this is going to be removed

fuel-relayer/src/config.rs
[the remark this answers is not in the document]
rakita: will remove new() and leave only default

fuel-core-interfaces/src/signer.rs
```diff
@@ -0,0 +1,28 @@
...
+}
+
+#[derive(Error, Debug)]
+pub enum SignerError {
+    #[error("Private key not loaded")]
+    KeyNotLoaded,
+}
+
+#[cfg(any(test, feature = "test_helpers"))]
+pub mod helpers {
```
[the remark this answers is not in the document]
rakita: Signer will be removed

fuel-relayer/src/pending_events.rs
rakita: I will probably use that inside the next iteration of Relayer, as I have already started in the other branch.

fuel-core-interfaces/src/relayer.rs
```diff
@@ -0,0 +1,141 @@
...
+            let _ = Storage::<Address,u64>::insert(self,address,stake);
+        }
+        self.set_validators_da_height(da_height).await;
+    }
+
+    /// current best block number
+    async fn get_block_height(&self) -> u64;
+
+    /// get validator set for current eth height
+    async fn get_validators(&self) -> HashMap<Address,u64>;
```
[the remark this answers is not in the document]
rakita: Maybe not, I forgot this is trait: `error: 'impl Trait' only allowed in function and inherent method return types, not in trait method return`

fuel-relayer/README.md
adlerjohn: The specs don't yet cover these cases. As such, preference should be to do the simplest thing possible at first (i.e. nothing), and document that it may change in the future.

fuel-relayer/README.md
adlerjohn: The specs don't yet cover these cases. As such, preference should be to do the simplest thing possible at first (i.e. nothing), and document that it may change in the future.

fuel-relayer/README.md
rakita: @adlerjohn @Voxelot this felt like a good response to the first comment. As already demonstrated with byte price and key rotation those were not the simplest solution for the problem but addons to already functional things.
I am not sure what I am missing here as a reward system seems like a pretty simple, logical and needed thing to do, and from both of you, I didn't get the reasoning why do you want it removed or put something else in place of it?

Voxelot (approved):
looks like most of the feedback is addressed, approving to make it easier to build on going forward.

PR #209, opened 2022-03-14
Deploy Fuel-Core Ingress with SSL Cert

rfuelsh:
The following PR covers the configuration and deployment of:

- NGINX Ingress Controller:

This is adapted from open-source (https://github.com/kubernetes/ingress-nginx) for AWS EKS Network Load Balancer 

Give its updated from open-source it will need to update per new releases of ingress-nginx in the fuel-core repo

- Cert Manager

This is straight forward install via helm chart and this can be easily update per new cert manager release - current version is v1.7.1

- Lets Encrypt Cert Issuer

This is a ClusterIssuer k8s resource to issue a SSL certificate for the custom ingress dns domain - its customizable per env file

- Fuel-Core Ingress

This is a custom k8s ingress resource that routes to the fuel-core NodePort backend service - its customizable per env file

- Fuel-Core NodePort Service

This replaces the original k8s External Load Balancer and is a NodePort Service

Currently this is all deployed manually on the fuel-core EKS cluster, but the plan is to tear down the ingress controller and cert-manager one more and deploy via the fuel-deployment pipeline to verify all code changes.

Future work will be focused on updating the ingress-controller and ingress resource as needed.


deployment/ingress/eks/ingress.yaml
```diff
@@ -0,0 +1,26 @@
...
+            pathType: Prefix
+            backend:
+              service:
+                name: ${k8s_namespace}-service
+                port:
+                  number: ${ingress_http_port}
+  tls:
+    - hosts:
+        - ${ingress_dns}
+      secretName: letsencrypt-prod
```
Voxelot, line 26: We should have docs or scripts for configuring this secret

deployment/ingress/eks/ingress-controller.yaml
```diff
@@ -0,0 +1,652 @@
+apiVersion: v1
```
Voxelot, line 1: Why do we need to maintain our own copy of the nginx helm chart templates? Could we use a stock ones via the deploy script using something like this?

`$ helm install my-ingress-controller nginx/nginx-ingress`

rfuelsh: @Voxelot - this is for specifically for AWS nginx ingress-controller and the specific annotations for AWS ingress controller 

the generic helm chart installs nginx-ingress that is not specific to aws and its load balancers

This article explains it better:

https://aws.amazon.com/blogs/opensource/network-load-balancer-nginx-ingress-controller-eks/

deployment/ingress/eks/ingress.yaml
rfuelsh: I can add a doc, but the secret is created automatically from ClusterIssuer , see

https://cert-manager.io/docs/configuration/acme/

Simply the deploy ingress script will take care of this -> I need to add instructions on it though

Voxelot:
Have we tested a deployment the latest iteration of this PR?

rfuelsh:
> Have we tested a deployment of the latest iteration of this PR?

Yes- I re-deployed the ingress controller, ingress, and cert-manager last night and it was a clean, successful deployment for https://node.swayswap.io

https://github.com/FuelLabs/fuel-deployment/runs/5580179805?check_suite_focus=true

I just need to write some documentation on it

Voxelot (approved):

Issue #133, opened 2022-01-15
Remove dependency on native OpenSSL


adlerjohn:
Reference: https://github.com/FuelLabs/sway/pull/651#issuecomment-1013628844

`fuel-gql-client` depends indirectly on `openssl-sys` (i.e. native OpenSSL), through `cynic`. Should be converted to something that's native Rust, to avoid users having to install OpenSSL system libraries to install `forc`.

We could simply wait until https://github.com/obmarg/cynic/issues/376 lands.

PR #219, opened 2022-03-21
Remove Openssl dependency from fuel-client

ControlCplusControlV:
Disables surf's default features (which compiles using openssl-sys) and instead enables rustls as a replacement. From my local testing it seems to have removed the need for openssl from fuel-gql client. But as fuel-gql-client is a dependency of other crates it would need a version bump to fully remove the openssl dependency

Would it then be better to version bump fuel-gql client in this PR, or save that for a full release to fully remove openssl?

### DIFF
--- a/fuel-client/Cargo.toml
+++ b/fuel-client/Cargo.toml
@@ -17,7 +17,7 @@ path = "src/main.rs"
 [dependencies]
 chrono = { version = "0.4", features = ["serde"] }
 clap = { version = "3.1", features = ["derive"] }
-cynic = { version = "0.14", features = ["surf", "rustls-tls"] }
+cynic = { version = "0.14", features = ["surf"] }
 derive_more = { version = "0.99" }
 fuel-storage = "0.1"
 fuel-tx = { version = "0.6", features = ["serde-types"] }
@@ -28,7 +28,7 @@ hex = "0.4"
 itertools = "0.10"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0", features = ["raw_value"] }
-surf = "2.2"
+surf = {version = "2.2",default-features = false ,features = ["h1-client-rustls"] }
 thiserror = "1.0"
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


ControlCplusControlV:
Done as part of #133

Voxelot:
In our release process, we always bump the versions in a separate PR once we're ready to schedule the release. That way we can review all changes accumulated on master and verify the version bump follows semver conventions.

fuel-client/Cargo.toml
```diff
@@ -28,7 +28,7 @@ hex = "0.4"
 itertools = "0.10"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0", features = ["raw_value"] }
-surf = "2.2"
+surf = {version = "2.2",default-features = false ,features = ["h1-client-rustls"] }
```
adlerjohn, line 31:
```suggestion
surf = { version = "2.2", default-features = false, features = ["h1-client-rustls"] }
```

adlerjohn (commented):
The Cargo lockfile should also be committed, no? I imagine there will be changes, specifically it should no longer have `openssl-sys`.

Voxelot (approved):

Voxelot:
[nit] we may be able to remove the openssl installation steps from the repo README.md / CONTRIBUTING.md

ControlCplusControlV:
Was waiting on a version bump to confirm that before removing that, because openssl-sys is still a dependency of older (latest release) fuel-gql-client versions used by other crates.

Voxelot:
Makes sense, we can defer that to another PR after this is released.

adlerjohn (changes requested):
The lockfile doesn't actually seem to have `openssl-sys` removed?

adlerjohn (approved):
NVM, I see the comment on a circular dependency

Voxelot:
This circular dep should get resolved when https://github.com/FuelLabs/fuels-rs/issues/62 is finished allowing the indexer to upgrade it's version of fuels-rs.

PR #221, opened 2022-03-22
Add sqlite support

tjsharp1:
Add sqlite support for fuel-indexer local running.
Had to create 2 new crates to separate sqlite dependencies because of a linker issue, (which kinda feels like an issue with cargo, I have these feature gated, but it complains anyway).

Here's the issue I ran into, fuel-indexer-schema has it's diesel dependencies behind feature flags, but cargo doesn't seem to understand that:
```
error: multiple packages link to native library `sqlite3`, but a native library can be linked only once

package `libsqlite3-sys v0.9.1`
    ... which satisfies dependency `libsqlite3-sys = ">=0.8.0, <0.23.0"` (locked to 0.9.1) of package `diesel v1.4.8`
    ... which satisfies dependency `diesel = "^1.4"` (locked to 1.4.8) of package `fuel-indexer-schema v0.0.0 (/home/teej/fuel-core/fuel-indexer/schema)`
    ... which satisfies path dependency `fuel-indexer-schema` (locked to 0.0.0) of package `fuel-indexer v0.0.0 (/home/teej/fuel-core/fuel-indexer/lib)`
    ... which satisfies path dependency `fuel-indexer` (locked to 0.0.0) of package `fuel-indexer-derive v0.0.0 (/home/teej/fuel-core/fuel-indexer/derive)`
links to native library `sqlite3`

package `sqlite3-src v0.3.0`
    ... which satisfies dependency `sqlite3-src = "^0.3"` (locked to 0.3.0) of package `sqlite3-sys v0.13.0`
    ... which satisfies dependency `sqlite3-sys = "^0.13"` (locked to 0.13.0) of package `sqlite v0.26.0`
    ... which satisfies dependency `sqlite = "^0.26"` (locked to 0.26.0) of package `api_server v0.1.0 (/home/teej/fuel-core/fuel-indexer/api_server)`
also links to native library `sqlite3`
```


fuel-indexer/api_server/src/main.rs
```diff
@@ -0,0 +1,81 @@
+cfg_if::cfg_if! {
```
Voxelot, line 1: What would happen if someone tried to build this project with `--all-features` enabled? This style of feature flagging seems potentially flakey. I think the database type should be selectable at runtime, rather than rely on compilation options. This way we could package a single docker image that could run in either configuration later on for example.

tjsharp1: I guess I could copy/paste models.rs and make two identical versions of it, but that seems like an even uglier solution. The main issue there is the `use` statements import different schema.rs files for the different database backends, and all that logic should be identical between the two. There's nothing stopping us from just building it twice and then wrapping a simple shell script around the two.

Voxelot: Even if it was uglier, I think it'd be preferable in order to avoid the footguns associated of not being able to do things like `cargo test --all-features` where you could potentially have large swaths of the code go untested (which has happened before in fuel crates and is why we are taking a hard stance on this now of always being able to use `--all-features`).

Fortunately, with some quick poking around of other projects depending on diesel it appears that hyperledger sawtooth already solved a similar problem with clever use of traits and connection wrappers.

They were able to unify most of their schema and make it SQL agnostic: https://github.com/hyperledger/transact/blob/bfe7b3fc73d9fc2b6f32a95cf63a4f047ec1fe2a/libtransact/src/state/merkle/sql/store/schema.rs#L18

In the areas where it was SQL dialect specific, rather than relying on the models directly, they use traits which specify the kinds of operations they'd like to do, and then implement them for each type of possible SQL connection. Here's an example of where they have two "conflicting" tables:

[postgres_merkle_radix_tree_node](https://github.com/hyperledger/transact/blob/bfe7b3fc73d9fc2b6f32a95cf63a4f047ec1fe2a/libtransact/src/state/merkle/sql/store/schema.rs#L48)
[sqlite_merkle_radix_tree_node](https://github.com/hyperledger/transact/blob/bfe7b3fc73d9fc2b6f32a95cf63a4f047ec1fe2a/libtransact/src/state/merkle/sql/store/schema.rs#L36)

And here is where they use alternate implementations of the conflicting models by implementing the same trait on different connection types:

[sqlite connection impl](https://github.com/hyperledger/transact/blob/e2def80b9c57f505bc8aaa0cbf98745b87d283d0/libtransact/src/state/merkle/sql/store/operations/write_changes.rs#L62)
[pg connection impl](https://github.com/hyperledger/transact/blob/e2def80b9c57f505bc8aaa0cbf98745b87d283d0/libtransact/src/state/merkle/sql/store/operations/write_changes.rs#L188)

Voxelot:
@tjsharp1 do we still need this pr open in this repo?

Voxelot:
closing as this is significantly out of date. The branch history should be preserved if needed later on.

PR #222, opened 2022-03-22
Deploy prometheus helm chart to k8s 

rfuelsh:
- deploy prometheus helm chart to k8s (https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack)

- updating deploy script comments

### DIFF
--- a/deployment/scripts/prometheus-deploy.sh
+++ b/deployment/scripts/prometheus-deploy.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+set -o allexport && source .env && set +o allexport 
+
+if [ "${k8s_provider}" == "eks" ]; then
+    echo "Updating your kube context locally ...."
+    aws eks update-kubeconfig --name ${TF_VAR_eks_cluster_name}
+    echo "Deploying kube-prometheus helm chart to ${TF_VAR_eks_cluster_name} ...."
+    helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+    helm repo update
+    helm upgrade kube-prometheus prometheus-community/kube-prometheus-stack \
+              --install \
+              --create-namespace \
+              --namespace=monitoring \
+              --wait \
+              --timeout 8000s \
+              --debug
+else
+   echo "You have inputted a non-supported kubernetes provider in your .env"
+fi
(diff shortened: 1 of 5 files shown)


deployment/scripts/prometheus-deploy.sh
```diff
@@ -0,0 +1,20 @@
...
+
+set -o allexport && source .env && set +o allexport 
+
+if [ "${k8s_provider}" == "eks" ]; then
+    echo "Updating your kube context locally ...."
+    aws eks update-kubeconfig --name ${TF_VAR_eks_cluster_name}
+    echo "Deploying kube-prometheus helm chart to ${TF_VAR_eks_cluster_name} ...."
+    helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+    helm repo update
+    helm upgrade kube-prometheus prometheus-community/kube-prometheus-stack \
```
Voxelot, line 11: we should be versioning external dependencies like third-party helm charts. Otherwise, we may run into unpredictable behavior.

Voxelot, line 11: we should be versioning external dependencies like third-party helm charts. Otherwise, we may run into unpredictable behavior.

rfuelsh: if we want to deploy a particular helm chart of prometheus kubernetes sure- but typically the latest released helm chart (in my experience) if verified and stable- (https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) is open source and maintained well

Lemme know

Voxelot: Could we atleast do some kind of semver based range? I.e. similar to how they specify dependencies here? https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/Chart.yaml#L42

Otherwise we risk pulling in some breaking new version of prometheus that isn't compatible with our metrics endpoint in the future and it would be painful to go back and pick versions.

rfuelsh: Sure I add the sem versio of 34.1.0- https://artifacthub.io/packages/helm/prometheus-community/kube-prometheus-stack from (17 Mar, 2022)

Voxelot (approved):

PR #225, opened 2022-03-22
Adjust to fuel-core-service to match helm templates

rfuelsh:
Not so much a typo or error, just continuation of the service name per se the helm charts service definition

### DIFF
--- a/deployment/ingress/eks/ingress.yaml
+++ b/deployment/ingress/eks/ingress.yaml
@@ -18,7 +18,7 @@ spec:
             pathType: Prefix
             backend:
               service:
-                name: ${k8s_namespace}-service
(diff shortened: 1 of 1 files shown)


Voxelot (approved):

deployment/ingress/eks/ingress.yaml
```diff
@@ -18,7 +18,7 @@ spec:
             pathType: Prefix
             backend:
               service:
-                name: ${k8s_namespace}-service
+                name: fuel-core-service
```
Voxelot, line 21: Where does this name come from? It looks like base/service.yaml uses `fuel-core-k8-lb`

PR #231, opened 2022-03-25
[Txpool] Add insert_tx and check mempool when finding tx's

### LINKED ISSUE #152: [txpool] Add insert/find RPCs for txpool
Three function can be exposed to outside world by RPC and they are:
* insert tx
* find tx by hash
* find_dependent (similar as find tx by hash, but it return found tx and its own dependencies if they exist)

While last two does not have a big priority and first one is a requirment for someone to use txpool.

ControlCplusControlV:
I made 2 main changes with this to fix #152 

- Check the mempool for a transaction if it's not found in the database with find_tx
- Utilize new mempool in submit tx under the hood to start using that code in lieu of the present dummy tx pool

Still used to writing this level of rust code though so feel free to let me know if I missed some best practices or did something wrong. Tests are passing for me, but as most of the functionality added is under the hood I may have missed something

### DIFF
--- a/fuel-core/src/tx_pool.rs
+++ b/fuel-core/src/tx_pool.rs
@@ -100,7 +100,7 @@ pub async fn submit_tx(&self, tx: Transaction) -> Result<Bytes32, Error> {
         let new_block_height = current_height + 1u32.into();
 
         // Next fetch the tx from mempool
-        let arc_txs_to_include  = self.fuel_txpool.includable().await.iter().map(|arc| Transaction::clone(&*arc)).collect();;
+        let arc_txs_to_include  = self.fuel_txpool.includable().await.iter().map(|arc| Transaction::clone(&*arc)).collect();
 
         let mut block = FuelBlock {
             header: FuelBlockHeader {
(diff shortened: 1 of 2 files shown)


ControlCplusControlV:
forgot tests were in another crate, looks like I forgot some checking on unwrap()'s, will fix

Voxelot:
moving this to draft since there's still some error handling to address and transaction status fallback handling todo.

fuel-core/src/tx_pool.rs
```diff
@@ -103,7 +116,7 @@ impl TxPool {
                 // TODO: compute the current merkle root of all blocks
                 prev_root: Default::default(),
             },
-            transactions: vec![tx],
+            transactions: txs_to_mine,
```
adlerjohn, line 119: I would avoid using terms like "mine," as it's a heavily-loaded term from a bygone era.

ControlCplusControlV: Yeah a couple variable names I wasn't sure and probably will change some when I get tests passing. Would you recommend using "include" then?

adlerjohn: Ye

Issue #233, opened 2022-03-26
Release for MOVI and misc fixes


Voxelot:
fuel-core will need a release once the following in-flight changes are released from fuel-vm:

- https://github.com/FuelLabs/fuel-vm/pull/100
- https://github.com/FuelLabs/fuel-vm/pull/93
- https://github.com/FuelLabs/fuel-vm/pull/102
- https://github.com/FuelLabs/fuel-vm/pull/103

PR #234, opened 2022-03-26
update to support movi instruction

Voxelot:
Update fuel-asm, fuel-vm, and fuel-tx to get latest bugfixes and opcodes.

related: #233

### DIFF
--- a/fuel-core/Cargo.toml
+++ b/fuel-core/Cargo.toml
@@ -28,15 +28,15 @@ clap = { version = "3.1", features = ["env", "derive"] }
 derive_more = { version = "0.99" }
 dirs = "3.0"
 env_logger = "0.9"
-fuel-asm = { version = "0.2", features = ["serde-types"] }
+fuel-asm = { version = "0.3", features = ["serde-types"] }
 fuel-core-interfaces = { path = "../fuel-core-interfaces", version = "0.4.2" }
 fuel-crypto = { version = "0.3" }
 fuel-merkle = "0.1"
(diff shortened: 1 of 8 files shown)


fuel-core/Cargo.toml
```diff
@@ -28,15 +28,15 @@ clap = { version = "3.1", features = ["env", "derive"] }
 derive_more = { version = "0.99" }
 dirs = "3.0"
 env_logger = "0.9"
-fuel-asm = { version = "0.2", features = ["serde-types"] }
+fuel-asm = { version = "0.3", features = ["serde-types"] }
 fuel-core-interfaces = { path = "../fuel-core-interfaces", version = "0.4.2" }
 fuel-crypto = { version = "0.3" }
```
adlerjohn, line 33: While you're here, should you update `fuel-crypto` to `0.4`?

Voxelot: unfortunately, looks like we'd have to update all the upstream deps first (fuel-tx and fuel-vm). We should probably make it so fuel-core just consumes whatever version of fuel-crypto is re-exported by the VM instead of managing the version here ourselves.

Voxelot: we'd have to update fuel-vm to make that re-export possible first though.

Voxelot: Made an issue to capture this here https://github.com/FuelLabs/fuel-core/issues/236

adlerjohn (approved):

Issue #185, opened 2022-02-23
Exclude pending UTXOs from coin selection API


Voxelot:
To ensure optimal concurrency, when performing coin selection (`coins_to_spend`) we should filter out any unspent but pending utxos from the txpool. This way users can submit multiple transactions at once without conflicting inputs.

We'll also need the ability to provide a list of utxo ids to exclude from the search. This would allow the wallet to track pending utxos which might not have been synced to the local txpool yet.

PR #237, opened 2022-03-27
feat: add `excluded_ids` argument to `Query.coinsToSpend`

AlicanC:
This PR partly implements https://github.com/FuelLabs/fuel-core/issues/185. The unimplemented part was excluding coins in the txpool, but this isn't an immediately necessary feature.

### DIFF
--- a/fuel-client/src/client.rs
+++ b/fuel-client/src/client.rs
@@ -284,14 +284,7 @@ impl FuelClient {
             })
             .try_collect()?;
         let excluded_ids: Option<Vec<schema::UtxoId>> = excluded_ids
-            .map(|ids| {
-                ids.iter()
-                    .map(
-                        #[allow(clippy::needless_question_mark)]
-                        |id| -> Result<_, ConversionError> { Ok(id.parse()?) },
-                    )
-                    .try_collect()
-            })
+            .map(|ids| ids.into_iter().map(UtxoId::from_str).try_collect())
             .transpose()?;
         let query = schema::coin::CoinsToSpendQuery::build(
             &(owner, spend_query, max_inputs, excluded_ids).into(),
--- a/fuel-client/src/client.rs
+++ b/fuel-client/src/client.rs
@@ -284,7 +284,7 @@ impl FuelClient {
             })
             .try_collect()?;
         let excluded_ids: Option<Vec<schema::UtxoId>> = excluded_ids
-            .map(|ids| ids.into_iter().map(UtxoId::from_str).try_collect())
+            .map(|ids| ids.into_iter().map(schema::UtxoId::from_str).try_collect())
             .transpose()?;
         let query = schema::coin::CoinsToSpendQuery::build(
             &(owner, spend_query, max_inputs, excluded_ids).into(),
--- a/fuel-client/src/client.rs
+++ b/fuel-client/src/client.rs
@@ -271,6 +271,7 @@ impl FuelClient {
         owner: &str,
         spend_query: Vec<(&str, u64)>,
         max_inputs: Option<i32>,
+        excluded_ids: Option<Vec<&str>>,
     ) -> io::Result<Vec<schema::coin::Coin>> {
         let owner: schema::Address = owner.parse()?;
         let spend_query: Vec<SpendQueryElementInput> = spend_query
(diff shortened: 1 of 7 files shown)


fuel-client/src/client.rs
```diff
@@ -282,8 +283,19 @@ impl FuelClient {
...
+        let excluded_ids: Option<Vec<schema::UtxoId>> = excluded_ids
+            .map(|ids| {
+                ids.iter()
+                    .map(
+                        #[allow(clippy::needless_question_mark)]
+                        |id| -> Result<_, ConversionError> { Ok(id.parse()?) },
+                    )
+                    .try_collect()
+            })
+            .transpose()?;
```
Voxelot, line 295:
```suggestion
            .map(|ids| ids.into_iter().map(|id| id.parse()).try_collect())
            .transpose()?;
```

fuel-client/src/client/schema/coin.rs
```diff
@@ -124,14 +124,31 @@ pub struct CoinsToSpendArgs {
...
+    )> for CoinsToSpendArgs
+{
+    fn from(
+        r: (
+            Address,
+            Vec<SpendQueryElementInput>,
+            Option<i32>,
+            Option<Vec<UtxoId>>,
+        ),
+    ) -> Self {
```
Voxelot, line 146:
```suggestion
pub(crate) type CoinSelectionArgTuple = (
    Address,
    Vec<SpendQueryElementInput>,
    Option<i32>,
    Option<Vec<UtxoId>>,
);

impl From<CoinSelectionArgTuple> for CoinsToSpendArgs {
    fn from(r: CoinSelectionArgTuple) -> Self {
```

fuel-core/src/coin_query.rs
```diff
@@ -59,10 +60,19 @@ pub fn largest_first(
...
                 .owned_coins_by_asset_id(owner, asset_id, None, None)
                 .try_collect()?;
 
+            // Filter excluded coins
+            coin_ids.retain(|&id| {
+                if let Some(excluded_ids) = excluded_ids {
+                    !excluded_ids.contains(&id)
+                } else {
+                    true
+                }
```
Voxelot, line 73: [nit] could use unwrap_or instead of if-let-else.
```suggestion
                excluded_ids
                    .map(|excluded_ids| !excluded_ids.contains(&id))
                    .unwrap_or(true)
```

fuel-core/src/coin_query.rs
```diff
@@ -131,10 +142,19 @@ pub fn random_improve(
...
                 .owned_coins_by_asset_id(*owner, *asset_id, None, None)
                 .try_collect()?;
 
+            // Filter excluded coins
+            coin_ids.retain(|&id| {
+                if let Some(excluded_ids) = excluded_ids {
+                    !excluded_ids.contains(&id)
+                } else {
+                    true
+                }
```
Voxelot, line 155: same nit as before

```suggestion
                excluded_ids
                    .map(|excluded_ids| !excluded_ids.contains(&id))
                    .unwrap_or(true)
```

Voxelot (approved):

Voxelot:
Just a couple non-blocking minor nits from me, approved!

fuel-core/src/coin_query.rs
```diff
@@ -255,29 +277,26 @@ mod tests {
 
     use super::*;
 
-    fn gen_test_db(owner: Address, asset_ids: &[AssetId]) -> Database {
+    static COIN_INDEX: AtomicU8 = AtomicU8::new(0);
```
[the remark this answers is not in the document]
AlicanC: I actually want to nuke this in favor of a more proper way to generate seeded databases.

If this was the TS repo I'd write a `class TestDatabase extends Database { ... }` with helper methods and put the index counter in there. I'm yet to come up with a similar thing here though.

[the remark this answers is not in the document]
Voxelot: The main issue is trying to use magic global variables. You could have a stateful db builder object that you initialize each test and call `make_coin` as a field on the builder. Then you could have an `impl From<TestDbBuilder> for Database`, which you can then use in your tests.

Voxelot (approved):

Issue #183, opened 2022-02-22
Transaction Status Handling


Voxelot:
We need to come up with a more comprehensive strategy for tracking transaction status. Ideally we avoid TTL based complexity, but there's also a need to balance error feedback UX so sdk users know when or why their transaction failed without storing too much junk data in the node.

We currently have 3 simple states:
 - Submitted
 - Success
 - Failed

However, if the transaction was very invalid (i.e. not includable in a block) this seems like it might need to be separate type of error from a tx that was partially processed within a block (i.e. utxos were spent and outputs created, but the script panicked). The former failure would likely be a transitory concern, while the latter would be a permanent part of the node history.

rakita:
all Included transctions are found in database, while most of pending transaction are found in txpool. We can probably use same mechanism of subscription writen here: #168 to notify on: insertion into txpool/removal from txpool/inclusion in block (and status of that inclusion) so that user can track what happens with its tx.

most troublesome is handling of `local` pending transactions, eth has priority on them so that user who sends tx to client will be sure that that tx will not be rejected because of low gas and it will have ability to track its state.

Voxelot:
yeah we might just want to do the following when we query for transactions status:

1. check if transaction status is saved to the db for success or failure state
2. If not saved to the db, check the txpool to see if it's still pending and return status as submitted.

If a tx is dropped from the txpool, then the status would no longer exist at all. I think this should be ok, as long as the submit API immediately returns any errors to the user if there are obvious problems with the tx.

ControlCplusControlV:
Continuing discussion from my last PR was planning taking on part of this issue, as ideally find_tx checks status to see where a tx is rather its current implementation which just checks if it exists in tx_pool.

AFAIK implementing this PR would expand status to
```

    - Submitted
    - Success
    - Failed
    - Invalid (With this one being an error more so than a status)
    - Reverted

```

What more does this issue entail, as I think atleast part of it could be addressed when I fix find_tx, unless I have overlooked something (which is definitely possible)?

Voxelot:
> Continuing discussion from my last PR was planning taking on part of this issue, as ideally find_tx checks status to see where a tx is rather its current implementation which just checks if it exists in tx_pool.

I don't think it would make sense to check status first for find_tx, since status is an optional field on the transaction object schema which is only loaded after the transaction is already resolved. https://github.com/FuelLabs/fuel-core/blob/v0.5.0/fuel-core/src/schema/tx/types.rs#L215

[...]

PR #238, opened 2022-03-27
Increase usage of new Fuel_txpool

ControlCplusControlV:
I've changed the submit tx and find tx methods to use the new fuel_txpool rather than the old one as part of #152 

One note I do want to make is that I currently use an ugly fallback method where if `includable` doesn't return the tx's correctly I default to previous behavior. I really did try to avoid this behavior but after spending so long diving through the codebase I think it should be a seperate issue as it will take significantly more time to fix and seems to be a bug with `precompute_metadata` as on only some transactions that are passed in that lack metadata precompute_metadata seems to be adding potentially wrong info, which results in them not being returned by includable, specifically in the

```
    tx::get_owned_transactions
    tx::get_transactions
 ```
 tests where this behavior was observed. 
 
 Otherwise submit_tx will use this new behavior adding transactions to the new txpool, and find tx performs a quick check in the mempool, although that won't be used yet until tx's can actually persist in the mempool for any amount of time (we execute them right away currently)

### CODE REVIEW

File: fuel-core/src/tx_pool.rs [hunk 1]
```diff
@@ -83,9 +83,44 @@ impl TxPool {
...
+        // TxPool code flows and prepare tx for execution
+        let mut tx_to_exec = tx.clone();
+
+        if tx_to_exec.metadata().is_none() {
+            tx_to_exec.precompute_metadata();
+        }
+
+        self.fuel_txpool
+            .insert(vec![Arc::new(tx_to_exec.clone())])
+            .await;
```
rakita, line 95: You need to check the result of the `insert`, tx needs to be valid to be included inside the pool.
Maybe we can make the `insert` and `insert_vec` functions to better facilitate this.

fuel-core/src/schema/tx.rs
```diff
@@ -41,8 +41,27 @@ impl TxQuery {
...
+        {
+            let tx_from_mem = found_tx.get(0).unwrap();
+
+            let rewrapped_tx = Some(Transaction(
+                Arc::try_unwrap(tx_from_mem.as_ref().unwrap().clone())
+                    .ok()
+                    .unwrap(),
+            ));
+
+            Ok(rewrapped_tx)
```
Voxelot, line 60:
```suggestion
            Ok((&transaction).clone())
```

fuel-core/src/tx_pool.rs
ControlCplusControlV: So on the tests that currently require the fallback (ie don't work with the new txpool) it is because I am running into the `[Err(Transaction is not inserted. UTXO is not existing: 0x99dd7fc1ad584d9b174275ef9de7bda04fc61e38899fdce22fd31a49f3fc47d6)]` error, which seems to stem from `TestContext`'s `transfer` method generating a `utxo_id:` via `self.rng.gen()`, so when `precompute_metadata` is called it requests an input transaction which doesn't exist causing it to not be inserted.

Or do you just mean runtime validation that inserting was successful?

fuel-core/src/schema/tx.rs
```diff
@@ -41,8 +41,17 @@ impl TxQuery {
...
         let key = id.0;
-        Ok(Storage::<fuel_types::Bytes32, FuelTx>::get(db, &key)?
-            .map(|tx| Transaction(tx.into_owned())))
+
+        let tx_pool = ctx.data::<Arc<TxPool>>().unwrap();
+
+        let found_tx = tx_pool.pool().find(&[key]).await;
+
+        if let Some(transaction) = found_tx.get(0) {
+            Ok(Transaction((&transaction).clone()))
```
Voxelot, line 50: ahh sorry, actually tested the code this time:

```suggestion
        if let Some(Some(transaction)) = found_tx.get(0) {
            Ok(Some(Transaction((transaction.deref()).clone())))
```

fuel-core/src/tx_pool.rs
rakita: yea, there are not check there and data can't be random, it is expected for them to fail. Changing those test to reflect error seems resonable.

runtime validation is needed here, you need to return error there for user to know what is happening.

Voxelot: Hmm, we've been spoofing utxos in tests in the early days, but it seems like the txpool actually validates their existence. Currently in other parts of the system, we validate utxo existence only when the `utxo-validation` flag is enabled. This is because full validation will be a fairly breaking change to all the downstream test suites in the SDK and in sway, so we've been holding off on enabling `utxo-validation` by default until they are adapted to initialize coins properly. 

Given that integrating this new txpool would conflict with our current `utxo-validation` default being `false`, we should gate usage of the new txpool using the same `utxo-validation` feature flag. That way we can begin writing tests with utxo-validation enabled, without breaking everything downstream.

Voxelot: Another note is that the utxo lookup mechanism used by the txpool will need to be adjusted, since it looks up coins only from other pre-existing transactions, instead of directly checking the utxo set. This will cause issues for us since we need to be able to add coins directly to the utxo set from the genesis block without an associated transaction.

ControlCplusControlV: Would moving this change to a separate feature mean making a separate function like `submit_tx_with_validation` and then only making that function available under a feature, or is there a way to overload a function based on feature?

Will work on adding runtime validation in the meantime though

Voxelot: Checking the feature flag should be as simple as something like this:

```
        if self.executor.config.utxo_validation {
            // your new txpool integration logic   
        } else {
            // the legacy dummy logic from before
        }

```

Voxelot: I'll make a new pr to fix the utxo lookup issue I described above

rakita: @Voxelot Yea, i should have used utxo db table for this, it seems like straightforward fix, if you want I can do it.

@ControlCplusControlV in general we should probably change error type in fuel-core to return `anyhow::Error` as we dont care what error is we just want to propagate it to user. But for `fuel-txpool` i feel like we need to chainge it to proper error type. 
In rust you have two main error libs `thiserror` and `anyhow` this is good read: https://github.com/dtolnay/anyhow

There is good comparison there:
```
Use Anyhow if you don't care what error type your functions return, you just want it to be easy. This is common in
application code. Use [thiserror](https://github.com/dtolnay/thiserror) if you are a library that wants to design your own 
dedicated error type(s) so that on failures the caller gets exactly the information that you choose.
```

adlerjohn:
converting to draft since still WIP

fuel-core/src/tx_pool.rs
Voxelot: In the context of the graphql function, we should be able to just use the `?` operator to make the error types match as long as the error has a display trait enabled.

ControlCplusControlV:
@rakita the error handling reading was very helpful, I now understand how Rust and fuel-core by extension is handling errors now. Changing the return type of `submit_tx` seemed to have lots of downstream implications even if it was just the error, so I added a value to Error which just wraps the value of `anyhow::Error`

Also wanted to thank both of you for helping me so much with this first PR, and if all is well on this one, hopefully my next PR will be much faster.

fuel-core/src/tx_pool.rs
```diff
@@ -37,6 +37,8 @@ pub enum Error {
     Database(Box<dyn StdError>),
     #[error("unexpected block execution error {0:?}")]
     Execution(crate::executor::Error),
+    #[error("Tx is invalid, insertion failed {0:?}")]
+    String(String),
```
Voxelot, line 41: This doesn't preserve the stack trace or other parts of the error context, however we could easily fix that with this:

```suggestion
    Other(#[from] anyhow::Result),
```

fuel-core/src/tx_pool.rs
```diff
@@ -83,9 +85,50 @@ impl TxPool {
...
+                .insert(vec![Arc::new(tx_to_exec.clone())])
+                .await;
+
+            // Error handling a WIP, just saving for now
+            if results.get(0).unwrap().is_err() {
+                // Return error if tx failed to insert
+                let failed_with = results.get(0).unwrap().as_ref().err().unwrap().to_string();
+
+                return Err(Error::String(failed_with));
+            }
```
Voxelot, line 109:
```suggestion
            results.into_iter().collect::<Result<Vec<_>, _>>()?;
```

Voxelot, line 109:
```suggestion
            self.fuel_txpool
                .insert(vec![Arc::new(tx_to_exec.clone())])
                .await
                .into_iter()
                .collect::<Result<Vec<_>, _>>()?;
```

Voxelot: https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect

ControlCplusControlV: Didn't know collect() could also handle errors, will start trying to look out for areas where using `iter()` can improve the elegance of my code, thanks for the doc page, that is helpful in explaining it

Voxelot:
We don't have any integration tests covering `utxo_enabled` so these changes currently aren't being tested. We'll need to update the `fuel-tests/tests/tx.rs` to use properly generated and signed transactions (see TransactionBuilder from fuel-tx) and enable utxo-validation there to ensure we don't break the SDK now that they have utxo-validation enabled by default.

ControlCplusControlV:
Ok, will check those out then and add in a test to make sure the `utxo_validation` submit_tx is working fully then, as well as fix the current generic issue with anyhow errors

fuel-core/src/tx_pool.rs
```diff
@@ -37,6 +37,8 @@ pub enum Error {
     Database(Box<dyn StdError>),
     #[error("unexpected block execution error {0:?}")]
     Execution(crate::executor::Error),
+    #[error("Tx is invalid, insertion failed {0:?}")]
+    Other(#[from] anyhow::Result),
```
Voxelot, line 41: whoops meant to use this

```suggestion
    Other(#[from] anyhow::Error),
```

fuel-core/src/tx_pool.rs
Voxelot: @rakita feel free to go ahead and make the fix, I haven't started on it yet.

As a side-note, is there any need for UtxoId to be still be txid + out_idx after this fix? It may help consolidate things like deposit coins vs utxos into the same db table if the id was just a simple 32 byte hash.

ControlCplusControlV:
@Voxelot Let me know if you want to change the test transaction as I know using just Default may not be ideal, `submit_tx` test was always working though even with `utxo_validation` turned on, so I modeled my test closer to that, with a similar test that just ensures the `utxo_validation` flag turned on.

Decided to put it in fuel-tests as since its an endpoint I felt it made the most sense for it to be tested as an external piece rather than an internal componet.

fuel-tests/tests/tx.rs
```diff
@@ -121,6 +121,29 @@ async fn submit() {
...
+
+    let id = client.submit(&tx).await.unwrap();
+    // verify that the tx returned from the api matches the submitted tx
+    let ret_tx = client
+        .transaction(&id.0.to_string())
+        .await
+        .unwrap()
+        .unwrap()
+        .transaction;
+    assert_eq!(tx.id(), ret_tx.id());
```
Voxelot, line 144: Should the test verify that the transaction was actually included in a block? Just returning the transaction object submitted via the API isn't a guarantee that the node is operating properly with the new txpool.

ControlCplusControlV: `submit_tx` also checks this way, however I definitely agree the block is a better source of truth than just a returned transaction, will modify my test to check the block

ControlCplusControlV: Test should check to ensure the transaction was actually included in the block now

Voxelot: Transparent transaction is just a different client encoding type (json instead of binary). We should actually check the `TransactionStatus` variant if it's successful. `client.transaction_status`

Voxelot: TransactionStatus looks like this:
```
pub enum TransactionStatus {
    Submitted {
        submitted_at: DateTime<Utc>,
    },
    Success {
        block_id: String,
        time: DateTime<Utc>,
        program_state: ProgramState,
    },
    Failure {
        block_id: String,
        time: DateTime<Utc>,
        reason: String,
        program_state: Option<ProgramState>,
    },
}
```

So we can confirm it was included in a block if we get the `Success` variant and the block_id is real.

Voxelot: We should also test that the pool is still well behaved after submitting multiple transactions. If we don't verify all these things properly then it could break the SDK if we release this.

fuel-tests/tests/tx.rs
```diff
@@ -121,6 +121,31 @@ async fn submit() {
...
+    // verify that the tx returned from the api matches the submitted tx
+    let ret_tx = client
+        .transaction(&id.0.to_string())
+        .await
+        .unwrap()
+        .unwrap()
+        .transaction;
+
+    let transaction_result = client.transaction_status(&ret_tx.id().to_string()).await;
+    assert!(transaction_result.is_ok());
```
Voxelot, line 146: This result indicates whether the client was able to successfully retrieve the status from the server. It doesn't actually interrogate the response to see if the status was actually successful. You'll need to use something like the `matches!` macro to verify the result value is actually successful.

ControlCplusControlV: Yeah I realized that shortly after I pushed it, almost done with the fix

ControlCplusControlV: Would you want another check to see `block_id`, and if so what would be the best way to check if a block id is valid vs invalid (I am guessing lookup the block and see if the transaction was included inside of it?)

ControlCplusControlV:
Test now 
- Runs with 10 transactions to ensure mempool is working and enduring (variation between tx's atm is gas price, if they need to be more complex lmk)
- Transaction Status is checked to make sure each transaction was a success
- The `block_id` of each transaction is then checked to ensure that the block its from actually exists

fuel-tests/tests/tx.rs
Voxelot: This is pretty good. I thought we'd have to do some genesis coins and transaction signing in the test, but since the fees are zero this works just fine.

Voxelot (approved):

ControlCplusControlV:
I have the ability to merge this, can I merge or do I need a second approval?

adlerjohn:
Ideally let's get a review from @rakita as the txpool OP

fuel-core/src/tx_pool.rs
rakita: will make bug issue.

txid+out_idx is still relevent and helpfull for dependencies

fuel-core/src/tx_pool.rs
```diff
@@ -83,9 +85,43 @@ impl TxPool {
...
+        }
+
+        let tx_id = tx_to_exec.id();
+
         // set status to submitted
-        db.update_tx_status(&tx_id, TransactionStatus::Submitted { time: Utc::now() })?;
+        db.update_tx_status(
+            &tx_id.clone(),
+            TransactionStatus::Submitted { time: Utc::now() },
+        )?;
```
rakita, line 124: @Voxelot  this seems like it needs a rework, updating status before it is executed is not a good flow, everything before block/tx inclusion shouldn't leave any trace in database, only after block is included we could have status of it. 
Expected status of transaction is:
* Pending (inside txpool)
* Included (included in chain)

Is sdk using this information?

fuel-tests/tests/tx.rs
rakita: > We should also test that the pool is still well behaved after submitting multiple transactions, since we had to add logic in this PR to manually remove txs from the pool after including them in a block. If we don't verify all these things properly then it could break the SDK if we release this.

If there is bug in txpool, shound't we just fix it if problem arises? Usage by sdk will be good test.

fuel-core/src/tx_pool.rs
Voxelot: I agree this needs rework and this was called out on the original [task](https://github.com/FuelLabs/fuel-core/issues/152):
> Update transaction status API to fallback to the txpool if the status doesn't exist in the database, using the Submitted status

I'm assuming there will be follow-up PR's to finish the rest of the planned work? @ControlCplusControlV

rakita (approved):
lgtm

fuel-core/src/tx_pool.rs
ControlCplusControlV: @Voxelot yes I am planning on using a fallback based on tx status rather than existence in a separate PR, as I figured there may be some overlap with #183 in part atleast, by tackling

> if the transaction was very invalid (i.e. not includable in a block) this seems like it might need to be separate type of error from a tx that was partially processed within a block (i.e. utxos were spent and outputs created, but the script panicked). 

As I could generate the first status based on if it errors when inserting into the tx pool (which would require utxo verification to be on to determine if a tx is invalid) and then the second would be a result of a failure but the tx was includable.

Unless you would rather that separate PR just cover lookup in mempool by status?

Issue #212, opened 2022-03-15
Use Bincode for p2p data serialization


leviathanbeak:
once #200 and #210 are implemented we need to serialize the data that is sent across the network

PR #239, opened 2022-03-28
implement Request Response for p2p

### LINKED ISSUE #210: implement RequestResponse protocol for fuel-p2p

leviathanbeak:
closes #210 

This PR also completes 1/2 of #212 
since it implements Serialization of Request/Response Messages, the other half is left for Serialization of Gossipsub messages

Further improvements:
- Need to clearly define Request/Response Messages
- Consequently set appropriate size of Request/Response Messages that are sent over the network

### DIFF
--- a/fuel-p2p/Cargo.toml
+++ b/fuel-p2p/Cargo.toml
@@ -19,10 +19,10 @@ futures = "0.3"
 futures-timer = "3.0"
 ip_network = "0.4"
 libp2p = "0.41"
+rand = "0.8.5"
 sha2 = "0.9"
 tokio = { version= "1.17", features = ["sync"] }
 tracing = "0.1"
-rand = "0.8.5"
 
 [dev-dependencies]
 env_logger = "0.9"
--- a/fuel-p2p/Cargo.toml
+++ b/fuel-p2p/Cargo.toml
@@ -13,12 +13,16 @@ description = "Fuel client networking"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+async-trait = "0.1.52"
+bincode = "2.0.0-rc.1"
 futures = "0.3"
 futures-timer = "3.0"
 ip_network = "0.4"
 libp2p = "0.41"
 sha2 = "0.9"
+tokio = { version= "1.17", features = ["sync"] }
 tracing = "0.1"
+rand = "0.8.5"
 
 [dev-dependencies]
 env_logger = "0.9"
--- a/fuel-p2p/Cargo.toml
+++ b/fuel-p2p/Cargo.toml
@@ -28,4 +28,4 @@ tracing = "0.1"
 [dev-dependencies]
 env_logger = "0.9"
 rand = "0.8"
-tokio = { version = "1.8", features = ["full"] }
+tokio = { version = "1.17", features = ["full"] }
--- a/fuel-p2p/Cargo.toml
+++ b/fuel-p2p/Cargo.toml
@@ -14,12 +14,13 @@ description = "Fuel client networking"
 
 [dependencies]
 async-trait = "0.1.52"
-bincode = "2.0.0-rc.1"
+bincode = "1.3"
 futures = "0.3"
 futures-timer = "3.0"
 ip_network = "0.4"
 libp2p = "0.41"
-rand = "0.8.5"
+rand = "0.8"
+serde = { version = "1.0", features = ["derive"] }
 sha2 = "0.9"
 tokio = { version= "1.17", features = ["sync"] }
 tracing = "0.1"
@@ -27,4 +28,4 @@ tracing = "0.1"
 [dev-dependencies]
 env_logger = "0.9"
 rand = "0.8"
-tokio = {version = "1.8", features = ["full"]}
+tokio = { version = "1.8", features = ["full"] }
--- a/fuel-p2p/src/behavior.rs
+++ b/fuel-p2p/src/behavior.rs
@@ -20,6 +20,7 @@ use std::{
     collections::{HashMap, VecDeque},
     task::{Context, Poll},
 };
+use tracing::debug;
 
 // todo: define which events outside world is intersted in
 #[derive(Debug)]
--- a/fuel-p2p/src/behavior.rs
+++ b/fuel-p2p/src/behavior.rs
@@ -1,10 +1,10 @@
 use crate::{
-    config::P2PConfig,
+    config::{P2PConfig, REQ_RES_TIMEOUT},
     discovery::{DiscoveryBehaviour, DiscoveryConfig, DiscoveryEvent},
     gossipsub,
     peer_info::{PeerInfo, PeerInfoBehaviour, PeerInfoEvent},
     request_response::{
-        codec::{MessageExchangeCodec, MessageExchangeProtocol},
+        codec::{MessageExchangeBincodeCodec, MessageExchangeBincodeProtocol},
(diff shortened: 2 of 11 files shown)


fuel-p2p/Cargo.toml
```diff
@@ -13,14 +13,18 @@ description = "Fuel client networking"
...
+rand = "0.8.5"
 sha2 = "0.9"
+tokio = { version= "1.17", features = ["sync"] }
 tracing = "0.1"
 
 [dev-dependencies]
 env_logger = "0.9"
 rand = "0.8"
-tokio = {version = "1.8", features = ["full"]}
+tokio = {version = "1.8", features = ["full"]}
```
adlerjohn, line 30: Is there an issue using two different versions of `tokio`?
```suggestion
tokio = { version = "1.8", features = ["full"] }

```

fuel-p2p/Cargo.toml
```diff
@@ -13,14 +13,18 @@ description = "Fuel client networking"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+async-trait = "0.1.52"
+bincode = "2.0.0-rc.1"
 futures = "0.3"
 futures-timer = "3.0"
 ip_network = "0.4"
 libp2p = "0.41"
+rand = "0.8.5"
```
adlerjohn, line 22: Unless there's a specific reason
```suggestion
rand = "0.8"
```

fuel-p2p/Cargo.toml
```diff
@@ -13,14 +13,18 @@ description = "Fuel client networking"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+async-trait = "0.1.52"
+bincode = "2.0.0-rc.1"
```
adlerjohn, line 17: Should we be using an rc for this dep?

Voxelot: We already use bincode in other parts of fuel-core so we should use the same version here for compiling efficiency as well.

fuel-p2p/src/request_response/codec.rs
```diff
@@ -0,0 +1,110 @@
...
+pub struct MessageExchangeProtocol;
+
+impl ProtocolName for MessageExchangeProtocol {
+    fn protocol_name(&self) -> &[u8] {
+        REQUEST_RESPONSE_PROTOCOL_ID
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct MessageExchangeCodec {}
```
Voxelot, line 29: Maybe we should denote that this is the bincode codec? I'm assuming we'll eventually have multiple codecs.

fuel-p2p/src/request_response/codec.rs
```diff
@@ -0,0 +1,110 @@
...
+
+// Since Bincode does not support async reads or writes out of the box
+// We prefix Request & Response Messages with the length of the data in bytes
```
Voxelot, line 32: I was confused by the `stream.close()`, but realized from the examples that the socket is a dedicated substream for this request-response flow. Copied this description from their docs since I figured it might help with readability here: 

```suggestion
// We prefix Request & Response Messages with the length of the data in bytes
/// We expect the substream to be properly closed when response channel is dropped.
/// Since the request protocol used here expects a response, the sender considers this
/// early close as a protocol violation which results in the connection being closed.
/// If the substream were not properly closed when dropped, the sender would instead
/// run into a timeout waiting for the response.
```

fuel-p2p/Cargo.toml
```diff
@@ -13,14 +13,19 @@ description = "Fuel client networking"
...
+async-trait = "0.1.52"
+bincode = "1.3"
 futures = "0.3"
 futures-timer = "3.0"
 ip_network = "0.4"
 libp2p = "0.41"
+rand = "0.8"
+serde = { version = "1.0", features = ["derive"] }
 sha2 = "0.9"
+tokio = { version= "1.17", features = ["sync"] }
```
Voxelot, line 25: shouldn't this match the version of tokio below?

```suggestion
tokio = { version= "1.8", features = ["sync"] }
```

fuel-p2p/src/behavior.rs
```diff
@@ -82,10 +112,35 @@ impl FuelBehaviour {
...
         let peer_info = PeerInfoBehaviour::new(local_public_key);
 
+        let msg_exchange_protocol =
+            std::iter::once((MessageExchangeProtocol, ProtocolSupport::Full));
+
+        let mut req_res_config = RequestResponseConfig::default();
+        req_res_config.set_request_timeout(
+            p2p_config
+                .set_request_timeout
+                .unwrap_or(Duration::from_secs(20)),
```
Voxelot, line 122: We should have a const for this default

fuel-p2p/src/service.rs
```diff
@@ -332,4 +375,67 @@ mod tests {
             };
         }
     }
+
+    #[tokio::test]
+    async fn request_response_works() {
```
Voxelot, line 380: should we have a negative test that verifies the outbound channel mappings are cleaned up when a timeout occurs?

fuel-p2p/src/request_response/codec.rs
leviathanbeak: yes, it's a one off

fuel-p2p/src/service.rs
leviathanbeak: added one

fuel-p2p/src/service.rs
```diff
@@ -108,23 +114,63 @@ impl FuelP2PService {
...
+
+    /// Sends RequestMessage to a peer
+    /// If the peer is not defined it will pick one at random
+    pub fn send_request_msg(
+        &mut self,
+        peer_id: Option<PeerId>,
+        message_request: RequestMessage,
+        tx_channel: oneshot::Sender<Result<ResponseMessage, ReqResNetworkError>>,
+    ) -> Result<RequestId, RequestError> {
+        let peer_id = match peer_id {
```
Voxelot, line 130: Would it make sense to extra this peer selector into a separate function? Perhaps even under an overridable trait?

fuel-p2p/src/service.rs
```diff
@@ -332,4 +380,130 @@ mod tests {
...
+        let mut p2p_config = build_p2p_config();
+        p2p_config.tcp_port = 4013;
+        p2p_config.bootstrap_nodes = vec![(node_a.local_peer_id, node_a_address.clone().unwrap())];
+        let mut node_b = build_fuel_p2p_service(p2p_config.clone()).await;
+
+        let (tx_test_end, mut rx_test_end) = tokio::sync::mpsc::channel(1);
+
+        loop {
+            tokio::select! {
+                _ = rx_test_end.recv() => {
```
Voxelot, line 467: might be easier to follow the steps 1,2,3,.. if we move this match arm down below the event_a handler.

fuel-p2p/src/service.rs
leviathanbeak: The peer selection will be implemented in PeerManager, until then we can leave it as it is, or I can extract it into an overridable trait / function, let me know what you would prefer.

Voxelot: functionality wise it's already easily overridable since the peer id is passed in as a param. A separate function would just be tidier in case we wanted to quickly identify or tweak areas of the code involving peer selection, but this is just a minor nit.

Voxelot (approved):

PR #256, opened 2022-04-03
Capitalize GQL field descriptions

### LINKED ISSUE #123: Capitalize GQL field descriptions
As seen in the image below, descriptions of fields in GQL Playground are uncapitalized and look weird.

image

We should note that some descriptions are derived from doc comments like in below:
image

For that reason, it might make sense to start capitalizing all doc comments (or even all comments) instead of just capitalizing the ones async-graphql read. (cc @Voxelot)

I think we could go even further and start capitalizing all comments on all repos.

While there are ESLint rules like `capitalized-comments` and `spaced-comment` to style comments a bit ([demo](https://eslint.org/demo#eyJ0ZXh0IjoiLy8gbG93ZXJcbi8vbm8gbGVhZGluZyBzcGFjZSIsIm9wdGlvbnMiOnsicGFyc2VyT3B0aW9ucyI6eyJlY21hVmVyc2lvbiI6ImxhdGVzdCIsInNvdXJjZVR5cGUiOiJzY3JpcHQiLCJlY21hRmVhdHVyZXMiOnt9fSwicnVsZXMiOnsiY2FwaXRhbGl6ZWQtY29tbWVudHMiOjIsInNwYWNlZC1jb21tZW50IjoyfSwiZW52Ijp7fX19)) it doesn't seem like equivalents exist in the Rust ecosystem.

There are generic CI tools that could enforce this or we could contribute these rules to Clippy, but I don't think we have to go there at this time.

ControlCplusControlV:
Done to fix #123  , I believe I regenerated the `schema.sdl` correctly but not too familiar so let me know if I missed a step. Went looking for fields of error in `fuel-core/schema/*` so its also possible I may have missed other GQL descriptions, I also tried to avoid capitalizing words which reference variables (like `utxo_id`) and instead added prefixes to those to fit in line with the new style while maintaining clarity.

### DIFF
--- a/fuel-core/src/schema/block.rs
+++ b/fuel-core/src/schema/block.rs
@@ -64,8 +64,8 @@ impl BlockQuery {
     async fn block(
         &self,
         ctx: &Context<'_>,
-        #[graphql(desc = "id of the block")] id: Option<BlockId>,
-        #[graphql(desc = "height of the block")] height: Option<U64>,
+        #[graphql(desc = "Id of the block")] id: Option<BlockId>,
+        #[graphql(desc = "Height of the block")] height: Option<U64>,
     ) -> async_graphql::Result<Option<Block>> {
         let db = ctx.data_unchecked::<Database>();
         let id = match (id, height) {
@@ -83,10 +83,10 @@ impl BlockQuery {
                     ));
                 } else {
                     db.get_block_id(height.try_into()?)?
-                        .ok_or("block height non-existent")?
+                        .ok_or("Block height non-existent")?
                 }
             }
-            (None, None) => return Err(async_graphql::Error::new("missing either id or height")),
+            (None, None) => return Err(async_graphql::Error::new("Missing either id or height")),
         };
 
         let block = Storage::<fuel_types::Bytes32, FuelBlockDb>::get(db, &id)?
--- a/fuel-core/src/schema/block.rs
+++ b/fuel-core/src/schema/block.rs
@@ -64,7 +64,7 @@ impl BlockQuery {
     async fn block(
         &self,
         ctx: &Context<'_>,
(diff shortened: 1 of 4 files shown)


fuel-core/src/schema/block.rs
```diff
@@ -64,8 +64,8 @@ impl BlockQuery {
     async fn block(
         &self,
         ctx: &Context<'_>,
-        #[graphql(desc = "id of the block")] id: Option<BlockId>,
-        #[graphql(desc = "height of the block")] height: Option<U64>,
+        #[graphql(desc = "Id of the block")] id: Option<BlockId>,
```
AlicanC, line 67: Replace `Id` with `ID` because here it's just an English abbreviation.

fuel-core/src/schema/tx.rs
```diff
@@ -37,7 +37,7 @@ impl TxQuery {
     async fn transaction(
         &self,
         ctx: &Context<'_>,
-        #[graphql(desc = "id of the transaction")] id: TransactionId,
+        #[graphql(desc = "The id of the transaction")] id: TransactionId,
```
AlicanC, line 40: s/id/ID

fuel-core/src/schema/coin.rs
```diff
@@ -69,7 +69,7 @@ impl CoinQuery {
     async fn coin(
         &self,
         ctx: &Context<'_>,
-        #[graphql(desc = "utxo_id of the coin")] utxo_id: UtxoId,
+        #[graphql(desc = "The utxo_id of the coin")] utxo_id: UtxoId,
```
AlicanC, line 72: Replace `utxo_id` with `ID`.

AlicanC (changes requested):

AlicanC (approved):
👍

Voxelot (approved):

PR #257, opened 2022-04-07
feat: add `Query.{balance,balances}` to GraphQL schema

AlicanC:
Adds appropriate fields to our GQL schema to query the total amount of assets an address has.

It works by fetching all coins and adding their amounts together which definitely is not what peak performance looks like. While this is very useful in the short term, I'd say these sums should be indexed somewhere in the future.

### DIFF
--- a/fuel-core/src/schema/balance.rs
+++ b/fuel-core/src/schema/balance.rs
@@ -1,0 +1,192 @@
+use crate::database::{Database, KvStoreError};
+use crate::model::coin::{Coin as CoinModel, CoinStatus};
+use crate::schema::scalars::{Address, AssetId, U64};
+use crate::state::IterDirection;
+use async_graphql::InputObject;
+use async_graphql::{
+    connection::{query, Connection, Edge, EmptyFields},
+    Context, Object,
+};
+use fuel_storage::Storage;
+use itertools::Itertools;
+
+pub struct Balance {
+    owner: fuel_types::Address,
+    amount: u64,
+    asset_id: fuel_types::AssetId,
+}
+
+#[Object]
+impl Balance {
+    async fn owner(&self) -> Address {
+        self.owner.into()
+    }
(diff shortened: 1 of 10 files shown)


fuel-core/src/schema/balance.rs
```diff
@@ -0,0 +1,192 @@
...
+        &self,
+        ctx: &Context<'_>,
+        #[graphql(desc = "address of the owner")] owner: Address,
+        #[graphql(desc = "asset_id of the coin")] asset_id: AssetId,
+    ) -> async_graphql::Result<Balance> {
+        let db = ctx.data_unchecked::<Database>();
+
+        let coin_ids: Vec<fuel_tx::UtxoId> =
+            db.owned_coins(owner.into(), None, None).try_collect()?;
+        let mut coins: Vec<(fuel_tx::UtxoId, CoinModel)> = coin_ids
```
Voxelot, line 55: If we combine all these iterator ops into one stream, we'll cut down memory overhead by not requiring all coins to be loaded into memory into a single vec (try_collect), and they could be summed and discarded as they load one by one from the db.

fuel-core/src/schema/balance.rs
```diff
@@ -0,0 +1,192 @@
...
+            .map(|(asset_id, coins)| {
+                let sum_amount = coins.map(|(_, coin)| coin.amount).sum();
+
+                Balance {
+                    owner: filter.owner.into(),
+                    amount: sum_amount,
+                    asset_id,
+                }
+            })
+            .collect::<Vec<Balance>>();
```
Voxelot, line 117: Same thing here, we should be able to output these sums without collecting all the coins into a vec first.

Voxelot (changes requested):

Voxelot (approved):
much cleaner now, thank you!

PR #259, opened 2022-04-08
update libp2p and import used features only

leviathanbeak:
This PR introduces following updates:

- update libp2p to the latest version, consequently some changes needed to be made to the code,
most notably: `ProtocolsHandler` was renamed to `ConnectionHandler` and 2 methods were completely removed
from the code `inject_connected()` and `inject_disconnected()` in favor of their code being moved to other 2 similar already existing methods

- we also define clearly which libp2p features we need instead of importing default-features

- we use our own production-ready `build_transport()` instead of libp2p's `development_transport()` for transport creation

### DIFF
--- a/fuel-p2p/Cargo.toml
+++ b/fuel-p2p/Cargo.toml
@@ -19,8 +19,8 @@ futures = "0.3"
 futures-timer = "3.0"
 ip_network = "0.4"
 libp2p = { version = "0.43", default-features = false, features = [
-    "dns-tokio", "gossipsub", "identify", "kad", "mdns", "mplex", "noise", 
-    "ping", "request-response", "secp256k1", "tcp-tokio", "yamux", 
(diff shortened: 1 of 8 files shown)


fuel-p2p/Cargo.toml
```diff
@@ -18,13 +18,20 @@ bincode = "1.3"
...
 tokio = { version= "1.17", features = ["sync"] }
 tracing = "0.1"
 
+[dependencies.libp2p]
+version = "0.43"
+features = [
+    "dns-tokio", "gossipsub", "identify", "kad", "mdns", "mplex", "noise", 
+    "ping", "request-response", "secp256k1", "tcp-tokio", "yamux", 
+]
+default-features = false
```
adlerjohn, line 33: Maybe putting this all into a `libp2p = ...` isn't the cleanest, but if the alternative is having to do more work on the toml linter, then just leave it ugly and files some good first issues to follow up.

Voxelot (approved):

PR #260, opened 2022-04-11
Debugger interface

Dentosal:
Extends the current dap interface to allow fine-grained execution control, such as breakpoints and single-steppng. The corresponding changes to https://github.com/FuelLabs/fuel-debugger/ should be merged simultaneously.

~~This PR currently also contains some profiling improvements, but they will be moved to a separate PR before un-draft-ing this PR.~~


adlerjohn (commented):
Can you update to the latest `master` and clean this up before review?

Dentosal:
> Can you update to the latest `master` and clean this up before review?

I will. This is a draft PR for a reason. I didn't realize github would request the reviews before the draft status is cleared; sorry for spam.

Dentosal:
Corresponding debugger client and CLI implementation PR at https://github.com/FuelLabs/fuel-debugger/pull/1 is waiting for this PR to be merged, but shows how this API is meant to be used.

adlerjohn:
cc @vlopes11 @Voxelot for review

fuel-client/assets/schema.sdl
```diff
@@ -65,6 +65,10 @@ type BlockEdge {
 	cursor: String!
 }
 scalar BlockId
+input Breakpoint {
+	contract: [Int!]!
```
vlopes11, line 69: Shouldn't this be a contract id? I wonder why its the same type of `$pc`, Is it an index of an internal mapping? If yes, then we might rather use the contract ID instead of relying in the debugger state in order to interact with it

fuel-client/assets/schema.sdl
```diff
@@ -276,6 +291,14 @@ enum ReturnType {
 	RETURN_DATA
 	REVERT
 }
+type RunResult {
+	state: RunState!
+	breakpoint: OutputBreakpoint
+}
+enum RunState {
+	COMPLETED
```
vlopes11, line 299: Wouldn't `CONTINUE` make more sense here, debug-wise?

fuel-core/src/schema/dap.rs
```diff
@@ -192,4 +192,272 @@ impl DapMutation {
...
+
+    async fn start_tx(
+        &self,
+        ctx: &Context<'_>,
+        id: ID,
+        tx_json: String,
+    ) -> async_graphql::Result<self::gql_types::RunResult> {
+        trace!("Spawning a new VM instance");
+
+        dbg!(&tx_json);
```
vlopes11, line 266: why not using `tracing` here?

fuel-core/src/schema/dap.rs
```diff
@@ -192,4 +192,272 @@ impl DapMutation {
...
+        _ctx: &Context<'_>,
+        _id: ID,
+    ) -> async_graphql::Result<self::gql_types::RunResult> {
+        Err(async_graphql::Error::new(
+            "Feature 'debug' is not compiled in",
+        ))
+    }
+
+    #[cfg(feature = "debug")]
+    async fn continue_tx(
```
vlopes11, line 318: Do we really need the `_tx` suffix here? As long we provide a valid state id, it should be implicit

fuel-core/src/schema/dap.rs
```diff
@@ -192,4 +192,272 @@ impl DapMutation {
...
+    //! GraphQL type wrappers
+    use async_graphql::*;
+
+    use fuel_types::Word;
+    #[cfg(feature = "debug")]
+    use fuel_vm::prelude::Breakpoint as FuelBreakpoint;
+
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, InputObject)]
+    pub struct ContractId {
+        value: [u8; 32],
```
vlopes11, line 368: It feels to me we are creating a tech debt by not using the fuel types here.

fuel-core/src/schema/dap.rs
```diff
@@ -192,4 +192,272 @@ impl DapMutation {
...
+    }
+    impl From<&fuel_types::ContractId> for ContractId {
```
vlopes11, line 370: Since we deref, does it make sense to have an additional `From` here? This will give the impression to the user that we use a transformed type internally - when in fact, we just copy & mirror the owned `From`

fuel-core/src/schema/dap.rs
```diff
@@ -192,4 +192,272 @@ impl DapMutation {
...
+    }
+    impl From<OutputContractId> for fuel_types::ContractId {
+        fn from(id: OutputContractId) -> Self {
+            Self::from(id.value)
+        }
+    }
+
+    #[derive(Debug, Clone, Copy, InputObject)]
+    pub struct Breakpoint {
+        contract: [u8; 32],
```
vlopes11, line 408: Why not use the `ContractId` created in this scope?

fuel-core/src/schema/dap.rs
```diff
@@ -192,4 +192,272 @@ impl DapMutation {
...
+
+    #[cfg(feature = "debug")]
+    impl From<Breakpoint> for FuelBreakpoint {
+        fn from(bp: Breakpoint) -> Self {
+            Self::new(bp.contract.into(), bp.pc)
+        }
+    }
+
+    #[derive(Debug, Clone, Copy, SimpleObject)]
+    pub struct OutputBreakpoint {
```
vlopes11, line 430: Why should an output breakpoint differ from a regular breakpoint? Consider adding a comment to clarify.

vlopes11 (commented):
Preliminary questions

fuel-core/src/schema/dap.rs
Dentosal: Just a left-over debug print, slipped past me. Removed.

We probably should deny `dbg_macro` lint in the CI.

fuel-core/src/schema/dap.rs
Dentosal: `continue` is a reserved identifier, so it would be quite awkward to use.

fuel-core/src/schema/dap.rs
Dentosal: When I first started working on the GraphQL, this was discussed with @Voxelot and at least his opinion was not to pollute the `fuel-*` crates with `async-graphql` annotations. I don't really see any clean way to do this apart from forking `async-graphql`, but I agree that this is tech dept.

fuel-core/src/schema/dap.rs
Dentosal: Only because the a single type cannot act as both input and output type in `async-graphql`, or at least I couldn't figure out how it could.  I'll add a comment to clarify that.

fuel-core/src/schema/dap.rs
Dentosal: Moreover, `async-graphql` doesn't support transparent wrapper types, (similar to `#[serde(transparent)]`) which makes this even more unergonomic. Contributing that addition might be possible, but I feel like it would already exist if it was easily done.

fuel-core/src/schema/dap.rs
Dentosal: Nevermind, updating to `async-graphql` version `3.0` seems to allow this.

Dentosal:
Blocked, waiting for https://github.com/async-graphql/async-graphql/pull/876 since I'm getting name conflict errors when trying to update to `v3` of `async-graphql`

Dentosal:
Attempted to upgrade to `async-graphql` `v3` but it's broken enough that upgrading directly to `v4` should make sense when it's released. Even version `2.10` breaks backwards compatibility, and we have been pinning to `=2.9` even before this PR.  Anyway, the library is quite painful to use, but migrating to something else would be quite a lot of work as well.

adlerjohn:
@Dentosal can you file an issue to recap your efforts on migrating to `async-graphql` `v3`, and eventually upgrading to `v4`?

Dentosal:
See https://github.com/FuelLabs/fuel-core/issues/310 for upgrading to newer `async-graphql` versions.

fuel-core/src/schema/dap.rs
Dentosal: Done

fuel-core/src/schema/dap.rs
Dentosal: That's a good point

fuel-core/src/schema/dap.rs
```diff
@@ -192,4 +192,264 @@ impl DapMutation {
...
+    }
+
+    #[cfg(feature = "debug")]
+    async fn set_breakpoint(
+        &self,
+        ctx: &Context<'_>,
+        id: ID,
+        breakpoint: self::gql_types::Breakpoint,
+    ) -> async_graphql::Result<bool> {
+        trace!("Continue exectuon of VM {:?}", id);
```
vlopes11, line 246: typo on `exectuon`

fuel-core/src/schema/dap.rs
```diff
@@ -192,4 +192,264 @@ impl DapMutation {
...
     }
+
+    #[cfg(not(feature = "debug"))]
+    async fn set_single_stepping(
+        &self,
+        _ctx: &Context<'_>,
+        _id: ID,
+        _enable: bool,
+    ) -> async_graphql::Result<bool> {
+        Err(async_graphql::Error::new(
```
vlopes11, line 203: Looks like this schema shouldn't be enabled if the `debug` feature isn't on. Since its not the scope of this PR, we could leave an open issue to disable this mod entirely if debug is off.

vlopes11 (dismissed):
LGTM, two optional nits

leviathanbeak (dismissed):

Voxelot:
Should we include these apis in the `fuel-gql-client` and have integ tests in this repo for these apis?

fuel-core/src/schema/dap.rs
Dentosal: I'll create an issue when this is merged.

Dentosal:
> Should we include these apis in the `fuel-gql-client` and have integ tests in this repo for these apis?

Done

fuel-core/src/schema/dap.rs
Voxelot: Since we publish schemas with all features enabled, this is the safest bet.

fuel-core/src/schema/dap.rs
```diff
@@ -192,4 +192,259 @@ impl DapMutation {
...
+    impl From<fuel_types::ContractId> for ContractId {
+        fn from(id: fuel_types::ContractId) -> Self {
+            Self(id)
+        }
+    }
+    impl From<ContractId> for fuel_types::ContractId {
+        fn from(id: ContractId) -> Self {
+            id.0
+        }
+    }
```
Voxelot, line 394: This should already exist in scalars

```suggestion
    pub use crate::schema::scalars::ContractId;
```

fuel-core/src/schema/dap.rs
Voxelot: Is there an issue to track this upgrade?

fuel-tests/tests/debugger.rs
```diff
@@ -0,0 +1,40 @@
+#![cfg(feature = "debug")]
+
+use fuel_core::service::{Config, FuelService};
+use fuel_gql_client::client::FuelClient;
+use fuel_vm::prelude::*;
+
+#[tokio::test]
+async fn debugger() {
```
Voxelot, line 8: good sanity check, but would be helpful to have some comments or more descriptive test names to distinguish where new cases should be made vs added onto this test.

Voxelot (approved):

Voxelot:
feel free to move my nits into followup issues.

fuel-core/src/schema/dap.rs
Dentosal: https://github.com/FuelLabs/fuel-core/issues/310

fuel-core/src/schema/dap.rs
Dentosal: Done

fuel-tests/tests/debugger.rs
Dentosal: I've clarified this a bit

Voxelot (approved):

PR #261, opened 2022-04-11
Added Endpoint to find dependent tx's and fixed find_tx

### LINKED ISSUE #152: [txpool] Add insert/find RPCs for txpool

ControlCplusControlV:
This PR is designed to close #152 .

It addresses @Voxelot request for the `transaction` endpoint to use transaction status an indicator of whether or not to check the mempool or db for a transaction

It also adds a `dependent_txs` endpoint which returns all dependent transactions given a transaction id. This was my first time working as in depth with the GraphQL endpoints, however I believe I tested and implemented this one correctly, I mirrored it's implementation off of `transactions` as they both have similar parameters, although different returns.

### DIFF
--- a/fuel-client/assets/schema.sdl
+++ b/fuel-client/assets/schema.sdl
@@ -1,5 +1,39 @@
 scalar Address
 scalar AssetId
+type Balance {
+	owner: Address!
+	amount: U64!
+	assetId: AssetId!
+}
+type BalanceConnection {
+	"""
+	Information to aid in pagination.
+	"""
+	pageInfo: PageInfo!
+	"""
+	A list of edges.
+	"""
+	edges: [BalanceEdge]
+}
+"""
+An edge in a connection.
+"""
+type BalanceEdge {
+	"""
+	The item at the end of the edge
+	"""
+	node: Balance!
+	"""
+	A cursor for use in pagination
+	"""
+	cursor: String!
+}
+input BalanceFilterInput {
+	"""
+	Filter coins based on the `owner` field
+	"""
+	owner: Address!
+}
 type Block {
 	id: BlockId!
 	height: U64!
@@ -181,6 +215,8 @@ type ProgramState {
 type Query {
 	register(id: ID!, register: U64!): U64!
 	memory(id: ID!, start: U64!, size: U64!): String!
+	balance(owner: Address!, assetId: AssetId!): Balance!
+	balances(filter: BalanceFilterInput!, first: Int, after: String, last: Int, before: String): BalanceConnection!
 	block(id: BlockId, height: U64): Block
 	blocks(first: Int, after: String, last: Int, before: String): BlockConnection!
 	chain: ChainInfo!
--- a/fuel-client/src/client.rs
+++ b/fuel-client/src/client.rs
@@ -11,6 +11,7 @@ pub mod schema;
 pub mod types;
 
 use schema::{
+    balance::BalanceArgs,
     block::BlockByIdArgs,
     coin::{Coin, CoinByIdArgs, SpendQueryElementInput},
     contract::{Contract, ContractByIdArgs},
@@ -299,6 +300,30 @@ impl FuelClient {
             schema::contract::ContractByIdQuery::build(ContractByIdArgs { id: id.parse()? });
         let contract = self.query(query).await?.contract;
         Ok(contract)
+    }
+
+    pub async fn balance(&self, owner: &str, asset_id: Option<&str>) -> io::Result<u64> {
+        let owner: schema::Address = owner.parse()?;
+        let asset_id: schema::AssetId = match asset_id {
+            Some(asset_id) => asset_id.parse()?,
+            None => schema::AssetId::default(),
+        };
+        let query = schema::balance::BalanceQuery::build(BalanceArgs { owner, asset_id });
+        let balance = self.query(query).await?.balance;
+        Ok(balance.amount.into())
+    }
+
+    // Retrieve a page of balances by their owner
+    pub async fn balances(
+        &self,
+        owner: &str,
+        request: PaginationRequest<String>,
+    ) -> io::Result<PaginatedResult<schema::balance::Balance, String>> {
+        let owner: schema::Address = owner.parse()?;
+        let query = schema::balance::BalancesQuery::build(&(owner, request).into());
+
+        let balances = self.query(query).await?.balances.into();
+        Ok(balances)
     }
 }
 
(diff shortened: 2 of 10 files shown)


fuel-core/src/schema/tx.rs
```diff
@@ -42,18 +42,41 @@ impl TxQuery {
         let db = ctx.data_unchecked::<Database>();
         let key = id.0;
 
-        let tx_pool = ctx.data::<Arc<TxPool>>().unwrap();
+        let tx_status = db.get_tx_status(&key).unwrap().unwrap();
```
Voxelot, line 45: Going back to the original context in [#152](https://github.com/FuelLabs/fuel-core/issues/152#issuecomment-1076666461), here's what I actually requested:

> - Update query transaction graphql schema to fallback to the txpool find by id method if the tx doesn't exist in the database
> - Update transaction status API to fallback to the txpool if the status doesn't exist in the database, using the Submitted status

Which doesn't quite to map to what you've interpreted here:
> It addresses @Voxelot request for the transaction endpoint to use transaction status an indicator of whether or not to check the mempool or db for a transaction

The difference was that finding the transactions vs fetching the status, were independent and separate things in my mind. We gain no additional information by querying the transaction status first, and functionally nothing would change if we took out this transaction status check. Rather than doubling our I/O per tx request (status column + tx column) we could skip disk I/O entirely in the best case if the tx is still in the pool because we know the status will be submitted.

Also, this wouldn't work at all once we stop storing the `Submitted` status in the db at all, and just deriving that fact from whether the tx exists in the txpool or not.

fuel-core/src/schema/tx.rs
```diff
@@ -42,18 +42,41 @@ impl TxQuery {
...
+
+        if let TransactionStatus::Submitted { .. } = tx_status {
+            let tx_pool = ctx.data::<Arc<TxPool>>().unwrap();
+
+            let found_tx = tx_pool.pool().find(&[key]).await;
 
-        let found_tx = tx_pool.pool().find(&[key]).await;
+            let return_val = Arc::try_unwrap(found_tx.get(0).unwrap().clone().unwrap()).unwrap();
```
Voxelot, line 52: Why did this [regress](https://github.com/FuelLabs/fuel-core/pull/238#discussion_r836646177) back to the both ugly and hazardous code we had before I reviewed the last changes here?

While [Arc::try_unwrap](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.try_unwrap) will return the inner object inside of an Arc, it has much larger consequences. The transaction pool was designed to avoid extra cloning of transactions, and also support async environments, which is why it returns all txs as an Arc. An Arc is essentially a reference counted item on the heap that is automatically dropped when no more references exist (similar to GC). Using `try_unwrap` like this could cause the API to panic (and potentially crash the entire node) if several callers requested the same tx id from the tx pool. This is because the txpool will clone the arc each time it hands out a tx, while `try_unwrap` says:

> Returns the inner value, if the Arc has exactly one strong reference.
> Otherwise, an [Err](https://doc.rust-lang.org/std/result/enum.Result.html#variant.Err) is returned with the same Arc that was passed in.

Which means that the second simultaneous transaction querier will get an error from try_unwrap, because the first caller which hasn't dropped their reference is still using it for their API response.

fuel-core/src/schema/tx.rs
ControlCplusControlV: My bad, in my mind I had swapped `transaction api` and `transaction status api` thinking you meant check transaction status as an indicator of whether or not to query the txpool vs the database to find a tx. Will revert the changes I've made then in favor of the correct fix.

To make sure future changes are targeting the right area, can you clarify where the `transaction status api` is (I believe you mean (this function)[https://github.com/FuelLabs/fuel-core/blob/master/fuel-client/src/client.rs#L163] in fuel client? And to make sure I am understanding the logic for this new fix the idea is that when fetching a transaction's status it can be checked if the transaction is in the txpool, and if it is just return `Submitted` which saves a db check?

fuel-core/src/schema/tx.rs
ControlCplusControlV: Will revert as this change is touching the wrong function. I wrote it like this after I changed to loop, but I definitely agree I should've kept the old way of handling this return value.

From now on I will avoid `unwrap` unless the value can either can be guaranteed to be non-err or there is explicit error handling for when it does err

fuel-core/src/schema/tx.rs
Voxelot: I see how that could be confusing, and may partially be an artifact of the GraphQL architecture vs a straight-forward rpc pattern to some extent. The status is actually a field on the transaction object, which is lazily resolved in this function here: https://github.com/FuelLabs/fuel-core/blob/master/fuel-core/src/schema/tx/types.rs#L215

> when fetching a transaction's status it can be checked if the transaction is in the txpool, and if it is just return Submitted which saves a db check

This is the intention yes. The more I think about it, we may want to even remove the `Submitted` variant from the status enum that's used in the db, and only have it on the graphql schema related types.

fuel-tests/tests/tx.rs
```diff
@@ -122,6 +122,47 @@ async fn submit() {
...
+        vec![],
+        vec![],
+    );
+
+    let id = client.submit(&tx).await.unwrap();
+
+    let dependent_txs = client.dependent_transactions(&id.to_string()).await;
+
+    println!("{:?}", dependent_txs);
+    assert!(dependent_txs.unwrap().is_some());
```
Voxelot, line 163: The test will only guarantee this api returns _something_ in the future. For example, if someone changed the server-side API to just return `Transaction::Default` the test would tell them everything was still ok.

One way to help with coming up with better test assertions is to think through all the criteria and behaviors a user might expect from this API, and then verify those are achieved. At least from a happy path perspective for integration tests like these. We can assume negative edge cases should be mostly handled at the unit test level.

One way to write these down might look like this:
* as a user of the find_dependent txs API, I expect that the transaction id I'm searching from is included in the response.
* as a user of the find_dependent txs API, I expect that transactions with overlapping txo inputs and outputs will form an ordered priority list based on their dependencies and then by gas price.

Then the test should verify scenarios written out like this, ensuring continuing functionality in the face of future changes and refactors.

Issue #50, opened 2021-11-03
Configurable Block Production Trigger


Voxelot:
Currently, blocks are produced instantly when transactions are submitted. In order to support multi-tx blocks, we need a configuration mechanism to enable/disable different block production triggers on startup.

e.g. `fuel-core run --block-production-mode interval` or `fuel-core run --block-production-mode instant`

This task should cover the implementation of the following block production options:

**`interval` Timer-based**: For simulating consensus block delays, the timer trigger will fire a block production event on a configurable interval
 **`instant` Instant trigger**: For integration testing contracts etc, make the tx pool trigger a block production event every time a tx is inserted.
**`none` No block production**: For passive nodes that simply listen for blocks

`pos` mode will be deferred to a later task.

Dentosal:
For proof-of-authority network, should there exist a hybrid mode of interval and instant, similar to e.g. Nagle's algorithm? 

We have three timers, with following expiration times:
* A: Maximum block time
* B: Minimum allowed block time
* C: Maximum transaction idle time (shorter than A)

When a block is produced, restart timers A and B and clear timer C.
When a new transaction arrives, if timer C is not running, start it.
When either of timers A or C expires, produce a new block.
When a block is full and B has expired, produce a new block.

This way. we can produce new blocks faster than some fixed interval, but still include multiple transactions per block. The last timer, C, makes sure we will not produce blocks too quickly, so other nodes can still keep up.

Voxelot:
This sounds great, and would also be useful in a rollup configuration without PoS. It helps avoid long strings of empty blocks when there's low activity.

PR #263, opened 2022-04-14
Fix Transaction Status API

### LINKED ISSUE #154: [txpool] Implement removal of txs that are included into chain
When new block is included, processed transaction needs to be removed from txpool. Steps for implementing this:
* remove tx from `by_hash`
* remove tx from `by_gas_price`
* remove tx from `by_dependenct`
  * Go over all remove tx outputs and compare them with `coins`. If it is variable outputs check if dependent tx has maching input, if not, remove it. Go over all dependency and decrease their depth.
  * go over all removed tx outputs and compare them with `contracts`. Set `origin` to None, `gas_price` to MAX and decrease depth for it, and all its dependencies found inside `used_by`. 

If block is for some reason reverted, transaction of that block need to be reinjected inside txpool for usage in next block. We can still use `block_update` function. Check how does reverting happen in our PoS consensus. 

this is medium effort to implement and it is hard requirement for client.

ControlCplusControlV:
Closes #154  

This updates the transaction status api to use the mempool as a shortcut to determine transaction status, as if a transaction is in the mempool it's status is submitted. There is a small TODO where the time used for the submitted status is the current time of when the status if fetched rather than the actual time it was submitted.

The `dependent_transactions` api was shelved as I understand it isn't that important, and I thought that if it is implemented it may make more sense to do with once utxo_validation is turned on.

### DIFF
--- a/fuel-core/src/schema/tx/types.rs
+++ b/fuel-core/src/schema/tx/types.rs
@@ -224,13 +224,9 @@ impl Transaction {
 
         let txpool = TxPoolService::new(tx_pooldb, Arc::new(Config::default()));
 
-        let self_id = self.0.id();
+        let transaction_in_pool = txpool.find(&[self.0.id()]).await;
 
-        let do_i_exist = txpool.find(&[self_id]).await;
-
-        let result = do_i_exist.get(0).unwrap().is_some();
-
-        if result {
+        if  transaction_in_pool.get(0).and_then(|y| y.as_ref()).is_some() {
             // TODO, fix this part where submitted time is lied about
             let time = chrono::Utc::now();
             Ok(Some(TransactionStatus::Submitted(SubmittedStatus(time))))
--- a/fuel-core/src/schema/tx/types.rs
+++ b/fuel-core/src/schema/tx/types.rs
@@ -10,7 +10,12 @@ use async_graphql::{Context, Enum, Object, Union};
 use chrono::{DateTime, Utc};
 use fuel_core_interfaces::db::KvStoreError;
 use fuel_storage::Storage;
+use fuel_core_interfaces::txpool::TxPoolDb;
+use fuel_core_interfaces::txpool::TxPool;
 use fuel_types::bytes::SerializableVec;
+use fuel_txpool::Config;
+use std::sync::Arc;
+use fuel_txpool::TxPoolService;
 use fuel_vm::prelude::ProgramState as VmProgramState;
 
 pub struct ProgramState {
@@ -214,8 +219,25 @@ impl Transaction {
 
     async fn status(&self, ctx: &Context<'_>) -> async_graphql::Result<Option<TransactionStatus>> {
         let db = ctx.data_unchecked::<Database>();
-        let status = db.get_tx_status(&self.0.id())?;
-        Ok(status.map(Into::into))
+
+        let tx_pooldb = Box::new(db.clone()) as Box<dyn TxPoolDb>;
+
+        let txpool = TxPoolService::new(tx_pooldb, Arc::new(Config::default()));
+
+        let self_id = self.0.id();
+
+        let do_i_exist = txpool.find(&[self_id]).await;
+
+        let result = do_i_exist.get(0).unwrap().is_some();
+
+        if result {
+            // TODO, fix this part where submitted time is lied about
+            let time = chrono::Utc::now();
+            Ok(Some(TransactionStatus::Submitted(SubmittedStatus(time))))
+        } else {
+            let status = db.get_tx_status(&self.0.id())?;
+            Ok(status.map(Into::into))
+        }
     }
 
     async fn receipts(&self, ctx: &Context<'_>) -> async_graphql::Result<Option<Vec<Receipt>>> {
--- a/fuel-core/src/schema/tx/types.rs
+++ b/fuel-core/src/schema/tx/types.rs
@@ -9,14 +9,14 @@ use crate::{database::Database, schema::block::Block};
 use async_graphql::{Context, Enum, Object, Union};
 use chrono::{DateTime, Utc};
 use fuel_core_interfaces::db::KvStoreError;
-use fuel_storage::Storage;
-use fuel_core_interfaces::txpool::TxPoolDb;
 use fuel_core_interfaces::txpool::TxPool;
-use fuel_types::bytes::SerializableVec;
+use fuel_core_interfaces::txpool::TxPoolDb;
+use fuel_storage::Storage;
 use fuel_txpool::Config;
-use std::sync::Arc;
 use fuel_txpool::TxPoolService;
+use fuel_types::bytes::SerializableVec;
 use fuel_vm::prelude::ProgramState as VmProgramState;
+use std::sync::Arc;
 
 pub struct ProgramState {
     return_type: ReturnType,
--- a/fuel-client/assets/schema.sdl
+++ b/fuel-client/assets/schema.sdl
@@ -222,7 +222,6 @@ type Query {
 	chain: ChainInfo!
 	version: String!
 	transaction(id: TransactionId!): Transaction
-	dependentTransactions(id: TransactionId!): [Transaction!]
 	transactions(first: Int, after: String, last: Int, before: String): TransactionConnection!
 	transactionsByOwner(owner: Address!, first: Int, after: String, last: Int, before: String): TransactionConnection!
 	"""
--- a/fuel-client/assets/schema.sdl
+++ b/fuel-client/assets/schema.sdl
@@ -222,6 +222,7 @@ type Query {
 	chain: ChainInfo!
 	version: String!
 	transaction(id: TransactionId!): Transaction
+	dependentTransactions(id: TransactionId!): [Transaction!]
 	transactions(first: Int, after: String, last: Int, before: String): TransactionConnection!
 	transactionsByOwner(owner: Address!, first: Int, after: String, last: Int, before: String): TransactionConnection!
 	"""
--- a/fuel-client/src/client.rs
+++ b/fuel-client/src/client.rs
@@ -160,7 +160,10 @@ impl FuelClient {
         Ok(transaction.map(|tx| tx.try_into()).transpose()?)
     }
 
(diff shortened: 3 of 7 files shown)


fuel-core/src/schema/tx/types.rs
```diff
@@ -214,8 +219,25 @@ impl Transaction {
 
     async fn status(&self, ctx: &Context<'_>) -> async_graphql::Result<Option<TransactionStatus>> {
         let db = ctx.data_unchecked::<Database>();
-        let status = db.get_tx_status(&self.0.id())?;
-        Ok(status.map(Into::into))
+
+        let tx_pooldb = Box::new(db.clone()) as Box<dyn TxPoolDb>;
+
+        let txpool = TxPoolService::new(tx_pooldb, Arc::new(Config::default()));
```
Voxelot, line 225: Isn't this creating an brand new instance of a txpool on each status request? We should be fetching the shared txpool out of the `ctx` like we do with the db. The db backs the txpool, but a lot of the data in the txpool is managed in memory, so creating an ephemeral txpool like this will not share any data with the txpool used in other places.

Voxelot:
tests?

ControlCplusControlV:
Oh I thought the transaction status api unit tests would cover this, however I will in one to specifically test this new behavior and make sure it is behaving properly

fuel-core/src/schema/tx/types.rs
ControlCplusControlV: I wasn't sure if this was the best way to access the txpool, I went with this approach as I wanted to avoid side effects (hence the clone), but I see why this could be extremely expensive and how the shared data could end up being helpful, so I will move this to fetch from the context

Voxelot: The clone is of the db, which is just cloning a pointer and virtually has no cost. The issue is that creating a new txpool instance wouldn't know about the transaction submitted by the user, as it's a different instance. So the transaction they're looking for would never exist in the temp pool you're making here.

ControlCplusControlV: Ok, will keep that in mind when deciding between whether to clone or pull in an object from the context.

What is the best way to fetch the TxPool from a context - `let txpool = ctx.data::<Arc<dyn TxPool>>().unwrap();` however this results in many tests breaking with `Custom { kind: Other, error: Curl("Send failed since rewinding of the data stream failed"): Send failed since rewinding of the data stream failed }`, what would be the correct way to fetch the TxPool or TxPoolService from the current context?

Voxelot: It's injected into the context here: https://github.com/FuelLabs/fuel-core/blob/master/fuel-core/src/service/graph_api.rs#L37

So based on the type use there it would be the concrete `Arc<TxPool>` instead of dyn TxPool.

ControlCplusControlV:
I've been running into some conversion issues. I noticed the `submit_utxo_verified_tx` test was actually able to catch a bug (with the api returning submitted too much), and the `get_transaction_by_id` test as well, so I was looking to base my new test off one of them. However, `submit_tx` on `FuelClient` currently deletes a transaction immediately after adding it to the `txpool`, and while to test the `Submitted`  status I need a transaction that is currently inside the `txpool`. Then when testing with initializing just a txpool from `TxPoolService` where it is possible to insert a tx without deleting it, the Transaction objects I am using are of `fuel_tx::Transaction` and cannot be converted to the type where their status is checkable through the status api (`fuel_core::schema::tx::types::Transaction` )

So my question boils down to how would this behavior be tested? As the only way to get a transaction from `FuelClient` which is of the type I could check its status using the api would mean its not in the txpool (so it couldn't be submitted, as we are deleting transactions from the txpool atomically), and using a fresh `TxPoolService` doesn't seem to make the transaction accessible in the correct type to check its status via the api.

ControlCplusControlV:
So the `Submitted` Status isn't currently testable with the new fix without significant modification to they way the txpool currently works. There is still some coverage testing `Success` and `Failed`, but there are no current calls which would result in a `Submitted` status because `submit_tx` deletes and executes a transaction from the txpool as soon as its added. The TxPool isn't directly accessible either otherwise I would do a quick test to insert a tx and check status.

I have been able to test it though by removing the block execution and deletion of the tx from txpool by commenting out parts of `submit_tx`. I would prefer to make tests for this to be made a new issue for when tx's will persist in the mempool as it will be a much more natural test, and as of now I haven't found a way to test this update without changes other parts of the Client.

fuel-core/src/schema/tx/types.rs
```diff
@@ -214,8 +216,20 @@ impl Transaction {
 
     async fn status(&self, ctx: &Context<'_>) -> async_graphql::Result<Option<TransactionStatus>> {
         let db = ctx.data_unchecked::<Database>();
-        let status = db.get_tx_status(&self.0.id())?;
-        Ok(status.map(Into::into))
+
+        let txpool = ctx.data::<Arc<TxPool>>().unwrap().pool();
+
+        let transaction_in_pool = txpool.find(&[self.0.id()]).await;
```
rakita, line 222: You can just extend TxPool to support `find_one` function it should be easy to add as you just need to see how `find` is made, this should simplify a bit logic here.
Other than that this seems okay.

Voxelot:
You could include the test that passed when you modified the txpool and add the `#[ignored]` annotation with a todo linking to #50, so that when that task is implemented the test can be re-enabled.

ControlCplusControlV:
I've added an empty test with a `todo!()` macro and explanation of what the test should do for testing the `Submitted` status and a link to the issue which is blocking it, then ignored the test the `todo!()`

ControlCplusControlV:
So to explain this (hopefully) final commit 

- Moves to new `find_one` method
- Extends `TxPool` trait in `fuel_core_interface/src/txpool.rs` to use the new `TxInfo` struct rather than just `Arc<Tx>` as mentioned to me by @rakita . This required `TxInfo` being moved to avoid a circular dependency between `fuel-core-interfaces` and `fuel-txpool`, which now resides in `fuel-core-interfaces` as its a type so I thought it was most fitting there.
- Fixes time so that the `Submitted Status` returned has the accurate time of when the transaction was submitted
- Added a mock test with annotations to an issue and a note to add in an automatic unit test for when execution is decoupled from `submit_tx`, so that `Submitted` can be tested in the future when it's a reachable status

I have re-done manual testing as well to ensure `Submitted` still works properly after this change (modifying `submit_tx` in `submit_utxo_verified` tx test to leave transactions in the txpool, and then checking their status)

fuel-core-interfaces/src/info.rs
```diff
@@ -1,7 +1,10 @@
...
 
-use crate::types::ArcTx;
 use chrono::{DateTime, Utc};
+use fuel_tx::Transaction;
+use std::sync::Arc;
+
+pub type ArcTx = Arc<Transaction>;
 
 #[derive(Debug, Clone)]
 pub struct TxInfo {
```
rakita, line 10: nit: I would just move this file to `fuel-core-interfaces/src/model/info.rs` and maybe even rename it to `txpool.rs` it feels like modules specific to txpool, that is why a little bit a specific naming of file.
other than that lgtm :)

ControlCplusControlV: Ok, I wasn't sure what name so I chose `info` but I agree`txpool` is better to make it more clear in what it does and easier to find. It is now moved it to `fuel-core-interfaces/src/model/txpool.rs`

fuel-core/src/schema/tx/types.rs
```diff
@@ -4,7 +4,9 @@ use super::receipt::Receipt;
 use crate::model::FuelBlockDb;
 use crate::schema::contract::Contract;
 use crate::schema::scalars::{AssetId, Bytes32, HexString, Salt, TransactionId, U64};
+use crate::schema::tx::Arc;
```
Voxelot, line 7: this seems like a strange way to import Arc?

fuel-core/src/schema/tx/types.rs
```diff
@@ -214,8 +216,18 @@ impl Transaction {
...
     async fn status(&self, ctx: &Context<'_>) -> async_graphql::Result<Option<TransactionStatus>> {
         let db = ctx.data_unchecked::<Database>();
-        let status = db.get_tx_status(&self.0.id())?;
-        Ok(status.map(Into::into))
+
+        let txpool = ctx.data::<Arc<TxPool>>().unwrap().pool();
+
+        let transaction_in_pool = txpool.find_one(&self.0.id()).await;
+
+        if transaction_in_pool.is_some() && db.get_tx_status(&self.0.id()).is_err() {
```
Voxelot, line 224: I think its a safe assumption that we don't need to check the db if it was found in the pool. 

```suggestion
        if transaction_in_pool.is_some() {
```

fuel-core/src/schema/tx/types.rs
ControlCplusControlV: My bad, I think I listened to the rustc suggestion without thinking that this wasn't a special type and just plain `Arc`

fuel-core/src/schema/tx/types.rs
ControlCplusControlV: Just tested and this check is redundant, fixed it

ControlCplusControlV: I put this 2nd check in there as I was running into an issue of `Submitted {}` being overreported, looking into it now it seems the root cause is that `submit_tx` isn't clearing transactions from the mempool correctly. Looking into why

fuel-core/src/schema/tx/types.rs
```diff
@@ -214,8 +216,18 @@ impl Transaction {
...
         let db = ctx.data_unchecked::<Database>();
-        let status = db.get_tx_status(&self.0.id())?;
-        Ok(status.map(Into::into))
+
+        let txpool = ctx.data::<Arc<TxPool>>().unwrap().pool();
+
+        let transaction_in_pool = txpool.find_one(&self.0.id()).await;
+
+        if transaction_in_pool.is_some() {
+            let time = transaction_in_pool.unwrap().submited_time();
```
Voxelot, line 225: [nit] use if-let syntax to avoid .is_some followed by unwrap.

Voxelot (approved):

PR #265, opened 2022-04-15
chainspec deployment config + chainspec ergonomics

Voxelot:
Now that we are moving towards supporting fully validated utxos in deployed environments, we needed a way to configure a chainspec that has preconfigured spendable coins.

- Setup a k8s configmap to load the chainspec into the container (this will have a 1mib cap)
- Configures a conventional path for loading chainspecs from when using the deployment scripts
- Exposes more options for running the node, making it possible to enable utxo validation from the deployment `.env`

misc fixes:
 - Make graphql API show the actual chain name from the chainspec instead of a hardcoded name
 - Allow hex numbers in the chainspec to be of variable length and automatically pad them.
 - Ensure line numbers are available in machine readable (JSON) logs
 - Make the pretty logs a little more compact
 - Improve error messages when the chainspec json isn't able to be parsed

This was deployed to `node.swayswap.io` where the genesis coin with value of `1<<50` can be seen with the following query:

```
curl 'https://node.swayswap.io/graphql' -H 'Accept-Encoding: gzip, deflate, br' -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Connection: keep-alive' -H 'DNT: 1' -H 'Origin: https://node.swayswap.io' --data-binary '{"query":"# Write your query or mutation here\n{\n  coins(first: 10, filter: {owner : \"0xf1e92c42b90934aa6372e30bc568a326f6e66a1a0288595e6e3fbd392a4f3e6e\", assetId:\"0x0000000000000000000000000000000000000000000000000000000000000000\"}) {\n    edges {\n      node {\n        amount,\n        assetId\n      }\n    }\n  }\n}"}' --compressed
```


fuel-core/Cargo.toml
```diff
@@ -59,7 +59,7 @@ tokio = { version = "1.8", features = ["macros", "rt-multi-thread"] }
 tower-http = { version = "0.2.1", features = ["set-header", "trace"] }
 tower-layer = "0.3"
 tracing = "0.1"
-tracing-subscriber = { version = "0.2", features = ["env-filter", "json"] }
+tracing-subscriber = { version = "0.3", features = ["env-filter", "json", "ansi"] }
```
adlerjohn, line 62: alphabetize features

adlerjohn (commented):
Looks fine to me on the dependencies end. Will defer to others for the infra stuff.

uberscott (approved):
It looks good.  And I was hoping to find something to gripe about.

deployment/charts/templates/fuel-core-deploy.yaml
```diff
@@ -72,10 +94,15 @@ spec:
...
+            - name: config
+              mountPath: "/mnt/config"
           env:
             - name: HUMAN_LOGGING
               value: {{ .Values.app.human_logging | quote }}
       volumes:
         - name: {{ .Values.app.volume.pvname }}
           persistentVolumeClaim:
             claimName: {{ .Values.app.volume.claimname }}
+        - name: config
```
rfuelsh, line 106: @Voxelot - this looks right but I am not sure, as I never referred configmaps in this fashion this way

I would like to test just to confirm everything makes sense

typically I use

`spec:
  containers:
    - name: env-var-configmap
      image: nginx:1.7.9 
      envFrom:
        - configMapRef:
            name: example-configmap`

the envFrom to reference the configmap, but again your code may work too.

deployment/test_chainspec.json
```diff
@@ -0,0 +1,16 @@
...
+  "initial_state": {
+    "coins": [
+      {
+        "owner": "0xf1e92c42b90934aa6372e30bc568a326f6e66a1a0288595e6e3fbd392a4f3e6e",
+        "amount": "0x04000000000000",
+        "asset_id": "0x0000000000000000000000000000000000000000000000000000000000000000"
+      }
+    ]
+  }
+}
```
rfuelsh, line 16: i guess something i have to correct myself on all the time- adding newline (i always forget)

deployment/charts/templates/fuel-core-deploy.yaml
Voxelot: envFrom would set the chainspec file as an env var. But since the chainspec could be quite large, I don't think it would fit into an env var. This approach simply mounts the configmap as a file volume on the container which allows it to be a much larger size.

rfuelsh (approved):
great work on helm side!

PR #266, opened 2022-04-15
fuel-gql-client support for "https://"

Voxelot:
Fuel gql client accepted a string for instantiation but always parsed it as a socket addr so that it could be formatted in a way that is guaranteed to be the correct route for the fuel node. The problem with this approach is that it always assumes the client URL will be on `http`. This change allows the string to be directly passed in and parsed as the URL of the fuel-core graphql endpoint, allowing users to set `https` in the prefix if needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2051,6 +2051,7 @@ dependencies = [
 name = "fuel-gql-client"
 version = "0.5.0"
 dependencies = [
+ "anyhow",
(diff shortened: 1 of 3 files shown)


Voxelot:
We will need this for rust sdk users hitting hosted environments behind SSL certs @digorithm

adlerjohn (approved):
utACK

Issue #206, opened 2022-03-08
refactor: accept an `Option` for creating a new `FuelClient` from URL


iqdecay:

Voxelot:
I don't think it makes sense to support this from the fuel core side. Why not just have a different constructor for contracts that doesn't require a client?

ControlCplusControlV:
Closing since this has fallen inactive and @Voxelot has indicated this is a won't fix

PR #268, opened 2022-04-15
FuelClient can be converted from an Option

amadeusluntz:
Tried my hand at #206

Keeping the generic `From` impl and having another one that takes an `Option` doesn't seem doable to me, at least not without [this](https://github.com/rust-lang/rust/issues/35121) and [this](https://github.com/rust-lang/rust/issues/35121) rfc merged.... or maybe with macros that expand to generic-like code but allow for overwriting (?)... or maybe I'm missing something obvious.

The one (dumb) solution I came upon is noticing that `FuleClient` was always converted from `net::SocketAddr` ... so I just removed the generic behavior to allow for the `Option` instead. However, it's unclear from the issue what the desired behavior for `None` is (currently it just panics).

80% sure I'm missing the point of the issue, in which case I'd love to hear it, then I could also test it properly (currently just added a random `Some` in another unittest to check if it works in principle)

### DIFF
--- a/fuel-client/src/client.rs
+++ b/fuel-client/src/client.rs
@@ -35,17 +35,25 @@ impl FromStr for FuelClient {
     }
 }
 
-impl<S> From<S> for FuelClient
-where
-    S: Into<net::SocketAddr>,
+impl From<net::SocketAddr> for FuelClient
 {
-    fn from(socket: S) -> Self {
-        let url = format!("http://{}/graphql", socket.into())
+    fn from(socket: net::SocketAddr) -> Self {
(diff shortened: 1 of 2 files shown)


Voxelot:
While this PR faithfully attempts to accomplish the linked issue, I'm not sure if it it really makes sense to use an optional URL for creating a client. There's no sensible usecase for creating a client from a None I can think of.

amadeusluntz:
That makes perfect sense, hence why I was confused about behavior in the `None` branch, no worries, was just an attempt to see if I can get the library and tests working (and also a bit of a refresh on rust generics for me, which is always good)

PR #258, opened 2022-04-08
Txpool utxo

rakita:
Txpool will search in utxo table for dependency check.

Moved `BlockHeader`, `Block`s, `Coin` to `fuel-core-interfaces`

PR #270, opened 2022-04-15
serialize gossipsub messages

leviathanbeak:
solves #212

### DIFF
--- a/fuel-p2p/src/gossipsub/messages.rs
+++ b/fuel-p2p/src/gossipsub/messages.rs
@@ -1,0 +1,24 @@
+use serde::{Deserialize, Serialize};
+use std::io;
+
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
+pub enum GossipsubMessage {
+    BroadcastNewTx,
+    BroadcastNewBlock,
+}
+
+impl GossipsubMessage {
+    pub fn encode(&self) -> Result<Vec<u8>, io::Error> {
+        match bincode::serialize(&self) {
+            Ok(encoded_data) => Ok(encoded_data),
+            Err(e) => Err(io::Error::new(io::ErrorKind::Other, e.to_string())),
+        }
+    }
+
+    pub fn decode(encoded_data: &[u8]) -> Result<Self, io::Error> {
+        match bincode::deserialize(encoded_data) {
+            Ok(decoded_data) => Ok(decoded_data),
+            Err(e) => Err(io::Error::new(io::ErrorKind::Other, e.to_string())),
+        }
+    }
+}
--- a/fuel-p2p/src/service.rs
+++ b/fuel-p2p/src/service.rs
@@ -1,6 +1,7 @@
 use crate::{
     behavior::{FuelBehaviour, FuelBehaviourEvent},
     config::{build_transport, P2PConfig},
+    gossipsub::messages::GossipsubMessage as FuelGossipsubMessage,
     peer_info::PeerInfo,
     request_response::messages::{
         ReqResNetworkError, RequestError, RequestMessage, ResponseError, ResponseMessage,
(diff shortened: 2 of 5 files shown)


fuel-p2p/src/gossipsub/messages.rs
```diff
@@ -0,0 +1,24 @@
+use serde::{Deserialize, Serialize};
+use std::io;
+
+#[derive(Serialize, Deserialize, PartialEq, Debug)]
+pub enum GossipsubMessage {
+    BroadcastNewTx,
+    BroadcastNewBlock,
+}
+
+impl GossipsubMessage {
```
Voxelot, line 10: I think we should consider a trait-based approach here so we could configure different codecs besides bincode in the future.

vlopes11: Agreed there, I don't think we can encapsulate here all the variants that will be trafficked in the network.

We should, instead, attempt to parse per domain. In the case of consensus events, it would be something that encapsulates this struct: https://github.com/FuelLabs/fuel-bft/blob/bb6b7dc5ad66905ad9d15c8a1fe35329bc37e770/src/vote.rs#L10-L17

leviathanbeak: @vlopes11 please refer to https://fuellabs.slack.com/archives/C01L1HUJSL8/p1650294323378019 
the correct messages will be defined in another PR since I was blocked by #258 (the core-interfaces)

vlopes11: Perfect! Do we have a tracking issue for that?

leviathanbeak: @vlopes11 I will create one, as a matter of fact the PR will come in asap as well

leviathanbeak: @Voxelot I have made those functions generic so that they accept any kind of Deserializer and Serializer

fuel-p2p/src/codecs/bincode.rs
```diff
@@ -10,34 +17,18 @@ use libp2p::{
...
-
-impl ProtocolName for MessageExchangeBincodeProtocol {
-    fn protocol_name(&self) -> &[u8] {
-        REQUEST_RESPONSE_PROTOCOL_ID
-    }
-}
-
-#[derive(Debug, Clone)]
-pub struct MessageExchangeBincodeCodec {}
+pub struct BincodeCodec {}
```
Voxelot, line 21: [nit]
```suggestion
pub struct BincodeCodec;
```

fuel-p2p/src/service.rs
```diff
@@ -44,7 +46,7 @@ impl FuelP2PService {
 
         // configure and build P2P Serivce
         let transport = build_transport(local_keypair.clone()).await;
-        let behaviour = FuelBehaviour::new(local_keypair, &config);
+        let behaviour = FuelBehaviour::new(local_keypair, &config, BincodeCodec {});
```
Voxelot, line 49: [nit]
```suggestion
        let behaviour = FuelBehaviour::new(local_keypair, &config, BincodeCodec);
```

fuel-p2p/src/service.rs
```diff
@@ -28,7 +30,7 @@ pub struct FuelP2PService {
     /// Store the local peer id
     pub local_peer_id: PeerId,
     /// Swarm handler for FuelBehaviour
-    swarm: Swarm<FuelBehaviour>,
+    swarm: Swarm<FuelBehaviour<BincodeCodec>>,
```
Voxelot, line 33: This is much better and way easier to ensure all aspects of the networking is using the same codec.

Voxelot (approved):

Voxelot (approved):

PR #275, opened 2022-04-19
Add txpool TxInfo time submitted

rakita:


rakita:
@ControlCplusControlV this adds time to txpool so it can be fetched from txpool for status info, and added `find_one` function to simplify some things.

Related to this: https://github.com/FuelLabs/fuel-core/pull/263

ControlCplusControlV (approved):
Both of these are very helpful, thank you! LGTM and with the way its implemented it looks like adding any additional information would be easy. Will refactor my changes to use this once merged

fuel-txpool/src/txpool.rs
```diff
@@ -53,7 +54,7 @@ impl TxPool {
         }
         // check and insert dependency
         let rem = self.by_dependency.insert(&self.by_hash, db, &tx).await?;
-        self.by_hash.insert(tx.id(), tx.clone());
+        self.by_hash.insert(tx.id(), TxInfo::new(tx.clone()));
         self.by_gas_price.insert(&tx);
```
Voxelot, line 58: Should we normalize these secondary indexes to just map to the tx hash instead of arc's?

Voxelot (approved):

fuel-txpool/src/txpool.rs
rakita: hm, maybe, i wrap it around `Arc` because i can clone it easily when i want to send list of transaction to block producer to use and i dont need to do deep copy, and second one is that it is used in multiple parts `by_pair` and  `by_hash`

PR #280, opened 2022-04-26
Fix spurious TransactionIdCollision errors

Voxelot:
Replace usage of a badly designed rocksdb api for checking if a key exists in a column. It was causing many false positives and causing transactions to randomly fail.


AlicanC (approved):
Tested with SwaySwap and LGTM!

PR #282, opened 2022-04-27
Update docker to rust 1.60 & productionalize

Voxelot:
- Use latest stable rust
- Update to latest ubuntu LTS (22.04)
- Cargo chef to cache cargo dependencies as layers

Caching will greatly improve the developer workflow for external development environments such as swayswap.

### DIFF
--- a/deployment/Dockerfile
+++ b/deployment/Dockerfile
@@ -1,25 +1,33 @@
 # Stage 1: Build
-FROM rust:1.59.0 as builder
-
-ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
-
+FROM lukemathwalker/cargo-chef:latest-rust-1.60 as chef
+WORKDIR /build/
 # hadolint ignore=DL3008
 RUN apt-get update && \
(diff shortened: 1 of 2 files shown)


deployment/Dockerfile
```diff
@@ -29,21 +37,21 @@ ENV IP="${IP}"
...
-    apt-get install -y --no-install-recommends \
-    curl \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*
-
 WORKDIR /root/
 
+RUN apt-get update -y \
+    && apt-get install -y --no-install-recommends ca-certificates \
+    # Clean up
```
adlerjohn, line 44: Are you sure this comment works? IIRC in bash you can't do this with `\` multi-line commands, but it might be okay if you use `&&` for multiple commands.

Voxelot: ran the dockerfile locally without any issues

adlerjohn (commented):
lgtm, but will defer to other who know Docker

PR #283, opened 2022-04-27
Use lld for linking

Voxelot:
Setup the fuel-core project to use lld as the linker, which is roughly twice as fast as the default linker.

For fuel-core CI, this makes the build ~2 minutes faster.

before: https://github.com/FuelLabs/fuel-core/runs/6180962633?check_suite_focus=true
after: https://github.com/FuelLabs/fuel-core/pull/283/checks

also cleans up openssl related stuff, since it's no longer used.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,2 +1,22 @@
 [alias]
 xtask = "run --package xtask --"
+
+# On Windows
+# ```
+# cargo install -f cargo-binutils
+# rustup component add llvm-tools-preview
+# ```
+[target.x86_64-pc-windows-msvc]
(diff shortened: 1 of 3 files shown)


adlerjohn (commented):
Does the user need to always use `lld`, or is it optional just for performance? Just thinking if we should introduce extra system dependencies that _need_ to be installed.

vlopes11:
Is there any way we can make this optional? So its a choice for the user to use lld or not; and if not, he will just get slower - but working - builds.

Also, maybe lets add some docs on why we chose to use lld?

Voxelot:
I'm looking into ways to use cargo features at compile time to selectively enable this

PR #289, opened 2022-04-27
Pin async-graphql 2.9 again

Voxelot:
don't use newer version of async-graphql until newer release is fixed


adlerjohn (commented):
Add a comment on why this is pinned, otherwise it looks like an error since we haven't been the most diligent with dep versions.

adlerjohn (approved):

PR #292, opened 2022-04-28
cleanup unused deps and scripts

Voxelot:
remove stray refs to llvm and openssl, cleanup unused k8s scripts


README.md
```diff
@@ -15,25 +15,23 @@ If you are interested in contributing to Fuel, see our [CONTRIBUTING.md](CONTRIB
...
 brew update
-brew install openssl cmake llvm
+brew install cmake
 ```
 
 ###### Debian
 ```bash
 apt update
-apt install -y cmake pkg-config libssl-dev git gcc build-essential git clang libclang-dev llvm
+apt install -y cmake pkg-config git gcc build-essential git clang libclang-dev
```
adlerjohn, line 29: Dup `git`. Aside: do we actually need all of these if we first install `build-essential`? I feel like `gcc` should be included in `build-essential` intuitively without actually looking into it.
```suggestion
apt install -y cmake pkg-config gcc build-essential git clang libclang-dev
```

Voxelot: based on some cursory searching it seems like you're correct.

PR #293, opened 2022-04-28
Update client URL parsing

Voxelot:
Before when creating the fuel-gql-client it was expecting the URL to also contain the expected endpoint `/graphql`. This PR allows users to just set a base URL for the node, and then the client will automatically set the correct path to the api.


adlerjohn (approved):
utACK

adlerjohn (approved):
utACK

PR #297, opened 2022-04-28
contract errors in txpool

Voxelot:


Cargo.lock
```diff
@@ -1593,6 +1593,7 @@ version = "0.0.0"
 dependencies = [
  "chrono",
  "fuel-core",
+ "fuel-crypto",
```
adlerjohn, line 1596: Why is the lockfile being updated? If it wasn't updated in a previous PR, can you separate this into a different PR that can be reviewed independently?

Voxelot: because I needed to add fuel-crypto as a dependency to the integration tests in order to write the test needed to repro the issue.

adlerjohn: Oh I see now, I missed the manifest file change.

Voxelot:
closing as this was just used to repro an issue which has since been fixed

PR #298, opened 2022-04-29
Fix contract output cleanup in txpool dependency graph

Voxelot:
When removing a transaction from the txpool that had a contract output, it was recursing on itself after it was already removed from the dependencies causing an unexpected failure. This fix makes it so that only contract created outputs will cause any sort of recursive cleanup, and ensures the tx can't recurse on itself. I've also replaced several panics with warns for now.

### DIFF
--- a/fuel-core/src/service.rs
+++ b/fuel-core/src/service.rs
@@ -145,6 +145,7 @@ impl FuelService {
             for coin in coins {
                 let utxo_id = UtxoId::new(
                     coin.tx_id.unwrap_or_default(),
+                    // TODO: come up with utxo id scheme if there are > 255 coins with default tx id
                     coin.output_index.map(|i| i as u8).unwrap_or_else(|| {
                         generated_output_index += 1;
                         generated_output_index
@@ -169,8 +170,7 @@ impl FuelService {
     fn init_contracts(db: &mut Database, state: &StateConfig) -> Result<(), std::io::Error> {
         // initialize contract state
         if let Some(contracts) = &state.contracts {
-            let mut generated_output_index = 0;
-            for contract_config in contracts {
+            for (generated_output_index, contract_config) in contracts.iter().enumerate() {
                 let contract = Contract::from(contract_config.code.as_slice());
                 let salt = contract_config.salt;
                 let root = contract.root();
@@ -186,9 +186,9 @@ impl FuelService {
                 let _ = Storage::<ContractId, UtxoId>::insert(
                     db,
                     &contract_id,
-                    &UtxoId::new(Default::default(), generated_output_index),
+                    // TODO: come up with utxo id scheme if there are > 255 genesis contracts
+                    &UtxoId::new(Default::default(), generated_output_index as u8),
                 )?;
-                generated_output_index += 1;
                 Self::init_contract_state(db, &contract_id, contract_config)?;
                 Self::init_contract_balance(db, &contract_id, contract_config)?;
             }
--- a/fuel-core/src/service.rs
+++ b/fuel-core/src/service.rs
@@ -5,8 +5,10 @@ use crate::tx_pool::TxPool;
 use fuel_storage::{MerkleStorage, Storage};
 use fuel_tx::UtxoId;
 use fuel_types::{AssetId, Bytes32, ContractId, Salt, Word};
+use fuel_vm::consts::WORD_SIZE;
 use fuel_vm::prelude::Contract;
 pub use graph_api::start_server;
+use itertools::Itertools;
 #[cfg(feature = "rocksdb")]
 use std::io::ErrorKind;
 use std::{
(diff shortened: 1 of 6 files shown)


Voxelot:
fixes #297

fuel-txpool/src/containers/dependency.rs
```diff
@@ -435,75 +439,72 @@ impl Dependency {
...
+                    if let Some(state) = self.coins.remove(&utxo).map(|c| c.is_spend_by) {
+                        // there may or may not be any dependents for this coin output
+                        if let Some(spend_by) = state {
+                            let rem_tx = txs.get(&spend_by).expect("Tx should be present in txs");
+                            removed_transactions.extend(
+                                self.recursively_remove_all_dependencies(txs, rem_tx.tx().clone()),
+                            );
+                        }
                     } else {
-                        panic!("InputIndex for contract should be always correct");
```
rakita, line 458: I prefere to panic here

fuel-txpool/src/containers/dependency.rs
```diff
@@ -435,75 +439,72 @@ impl Dependency {
...
-                            self.recursively_remove_all_dependencies(txs, rem_tx.tx().clone()),
-                        );
+                Output::ContractCreated { contract_id, .. } => {
+                    // remove any other pending txs that depend on our yet to be created contract
+                    if let Some(contract) = self.contracts.remove(contract_id) {
+                        for spend_by in contract
+                            .used_by
+                            .into_iter()
+                            // skip the current tx
+                            .filter(|used_by| used_by != &id)
```
rakita, line 481: this `.filter` is redundent as it is should not be possible that tx uses its own Output::ContractCreated

fuel-txpool/src/containers/dependency.rs
```diff
@@ -512,15 +513,15 @@ impl Dependency {
...
-                    {
-                        let state = self
-                            .contracts
-                            .get_mut(contract_id)
-                            .expect("Contract should be present");
+                    if let Some(state) = self.contracts.get_mut(contract_id) {
                         state.used_by.remove(&tx.id());
-                        // if contract list is empty, remove contract from dependency.
+                        // if contract list is empty, flag contract state for removal.
                         if state.used_by.is_empty() {
```
rakita, line 525: while looking at it, this check is missing  a `&& state.depth == 0` as only contracts loaded from db need to be remove from dependencies in this stage.

fuel-txpool/src/containers/dependency.rs
```diff
@@ -435,75 +439,72 @@ impl Dependency {
...
 
-        // set all inputs as unspend.
+        // remove this transaction as a dependency of its' inputs.
         for input in tx.inputs() {
             match input {
                 Input::Coin { utxo_id, .. } => {
+                    // Input coin cases
+                    // 1. coin state was already removed if parent tx was also removed, no cleanup required.
+                    // 2. coin state spent_by needs to be freed from this tx if parent tx isn't being removed
+                    // 3. coin state can be removed if this is a database coin, as no other txs are involved.
```
rakita, line 500: This feels cleaner

fuel-txpool/src/containers/dependency.rs
rakita: Is there maybe a reason why you put it there?

fuel-txpool/src/containers/dependency.rs
Voxelot: I think this case isn't so egregious that we need to crash the node. I'd rather come up with a way to reset the mempool if it's unhappy about its' own state management than to panic and crash the rest of the node.

fuel-txpool/src/containers/dependency.rs
```diff
@@ -246,6 +247,8 @@ impl Dependency {
                                 .expect("Tx should be always present in txpool");
                             // compare if tx has better price
                             if txpool_tx.gas_price() > tx.gas_price() {
+                                // TODO: should we consider the value of all dependant txs to avoid
+                                //       large graphs being cancellable by single tx with a slightly higher gas price?
```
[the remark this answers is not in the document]
Voxelot: ah good point, I was thinking more in terms of someone griefing a bunch of queue txs on a contract creation output, since anyone could observe and then make the same contract id output. However, that's a totally different path than this part.

fuel-txpool/src/containers/dependency.rs
Voxelot: good catch, added this case and also refactored to use a helper `is_in_database()` to make the code more readable.

fuel-core/src/service.rs
```diff
@@ -145,6 +145,7 @@ impl FuelService {
             for coin in coins {
                 let utxo_id = UtxoId::new(
                     coin.tx_id.unwrap_or_default(),
+                    // TODO: come up with utxo id scheme if there are > 255 coins with default tx id
```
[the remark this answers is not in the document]
Voxelot: done

rakita (approved):
Lgtm

PR #299, opened 2022-04-29
Cleanup all removed dependencies

rakita:
This is bug where we didnt cleanup transaction if it is removed as dependency.


fuel-txpool/src/txpool.rs
```diff
@@ -99,10 +99,14 @@ impl TxPool {
     /// remove transaction from pool needed on user demand. Low priority
     pub fn remove_by_tx_id(&mut self, tx_id: &TxId) -> Vec<ArcTx> {
         if let Some(tx) = self.by_hash.remove(tx_id) {
-            self.by_gas_price.remove(tx.tx());
-            return self
+            let removed = self
                 .by_dependency
                 .recursively_remove_all_dependencies(&self.by_hash, tx.tx().clone());
+            for remove in removed.iter() {
```
Voxelot, line 105: [nit] this looks slightly cleaner
```suggestion
            for remove in &removed {
```

Voxelot (approved):

PR #301, opened 2022-04-30
Configurable service vars

Voxelot:
The following parameters were exposed but not fully adjustable on a per-env basis. This PR makes it so that the following vars can be configured and modified on a per deployment basis via .env files:

- human_logging
- utxo_validation
- vm_backtrace


rfuelsh (approved):
looks good!

PR #306, opened 2022-05-05
ensure transactions have at least one coin input

Voxelot:
implements https://github.com/FuelLabs/fuel-tx/issues/118 in the validator node for now, while we wait for the fuel-tx rules to propagate.

### DIFF
--- a/fuel-core/src/executor.rs
+++ b/fuel-core/src/executor.rs
@@ -8,7 +8,7 @@ use fuel_asm::Word;
 use fuel_merkle::{binary::MerkleTree, common::StorageMap};
 use fuel_storage::Storage;
 use fuel_tx::{
-    Address, AssetId, Bytes32, Input, Output, Receipt, Transaction, UtxoId, ValidationError,
+    Address, AssetId, Bytes32, Input, Output, Receipt, Transaction, TxId, UtxoId, ValidationError,
 };
 use fuel_types::{bytes::SerializableVec, ContractId};
 use fuel_vm::{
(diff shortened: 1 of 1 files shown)


adlerjohn (commented):
Is there a reason not to bump fuel-tx? My concern is that the error variant here is being defined in the wrong place potentially.

Voxelot:
This is simply a convenience to avoid extra churn, but I don't think there's anything majorly blocking a new fuel-tx release if we'd rather do that.

Voxelot:
@adlerjohn it looks like there are also a lot of tests that will break in fuel-{tx, vm, core} if this is required by default, making that a bigger lift. This PR selectively performs this validation only when utxo-validation is enabled, and the tests that are written for utxo-validation == true already have coins to cover gas fees.

adlerjohn (commented):
clippy

adlerjohn (commented):
Will defer to @vlopes11

fuel-core/src/executor.rs
```diff
@@ -266,6 +268,18 @@ impl Executor {
         Ok(())
     }
 
+    /// Verify the transaction has at least one coin.
+    ///
+    /// TODO: This verification really belongs in fuel-tx, and can be removed once
+    ///       https://github.com/FuelLabs/fuel-tx/issues/118 is resolved.
+    fn verify_tx_has_at_least_one_coin(&self, tx: &Transaction) -> Result<(), Error> {
+        if tx.inputs().iter().filter(|input| input.is_coin()).count() == 0 {
```
vlopes11, line 276: Optional nit: this looks like `!iter().filter(Input::is_coin).next().map(|_| ()).ok_or(TransactionValidityError::NoCoinInput(tx.id()).into())`

That way, instead of counting the iterators, we just lazy-check the next one - that already satisfies the count != 0 criteria

vlopes11 (approved):
LGTM

fuel-core/src/executor.rs
Voxelot: That's maybe slightly less readable but seems like a good technique. Maybe we can do it that way when we impl on the fuel-tx side.

PR #307, opened 2022-05-06
Add ability to set minimum required gas / byte fees on transactions

Voxelot:
Adds the following transaction pool configuration parameters to fuel-core:

- minimum gas price
- minimum byte price

```rust
/// The minimum allowed gas price
pub min_gas_price: u64,

/// The minimum allowed byte price
pub min_byte_price: u64,
```

These are enforced in the transaction pool rather than the consensus rules, since fee market management isn't the job of consensus. 

It also does some miscellaneous cleanup of unneeded Arcs.

### DIFF
--- a/fuel-txpool/src/txpool.rs
+++ b/fuel-txpool/src/txpool.rs
@@ -1,27 +1,30 @@
-use crate::containers::dependency::Dependency;
-use crate::Error;
-use crate::{containers::price_sort::PriceSort, types::*, Config};
-use fuel_core_interfaces::model::{ArcTx, TxInfo};
-use fuel_core_interfaces::txpool::TxPoolDb;
+use crate::{
+    containers::{dependency::Dependency, price_sort::PriceSort},
+    types::*,
+    Config, Error,
+};
+use fuel_core_interfaces::{
+    model::{ArcTx, TxInfo},
+    txpool::TxPoolDb,
+};
 use std::collections::HashMap;
-use std::sync::Arc;
 
 #[derive(Debug, Clone)]
 pub struct TxPool {
     by_hash: HashMap<TxId, TxInfo>,
     by_gas_price: PriceSort,
     by_dependency: Dependency,
-    config: Arc<Config>,
+    config: Config,
 }
 
 impl TxPool {
-    pub fn new(config: Arc<Config>) -> Self {
+    pub fn new(config: Config) -> Self {
         let max_depth = config.max_depth;
         Self {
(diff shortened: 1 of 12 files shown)


adlerjohn (commented):
Seems okay, but will defer to @rakita

ControlCplusControlV:
Would it be worth adding some tests to cover if fees change? Gas price is usually dynamic so it may be nice to have some tests ensuring a tx isn't valid if gas price increases, or is now valid if gas price drops. Although as execution isn't decoupled yet that may be harder to test (as you can't insert a tx, raise price, then make sure it doesn't get executed)

Voxelot:
These are just static minimums, so we only need to check the price at insertion time. However, we will need to recheck transactions eventually once we have additional dynamic pricing.

fuel-txpool/src/txpool.rs
```diff
@@ -110,6 +120,30 @@ impl TxPool {
...
+    fn verify_tx_min_byte_price(&mut self, tx: &Transaction) -> Result<(), Error> {
+        if tx.byte_price() < self.config.min_byte_price {
+            return Err(Error::NotInsertedBytePriceTooLow);
+        }
+        Ok(())
+    }
+
+    fn verify_tx_min_byte_price_ratio(&mut self, tx: &Transaction) -> Result<(), Error> {
+        if self.config.max_gas_per_byte_price > 0
+            && tx.byte_price() * self.config.max_gas_per_byte_price < tx.gas_price()
```
rakita, line 140: maybe we should update fuel-spec to address this, it is currently set as `bytePrice != gasPrice * GAS_PER_BYTE`
`

rakita (approved):
lgtm

fuel-txpool/src/txpool.rs
Voxelot: Removed this since configurable min fees should be sufficient for now. We can revisit multi-modal pricing once we have legit block production and more interesting resource contention issues.

adlerjohn (approved):
lgtm piggy-backing off @rakita's review since subsequent commit only addressed review comment.

rakita:
> lgtm piggy-backing off @rakita's review since subsequent commit only addressed review comment.

Don't include me, `min_byte_price` now does not make sense other than pushing the agenda.

Voxelot:
It's necessary until we have https://github.com/FuelLabs/fuel-tx/issues/119 implemented, after which it won't really be needed until we revisit this "agenda".

PR #314, opened 2022-05-09
Add Docker image building to standard CI flow

Voxelot:
- Fixes a docker image building issue due to a [bug](https://github.com/LukeMathWalker/cargo-chef/issues/128) in cargo-chef that was triggered by the addition of the relayer
- Make docker image building a standard part of CI, so that these issues will be caught earlier. This will also open up the possibility of us having similar capabilities to GitLab [review apps](https://about.gitlab.com/blog/2016/11/22/introducing-review-apps/) from a feature branch of fuel-core (e.g. spin up a hosted env with predicates enabled)
- Also found a bug with the way our rust caching action was interacting with the new matrix style builds. The matrix commands need to be included in the cache key, otherwise the cache will be corrupted by other jobs using different feature flags. This drops our cached build times by 50% in some cases.
- Add timeout to cargo-verifications to prevent stalled jobs from running for hours.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,7 +146,7 @@ jobs:
     needs:
       - lint-toml-files
       - cargo-verifications
-      - publish
+      - publish-crates
 
     if: always()
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,8 @@ jobs:
         with:
           toolchain: stable
       - uses: Swatinem/rust-cache@v1
+        with:
+          key: "${{ matrix.command }} ${{ matrix.args }}"
       - name: ${{ matrix.command }} ${{ matrix.args }}
         uses: actions-rs/cargo@v1
         with:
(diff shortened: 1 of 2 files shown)


adlerjohn (commented):
CI failing

Voxelot:
looks like that was due to the flakey p2p test that is stalling: https://github.com/FuelLabs/fuel-core/runs/6362763697?check_suite_focus=true#step:6:161 

the new timeout made it fail faster.

.github/workflows/ci.yml
```diff
@@ -142,20 +146,28 @@ jobs:
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_NOTIFY_BUILD }}
 
-  build-publish-master-image:
+  publish-docker-image:
     needs:
       - lint-toml-files
       - cargo-verifications
-    if: github.ref == 'refs/heads/master'
+      - publish-crates
```
adlerjohn, line 153: Is this needed to depend on `publish-crates`? So long as `cargo-verification` passes, then the Docker image should be built, no?

PR #320, opened 2022-05-12
add a manual dialing of a peer and panic! if it stalls in reconnect test

leviathanbeak:
added 2 timeouts:

- on first timeout, we try to reconnect to the peer manually
- on second we panic! exit the test since the peer is unreachable


leviathanbeak:
maybe adding manual reconnect + timeout and panic! exit to each test would improve it as well

PR #323, opened 2022-05-12
fix: has next page

luizstacio:
This PR fixes an issue on `hasPreviousPage` and `hasNextPage`;

## Use case
To navigate from the first page to the last page, being the first page `the latest transactions`.

### Page 1
To do that we query transactions using;

`transactions(last: 5) {....`

###  Next -> Page 2
To go to the next page we use the `startCursor` returned from the previous query and pass it as a cursor;

`transactions(last: 5, before: "<startCursor>") {....`

### Previous -> Page 1
Now navigate to the previous page we use the `endCursor` returned from the previous query;

`transactions(first: 5, after: "<endCursor>") {....`

This returns `hasPreviousPage` and `hasNextPage` as `true`. But it should return `hasNextPage` as `false`.

### DIFF
--- a/fuel-client/src/client/schema.rs
+++ b/fuel-client/src/client/schema.rs
@@ -175,6 +175,8 @@ impl<T: Into<String>> From<PaginationRequest<T>> for ConnectionArgs {
 pub struct PaginatedResult<T, C> {
     pub cursor: Option<C>,
     pub results: Vec<T>,
+    pub has_next_page: bool,
+    pub has_previous_page: bool,
 }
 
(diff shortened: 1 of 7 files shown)


luizstacio:
Do we need tests here? I want to move from `draft` to `PR`.

Voxelot:
We should have an integ test in `fuel-tests` to verify this. It could build on or copy this existing test: https://github.com/FuelLabs/fuel-core/blob/master/fuel-tests/tests/tx.rs#L208

PR #342, opened 2022-05-19
Publish compiled fuel-core binaries on release

Voxelot:
This PR adds a new ci job which will cross-compile fuel-core binaries to various platforms when released. We had to deviate from the approach used in sway, and go with the cross framework instead because of the following rust-bindgen issues: 
- https://github.com/rust-rocksdb/rust-rocksdb/issues/550
- https://github.com/rust-lang/rust-bindgen/issues/1229

Since rocksdb also requires clang, this also required extending the default cross dockerfiles to either update or install the llvm toolchain for each target.

Tested the release process using this branch here: https://github.com/FuelLabs/fuel-dummy-test/releases/tag/v0.2.57

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -265,11 +265,8 @@ jobs:
             platform: darwin-arm
             target: aarch64-apple-darwin
     steps:
-      - name: Checkout sources
+      - name: Checkout repository
         uses: actions/checkout@v2
-        with:
-          repository: 'FuelLabs/fuel-core'
-          ref: 'cross-compiling'
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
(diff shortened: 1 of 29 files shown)


.github/workflows/ci.yml
```diff
@@ -230,3 +230,153 @@ jobs:
           notify_when: 'failure'
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_NOTIFY_BUILD }}
+
+  publish-fuel-core-binary:
+    name: Release fuel-core ${{ matrix.job.target }}
```
adlerjohn, line 235: Matrix doesn't exist here, so the name shows up literally as `Release fuel-core ${{ matrix.job.target }}`.
```suggestion
    name: Release fuel-core binaries
```

mitchmindtree (approved):
Wow, `cross` looks like a big improvement on messing around with getting different target toolchains installed locally! What an awesome project.

I won't pretend to fully understand the Docker stuff but all looks simple enough and makes sense. The README is a great help, nice one.

Voxelot:
Ideally we'd prebuild all these docker images and publish them somewhere. But building them on the fly works for now.

PR #345, opened 2022-05-20
Relayer block commit and deposit handling

rakita:
Info about fuel-relayer structure:
* `Service`: a way to initialize relayer
  * `Relayer`: the main mod that contains initial sync and handling of inbound channels from both ethereum and fuel:
    * `FinalizationQueue`: before events get finalized they are saved in memory here, we handle the finalization of events here.
      * `PendingBlocks`: Handles what block are committed to the contract and what is the best block inside fuel network. This is the place where we do packing of eth transaction and bundling of block commits.
      *  `Validators`:  Has a last finalized validator set. Handles saving of diffs to database and applying that diff current validator set.

`Log` mod contains type casting from eth log to event that we can digest. `bin/testrun.rs` is simple run with DummyDb for manual testing.

Communication channels that we have:
* Inbound `fuel_block_importer` with `NewBlockEvent`: new block event from fuel, , mostly send block commit if this client created new fuel block
* Inbound `requests` with `RelayerEvent`: request for relayer for example for validator set.
* `da_block_watcher`: do finalization on this event
* `logs_watcher`: insert logs into finalization_queue.

On the database side we have:
* `StakingDIffs` table: for every da_block it contain staking diff. Staking diff contains un/registration of validator and delegations
* `ValidatorSet` table: validator set at last finalized da block. 
* `DelegatesIndex` table: One delegation has a list of validators and delegation stakes, when undelegating or adding a new delegation, the old delegation needs to be removed. This index contains list of blocks where delegation happened so that we can traverse it and get that old delegation. 
* `DepositCoin`: table of finalized coins. It will become the table of messages in the new bridging architecture.

And some DB variables:
* `validator_da_height`: Relater to `ValidatorSet` table
* `finalized_da_height`:  Updated to become `new_da_height - finalization_period`
* `last_commited_finalized_fuel_height`: this is an interesting one, as it name says it represents the last committed finalized fuel height, a block that we know is committed and finalized in Ethereum contract, for blocks after this height we will still need to have consensus votes (`SealedBlock`)

There are some models that are added as `SealedFuelBlock` and `block_importer::NewBlockEvent` enum but those are more like placeholders and will probably be changed when the sources of those items get implemented. Same as `get_sealed_block` function in db, this is something that is still WIP in fuel.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,10 +122,15 @@ jobs:
         run: |
           curl -sSLf "https://github.com/TomWright/dasel/releases/download/v1.24.3/dasel_linux_amd64" -L -o dasel && chmod +x dasel
           mv ./dasel /usr/local/bin/dasel
+          ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuel-block-executor/Cargo.toml
+          ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuel-block-importer/Cargo.toml
+          ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuel-block-producer/Cargo.toml
           ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuel-client/Cargo.toml
           ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuel-core/Cargo.toml
+          ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuel-core-bft/Cargo.toml
           ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuel-core-interfaces/Cargo.toml
           ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuel-p2p/Cargo.toml
+          ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuel-sync/Cargo.toml
           ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuel-txpool/Cargo.toml
 
       - name: Verify helm chart version
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -97,7 +97,8 @@ This is a rough outline of what a contributor's workflow looks like:
 - Write code, add test cases, and commit your work.
 - Run tests and make sure all tests pass.
 - If the PR contains any breaking changes, add the breaking label to your PR.
-- Push your changes to a branch in your fork of the repository and submit a pull request.
+- If you are part of the FuelLabs Github org, please open a PR from the repository itself.
+- Otherwise, push your changes to a branch in your fork of the repository and submit a pull request.
     * Make sure mention the issue, which is created at step 1, in the commit message.
 - Your PR will be reviewed and some changes may be requested.
     * Once you've made changes, your PR must be re-reviewed and approved.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2160,6 +2160,38 @@ dependencies = [
 ]
 
 [[package]]
+name = "fuel-block-executor"
+version = "0.8.0"
+dependencies = [
+ "anyhow",
+ "fuel-core-interfaces",
+ "fuel-types",
+ "tokio",
+]
+
+[[package]]
+name = "fuel-block-importer"
+version = "0.8.0"
+dependencies = [
+ "anyhow",
+ "fuel-core-interfaces",
+ "fuel-types",
+ "parking_lot 0.12.1",
+ "tokio",
+]
+
+[[package]]
+name = "fuel-block-producer"
+version = "0.8.0"
+dependencies = [
+ "anyhow",
+ "fuel-core-interfaces",
+ "fuel-types",
+ "parking_lot 0.12.1",
+ "tokio",
+]
+
+[[package]]
 name = "fuel-core"
 version = "0.8.0"
 dependencies = [
@@ -2176,10 +2208,15 @@ dependencies = [
  "dirs",
  "env_logger",
  "fuel-asm",
+ "fuel-block-executor",
+ "fuel-block-importer",
+ "fuel-block-producer",
+ "fuel-core-bft",
  "fuel-core-interfaces",
  "fuel-crypto",
  "fuel-merkle 0.1.1",
  "fuel-storage 0.1.0",
+ "fuel-sync",
  "fuel-tx",
  "fuel-txpool",
  "fuel-types",
@@ -2206,6 +2243,17 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "uuid",
+]
+
+[[package]]
+name = "fuel-core-bft"
+version = "0.8.0"
+dependencies = [
+ "anyhow",
+ "fuel-core-interfaces",
+ "fuel-types",
+ "parking_lot 0.12.1",
+ "tokio",
 ]
 
 [[package]]
@@ -2363,6 +2411,17 @@ name = "fuel-storage"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34b9161e86d434a93088409530a4f71f42e074b3bbcbb7a27bfe666583f92fd7"
+
+[[package]]
+name = "fuel-sync"
+version = "0.8.0"
+dependencies = [
+ "anyhow",
+ "fuel-core-interfaces",
+ "fuel-types",
+ "parking_lot 0.12.1",
+ "tokio",
+]
 
 [[package]]
 name = "fuel-tests"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,11 +2,16 @@
 # Use the new resolver to prevent dev-deps and build-deps from enabling debugging or test features in production.
 resolver = "2"
 members = [
+  "fuel-block-executor",
+  "fuel-block-importer",
+  "fuel-block-producer",
   "fuel-client",
   "fuel-core",
+  "fuel-core-bft",
   "fuel-core-interfaces",
   "fuel-p2p",
   "fuel-relayer",
+  "fuel-sync",
   "fuel-tests",
   "fuel-txpool",
   "xtask",
--- a/fuel-block-executor/Cargo.toml
+++ b/fuel-block-executor/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "fuel-block-executor"
+version = "0.8.0"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+edition = "2021"
+homepage = "https://fuel.network/"
+keywords = ["blockchain", "fuel", "fuel-vm"]
+license = "BUSL-1.1"
+repository = "https://github.com/FuelLabs/fuel-core"
+description = "Fuel Block Executor"
+
+[dependencies]
+anyhow = "1.0"
+fuel-core-interfaces = { path = "../fuel-core-interfaces", version = "0.8.0" }
+fuel-types = { version = "0.5", default-features = false }
+tokio = { version = "1.14", features = ["full"] }
--- a/fuel-block-executor/README.md
+++ b/fuel-block-executor/README.md
@@ -1,0 +1,1 @@
+# Fuel Block Executor
--- a/fuel-block-executor/src/lib.rs
+++ b/fuel-block-executor/src/lib.rs
@@ -1,0 +1,1 @@
+pub struct BlockExecutor {}
--- a/fuel-block-importer/Cargo.toml
+++ b/fuel-block-importer/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "fuel-block-importer"
+version = "0.8.0"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+edition = "2021"
+homepage = "https://fuel.network/"
+keywords = ["blockchain", "fuel", "fuel-vm"]
+license = "BUSL-1.1"
+repository = "https://github.com/FuelLabs/fuel-core"
+description = "Fuel Block Importer"
+
+[dependencies]
+anyhow = "1.0"
+fuel-core-interfaces = { path = "../fuel-core-interfaces", version = "0.8.0" }
+fuel-types = { version = "0.5", default-features = false }
+parking_lot = "0.12"
+tokio = { version = "1.14", features = ["full"] }
--- a/fuel-block-importer/README.md
+++ b/fuel-block-importer/README.md
@@ -1,0 +1,1 @@
+# Fuel Block Importer
--- a/fuel-block-importer/src/config.rs
+++ b/fuel-block-importer/src/config.rs
@@ -1,0 +1,2 @@
+#[derive(Default)]
+pub struct Config {}
--- a/fuel-block-importer/src/lib.rs
+++ b/fuel-block-importer/src/lib.rs
@@ -1,0 +1,5 @@
+pub mod config;
+pub mod service;
+
+pub use config::Config;
+pub use service::Service;
--- a/fuel-block-importer/src/service.rs
+++ b/fuel-block-importer/src/service.rs
@@ -1,0 +1,48 @@
+use crate::Config;
+use fuel_core_interfaces::block_importer::{ImportBlockBroadcast, ImportBlockMpsc};
+use parking_lot::Mutex;
+use tokio::{
+    sync::{broadcast, mpsc},
+    task::JoinHandle,
+};
+
+pub struct Service {
+    join: Mutex<Option<JoinHandle<()>>>,
+    sender: mpsc::Sender<ImportBlockMpsc>,
+    broadcast: broadcast::Sender<ImportBlockBroadcast>,
+}
+
+impl Service {
+    pub async fn new(_config: &Config, _db: ()) -> Result<Self, anyhow::Error> {
+        let (sender, _receiver) = mpsc::channel(100);
+        let (broadcast, _receiver) = broadcast::channel(100);
+        Ok(Self {
+            sender,
+            broadcast,
+            join: Mutex::new(None),
+        })
+    }
+
+    pub async fn start(&self) {
+        let mut join = self.join.lock();
+        if join.is_none() {
+            *join = Some(tokio::spawn(async {}));
+        }
+    }
+
+    pub async fn stop(&self) -> Option<JoinHandle<()>> {
+        let join = self.join.lock().take();
+        if join.is_some() {
+            let _ = self.sender.send(ImportBlockMpsc::Stop);
+        }
+        join
+    }
+
+    pub fn subscribe(&self) -> broadcast::Receiver<ImportBlockBroadcast> {
+        self.broadcast.subscribe()
+    }
+
+    pub fn sender(&self) -> &mpsc::Sender<ImportBlockMpsc> {
+        &self.sender
+    }
+}
--- a/fuel-block-producer/Cargo.toml
+++ b/fuel-block-producer/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "fuel-block-producer"
+version = "0.8.0"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+edition = "2021"
+homepage = "https://fuel.network/"
+keywords = ["blockchain", "fuel", "fuel-vm"]
+license = "BUSL-1.1"
+repository = "https://github.com/FuelLabs/fuel-core"
+description = "Fuel Block Producer"
+
+[dependencies]
+anyhow = "1.0"
+fuel-core-interfaces = { path = "../fuel-core-interfaces", version = "0.8.0" }
+fuel-types = { version = "0.5", default-features = false }
+parking_lot = "0.12"
+tokio = { version = "1.14", features = ["full"] }
--- a/fuel-block-producer/README.md
+++ b/fuel-block-producer/README.md
@@ -1,0 +1,1 @@
+# Fuel Block Producer
--- a/fuel-block-producer/src/config.rs
+++ b/fuel-block-producer/src/config.rs
@@ -1,0 +1,2 @@
+#[derive(Default)]
+pub struct Config {}
--- a/fuel-block-producer/src/lib.rs
+++ b/fuel-block-producer/src/lib.rs
@@ -1,0 +1,5 @@
+pub mod config;
+pub mod service;
+
+pub use config::Config;
+pub use service::Service;
--- a/fuel-block-producer/src/service.rs
+++ b/fuel-block-producer/src/service.rs
@@ -1,0 +1,38 @@
+use crate::Config;
+use fuel_core_interfaces::block_producer::BlockProducerMpsc;
+use parking_lot::Mutex;
+use tokio::{sync::mpsc, task::JoinHandle};
+
+pub struct Service {
+    join: Mutex<Option<JoinHandle<()>>>,
+    sender: mpsc::Sender<BlockProducerMpsc>,
+}
+
+impl Service {
+    pub async fn new(_config: &Config, _db: ()) -> Result<Self, anyhow::Error> {
+        let (sender, _receiver) = mpsc::channel(100);
+        Ok(Self {
+            sender,
+            join: Mutex::new(None),
+        })
+    }
+
+    pub async fn start(&self, _txpool: ()) {
+        let mut join = self.join.lock();
+        if join.is_none() {
+            *join = Some(tokio::spawn(async {}));
+        }
+    }
+
+    pub async fn stop(&self) -> Option<JoinHandle<()>> {
+        let join = self.join.lock().take();
+        if join.is_some() {
+            let _ = self.sender.send(BlockProducerMpsc::Stop);
+        }
+        join
+    }
+
+    pub fn sender(&self) -> &mpsc::Sender<BlockProducerMpsc> {
+        &self.sender
+    }
+}
--- a/fuel-core-bft/Cargo.toml
+++ b/fuel-core-bft/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "fuel-core-bft"
+version = "0.8.0"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+edition = "2021"
+homepage = "https://fuel.network/"
+keywords = ["blockchain", "fuel", "consensus", "bft"]
+license = "BUSL-1.1"
+repository = "https://github.com/FuelLabs/fuel-core"
+description = "Fuel Core BFT"
+
+[dependencies]
+anyhow = "1.0"
+fuel-core-interfaces = { path = "../fuel-core-interfaces", version = "0.8.0" }
+fuel-types = { version = "0.5", default-features = false }
+parking_lot = "0.12"
+tokio = { version = "1.14", features = ["full"] }
--- a/fuel-core-bft/README.md
+++ b/fuel-core-bft/README.md
@@ -1,0 +1,1 @@
+# Fuel Block Producer
--- a/fuel-core-bft/src/config.rs
+++ b/fuel-core-bft/src/config.rs
@@ -1,0 +1,2 @@
+#[derive(Default)]
+pub struct Config {}
--- a/fuel-core-bft/src/lib.rs
+++ b/fuel-core-bft/src/lib.rs
@@ -1,0 +1,5 @@
+pub mod config;
+pub mod service;
+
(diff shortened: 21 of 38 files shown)


fuel-core-interfaces/src/db.rs
```diff
@@ -66,6 +66,7 @@ impl From<KvStoreError> for InterpreterError {
 pub mod helpers {
 
     use async_trait::async_trait;
+    use chrono::Utc;
```
vlopes11, line 69: Any reason to favor `chrono` over `time`? `time` is more maintained with better interface compared to chrono, that relies in an old version of `time` (0.1)

fuel-core-interfaces/src/db.rs
```diff
@@ -140,17 +144,19 @@ pub mod helpers {
 
     #[derive(Clone, Debug)]
     pub struct Data {
+        /// Used for Storage<Address, (u64, Option<Address>)>
+        pub validators: HashMap<Address, (ValidatorStake, Option<Address>)>,
```
vlopes11, line 148: an address mapping to an address isn't obvious without external context. consider adding some documentation here explaining what is what

fuel-core-interfaces/src/db.rs
```diff
@@ -140,17 +144,19 @@ pub mod helpers {
...
+        /// variable for current validators height, at height our validator set is
+        pub validators_height: DaBlockHeight,
+        /// diffs between blocks regarding validator set and delegation stake
+        pub staking_diffs: BTreeMap<DaBlockHeight, StakingDiff>,
```
vlopes11, line 152: maybe explain a bit better here what is the purpose of this map

fuel-core-interfaces/src/db.rs
```diff
@@ -160,9 +166,17 @@ pub mod helpers {
         pub contract: HashSet<ContractId>,
         /// Dummy deposit coins.
         pub deposit_coin: HashMap<Bytes32, DepositCoin>,
+        /// variable for last commited and finalized fuel height
+        pub last_commited_finalized_fuel_height: BlockHeight,
     }
 
     impl DummyDb {
+        pub fn insert_sealed_block(&self, sealed_block: Arc<SealedFuelBlock>) {
```
vlopes11, line 174: won't this lock be async somewhere?

fuel-core-interfaces/src/model/block.rs
```diff
@@ -39,18 +54,27 @@ impl FuelBlockHeader {
         hasher.input(self.producer.as_ref());
         hasher.digest()
     }
+
+    pub fn id(&self) -> Bytes32 {
+        if let Some(ref metadata) = self.metadata {
+            metadata.id
+        } else {
+            self.hash()
```
vlopes11, line 62: shouldn't we replace the metadata once it is calculated?

fuel-core-interfaces/src/model/block.rs
```diff
@@ -88,4 +112,57 @@ impl FuelBlock {
...
+
+/// This structure is ceated as placeholder for future usage.
+/// It represent commitment for next child block
+#[derive(Clone, Debug, Default)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct FuelBlockConsensus {
+    /// required stake for next block
+    pub required_stake: u64,
+    /// Map of Validator consensus key and pair of stake and signature
+    pub validators: HashMap<Address, (ValidatorStake, Bytes32)>,
```
vlopes11, line 152: in the other struct, we map to a stake+address, and here to bytes32. why diverge the types?

fuel-core-interfaces/src/model/deposit_coin.rs
```diff
@@ -0,0 +1,20 @@
...
+    pub amount: Word,
+    pub asset_id: AssetId,
+    pub nonce: Bytes32,
+    pub deposited_da_height: DaBlockHeight,
+    pub fuel_block_spend: Option<BlockHeight>,
+}
+
+impl DepositCoin {
+    pub fn id(&self) -> Bytes32 {
+        self.nonce
```
vlopes11, line 18: why force copy? also, if the field and is public, why create a getter with a different name? maybe this will cause confusion

fuel-relayer/src/bin/testrun.rs
```diff
@@ -1,32 +1,80 @@
-use std::str::FromStr;
-
-use fuel_relayer::{Config, Service};
-
+use chrono::Utc;
```
vlopes11, line 1: committing so much to chrono could be a long term problem. reconsider using `time`

fuel-relayer/src/validators.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    /// Get validator set
+    pub async fn get(
+        &mut self,
+        da_height: DaBlockHeight,
+    ) -> Option<HashMap<Address, (u64, Option<Address>)>> {
+        // TODO apply down drift https://github.com/FuelLabs/fuel-core/issues/365
+        if self.da_height == da_height {
+            return Some(self.set.clone());
```
vlopes11, line 36: is cloning the entire set the best option here?

vlopes11 (commented):
initial round of comments

fuel-relayer/Cargo.toml
```diff
@@ -11,18 +11,24 @@ publish = false
 anyhow = "1.0"
 async-trait = "0.1"
 bytes = "1.1"
+chrono = "0.4"
 env_logger = "0.9"
-ethers-core = { git = "https://github.com/rakita/ethers-rs.git", branch = "master", package = "ethers-core", default-features = false }
-ethers-middleware = { git = "https://github.com/rakita/ethers-rs.git", branch = "master", package = "ethers-middleware", default-features = false }
-ethers-providers = { git = "https://github.com/rakita/ethers-rs.git", branch = "master", package = "ethers-providers", default-features = false, features = [
+ethers-contract = { git = "https://github.com/rakita/ethers-rs.git", branch = "master", default-features = false, features = [
```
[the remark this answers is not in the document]
adlerjohn: This should be fine for now since `fuel-relayer` isn't published. It can be tackled in a future PR since it's orthogonal to the changes at hand.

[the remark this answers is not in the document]
rakita: Currently waiting for this as there is some conflicts: https://github.com/FuelLabs/fuel-core/issues/367

fuel-core-interfaces/src/model/block.rs
rakita: Needs to be changed, good catch i missed that one

fuel-core-interfaces/src/model/deposit_coin.rs
rakita: It was some kind of placeholder as we were talking about hashing all fields in the hash but wasn't sure exactly what, but in the last iteration, we are removed `DepositCoin` and introduces bridge `Messages` so this part is going to be changed when details in contracts become known.

fuel-relayer/src/validators.rs
rakita: Yeah, this function is called in request by `fuel-bft` and the relayer need to return the copy of the requested validator set to it. 
I will implement drift in past in this commit: https://github.com/FuelLabs/fuel-core/issues/365

fuel-core-interfaces/src/model/block.rs
rakita: A new block is mostly going to be broadcasted as `Arc` in that case having `&mut` self is not going to work. This else branch is left here as to not break other parts of fuel-core by requiring `recalculate_metadata`  to be called beforehand for the id to be valid.

fuel-core/src/database.rs
```diff
@@ -263,15 +269,20 @@ impl InterpreterStorage for Database {
 
 #[async_trait]
 impl RelayerDb for Database {
```
Voxelot, line 271: Could we move all this RelayerDb impl logic into a separate mod? Would help keep the main DB file more manageable.

fuel-relayer/src/config.rs
```diff
@@ -1,40 +1,61 @@
 use ethers_core::types::{H160, H256};
+use fuel_core_interfaces::model::DaBlockHeight;
 use std::{str::FromStr, time::Duration};
 
 lazy_static::lazy_static! {
-    pub static ref ETH_ASSET_DEPOSIT : H256 = H256::from_str("0x0000000000000000000000000000000000000000000000000000000000000001").unwrap();
-    pub static ref ETH_VALIDATOR_DEPOSIT : H256 = H256::from_str("0x0000000000000000000000000000000000000000000000000000000000000002").unwrap();
-    pub static ref ETH_VALIDATOR_WITHDRAWAL : H256 = H256::from_str("0x0000000000000000000000000000000000000000000000000000000000000003").unwrap();
-    pub static ref ETH_FUEL_BLOCK_COMMITED : H256 = H256::from_str("0x0000000000000000000000000000000000000000000000000000000000000004").unwrap();
+    /// Hash over solidity function: `DepositMade(uint32,address,address,uint8,uint256,uint256)`
```
Voxelot, line 6: I think we should file a tech debt issue to replace this with the actual hashing code, so that it's easier to update these in the future.

fuel-core-interfaces/src/db.rs
Voxelot: I'm in favor of finding an alternative to chrono, but this can be deferred to a tech debt item.

fuel-relayer/src/pending_blocks.rs
```diff
@@ -0,0 +1,617 @@
...
+    fn handle_block_commit_append(
+        &mut self,
+        block_root: Bytes32,
+        height: BlockHeight,
+        da_height: DaBlockHeight,
+    ) {
+        if self.pending_block_commits.is_empty() {
+            let lcffh = self.last_commited_finalized_fuel_height;
+            if lcffh + 1u64.into() != height {
+                error!("We have missing logs from LCFFH {lcffh} to new height {height}")
```
Voxelot, line 213: Do we expect to receive logs from every block even without any activity on the L1 contract?

fuel-relayer/src/pending_blocks.rs
```diff
@@ -0,0 +1,617 @@
...
+
+/// Pending Fuel Blocks waiting to be finalized inside client. Until then
+/// there is possibility that they are going to be reverted
+pub struct PendingBlocks {
+    signer: LocalWallet,
+    contract_address: H160,
+    // Pending block commits seen on DA layer and waiting to be finalized
+    pending_block_commits: VecDeque<PendingBlock>,
+    // Highest known chain height, used to check if we are seeing lag between block commits and our fule chain
+    chain_height: BlockHeight,
```
Voxelot, line 31: It's confusing to keep track of which is da and which is fuel when we sometimes use `da_height` vs `height`, and then use `chain_height` vs `fuel_height` here. Without seeing these two vars side by side, I'm not sure what chain_height means when it's being used.

```suggestion
    best_da_height: BlockHeight,
```

fuel-relayer/src/pending_blocks.rs
```diff
@@ -0,0 +1,617 @@
...
+        }
+
+        let front_height = self.pending_block_commits.front().unwrap().block_height;
+        let back_height = self.pending_block_commits.back().unwrap().block_height;
+
+        if height < back_height {
+            // ignore block that are not inside pending queue.
+            error!("All pending block commits should be present in block queue. height:{height} last_known:{back_height}");
+        } else if height > front_height {
+            error!("Something unexpected happened.Reverted block commits are not something found in the future.");
```
Voxelot, line 287: Related to my other point, could this occur fairly regularly if the L1 contract doesn't emit any logs for a few blocks?

fuel-relayer/src/pending_blocks.rs
```diff
@@ -0,0 +1,617 @@
...
+        } else {
+            // happy path. iterate over pending blocks and set it as commited.
+            for pending in self.pending_block_commits.iter_mut() {
+                if pending.block_height == height {
+                    if pending.reverted {
+                        error!("We received block {height} commit that was already reverted");
+                    }
+                    pending.da_height = da_height;
+                    pending.reverted = true;
+                    pending.block_height = height;
```
Voxelot, line 297: [nit] seems unnecessary to set block_height here if it doesn't change

fuel-relayer/src/pending_blocks.rs
```diff
@@ -0,0 +1,617 @@
...
+            // iterate over pending blocks and set reverted commit as commited.
+            for pending in self.pending_block_commits.iter_mut() {
+                if pending.block_height == height {
+                    if !pending.reverted {
+                        error!("We received block {height} commit that was not reverted")
+                    }
+                    pending.da_height = da_height;
+                    pending.reverted = false;
+                    pending.block_height = height;
+                    pending.block_root = block_root;
```
Voxelot, line 259: unnecessary variable setting

```suggestion
                    pending.reverted = false;
```

fuel-relayer/src/finalization_queue.rs
```diff
@@ -0,0 +1,521 @@
...
+            // to be reinserted in append eth events.
+            self.pending
+                .retain(|diff| diff.da_height < lowest_removed_da_height);
+        }
+        // apply new event to pending queue
+        self.append_da_events(event, da_height).await;
+    }
+
+    /// At begining we will ignore all event until event for new fuel block commit commes
+    /// after that syncronization can start.
```
Voxelot, line 188: We should be able to track the finalized state of the contract with or without waiting and relying on blocks being committed to L1. For example, we may slow down L1 block commitments if the chain is very expensive or congested, but that shouldn't prevent or limit deposits from being processed to L2.

fuel-relayer/src/finalization_queue.rs
```diff
@@ -0,0 +1,521 @@
...
+            }
+        } else {
+            // if there are not pending block for removal just add it.
+            self.bundled_removed_eth_events
+                .push((da_height, vec![event]));
+        }
+    }
+
+    /// propagate new fuel block to pending_blocks
+    pub fn handle_fuel_block(&mut self, block: &Arc<SealedFuelBlock>) {
```
Voxelot, line 115: `&Arc` is an anti-pattern imo. Since Arc impls `Borrow<T>`, if the block comes from an arc we can just borrow it before passing to this function.

```suggestion
    pub fn handle_fuel_block(&mut self, block: &SealedFuelBlock) {
```

fuel-relayer/src/pending_blocks.rs
Voxelot, line 31: Is this the same thing as `last_commited_finalized_fuel_height`, but includes the best height seen even in pending commitments? This name is way too ambiguous for me. I shouldn't have to read this comment each time to know what this is.

```suggestion
    best_pending_chain_height_commit: BlockHeight,
```

fuel-relayer/src/pending_blocks.rs
```diff
@@ -0,0 +1,617 @@
...
+        //get blocks range that start from last pending queue item that is not reverted and goes to current block.
+        for pending in self.pending_block_commits.iter() {
+            if !pending.reverted {
+                from_height = pending.block_height;
+                break;
+            }
+        }
+
+        debug!("Bundle from:{from_height}, to:{to_height}");
+        //.expect("This block should be present as we couldn't create new block");
```
Voxelot, line 157:
```suggestion
```

fuel-relayer/src/pending_blocks.rs
```diff
@@ -0,0 +1,617 @@
...
+        //from_height = from_height + BlockHeight::from(1u64);
```
Voxelot, line 158:
```suggestion
```

fuel-relayer/src/pending_blocks.rs
Voxelot, line 213: LCFFH is pretty meaningless to a someone debugging the chain, since we never use this abbreviation anywhere else. 

```suggestion
                error!("Missing block commitments between last finalized commitment {lcffh} to new height {height}")
```

fuel-relayer/src/pending_blocks.rs
```diff
@@ -0,0 +1,617 @@
...
+                    pending.block_height = height;
+                    pending.block_root = block_root;
+                    break;
+                }
+            }
+        }
+    }
+
+    /// Handle revert of block commit from contract.
+    fn handle_block_commit_revert(
```
Voxelot, line 267: Why don't we just retain all pending_block_commits with da_height < reorg? It seems like the new logs after the reorg would just simply add them back, rather than trying to search and edit their `reverted` and `da_height` vars back and forth.

fuel-relayer/src/relayer.rs
```diff
@@ -215,15 +169,13 @@ impl Relayer {
 
         loop {
             // 1. get best block and its hash sync over it, and push it over
-            self.pending.clear();
-            self.pending.push_back(last_diff.clone());
+            self.queue.clear();
+            //sself.queue.push_back(last_diff.clone());
```
Voxelot, line 173:
```suggestion
```

fuel-relayer/src/finalization_queue.rs
```diff
@@ -0,0 +1,521 @@
...
+        self.finalized_da_height = finalized_da_height;
+        // bump validator set to last finalized block
+        self.validators
+            .bump_set_to_da_height(finalized_da_height, db)
+            .await
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
Voxelot, line 297: There are no tests involving removed logs (reorgs) while there is a lot of complex logic above to deal with them.

fuel-relayer/src/finalization_queue.rs
adlerjohn: That sounds right, but is this comment referring only to the initial sync, or consensus sync (i.e. already at the tip)?

fuel-relayer/src/relayer.rs
```diff
@@ -278,95 +213,110 @@ impl Relayer {
...
-        self.current_validator_set
-            .load_get_validators(self.db.as_mut())
-            .await;
+        self.queue.load_validators(self.db.as_mut()).await;
 
-        loop {
-            // initial sync
-            let (mut blocks_watcher, mut logs_watcher) = match self.initial_sync(&provider).await {
-                Ok(watcher) => watcher,
+        let mut number_of_tries = 10;
```
Voxelot, line 238: 10 is a magic number. Max retries should always be configurable constants.

fuel-relayer/src/relayer.rs
```diff
@@ -233,40 +185,23 @@ impl Relayer {
...
+            self.queue.append_eth_logs(logs).await;
 
             // 3. Start listening to eth events
             let eth_log_filter = Filter::new().address(ValueOrArray::Array(contracts.clone()));
             watchers = Some((
-                self.stop_handle(|| provider.watch_blocks()).await??,
-                self.stop_handle(|| provider.watch(&eth_log_filter))
-                    .await??,
+                handle_interrupt!(self, provider.watch_blocks())??,
+                handle_interrupt!(self, provider.watch(&eth_log_filter))??,
```
Voxelot, line 195: why recreate the watchers in a loop and wait 500ms each time? 

1. The watchers do nothing unless polled as a stream.
2. The default poll interval is 7000ms.

fuel-relayer/src/finalization_queue.rs
rakita: The comment is wrong, it should be updated. As you said Brandon, the state is finalized without relying on the block being committer.

fuel-relayer/src/relayer.rs
rakita: `watch_blocks` and `watch` are sending the request to eth client to subscribe to a particular event here: https://github.com/gakonst/ethers-rs/blob/598c00e55ddbe98267c82de477b355f7f73010b7/ethers-providers/src/provider.rs#L668-L693 
Here is how pub/sub API works: https://geth.ethereum.org/docs/rpc/pubsub
There is a possibility that in-between times I ask for the newest block and start watching events new block gets mined and I can miss it, that is why I get best_block before subscribing and after, and compare if they are the same.
50ms is really arbitrary as to introduce some delay between subscription next request for the best block.

rakita: When we register the subscriber on the eth client, the eth client will start to queue pending logs and they will become available on the next watcher pull request.

```
notifications are stored in an internal buffer and sent from this buffer to the client. If the client is unable to keep up and the number of buffered notifications reaches a limit (currently 10k) the connection is closed. Keep in mind that subscribing to some events can cause a flood of notifications, e.g. listening for all logs/blocks when the node starts to synchronize.
```

fuel-core-interfaces/src/db.rs
rakita: created issue: https://github.com/FuelLabs/fuel-core/issues/384
I would use `u64` as it is easier to serialize as a native type.

fuel-core/src/database.rs
rakita: Makes sense, will do the same with VM related impl, and i would like to move db tables to separate submodules to group them together. https://github.com/FuelLabs/fuel-core/issues/385

fuel-relayer/src/config.rs
rakita: here it is: https://github.com/FuelLabs/fuel-core/pull/345/commits/8a1fbbd182ced087a2d2142f91013f3fe112d1c7

fuel-relayer/src/pending_blocks.rs
rakita: It is better to be sure and override all fields. Will remove `block_height` for it does not make sense.

fuel-relayer/src/pending_blocks.rs
rakita: It is `Highest known chain height`

fuel-relayer/src/relayer.rs
rakita: introduced constant in config

fuel-relayer/src/pending_blocks.rs
```diff
@@ -0,0 +1,617 @@
...
+    ) -> Self {
+        // it is some random key for now
+        let sk = SigningKey::from_bytes(private_key).unwrap();
+        let signer: LocalWallet = sk.into();
+        let signer = signer.with_chain_id(chain_id);
+
+        Self {
+            signer,
+            contract_address,
+            chain_height: BlockHeight::from(10u64), // TODO
```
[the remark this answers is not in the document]
rakita: Properly set chain_height

fuel-relayer/src/relayer.rs
rakita: remove d line in different commit

fuel-relayer/src/pending_blocks.rs
rakita: Code was already made in the previous iteration, so i left it there as It gives us additional information that we can reuse or at least show addition info. A potential use case is if we see that there are reverted commits to delay new fuel commits for some time, just wait for them.

Inclusion of transaction after reorg is not guaranteed, it is most probably outcome, but it depends on multiple factors, reorg block could be empty for example.

fuel-relayer/src/pending_blocks.rs
rakita: Reworded in different commit

fuel-relayer/src/pending_blocks.rs
rakita: Removed in different commit

fuel-relayer/src/relayer.rs
```diff
@@ -154,55 +141,24 @@ impl Relayer {
...
-                // we are sending dummy eth block num bcs we are sure that it is finalized
-                self.pending
-                    .commit_diffs(self.db.as_mut(), best_finalized_block)
-                    .await;
-            }
-            self.pending.pop_back().unwrap()
-        };
+            // we are sending dummy eth block bcs we are sure that it is finalized
+            self.queue.commit_diffs(self.db.as_mut(), end).await;
+        }
```
Voxelot, line 161: I have an idea to simplify this process and make the relayer a bit more durable.

If I understand this next section of code properly, we're trying to backfill all the unfinalized events so that we'll be ready to just start streaming/appending new logs as they're observed. We have these retries in an attempt to ensure:
1. No logs are skipped between initialization and active log listening during regular operation
2. Protect against potential re-orgs.

What if we instead go with the following approach:

Whenever a new log is received, we perform the necessary checks to ensure there were no gaps between the last events we received (e.g. check that block number delta == 1, or that parent block hash matches our last block hash etc), and if we detect any jumps, we fetch the missing events using `get_logs()` until we're caught up to the event we just received.

In the happy path, get_logs is never called, most of these lookups could be done using cached data, and then we simply just append the log.

In the unhappy path, we stop polling for new events until we are resync'd.

Initialization could be drastically simplified to simply grabbing and storing all the finalized events, and then we consolidate concerns like handling pending events and reorgs to a single location. This would also be more reliable in the event some messages were dropped between the eth client subscription and the relayer, as the relayer would be able to automatically detect a divergence in expected state and retrieve the missing history.

Voxelot: Another more radical idea is to ignore any events beyond the finalization window simply. We could just subscribe only to new block events and then call `get_logs()` on (new block - finalization period). Maybe not as efficient as streaming each event via pubsub, but we wouldn't need all the complicated queueing logic to handle reorgs or ever worry about dropped messages.

fuel-relayer/src/relayer.rs
Voxelot: I see, so is the idea that we want to reset the subscription buffer on the eth node to clear any log events that could be preemptively discarded when a reorg happens, instead of relying on the reorg handling logic in the finalization queue later on?

This seems like it could congest the eth node since the watchers don't automatically call `uninstall_filter` for their subscription id when dropped. Would this cause each abandoned watcher queue to pile up to 10k events before being cleaned up by the eth node?

Voxelot: I'm concerned about relying so heavily on the state of the subscription mechanism and trying to finesse the perfect order of events it returns by all this special case handling around starting and stopping watchers etc. 

For example, if the eth node crashed, restarted, updated, or our sticky session needed to be dropped due to infura rescaling, the relayer would just crap out and stop running. See my other comments about coming up with a more durable solution that avoids the complexities of reorg handling etc.

fuel-relayer/src/finalization_queue.rs
rakita: Will add tests

fuel-relayer/src/relayer.rs
rakita: As time goes filter will be droped by eth client.

Here is point where rechecking blocks is needed to be done: https://github.com/FuelLabs/fuel-core/blob/feac06c7f883cbba91df6df1dc05520f878a4ebc/fuel-relayer/src/finalization_queue.rs#L259
Or even before this loop.

For recovery scenearions you can put tasks in the backlog, when they become problem we can focus on them.

fuel-relayer/src/relayer.rs
rakita: Having pending logs (except for block commit) is to have additional information on pending things, it is leftover from the first iteration and it is a working mechanism.
The second proposal can act as an addon to what we already have. As you said we can check finalized blocks and ask eth client for logs on that block this is a more robust way to do things, but I didn't implement it as current mechanism should work just fine, and moving my focus to build missing parts of the client seems like a higher priority.

fuel-relayer/src/relayer.rs
Voxelot: Wrote up my summary of how I think this should be handled here: https://github.com/FuelLabs/fuel-core/issues/398

I'm good with deferring this as follow-up work, given that we identified use-cases for pending events in the UX.

fuel-relayer/src/pending_blocks.rs
Voxelot: ok I'll close my other PR since we may use this data for improving the efficiency of committing blocks to L1.

fuel-relayer/src/relayer.rs
Voxelot: > moving my focus to build missing parts of the client seems like a higher priority.

I'm not saying building the rest of the client is low priority, but we should be making our best effort to ensure loose ends are tied up in one area before opening several other cans of worms. Otherwise, the cognitive overhead of tracking how "done" things actually are will be too hard to manage and we may end up shipping unstable products.

Voxelot:
I think after the tests, we should be good to go on this. Future improvements to pubsub handling can be deferred to https://github.com/FuelLabs/fuel-core/issues/398

Voxelot (approved):

Voxelot (approved):

PR #346, opened 2022-05-20
Improve p2p tests

### LINKED ISSUE #324: Fix p2p-tests stalling issues
Investigate why the reconnect is not always successful.

Rename network name for each test.

leviathanbeak:
Closes #324 

After creating a small local p2p network and testing it out here are some findings and respective changes:

**Reconnect is guaranteed**
As long as Kademlia's random walk is enabled in the config (as is in the tests), the node will try to reconnect to another node if they disconnected previously. It will try to do that every 60 seconds, which is the max interval currently set in random walk.

Random walk tries to discover a random peer + it gets closest peers it already knows and tries to connect to them if not already connected.

_Changes:_ 
- Removed the sanity reconnect test, since it is not needed and is buggy (more on it in the next section)

**Multiple notifications for a single event might've caused issues with logic in the tests**
I have noticed some events like `PeerConnected` or `PeerDisconnected` appear multiple times from `libp2p` and consequently are propagated to our p2p service. Some tests that relied on those events, like now removed `peer_reconnects_after_disconnect()` potentially got their flow messed up because of it and they ended up stalling.

While debugging this I have also found a flaw in our logic when it comes to counting connected peers, since I would just increment a counter on each established connection, since some events were propagated multiple times for the same event, we would end up incrementing the `connected_peers` count multiple times for a single peer!

_Changes:_
- removed connected_peers count in favor of a `HashSet<PeerId>`, this way each original connect/disconnect is counted only once
- only propagate unique  Connect/Disconnect events once to the p2p service by utilizing newly added HashSet<PeerId> where we track connected peers

**Peer Info intervals can be customized**
Up until now we depended on default values for how often would a peer get identified and/or pinged.
I have added 2 more config field options in order to customize these intervals, giving us more flexibility.

Using smaller intervals for our tests should in theory drive tests faster to completion.

**Other changes**
Our tests were ran concurrently and all of them had the same network name - which could cause some interference.
While we already moved to executing tests synchronously - I have changed the name of each test's network name to match the name of the test itself.

**Does this mean there won't be any stalling tests like EVER?**
Unfortunately no, we cannot guarantee that at this point. What we've done here is removed most of the possible causes.
It's hard to detect from the CI did the connection break between the peers or there was some bug in the test flow.
Hence we should run some cron jobs with tracing enabled in order to discover possible issues.

### DIFF
--- a/fuel-p2p/src/behavior.rs
+++ b/fuel-p2p/src/behavior.rs
@@ -113,7 +113,7 @@ impl<Codec: NetworkCodec> FuelBehaviour<Codec> {
             discovery_config
         };
 
-        let peer_info = PeerInfoBehaviour::new(local_public_key);
+        let peer_info = PeerInfoBehaviour::new(local_public_key, p2p_config);
 
         let req_res_protocol =
             std::iter::once((codec.get_req_res_protocol(), ProtocolSupport::Full));
--- a/fuel-p2p/src/config.rs
+++ b/fuel-p2p/src/config.rs
@@ -36,10 +36,10 @@ pub struct P2PConfig {
     // `PeerInfo` fields
     /// The interval at which identification requests are sent to
     /// the remote on established connections after the first request
-    pub identify_interval: Duration,
+    pub identify_interval: Option<Duration>,
     /// The duration between the last successful outbound or inbound ping
     /// and the next outbound ping
-    pub info_interval: Duration,
+    pub info_interval: Option<Duration>,
 
     // `Gossipsub` related fields
     pub topics: Vec<String>,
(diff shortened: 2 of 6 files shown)


leviathanbeak:
the irony would be if the CI failed now 😅

Voxelot:
The tests look much faster now, awesome! Is there any tracing you'd want to add before we turn on the cron job?

leviathanbeak:
> Is there any tracing you'd want to add before we turn on the cron job?

Yes, I will try to add some more tracing features in order for us to get a nicer overview of what happens

Voxelot (approved):
approving for now, feel free to do the instrumentation for a cron job in a separate pr

PR #347, opened 2022-05-21
Allow dry-run to override utxo_validation setting

Voxelot:
While ethereum has `.call` for querying read-only methods on contracts, our dry-run has the ability to fully simulate the end to end execution of a transaction including the txpool rules. While this is great for things like estimating gas costs and predicting result of a transaction execution which modifies state, it's inconvenient for simple read-only use-cases if simulated txs require real coin inputs and valid signatures. 

To improve the read-only `call`-style experience, this PR adds a new optional `utxo_validation` parameter to the dry run API. This allows the SDK to bypass coin and signature validation normally applied to a transaction. This was added in a non-breaking way, allowing for this to be a patch release to fuel-core.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,6 +153,7 @@ jobs:
 
   publish-docker-image:
     needs:
+      - cancel-previous-runs
       - lint-toml-files
       - cargo-verifications
       - publish-crates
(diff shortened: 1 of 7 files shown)


luizstacio (approved):
LGTM

AlicanC (approved):
LGTM

PR #350, opened 2022-05-21
Update Contracts API Endpoints

ControlCplusControlV:
This PR implements a couple things
     - The Salt object on the Contract Object, as a new field on an existing structure
     - A new endpoint for querying the balance of a Contract
     - Added this new endpoint for fuel-client
     - Updated the appropriate snapshot tests, and schema for the above

I wrote 3 tests for this, as well as reshuffled around the `TestSetupBuilder` into a new helpers file, as it was needed in my contracts tests as well, but was only avaliable in the `/tx` tests, so it's since been moved to a new helpers.rs

Tests cover that the Salt contract object exists, and a test to cover that when a contract has no balance 0 is returned, and a test to ensure it returns the correct balance when the contract does have a balance.

To fully close the Contracts API issue a cursor paginated query still needs to be implemented, and contract connection, but as both of the components usually build on an underlying request, I thought it would be better to get contract balance in, before then adding features to extend its functionality


fuel-core/src/schema/contract.rs
```diff
@@ -49,3 +60,34 @@ impl ContractQuery {
...
+
+        let db = ctx.data_unchecked::<Database>().clone();
+
+        let asset_id: fuel_types::AssetId = asset.into();
+
+        let result = fuel_vm::storage::InterpreterStorage::merkle_contract_asset_id_balance(
+            &db,
+            &contract_id,
+            &asset_id,
+        );
```
AlicanC, line 85: Nice. Even though I was very comfortable with GQL, I found it hard to find ways to reach the data I actually needed.

So this is how we get contract balances, huh? Is there a way to use this to build the plural API `Query.contractBalances`? Something to get all balances of a contract?

AlicanC (changes requested):
Other than the `U64 -> ContractBalance` comment, nothing that really needs to be addressed.

If we don't want to expand the scope of this PR, then a follow-up issue for adding these would be great:
- Query.contractBalances
- Contract.balance
- Contract.balances

fuel-client/assets/schema.sdl
```diff
@@ -232,6 +233,7 @@ type Query {
 	coins(filter: CoinFilterInput!, first: Int, after: String, last: Int, before: String): CoinConnection!
 	coinsToSpend(owner: Address!, spendQuery: [SpendQueryElementInput!]!, maxInputs: Int, excludedIds: [UtxoId!]): [Coin!]!
 	contract(id: ContractId!): Contract
+	contractBalance(contract: ContractId!, asset: AssetId!): U64!
```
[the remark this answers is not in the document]
ControlCplusControlV: Ok, I can make a new `ContractBalance` to return rather than just a U64, swapping `owner` for `contract_id`

fuel-client/assets/schema.sdl
```diff
@@ -131,6 +131,7 @@ enum CoinStatus {
 type Contract {
 	id: ContractId!
 	bytecode: HexString!
+	salt: Salt!
```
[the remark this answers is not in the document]
ControlCplusControlV: I originally tried doing it this way, but I found args are only valid on queries, not plain Graphql types. Would there be a way to do it like this? When I talked with @Voxelot he suggested making a single specific balance query and breaking away balances from the default contract object (in addition to a balances query which I am planning on covering in the next PR)

fuel-core/src/schema/contract.rs
ControlCplusControlV: I had was having trouble as well accessing the data until I found the [fuel-vm docs](https://docs.rs/fuel-core/0.6.3/fuel_core/database/struct.Database.html#method.merkle_contract_asset_id_balance) which are very helpful for fetching a data from the db.

I don't know of a specific way to fetch all the balances fro a contract, but my thinking was using just accepting a vec of assetId's for `Query.contractBalances` and looping through each of them when fetching balances.

fuel-client/assets/schema.sdl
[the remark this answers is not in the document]
Voxelot: I'd have to take a closer look to see if async-graphql / cynic support this type of query structure @AlicanC. I think deferring balances to a separate PR is a good idea.

fuel-core/src/schema/contract.rs
Voxelot: All tables are iterable, performing these kinds of queries depends on the key structure, as all k/v pairs are lexicographically sorted by their bytes. You can see that balance keys are inserted in [database/balances.rs](https://github.com/FuelLabs/fuel-core/blob/master/fuel-core/src/database/balances.rs#L24) as a (contract_id, asset_id) concatenated tuple. So for enumerating all the balances for a contract, you'd provide just the `contract_id` prefix to the query and the database could iterate over all the assets for that prefix. Operations such as sorting based on balance value would be more suited for fuel-indexer though.

fuel-tests/tests/contract.rs
```diff
@@ -1,29 +1,201 @@
...
+                input_index: 1,
+                balance_root: Default::default(),
+                state_root: Default::default(),
+            })
+            .finalize()
+        })
+        .collect_vec();
+
+    // setup genesis block with coins that transactions can spend
+    test_builder.config_coin_inputs_from_transactions(&transactions);
```
Voxelot, line 56: It doesn't look like you need these transactions for this test

```suggestion
```

fuel-tests/tests/contract.rs
```diff
@@ -1,29 +1,201 @@
...
+
```
Voxelot, line 120: Likewise, it seems like these txs don't impact the subject under test.

```suggestion
```

fuel-tests/tests/contract.rs
```diff
@@ -1,29 +1,201 @@
...
+                input_index: 1,
+                balance_root: Default::default(),
+                state_root: Default::default(),
+            })
+            .finalize()
+        })
+        .collect_vec();
+
+    // setup genesis block with coins that transactions can spend
+    test_builder.config_coin_inputs_from_transactions(&transactions);
```
Voxelot, line 119: Seems like these txs / coins are unused.

```suggestion
```

fuel-tests/tests/helpers.rs
```diff
@@ -0,0 +1,153 @@
...
+impl TestSetupBuilder {
+    pub fn new(seed: u64) -> TestSetupBuilder {
+        Self {
+            rng: StdRng::seed_from_u64(seed),
+            ..Default::default()
+        }
+    }
+
+    /// setup a contract and add to genesis configuration
+    pub fn setup_contract(&mut self, code: Vec<u8>) -> (Salt, ContractId) {
```
Voxelot, line 33:
```suggestion
    pub fn setup_contract(&mut self, code: Vec<u8>, balances: Option<HashMap<AssetId, u64>>) -> (Salt, ContractId) {
```

fuel-tests/tests/helpers.rs
```diff
@@ -0,0 +1,153 @@
...
+        let salt: Salt = self.rng.gen();
+        let contract_id = contract.id(&salt.clone(), &root, &Contract::default_state_root());
+
+        self.contracts.insert(
+            contract_id,
+            ContractConfig {
+                code,
+                salt,
+                state: None,
+                balances: None,
```
Voxelot, line 45:
```suggestion
                balances,
```

fuel-tests/tests/contract.rs
```diff
@@ -1,29 +1,201 @@
...
+    // Check that salt is 0x Hex prefixed
+    let salt = contract.unwrap().salt;
+    assert_eq!("0x", &salt.to_string()[..2]);
+}
+
+#[tokio::test]
+async fn test_contract_balance_default() {
+    let mut rng = StdRng::seed_from_u64(2322);
+    let mut test_builder = TestSetupBuilder::new(2322);
+    let (_, contract_id) = test_builder.setup_contract(vec![]);
```
Voxelot, line 75: See my other posts about updating the test builder to accept initial balances for contracts.

fuel-client/assets/schema.sdl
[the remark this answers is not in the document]
AlicanC: > I originally tried doing it this way, but I found args are only valid on queries, not plain Graphql types.

In GraphQL speak, `contractBalance` is a field of the type `Query` and it has arguments. `Contract` is yet another type like `Query` and it could as well have a similar field.

I thought there was an example of this in the repo but couldn't find any. I guess we'll leave it to whoever takes the follow-up issue :D

fuel-tests/tests/helpers.rs
ControlCplusControlV: Changed to now accept an Option of a Vec of Tuples

fuel-tests/tests/contract.rs
ControlCplusControlV: I've updated this and the other `setup_contract` references to now initialize with a balance

fuel-tests/tests/contract.rs
```diff
@@ -1,29 +1,64 @@
...
-    assert!(contract.is_some());
+
+    assert_eq!(balance, 0); // Ensures default value is working properly
+}
+
+#[tokio::test]
+async fn test_contract_balance() {
+    let mut test_builder = TestSetupBuilder::new(2322);
+    let (_, contract_id) =
+        test_builder.setup_contract(vec![], Some(vec![(AssetId::new([1u8; 32]), 500)]));
```
Voxelot, line 48: move magic number into constant variable at start of the test.

ControlCplusControlV: Moved it to a const at the beginning of the file and then pointed all the test_build initializations to it

Voxelot: oh i meant the balance var `500`

ControlCplusControlV: I did slightly more than that, instead of making it a const I just merged it with the previous test (which did the same thing except with a balance of 0) to instead have on test which tries with various different potential balances

fuel-client/assets/schema.sdl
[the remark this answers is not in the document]
ControlCplusControlV: @AlicanC I added in `ContractBalance`, did I implement it how you were expecting?

Voxelot (approved):

Voxelot (approved):

fuel-client/assets/schema.sdl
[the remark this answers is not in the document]
AlicanC: LGTM

AlicanC (commented):
👍

AlicanC (approved):

PR #354, opened 2022-05-24
readme update

chappie1998:
"cargo xtask build" command only runs "cargo build" command and "cargo run" build only for development. It only creates a debug directory. It doesn't create a release directory. To create a release directory or run it on production and optimized "cargo xtask build" should run "cargo build --release" command.


Voxelot:
Ah we should probably skip docker image publishing from fork branches. lgtm

PR #355, opened 2022-05-25
Support split of Input::Coin in to CoinSignature & CoinPredicate

Voxelot:
Adds support for the new Input type `CoinPredicate`. This will only break Rust dependents, graphql and ts consumers won't be affected.

This also adds a feature flag called `predicates` to the fuel-core service config which allows the node to reject tx's with predicates until they are stabilized.

### DIFF
--- a/fuel-client/Cargo.toml
+++ b/fuel-client/Cargo.toml
@@ -21,9 +21,9 @@ clap = { version = "3.1", features = ["derive"] }
 cynic = { version = "1.0", features = ["surf"] }
 derive_more = { version = "0.99" }
 fuel-storage = "0.1"
-fuel-tx = { version = "0.10", features = ["serde"] }
+fuel-tx = { version = "0.11", features = ["serde"] }
 fuel-types = { version = "0.5", features = ["serde"] }
-fuel-vm = { version = "0.9", features = ["serde"] }
+fuel-vm = { version = "0.10", features = ["serde"] }
 futures = "0.3"
 hex = "0.4"
 itertools = "0.10"
--- a/fuel-core-interfaces/src/db.rs
+++ b/fuel-core-interfaces/src/db.rs
@@ -4,6 +4,7 @@ use fuel_vm::prelude::InterpreterError;
 use thiserror::Error;
 
 #[derive(Error, Debug)]
+#[non_exhaustive]
 pub enum Error {
     #[error("error performing binary serialization")]
     Codec,
@@ -180,15 +181,13 @@ pub mod helpers {
                 receipts_root: Default::default(),
                 script,
                 script_data: vec![],
-                inputs: vec![Input::Coin {
+                inputs: vec![Input::CoinSigned {
                     utxo_id: UtxoId::new(*TX_ID_DB1, 0),
                     owner: Address::default(),
                     amount: 100,
                     asset_id: Default::default(),
                     witness_index: 0,
                     maturity: 0,
-                    predicate: vec![],
-                    predicate_data: vec![],
                 }],
                 outputs: vec![
                     Output::Coin {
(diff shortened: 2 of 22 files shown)


adlerjohn (approved):
This seems fine, but would like @vlopes11 to review.

fuel-client/Cargo.toml
```diff
@@ -21,9 +21,9 @@ clap = { version = "3.1", features = ["derive"] }
 cynic = { version = "1.0", features = ["surf"] }
 derive_more = { version = "0.99" }
 fuel-storage = "0.1"
-fuel-tx = { version = "0.10", features = ["serde"] }
+fuel-tx = { version = "0.11", features = ["serde"] }
```
vlopes11, line 24: Consider upgrading to 0.12 instead

fuel-core-interfaces/src/db.rs
```diff
@@ -180,15 +181,13 @@ pub mod helpers {
                 receipts_root: Default::default(),
                 script,
                 script_data: vec![],
-                inputs: vec![Input::Coin {
+                inputs: vec![Input::CoinSigned {
```
vlopes11, line 184: We could consider including a predicate too, for test purposes

fuel-core/src/args.rs
```diff
@@ -53,9 +53,15 @@ pub struct Opt {
...
     pub min_gas_price: u64,
+
     /// The minimum allowed byte price
     #[clap(long = "min-byte-price", default_value = "0")]
     pub min_byte_price: u64,
+
+    /// Enable predicate execution on transaction inputs.
+    /// Will reject any transactions with predicates if set to false.
+    #[clap(long = "predicates")]
+    pub predicates: bool,
```
vlopes11, line 64: contract creation currently rejects tx create with invalid predicates. should we disable that default check?

https://github.com/FuelLabs/fuel-vm/blob/ffaf70c6d0f7974de4bd316c1742f3698ec8c219/src/interpreter/executors/main.rs#L132-L137

fuel-core/src/executor.rs
```diff
@@ -268,6 +270,21 @@ impl Executor {
         Ok(())
     }
 
+    /// Verify all the predicates of a tx.
+    pub fn verify_tx_predicates(&self, tx: &Transaction) -> Result<(), Error> {
+        // fail if tx contains any predicates when predicates are disabled
+        if !self.config.predicates {
```
vlopes11, line 276: If the predicates verification is disabled in the config, shouldn't we ignore them?

fuel-tests/tests/tx/predicates.rs
```diff
@@ -0,0 +1,75 @@
...
+use rand::{rngs::StdRng, Rng, SeedableRng};
+
+#[tokio::test]
+async fn transaction_with_predicates_is_rejected_when_feature_disabled() {
+    let mut rng = StdRng::seed_from_u64(2322);
+
+    // setup tx with a predicate input
+    let asset_id = rng.gen();
+    let predicate_tx = TransactionBuilder::script(Default::default(), Default::default())
+        .add_input(Input::coin_predicate(
```
vlopes11, line 16: This should break the validation for the next bump

https://github.com/FuelLabs/fuel-tx/blob/c9704af3ba826beb559623912fbd735a2303b32e/src/transaction/types/input.rs#L255-L261

We should set the owner as the code root of the validator

vlopes11 (approved):
LG overall; couple of questions and optional nits

fuel-client/Cargo.toml
Voxelot: I think I'll do that in a separate PR, just to reduce make the changes more isolated.

fuel-core-interfaces/src/db.rs
Voxelot: yeah i'll do that in the future, hence why predicates are under a feature flag for now as unstable

fuel-core/src/args.rs
Voxelot: yes probably, triggering those checks should be outside the VM (e.g. from the txpool)

Voxelot: not a blocker atm but could you file a tech debt issue?

fuel-core/src/executor.rs
Voxelot: I wanted to play it safe and just outright reject txs that include predicates, that way someone can't spend invalid predicates in our stable environment

fuel-tests/tests/tx/predicates.rs
Voxelot: thanks for the heads up, will deal with it in the next pr

PR #370, opened 2022-05-27
async-graphql upgrade to v4

### LINKED ISSUE #310: async-graphql upgrade
We currently pin `async-grapqhql` dependency to version `=2.9`, since version `2.10` includes some breaking changes. This issue tracks the efforts on upgrading the dependency. Currently version `4` of the library is in active development, and seems to fix most of the issues, but it hasn't been released yet. As the version `3` seems to be somewhat broken, upgrading directly to `v4` when it's released seems like a good idea. Some of the required migration steps are described here.

## Async and errors

Version `2.10` doesn't work with the current error types, as we don't bound the `dyn std::error::Error` types strictly enough to be used in async contexts. To fix this, some  instances of  `Box<dyn std::error::Error>` and it's alias `Box<dyn StdError>` must be replaced with `Box<Error + Send + Sync>`. The compiler errors for this are somewhat hard to read, but searching for boxed std error types finds them all quickly.

## Type inference inside `query` calls

EIther version `3` or `2.10` (not sure which), requires `Ok(...)` return values within `query` calls to mark their return type as type inference seems to fail. The correct type can be annotated like this: `Ok::<_, KvStoreError>(...)`, or in some cases with `Ok::<_, Infallible>(...)`

## Separate types required for input and output

Until version `3` of `async-graphql`, separate types are required for input and output values, as `SimpleObject` and `InputObject` cannot be both derived for the same type. This requires lots of duplicated code. Version `3` fixes this limitation. See [this library docs](https://async-graphql.github.io/async-graphql/en/define_simple_object.html#used-for-both-input-and-output) for this.

```rust
#[derive(SimpleObject, InputObject)]
#[graphql(input_name = "MyObjInput")]
struct MyObj {
    field: i32
}
```

## Duplicate GraphQL type names

Tracking issue https://github.com/async-graphql/async-graphql/pull/876

Version `3` of `async-graphql` introduces a regression that doesn't allow multiple types like `Connection<Cursor, Transaction, _, _>` where the `Cursor` type varies but `Transaction` does not. One such instance in this codbase is [`Connection<SortedTxCursor, Transaction, EmptyFields, EmptyFields>`](https://github.com/FuelLabs/fuel-core/blob/735ea7ffe3218b5c1ead3f56dbb9b4a81066eac1/fuel-core/src/schema/tx.rs#L64) and [`Connection<HexString, Transaction, EmptyFields, EmptyFields>`](https://github.com/FuelLabs/fuel-core/blob/735ea7ffe3218b5c1ead3f56dbb9b4a81066eac1/fuel-core/src/schema/tx.rs#L174). Note that this causes a runtime panic, instead of a compilation error. This is caused by the [`type_name` implementation of `Connection`](https://docs.rs/async-graphql/3.0.38/src/async_graphql/types/connection/connection_type.rs.html#163) (and Edge as well), which ignore the type of the cursor (`C`) when generating graphql names. This could be fixed with something like this in the above-mentioned file.

```rust
/// Name of a Rust type, cleaned of punctuation so it's usable with GraphQL
fn clean_type_name<T>() -> String {
    std::any::type_name::<T>()
        .chars()
        .map(|c| if c.is_ascii_alphanumeric() { c } else { '_' })
        .fold(String::new(), |mut s, c| {
            if (!s.ends_with('_')) || c != '_' {
                s.push(c);
            }
            s
        })
}

impl  ... for Connection ... {
  fn type_name() -> Cow<'static, str> {
    Cow::Owned(format!("{}{}Connection", clean_type_name::<C>(), T::type_name()))
  }
}
```

Unfortunately this also renames the cases where the cursor is not duplicated, breaking backwards compatibility with the older graphql API. After this most of the queries in the client module must be rewritten as well.

It could also be worked around with using a separate wrapper type for `Transaction`, but after wasting hours on attempting that workaround, I've given up on it.

## Version `4` breaking changes

The version `4` includes more breaking changes than mentioned here, including soem changes to type signatures. I have not gone through all these yet, and probably more will be coming before the v4 release.

leviathanbeak:
Closes #310 

I updated `async-graphql` to v4

In this latest version they removed `append()` method from `Connection` so I had to extend the edges with
`conn.edges.extend()` directly

Our other problem was that our Errors weren't constrained to Send + Sync, so I added those constraints and now it compiles

### DIFF
--- a/fuel-client/assets/schema.sdl
+++ b/fuel-client/assets/schema.sdl
@@ -1,10 +1,13 @@
 scalar Address
+
 scalar AssetId
+
 type Balance {
 	owner: Address!
 	amount: U64!
 	assetId: AssetId!
 }
+
 type BalanceConnection {
 	"""
 	Information to aid in pagination.
@@ -13,27 +16,30 @@ type BalanceConnection {
 	"""
 	A list of edges.
 	"""
-	edges: [BalanceEdge]
+	edges: [BalanceEdge!]!
 }
+
 """
 An edge in a connection.
 """
 type BalanceEdge {
 	"""
-	The item at the end of the edge
-	"""
-	node: Balance!
-	"""
 	A cursor for use in pagination
 	"""
 	cursor: String!
+	"""
+	"The item at the end of the edge
+	"""
+	node: Balance!
 }
+
 input BalanceFilterInput {
 	"""
 	Filter coins based on the `owner` field
 	"""
 	owner: Address!
 }
+
 type Block {
 	id: BlockId!
 	height: U64!
@@ -41,6 +47,7 @@ type Block {
 	time: DateTime!
 	producer: Address!
 }
+
 type BlockConnection {
 	"""
 	Information to aid in pagination.
@@ -49,38 +56,46 @@ type BlockConnection {
 	"""
 	A list of edges.
 	"""
-	edges: [BlockEdge]
+	edges: [BlockEdge!]!
 }
+
 """
 An edge in a connection.
 """
 type BlockEdge {
 	"""
-	The item at the end of the edge
-	"""
-	node: Block!
-	"""
 	A cursor for use in pagination
 	"""
 	cursor: String!
+	"""
+	"The item at the end of the edge
+	"""
+	node: Block!
 }
+
 scalar BlockId
+
+
 input Breakpoint {
 	contract: ContractId!
 	pc: U64!
 }
+
 scalar Bytes32
+
 type ChainInfo {
 	name: String!
 	latestBlock: Block!
 	baseChainHeight: U64!
 	peerCount: Int!
 }
+
 type ChangeOutput {
 	to: Address!
 	amount: U64!
 	assetId: AssetId!
 }
+
 type Coin {
 	utxoId: UtxoId!
 	owner: Address!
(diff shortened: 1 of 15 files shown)


leviathanbeak:
While `Connection<_, KvStoreError>` could've worked I opted to be more explicit and define the exact type :)

leviathanbeak:
In order to fix `Duplicate GraphQL type names` issue that @Dentosal mentioned, I have added `TransactionConnection` to be ignored in name conflicts by using this latest fix from `async-graphql` => https://github.com/async-graphql/async-graphql/pull/876#issuecomment-1135859052

fuel-client/assets/schema.sdl
```diff
@@ -98,21 +114,23 @@ type CoinConnection {
...
-	"""
-	The item at the end of the edge
-	"""
-	node: Coin!
 	"""
 	A cursor for use in pagination
 	"""
 	cursor: String!
+	"""
+	"The item at the end of the edge
```
Voxelot, line 129: This kind of looks like a bug in async-grapqhl, why do all these connection edges start with quotes

```suggestion
	"The item at the end of the edge
```

Voxelot (approved):

PR #372, opened 2022-05-29
add node info and version

### LINKED ISSUE #361: Add node info to graphql schema
What:

Add node info schema to the fuel-core api, this should include things like:

- The version of the node (from `env!("CARGO_PKG_VERSION")`)
- The settings of the node:
  - utxo-validation enabled: true/false
  - predicates enabled: true/false
  - vm-backtraces enabled: true/false
  - min-gas-price (from txpool::Config): u64
  - min-byte-price (from txpool::Config): u64
  - max-tx (from txpool::Config): u64
  - max_depth max_depth: u64

Why:

This will make it easier to troubleshoot issues users may have in different environments by allowing them to run a simple curl command to gather most of the info we'd need.

leviathanbeak:
Closes #361


Voxelot:
Since graphql allows clients to only fetch the fields they want, I don't see the utility in splitting this up into two separate queries?

ControlCplusControlV (approved):
LGTM! Do you we want integration tests for this endpoint?

leviathanbeak:
> Since graphql allows clients to only fetch the fields they want, I don't see the utility in splitting this up into two separate queries?

I kind of went with what was mentioned in the issue, `settings/info` & `version` but yeah I can add `version` as part of the `node info`

Voxelot:
Ah my bad, I intended them to all be under one ChainInfo endpoint, but the way I structured that list of details was based on how fuel-core internally groups these values (which isn't actually that relevant to the API).

fuel-core/src/schema/node_info.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+}
+
+#[Object]
+impl NodeQuery {
+    async fn node_version(&self, _ctx: &Context<'_>) -> async_graphql::Result<String> {
+        const VERSION: &str = env!("CARGO_PKG_VERSION");
+
+        Ok(VERSION.to_owned())
+    }
```
AlicanC, line 55: We already have this as `Query.version`: https://github.com/FuelLabs/fuel-core/blob/master/fuel-core/src/schema/tx.rs#L31-L35

I think we should remove that and keep the new `node_version`.

leviathanbeak: yeah I actually copy/pasted from it, forgot to remove it

leviathanbeak:
@Voxelot @AlicanC @ControlCplusControlV updated the PR, please (re)review :)

fuel-client/src/client/schema/node_info.rs
```diff
@@ -0,0 +1,32 @@
...
+pub struct QueryNodeInfo {
+    pub node_info: NodeInfo,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn node_info_query_gql_output() {
```
ControlCplusControlV, line 27: Would be nice to have end-to-end tests for this new endpoint in `fuel-tests`, so that it can be tested against a variety of configured nodes

ControlCplusControlV (commented):
LGTM, can approve if needed but would like end to end tests for this new endpoint using the fuel-client endpoint to call in and validate in a specific env to match with the other endpoints which have tests there

fuel-client/src/client/schema/snapshots/fuel_gql_client__client__schema__node_info__tests__node_settings_query_gql_output.snap
```diff
@@ -0,0 +1,8 @@
+---
```
Voxelot, line 1: this snapshot doesn't have a corresponding test

Voxelot (approved):

fuel-client/src/client/schema/snapshots/fuel_gql_client__client__schema__node_info__tests__node_settings_query_gql_output.snap
leviathanbeak: good catch, will remove it

fuel-client/src/client/schema/node_info.rs
leviathanbeak: good point, I will add one test there

Voxelot (approved):

PR #373, opened 2022-05-29
tx consensus parameters

### LINKED ISSUE #362: Add transaction consensus parameters to ChainInfo graphql schema
What:

Now that the parameters used for validating transactions are [configurable via chainspec](https://github.com/FuelLabs/fuel-core/blob/29366c1932321932b87792088984c418f014b0ea/fuel-core/src/chain_config.rs#L25), add a new field in the [`ChainInfo`](https://github.com/FuelLabs/fuel-core/blob/29366c1932321932b87792088984c418f014b0ea/fuel-core/src/schema/chain.rs#L13) schema to return these to users.

Why:

This will make it easier for the SDK or other fuel-core consumers to know what settings to use for things like MAX_GAS_LIMIT when building transactions.

leviathanbeak:
closes #362 

Created `ConsensusParameters` struct with all the necessary fields, and made it part of the `ChainInfo` query - that is, gave it its own queryable method `consensus_parameters`.

### DIFF
--- a/fuel-client/assets/schema.sdl
+++ b/fuel-client/assets/schema.sdl
@@ -88,6 +88,7 @@ type ChainInfo {
 	latestBlock: Block!
 	baseChainHeight: U64!
 	peerCount: Int!
+	transactionParameters: ConsensusParameters!
 }
 
 type ChangeOutput {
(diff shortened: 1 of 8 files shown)


adlerjohn (commented):
looks fairly straightforward

Voxelot:
> Would unit tests be necessary as this looks like it's adding a new field on an existing endpoint? Otherwise it looks fairly straightforward and can approve if tests are not needed

Agree, even though this API is very simple and basic, we should have at least one happy path test per public interface.

leviathanbeak:
> need to remove unused snapshot, otherwise looks good

ah kept that by accident while rebasing since I started off from where I left the previous PR

leviathanbeak:
@Voxelot @ControlCplusControlV adding a e2e test was actually good idea, since there wasn't `chain_info` method on the client at all. I have added it and added a test.

ControlCplusControlV (approved):

AlicanC (commented):
LGTM

Voxelot (approved):

Issue #71, opened 2021-11-19
Contracts API


Voxelot:
Add APIs for retrieving information about deployed contracts.

- [ ] Components
  - [x] add to rust api in fuel-client
    - [x] snapshot test
    - [x] integration tests
  - [x] add to GQL schema in fuel-core
- [x] API
  - [x] get contract by id
  - [ ] #395
- [x] Data
  - [x] id
  - [x] salt
  - [x] code
  - [x] balances

PR #376, opened 2022-05-31
Add Contract Balances API

ControlCplusControlV:
Should fix and mean everything in #71  has now been addressed, closing up on the contracts API.

Adds an endpoint for contract balances, using `iter_all` to fetch the balance for each asset, of all possible assets the contract has. This is my first time writing a query which requires a cursor and pages, so it's heavily based off of the existing balances endpoint. 

I also haven't been to close to the db yet, so safety of the db query is something I wasn't sure on.

### DIFF
--- a/fuel-client/src/client/schema/contract.rs
+++ b/fuel-client/src/client/schema/contract.rs
@@ -88,7 +88,7 @@ pub struct ContractBalanceEdge {
 #[derive(cynic::QueryFragment, Debug)]
 #[cynic(schema_path = "./assets/schema.sdl")]
 pub struct ContractBalanceConnection {
-    pub edges: Option<Vec<Option<ContractBalanceEdge>>>,
+    pub edges: Vec<ContractBalanceEdge>,
     pub page_info: PageInfo,
 }
 
@@ -109,12 +109,7 @@ impl From<ContractBalanceConnection> for PaginatedResult<ContractBalance, String
             has_next_page: conn.page_info.has_next_page,
             has_previous_page: conn.page_info.has_previous_page,
             cursor: conn.page_info.end_cursor,
-            results: conn
-                .edges
-                .unwrap_or_default()
-                .into_iter()
-                .filter_map(|e| e.map(|e| e.node))
-                .collect(),
+            results: conn.edges.into_iter().map(|e| e.node).collect(),
         }
     }
 }
--- a/fuel-client/src/client/schema/contract.rs
+++ b/fuel-client/src/client/schema/contract.rs
@@ -1,4 +1,5 @@
-use crate::client::schema::{schema, AssetId, ContractId, HexString, Salt, U64};
+use crate::client::schema::{schema, AssetId, ContractId, PageInfo, HexString, Salt, U64};
+use crate::client::{PageDirection, PaginatedResult, PaginationRequest};
 
 #[derive(cynic::FragmentArguments, Debug)]
 pub struct ContractByIdArgs {
@@ -53,6 +54,90 @@ pub struct Contract {
 #[cynic(schema_path = "./assets/schema.sdl", graphql_type = "Contract")]
 pub struct ContractIdFragment {
     pub id: ContractId,
+}
+
+#[derive(cynic::InputObject, Clone, Debug)]
+#[cynic(schema_path = "./assets/schema.sdl")]
+pub struct ContractBalanceFilterInput {
+    /// Filter coins based on the `owner` field
+    pub contract: ContractId,
+}
+
+#[derive(cynic::FragmentArguments, Debug)]
+pub struct ContractBalancesConnectionArgs {
+    /// Filter coins based on a filter
+    filter: ContractBalanceFilterInput,
+    /// Skip until coin id (forward pagination)
+    pub after: Option<String>,
+    /// Skip until coin id (backward pagination)
+    pub before: Option<String>,
+    /// Retrieve the first n coins in order (forward pagination)
+    pub first: Option<i32>,
+    /// Retrieve the last n coins in order (backward pagination).
+    /// Can't be used at the same time as `first`.
+    pub last: Option<i32>,
+}
+
+#[derive(cynic::QueryFragment, Debug)]
+#[cynic(schema_path = "./assets/schema.sdl")]
+pub struct ContractBalanceEdge {
+    pub cursor: String,
+    pub node: ContractBalance,
+}
+
+#[derive(cynic::QueryFragment, Debug)]
+#[cynic(schema_path = "./assets/schema.sdl")]
+pub struct ContractBalanceConnection {
+    pub edges: Option<Vec<Option<ContractBalanceEdge>>>,
+    pub page_info: PageInfo,
+}
+
+#[derive(cynic::QueryFragment, Debug)]
+#[cynic(
+    schema_path = "./assets/schema.sdl",
+    graphql_type = "Query",
+    argument_struct = "ContractBalancesConnectionArgs"
+)]
+pub struct ContractBalancesQuery {
+    #[arguments(filter = &args.filter, after = &args.after, before = &args.before, first = &args.first, last = &args.last)]
+    pub contract_balances: ContractBalanceConnection,
+}
+
+impl From<ContractBalanceConnection> for PaginatedResult<ContractBalance, String> {
+    fn from(conn: ContractBalanceConnection) -> Self {
+        PaginatedResult {
+            has_next_page: conn.page_info.has_next_page,
+            has_previous_page: conn.page_info.has_previous_page,
+            cursor: conn.page_info.end_cursor,
+            results: conn
+                .edges
+                .unwrap_or_default()
+                .into_iter()
+                .filter_map(|e| e.map(|e| e.node))
+                .collect(),
+        }
+    }
+}
+
+impl From<(ContractId, PaginationRequest<String>)> for ContractBalancesConnectionArgs {
+    fn from(r: (ContractId, PaginationRequest<String>)) -> Self {
+        match r.1.direction {
+            PageDirection::Forward => ContractBalancesConnectionArgs {
+                filter: ContractBalanceFilterInput { contract: r.0 },
+                after: r.1.cursor,
+                before: None,
+                first: Some(r.1.results as i32),
+                last: None,
+            },
+            PageDirection::Backward => ContractBalancesConnectionArgs {
+                filter: ContractBalanceFilterInput { contract: r.0 },
+                after: None,
+                before: r.1.cursor,
+                first: None,
+                last: Some(r.1.results as i32),
+            },
+        }
+    }
 }
 
 #[cfg(test)]
--- a/fuel-client/src/client/schema/contract.rs
+++ b/fuel-client/src/client/schema/contract.rs
@@ -67,7 +67,7 @@ pub struct ContractBalanceFilterInput {
 pub struct ContractBalancesConnectionArgs {
     /// Filter balances based on a filter
     filter: ContractBalanceFilterInput,
-    /// Skip until coin id (forward pagination)
+    /// Skip until asset id (forward pagination)
     pub after: Option<String>,
     /// Skip until coin id (backward pagination)
     pub before: Option<String>,
--- a/fuel-client/src/client/schema/contract.rs
+++ b/fuel-client/src/client/schema/contract.rs
@@ -71,7 +71,7 @@ pub struct ContractBalancesConnectionArgs {
     pub after: Option<String>,
     /// Skip until asset id (backward pagination)
     pub before: Option<String>,
-    /// Retrieve the first n coins in order (forward pagination)
+    /// Retrieve the first n asset balances in order (forward pagination)
     pub first: Option<i32>,
     /// Retrieve the last n coins in order (backward pagination).
     /// Can't be used at the same time as `first`.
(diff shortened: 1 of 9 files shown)


fuel-client/src/client/schema/contract.rs
```diff
@@ -55,6 +56,85 @@ pub struct ContractIdFragment {
     pub id: ContractId,
 }
 
+#[derive(cynic::InputObject, Clone, Debug)]
+#[cynic(schema_path = "./assets/schema.sdl")]
+pub struct ContractBalanceFilterInput {
+    /// Filter coins based on the `owner` field
```
Voxelot, line 62:
```suggestion
    /// Filter asset balances based on the `contract` field
```

fuel-client/src/client/schema/contract.rs
```diff
@@ -55,6 +56,85 @@ pub struct ContractIdFragment {
...
+    pub contract: ContractId,
+}
+
+#[derive(cynic::FragmentArguments, Debug)]
+pub struct ContractBalancesConnectionArgs {
+    /// Filter coins based on a filter
```
Voxelot, line 68:
```suggestion
    /// Filter balances based on a filter
```

fuel-client/src/client/schema/contract.rs
```diff
@@ -55,6 +56,85 @@ pub struct ContractIdFragment {
...
+    filter: ContractBalanceFilterInput,
+    /// Skip until coin id (forward pagination)
```
Voxelot, line 70:
```suggestion
    /// Skip until asset id (forward pagination)
```

fuel-client/src/client/schema/contract.rs
```diff
@@ -55,6 +56,85 @@ pub struct ContractIdFragment {
...
+    pub after: Option<String>,
+    /// Skip until coin id (backward pagination)
```
Voxelot, line 72:
```suggestion
    /// Skip until asset id (backward pagination)
```

fuel-client/src/client/schema/contract.rs
```diff
@@ -55,6 +56,85 @@ pub struct ContractIdFragment {
...
+    pub before: Option<String>,
+    /// Retrieve the first n coins in order (forward pagination)
+    pub first: Option<i32>,
+    /// Retrieve the last n coins in order (backward pagination).
```
Voxelot, line 76:
```suggestion
    /// Retrieve the last n asset balances in order (backward pagination).
```

fuel-core/src/database/contracts.rs
```diff
@@ -49,6 +65,23 @@ impl Storage<ContractId, UtxoId> for Database {
...
+    pub fn contract_balances(
+        &self,
+        contract: ContractId,
+        start_asset: Option<AssetId>,
+        direction: Option<IterDirection>,
+    ) -> impl Iterator<Item = Result<AssetId, Error>> + '_ {
+        self.iter_all::<Vec<u8>, Word>(
+            BALANCES,
+            Some(contract.as_ref().to_vec()),
+            start_asset.map(|b| contract_asset_id_id_key(&contract, &b)),
```
Voxelot, line 78: [nit] consider reusing the same compound key wrapper type used for fetching and inserting balances in other places:

```suggestion
            start_asset.map(|b| MultiKey::new((&contract, &b)).as_ref().to_vec()),
```

fuel-core/src/schema/contract.rs
```diff
@@ -1,9 +1,11 @@
 use crate::database::{Database, KvStoreError};
 use crate::schema::scalars::{AssetId, ContractId, HexString, Salt, U64};
+use crate::state::IterDirection;
+use async_graphql::connection::{query, Connection, Edge, EmptyFields};
```
Voxelot, line 4: [nit] nest `async_graphql` imports

fuel-core/src/database/contracts.rs
ControlCplusControlV: I had seen this method of conversion somewhere else, but `MultiKey` is much cleaner. I committed this but had to do it myself because this also means the removal of 2 functions in the same file and the addition of an import

fuel-core/src/schema/contract.rs
```diff
@@ -113,4 +121,112 @@ impl ContractBalanceQuery {
...
+        ctx: &Context<'_>,
+        filter: ContractBalanceFilterInput,
+        first: Option<i32>,
+        after: Option<String>,
+        last: Option<i32>,
+        before: Option<String>,
+    ) -> async_graphql::Result<Connection<AssetId, ContractBalance, EmptyFields, EmptyFields>> {
+        let db = ctx.data_unchecked::<Database>().clone();
+
+        let balances: Vec<ContractBalance> = db
```
Voxelot, line 136: this should be moved inside of the query section below, where the start asset and number of results to fetch are specified. 

Otherwise, if we collect the iterator into a vec now, and then filter it based on the query this is very inefficient. 

For example, let's say I have an exchange contract with 1000 asset balances. Even if my application only showed 5 balances per page, collecting the iterator right will cause all 1000 to load into memory, and then discard 995 of the results below. 

We should use the before/after cursors to tell the database where to start the iterator, to avoid loading all these extra values into memory. Then we should only consume as many items from the database iterator as the user requested, so that we don't fetch anything past that point.

Move this into the query section below, and populate the two `None` values with the query params. Only collect once the iterator is configured to touch the least amount of state.

Voxelot: We should update that in a separate PR. Coin balances is more complicated, because the cursor is across asset id's, but to compute a balance it needs to sum up all the coins for each asset. We need to be careful to ensure we don't accidentally compute partial balances by applying the paging params like first/last directly to the coins iterator.

ControlCplusControlV: Ok, will resolve this one first and make that a seperate issue then

fuel-core/src/schema/contract.rs
ControlCplusControlV: Fixed in recent PR

fuel-core/src/schema/contract.rs
```diff
@@ -1,9 +1,11 @@
 use crate::database::{Database, KvStoreError};
 use crate::schema::scalars::{AssetId, ContractId, HexString, Salt, U64};
-use async_graphql::{Context, Object};
+use crate::state::IterDirection;
+use async_graphql::connection::{query, Connection, Edge, EmptyFields};
+use async_graphql::{Context, InputObject, Object};
```
Voxelot, line 5:
```suggestion
use async_graphql::{
    connection::{query, Connection, Edge, EmptyFields},
    Context, InputObject, Object,
};
```

fuel-core/src/schema/contract.rs
```diff
@@ -1,9 +1,11 @@
...
 use fuel_storage::Storage;
 use fuel_vm::prelude::Contract as FuelVmContract;
-
+use std::iter::IntoIterator;
```
Voxelot, line 8:
```suggestion
use std::iter::IntoIterator;

```

fuel-core/src/schema/contract.rs
```diff
@@ -113,4 +121,112 @@ impl ContractBalanceQuery {
...
+                let end;
+
+                if direction == IterDirection::Forward {
+                    start = after;
+                    end = before;
+                } else {
+                    start = before;
+                    end = after;
+                }
+                // impl Iterator<Item = ContractBalance>
```
Voxelot, line 163:
```suggestion

```

fuel-core/src/schema/contract.rs
```diff
@@ -113,4 +121,112 @@ impl ContractBalanceQuery {
...
+                            .unwrap();
+
+                        let amount = result.unwrap_or(0);
+
+                        ContractBalance {
+                            contract: filter.contract.into(),
+                            amount,
+                            asset_id,
+                        }
+                    })
```
Voxelot, line 184: This call to collect will load every balance after the starting asset id. We need to limit how many items the iterator will possibly consume:

```suggestion
                    })
                    .take(records_to_fetch)
```

fuel-core/src/schema/contract.rs
```diff
@@ -113,4 +121,112 @@ impl ContractBalanceQuery {
...
+                    .map(|balance| -> ContractBalance {
+                        let asset_id = balance.unwrap();
+
+                        let result =
+                            fuel_vm::storage::InterpreterStorage::merkle_contract_asset_id_balance(
+                                &db,
+                                &filter.contract.into(),
+                                &asset_id,
+                            )
+                            .unwrap();
```
Voxelot, line 175: unsafe unwrap, this would crash the whole server if a spurious I/O error occurred.

fuel-core/src/schema/contract.rs
```diff
@@ -113,4 +121,112 @@ impl ContractBalanceQuery {
...
+                if direction == IterDirection::Forward {
+                    start = after;
+                    end = before;
+                } else {
+                    start = before;
+                    end = after;
+                }
+                // impl Iterator<Item = ContractBalance>
+                let mut balances = db
+                    .contract_balances(filter.contract.into(), start, Some(IterDirection::Forward))
```
Voxelot, line 165: You've hardcoded the iterator direction instead of using the direction calculated above. When using backwards pagination we need to tell rocksdb to start from the end of the table instead of the beginning.

See the graphql connection specs for more details: https://relay.dev/graphql/connections.htm#sec-Forward-pagination-arguments

fuel-core/src/schema/contract.rs
```diff
@@ -113,4 +121,112 @@ impl ContractBalanceQuery {
...
+                            contract: filter.contract.into(),
+                            amount,
+                            asset_id,
+                        }
+                    })
+                    .collect::<Vec<ContractBalance>>()
+                    .into_iter();
+
+                if direction == IterDirection::Reverse {
+                    balances = balances.rev().collect::<Vec<ContractBalance>>().into_iter();
```
Voxelot, line 189: why do we reverse here and then re-reverse it again later?

fuel-core/src/schema/contract.rs
```diff
@@ -113,4 +121,112 @@ impl ContractBalanceQuery {
...
+                }
+                if let Some(start) = start {
+                    balances = balances
+                        .skip_while(|balance| balance.asset_id == start)
+                        .collect::<Vec<ContractBalance>>()
+                        .into_iter();
+                }
```
Voxelot, line 196: I don't think this is necessary
```suggestion
```

fuel-core/src/schema/contract.rs
```diff
@@ -113,4 +121,112 @@ impl ContractBalanceQuery {
...
+                let mut started = None;
+                if start.is_some() {
+                    // skip initial result
+                    started = balances.next();
```
Voxelot, line 200: we should do this before mapping the iterator above

fuel-tests/tests/contract.rs
```diff
@@ -44,3 +45,38 @@ async fn test_contract_balance() {
         assert_eq!(balance, test_bal);
     }
 }
+
+#[rstest]
+#[case(PageDirection::Forward)]
+#[case(PageDirection::Backward)]
+#[tokio::test]
+async fn test_first_5_contract_balances(#[case] direction: PageDirection) {
```
Voxelot, line 53: instead of testing only the "first 5", could we parameterize the number of balances, the start cursor and the page size? That way we can also test whether things like has_previous_page or has_next_page work correctly, or that the number of results returned is always correct etc

Voxelot: Alternately, maybe we can file a tech debt issue to refactor all pagination logic into a common function across all APIs and then we only need these kinds of tests in one place instead of every api that has pagination logic.

Voxelot: https://github.com/FuelLabs/fuel-core/issues/382

fuel-tests/tests/contract.rs
```diff
@@ -1,6 +1,7 @@
 use crate::helpers::{TestContext, TestSetupBuilder};
+use fuel_gql_client::client::{PageDirection, PaginationRequest};
 use fuel_vm::prelude::*;
-
+use rstest::rstest;
```
Voxelot, line 4: keep imports separated from other vars

```suggestion
use rstest::rstest;

```

fuel-core/src/schema/contract.rs
```diff
@@ -113,4 +124,87 @@ impl ContractBalanceQuery {
...
+                        let balance = balance?;
+
+                        Ok(ContractBalance {
+                            contract: filter.contract.into(),
+                            amount: balance.1,
+                            asset_id: balance.0,
+                        })
+                    })
+                    .collect();
+                let mut balances = balances?;
```
Voxelot, line 187: [nit] you could use the turbofish syntax to combine these lines:

```suggestion
                let balances  = balances_iter
                    .take(records_to_fetch + 1)
                    .map(|balance| {
                        let balance = balance?;

                        Ok(ContractBalance {
                            contract: filter.contract.into(),
                            amount: balance.1,
                            asset_id: balance.0,
                        })
                    })
                    .collect::<Result<Vec<ContractBalance>, KvStoreError>>()?;
```

fuel-core/src/schema/contract.rs
```diff
@@ -113,4 +124,86 @@ impl ContractBalanceQuery {
...
+                // Calculate direction of which to iterate through rocksdb
+                let (records_to_fetch, direction) = if let Some(first) = first {
+                    (first, IterDirection::Forward)
+                } else if let Some(last) = last {
+                    (last, IterDirection::Reverse)
+                } else {
+                    (0, IterDirection::Forward)
+                };
+
+                if (first.is_some() && before.is_some()) || (after.is_some() && before.is_some()) {
```
Voxelot, line 154: should we also prevent `last.is_some && after.is_some()`?

Voxelot (approved):

AlicanC (approved):

PR #377, opened 2022-05-31
Validators get

rakita:
Implemented `validator_get` function that can get validator set that is in past in regards to the current finalized set.

StakingDiff DB table is extended to contain both old and new validator `consensus_key`.

### DIFF
--- a/fuel-core-interfaces/src/relayer.rs
+++ b/fuel-core-interfaces/src/relayer.rs
@@ -7,16 +7,25 @@ use tokio::sync::oneshot;
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[derive(Debug, Clone)]
 pub struct StakingDiff {
-    /// Validator registration, it can be new consensus address if registered or
-    /// None if unregistration happened.
-    pub validators: HashMap<Address, Option<Address>>,
+    /// Validator registration, it is pair of old consensu key and new one, where consensus address
+    /// if registered is Some or None if unregistration happened.
+    pub validators: HashMap<Address, ValidatorDiff>,
     /// Register changes for all delegations inside one da block.
     pub delegations: HashMap<Address, Option<HashMap<Address, ValidatorStake>>>,
 }
 
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug, Clone)]
+pub struct ValidatorDiff {
+    /// Previous consensus key, None if validator was not set.
+    pub previous_consensus_key: Option<Address>,
+    /// New consensus key or None if unregistration happened.
+    pub new_consensus_key: Option<Address>,
+}
+
 impl StakingDiff {
     pub fn new(
-        validators: HashMap<Address, Option<Address>>,
+        validators: HashMap<Address, ValidatorDiff>,
         delegations: HashMap<Address, Option<HashMap<Address, ValidatorStake>>>,
     ) -> Self {
         Self {
@@ -53,10 +62,9 @@ pub trait RelayerDb:
     }
 
     /// Query delegate index to find list of blocks that delegation changed
-    /// iterate over list of indexed to find height that is less and closest to da_height
+    /// iterate over list of indexed to find height that is less but closest to da_height
     /// Query that block StakeDiff to find actual delegation change.
-    /// TODO not needed
-    async fn get_last_delegation(&mut self,delegate: &Address, da_height: DaBlockHeight) ->  Option<HashMap<Address,u64>> {
+    async fn get_first_lesser_delegation(&mut self,delegate: &Address, da_height: DaBlockHeight) ->  Option<HashMap<Address,u64>> {
         // get delegate index
         let delegate_index = Storage::<Address,Vec<DaBlockHeight>>::get(self,delegate).expect("Expect to get data without problem")?;
         let mut last_da_height = 0;
--- a/fuel-relayer/src/finalization_queue.rs
+++ b/fuel-relayer/src/finalization_queue.rs
@@ -7,11 +7,11 @@ use ethers_core::types::{Log, H160};
 use ethers_providers::Middleware;
 use fuel_core_interfaces::{
     model::{BlockHeight, DaBlockHeight, SealedFuelBlock},
-    relayer::{RelayerDb, StakingDiff},
+    relayer::{RelayerDb, StakingDiff, ValidatorDiff},
 };
 use fuel_tx::{Address, Bytes32};
 use std::{
-    collections::{HashMap, VecDeque},
+    collections::{hash_map::Entry, HashMap, VecDeque},
     sync::Arc,
 };
 use tracing::{debug, error, info, warn};
@@ -85,8 +85,9 @@ impl FinalizationQueue {
     pub async fn get_validators(
         &mut self,
         da_height: DaBlockHeight,
+        db: &mut dyn RelayerDb,
     ) -> Option<HashMap<Address, (u64, Option<Address>)>> {
-        self.validators.get(da_height).await
+        self.validators.get(da_height, db).await
     }
 
     pub fn clear(&mut self) {
@@ -265,17 +266,42 @@ impl FinalizationQueue {
         }
         self.remove_bundled_reverted_events();
 
+        //TODO to be paranoid, recheck every block and all events got from eth client.
+
+        let mut validators: HashMap<Address, Option<Address>> = HashMap::new();
         while let Some(diff) = self.pending.front_mut() {
             if diff.da_height > finalized_da_height {
                 break;
             }
             info!("flush eth log:{:?} diff:{:?}", diff.da_height, diff);
-            //TODO to be paranoid, recheck events got from eth client.
+
+            let validator_diff: HashMap<Address, ValidatorDiff> = diff
+                .validators
+                .iter()
+                .map(|(val, &new_consensus_key)| {
+                    let previous_consensus_key = match validators.entry(*val) {
+                        Entry::Occupied(mut entry) => {
+                            core::mem::replace(entry.get_mut(), new_consensus_key)
+                        }
+                        Entry::Vacant(entry) => {
+                            entry.insert(new_consensus_key);
+                            self.validators.set.get(val).and_then(|(_, i)| *i)
+                        }
+                    };
+                    (
+                        *val,
+                        ValidatorDiff {
+                            previous_consensus_key,
+                            new_consensus_key,
+                        },
+                    )
+                })
+                .collect();
 
             // apply staking diffs
             db.insert_staking_diff(
                 diff.da_height,
-                &StakingDiff::new(diff.validators.clone(), diff.delegations.clone()),
+                &StakingDiff::new(validator_diff, diff.delegations.clone()),
             )
             .await;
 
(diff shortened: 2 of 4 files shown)


fuel-core-interfaces/src/relayer.rs
```diff
@@ -7,16 +7,23 @@ use tokio::sync::oneshot;
...
+    /// if registered is Some or None if unregistration happened.
+    pub validators: HashMap<Address, ValidatorDiff>,
     /// Register changes for all delegations inside one da block.
     pub delegations: HashMap<Address, Option<HashMap<Address, ValidatorStake>>>,
 }
 
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug, Clone)]
+pub struct ValidatorDiff {
+    pub old: Option<Address>,
```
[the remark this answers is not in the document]
rakita: They do sounds wongly, renamed them to `previous_consensus_key` and `new_consensus_key`

fuel-relayer/src/finalization_queue.rs
```diff
@@ -250,17 +251,37 @@ impl FinalizationQueue {
...
+                    let old = match validators.entry(*val) {
+                        Entry::Occupied(mut entry) => {
+                            let old = *entry.get();
+                            entry.insert(new);
+                            old
+                        }
+                        Entry::Vacant(entry) => {
+                            entry.insert(new);
+                            self.validators.set.get(val).and_then(|(_, i)| *i)
+                        }
```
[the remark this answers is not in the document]
rakita: I have a need for old value or to get value from `self.validators.set`. I did clean it up and used `mem::replace`

Voxelot:
This seems like a lot of code changes, is there any way this could be broken into smaller PRs?

rakita:
> This seems like a lot of code changes, is there any way this could be broken into smaller PRs?

It was not rebased and contained merged relayer commit

fuel-core-interfaces/src/relayer.rs
```diff
@@ -53,10 +62,9 @@ pub trait RelayerDb:
     }
 
     /// Query delegate index to find list of blocks that delegation changed
-    /// iterate over list of indexed to find height that is less and closest to da_height
+    /// iterate over list of indexed to find height that is less but closest to da_height
     /// Query that block StakeDiff to find actual delegation change.
-    /// TODO not needed
-    async fn get_last_delegation(&mut self,delegate: &Address, da_height: DaBlockHeight) ->  Option<HashMap<Address,u64>> {
+    async fn get_first_lesser_delegation(&mut self,delegate: &Address, da_height: DaBlockHeight) ->  Option<HashMap<Address,u64>> {
```
Voxelot, line 67: fmt
```suggestion
    async fn get_first_lesser_delegation(&mut self, delegate: &Address, da_height: DaBlockHeight) -> Option<HashMap<Address, u64>> {
```

fuel-relayer/src/finalization_queue.rs
```diff
@@ -265,17 +266,42 @@ impl FinalizationQueue {
...
+                .validators
+                .iter()
+                .map(|(val, &new_consensus_key)| {
+                    let previous_consensus_key = match validators.entry(*val) {
+                        Entry::Occupied(mut entry) => {
+                            core::mem::replace(entry.get_mut(), new_consensus_key)
+                        }
+                        Entry::Vacant(entry) => {
+                            entry.insert(new_consensus_key);
+                            self.validators.set.get(val).and_then(|(_, i)| *i)
```
Voxelot, line 288: [nit] isn't this the same thing as `new_consensus_key`?

```suggestion
                            new_consensus_key
```

Voxelot: Also, if this was vacant before, shouldn't previous_consensus_key be None?

fuel-relayer/src/validators.rs
```diff
@@ -26,16 +26,94 @@ impl Validators {
...
+                // Probably there is eth client sync problem
+                error!(
+                    "current height {} is less then requested validator set height: {da_height}",
+                    self.da_height
+                );
+                return None;
+            }
+            std::cmp::Ordering::Equal => {
+                // unusual but do nothing
+                info!("Get last finalized set height: {da_height}");
```
Voxelot, line 47: [nit] Seems like `info!`-level may be too noisy?

```suggestion
                debug!("Get last finalized set height: {da_height}");
```

fuel-relayer/src/validators.rs
```diff
@@ -26,16 +26,94 @@ impl Validators {
...
+                let set = self
+                    .set
+                    .iter()
+                    .filter(|(_, (stake, consensus_key))| *stake != 0 && consensus_key.is_some())
+                    .map(|(k, v)| (*k, *v))
+                    .collect();
+                return Some(set);
+            }
+            std::cmp::Ordering::Greater => {
+                // slightly drift to past
```
Voxelot, line 57: [nit] This code is slightly awkward to follow with the multiple return statements above and then nothing in this match arm here. Ideally, we could leverage implicit returns from each match arm to consolidate the control flow and code readability.

fuel-relayer/src/validators.rs
```diff
@@ -26,16 +26,94 @@ impl Validators {
...
     }
 
-    /// Get validator set
+    /// Get validator set for da_height.
     pub async fn get(
-        &mut self,
+        &self,
         da_height: DaBlockHeight,
+        db: &mut dyn RelayerDb,
     ) -> Option<HashMap<Address, (u64, Option<Address>)>> {
```
Voxelot, line 34: [nit] improve readability by aliasing return types.

```suggestion
    ) -> Option<HashMap<StakingAddress, (u64, Option<ConsensusAddress>)>> {
```

fuel-core-interfaces/src/relayer.rs
```diff
@@ -7,16 +7,25 @@ use tokio::sync::oneshot;
...
+    pub validators: HashMap<Address, ValidatorDiff>,
     /// Register changes for all delegations inside one da block.
     pub delegations: HashMap<Address, Option<HashMap<Address, ValidatorStake>>>,
 }
 
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug, Clone)]
+pub struct ValidatorDiff {
+    /// Previous consensus key, None if validator was not set.
+    pub previous_consensus_key: Option<Address>,
```
Voxelot, line 21: [nit] would it be more correct to call this a consensus_address, or should this type really be a Option<PublicKey>?

```suggestion
    pub previous_consensus_key: Option<PublicKey>,
```

fuel-relayer/src/validators.rs
```diff
@@ -26,16 +26,94 @@ impl Validators {
...
+            }
+            std::cmp::Ordering::Greater => {
+                // slightly drift to past
+            }
+        }
+
+        let mut validators = self.set.clone();
+        // get staking diffs to revert from our current set.
+        let diffs = db
+            .get_staking_diffs(da_height + 1, Some(self.da_height))
```
Voxelot, line 64: Should we use `checked_add` to prevent overflow from a user input causing use to diff way more history than expected?

fuel-relayer/src/validators.rs
```diff
@@ -26,16 +26,94 @@ impl Validators {
...
+            .await;
+
+        for (diff_height, diff) in diffs.into_iter().rev() {
```
Voxelot, line 67: since `.rev()` requires collecting an entire iterator into memory in order to reverse it, it seems like it would be more efficient to just use backward pagination on the underlying db impl so that we iterate it the reversed order by default. Then we could stream / iterate to an arbitrary point in history without loading all the diffs into memory at once.

fuel-relayer/src/validators.rs
```diff
@@ -26,16 +26,94 @@ impl Validators {
...
+            {
+                if let Some((_, consensus_key)) = validators.get_mut(&validator) {
+                    *consensus_key = previous_consensus_key;
+                } else {
+                    panic!("Validator should be present when reverting diff");
+                }
+            }
+
+            // for every delegates, cache it and if it is not in cache query db's delegates_index for earlier delegate set.
+            for (delegator, delegation) in diff.delegations.into_iter() {
```
Voxelot, line 85: [nit] unneeded cast to iter in for loop
```suggestion
            for (delegator, delegation) in diff.delegations {
```

Voxelot (commented):
my only non-nit feedback is wrt previous consensus key always being set, even for previously nonexistent validators.

fuel-core-interfaces/src/relayer.rs
vlopes11: Keep in mind we use uncompressed public keys, so they need 64 bytes while address is only 32

fuel-relayer/src/validators.rs
vlopes11: Maybe even create concrete types for these somewhat complex returns. Its a bit hard to read if we don't

fuel-relayer/src/finalization_queue.rs
rakita: if it is inside `validator` it means that the previous diff changed it in some way, but if it is not inside `validator` we need to fetch original (previous) value and that is why we call `self.validators`

fuel-relayer/src/validators.rs
rakita: New types look a lot nicer and more readable, but I will create new branch to propagate this. My preference would be `StakingPublicKey` and `ConsensusPublicKey`. @Voxelot & @vlopes11 is this naming more aligned or does `StakingAddress` and `ConsensusAddress` sound better for you?

fuel-core-interfaces/src/relayer.rs
rakita: New type will ba made

fuel-relayer/src/validators.rs
rakita: This is not just a nit, as you said memory footprint can grow and in that case, it would be better to have an iterator here.  The reason why I am fine with the current code is that we don't expect a lot of movement on validators staking and in this case it is more efficient to fetch all values from db at the same time.

fuel-relayer/src/validators.rs
rakita: Or to be more precise it is not `StakingAddress` but more like `ValidatorAddress` or `ValidatorPublicKey`

fuel-relayer/src/finalization_queue.rs
Voxelot: ahh ok I was confused by all the different validator sets and what they were being used for. So IIUC, to determine the previous key, we first check a cross-block cache scoped to this function call for any current keys, and if that's empty we fallback to the latest validator set snapshot loaded from the database.

I don't see anything in this function to update the validator set or reload it except during initialization. So if there were several calls to this commit_diffs function, could we end up using the wrong previous keys once the `validators` hashmap drops between calls?

rakita: Yep, that 's the logic.

update for `self.validator.set` is done in function below, `bump_set_to_da_height` where i read from db and update it, this is not the best but functions were already made before last iteration of relayer was devised

Voxelot (approved):

adlerjohn (approved):
utACK

fuel-relayer/src/validators.rs
Voxelot: I guess what I'm not sure about is whether we use the consensus address (32 bytes) or key (64 bytes) in the contracts etc. We could go really generic for now and just do the bare minimum to differentiate these vars, i.e. `ValidatorId` and `ConsensusSessionId`. I like SessionId because it illustrates the transitory nature of the key/address in relation to the ValidatorId.

rakita: `Consensus` part is a public key but in the compressed form of 32bytes

PR #379, opened 2022-06-01
evict reverted pending block commitments [relayer]

Voxelot:
Attempt to simplify the pending_blocks structure a lot by simply removing reverted pending blocks since they aren't used by anything. It's hard to say if everything works as intended because the test coverage for handling revert logic in the finalization_queue module isn't implemented.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1270,9 +1270,9 @@ dependencies = [
 
 [[package]]
 name = "ctor"
-version = "0.1.21"
+version = "0.1.22"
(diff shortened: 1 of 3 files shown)


vlopes11 (commented):
Its definitely the right direction but we have a CI nit

adlerjohn:
To be merged into #345

Voxelot:
closing as @rakita pointed out we may want to keep track of pending fuel block commitments on L1 to avoid contention.

PR #380, opened 2022-06-02
Rakita/project skeleton

rakita:
Introduced missing modules into fuel core:
* `fuel-block-executor`
* `fuel-block-importer`
* `fuel-block-producer`
* `fuel-core-bft`
* `fuel-sync`

Introduces events defined in wiki: https://github.com/FuelLabs/fuel-core/wiki/Interfaces-between-fuel-client-modules

Bounded them together and introduced a skeleton for event-driven architecture.


fuel-core-interfaces/src/bft.rs
```diff
@@ -0,0 +1,17 @@
+use super::model::{FuelBlock, SealedFuelBlock};
+use anyhow::Result;
+use std::sync::Arc;
+use tokio::sync::oneshot;
+
+pub enum BFTMpsc {
```
vlopes11, line 6: I don't think we should define the specifics of the behavior in the consumer level.

Instead, we can use [fuel_bft::Message](https://github.com/FuelLabs/fuel-bft/blob/7acc238ee02ec8cbc726deb97519b075fc0ab538/src/reactor/message.rs#L11-L20) so we inherit the expected behavior of the consensus implementation.

The proposed approach will lead to duplicated logic and restriction of functionalities since this interface cannot be aware of the capabilities of fuel-bft.

As example, the variant `BFTMpsc::Stop` is supersede by [Notification::Kill](https://github.com/FuelLabs/fuel-bft/blob/7acc238ee02ec8cbc726deb97519b075fc0ab538/src/reactor/message/notification.rs#L9-L10)

Defining the structures here might lead to either duplication or circular dependencies since either fuel-bft will reimplement these variants or will require fuel-core

fuel-core-interfaces/src/bft.rs
```diff
@@ -0,0 +1,17 @@
...
+    CheckBlockConsensus {
+        block: Arc<SealedFuelBlock>,
+        ret: oneshot::Sender<Result<()>>,
+    },
+    CheckBlockLeader {
+        block: Arc<FuelBlock>,
```
vlopes11, line 12: I just noticed that we don't have the total committed rounds as part of the block header. We need to add it because its a requirement for bft to calculate the current round; also provided the timestamp.

For more info:
https://github.com/FuelLabs/fuel-bft/blob/7acc238ee02ec8cbc726deb97519b075fc0ab538/src/reactor.rs#L67-L79

rakita:
> It seems like we could make a standard trait or even generic service implementation that is configurable for each component, rather than duplicating this pattern around?

Signatures of functions are not the same.

fuel-core-interfaces/src/bft.rs
rakita: `fuel-bft` has a very opinionated interface, but in `fuel-core` it is just 1 out of 7/8 modules.
This is binding that I talked about in chat, and it can be done inside `fuel-core-bft`.

fuel-core-interfaces/src/bft.rs
rakita: This should be added, can you create a github issue for it as it is not related to this pull request.

Voxelot:
> Signatures of the functions are not the same.

Seems like workaroundable problem using generics? I was just thinking it would make it cleaner to do things like kill and join on all the services if they all used a standard interface to manage their lifecycle, similar to actors.

fuel-block-importer/src/service.rs
```diff
@@ -0,0 +1,48 @@
...
+            *join = Some(tokio::spawn(async {}));
+        }
+    }
+
+    pub async fn stop(&self) -> Option<JoinHandle<()>> {
+        let join = self.join.lock().take();
+        if join.is_some() {
+            let _ = self.sender.send(ImportBlockMpsc::Stop);
+        }
+        join
```
Voxelot, line 38: I feel like these Stop messages are not necessary and don't add much. We could simply abort the task:

```suggestion
        let join = self.join.lock().take();
        if let Some(handle) = join {
            handle.abort();
        }
        join
```

rakita: `Stop` allow the module to gracefully finish whatever it was working on or do cleanup if needed.

rakita:
> > Signatures of the functions are not the same.
> 
> Seems like a workaroundable problem using generics? I was just thinking it would make it cleaner to do things like kill and join on all the services if they all used a standard interface to manage their lifecycle, similar to actors.

I didn't like to overengineer, atm it seems good as it is in my opinion, and in the future, if there is a need for it we can always add it.

fuel-core-interfaces/src/bft.rs
Voxelot, line 6: [nit] 
```suggestion
pub enum BftMpsc {
```

Voxelot, line 6: [nit] 
```suggestion
pub enum BftMpsc {
```

Voxelot: I think it's fine for fuel-core to define it's own messages between components, and wrap / adapt the message interfaces defined in fuel-bft.

Voxelot: I think it's fine for fuel-core to define it's own messages between components, and wrap / adapt the message interfaces defined in fuel-bft.

fuel-core-interfaces/src/bft.rs
Voxelot: tracked: https://github.com/FuelLabs/fuel-core/issues/399

fuel-core-interfaces/src/block_importer.rs
```diff
@@ -1,4 +1,9 @@
+use super::model::{ConsensusVote, FuelBlock, SealedFuelBlock};
+use fuel_types::Bytes32;
+use std::sync::Arc;
+
 /// Currently just placeholder for new block included and new block created events.
+/// TODO remove this after relayer pull request passes
```
Voxelot, line 6: Not sure what this todo is referring to. New blocks come either from the synchronizer or peers during bft, not the relayer.

Voxelot (approved):
while very skeletal, I think this is a fine place to start for now. We can address issues regarding message variants and their fields in subsequent PRs related to the component in question.

Voxelot (approved):

PR #394, opened 2022-06-06
p2p messages with dummy consensus Vote

leviathanbeak:
This in favor of #313 

I have defined:
- Gossipsub Messages
- Request/Response Messages

In order to unblock myself I have added a placeholder for Vote struct from FuelLabs/fuel-bft#9
in `fuel-core-interfaces` .

In this PR I also fixed a bug in `serde` feature in `fuel-core-interfaces` where `time` field of `FuelBlockHeader` is not properly serialized, by importing `serde` feature from `chrono` dep.

I have added a `max_response_size` field to `BincodeCodec` since max FuelBlock size is configurable and we need to configure max size of both Request and Response messages. Max request size is pretty straightforward but response message size depends on max block size.

### DIFF
--- a/fuel-p2p/src/service.rs
+++ b/fuel-p2p/src/service.rs
@@ -376,6 +376,7 @@ mod tests {
     #[instrument]
     async fn gossipsub_exchanges_messages() {
         use crate::gossipsub::messages::GossipsubMessage as FuelGossipsubMessage;
+        use fuel_tx::Transaction;
 
         let mut p2p_config = build_p2p_config("gossipsub_exchanges_messages");
         let topics = vec!["create_tx".into(), "send_tx".into()];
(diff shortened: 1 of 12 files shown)


leviathanbeak:
seems adding `#[serde(with = "ts_seconds")]` to field `time` breaks the graphql Block query and `submit_utxo_verified_tx_with_min_gas_price` test

leviathanbeak:
> seems adding `#[serde(with = "ts_seconds")]` to field `time` breaks the graphql Block query and consequently `submit_utxo_verified_tx_with_min_gas_price` test

Turned out adding that wasn't necessary, just importing `serde` feature from `chrono` was sufficient.

fuel-p2p/src/service.rs
```diff
@@ -183,6 +188,7 @@ mod tests {
             network_name: network_name.into(),
             address: IpAddr::V4(Ipv4Addr::from([0, 0, 0, 0])),
             tcp_port: 4000,
+            max_block_size: 100_000,
```
adlerjohn, line 191: Make this a parameter

PR #402, opened 2022-06-10
Network Orchestrator service proposal

leviathanbeak:
Network Orchestrator is the service that:
- holds and runs p2p service
- holds multiple `tokio::sync::mpsc::Sender` values for propagating p2p messages to different components
- holds and listens from `tokio::sync::mpsc::Receiver` to get messages from different fuel-core components

This initial version is (loosely) based on [miro design](https://miro.com/app/board/o9J_lhccOik=/).

Apart from general design I would need feedback on specific components


fuel-p2p/src/orchestrator.rs
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    pub async fn run(&mut self) {
+        loop {
+            tokio::select! {
+                p2p_event = self.p2p_service.next_event() => {
+                    if let FuelP2PEvent::Behaviour(behaviour_event) = p2p_event {
+                        match behaviour_event {
+                            FuelBehaviourEvent::GossipsubMessage { message, topic_hash, peer_id } => {
```
leviathanbeak, line 59: Gossipsub Messages, I would need to know for each message to which component I send it to.

fuel-p2p/src/orchestrator.rs
```diff
@@ -0,0 +1,98 @@
...
+                                    },
+                                    GossipsubMessage::NewBlock(block) => {
+
+                                    },
+                                    GossipsubMessage::ConensusVote(vote) => {
+
+                                    },
+                                }
+                            },
+                            FuelBehaviourEvent::RequestMessage { request_message, request_id } => {
```
leviathanbeak, line 72: Here we receive a request message from another peer,
currently this is only a request for a FuelBlock,

I need to get the Block and respond back.

From where do I fetch it?

fuel-p2p/src/orchestrator.rs
```diff
@@ -0,0 +1,98 @@
...
+                                    //let response_message = ResponseMessage::ResponseBlock(block);
+                                    //let _ = self.p2p_service.send_response_msg(request_id, response_message);
+                                });
+                            },
+                            _ => {}
+                        }
+                    }
+
+                },
+                bft_vote = self.rx_bft.recv() => {
```
leviathanbeak, line 84: Which messages do I receive from fuel-bft?

fuel-p2p/src/orchestrator.rs
```diff
@@ -0,0 +1,98 @@
...
+                    if let Some(vote) = bft_vote {
+                        // vote
+                    }
+                },
+
+                block_producer_msg = self.rx_block_producer.recv() => {
```
leviathanbeak, line 90: Which messages do I receive from Block Producer

fuel-p2p/src/orchestrator.rs
rakita: It can be modelled as broadcast, as you would have the ability to subscribe and sniff what is going on in the p2p layer without disturbing other parts of the system. https://github.com/FuelLabs/fuel-core/wiki/Interfaces-between-fuel-client-modules#p2p-channels
But it can be done with mpsc channel

fuel-p2p/src/orchestrator.rs
rakita: `fuel-txpool` and `fuel-relayer` take the trait that db needs to implement, and with that, you can directly access the database to fetch block body/header.

fuel-p2p/src/orchestrator.rs
rakita: Not exactly from block producer, as block-producer only creates the block, the proposed block should come from `core-bft` or most probably from `block-importer` as he is the one that contains pending blocks before they are finalized.

fuel-p2p/src/orchestrator.rs
rakita: I think one thing is missing, the channel that `fuel-sync` is going to use to request blocks from the network.

Maybe it will be cleaner for discussion to model channels in this format: https://github.com/FuelLabs/fuel-core/wiki/Interfaces-between-fuel-client-modules#p2p-channels
and then just apply changes to the wiki afterwards.

fuel-p2p/src/orchestrator.rs
```diff
@@ -0,0 +1,98 @@
...
+    config::P2PConfig,
+    gossipsub::messages::GossipsubMessage,
+    service::{FuelP2PEvent, FuelP2PService},
+};
+
+pub struct NetworkOrchestrator {
+    p2p_service: FuelP2PService,
+
+    // receivers
+    rx_block_producer: Receiver<()>,
```
vlopes11, line 17: This isn't entirely clear. Why would network orchestrator receive `()`?

fuel-p2p/src/orchestrator.rs
```diff
@@ -0,0 +1,98 @@
...
+    rx_bft: Receiver<Vote>,
+
+    // senders
+    tx_synchronizer: Sender<()>,
+    tx_transaction_pool: Sender<Transaction>,
+    tx_bft: Sender<()>,
```
vlopes11, line 23: Also don't understand this `()`. Is it WiP?

fuel-p2p/src/orchestrator.rs
```diff
@@ -0,0 +1,98 @@
...
+        rx_block_producer: Receiver<()>,
+        rx_bft: Receiver<Vote>,
+
+        tx_synchronizer: Sender<()>,
+        tx_transaction_pool: Sender<Transaction>,
+        tx_bft: Sender<()>,
+    ) -> Self {
+        let p2p_service = FuelP2PService::new(local_keypair, p2p_config)
+            .await
+            .unwrap();
```
vlopes11, line 39: Maybe return a result instead of internal panic?

fuel-p2p/src/orchestrator.rs
```diff
@@ -0,0 +1,98 @@
...
+            tx_bft,
+            tx_synchronizer,
+            tx_transaction_pool,
+        }
+    }
+
+    pub async fn run(&mut self) {
+        loop {
+            tokio::select! {
+                p2p_event = self.p2p_service.next_event() => {
```
vlopes11, line 56: instead of nesting all these match patterns, maybe create a single level identity who will check the cases?

fuel-p2p/src/orchestrator.rs
vlopes11: the only thing the network should expect is a vote

fuel-p2p/src/orchestrator.rs
leviathanbeak: @vlopes11 yes, it's a work in progress :)

leviathanbeak:
Most of the reasoning for these changes can be found at: https://fuellabs.slack.com/archives/C01L1HUJSL8/p1656343694067609

It could be argued that it's a bit over-engineered but on the flip side we do not `.clone()` any values from `Arc<T>` that we receive/send so it is definitely cleaner and more performant in that regard.

rakita (approved):
tbh it is hard to do indepth review without knowing libp2p lib. Reviewing the code it looked good so I am okay to approve it.
I think a few things will pop out when starting to integrate p2p with the rest of the system. In wiki It was hard to specify everything in one go here so i see there is inconsistencies as block producer broadcasts block (it is used by multiple parties) but p2p has mpsc that wants to receive it.

fuel-p2p/src/behavior.rs
```diff
@@ -175,34 +194,42 @@ impl<Codec: NetworkCodec> FuelBehaviour<Codec> {
...
+                }
+            }
+            (Ok(_), None) => {
+                debug!("ResponseChannel for {:?} does not exist!", request_id);
+                return Err(ResponseError::ResponseChannelDoesNotExist);
             }
-        } else {
-            debug!("ResponseChannel for {:?} does not exist!", request_id);
-            return Err(ResponseError::ResponseChannelDoesNotExist);
+            _ => {}
```
Voxelot, line 232: Should we log anything in the event of an error?

```suggestion
            (Err(e), channel) => { debug!("unexpected error occurred while processing message response: {:?}", e); }
```

fuel-p2p/src/orchestrator.rs
```diff
@@ -0,0 +1,140 @@
...
+                                let broadcast = GossipsubBroadcastRequest::ConensusVote(vote);
+                                let _ = self.p2p_service.publish_message(broadcast);
+                            }
+                        }
+                    } else {
+                        warn!(target: "fuel-libp2p", "Failed to receive P2PRequestEvent");
+                    }
+                }
+
+
```
Voxelot, line 136: normally `cargo fmt` would trim this, but don't think it works inside of select! macros. 
```suggestion
```

fuel-p2p/src/orchestrator.rs
```diff
@@ -0,0 +1,140 @@
...
+                                }
+                            },
+                            FuelBehaviourEvent::RequestMessage { request_message, request_id } => {
+                                match request_message {
+                                    RequestMessage::RequestBlock(block_height) => {
+                                        let db = self.db.clone();
+
+                                        self.outbound_responses.push(
+                                            Box::pin(async move {
+                                                db.get_sealed_block(block_height).await.map(|block| (OutboundResponse::ResponseBlock(block), request_id))
```
Voxelot, line 99: It may make sense to have some other service/module handle fetching blocks from the DB for the orchestrator. For instance, if there are many outstanding requests (e.g. lots of full-syncs) having a dedicated block-fetching worker on the [`spawn_blocking`](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) threadpool would help avoid stalling the main async runtime.

fuel-p2p/src/gossipsub/topics.rs
```diff
@@ -0,0 +1,59 @@
...
+            hash if hash == &new_tx_topic.0 => Some(GossipTopicTag::NewTx),
+            hash if hash == &new_block_topic.0 => Some(GossipTopicTag::NewBlock),
+            hash if hash == &consensus_vote_topic.0 => Some(GossipTopicTag::ConensusVote),
+            _ => None,
+        }
+    }
+
+    /// Given a `GossipsubBroadcastRequest` retruns a `GossipTopic`
+    /// which is broadcast over the network with the serialized inner value of `GossipsubBroadcastRequest`
+    pub fn get_gossipsub_topic(&self, outgoing_request: &GossipsubBroadcastRequest) -> GossipTopic {
```
Voxelot, line 52: Do we have tests to verify all of these lookups and topic encodings are accurate? Since topic setup relies on manual string formatting and setup, it will be harder for the compiler to detect errors during refactoring etc.

fuel-p2p/src/service.rs
```diff
@@ -412,8 +416,8 @@ mod tests {
                             // verifies that we've got at least a single peer address to send message to
                             if !peer_addresses.is_empty() && !message_sent  {
                                 message_sent = true;
-                                let default_tx = FuelGossipsubMessage::NewTx(Transaction::default());
-                                node_a.publish_message(selected_topic.clone(), default_tx).unwrap();
+                                let default_tx = GossipsubBroadcastRequest::NewTx(Arc::new(Transaction::default()));
```
Voxelot, line 419: We should also have tests that ensure message routing and serialization work as expected across _all_ network event types we have. Currently, it seems like we only exercise the `NewTx` and `RequestBlock` code paths in our tests.

Voxelot:
ready to approve once test coverage for new message types is improved.

fuel-p2p/src/gossipsub/topics.rs
leviathanbeak: good observation, added a test that checks all the options

fuel-p2p/src/orchestrator.rs
leviathanbeak: yes `select!` support is terrible

fuel-p2p/src/service.rs
leviathanbeak: added tests for all Gossipsub Messages, as for Request-Response, currently we only support RequestBlock anyway

leviathanbeak:
did some additional cleaning:
- remove `unwrap()` when initializing the Orchestrator::new()
- removed some unused methods
- only made public the Orchestrator and what is needed for it to start

fuel-p2p/src/orchestrator.rs
leviathanbeak: since the fetching itself is already async, shouldn't it be just fine to spawn another async task with `tokio::spawn` instead ?

Voxelot: good point, it should be the responsibility of this async-trait implementor to deal with any potential blocking issues.

fuel-p2p/src/service.rs
```diff
@@ -380,17 +358,64 @@ mod tests {
...
+            GossipsubBroadcastRequest::NewTx(Arc::new(Transaction::default())),
+            4008,
+            4009,
+        )
+        .await;
+    }
 
+    #[tokio::test]
+    #[instrument]
+    async fn gossipsub_broadcast_vote() {
```
Voxelot, line 372: I like these granular clean looking tests

Voxelot (approved):

PR #271, opened 2022-04-18
Deploy fuel-core-dev1 When Merging To Master

ellioty:

PR #403, opened 2022-06-13
Fix CI.yml CI/CD

ellioty:
This PR is meant to fix issues that were not properly addressed in the previous PR: https://github.com/FuelLabs/fuel-core/pull/271

Essentially the Docker step has a dependency on Publishing to Crates, however, we manually ignore this since the Docker step has `if: always()` so all dependencies within the Docker step are currently being ignored.


.github/workflows/ci.yml
```diff
@@ -162,7 +162,7 @@ jobs:
       - cancel-previous-runs
       - lint-toml-files
       - cargo-verifications
-      - publish-crates
+    # - publish-crates
```
adlerjohn, line 165: Why is this commented out?

Voxelot: Agreed, we should not be disabling our entire publishing pipeline from master for the sake of debugging this deployment.

adlerjohn:
Can you update the PR description @ellioty with some information of what this PR is actually for?

.github/workflows/ci.yml
ellioty: This is only a dependency, which actually isn't even being respected, if you see 2 lines below we're doing `if: always()`

I actually meant to completely remove it

Voxelot (approved):

PR #404, opened 2022-06-13
Database/RocksDb Metrics

### LINKED ISSUE #167: Database/RocksDb Metrics
Instrument the fuel-core database to collect telemetry metrics.

Some basic metrics for starters:
- read rate
- write rate
- cache statistics
- (additional metrics deemed useful)

ControlCplusControlV:
Designed to close #167 

**Design**

Uses Lazy Static to host Prometheus counters, which are then written to whenever the database performs read and write operations. This is then hosted on a `/metrics` endpoint using the same design as @uberscott 's rustometer.

**Test**
Test is a bit messy with turbofish, I wasn't sure how to clean it up but if is too ugly I can look for a different way. I ran into some weird behavior with the counter returning inconsistent results when run with all the tests (so it was different between `cargo test` and `cargo test metrics`). I think this is mostly to do with the testing harness starting and stopping tests weirdly async, and won't be replicated in actual use.


ControlCplusControlV:
@uberscott thoughts?

fuel-core/Cargo.toml
```diff
@@ -87,6 +88,7 @@ fuel-vm = { version = "0.11", features = ["serde", "random", "test-helpers"] }
 insta = "1.8"
 
 [features]
-default = ["rocksdb", "debug"]
+prometheus = ["dep:prometheus"]
+default = ["rocksdb","prometheus" ,"debug"]
```
Voxelot, line 92:
```suggestion
default = ["rocksdb", "prometheus", "debug"]
```

fuel-core/src/service/graph_api.rs
```diff
@@ -43,6 +45,7 @@ pub async fn start_server(
         .data(modules.bft.clone());
     let schema = dap::init(schema, params).extension(Tracing).finish();
 
+    #[cfg(not(feature = "prometheus"))]
```
Voxelot, line 48:
```suggestion
```

fuel-core/src/service/graph_api.rs
```diff
@@ -43,6 +45,7 @@ pub async fn start_server(
...
     let router = Router::new()
         .route("/playground", get(graphql_playground))
         .route("/graphql", post(graphql_handler).options(ok))
```
Voxelot, line 51:
```suggestion
        .route("/graphql", post(graphql_handler).options(ok))
        .route("/metrics", get(metrics))
```

fuel-core/src/service/graph_api.rs
```diff
@@ -1,6 +1,8 @@
 use super::modules::Modules;
 use crate::database::Database;
 use crate::schema::{build_schema, dap, CoreSchema};
+#[cfg(feature = "prometheus")]
+use crate::service::metrics::metrics;
```
Voxelot, line 5: I think it might be better to leave the metrics route always configured, but use the feature flags to adjust the API handler response when disabled.

```suggestion
use crate::service::metrics::metrics;
```

ControlCplusControlV: Ok, will adjust `/metrics` to just return a blank body then if Prometheus is not enabled

fuel-core/src/database.rs
```diff
@@ -174,9 +183,21 @@ impl Database {
...
+        }
         self.data
             .get(key, column)?
-            .map(|val| bincode::deserialize(&val).map_err(|_| Error::Codec))
+            .map(|val| {
+                #[cfg(feature = "prometheus")]
+                {
+                    DATABASE_METRICS
+                        .bytes_read_meter
+                        .inc_by(bincode::serialized_size(&val).unwrap_or(0));
```
Voxelot, line 197: since `val` isn't deserialized yet, we could just get the length directly here:

```suggestion
                        .inc_by(val.len() as u64);
```

fuel-core/src/database.rs
```diff
@@ -148,6 +150,13 @@ impl Database {
         column: ColumnId,
         value: V,
     ) -> Result<Option<V>, Error> {
+        #[cfg(feature = "prometheus")]
+        {
+            DATABASE_METRICS.write_meter.inc();
+            DATABASE_METRICS
+                .bytes_written_meter
+                .inc_by(bincode::serialized_size(&value).unwrap_or(0));
```
Voxelot, line 158: would it be more efficient to first perform the serialization and assign to a tmp var, then check the length of the serialized data for metrics afterwards?

fuel-core/src/database.rs
ControlCplusControlV: Was looking for a away to do this, didn't realize I could just call length directly

fuel-core/src/database.rs
ControlCplusControlV: Yes that would be much better rather than doubling serialization times, just pushed a commit which addresses that

ControlCplusControlV:
Moving back to draft as tests now must specifically use rocksdb

Voxelot (approved):

fuel-core/src/state/rocks_db.rs
```diff
@@ -171,7 +193,16 @@ impl KeyValueStore for RocksDb {
...
             .db
             .iterator_cf_opt(&self.cf(column), opts, iter_mode)
-            .map(|(key, value)| (key.to_vec(), value.to_vec()));
+            .map(|(key, value)| {
+                #[cfg(feature = "prometheus")]
+                {
+                    DATABASE_METRICS.read_meter.inc();
+                    DATABASE_METRICS
+                        .bytes_read_meter
+                        .inc_by(value.to_vec().len() as u64);
```
Voxelot, line 202: 1. We should convert value to vec before logging the metrics so we don't need to double copy the bytes just to check the length.
2. the key byte length should also be included

ControlCplusControlV: Had to fix merge conflicts so corrected this anyway

Voxelot (approved):

fuel-core/src/state/rocks_db.rs
Voxelot: ok we should still file a tech debt to make sure the key length is also included in the read bytes.

We also need to instrument the batch_writes

ControlCplusControlV: Filed in #416

PR #405, opened 2022-06-13
Debugger API: Return new receipts when execution is paused

Dentosal:
Currently the GraphQL debugger API doesn't make receipts availble. This PR adds receipts to `RunResult` responses, so that they are communicated to the caller every time a breakpoint is hit and when the transaction terminates.

### DIFF
--- a/fuel-client/src/client/schema.rs
+++ b/fuel-client/src/client/schema.rs
@@ -6,7 +6,7 @@ pub mod schema {
 
 use hex::FromHexError;
 use std::array::TryFromSliceError;
-use std::fmt::Debug;
+use std::fmt::{self, Debug};
 use std::io::ErrorKind;
 use std::num::TryFromIntError;
 use thiserror::Error;
@@ -191,6 +191,25 @@ pub struct ContinueTx {
 #[cynic(schema_path = "./assets/schema.sdl")]
 pub struct RunResult {
     pub breakpoint: Option<OutputBreakpoint>,
+    pub json_receipts: Vec<String>,
+}
+
+impl RunResult {
+    pub fn receipts(&self) -> impl Iterator<Item = fuel_tx::Receipt> + '_ {
+        self.json_receipts.iter().map(|r| {
+            serde_json::from_str::<fuel_tx::Receipt>(r).expect("JSON deserialization failed")
+        })
+    }
+}
+
+impl fmt::Display for RunResult {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let receipts: Vec<fuel_tx::Receipt> = self.receipts().collect();
+        f.debug_struct("RunResult")
+            .field("breakpoint", &self.breakpoint)
+            .field("receipts", &receipts)
+            .finish()
+    }
 }
 
 #[derive(cynic::QueryFragment, Debug)]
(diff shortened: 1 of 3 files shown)


fuel-core/src/schema/dap.rs
```diff
@@ -423,9 +443,10 @@ mod gql_types {
         Breakpoint,
     }
 
-    #[derive(Debug, Clone, Copy, SimpleObject)]
+    #[derive(Debug, Clone, SimpleObject)]
     pub struct RunResult {
         pub state: RunState,
         pub breakpoint: Option<OutputBreakpoint>,
+        pub json_receipts: Vec<String>,
```
Voxelot, line 450: [nit] You could try using the graphql types instead of json serialization. Although we'd need to move the corresponding transparent_receipt in the fuel-gql-client out of the test module.

```suggestion
        pub json_receipts: Vec<crate::schema::tx::receipt::Receipt>,
```

Voxelot (approved):

fuel-client/src/client/schema.rs
```diff
@@ -191,6 +191,25 @@ pub struct ContinueTx {
 #[cynic(schema_path = "./assets/schema.sdl")]
 pub struct RunResult {
     pub breakpoint: Option<OutputBreakpoint>,
+    pub json_receipts: Vec<String>,
+}
+
+impl RunResult {
+    pub fn receipts(&self) -> impl Iterator<Item = fuel_tx::Receipt> + '_ {
+        self.json_receipts.iter().map(|r| {
+            serde_json::from_str::<fuel_tx::Receipt>(r).expect("JSON deserialization failed")
```
vlopes11, line 200: Since we will panic, maybe put something meaningful in the function name such as `unchecked`. Alternatively, or maybe better, we might just return a `Result`

fuel-core/src/schema/dap.rs
```diff
@@ -288,6 +288,12 @@ impl DapMutation {
             .transact(tx)
             .map_err(|err| async_graphql::Error::new(format!("Transaction failed: {err:?}")))?;
 
+        let json_receipts = state_ref
+            .receipts()
+            .iter()
+            .map(|r| serde_json::to_string(&r).expect("JSON serialization failed"))
```
vlopes11, line 294: Ditto; maybe don't overuse panic?

fuel-client/src/client/schema.rs
Voxelot: It might be better to find a way to perform all the deserialization up-front, rather than lazily when users call these getter functions. Since this type is setup for graphql compatibility, we may need to do the unspeakable and create a duplicate type that is more user friendly and automatically converts the gql response when created.

Voxelot:
@Dentosal any updates on this?

fuel-core/src/schema/dap.rs
Dentosal: I think this is good enough for now

Dentosal:
I've been pondering on how to proceed with this. Maybe the duplicate type would be the right solution, at least it would be more performant and give actual type-checking at parse/graphql layer? OTOH, this is code that's run on the debugger, so performance shouldn't matter much. The panics discussed above cannot occur unless the API changes, and if we duplicate the types, then we have to remember to update those anyway, so I'm not sure if that helps at all.

PR #407, opened 2022-06-13
txpool channels

rakita:
Add channels to txpool service


fuel-core-interfaces/src/txpool.rs
```diff
@@ -26,61 +26,66 @@ pub trait TxPoolDb:
...
-    async fn removed(&self, tx: Arc<Transaction>, error: &Error);
-}
-
-#[async_trait]
-pub trait TxPool: Send + Sync {
+#[derive(Debug)]
+pub enum TxPoolMpsc {
+    /// Return all sorted transactions that are includable in next block.
+    /// This is going to be heavy operation, use it with only when needed.
+    Includable {
```
Voxelot, line 33: [nit] should we create a tech debt item to explore parameterizing this response a bit? For example we could put an upper-bound on the number of txs to return.

fuel-txpool/src/service.rs
```diff
@@ -1,143 +1,137 @@
...
 
-    /// Return all sorted transactions that are includable in next block.
-    /// This is going to be heavy operation, use it with only when needed.
-    async fn includable(&self) -> Vec<ArcTx> {
-        let pool = self.txpool.read().await;
-        pool.sorted_includable()
+    pub async fn find_one(&self, id: TxId) -> oneshot::Receiver<Option<TxInfo>> {
+        let (response, receiver) = oneshot::channel();
+        let _ = self.sender.send(TxPoolMpsc::FindOne { id, response }).await;
+        receiver
```
Voxelot, line 92: We should await the oneshot here to clean up the interface (and not require consumers to use `.await.await`)

```suggestion
        receiver.await
```

fuel-txpool/src/service.rs
```diff
@@ -1,143 +1,137 @@
...
-    async fn filter_by_negative(&self, tx_ids: &[TxId]) -> Vec<TxId> {
-        let mut res = Vec::new();
-        let pool = self.txpool.read().await;
-        for tx_id in tx_ids {
-            if pool.txs().get(tx_id).is_none() {
-                res.push(*tx_id)
-            }
-        }
-        res
+    pub async fn find(&self, ids: Vec<TxId>) -> oneshot::Receiver<Vec<Option<TxInfo>>> {
```
Voxelot, line 83: consider awaiting the oneshot here to simplify the interface.

Voxelot, line 83: consider awaiting the oneshot here to simplify the interface.

```suggestion
    pub async fn find(&self, ids: Vec<TxId>) -> Vec<Option<TxInfo>> {
```

fuel-txpool/src/service.rs
```diff
@@ -1,143 +1,137 @@
...
+    pub async fn find_one(&self, id: TxId) -> oneshot::Receiver<Option<TxInfo>> {
+        let (response, receiver) = oneshot::channel();
+        let _ = self.sender.send(TxPoolMpsc::FindOne { id, response }).await;
+        receiver
     }
 
-    /// When block is updated we need to receive all spend outputs and remove them from txpool
-    async fn block_update(&self /*spend_outputs: [Input], added_outputs: [AddedOutputs]*/) {
-        self.txpool.write().await.block_update()
+    pub async fn find_dependent(&self, ids: Vec<TxId>) -> oneshot::Receiver<Vec<Arc<Transaction>>> {
```
Voxelot, line 95: same feedback as here: https://github.com/FuelLabs/fuel-core/pull/407/files#r896344625

fuel-txpool/src/service.rs
```diff
@@ -1,143 +1,137 @@
...
-    async fn remove(&self, tx_ids: &[TxId]) {
-        let mut removed = Vec::new();
-        for tx_id in tx_ids {
-            let rem = { self.txpool.write().await.remove_by_tx_id(tx_id) };
-            removed.extend(rem.into_iter());
-        }
-        for removed in removed {
-            self.subs.removed(removed, &Error::Removed).await
-        }
+    pub async fn filter_by_negative(&self, ids: Vec<TxId>) -> oneshot::Receiver<Vec<TxId>> {
```
Voxelot, line 104: same feedback as here: https://github.com/FuelLabs/fuel-core/pull/407/files#r896344625

fuel-txpool/src/service.rs
```diff
@@ -1,143 +1,137 @@
...
+        let _ = self
+            .sender
+            .send(TxPoolMpsc::FilterByNegative { ids, response })
+            .await;
+        receiver
     }
 
-    async fn subscribe(&self, sub: Arc<dyn Subscriber>) {
-        self.subs.sub(sub);
+    pub async fn includable(&self) -> oneshot::Receiver<Vec<Arc<Transaction>>> {
```
Voxelot, line 113: same feedback as here: https://github.com/FuelLabs/fuel-core/pull/407/files#r896344625

fuel-txpool/src/service.rs
```diff
@@ -1,143 +1,137 @@
...
-        list.sort_by_key(|tx| Reverse(tx.gas_price()));
+    pub fn sender(&self) -> &mpsc::Sender<TxPoolMpsc> {
+        &self.sender
+    }
 
-        list
+    pub async fn insert(
+        &self,
+        txs: Vec<Arc<Transaction>>,
+    ) -> oneshot::Receiver<Vec<anyhow::Result<Vec<Arc<Transaction>>>>> {
```
Voxelot, line 77: same feedback as here https://github.com/FuelLabs/fuel-core/pull/407/files#r896344625

fuel-txpool/src/service.rs
```diff
@@ -1,143 +1,137 @@
...
-    pub fn new(db: Box<dyn TxPoolDb>, config: Config) -> Self {
-        Self {
-            txpool: RwLock::new(TxPoolImpl::new(config)),
-            db,
-            subs: MultiSubscriber::default(),
-        }
+impl Service {
+    pub fn new(db: Box<dyn TxPoolDb>, config: Config) -> Result<Self, anyhow::Error> {
+        let (sender, receiver) = mpsc::channel(100);
+        let (broadcast, _receiver) = broadcast::channel(100);
```
Voxelot, line 22: if we want the channels to be recyclable (due to service restart etc), maybe we should consolidate the initialization of all of them into the main fuel-core service?

fuel-txpool/src/interface.rs
```diff
@@ -0,0 +1,193 @@
...
+use std::sync::Arc;
+use tokio::sync::{broadcast, mpsc, RwLock};
+
+pub struct Interface {
+    txpool: RwLock<TxPoolImpl>,
+    broadcast: broadcast::Sender<TxStatusBroadcast>,
+    db: Box<dyn TxPoolDb>,
+}
+
+impl Interface {
```
Voxelot, line 16: This isn't really an interface? Seems like it's all impl with no traits.

fuel-txpool/src/service.rs
rakita: It is the same problem, if I want to receive from `Receiver` i need a `mut` version of it, I could wrap it inside Mutex but this is similar to what I am doing now

fuel-core-interfaces/src/txpool.rs
rakita: In most cases, you would like to fetch all transactions from txpool as you don't know how efficient you can pack the new block as it depends on gas that is spent by transaction. There could possible be some consensus wide limit calculated as `block_gas_limit` / `  system minimum of transaction gas spend` but those things are like a small detail.

fuel-txpool/src/interface.rs
rakita: It acts as an interface between service and txpool, but maybe proxy is a better word

fuel-txpool/src/service.rs
```diff
@@ -168,19 +156,40 @@ pub mod tests {
...
+        let out = receiver.await.unwrap();
         assert_eq!(out.len(), 2, "Shoud be len 2:{:?}", out);
         assert!(out[0].is_some(), "Tx1 should be some:{:?}", out);
         let id = out[0].as_ref().unwrap().id();
         assert_eq!(id, tx1_hash, "Found tx id match{:?}", out);
         assert!(out[1].is_none(), "Tx3 should not be found:{:?}", out);
+        service.stop().await.unwrap().await.unwrap();
     }
 
+    /*
```
ControlCplusControlV, line 192: Is this test blocked from working by something or no longer needed? I saw `Subscriber` was removed so if it's no longer needed I would suggest removing the test. If it is blocked by something could you make issue to track this for when it should be reinstated?

fuel-core-interfaces/src/txpool.rs
```diff
@@ -26,61 +26,66 @@ pub trait TxPoolDb:
...
+    /// stop txpool
+    Stop,
+}
 
-    /// remove transaction from pool needed on user demand. Low priority
-    async fn remove(&self, hashes: &[TxId]);
+#[derive(Clone, Debug)]
+pub enum TxStatus {
+    /// Included into txpool.
+    Included,
```
ControlCplusControlV, line 78: Included could be confusing terminology as Included is often used to say a transaction has been Included in a Block. You included a doc comment though so should be fine, just wanted to mention as it was my initial thought

fuel-txpool/src/service.rs
```diff
@@ -1,143 +1,108 @@
...
-
-pub struct TxPoolService {
-    txpool: RwLock<TxPoolImpl>,
-    db: Box<dyn TxPoolDb>,
-    subs: MultiSubscriber,
+use tokio::sync::{broadcast, mpsc, Mutex};
+use tokio::task::JoinHandle;
+use tracing::warn;
+
+pub struct Service {
```
ControlCplusControlV, line 9: Is there a reason the name `TxPoolService` was changed? I could see `Service` leading to conflicts in the future as new parts are introduced, something more distinct could avoid needing alias it on imports

ControlCplusControlV (changes requested):

fuel-txpool/src/service.rs
rakita: https://github.com/FuelLabs/fuel-core/pull/407#discussion_r900116952

fuel-txpool/src/service.rs
rakita: https://github.com/FuelLabs/fuel-core/pull/407#discussion_r900116952

fuel-txpool/src/service.rs
rakita: https://github.com/FuelLabs/fuel-core/pull/407#discussion_r900116952

fuel-txpool/src/service.rs
rakita: All modules have the same naming, all of them now have `Service` and `Config` structures inside

fuel-txpool/src/service.rs
rakita: I left it as maybe code that can be reused, the `subscription` is now done with `tokio::broadcast`. I will remove the comment.

fuel-txpool/src/service.rs
rakita: https://github.com/FuelLabs/fuel-core/pull/407#discussion_r900116952

ControlCplusControlV (dismissed):
LGTM

fuel-core-interfaces/src/txpool.rs
Voxelot: Agreed - this will come up later if needed

fuel-txpool/src/service.rs
Voxelot: Ultimately it's the same behavior for the client/caller either way right? They should still use timeouts on potentially long-running async-await methods if needed, I'm having a hard time imagining how I would handle the oneshot result differently than any other regular future.

fuel-txpool/src/service.rs
Voxelot: I feel like it's a smell to have commented out tests without any explanation in the code. If we want this code as a reference for later it should be put in a doc-test or something else to avoid bit-rot.

Voxelot (dismissed):
This looks correct based on the events proposal. Although I wonder if we're really gaining much value from converting async traits into oneshot channels. I could see this being useful if we embraced an actor-style fault-tolerant architecture, but right now it seems like they add a lot of extra boilerplate while increasing the overhead of every call.

Approving since this is the work we set out to do and it looks correct. But I'm wondering if others on the team feel the same way about all this oneshot channel management.

fuel-txpool/src/interface.rs
leviathanbeak: yeah, the name is too generic, I'd prefer having some prefix/sufix with it so that it's clear from the name to what exactly it is an interface/proxy

fuel-txpool/src/interface.rs
```diff
@@ -0,0 +1,192 @@
...
+    pub async fn run(
+        interface: Arc<Interface>,
+        mut new_block: broadcast::Receiver<ImportBlockBroadcast>,
+        mut receiver: mpsc::Receiver<TxPoolMpsc>,
+    ) -> mpsc::Receiver<TxPoolMpsc> {
+        let mut running = true;
+        while running {
```
leviathanbeak, line 24: why not just `break` out of the loop instead of having the `running` flag ?

fuel-txpool/src/interface.rs
```diff
@@ -0,0 +1,192 @@
...
+
+/// Acts an interface between Service and Txpool
+pub struct Interface {
+    txpool: RwLock<TxPoolImpl>,
+    broadcast: broadcast::Sender<TxStatusBroadcast>,
+    db: Box<dyn TxPoolDb>,
+}
+
+impl Interface {
+    pub async fn run(
```
leviathanbeak, line 18: Would it be better that `run_interface()` is either a standalone function that takes `interface: Arc<Interface>` or `Interface` method that takes `self: Arc<Self>`?
Currently it's an associated function with a bit confusing API: `Interface::run(Arc::new(Interface::new()))`

fuel-txpool/src/interface.rs
```diff
@@ -0,0 +1,192 @@
...
+        for tx_id in tx_ids {
+            if pool.txs().get(tx_id).is_none() {
+                res.push(*tx_id)
+            }
+        }
+        res
+    }
+
+    /// Return all sorted transactions that are includable in next block.
+    /// This is going to be heavy operation, use it with only when needed.
```
leviathanbeak, line 165: nit: "use it only when needed"

fuel-core/src/schema/tx.rs
```diff
@@ -266,20 +274,72 @@ impl TxMutation {
...
     }
 
     /// Submits transaction to the txpool
     async fn submit(&self, ctx: &Context<'_>, tx: HexString) -> async_graphql::Result<Transaction> {
-        let tx_pool = ctx.data::<Arc<TxPool>>().unwrap();
-        let tx = FuelTx::from_bytes(&tx.0)?;
-        tx_pool.submit_tx(tx.clone()).await?;
-        let tx = Transaction(tx);
+        let db = ctx.data_unchecked::<Database>();
+        let txpool = ctx.data::<Arc<TxPoolService>>().unwrap();
```
leviathanbeak, line 296: why not the same approach: `let txpool = ctx.data_unchecked::<Arc<TxPoolService>>();` ?

fuel-txpool/src/interface.rs
rakita: I agree, left it as run but made first input argument as `self: Arc<Self>`

fuel-txpool/src/service.rs
rakita: test was rewriten

fuel-core-interfaces/src/txpool.rs
```diff
@@ -26,61 +26,66 @@ pub trait TxPoolDb:
...
+    Insert {
+        txs: Vec<Arc<Transaction>>,
+        response: oneshot::Sender<Vec<anyhow::Result<Vec<Arc<Transaction>>>>>,
+    },
     /// find all tx by their hash
-    async fn find(&self, hashes: &[TxId]) -> Vec<Option<TxInfo>>;
-
+    Find {
+        ids: Vec<TxId>,
+        response: oneshot::Sender<Vec<Option<TxInfo>>>,
```
vlopes11, line 49: Does vec of option make sense here? If the tx isn't found, then the vec should be empty

fuel-core-interfaces/src/txpool.rs
```diff
@@ -26,61 +26,66 @@ pub trait TxPoolDb:
...
+    },
     /// find all dependent tx and return them with requsted dependencies in one list sorted by Price.
-    async fn find_dependent(&self, hashes: &[TxId]) -> Vec<Arc<Transaction>>;
-
+    FindDependent {
+        ids: Vec<TxId>,
+        response: oneshot::Sender<Vec<Arc<Transaction>>>,
+    },
+    /// remove transaction from pool needed on user demand. Low priority
+    Remove { ids: Vec<TxId> },
```
vlopes11, line 62: Shouldn't we separate debug/helpers from the core functionality?

fuel-core/src/executor.rs
```diff
@@ -44,6 +48,32 @@ pub struct Executor {
...
+        let db = self.database.clone();
+
+        for tx in txs.iter() {
+            // set status to submitted
+            db.update_tx_status(&tx.id(), TransactionStatus::Submitted { time: Utc::now() })?;
+        }
+
+        // setup and execute block
+        let current_height = db.get_block_height()?.unwrap_or_default();
+        let current_hash = db.get_block_id(current_height)?.unwrap_or_default();
```
vlopes11, line 61: unwraping these might be source of inconsistencies. Maybe fail the function instead? both height and block id are critical for the correctness of the system

fuel-core-interfaces/src/txpool.rs
```diff
@@ -26,61 +26,66 @@ pub trait TxPoolDb:
...
-
-    async fn inserted_on_block_revert(&self, tx: Arc<Transaction>);
-
-    async fn removed(&self, tx: Arc<Transaction>, error: &Error);
-}
-
-#[async_trait]
-pub trait TxPool: Send + Sync {
+#[derive(Debug)]
+pub enum TxPoolMpsc {
```
vlopes11, line 30: Isn't it over-engineering converting functionalities that can trivially be functions into async channels? If we have many enum variants for the things that should actually be async, it might be confusing.

fuel-core-interfaces/src/txpool.rs
```diff
@@ -26,61 +26,66 @@ pub trait TxPoolDb:
...
+    /// Return all sorted transactions that are includable in next block.
```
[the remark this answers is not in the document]
rakita: They are sorted in the only way that pool sorts transactions, by priority, or in the current case by gas price. It needs to be sorted as block_producer needs to know what transaction to include first in the block.

fuel-core-interfaces/src/txpool.rs
rakita: Find is for multiple transactions and for every transaction result is returned.

fuel-core-interfaces/src/txpool.rs
rakita: `Remove` is under core functionality, as having the ability to remove transactions from txpool. blacklisting transaction from pool is something that we can think about in future.

fuel-core-interfaces/src/txpool.rs
rakita: having a trait or channel does not in essence change a lot in terms of the interface, that is why firstly I implemented this as `trait TxPool` and now I am changing it to channels as it seems that it will give us a little bit better handling for requests in more concurrent mode.
Sender channel can emulate ordinary function call with a wrapped structure, so the channel is hidden behind it: https://github.com/FuelLabs/fuel-core/compare/rakita/txpool_channels...rakita/txpool_sender

fuel-core/src/executor.rs
Voxelot: Since the return type is `Result<Option<T>>`, errors are properly handled here via `?`, while `unwrap_or_default` is used on the option to initialize the block number if it hasn't been set in the db yet.

fuel-txpool/src/service.rs
rakita: sometimes we want to wait for a return, sometimes we don't care. TxPool return of `include` from the user side matters as it gives a reason for not including, TxPool return of `include` from p2p side can be dropped. Having more options is not a bad thing and it is addressed here: https://github.com/FuelLabs/fuel-core/compare/rakita/txpool_channels...rakita/txpool_sender
Where you can use both direct call and with `DerefMut` use `channel` `send` if needed.

fuel-txpool/src/service.rs
rakita: You are now talking about full restart of fuel-client? The better way is to kill the app and start it again as it will be the most usual and most tested way how you start the client.

fuel-txpool/src/interface.rs
rakita: It is internally used and not exposed.

rakita:
@adlerjohn you are only one left that didn't comment on this pull request :)

adlerjohn (commented):
Looks fine to me, no immediate comments. Will defer to @Voxelot and @vlopes11 for approval.

adlerjohn (commented):
merge conflicts!

rakita:
This conflict is from last night: https://github.com/FuelLabs/fuel-core/commit/4c1f9d287f9806e33cb96c313de39e152aa172b7
With this pull request legacy txpool is removed and execution is moved to `Executor`

Voxelot (approved):

PR #409, opened 2022-06-14
Standalone  Multi-TX Block Producer

Voxelot:
Initial base for the block producer service *WIP*

Leveraging ports & adapters architecture to unblock development from depending on other components & also to improve composability & testability. 

The main principle behind the P&A design is to allow the component to specify all external dependencies via its' own traits which don't need to be depended on by others. Fuel Core can provide adapter impls at a later time when feasible, while allowing the block producer to be logically implemented and functionally tested in the meantime as a self-contained unit.

I'm also experimenting with the service design a bit, it seems like treating the service as a dummy wrapper for spawning the stateful block producer task makes the borrow checker happier and avoids the need for mutexes and other concurrency primitives.


rakita:
What is intention behind this change?

Voxelot:
Most of this PR is just additional logic for the block producer. The P&A concepts started introducing as I went along because I ran into blockers with directly integrating with the relayer and txpool since they aren't done yet. Then I realized the ports pattern could be used to start developing the other modules in parallel, so that we don't need to have to serialize the development of these modules based on their order of dependencies. I'd consider it as a complementary to the event-driven arch, which makes the core domain logic of the modules more pure, testable and easier to follow.

rakita:
Why i asked that directly for intention is that I was not sure if you looked at how it is now, and for example how `fuel-relayer` did this, as it does rely on `database` and `fuel-importer` and can be standalone tested: https://github.com/FuelLabs/fuel-core/blob/f549a3f315380202dec60893a14bd7c9990c27e0/fuel-relayer/src/service.rs#L26-L27 
For relayer channels, I need to move them from `new` function but the idea is there.

same for txpool: https://github.com/FuelLabs/fuel-core/blob/887f9443d9a362cb150967c1953d82b68c515ea1/fuel-txpool/src/service.rs#L31

Abstraction is done over channel message enums.

Maybe there is a need for having `new`, `init` and `start` functions. `new` to create, `init` to bind channels, and `start/stop` to stop and start service.

Dentosal:
As discussed with @Voxelot earlier today, this PR now contains the standalone block producer module with minimal implementation. It will not be used elsewhere in the code yet, but to keep PRs small enough to review, that integration will be done later.

Follow-up PRs will include:
* Integrating the block producer into Fuel node service
* Configurable block production trigger https://github.com/FuelLabs/fuel-core/issues/50
* A minimal block importer, see also https://github.com/FuelLabs/fuel-core/issues/465
* Proper e2e testing

fuel-block-producer/src/block_producer/transaction_selector.rs
```diff
@@ -0,0 +1,172 @@
...
+                }
+            } else {
+                false
+            }
+        })
+        .collect()
+}
+
+#[cfg(test)]
+mod tests {
```
Voxelot, line 42: I'd rather see this kind of integration testing after the block producer is integrated into fuel-core, and for the tests to co-exist with the other integ tests in the standard `fuel-tests` package.

Keeping the dev-dependencies light in these modules allows for rapid iteration & testing of logical changes in a localized way. This is one of the key benefits of a P&A approach, as these "domain modules" can be iterated on and be well-tested independently of each other.

I'm ok with keeping this here temporarily until we finish the integration with fuel-core, but it should be relocated & refactored as part of the integration task.

fuel-core-interfaces/src/db.rs
```diff
@@ -60,3 +60,1118 @@ impl From<KvStoreError> for InterpreterError {
         InterpreterError::Io(std::io::Error::new(std::io::ErrorKind::Other, e))
     }
 }
+
+#[cfg(any(test, feature = "test-helpers"))]
```
Voxelot, line 64: this section should be removed, as this test data should no longer be used.

fuel-core-interfaces/src/model/block.rs
```diff
@@ -74,14 +84,25 @@ impl FuelBlockHeader {
...
+
+    pub fn transactions_root(txs: &[Transaction]) -> Bytes32 {
+        let mut storage = StorageMap::new();
+        let mut tree = MerkleTree::new(&mut storage);
+        for tx in txs {
+            // serialize tx into canonical format for hashing
+            let ser_tx = tx.clone().to_bytes();
+            tree.push(&ser_tx).expect("infallible storage");
+        }
+        tree.root().expect("infallible storage").into()
```
bvrooman, line 96: You can instead use the in-memory version of the Merkle tree to provide in-memory, infallible storage for calculating the transactions root. This would look something like:

```
use fuel_merkle::binary::in_memory::MerkleTree;
...

let merkle_tree = MerkleTree::new();
for tx in txs {
    // serialize tx into canonical format for hashing
    let ser_tx = tx.clone().to_bytes();
    tree.push(&ser_tx);
}
tree.root().into()
```

bvrooman:
It appears that we are reintroducing the concept of a prefilled test database in the form of `DummyDb` in fuel-core-interfaces/src/db/helpers.rs. 

We removed all instances of `DummyDb` in unit tests and replaced it with an empty database that unit tests could instantiate and fill themselves. 

Is there a reason we want to reintroduce the `DummyDb` struct and all its prefilled data?

Dentosal:
> Is there a reason we want to reintroduce the DummyDb struct and all its prefilled data?

No, the `helpers` file just slipped by me when merging from master. I'll remove it.

fuel-txpool/src/txpool.rs
```diff
@@ -239,10 +241,16 @@ impl TxPool {
...
     /// When block is updated we need to receive all spend outputs and remove them from txpool.
     pub async fn block_update(
-        txpool: &RwLock<Self>, // spend_outputs: [Input], added_outputs: [AddedOutputs]
+        txpool: &RwLock<Self>,
+        block: Arc<FuelBlock>,
+        // spend_outputs: [Input], added_outputs: [AddedOutputs]
     ) {
-        let _ = txpool.write().await;
+        let mut guard = txpool.write().await;
         // TODO https://github.com/FuelLabs/fuel-core/issues/465
```
ControlCplusControlV, line 249: Should this still be TODO if this PR resolves that issue?

Dentosal: The issue is only partially resolved here, and the TODO should indeed remain.

Voxelot:
Just pushed a bunch of changes:

- No longer using the mpsc channels, using a simpler async trait approach instead.
- Cleaned up the integ test to no longer require fuel-core by using a mockdb for the txpool
- No longer use a trailing da_height. This just slows down the deposit process and creates a bunch of extra corner cases when selecting transaction from the txpool that contain messages. Instead, the block importer should queue blocks until the local relayer finalized da_height catches up to the block da_height.
- unit test!

fuel-block-producer/src/block_producer.rs
```diff
@@ -0,0 +1,194 @@
...
+        Ok(selected_txs)
+    }
+
+    async fn previous_block_info(
+        &self,
+        height: BlockHeight,
+    ) -> Result<PreviousBlockInfo> {
+        // block 0 is reserved for genesis
+        if height == 0u32.into() {
+            return Err(GenesisBlock.into())
```
ControlCplusControlV, line 159:
```suggestion
            Err(GenesisBlock.into())
```

ControlCplusControlV: Fixes clippy issue

Dentosal (commented):
LGTM, but as I've written a large chunk of this, someone else should review it as well.

fuel-block-producer/src/adapters.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub consensus_params: ConsensusParameters,
+}
+
+#[async_trait::async_trait]
+impl TxPool for TxPoolAdapter {
+    async fn get_includable_txs(
+        &self,
+        block_height: BlockHeight,
+    ) -> anyhow::Result<Vec<Arc<CheckedTransaction>>> {
+        let includable_txs = self.sender.includable().await?;
```
xgreenx, line 23: Maybe for production better to extract all transactions and move them to the block producer? BP will return them to the TX pool if the block can't be created.

Why do we want to duplicate transactions in the TX pool? Only for the GraphQL cases?

fuel-block-producer/src/block_producer.rs
```diff
@@ -0,0 +1,194 @@
...
+
+pub struct Producer<'a> {
+    pub config: Config,
+    pub db: &'a dyn BlockProducerDatabase,
+    pub txpool: &'a dyn TxPool,
+    pub executor: &'a dyn Executor,
+    pub relayer: &'a dyn Relayer,
+}
+
+#[async_trait::async_trait]
```
xgreenx, line 58: If we remove all unjustified async methods in dependencies, we can also remove it here.

fuel-block-producer/src/block_producer.rs
```diff
@@ -0,0 +1,194 @@
...
+    async fn produce_block(&self, height: BlockHeight) -> Result<FuelBlock> {
+        //  - get previous block info (hash, root, etc)
+        //  - select best da_height from relayer
+        //  - get available txs from txpool
+        //  - select best txs based on factors like:
+        //      1. fees
+        //      2. parallel throughput
+        //  - Execute block with production mode to correctly malleate txs outputs and block headers
+
+        let previous_block_info = self.previous_block_info(height).await?;
```
xgreenx, line 70: Seems `previous_block_info` is sync method

fuel-block-producer/src/block_producer.rs
```diff
@@ -0,0 +1,194 @@
...
+#[cfg(test)]
+mod tests;
+mod transaction_selector;
+
+pub struct Producer<'a> {
+    pub config: Config,
+    pub db: &'a dyn BlockProducerDatabase,
+    pub txpool: &'a dyn TxPool,
+    pub executor: &'a dyn Executor,
+    pub relayer: &'a dyn Relayer,
```
xgreenx, line 55: Do we need to put it under `feature = "relayer"`?

fuel-block-producer/src/block_producer.rs
```diff
@@ -0,0 +1,194 @@
...
+            );
+        }
+
+        let _ = result.context(format!(
+            "Failed to produce block {:?} due to execution failure",
+            block
+        ))?;
+
+        debug!("Produced block: {:?}", &block);
+        Ok(block)
```
xgreenx, line 122: When do we plan to sign it? Seems BlockProducer should have `Signer` and sign it

fuel-block-producer/src/block_producer/tests.rs
```diff
@@ -0,0 +1,231 @@
...
+}
+
+#[tokio::test]
+async fn can_produce_initial_block() {
+    // simple happy path for producing the first block
+    let ctx = TestContext::default();
+    let producer = ctx.producer();
+
+    let err = producer
+        .produce_block(0u32.into())
```
xgreenx, line 57: I guess it should be `1`? And test should be positive?

fuel-block-producer/src/block_producer/transaction_selector.rs
```diff
@@ -0,0 +1,180 @@
...
+            TxGas { price: 3, limit: 2000 },
+            TxGas { price: 1, limit: 1000 },
+            TxGas { price: 4, limit: 3000 },
+            TxGas { price: 5, limit: 1000 },
+            TxGas { price: 2, limit: 1000 },
+        ];
+
+        for k in 0..original.len() {
+            for perm in original.into_iter().permutations(k) {
+                for gas_limit in [999, 1000, 2000, 2500, 3000, 5000, 6000, 10_000] {
```
xgreenx, line 172: Maybe checking the number of transactions also is a good idea.
`[(999, 0), (1000, 1), (2000, 2), (2500, 2), (3000, 1), (5000, 3), (6000, 3), (10_000, 5)]`

fuel-block-producer/src/config.rs
```diff
@@ -1,2 +1,25 @@
...
+        Address,
+        Word,
+    },
+};
+
+#[derive(Clone, Debug)]
+pub struct Config {
+    /// The primary identity associated with the validator (i.e. staking key)
+    pub validator_id: Address,
+    pub max_gas_per_block: Word,
```
xgreenx, line 13: Maybe it should be part of the consensus?

fuel-block-producer/src/ports.rs
```diff
@@ -0,0 +1,25 @@
...
+        BlockHeight,
+        DaBlockHeight,
+    },
+};
+use std::sync::Arc;
+
+#[async_trait]
+pub trait Relayer: Sync + Send {
+    /// Get the best finalized height from the DA layer
+    async fn get_best_finalized_da_height(&self) -> Result<DaBlockHeight>;
```
xgreenx, line 15: I guess it also should be sync=)

fuel-block-producer/src/adapters.rs
```diff
@@ -0,0 +1,38 @@
...
+    txpool::Sender,
+};
+use std::sync::Arc;
+
+pub struct TxPoolAdapter {
+    pub sender: Sender,
+    pub consensus_params: ConsensusParameters,
+}
+
+#[async_trait::async_trait]
```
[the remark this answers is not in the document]
Voxelot: I had a similar conversation with @freesig about our overuse of async. However, I'm not totally convinced block production shouldn't be async yet. Given that block production has to also call `execute` on the block which may take a few seconds to process, do you think it makes sense to allow the caller of produce to yield while we do the actual execution on a worker pool (like rayon)?

fuel-block-producer/src/adapters.rs
Voxelot: Maybe it would be simpler to only deal with the removal of txs from the block importer/committer, that way txs are only removed once the block is actually applied? Since in PoS mode, we might not know right away whether the block is good and then we need come up with a new way to re-include the txs from a failed round.

Less copying is generally a good thing, but we'd also have to factor in the cost of re-insertion of txs that didn't fit in the block. Let's say our txpool is holding more than 100 blocks worth of txs, this repeated txpool dumping and reinsertion may be higher than just copying the txs we selected for the block.

fuel-block-producer/src/block_producer.rs
Voxelot: I think it would be simpler for this module if fuel-core just passed in a dummy relayer implementation when the feature is disabled. Since it's behind a trait, the behavior is already configurable.

fuel-block-producer/src/block_producer/tests.rs
Voxelot: oops copy paste bug

fuel-block-producer/src/config.rs
Voxelot: Ah good point, this totally seems like something that should be part of consensus params.

I'll make a PR to fuel-tx for this..

fuel-block-producer/src/ports.rs
Voxelot: Yeah this could be sync, since we making a bunch of changes to the relayer right now anyways I'll just make this change since we won't be stuck with using a `block_on`.

fuel-block-producer/src/config.rs
Voxelot: As we discussed on slack, we should bundle consensus parameters with the appropriate domain and then aggregate all the various domain specific settings in fuel-core. This will prevent fields being added to fuel-tx that are unused by fuel-tx or the VM, and avoid causing breaking changes in fuel tx when settings unrelated to tx validation are updated.

fuel-block-producer/src/block_producer.rs
Voxelot: fixed

fuel-block-producer/src/block_producer/tests.rs
Voxelot: fixed

fuel-block-producer/src/ports.rs
Voxelot: fixed

fuel-block-producer/src/block_producer.rs
Voxelot: fixed by removing producer identity from the config. This will be deferred to the consensus module.

fuel-block-producer/src/adapters.rs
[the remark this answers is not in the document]
xgreenx: I agree that for `BlockExecuter::execute` we want `async`, because, with the metadata, we can parallelize execution.

But the block **production** should be synchronous because it is our source of metadata. We can't parallelize it, so it will consume the thread's resources entirely. We also can't start the next block production while the previous is not finished.

It is one of two heavies operations(the second is block commit). And so it should spawn a separate thread that fully focuses on it.

fuel-block-producer/src/adapters.rs
xgreenx: We can move the selection strategy into the tx pool. In that case, we can call `txpool.includable(MAX_GAS)`, it will return only transactions that cumulative gas is less than `MAX_GAS`.

TXs can be sorted in the `TxPool` by the gas, and selecting the last `X` transactions is easy. Insertion into this vector will be for  `log(n)` with a binary search approach.

Dentosal: That might be a good idea. Maybe the txpool could even store "next block tx set" separately to make it even faster. I'm not sure this optimization is necessary though.

Voxelot: This does make more sense than using this heavy "fetch all potentially incluable txs and then filter them" design IMO.

> TXs can be sorted in the TxPool by the gas, and selecting the last X transactions is easy. 

If you take a look at the current txpool impl you'll see this is already done :)

I'll update the txpool port for block producer to provide a max_gas to the includable API. For now can move the selection logic into the TxPool adapter, and then make a separate PR to update the txpool API and remove the need for the adapter to do any advanced logic.

fuel-block-producer/src/adapters.rs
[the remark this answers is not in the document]
Voxelot: One risk of making this not async, is that to perform block production it will likely need to be spawned onto the blocking pool. The same blocking pool as the executor would be running on. This busy waiting on the spawned block execution task is an inefficient use of threads and is exactly the sort of thing async is designed to avoid.

> We also can't start the next block production while the previous is not finished.

This is also a valid concern, but could also be dealt with by adding a semaphore to this method and leaving it async.

fuel-block-producer/src/adapters.rs
Voxelot: moved transaction selection to the adapter and updated the local txpool interface to include max_gas.

fuel-block-producer/src/block_producer/transaction_selector.rs
Voxelot: I think this one is less important, since the number of txs doesn't really matter as much as verifying we can never select more than the allowed amount of gas. It's also not that straightforward to configure the expected number of txs since we are performing these selections across all permutations, which in some cases means we have a set size of zero to select from for example.

Voxelot: I'll look into adding the selection correctness to the preceding test: `selector_prefers_highest_gas_txs_and_sorts`

Voxelot: Making these tests revealed a flaw in the transaction selection which I've now addressed!

The selection was using `tx.max_fee()` to calculate the block capacity, when it should've been using only gas instead (block capacity isn't dependent on gas price). This required a small change to fuel-tx, and so the patch section is just temporary until https://github.com/FuelLabs/fuel-tx/pull/189 gets merged.

fuel-block-producer/src/adapters.rs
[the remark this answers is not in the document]
Voxelot: Added an async-friendly semaphore to prevent concurrent block production requests.

Voxelot:
blocked by: https://github.com/FuelLabs/fuel-tx/pull/189

Dentosal:
Blocker cleared. Now this has merge conflicts.

fuel-block-producer/src/adapters.rs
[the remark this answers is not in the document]
Voxelot: > so it should spawn a separate thread that fully focuses on it.

If this thread spends most of its time dedicated to being blocked on the executor threadpool, why not just make it an async task on a shared runtime?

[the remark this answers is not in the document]
Voxelot: I created an issue to defer this discussion to a future discovery spike: https://github.com/FuelLabs/fuel-core/issues/642

Let's just keep this async for now.

xgreenx (approved):
Look good!=) It is hard to read the integration test, but maybe in the future, we will find how to make it more readable=D

Voxelot:
> Look good!=) It is hard to read the integration test, but maybe in the future, we will find how to make it more readable=D

I think once the followup tasks are implemented this test should look a lot simpler and not have to simulate as much of the node behavior.

PR #411, opened 2022-06-15
Dependency consolidation using fuel-core-interfaces

Voxelot:
To make version bumps less painful, all external fuel-* deps (vm, asm, crypto, merkle, storage, ...) are centrally managed by `fuel-core-interfaces` which re-exports them for other modules. This will help avoid issues with missed version bumps, and also generally make version bumping in fuel-core less painful.

The only crate these re-exports won't affect is fuel-gql-client, as it doesn't use fuel-core-interfaces (nor should it, as the interfaces are for fuel-core internal module integrations).

If we wanted to fully dedupe these imports, we'd need to make a new workspace crate that simply imports all the common deps and re-exports them back out. This could then be used by fuel-core-interfaces and fuel-gql-client. I opted not to do this for now, since bumping in 2 places is still better than 12 and didn't want the overhead of another published crate just yet.

### DIFF
--- a/fuel-core-interfaces/Cargo.toml
+++ b/fuel-core-interfaces/Cargo.toml
@@ -13,10 +13,11 @@ description = "Fuel core interfaces"
 [dependencies]
 anyhow = "1.0"
 async-trait = "0.1"
-chrono = { version = "0.4", features = ["serde"] }
+chrono = { version = "0.4" }
 derive_more = { version = "0.99" }
 fuel-asm = "0.5"
 fuel-crypto = { version = "0.5", default-features = false }
+fuel-merkle = { version = "0.2" }
 fuel-storage = "0.1"
 fuel-tx = { version = "0.12", default-features = false }
 fuel-types = { version = "0.5", default-features = false }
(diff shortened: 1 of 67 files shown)


fuel-core-interfaces/Cargo.toml
```diff
@@ -13,10 +13,11 @@ description = "Fuel core interfaces"
 [dependencies]
 anyhow = "1.0"
 async-trait = "0.1"
-chrono = { version = "0.4", features = ["serde"] }
+chrono = { version = "0.4" }
```
leviathanbeak, line 16: why was "serde" feature removed? I think it was necessary for "FuelBlockHeader" to be able to be correctly serialize `pub time: DateTime<Utc>`

Voxelot: It wasn't removed, I moved it to the `serde` feature below so that we don't pull in serde when default features are disabled.

leviathanbeak: ah good point

leviathanbeak (approved):
lgtm

PR #412, opened 2022-06-15
add gas price factor

Voxelot:
Update to the latest VM to make use of `gas_price_factor`. Also, make the necessary changes to support configuring this param from the chainspec.

ref: https://github.com/FuelLabs/fuel-specs/pull/343


rakita:
It needs to be rebased

vlopes11 (approved):
LGTM!

PR #421, opened 2022-06-16
ValidatorAddress and ConsensusPublicKey types

rakita:
Semantic change and introduction of `ValidatorAddress` and `ConsensusPublicKey` types.
Discussion started here: https://github.com/FuelLabs/fuel-core/pull/377/files#r896284757

### DIFF
--- a/fuel-core-interfaces/src/db.rs
+++ b/fuel-core-interfaces/src/db.rs
@@ -145,7 +145,7 @@ pub mod helpers {
 
     #[derive(Clone, Debug)]
     pub struct Data {
-        /// Used for Storage<Address, (u64, Option<Address>)>
+        /// Used for Storage<ValidatorAddress, (ValidatorStake, Option<ConsensusPublicKey>)>
         /// Contains test validator set: <validator_address, (stake and consensus_key)>;
         pub validators: HashMap<ValidatorAddress, (ValidatorStake, Option<ConsensusPublicKey>)>,
         /// variable for current validators height, at height our validator set is
--- a/fuel-core-interfaces/src/db.rs
+++ b/fuel-core-interfaces/src/db.rs
@@ -128,8 +128,8 @@ pub mod helpers {
 
     use crate::{
         model::{
-            BlockHeight, Coin, CoinStatus, ConsensusPublicKey, DaBlockHeight, DepositCoin,
-            FuelBlock, FuelBlockConsensus, FuelBlockHeader, SealedFuelBlock, ValidatorAddress,
+            BlockHeight, Coin, CoinStatus, ConsensusId, DaBlockHeight, DepositCoin,
+            FuelBlock, FuelBlockConsensus, FuelBlockHeader, SealedFuelBlock, ValidatorId,
             ValidatorStake,
         },
         relayer::{RelayerDb, StakingDiff},
@@ -145,9 +145,9 @@ pub mod helpers {
 
     #[derive(Clone, Debug)]
     pub struct Data {
-        /// Used for Storage<ValidatorAddress, (ValidatorStake, Option<ConsensusPublicKey>)>
+        /// Used for Storage<ValidatorId, (ValidatorStake, Option<ConsensusId>)>
         /// Contains test validator set: <validator_address, (stake and consensus_key)>;
-        pub validators: HashMap<ValidatorAddress, (ValidatorStake, Option<ConsensusPublicKey>)>,
+        pub validators: HashMap<ValidatorId, (ValidatorStake, Option<ConsensusId>)>,
         /// variable for current validators height, at height our validator set is
         pub validators_height: DaBlockHeight,
         /// Used for Storage<DaBlockHeight, StakingDiff>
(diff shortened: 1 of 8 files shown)


fuel-core-interfaces/src/db.rs
```diff
@@ -146,7 +147,7 @@ pub mod helpers {
     pub struct Data {
         /// Used for Storage<Address, (u64, Option<Address>)>
         /// Contains test validator set: <validator_address, (stake and consensus_key)>;
-        pub validators: HashMap<Address, (ValidatorStake, Option<Address>)>,
+        pub validators: HashMap<ValidatorAddress, (ValidatorStake, Option<ConsensusPublicKey>)>,
```
adlerjohn, line 150: Probably want to update the docstrings for this and other places the name was changed.

fuel-core-interfaces/src/model.rs
```diff
@@ -9,12 +9,16 @@ pub use block::{FuelBlock, FuelBlockConsensus, FuelBlockDb, FuelBlockHeader, Sea
...
 pub use deposit_coin::DepositCoin;
+use fuel_types::Address;
 pub use txpool::{ArcTx, TxInfo};
 pub use vote::Vote;
 
 pub type DaBlockHeight = u32;
 pub type ValidatorStake = u64;
 
+pub type ValidatorAddress = Address;
+pub type ConsensusPublicKey = Address;
```
adlerjohn, line 20: An address is the hash of a public key. In this case, is `ConsensusPublicKey` actually a public key (compressed)? If so, it might be better to have it have a different _direct_ underlying type, even though the ultimate representation is still 32 bytes.

fuel-core-interfaces/src/model.rs
```diff
@@ -9,12 +9,16 @@ pub use block::{FuelBlock, FuelBlockConsensus, FuelBlockDb, FuelBlockHeader, Sea
...
 pub use coin::{Coin, CoinStatus};
 pub use deposit_coin::DepositCoin;
+use fuel_types::Address;
 pub use txpool::{ArcTx, TxInfo};
 pub use vote::Vote;
 
 pub type DaBlockHeight = u32;
 pub type ValidatorStake = u64;
 
+pub type ValidatorAddress = Address;
```
adlerjohn, line 19: Can you document this and the consensus public key types?

fuel-core-interfaces/src/model.rs
rakita: ValidatorAddress is eth related address.
But for `ConsensusPublicKey`, my first thought is that this must be a public key as we are using it for verification, but as we can do recovery of the public key from signature, `ConsensusPublicKey` can be made as a hash. Maybe leave it as a compressed public key without prefix for now. In future, it could be a consensus change to make it into hash.

rakita: For this pull request i just changed `Adress` to `Bytes32`

rakita:
@Voxelot about naming, take 10s, think about it and decide on what we want to use here, I started to overthink it :). The change will be a search and replace.

Voxelot:
`ValidatorId` & `ConsensusId`, that way we are flexible wrt Address vs Public Key semantics.

adlerjohn:
That's a good suggestion, as it abstracts away the implementation (uncompressed pubkey, hash of pubkey, 20 bytes from hash of pubkey, compressed pubkey).

rakita:
renamed to `ValidatorId` and `ConsensusId`

adlerjohn (commented):
looks good to me, will defer to @Voxelot and @vlopes11 for final review

fuel-core-interfaces/src/db.rs
```diff
@@ -792,33 +792,33 @@ pub mod helpers {
...
             Ok(self.data.lock().validators.insert(*key, *value))
         }
 
         fn remove(
             &mut self,
-            _key: &Address,
-        ) -> Result<Option<(ValidatorStake, Option<Address>)>, Self::Error> {
+            _key: &ValidatorId,
+        ) -> Result<Option<(ValidatorStake, Option<ConsensusId>)>, Self::Error> {
             unreachable!()
```
vlopes11, line 810: Unrelated to this PR, but how is that unreachable since the inner counterpart is a hashmap that trivially implements `remove`? If we really want to block keys removal, we should return an `Err` instead - and not panic.

fuel-core-interfaces/src/db.rs
```diff
@@ -792,33 +792,33 @@ pub mod helpers {
...
-        ) -> Result<Option<std::borrow::Cow<'a, (ValidatorStake, Option<Address>)>>, Self::Error>
+            key: &ValidatorId,
+        ) -> Result<Option<std::borrow::Cow<'a, (ValidatorStake, Option<ConsensusId>)>>, Self::Error>
         {
             Ok(self.data.lock().validators.get(key).map(|i| Cow::Owned(*i)))
         }
 
-        fn contains_key(&self, _key: &Address) -> Result<bool, Self::Error> {
+        fn contains_key(&self, _key: &ValidatorId) -> Result<bool, Self::Error> {
             unreachable!()
```
vlopes11, line 822: Ditto; out of context `unreachable`

fuel-core-interfaces/src/relayer.rs
```diff
@@ -64,7 +64,7 @@ pub trait RelayerDb:
     /// Query delegate index to find list of blocks that delegation changed
     /// iterate over list of indexed to find height that is less but closest to da_height
     /// Query that block StakeDiff to find actual delegation change.
-    async fn get_first_lesser_delegation(&mut self,delegate: &Address, da_height: DaBlockHeight) ->  Option<HashMap<Address,u64>> {
+    async fn get_first_lesser_delegation(&mut self,delegate: &Address, da_height: DaBlockHeight) ->  Option<HashMap<ValidatorId,ValidatorStake>> {
```
vlopes11, line 67: Also unrelated to the PR, but we are forcing the user to reallocate a new map when we can trivially return an iterator.

fuel-core-interfaces/src/relayer.rs
```diff
@@ -106,10 +106,10 @@ pub trait RelayerDb:
 
     /// Apply validators diff to validator set and update validators_da_height. This operation needs
     /// to be atomic.
-    async fn apply_validator_diffs(&mut self, da_height: DaBlockHeight, changes: &HashMap<Address,(ValidatorStake,Option<Address>)>) {
+    async fn apply_validator_diffs(&mut self, da_height: DaBlockHeight, changes: &HashMap<ValidatorId,(ValidatorStake,Option<ConsensusId>)>) {
```
vlopes11, line 109: Another unrelated nit; we should either take the map as owned so we won't copy every element, or take an iterator.

vlopes11 (approved):
LGTM, couple of unrelated nits for future discussion

fuel-core-interfaces/src/relayer.rs
rakita: One of the reasons is this: https://github.com/FuelLabs/fuel-core/issues/430
And I think I already tried it but don't remember what other problem was there, I think it is returning impl iter from trait.
I can agree that iterator seems a better and cleaner way, but this is good enough as we are always in need of a full set.

fuel-core-interfaces/src/relayer.rs
rakita: Storage `insert` works on references, so it is internally copied. Either `impl Iter` would look nicer here

fuel-core-interfaces/src/db.rs
rakita: It is DummyDB used for testing. I created task for it: https://github.com/FuelLabs/fuel-core/issues/431

fuel-core-interfaces/src/db.rs
rakita: Task for it: https://github.com/FuelLabs/fuel-core/issues/431

PR #422, opened 2022-06-17
fix: use gas_price_factor to verify FeeAmount

luizstacio:
After we insert the `gas_price_factor` to calculate gas, the `verify_gas` started to break, because it didn't consider the  `gas_price_factor` when calculating the `gas.checked_sub(total_gas_required)` causing the TX to not pass the validation.

On this fix I made:
- Change the retrieve `gas` available amount to also filter by `AssetId:default()` as outputs do.
- Calculate the `total_gas_required` and divide it by the `gas_price_factor` and `ceil` the result.

Now when returning the error `InsufficientFeeAmount` it would present the two values in the final amount as ex.: `{ provided: 1, required: 2 }`.

Use case:
In a wallet with 1000000 `Native Coins`
I'm going to do a TX that consumes ~2000000 Gas Units which cost in native coins `(<Gas Units>/<Gas Price Factor>) = 2 Native Coins`
When sending the TX it was failing with the return `InsufficientFeeAmount { provided: 1000000, required: 1999992 }`
Wich is incorrect because the `required fee` should be `1999992/1000000 = 2 Native Coins`.

Notes:
If I add an amount of 3000000 `Native Coins` to the wallet, it would work and consume just the correct amount ~2 Native Coins.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2051,7 +2051,7 @@ dependencies = [
 
 [[package]]
 name = "fuel-block-executor"
-version = "0.9.0"
+version = "0.9.1"
 dependencies = [
  "anyhow",
  "fuel-core-interfaces",
@@ -2060,7 +2060,7 @@ dependencies = [
 
 [[package]]
 name = "fuel-block-importer"
-version = "0.9.0"
+version = "0.9.1"
 dependencies = [
  "anyhow",
  "fuel-core-interfaces",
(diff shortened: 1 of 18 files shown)


Voxelot (approved):
Thanks for fixing this! Good deep dive on finding this problem.

This reminds me that we need to consolidate this logic under fuel-tx: https://github.com/FuelLabs/fuel-vm/issues/82#issuecomment-1079307265

vlopes11 (approved):
LGTM! We should open an issue to cover this case with a test

PR #427, opened 2022-06-20
Add Metrics to Batch Write and Key Length Fix

### LINKED ISSUE #416: Enhance DB Metrics
[Explained here](https://github.com/FuelLabs/fuel-core/pull/404#discussion_r898317340)

Metric meters should be incremented when calling batch db operations, and include key_length when incrementing bytes for `iter_all`

ControlCplusControlV:
Closes #416 , added a script and a submit call to ensure `batch_write` is actually tested with metrics. Working fine and minimal changes, just a quick fix.

### DIFF
--- a/fuel-core/src/state/rocks_db.rs
+++ b/fuel-core/src/state/rocks_db.rs
@@ -195,14 +195,15 @@ impl KeyValueStore for RocksDb {
             .iterator_cf_opt(&self.cf(column), opts, iter_mode)
             .map(|(key, value)| {
                 let value_as_vec = value.to_vec();
+                let key_as_vec = key.to_vec();
                 #[cfg(feature = "prometheus")]
                 {
                     DATABASE_METRICS.read_meter.inc();
                     DATABASE_METRICS
                         .bytes_read_meter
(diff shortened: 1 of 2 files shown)


fuel-core/src/state/rocks_db.rs
```diff
@@ -229,7 +230,13 @@ impl BatchOperations for RocksDb {
                 }
             }
         }
-
+        #[cfg(feature = "prometheus")]
+        {
+            DATABASE_METRICS.write_meter.inc();
+            DATABASE_METRICS
+                .bytes_written_meter
+                .inc_by(batch.len() as u64);
```
Voxelot, line 238: Looking at the [docs](https://docs.rs/rocksdb/0.17.0/rocksdb/struct.WriteBatch.html#method.len), it seems like we should be using `size_in_bytes` instead:

```suggestion
                .inc_by(batch.size_in_bytes() as u64);
```

ControlCplusControlV: good catch

Voxelot (approved):

Issue #75, opened 2021-11-22
Snapshot Node State


Voxelot:
Add the ability to export the state of a node to a chain / genesis configuration file. This will allow for things like "forking main net", regenesis, or other testing utilities.

Related to #53

PR #435, opened 2022-06-23
Snapshot to chainspec command

ControlCplusControlV:


ControlCplusControlV:
Moved to PR so I can see everything to clean up easier, mostly working but still sorting out a cleaner test and potential state_config issue

rakita:
I did a little bit of refactoring of the main config: https://github.com/FuelLabs/fuel-core/pull/440 just a heads up as there will be conflicts.

ControlCplusControlV:
Ok, going to hold off requesting review until those are merged then

ControlCplusControlV:
Converting to draft while I removed feature flags as requested

ControlCplusControlV:
Closes #75

fuel-core/src/config/chain_config.rs
```diff
@@ -121,6 +120,16 @@ pub struct StateConfig {
     pub height: Option<BlockHeight>,
 }
 
+impl StateConfig {
+    pub fn generate_state_config(db: Database) -> Self {
+        StateConfig {
+            coins: db.get_coin_config().unwrap(),
```
leviathanbeak, line 126: are these things always assumed to work or should we be more careful with `unwraps` here ?

fuel-core/src/database/coin.rs
```diff
@@ -120,4 +123,36 @@ impl Database {
                 == asset_id
         })
     }
+
+    pub fn get_coin_config(&self) -> Result<Option<Vec<CoinConfig>>, anyhow::Error> {
+        let configs = self
+            .iter_all::<Vec<u8>, Word>(COIN, None, None, None)
+            .map(|raw_coin| -> CoinConfig {
+                let coin = raw_coin.unwrap();
```
leviathanbeak, line 131: few `unwraps` here as well - would it be better that we rather return an `error` instead of panicking here?

fuel-core/src/database/contracts.rs
```diff
@@ -67,6 +69,46 @@ impl Database {
...
     }
+
+    pub fn get_contract_config(&self) -> Result<Option<Vec<ContractConfig>>, anyhow::Error> {
+        let configs = self
+            .iter_all::<Vec<u8>, Word>(CONTRACTS, None, None, None)
+            .map(|raw_contract_id| -> ContractConfig {
+                let contract_id =
+                    ContractId::new(raw_contract_id.unwrap().0[..32].try_into().unwrap());
+
+                // Working?
```
leviathanbeak, line 80: `Working?` comments here - what are they for :D

leviathanbeak (dismissed):
I think there's a lot of `unwraps` everywhere, in methods that actually return `Result<_, anyhow::Error>` so it would be better to cast those errors into `anyhow::Error` and return them rather than potentially `unwrap()` panicking in the middle of the execution.

fuel-core/src/database/coin.rs
ControlCplusControlV: I initialing had trouble unwrapping inside of that closure because I wasn't returning an error, but I agree that it is dangerous and will adjust this to properly propagate errors to the top

fuel-core/src/config/chain_config.rs
ControlCplusControlV: Moving this to `unwrap_or_default`, these should always work and properly return a `None` value if nothing exists, but moving to `unwrap_or_default` so an error in one of them doesn't ruin the generation of the entire config

ControlCplusControlV:
> I think there's a lot of `unwraps` everywhere, in methods that actually return `Result<_, anyhow::Error>` so it would be better to cast those errors into `anyhow::Error` and return them rather than potentially `unwrap()` panicking in the middle of the execution.

I agree, I moved both coin and contract config generation to now `unwrap` much smarter and propagate the error upward if one does occur. I also dropped in better error handling in `state_config` generation by falling back to a default value in an error occurs so the rest of the config can still be exported in the event there is in error in one part of it.

fuel-core/src/args.rs
```diff
@@ -12,14 +22,17 @@ lazy_static::lazy_static! {
...
+#[derive(Parser, Clone, Debug)]
 #[clap(
     name = "fuel-core",
     about = "Fuel client implementation",
     version,
     rename_all = "kebab-case"
 )]
 pub struct Opt {
+    #[clap(subcommand)]
+    pub _snapshot: Option<Snapshot>,
```
Voxelot, line 34: [nit] any reason for the `_` prefix here?

fuel-core/src/args.rs
```diff
@@ -137,3 +178,34 @@ impl Opt {
...
+    {
+        Err(anyhow!(
+            "Rocksdb must be enabled to use the database at {}",
+            path.display()
+        ))
+    }
+
+    #[cfg(feature = "rocksdb")]
+    {
+        let db = Database::open(&path)?;
```
Voxelot, line 193:
```suggestion
        let db = Database::open(&path).context(|| format!("failed to open database at path {}", path))?;
```

fuel-core/src/args.rs
```diff
@@ -137,3 +178,34 @@ impl Opt {
...
+            chain_name: _config.chain_name,
+            block_production: _config.block_production,
+            initial_state: Some(state_conf),
+            transaction_parameters: _config.transaction_parameters,
+        };
+
+        let serialized = serde_json::to_string(&chain_conf).unwrap();
+        let mut stdout = io::stdout().lock();
+
+        stdout.write_all(serialized.as_bytes())?;
```
Voxelot, line 207: 1. No need to unwrap here when we have the convenience of anyhow available
2. Consider using streaming serialization instead of copying the whole string into memory

```suggestion
        let mut stdout = io::stdout().lock();
        let serialized = serde_json::to_writer(stdout, &chain_conf).context("failed to dump snapshot to JSON")?;
```

fuel-core/src/database/coin.rs
```diff
@@ -120,4 +123,36 @@ impl Database {
                 == asset_id
         })
     }
+
+    pub fn get_coin_config(&self) -> Result<Option<Vec<CoinConfig>>, anyhow::Error> {
+        let configs = self
+            .iter_all::<Vec<u8>, Word>(COIN, None, None, None)
+            .map(|raw_coin| -> Result<CoinConfig, anyhow::Error> {
+                let coin = raw_coin.unwrap();
```
Voxelot, line 131: is unwrap needed here?

fuel-core/src/database/coin.rs
```diff
@@ -120,4 +123,36 @@ impl Database {
...
+
+                let byte_id = Bytes32::new(coin.0[..32].try_into()?);
+                let output_index = coin.0[32];
+                // Potentially chop off a byte for output index
+                let tx_id = UtxoId::new(byte_id, output_index);
+
+                let ref_coin = Storage::<UtxoId, CoinModel>::get(self, &tx_id)?;
+
+                let ref_coin = ref_coin.unwrap();
```
Voxelot, line 140: more unwrap

fuel-core/src/database/contracts.rs
```diff
@@ -67,6 +69,40 @@ impl Database {
...
+                    .into_owned()
+                    .into();
+
+                let salt = InterpreterStorage::storage_contract_root(self, &contract_id)?
+                    .unwrap()
+                    .into_owned()
+                    .0;
+
+                let state: Option<Vec<(Bytes32, Bytes32)>> =
+                    self.get(contract_id.as_ref(), CONTRACTS_STATE)?;
```
Voxelot, line 90: `get` only fetches a single key, we need to use `iter_all` to collect _all_ the state of the contract. Also these columns use a `MultiKey`, where contract id is just the prefix to the actual state key.

fuel-tests/tests/snapshot.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    // setup config
+    let mut config = Config::local_node();
+    config.chain_conf.initial_state = Some(StateConfig {
+        height: None,
+        contracts: Some(vec![ContractConfig {
+            code: vec![8; 32],
+            salt: Salt::new([9; 32]),
+            state: None,
+            balances: None,
```
Voxelot, line 33: test should cover state and balances as well, as it would reveal current problems in the implementation

fuel-core/src/args.rs
```diff
@@ -137,3 +178,34 @@ impl Opt {
         })
     }
 }
+
+pub fn dump_snapshot(path: PathBuf, _config: ChainConfig) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
ControlCplusControlV: config is unused if rocksdb isn't enabled because a snapshot couldn't be taken, and so it's never used. I got around this with path by printing it in the error, would you like me to do the same with config?

fuel-core/src/args.rs
ControlCplusControlV: I added this in but not commit suggestion because it actually further can have some changes such to make this look like
```rust
        let stdout = io::stdout().lock();
        serde_json::to_writer(stdout, &chain_conf).context("failed to dump snapshot to JSON")?;
        ```

File: fuel-core/src/args.rs (hunk 11 above)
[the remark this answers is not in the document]
Comment (Voxelot) [reply fuel-core/src/args.rs:182]: Ah makes sense, I think it's fine as is then.

File: fuel-tests/Cargo.toml [hunk 12]
```diff
@@ -32,10 +35,12 @@ fuel-txpool = { path = "../fuel-txpool" }
 futures = "0.3"
 insta = "1.8"
 itertools = "0.10"
+portpicker = "0.1.1"
```
Voxelot, line 38: This doesn't seem necessary now that we removed full integ testing of the CLI

fuel-tests/Cargo.toml
```diff
@@ -16,14 +16,17 @@ name = "integration_tests"
...
 [[test]]
 name = "metrics_test"
 path = "tests/metrics.rs"
 required-features = ["fuel-core/prometheus", "fuel-core/rocksdb"]
 harness = true
 
 [dependencies]
+async-std = "1.12"
 chrono = { version = "0.4", features = ["serde"] }
+escargot = "0.5.7"
```
Voxelot, line 29: Also not needed anymore

fuel-tests/Cargo.toml
ControlCplusControlV: will remove

fuel-tests/Cargo.toml
ControlCplusControlV: will remove

fuel-core/src/args.rs
```diff
@@ -12,14 +19,17 @@ lazy_static::lazy_static! {
...
 
-#[derive(Parser, Debug)]
+#[derive(Parser, Clone, Debug)]
 #[clap(
     name = "fuel-core",
     about = "Fuel client implementation",
     version,
     rename_all = "kebab-case"
 )]
 pub struct Opt {
```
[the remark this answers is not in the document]
rakita: https://github.com/FuelLabs/fuel-core/issues/457

fuel-core/src/args.rs
```diff
@@ -64,8 +74,35 @@ pub struct Opt {
     pub predicates: bool,
 }
 
+#[derive(clap::Subcommand, Clone, Debug)]
+pub enum Snapshot {
+    Snapshot(SnapshotCommand),
+}
```
rakita, line 80: [nit] This probably deserved it's own mod, as in `args/snapshot.rs`, can be put under effort of refactoring cli subcommand: https://github.com/FuelLabs/fuel-core/issues/457

fuel-core/src/args.rs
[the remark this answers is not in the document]
rakita: Usually, you specify two functions one with an error and one without it and use cfg to enable/disable them.

fuel-core/src/args.rs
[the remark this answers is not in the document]
ControlCplusControlV: I was wondering if it was worth re-vamping the cli structure with this PR but wanted to keep changes minimal. I definitely agree though because this implementation is a bit clunky

fuel-core/src/database/coin.rs
ControlCplusControlV: This is an option Unwrap not a result unwrap, so should be safe right?

fuel-core/src/args/run.rs
```diff
@@ -0,0 +1,149 @@
...
+            // use machine parseable structured logs
+            sub
+                // disable terminal colors
+                .with_ansi(false)
+                .with_level(true)
+                .with_line_number(true)
+                // use json
+                .json()
+                .init();
+        }
```
Voxelot, line 97: As this logging setup could be used for both snapshot and running, it should be moved to the central `run_cli` rather than only happening when a node is run.

fuel-core/src/database/coin.rs
Voxelot: Why do we need to fetch the coin by ID if we've already unwrapped it from `raw_coin` above?

fuel-core/src/args/run.rs
ControlCplusControlV: Moved it into the outer `args.rs` so it should be accessible for subcommands, and starts regardless of the subcommand being run

fuel-tests/tests/snapshot.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    let state_conf = StateConfig::generate_state_config(db);
+
+    assert!(state_conf.contracts.is_some());
+    assert!(state_conf.coins.is_some());
+
+    let contract_config = &state_conf.contracts.unwrap()[1];
+
+    assert!(contract_config.balances.is_some());
+    assert!(contract_config.state.is_some());
```
Voxelot, line 90: We should have assertions that the state key, state values, and balances are correct, and not just set to any random value. Perhaps we should split this test up across the various areas of concern, such as a test for each of these:
- contract deployments (eg. bytecode, salt)
- contract balances
- contract state
- coins
- block height

These tests should verify the exact expected values.

ControlCplusControlV:
I removed the expensive and ugly double collection, and then remove the extra get within the balances state collection (fixing what @Voxelot had asked in using all of the values provided in `iter_all`)

fuel-tests/tests/snapshot.rs
ControlCplusControlV: Ok, will expand this test. Sorry didn't see this until I had just re-requested review in the 2nd commit

fuel-core/src/config/chain_config.rs
```diff
@@ -121,6 +120,16 @@ pub struct StateConfig {
     pub height: Option<BlockHeight>,
 }
 
+impl StateConfig {
+    pub fn generate_state_config(db: Database) -> Self {
+        StateConfig {
+            coins: db.get_coin_config().unwrap_or_default(),
+            contracts: db.get_contract_config().unwrap_or_default(),
+            height: db.get_block_height().unwrap_or_default(),
```
Voxelot, line 128: This method should propagate the errors from the database methods without unwrap. Otherwise, the snapshot could silently fail while still providing partially valid results without the user knowing.

```suggestion
    pub fn generate_state_config(db: Database) -> anyhow::Result<Self> {
        StateConfig {
            coins: db.get_coin_config()?,
            contracts: db.get_contract_config()?,
            height: db.get_block_height()?,
```

ControlCplusControlV: I initially did this so corruption in one column would allow the rest to be saved, but I see how this silent failure could be more harmful

Voxelot: This could be enabled in the future using a `continue-on-error` flag, however, any errors should be reported via `error!` or `warn!` before `unwrap_or_default` so the user is aware that their snapshot may come with some issues.

Voxelot (approved):

PR #436, opened 2022-06-24
add graceful shutdown to api server

Voxelot:
this will handle sigkill commands to drain existing api connections before terminating the server


rakita:
This is great, but pin can be removed, check this: https://github.com/FuelLabs/fuel-core/pull/437

rakita (approved):
lgtm

PR #440, opened 2022-06-24
Move config to separate mod. Include module configurations.

rakita:
Refactoring of main config structure and inclusion of other module configurations into one config.


ControlCplusControlV:
Is there a reason p2p doesn't get it's config? I know its still a WIP so fine if it doesn't get one now, but may be nice to make an issue so it eventually gets added as I know lots of people tweak p2p config on Geth

Voxelot:
Yeah P2P is still early but will eventually be included in the main fuel-core service config just like relayer / txpool.

Voxelot (approved):

fuel-core/src/config/chain_config.rs
```diff
@@ -21,12 +19,22 @@ pub const TESTNET_INITIAL_BALANCE: u64 = 10_000_000;
 pub struct ChainConfig {
```
Voxelot, line 19: I'm thinking we should maybe rename this to chainspec and keep it in a different module namespace. It's more focused on genesis config, while the other config concerns are primarily runtime configurable params for tuning various modules within the client.

ControlCplusControlV (approved):

adlerjohn (approved):
utACK

fuel-core/src/config/chain_config.rs
rakita: Yeah, `Chainspec` is a lot nicer.

adlerjohn (approved):

PR #441, opened 2022-06-24
implement DummyDB functionality for tests

### LINKED ISSUE #431: Fill `unrechable!` function in DummyDB that is used for tests
DummyDB is testing structure that implements required `Storage` traits and have test data that can be easily used.
This task is low priority as it does not effect anyone but we can remove `unreachable!` and implement functionality found there:
https://github.com/FuelLabs/fuel-core/blob/981d1a45ad5190725c2ca4a6a7d893ede76e555b/fuel-core-interfaces/src/db.rs#L678

thdaraujo:
[fixes #431]

Implements DummyDB functionality to be used on other tests.


fuel-core-interfaces/src/db.rs
```diff
@@ -124,7 +124,7 @@ pub mod helpers {
         Address, Bytes32, ContractId, Input, Metadata, Output, Transaction, TxId, UtxoId,
     };
     use fuel_vm::prelude::Contract;
-    use std::collections::{HashMap, HashSet};
+    use std::collections::HashMap;
```
thdaraujo, line 127: Here I'm assuming this should be a Map and not a Set.

fuel-core-interfaces/src/db.rs
```diff
@@ -166,7 +166,7 @@ pub mod helpers {
         /// Dummy coins
         pub coins: HashMap<UtxoId, Coin>,
         /// Dummy contracts
-        pub contract: HashSet<ContractId>,
+        pub contract: HashMap<ContractId, Contract>,
```
thdaraujo, line 169: I'm assuming here that this should be a Map and not a Set.

ControlCplusControlV:
Do you have any tests for this? Or how have you been testing your changes?

thdaraujo:
> Do you have any tests for this? Or how have you been testing your changes?

Yep! I do have some tests, but they look a bit repetitive. I was trying to make them more generic, but probably the only way is by using macros.

I'll commit my tests for transactions to show you what I mean,

fuel-core-interfaces/src/db.rs
```diff
@@ -672,51 +676,72 @@ pub mod helpers {
...
-            _key: &Bytes32,
-            _value: &Transaction,
+            key: &Bytes32,
+            value: &Transaction,
         ) -> Result<Option<Transaction>, Self::Error> {
-            unreachable!()
+            let arc = Arc::new(value.clone());
+
+            if let Some(tx) = self.data.lock().tx.insert(*key, arc) {
+                Ok(Some(tx.as_ref().clone()))
```
thdaraujo: FYI: this method will return the old value (the replaced value) that was overwritten. That was my understanding after reading this:
https://github.com/FuelLabs/fuel-storage/blob/62f163dc21c712e2fa8dd9a6a76cf94f82ffc0d0/src/lib.rs#L27

fuel-core-interfaces/src/db.rs
```diff
@@ -672,51 +676,72 @@ pub mod helpers {
...
             &mut self,
-            _key: &Bytes32,
-            _value: &Transaction,
+            key: &Bytes32,
+            value: &Transaction,
         ) -> Result<Option<Transaction>, Self::Error> {
-            unreachable!()
+            let arc = Arc::new(value.clone());
+
+            if let Some(previous_tx) = self.data.lock().tx.insert(*key, arc) {
```
thdaraujo, line 684: FYI: this method will return the old value (the replaced value) that was overwritten. That was my understanding after reading this:
https://github.com/FuelLabs/fuel-storage/blob/62f163dc21c712e2fa8dd9a6a76cf94f82ffc0d0/src/lib.rs#L27

fuel-core-interfaces/src/db.rs
```diff
@@ -918,3 +948,42 @@ pub mod helpers {
...
+    #[test]
+    fn transactions_db() -> Result<(), KvStoreError> {
+        let mut db: Box<dyn Storage<Bytes32, Transaction, Error = KvStoreError>> =
+            Box::new(DummyDb::filled());
+
+        let value: Transaction = Transaction::default();
+        let key = value.id();
+
+        //before insert
+        assert!(!db.contains_key(&key)?);
```
thdaraujo, line 969: I would like to reuse all the tests below and only change the db, value and key for the specific implementation. That way, I could test remove some duplication.
 
But not entirely sure how to do it without macros or leveraging generics. Maybe you have a better suggestion.

Voxelot: I'd recommend looking into [rstest](https://github.com/la10736/rstest) as we already use it in other areas. The db could be instantiated via fixture.

thdaraujo: Thanks for the info!

I added more test cases and removed most of the duplication, and also found a way to to instantiate the db without needing fixtures after all.

Voxelot (approved):
nice and simple approach 👍🏻

ControlCplusControlV (approved):
LGTM

PR #446, opened 2022-06-27
refactor: add WASM build support

Br1ght0ne:
This PR adds WASM build support for all crates in this repo *except* for:
- `fuel-relayer` (`libp2p` does not provide some required functionality on WASM)
- the TCP `Service` in `fuel-core` (`hyper`'s TCP functionality is also not available)

Needed for continuing work on https://github.com/FuelLabs/fuels-rs/pull/429.
Impact on CI time is negligible, as `rustup target add wasm32-unknown-unknown` only downloads WASM stdlib (<1 sec).


Br1ght0ne:
`[fuel-client]$ cargo tree | grep openssl` gives nothing. `openssl` is not a dependency, but is still in `Cargo.lock`. Not sure why.

xgreenx (approved):

Br1ght0ne:
Need to debug build issues after rebasing on fresh `master`.

Voxelot:
What is the usecase for making all of fuel-core wasm compatible? This seems like it will inflict more pain on the fuel-core dev process without solving any of our immediate needs. It could be a cool way to run something like a light-client in the browser, but the extra build and test complexity of making everything in `FuelService` / `fuel-core-bin` wasm friendly doesn't seem like something we should prioritize given we plan to ship fuel-core as a binary that others run in docker etc.

xgreenx:
I close this PR as outdated, but you are free to create a new one based on the master=D

BTW, it is hard to make full `fuel-core` WASM compatible while `fuel-vm` is not WASM compatible.

Issue #447, opened 2022-06-27
Integrate fuel-relayer in fuel-core


rakita:
Integrate `fuel-relayer` and allow it to be enabled and run with `fuel-core`

PR #450, opened 2022-06-29
Relayer integration. ServiceBuilder. Context

rakita:
Integrate `fuel-relayer` as `fuel-core` module and connect it to other module channels. Introduce `relayer::Sender` 

Introduce `ServiceBuilder` in relayer so that we can build service and service can be called with  `start()` without any inputs.
Introduce notion of `Context` inside relayer so that we know what to share to `fuel_relayer::run` and what to expect to be returned back and kept in relayer server.


fuel-core/src/database/metadata.rs
```diff
@@ -1,24 +1,41 @@
...
 
+pub(crate) const DB_VERSION_KEY: &[u8] = b"version";
 pub(crate) const CHAIN_NAME_KEY: &[u8] = b"chain_name";
 pub(crate) const CHAIN_HEIGHT_KEY: &[u8] = b"chain_height";
 pub(crate) const FINALIZED_DA_HEIGHT_KEY: &[u8] = b"finalized_da_height";
 pub(crate) const VALIDATORS_DA_HEIGHT_KEY: &[u8] = b"current_validator_set";
 pub(crate) const LAST_COMMITED_FINALIZED_BLOCK_HEIGHT_KEY: &[u8] =
     b"last_commited_finalized_block_height";
 
+pub(crate) const DB_VERSION: u32 = 0;
```
Voxelot, line 15: We should cleanup existing db version logic as well:

https://github.com/FuelLabs/fuel-core/blob/master/fuel-core/src/database.rs#L55
https://github.com/FuelLabs/fuel-core/blob/master/fuel-core/src/state/rocks_db.rs#L57

rakita: I didnt even see that there is a version there, I moved constants in `metadata.rs`

fuel-core/src/service/genesis.rs
```diff
@@ -39,14 +41,6 @@ impl FuelService {
         Ok(())
     }
 
-    /// initialize starting block height if set
-    fn init_block_height(db: &Database, state: &StateConfig) -> Result<()> {
```
Voxelot, line 43: This isn't the same as the other init fns, it allows the chain config to specify the starting block height of a chain. My thinking was this could be used to re-genesis / fork a network using only the state config (e.g. fork main-net at given height for contract testing)

fuel-core/src/service/modules.rs
```diff
@@ -1,6 +1,7 @@
 use crate::config::Config;
 use crate::database::Database;
 use anyhow::Result;
+use fuel_core_interfaces::relayer::RelayerDb;
```
Voxelot, line 4: [nit] Unrelated to this PR, but what do you think about moving the RelayerDb trait to the relayer crate? Since fuel-core already depends on the relayer, it would largely just be changing the import paths. My thinking is that we don't expect other modules outside of the relayer to access the RelayerDb directly, they should only use the Sender interface. So we probably shouldn't share guarded / lower level implementation requirements of modules with each other.

fuel-core/src/service/modules.rs
```diff
@@ -38,8 +40,17 @@ pub async fn start_modules(config: &Config, database: &Database) -> Result<Modul
...
-    // let mut relayer = FuelRelayer::new(FuelRelayerConfig::default());
-    // let mut p2p = FuelP2P::new(FuelP2PConfig::default());
+
+    let mut relayer_builder = fuel_relayer::ServiceBuilder::new();
+    relayer_builder
+        .config(config.relayer.clone())
+        .db(Box::new(database.clone()) as Box<dyn RelayerDb>)
+        .import_block_event(block_importer.subscribe())
+        .private_key(
+            hex::decode("c6bd905dcac2a0b1c43f574ab6933df14d7ceee0194902bce523ed054e8e798b")
```
Voxelot, line 50: Shouldn't this be part of the relayer config?

fuel-relayer/src/service.rs
```diff
@@ -1,67 +1,197 @@
...
-    pub async fn stop(&mut self) {
-        let _ = self.sender.send(RelayerEvent::Stop);
-        if let Some(join) = self.stop_join.take() {
-            let _ = join.await;
+    pub fn config(&mut self, config: Config) -> &mut Self {
+        self.config = config;
+        self
+    }
+
+    pub fn build(self) -> Result<Service, anyhow::Error> {
```
Voxelot, line 70: [very minor nit]
```suggestion
    pub fn build(self) -> anyhow::Result<Service> {
```

fuel-relayer/src/service.rs
```diff
@@ -1,67 +1,197 @@
...
+                let uri = context.config.eth_client.clone();
+                if uri.starts_with("http") {
+                    self.run(&mut *join, context, Self::provider_http(&uri))
+                        .await
+                } else if uri.starts_with("ws") {
+                    self.run(&mut *join, context, Self::provider_ws(&uri).await)
+                        .await
+                } else {
+                    *self.context.lock().await = Some(context);
+                    Err(anyhow!("Relayer uclient uri is not http or ws"))
```
Voxelot, line 169: `uclient` -> `client`

```suggestion
                    Err(anyhow!("Relayer client uri is not http or ws"))
```

Voxelot (approved):
Very clean and nice PR! The use of context as a way to guard singleton resources in a threadsafe way seems like a decent approach. Although I wonder if the same simplicity could be accomplished by merging the context and relayer together?

Thanks for all the database tech debt cleanup as well.

adlerjohn (dismissed):
utACK

fuel-core/src/service/modules.rs
rakita: The best way would be to create `fuel-storage` and extract database related stuff from `fuel-core`. In this case leaving `RelayerDb` in `fuel-core-interface` makes sense. I feel like those traits for db are going to be refactored in future to aggregate them and make better intention/clarification for usage, but for now, they seem good enough.

fuel-core/src/service/modules.rs
rakita: I missed removing this. I have one last thing to do with the integration of `fuel-relayer` and that would be to include config in cli. This will be resolved there: https://github.com/FuelLabs/fuel-core/issues/447

fuel-relayer/src/service.rs
rakita: https://github.com/FuelLabs/fuel-core/pull/450/commits/6bd6fb4e32e3c64df29373f64e561f571b68a171

rakita:
> Very clean and nice PR! The use of context as a way to guard singleton resources in a threadsafe way seems like a decent approach. Although I wonder if the same simplicity could be accomplished by merging the context and relayer together?
> 

It is a good improvement, context can contain externalities to connect to other modules, and there is a clear separation of external dependencies and it allows us to have `start` without needing inputs.

I didn't think a lot about this, for now, it is still just an idea for the future, but having a separate context allows us to have a clearer idea about what  external connections  we can switch. As in inserting a sniffing channel that would log interaction/calls between for example `fuel-bft` and `fuel-relayer` or `fuel-producer` and `fuel-txpool` it is an interesting idea to think about but I wouldn't want to invest time atm until we better know about requirements and restrictions in our system.

fuel-core/src/service/modules.rs
rakita: i meant `fuel-core-storage` or `fuel-database` as we already have `fuel-storage`

adlerjohn (approved):

PR #461, opened 2022-07-05
Controlc/block api

### LINKED ISSUE #455: feat: add API to mine block(s) without requiring signing a transaction
Currently, the way to mine blocks is to submit empty transactions. This means having  to sign transactions with a wallet, and potentially even having to pay for gas. It would be cool to have an API that just mines blocks, similar to Foundry/Hardhat.

ControlCplusControlV:
This PR adds a simple `advance_block` endpoint, which takes in an option of the amount to advance forward the block-height by (defaulting to 1 if nothing is supplied). I then inserts n empty `FuelBlocks`, and returns the new block-height. I was hoping to confirm this was inline with what you were looking for @vnepveu ?

Closes #455


Voxelot:
We should have a config option for the node to disable this behavior in production nodes. Otherwise strange things could happen.

ControlCplusControlV:
> We should have a config option for the node to disable this behavior in production. Otherwise strange things could happen.

Oh true, wasn't sure how exactly best to do this. Should this endpoint be specifically addressed (ie config option like `enable rpc-block-advance`) or under a more general `dev/testnet` setting?

ControlCplusControlV:
@Voxelot I've updated this to require a config option to enabled (which is disabled by default) for this endpoint to work, and added a negative test to ensure it's off by default

fuel-core/src/schema/block.rs
```diff
@@ -175,3 +178,53 @@ impl BlockQuery {
...
+
+#[derive(Default)]
+pub struct BlockMutation;
+
+#[Object]
+impl BlockMutation {
+    async fn advance_block(
+        &self,
+        ctx: &Context<'_>,
+        advance_by: Option<U64>,
```
vnepveu, line 190: Same here: the `Option` wrapper just adds complexity for little gain IMO (with the unwrap eg)

fuel-tests/tests/blocks.rs
```diff
@@ -32,6 +32,38 @@ async fn block() {
...
+
+    let srv = FuelService::from_database(db, Config::local_node())
+        .await
+        .unwrap();
+
+    let client = FuelClient::from(srv.bound_address);
+
+    let new_height = client.advance_block(Some(5)).await;
+
+    assert!(new_height.is_err());
```
vnepveu, line 64: Check that the error has the right message about RPC control having to be enabled

vnepveu (changes requested):
vgud! ty for being so quick! Just some questions about naming :)

fuel-core/src/schema/block.rs
ControlCplusControlV: My thinking with this was the someone could call `mine()` if they just want to advance, then have the option to specify an amount `mine(5)`, with the thinking that it could then be a optional arg. I get now though with Rust that style of design isn't supported and would have to be `mine(None)`, will change though

fuel-client/assets/schema.sdl
```diff
@@ -282,6 +282,7 @@ type Mutation {
 	Submits transaction to the txpool
 	"""
 	submit(tx: HexString!): Transaction!
+	advanceBlock(advanceBy: U64): U64!
```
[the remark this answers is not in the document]
ControlCplusControlV: I think other libraries use `mineBlock` for this but since that isn't really what's happening here I used `advance`, I can move it over to produce though as that also conveys the same message but more detailed

fuel-core/src/args.rs
```diff
@@ -41,6 +41,10 @@ pub struct Opt {
     #[clap(name = "CHAIN_CONFIG", long = "chain", default_value = "local_testnet")]
     pub chain_config: String,
 
+    /// Allows RPC Endpoints to arbitrarily modify the VM. Used for local development only
+    #[clap(long = "enable_rpc_control")]
+    pub enable_rpc_control: bool,
```
Voxelot, line 46: We don't refer to our endpoints as RPC, as we don't use any rpc frameworks like json-rpc or grpc. I think we should have a special feature flag just for manual block production, and we could make super features out of the smaller ones later if we find that they are commonly enabled together.

Also, we should use kebab casing like our other arg flags (`vm-backtrack`, `utxo-validation`).

```suggestion
    #[clap(long = "manual-blocks")]
    pub manual_blocks_enabled: bool,
```

fuel-tests/tests/blocks.rs
```diff
@@ -32,6 +32,41 @@ async fn block() {
...
+    let mut config = Config::local_node();
+
+    config.manual_blocks_enabled = true;
+
+    let srv = FuelService::from_database(db, config).await.unwrap();
+
+    let client = FuelClient::from(srv.bound_address);
+
+    let new_height = client.produce_block(5).await.unwrap();
+
```
Voxelot, line 48: We should probably strengthen this test to verify that the block height was actually increased. Otherwise this test could continue to pass even if the underlying implementation of the API was completely gutted while returning the expected response.

This might look something like:
 - When I call produce_block with 5,
 - Any transactions I submit afterward are 5 blocks higher than without calling produce_block.

This establishes an expected cause and effect in behavior and also ensures the node continues to function as expected (within reason) after manually creating some blocks.

fuel-core/src/args.rs
```diff
@@ -42,8 +42,8 @@ pub struct Opt {
     pub chain_config: String,
 
     /// Allows RPC Endpoints to arbitrarily modify the VM. Used for local development only
```
vnepveu, line 44: IIUC refering to "RPC" here isn't accurate?

vnepveu (changes requested):
One nit left, otherwise LGTM :)

fuel-core/src/args.rs
ControlCplusControlV: Good catch, forgot to update the comment when I changed the name

fuel-tests/tests/blocks.rs
ControlCplusControlV: Just added in my tests this logic to ensure the block height actually advances and is represented as such

Voxelot (approved):

PR #466, opened 2022-07-07
relayer cli

rakita:
some small renamings and adding `cli` for relayer


Voxelot:
we should ensure the relayer doesn't log any error messages if the eth_client URL is unset.

rakita:
@Voxelot  already done that in this pull request

Voxelot:
ok great, looks like this just needs a rebase with master since @ControlCplusControlV's PR got merged.

fuel-core/src/cli/run/relayer.rs
```diff
@@ -0,0 +1,61 @@
...
+use fuel_relayer::{Config, H160};
+use std::str::FromStr;
+use std::time::Duration;
+
+#[derive(Debug, Clone, Args)]
+pub struct RelayerArgs {
+    /// Uri address to ethereum client. It can be in format of `http://localhost:8545/` or `ws://localhost:8545/`.
+    /// If not set relayer will not start.
+    #[clap(long = "relayer", default_value = "none")]
+    pub eth_client: String,
```
Voxelot, line 12: Why not just use `Option` instead of setting the string to "none" by default?

Voxelot (dismissed):

Voxelot (approved):

PR #485, opened 2022-07-12
Relayer DaMessage

rakita:
Replace DepositCoin with DaMessage


fuel-core-interfaces/src/model/messages.rs
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    pub fn lock(self) -> DaMessageLocked {
+        let id = self.id();
+        DaMessageLocked { message: self, id }
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct DaMessageLocked {
```
rakita, line 38: Hm i used here `Locked` naming, while Transaction has `Checked`.
Idea is that we `lock` the data that is not possible to be changed in a locked state (It has `AsRef` and `unlock` function). Are we okay to have two namings or should i rename it to `Checked` for consistency?

pixelcircuits:
Just curious. Why did you name it DaMessage? Is the DA short for data availability?

fuel-core-interfaces/src/model/messages.rs
pixelcircuits: If they're the same thing, then we should make it a trait and name them under a similar convention both implementing that trait. If they don't behave similar enough to share a trait, then I don't have a problem with them just being named in a different style

fuel-relayer/src/finalization_queue.rs
```diff
@@ -208,8 +209,9 @@ impl FinalizationQueue {
         }
         let last_diff = self.pending.back_mut().unwrap();
         match fuel_event {
-            EthEventLog::AssetDeposit(deposit) => {
-                last_diff.assets.insert(deposit.deposit_nonce, deposit);
+            EthEventLog::DaMessage(message) => {
+                let msg = DaMessage::from(&message).lock();
+                last_diff.messages.insert(*msg.id(), msg);
             }
             EthEventLog::Deposit { .. } => {
```
pixelcircuits, line 216: Is this `Deposit` in the enumeration necessary anymore? I'm not sure why it was there in the old deposit/withdrawal setup either since `AssetDeposit` already existed

fuel-core-interfaces/src/model/messages.rs
Voxelot: Lock may be confusing with concurrency related issues. Since we've already discussed using "Checked" types before, and have already used that convention in fuel-tx and fuel-vm for some time, I'd prefer to keep things consistent with Checked. Although if you think Checked isn't a great name, then we should have a team vote and then change it in all the other places to be consistent.

Voxelot (dismissed):

fuel-relayer/src/finalization_queue.rs
Voxelot: https://github.com/FuelLabs/fuel-core/issues/515

Voxelot (approved):

Issue #478, opened 2022-07-11
Include valid broadcast transactions from peers into the local txpool


bvrooman:
There are two halves for implementing the share txs feature, broadcast and receiving. This task is only for tracking the receiving side of this effort.

Expected Test Coverage for completion:

- Unit testing in fuel-txpool to ensure `NewTx` messages are listened to and that the txs are included into the txpool set.

PR #491, opened 2022-07-14
Controlc/tx p2p

### LINKED ISSUE #477: Broadcast transactions to peers when they're inserted into txpool via GraphQL API
There are two halves for implementing the share txs feature, broadcast and receiving. This task is only for tracking the broadcast side of this effort.

Expected Test Coverage for completion:

- GrapqhQL API testing of fuel-core to ensure a proper `TransactionBroadcast::NewTx` event is sent when a tx is submitted via GraphQL, only after it is successfully included in the transaction pool.
- Transactions directly inserted into the txpool do not emit a `NewTx` event, to ensure no feedback loop happens when receiving txs.

ControlCplusControlV:
Design to closes #477 and #478 . Adds 2 new channels to the txpool designed for p2p to hook into. One channel is a `mpsc` channel designed to send transactions received on the txpool from the gql endpoint to p2p, so that it can broadcast them to other nodes. The other is a simple `broadcast` channel which is used to receive transactions from p2p which should be included in the txpool. Tests are bit rough atm and need some polishing, but making a draft PR for general design review and feedback


fuel-txpool/src/service.rs
```diff
@@ -10,9 +10,11 @@ pub struct ServiceBuilder {
     sender: txpool::Sender,
     receiver: mpsc::Receiver<TxPoolMpsc>,
     config: Config,
-    broadcast: broadcast::Sender<TxStatusBroadcast>,
+    consumer: broadcast::Sender<TxStatusBroadcast>,
     db: Option<Box<dyn TxPoolDb>>,
     import_block_events: Option<broadcast::Receiver<ImportBlockBroadcast>>,
+    broadcast_txs: Option<mpsc::Sender<TransactionBroadcast>>,
```
Voxelot, line 16: Based on our earlier call today, I think we should use `P2pRequestEvent` instead of the TransactionBroadcast enum for sending. 

Another minor nit is that `broadcast_txs` isn't totally clear which direction the broadcast is going. Maybe something like `network_outbox` would help clarify the intention?

```suggestion
    network_outbox: Option<mpsc::Sender<P2pRequestEvent>>,
```

ControlCplusControlV: I moved this to P2pRequestEvent, I do remember talking about this, I think it got mixed up between `fuel-core-interfaces/p2p/P2pRequestEvent::BroadcastNewTransaction` and `fuel-core-interfaces/p2p/TransactionBroadcast::NewTransaction` as both have similar names

fuel-txpool/src/service.rs
```diff
@@ -10,9 +13,11 @@ pub struct ServiceBuilder {
     sender: txpool::Sender,
     receiver: mpsc::Receiver<TxPoolMpsc>,
     config: Config,
-    broadcast: broadcast::Sender<TxStatusBroadcast>,
+    consumer: broadcast::Sender<TxStatusBroadcast>,
```
Voxelot, line 16: consumer could mean a lot of things... we could just call this `tx_status_updates`?

bvrooman: I agree. I am aiming to disambiguate some of the naming: `sender`,  `receiver`, `consumer`, etc., are pretty vague and I find it can be confusing trying to understand which channels are used for what. I am taking a stab at that now.

bvrooman:
I noticed that the test `insert_from_p2p` seems to be flaky when I run it locally: most of the times it passes but sometimes it fails. It looks like there's a race condition. We're expecting the same transaction to be sent twice and for an error to raise on the second submission. However, in some cases, it looks like we get to the assert and check the receiver before both submissions are received, causing the test to fail.

ControlCplusControlV:
> I noticed that the test `insert_from_p2p` seems to be flaky when I run it locally: most of the times it passes but sometimes it fails. It looks like there's a race condition. We're expecting the same transaction to be sent twice and for an error to raise on the second submission. However, in some cases, it looks like we get to the assert and check the receiver before both submissions are received, causing the test to fail.

@bvrooman Race condition is most likely because send isn't awaited in any way, so if the send doesn't process in time then insertion could possible go through (as the previous transaction hasn't been inserted yet), leading to no error and the test failing. This is one thing that needs to be fixed in the test, which ideally should just have a base way to test the sending of the transaction rather than measuring side affects

bvrooman:
I have some additional thoughts:

As currently designed in this PR, the txpool service (and inner context) has a new receiver that will receive events for incoming transactions. When it receives one, it will insert it into the txpool. 

This is similar to when the existing txpool receiver gets an insert event, where it then signals the service to insert the transactions into the txpool. However, the difference between inserting transactions and receiving a new transaction is that we must broadcast the transaction and status (submitted, squeezed out) on insertion, but not on new transactions.

Because this PR currently uses the same `TxPool::insert` method in both cases, we had to augment the signature to accept an optional broadcaster:  we pass `Some(network_outbox)` on insertion, but `None` on new transactions. Inside `insert`, we then check against the `Option`, to see if we need to put anything in that outbox. This is a bit of a hack. I'd argue that this approach can be improved upon, and the `insert` method isn't well suited for the latter case of new transactions. Rather, we can introduce a new `TxPool` method, something like `TxPool::new_tx` to handle this case. This is further evidenced by the fact that `insert` is designed to operate on a vector of transactions, rather than a single transaction, and we have to wrap a new transaction in a 1-sized vector to accommodate this interface. 

When I get a chance, I will look at this refactor and push a commit here, along with the ongoing refactoring of the existing service and context channel names.

bvrooman:
We also do not seem to have much consistency on our channel naming conventions. In a number of places, we use terms like "sender", "producer",  and now "outbox" to refer to objects of `Sender`. We use terms like "receiver", "incoming", and "events", for objects of type `Receiver`. Personally, I find this can be fairly confusing and unintuitive to read. It would be great to consolidate on some naming convention for channels.

bvrooman:
I had a conversation on Monday with @Voxelot regarding the feature as well as the currently proposed code:

One key discussion point was that we do not need to perform any broadcasting on new local transactions (at least, not at this time). The agreed action item was to simplify the current implementation, and to remove the broadcasting on this event.

This means a couple of things:
- The handler for this event becomes simpler, and we do not need to send anything to the `network_outbox`
- We do not need to test for broadcasts on local insertions
- The `network_outbox` itself is unnecessary, and it can be removed entirely from the service and context, along with the corresponding service builder methods
- The existing `TxPool::insert` function signature can be reverted back to its original signature, and we no longer need the `Option<Sender>` parameter

This simplifies things in the current PR. I am putting up the commit for this shortly. 

I'd just like to get @Voxelot to confirm that my understanding of this is correct.

CC: @ControlCplusControlV

Voxelot:
I think there's some confusion around the terminology of what was meant by local insert. What I thought we were discussing is that that we don't need to do any additional p2p broadcast for `insert_local` (which is confusingly only actually used when receiving txs from non-local sources like the p2p network). We would still need to broadcast from the main `TxPool::insert` method since that's what the GraphQL API uses to insert txs from a locally submitted txn. If we rip p2p request sender/outbox entirely, I don't know how other nodes would be expected to receive anything in that case if there's no mechanism at all to send messages.

fuel-core/src/service/modules.rs
```diff
@@ -56,9 +59,15 @@ pub async fn start_modules(config: &Config, database: &Database) -> Result<Modul
                 .unwrap(),
         );
 
+    // Meant to simulate p2p's channels which hook in to communicate with txpool
+    let (network_sender, _) = mpsc::channel(100);
```
Voxelot, line 63: Without any network_receiver, this queue will fill up and may stop the txpool from working once it's reached 100 txs.

We likely need to spawn a task to drain this queue whenever txs are added.

bvrooman:
I am squashing & merging this branch into another dev branch to make future rebases easier. We are going to rebase with [Elvis's development branch](https://github.com/FuelLabs/fuel-core/pull/492) to continue development.

Issue #458, opened 2022-07-04
Configuration from file


rakita:
In most cases you would like to specify some cli argument in the config file so that you don't have to use it every time when running `fuel-core` and can just specify the config file you want to run `fuel-core run --config=local.toml`

I don't think it is supported in `clap`  and we need to use additional structure and lib to parse configuration from toml into `fuel-core`

ControlCplusControlV:
This would be a huge UX improvement rather than pasting 100 lines of json into a command, is there a reason the config file would be `.toml` over `.json`/`.yaml` other than homogeneity with Rust's `Cargo.toml`? Right now chain_config is exported via json, so would the config have json values within a `toml` file?

rakita:
there is a good amount of discussion about `toml` vs `yaml` where it is said that `toml` is better for simple configurations. And having it used by rust cargo is good alignment for us.
having `.json` as config is considered unredable and it is used mostly as it is easy to serialize/unserialize data. We can probably have a field in the config that points to `chain_config` json file (renaming it to chainspec.json feels better) as it will contain data for the initialization of the chain. This is something that is not expected to change when setting the genesis block so it seems okay to put that data in json.

Voxelot:
We should still support CLI args as well, since they are easier to templatize and configure in helm charts. 

Toml based config files would align with other parts of our eco-system such as `Forc.toml`. However, there's likely parsers out there that are file-format agnostic since you can generally represent the same kinds of underlying structures using json, yaml or toml.

MitchTurner:
We now have the option of including an env file to specify many of the values you would pass to the cli. However, not all arguments can be read/overridden with the env variables.

If we added support for all the arguments, that would resolve this issue.

thedevbirb:
Hi @MitchTurner, I'd love to start contributing to Fuel. Can I take this issue?
By the way, what is an example of argument that cannot be read/overridden by the env file?

Thanks!

MitchTurner:
@thedevbirb Hey! Welcome! We are always open to contributions from the community. Feel free to work on this and let us know if you have questions.

MitchTurner:
Probably the simplest solution to this is making sure all flags are set-able using env-vars. 

We can also include a test that sets all variables to avoid regressions, etc.

MitchTurner:
@xgreenx 
We currently have env-vars on _most_ fields. Idk if that is good enough to close this issue?

[...]

PR #492, opened 2022-07-15
Include p2p into fuel core

### LINKED ISSUE #480: Integrate p2p-orchestrator into fuel-core and expose configuration
- [ ] include fuel-p2p into fuel-core and integrate orchestrator into services/modules.rs
- [ ] create cli args for the p2p service configs, using the same flattened approach as the relayer.

I suggest doing these as different PR's if possible, in order to unblock downstream tasks.

leviathanbeak:
Resolves #480 

I have included both things in a single PR since once I included p2p into fuel-core and added `p2p` field to `config` the cli part of the codebase was complaining.

For clarity, there is:
- P2pService
- NetworkOrchestrator which contains P2pService 
- Service which wraps the NetworkOrchestrator (I have named it `Service` to keep in line with other modules that had top level `Service`)

### Creating the top level `Service`
I have went and implemented a similar pattern for the network as our other Services have done, by adding `start()` and `stop()` methods.

I have achieved that by wrapping the NetworkOrchestrator into a `Service` but I had to change some things:

#### P2P Service and Network needed to shut down
Unlike  some other simpler Services, once network `Service` was stopped, we needed to shut down the whole p2p network.
I have achieved that by moving `p2p_service` from a `NetworkOrchestrator` field into a simple `variable` within `run()` method, so once `run()` ended (ie `stop()` was called) the `p2p_service` would go out of scope and get removed completely.

And if `start()` was called again, a new `p2p_service` would be instantiated - I had to keep the p2p_config within the NetworkOrchestrator for this case.

#### FuturesUnordered and their blocking of the executor
While testing out `start()` and `stop()` I have discovered a bug in our NetworkOrchestrator.

It's field `outbound_responses` which was of type 
```
type ResponseFuture = Pin<Box<dyn Future<Output = Option<(OutboundResponse, RequestId)>>>>;
FuturesUnordered<ResponseFuture>
```
once called `next()` on it, it would return `None` so in our `loop { tokio::select!...` pattern it was constantly returning `None`
without letting any other items to be polled, basically it was blocking the executor. I discovered this while adding tokio's `sleep` within a test, and it never got to finish! So I have changed this part of the code slightly by using tokio's channels instead of FuturesUnordered.

There were few other changes and clean ups.


leviathanbeak:
@rakita I have just passed some of the p2p channels into other services and removed the placeholders, please review and let me know if they are correct

fuel-core/src/cli/run/p2p.rs
```diff
@@ -0,0 +1,150 @@
...
+                    let mut bytes = std::fs::read(path)?;
+                    Keypair::secp256k1_from_der(&mut bytes)?
+                }
+                _ => Keypair::generate_secp256k1(),
+            }
+        };
+
+        let bootstrap_nodes = {
+            match args.bootstrap_nodes {
+                Some(path) => std::fs::read_to_string(path)?
```
Voxelot, line 107: I'd prefer taking the bootnodes via comma separated cli args instead of loading from a file.

1. There is no snapshot test of the file format or automated tests that exercise this file parsing works as expected
2. For small network configurations it's inconvenient to require a separate dedicated config file just to pass a handful of bootstrap URLs
3. We have other tasks related to supporting a config file-based approach (#458), and I think long lists of bootnode URLs should be included in the main YAML/TOML for the whole service instead of an independent config file format.

fuel-core/src/service/modules.rs
```diff
@@ -96,12 +99,29 @@ pub async fn start_modules(config: &Config, database: &Database) -> Result<Modul
     }
     txpool.start().await?;
 
+    let p2p_db: Arc<Box<dyn P2pDb>> = Arc::new(Box::new(database.clone()));
```
Voxelot, line 102: Why do we need the extra box inside the arc (`Arc<Box<dyn T>>`), shouldn't `Arc<dyn T>` suffice? Both Box and Arc allocate on the heap.

fuel-core/src/cli/run/p2p.rs
```diff
@@ -0,0 +1,150 @@
...
+    #[clap(long = "network")]
+    pub network: String,
+
+    /// p2p network's IP Address
+    #[clap(long = "address")]
+    pub address: Option<IpAddr>,
+
+    /// p2p network's TCP Port
+    #[clap(long = "port", default_value = "4000")]
+    pub port: u16,
```
Voxelot, line 28: This conflicts with the graphql port cli arg. Attempting to start fuel-core causes the following error:
```
thread 'main' panicked at 'Command fuel-core: Argument names must be unique, but 'port' is in use by more than one argument or group'
```

```suggestion
    #[clap(long = "peering-port", default_value = "4000")]
    pub port: u16,
```

fuel-p2p/src/orchestrator.rs
```diff
@@ -20,58 +19,58 @@ use crate::{
...
+            tx_outbound_responses,
             db,
-            outbound_responses: Default::default(),
-        })
+        }
     }
 
-    pub async fn run(&mut self) {
+    pub async fn run(mut self) -> anyhow::Result<Self> {
+        let mut p2p_service = FuelP2PService::new(self.p2p_config.clone()).await?;
```
Voxelot, line 64: We should have a bypass for running this service if p2p networking is disabled / not configured. fuel-core is heavily utilized via SDK integration tests for contract development, and we don't want to increase the cost of localized fuel-core testing with lots of libp2p connection setup and teardown.

fuel-core/src/service/modules.rs
```diff
@@ -61,25 +65,24 @@ pub async fn start_modules(config: &Config, database: &Database) -> Result<Modul
         .db(Box::new(database.clone()) as Box<dyn TxPoolDb>)
         .import_block_event(block_importer.subscribe());
 
-    let p2p_mpsc = ();
-    let p2p_broadcast_consensus = ();
-    let p2p_broadcast_block = ();
+    let (tx_request_event, rx_request_event) = mpsc::channel(100);
```
ControlCplusControlV, line 68: Is there a reason these channels are set at 100? I think this may have similar concerns as @Voxelot 's comment here https://github.com/FuelLabs/fuel-core/pull/491#discussion_r931550844

fuel-p2p/src/orchestrator.rs
```diff
@@ -110,26 +109,157 @@ impl NetworkOrchestrator {
...
+}
+
+pub struct Service {
+    join: Mutex<Option<JoinHandle<Result<NetworkOrchestrator, anyhow::Error>>>>,
+    network_orchestrator: Arc<Mutex<Option<NetworkOrchestrator>>>,
+    tx_request_event: Sender<P2pRequestEvent>,
+}
+
+impl Service {
+    pub fn new(
```
ControlCplusControlV, line 146: TxPool leverages a different patten for initialization, using a `ServiceBuilder` with an empty new that requires the calling of multiple setter functions before initializing the Service. Moving to something similar here could make it easier to pickup from across the codebase. https://github.com/FuelLabs/fuel-core/blob/master/fuel-txpool/src/service.rs#L24 an example of ServiceBuilder can be seen there

fuel-core/src/service/modules.rs
leviathanbeak: you're right, guess my mind was already on vacation

fuel-core/src/cli/run/p2p.rs
```diff
@@ -0,0 +1,150 @@
...
+    pub connection_keep_alive: Option<u64>,
+}
+
+impl From<P2pArgs> for anyhow::Result<P2PConfig> {
+    fn from(args: P2pArgs) -> Self {
+        let local_keypair = {
+            match args.keypair {
+                Some(path) => {
+                    let mut bytes = std::fs::read(path)?;
+                    Keypair::secp256k1_from_der(&mut bytes)?
```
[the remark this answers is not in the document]
digorithm: Yeah, the SDK and `fuel-crypto` have some primitives around keygen/mnemonics. Maybe we should use those here. 

SDK documentation for wallets/mnemonics: https://fuellabs.github.io/fuels-rs/v0.19.0/wallets/mnemonic-wallet.html

`fuel-crypto` documentation for `SecretKey` generation: https://docs.rs/fuels/latest/fuels/signers/fuel_crypto/struct.SecretKey.html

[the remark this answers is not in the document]
Voxelot: Seems like we should migrate some of the mnemonics logic to fuel-crypto, to avoid requiring the whole SDK for key parsing.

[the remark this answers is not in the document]
digorithm: Makes sense to me.

fuel-p2p/src/orchestrator.rs
leviathanbeak: good point, I think this should be somewhere where [this service is started](https://github.com/FuelLabs/fuel-core/pull/492/files#diff-9687cbe0598079f7848f4f130bdb462da74313456432c93874f3e8473fcda142R116) instead of the `run()` method itself.

Maybe having different `start_modules()` methods, like `start_with_network()` and without ?

fuel-core/src/cli/run/p2p.rs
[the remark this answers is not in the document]
leviathanbeak: great, should we then create a separate ticket for migrating that logic to fuel-crypto and another for incorporating it in a new cli command ?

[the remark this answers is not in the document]
digorithm: @Voxelot already created one for migrating the logic to `fuel-crypto`: https://github.com/FuelLabs/fuel-crypto/issues/27 

I guess we need another one for the latter like you said!

fuel-p2p/src/orchestrator.rs
Voxelot: I think it should be up to each module to decide whether it's going to run based on the module related config, rather than branching at the top level. So in this case, I think p2p should not run by default unless things like peering keys and bootstrap nodes are setup.

fuel-core/src/service/modules.rs
leviathanbeak: his concern was mainly because no one would be consuming those messages on the other end, in this case there is the consumer `rx_request_event` so it's fine.

fuel-p2p/src/orchestrator.rs
leviathanbeak: I don't disagree, but we already have [conditional module starting](https://github.com/FuelLabs/fuel-core/blob/46e952895ee32521f69262c5992af137ddba40a5/fuel-core/src/service/modules.rs#L97) so I figured we add another one for the network, instead of having a `start()` method on a module which would not always start actually.

fuel-p2p/src/orchestrator.rs
leviathanbeak: presently, I don't see much benefit of using a builder pattern just in order to initialize a service with it

fuel-p2p/src/orchestrator.rs
```diff
@@ -110,26 +109,157 @@ impl NetworkOrchestrator {
...
+                    validators: Default::default(),
+                },
+            }))
+        }
+    }
+
+    #[tokio::test]
+    async fn start_stop_works() {
+        let mut p2p_config = P2PConfig::default_with_network("start_stop_works");
+        p2p_config.tcp_port = 4018; // an unused port
```
bvrooman, line 237: nit: Would specifying port 0 to ask the OS for an available port be useful?

bvrooman (commented):
I know the convention for tokio channels is to name the senders and receivers `tx_` and `rx_` respectively, but I find that can be confusing, since we use `tx` to also mean a transaction. Do you find this naming convention effective?

fuel-p2p/src/orchestrator.rs
```diff
@@ -20,58 +19,58 @@ use crate::{
...
             tx_transaction,
+            tx_outbound_responses,
             db,
-            outbound_responses: Default::default(),
-        })
+        }
     }
 
-    pub async fn run(&mut self) {
+    pub async fn run(mut self) -> anyhow::Result<Self> {
```
bvrooman, line 63: Since this returns a Result, can we add test coverage for the success and failure cases?

bvrooman (dismissed):
Approved with some nits

fuel-p2p/src/orchestrator.rs
ControlCplusControlV: Using something like Portpicker would be beneficial here as when running multithreaded tests unless each port is individually assigned there can be collision

fuel-p2p/src/orchestrator.rs
ControlCplusControlV: The benefit of builder comes from not requiring parameters on intialization, I can see why you want to avoid it but we should probably decide at a higher level if we want the other modules to have it too for consistency

leviathanbeak: Sure, if it makes sense of not having a parameter available at initialization time or some similar case, but in this case there is no need for that. If it makes sense for another module to do it, that should not require the rest to the same if there is no need.

fuel-p2p/src/orchestrator.rs
leviathanbeak: ah this is a good point!, I will try and add this to all of my p2p tests

fuel-p2p/src/orchestrator.rs
leviathanbeak: The only case where this would fail is if the p2p configuration was incorrect. As for the success cases, the p2p service has been initialized and tested in dozens of p2p tests already.

leviathanbeak:
> I know the convention for tokio channels is to name the senders and receivers `tx_` and `rx_` respectively, but I find that can be confusing, since we use `tx` to also mean a transaction. Do you find this naming convention effective?

@bvrooman I'm pretty used to the pattern of naming channels like that, but I see what you mean, I wouldn't mind if we come up with a better prefix/suffix for our channel names.

ControlCplusControlV:
Sorry for re-running clippy showing the rust 1.63 errors, I needed a sanity check against my upstream branch since I wasn't sure if clippy changes had been moved to this branch yet. Once you can finish merge conflicts @leviathanbeak I can re-review and then we can get this moved as git history is starting to get very long

fuel-core/src/cli/run.rs
```diff
@@ -71,9 +75,18 @@ impl Command {
...
             relayer_args,
+            p2p_args,
         } = self;
 
         let addr = net::SocketAddr::new(ip, port);
+
+        let p2p = {
+            match p2p_args.into() {
+                Ok(value) => value,
+                Err(e) => return Err(io::Error::new(io::ErrorKind::Other, e)),
```
ControlCplusControlV, line 86: Any particular reason for `io::Error` over something like `anyhow::Error`?

fuel-core/src/cli/run/p2p.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// Choose min mesh size for gossipsub protocol
+    #[clap(long = "min_mesh_size", default_value = "4")]
+    pub min_mesh_size: usize,
+
+    /// Choose ideal mesh size for gossipsub protocol
+    #[clap(long = "ideal_mesh_size", default_value = "6")]
+    pub ideal_mesh_size: usize,
+
+    /// Choose timeout for sent requests in RequestResponse protocol
+    #[clap(long = "request_timeout")]
```
ControlCplusControlV, line 86: Would a good default `request_timeout` help? I think it may be good to have some conservative default just in case?

ControlCplusControlV (approved):
LGTM, nothing block on my end, just some questions.

fuel-core/src/cli/run/p2p.rs
Voxelot: usually 30s is a default for most load balancers, although we could shorten that if we expect nodes to respond faster.

Voxelot:
I'm starting to get concerned with the weight of all the extra dependencies being linked to fuel-core. Can we put the service under a feature flag, so that it's not compiled or included when `--no-default-features` is used when building? We should also do this for the relayer.

fuel-core/src/cli/run/p2p.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub connection_keep_alive: Option<u64>,
+}
+
+impl From<P2pArgs> for anyhow::Result<P2PConfig> {
+    fn from(args: P2pArgs) -> Self {
+        let local_keypair = {
+            match args.keypair {
+                Some(path) => {
+                    let mut bytes = std::fs::read(path)?;
+                    Keypair::secp256k1_from_der(&mut bytes)?
```
Voxelot, line 100: This should use the new fuel-crypto mnemonics to load/generate the secret key.

fuel-core/src/cli/run.rs
leviathanbeak: mainly because the return value of the function is `io::Result` not `anyhow::Result`

fuel-core/src/cli/run/p2p.rs
leviathanbeak: The default value for both `request_timeout` and `connection_keep_alive` was provided within the initialization of RequestResponse protocol, but I have now moved out, now it's with all other default values at `cli/run/p2p.rs`

fuel-core/src/cli/run/p2p.rs
leviathanbeak: I switched to using `fuel-crypto` for both loading a mnemonic and generating a random key,

since `libp2p`'s `Keypair` struct was needed for `p2pConfig` I have added a method to convert from `fuel-crypto`'s SecretKey into `libp2p`'s `SecretKey -> Keypair`

leviathanbeak:
ok, now the last thing is to added the service into optional features

leviathanbeak:
added `p2p` optional feature to `fuel-core`

fuel-core/Cargo.toml
```diff
@@ -87,3 +89,4 @@ prometheus = ["dep:prometheus"]
 default = ["rocksdb", "prometheus", "debug"]
 debug = ["fuel-core-interfaces/debug"]
 test-helpers = []
+p2p = []
```
Voxelot, line 92:
```suggestion
p2p = ["dep:fuel-p2p"]
```

Voxelot (approved):

ControlCplusControlV (approved):
clippy-ship-it

PR #501, opened 2022-07-23
Deunify build

ellioty:


.github/workflows/ci.yml
```diff
@@ -10,11 +10,12 @@ on:
     types: [published]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.ref }}
```
Voxelot, line 13: Is there a reason why we would want this workflow to cancel other unrelated workflows?

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow


```suggestion
  group: ${{ github.workflow }}-${{ github.ref }}
```

ellioty: I didn't mean to change this, apologies, reverted!

Voxelot:
The most recent run seems to have taken 47mins compared to 20mins on master, so I think we need to do some more investigating before merging this in.

Issue #500, opened 2022-07-23
Eliminate libssl dependency


Voxelot:
On a fresh Ubuntu install, users are reporting that fuel-core requires libssl to be installed. This is a regression, and libssl should be rooted out from the dependency tree if possible. This is likely due to the recent inclusion of Prometheus, but requires further analysis.

PR #518, opened 2022-07-30
fix openssl dep

Voxelot:
Integration tests were creating a dependency on openssl, preventing the repo from being tested without openssl installed on the system. This PR also also includes a new CI change to prevent openssl from being included in the future.

closes: #500 

Tested using a clean install of ubuntu in virtual box


ControlCplusControlV (approved):
Tested locally on a fresh docker image without openssl, works great

PR #532, opened 2022-08-04
Message verification in the block executor

### LINKED ISSUE #512: Properly verify message id existence and spend in the block executor
When utxo_validation is enabled, the block execution logic should verify the existence of bridge messages. The executor also needs to mark the message as spent to prevent reuse.

This work was deferred from the main task (#510) of upgrading to the latest versions of fuel-tx and fuel-vm, ~~and is blocked on #485~~ as well to check if the message exists in the database.

Test Cases:

- Block execution succeeds when a tx includes a message with message da_height <= block da_height
- Block execution fails if tx spends a message id that does not exist in db
- Block execution fails if tx spends a message id with da_height > block da_height
- Block execution fails if tx spends a message id that has already been spent

Dentosal:
Closes #512

### DIFF
--- a/fuel-core/Cargo.toml
+++ b/fuel-core/Cargo.toml
@@ -43,6 +43,7 @@ fuel-core-interfaces = { path = "../fuel-core-interfaces", version = "0.9.5", fe
 fuel-relayer = { path = "../fuel-relayer", version = "0.9.5" }
 fuel-sync = { path = "../fuel-sync", version = "0.9.5" }
 fuel-txpool = { path = "../fuel-txpool", version = "0.9.5" }
+fuel-types = "0.5"
 futures = "0.3"
 graphql-parser = "0.3.0"
 hex = { version = "0.4", features = ["serde"] }
--- a/fuel-core/Cargo.toml
+++ b/fuel-core/Cargo.toml
@@ -43,7 +43,6 @@ fuel-core-interfaces = { path = "../fuel-core-interfaces", version = "0.9.5", fe
 fuel-relayer = { path = "../fuel-relayer", version = "0.9.5" }
 fuel-sync = { path = "../fuel-sync", version = "0.9.5" }
 fuel-txpool = { path = "../fuel-txpool", version = "0.9.5" }
-fuel-types = "0.5"
 futures = "0.3"
 graphql-parser = "0.3.0"
 hex = { version = "0.4", features = ["serde"] }
--- a/fuel-core/src/executor.rs
+++ b/fuel-core/src/executor.rs
@@ -5,7 +5,10 @@ use crate::{
     tx_pool::TransactionStatus,
 };
 use chrono::Utc;
-use fuel_core_interfaces::common::fuel_tx::{CheckedTransaction, TransactionFee};
+use fuel_core_interfaces::{
+    common::fuel_tx::{CheckedTransaction, TransactionFee},
+    model::DaMessage,
+};
 use fuel_core_interfaces::{
     common::{
         fuel_asm::Word,
@@ -23,6 +26,7 @@ use fuel_core_interfaces::{
     },
     model::FuelBlockHeader,
 };
+use fuel_types::MessageId;
 use std::{
     error::Error as StdError,
     ops::{Deref, DerefMut},
@@ -117,11 +121,15 @@ impl Executor {
                 // validate transaction has at least one coin
                 self.verify_tx_has_at_least_one_coin_or_message(tx)?;
                 // validate utxos exist and maturity is properly set
-                self.verify_input_state(block_db_transaction.deref(), tx, block.header.height)?;
+                self.verify_input_state(
+                    block_db_transaction.deref(),
+                    tx,
+                    block.header.height,
+                    block.header.number,
+                )?;
                 // validate transaction signature
                 tx.validate_input_signature()
                     .map_err(TransactionValidityError::from)?;
-                // verify that brindge messages exist
             }
 
             // index owners of inputs and outputs with tx-id, regardless of validity (hence block_tx instead of tx_db)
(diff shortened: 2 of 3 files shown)


fuel-core/Cargo.toml
```diff
@@ -43,6 +43,7 @@ fuel-core-interfaces = { path = "../fuel-core-interfaces", version = "0.9.5", fe
 fuel-relayer = { path = "../fuel-relayer", version = "0.9.5" }
 fuel-sync = { path = "../fuel-sync", version = "0.9.5" }
 fuel-txpool = { path = "../fuel-txpool", version = "0.9.5" }
+fuel-types = "0.5"
```
Voxelot, line 46: This should be re-exported by `fuel-core-interfaces`. If we let these fuel-* deps proliferate into the `Cargo.toml` of all our crates, we increase the likelihood of a missed version bump. It also just makes a lot of extra work when it's time to upgrade these versions if we have to update in several places instead of just `fuel-core-interfaces`.

fuel-core/src/executor.rs
```diff
@@ -5,7 +5,10 @@ use crate::{
     tx_pool::TransactionStatus,
 };
 use chrono::Utc;
-use fuel_core_interfaces::common::fuel_tx::{CheckedTransaction, TransactionFee};
+use fuel_core_interfaces::{
+    common::fuel_tx::{CheckedTransaction, TransactionFee},
+    model::DaMessage,
+};
```
Voxelot, line 11:
```suggestion
use fuel_core_interfaces::{
    common::{fuel_tx::{CheckedTransaction, TransactionFee}, fuel_type::MessageId},
    model::DaMessage,
};
```

fuel-core/src/executor.rs
```diff
@@ -23,6 +26,7 @@ use fuel_core_interfaces::{
     },
     model::FuelBlockHeader,
 };
+use fuel_types::MessageId;
```
Voxelot, line 29:
```suggestion
```

fuel-core/Cargo.toml
Voxelot, line 46:
```suggestion
```

fuel-core/src/executor.rs
```diff
@@ -304,9 +318,22 @@ impl Executor {
...
+                        if message.fuel_block_spend.is_some() {
+                            return Err(TransactionValidityError::MessageAlreadySpent(*message_id));
+                        }
+                        if BlockHeight::from(message.da_height) > block_da_height {
+                            return Err(TransactionValidityError::MessageSpendTooEarly(
+                                *message_id,
+                            ));
+                        }
+                    } else {
+                        println!("NOTFOUND {message_id:?}");
```
Voxelot, line 333: We should use tracing instead of println's for logging messages. I'm suggesting we remove this message since the error should already have a trace/context.

```suggestion
```

fuel-core/src/executor.rs
```diff
@@ -345,57 +372,115 @@ impl Executor {
...
+                            .ok_or(Error::TransactionValidity(
+                                TransactionValidityError::MessageDoesNotExist(*message_id),
+                            ))?
+                            .da_height
+                    } else {
+                        // if utxo validation is disabled, just assignto the original block
+                        Default::default()
+                    };
+
+                    println!("INSERT {message_id:?}");
```
Voxelot, line 467:
```suggestion
```

fuel-core/src/executor.rs
```diff
@@ -345,57 +372,115 @@ impl Executor {
...
-        } else {
+    fn verify_tx_has_at_least_one_coin_or_message(&self, tx: &Transaction) -> Result<(), Error> {
+        if tx
+            .inputs()
+            .iter()
+            .any(|input| input.is_coin() || input.is_message())
+        {
             Ok(())
+        } else {
+            Err(TransactionValidityError::NoCoinInput(tx.id()).into())
```
Voxelot, line 383: This error should be updated to mention that there's no coin OR message.

fuel-core/src/executor.rs
```diff
@@ -304,9 +317,22 @@ impl Executor {
...
+                        if message.fuel_block_spend.is_some() {
+                            return Err(TransactionValidityError::MessageAlreadySpent(*message_id));
+                        }
+                        if BlockHeight::from(message.da_height) > block_da_height {
+                            return Err(TransactionValidityError::MessageSpendTooEarly(
+                                *message_id,
+                            ));
+                        }
+                    } else {
+                        println!("NOTFOUND {message_id:?}");
```
Voxelot, line 332: More printlns, plz cleanup

Voxelot (changes requested):
One remaining println

Voxelot (approved):

fuel-core/src/executor.rs
```diff
@@ -1594,4 +1686,183 @@ mod tests {
...
+            .finalize();
+
+        if let Input::MessageSigned { owner, .. } = tx.inputs()[0] {
+            message.owner = owner;
+        } else {
+            unreachable!();
+        }
+
+        (tx, message.check())
+    }
```
bvrooman, line 1721: Personally, I like to put helper functions at the top of the `tests` module in its own `helpers` submodule, i.e.: `tests::helpers::make_tx_and_message`. This way:
- When reading the code the test suite isn't interrupted by helper fns
- The test code makes it clear you're calling a test function, i.e. `helpers::make_tx_and_message(..)`

bvrooman (approved):

fuel-core/src/executor.rs
Dentosal: I like to keep code like this near the test cases that use it, so I'm keeping it here for now. I agree that more-generic functions should be in helpers module.

Issue #520, opened 2022-08-01
Add db column for `MessageOwner` -> `[MessageId]` where message owner is address or predicate hash


Voxelot:
This will enable the SDK to find message inputs to spend. This should largely be the same as `OWNED_COINS`.

Tests:

- [x] when a DaMessage is inserted using `RelayerDb::insert(MessageId, DaMessage)`, the MessageId is included in the set of ids associated with the owner.
- [x] when a DaMessage is removed using `RelayerDb::remove(MessageId)`, the MessageId is no longer included in the set of ids associated with the owner.

PR #533, opened 2022-08-04
add db column for message owner

leviathanbeak:
solves #520 

* added the column for storing the [MessageId]
* updated `insert()` and `remove()` methods so that MessageIds get persisted
* added `owned_message_ids()` query method to the Database
* added a test that verifies the functionality of the changes

### DIFF
--- a/fuel-core/src/database/message.rs
+++ b/fuel-core/src/database/message.rs
@@ -86,3 +86,39 @@ fn owner_msg_id_key(owner: &Address, msg_id: &MessageId) -> Vec<u8> {
         .copied()
         .collect()
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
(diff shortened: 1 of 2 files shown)


fuel-core/src/database/message.rs
```diff
@@ -13,12 +19,31 @@ impl Storage<MessageId, DaMessage> for Database {
         key: &MessageId,
         value: &DaMessage,
     ) -> Result<Option<DaMessage>, KvStoreError> {
+        // insert secondary record by owner
+        Database::insert(
```
Voxelot, line 23: It seems like we should do the primary insert of the record before the updating the index. Just in case something tries to query the index and the underlying data isn't persisted yet.

Voxelot (dismissed):

Voxelot (approved):

PR #540, opened 2022-08-09
consolidate schema interface

Voxelot:
Use optional owner to collapse nearly duplicate graphql interfaces.


fuel-core/src/database/message.rs
```diff
@@ -78,27 +78,14 @@ impl Database {
         })
     }
 
-    pub fn all_owners_and_message_ids(
+    pub fn all_messages(
         &self,
-        start: Option<OwnerAndMessageId>,
+        start: Option<MessageId>,
```
leviathanbeak, line 83: good one, I forgot that we could just query by ID instead of combing the owner and the ID

leviathanbeak (approved):
overall good improvement of the existing PR

bvrooman:
LGTM

Issue #539, opened 2022-08-08
implement time instruction in fuel-core


Voxelot:
depends on [fuel-vm v0.15](https://github.com/FuelLabs/fuel-vm/pull/194)

PR #545, opened 2022-08-10
current time impl for VM

Voxelot:
implement current time for the vm, also fixes an issue where the currently executing block details were not available in the vm.


bvrooman (approved):

Voxelot:
fixes #539

bvrooman (approved):

PR #550, opened 2022-08-10
Restructure configuration

Voxelot:
This PR moves genesis chainspec configuration back to being a top-level module, and moves "Config" from being a top level module to being a submodule of service. This is because Config is only useful in the context of starting a FuelService, while chain_config is consumed in a larger variety of ways (e.g. node snapshot tooling).

This also reduces the number of breaking changes when migrating from v0.9.x.


ControlCplusControlV (commented):
Looking at failing tests there ares still some files you missed, I did a quick grep and it showed

```
./fuel-core-bft/src/lib.rs:pub use config::Config;
./fuel-txpool/src/lib.rs:pub use config::Config;
./fuel-block-importer/src/lib.rs:pub use config::Config;
./fuel-sync/src/lib.rs:pub use config::Config;
./fuel-block-producer/src/lib.rs:pub use config::Config;
./fuel-block-executor/src/lib.rs:pub use config::Config;
./fuel-core/src/database/metadata.rs:use crate::service::config::Config;
./fuel-core/src/service/genesis.rs:    service::{config::Config, FuelService},
./fuel-core/src/service/genesis.rs:    use crate::service::config::Config;
./fuel-relayer/src/lib.rs:pub use config::Config;
./fuel-tests/tests/chain_conf.rs:use fuel_core::{config::Config, service::FuelService};
./fuel-tests/tests/debugger.rs:use fuel_core::{config::Config, service::FuelService};
```

Voxelot:
Those configs have the same name, but this refactor only applied to the one in fuel-core. I've pushed a commit which should address the current failure.

bvrooman (approved):
👍

PR #575, opened 2022-08-22
Add rustfmt rules to the project.

xgreenx:
Added rustfmt rules. I used my favorite rules but we can discuss your wishes=)
It requires usage of nightly rustfmt because it contains many new configurations.
Formatted the whole project, so it will make ongoing PR harder(you need to fix conflicts)
Added CI to check the fmt.


xgreenx:
You can check what the code looks like and decide whether it is okay or not=)
And maybe better to merge other PR first before this one.

.github/workflows/ci.yml
```diff
@@ -19,6 +19,23 @@ env:
   REGISTRY: ghcr.io
 
 jobs:
+  rustfmt:
+    concurrency:
```
Voxelot, line 23: Do we need job level concurrency settings here? Seems like this is superseded by the workflow-level concurrency setting on line 12.

xgreenx:
We actually can remove it because it is very fas check(~10 seconds) =)

.github/workflows/ci.yml
```diff
@@ -75,8 +92,6 @@ jobs:
     strategy:
```
Voxelot, line 92: we should make this job require rustfmt as a pre-requisite, otherwise CI will continue even if the formatter checks fail.

.rustfmt.toml
```diff
@@ -0,0 +1,64 @@
...
+reorder_imports = true
+reorder_modules = true
+reorder_impl_items = false
+type_punctuation_density = "Wide"
+space_before_colon = false
+space_after_colon = true
+spaces_around_ranges = false
+binop_separator = "Front"
+remove_nested_parens = true
+combine_control_expr = false # changed
```
vlopes11, line 31: mixed feelings with this one too since we will end with more lines of code

.rustfmt.toml
```diff
@@ -0,0 +1,64 @@
...
+overflow_delimited_expr = false
+struct_field_align_threshold = 0
+enum_discrim_align_threshold = 0
+match_arm_blocks = true
+force_multiline_blocks = true # changed
```
vlopes11, line 36: Ditto; is it desirable to enforce more LOC?

.rustfmt.toml
```diff
@@ -0,0 +1,64 @@
...
+fn_args_layout = "Tall"
+brace_style = "SameLineWhere"
+control_brace_style = "AlwaysSameLine"
+trailing_semicolon = false # changed
```
vlopes11, line 40: This one is very good

.rustfmt.toml
```diff
@@ -0,0 +1,64 @@
...
+trailing_comma = "Vertical"
+match_block_trailing_comma = false
+blank_lines_upper_bound = 1
+blank_lines_lower_bound = 0
+edition = "2021" # changed
+version = "One"
+merge_derives = true
+use_try_shorthand = true # changed
+use_field_init_shorthand = true # changed
```
vlopes11, line 49: Also another cool improvement!

vlopes11 (commented):
I think its a good thing to have a fmt standard. But then if we are diverging from the default of rustfmt, we might want to enforce that to all our repos so we don't end up having different styles across the org

.rustfmt.toml
```diff
@@ -0,0 +1,64 @@
+max_width = 90 # changed
+hard_tabs = false
```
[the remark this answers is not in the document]
xgreenx: Of course, we can do that, but if in the future Rust team decides on default values, it will affect us=)

.rustfmt.toml
```diff
@@ -0,0 +1,64 @@
...
+tab_spaces = 4
+newline_style = "Auto"
+use_small_heuristics = "Default"
+indent_style = "Block"
+wrap_comments = false
+format_code_in_doc_comments = false
+comment_width = 80
+normalize_comments = true # changed
```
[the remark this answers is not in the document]
xgreenx: Many fields from here are unstable=) It is why it is nightly rust fmt

.rustfmt.toml
```diff
@@ -0,0 +1,64 @@
...
+format_strings = false
+format_macro_matchers = false
+format_macro_bodies = true
+empty_item_single_line = true
+struct_lit_single_line = true
+fn_single_line = false
+where_single_line = false
+imports_indent = "Block"
+imports_layout = "Vertical" # changed
+imports_granularity = "Crate" # changed
```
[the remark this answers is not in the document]
xgreenx: Example from our codebase("Crate"):
изображение

Example from our codebase("Module"):
изображение


I'm okay with both variants but prefer more "Crate" because I can easy collapse imports from one crate or add a `cfg` flag for whole crate=)

.rustfmt.toml
xgreenx: I think the number of lines is not a problem. The readability of the code is more important=) With that flag each block of code will have its idents.

.rustfmt.toml
xgreenx: Again, better to write readable code(for reviewers) than try to put everything in one line with `if a > b { 24 } else { 42 }`. IDE will collapse it for you in code during navigation, so you will not see it

.rustfmt.toml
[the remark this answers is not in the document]
adlerjohn: Hmmm. Does this require users to switch between `nightly` and `stable` to run `cargo fmt`? The codebase itself builds with `stable`.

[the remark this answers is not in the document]
xgreenx: They need to do `cargo +nightly fmt`. But all IDE support a separate channel for rustfmt.
изображение

.rustfmt.toml
```diff
@@ -0,0 +1,64 @@
+max_width = 90 # changed
```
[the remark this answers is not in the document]
xgreenx: I'm using only a laptop=) The default value is 120. I didn't try 100 by I know that with 90, it always fits into GitHub split mode review=)

xgreenx:
> I think its a good thing to have a fmt standard. But then if we are diverging from the default of rustfmt, we might want to enforce that to all our repos so we don't end up having different styles across the org

Yea, if we are okay to apply rustfmt for `fuel-core` then better to do the same for all rust repos=)

adlerjohn:
If this requires an org-wide change then it should have org-wide buy-in. Might be time to follow-up on https://github.com/FuelLabs/rfcs/pull/8

.github/workflows/ci.yml
Voxelot: Specifically the "needs" section for `cargo-verifications` should include `rustfmt`

.rustfmt.toml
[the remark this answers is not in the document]
Voxelot: the cfg flag can go both ways. For example, it is also likely that a cfg flag will affect a subset of imports within a crate, rather than the entire crate. 


```
use crate::{ThingA, ThingB}
#[cfg(feature = "feature-c")]
use crate::ThingC
```

Will rustfmt avoid collapsing the import if we have a cfg flag on a specific set of imports from a crate?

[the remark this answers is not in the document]
xgreenx: Yes, it will not touch `cfg` imports. Also, if you have an "enter" between to import section, it will not merge them into one, but it will collapse and sort all inside each section(the same for `cfg`).

.rustfmt.toml
Voxelot: Just to iterate on the reviewability point a little more, more LOC seems like it would make diffs in github easier to follow, as the changes would be more specific and isolated.

Dentosal: I really dislike this one, especially for nested match blocks and single-expression closures. It hurts readibility quite a bit, since I'm accustomed to block having multiple expressions.

xgreenx: Better to not have nested match blocks=D It will force you to solve them

Voxelot: seems like this one isn't so popular, maybe we should revert to default here

.rustfmt.toml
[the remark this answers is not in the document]
leviathanbeak: the default value is 100 -> https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#max_width

leviathanbeak:
Reviewed all the config options, apart from maybe `imports_granularity = "Crate/Module"` I don't see much use in enforcing other ones, that is, diverging from the default values.

Voxelot (dismissed):
lgtm, but since this is a big change plz wait for at least one other approval before merging.

Dentosal (dismissed):
lgtm as well

Voxelot:
@adlerjohn we need to update the ci requirements on the repo to merge this pr

Voxelot:
@xgreenx can you also update the `CONTRIBUTING.md` with the updated instructions for running the nightly formatter?

Voxelot (approved):

Voxelot (approved):

Issue #517, opened 2022-07-29
Decompose DummyDb from fuel-core-interfaces into module specific dummy databases


Voxelot:
Right now we are overloading a global database test fixture that implements all the traits from all the different modules in the client. Each module already defines a database interface for the areas it's concerned about, and shouldn't require additional test data from other modules. Test data should be isolated to the module it is used in. Global reuse of test data may lead to unforeseen and difficult to solve side effects and unexpected behavior in unit tests.

For example, the txpool uses globally defined dummy transactions (TX_1, TX_2, etc) to fill the dummy db. This creates a lot of noise, which isn't scalable as more modules add test data, and generally increases the complexity of writing tests.

Eliminate DummyDb from fuel-core-interfaces, and migrate all related test data & setup to each respective crate module.

bvrooman:
Related:  https://github.com/FuelLabs/fuel-core/issues/561

bvrooman:
I am considering moving the txpool tests and `MockDb` to a separate `tests` folder under `fuel-txpool`. I think this would make code cleaner and more maintainable. In order to have proper reuse of `MockDb` in other areas of `fuel-txpool`, this may mean moving other unit test suites into `tests` as well, causing a cascade of changes. In the meantime, I've left `mock_db.rs` under `src` under a test cfg flag.

PR #576, opened 2022-08-23
feat: Update Relayer to use module level MockDb

bvrooman:
Related issues:
- https://github.com/FuelLabs/fuel-core/issues/517

This PR removes the Relayer tests' dependency on the existing `DummyDb` that comes pre-seeded, and uses a module level database, `MockDb` instead. `MockDb` does not come seeded with any data, and tests are expected to perform explicit setup. This means tests become more descriptive, self-documenting, and stable by removing sensitivity to data changes across shared test configurations. 

The Relayer module is also the last area where the `DummyDb` is used. Once this PR is merged, we can also remove the `DummyDb` entirely, allowing us to move closer towards a convention of explicit test setup.

### DIFF
--- a/fuel-relayer/src/mock_db.rs
+++ b/fuel-relayer/src/mock_db.rs
@@ -1,0 +1,227 @@
+use async_trait::async_trait;
+
+use std::borrow::Cow;
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+
+use fuel_core_interfaces::common::fuel_tx::Address;
+use fuel_core_interfaces::model::{
+    BlockHeight, ConsensusId, DaBlockHeight, SealedFuelBlock, ValidatorId, ValidatorStake,
+};
+use fuel_core_interfaces::relayer::{RelayerDb, StakingDiff, ValidatorSet};
+use fuel_core_interfaces::{
+    common::{fuel_storage::Storage, fuel_tx::MessageId},
+    db::KvStoreError,
+    model::Message,
+};
+
+#[derive(Default)]
+pub(crate) struct Data {
+    pub messages: HashMap<MessageId, Message>,
+    pub validators: HashMap<ValidatorId, (ValidatorStake, Option<ConsensusId>)>,
+    pub delegator_index: HashMap<Address, Vec<DaBlockHeight>>,
+    pub staking_diffs: HashMap<DaBlockHeight, StakingDiff>,
+
+    pub chain_height: BlockHeight,
+    pub sealed_blocks: HashMap<BlockHeight, Arc<SealedFuelBlock>>,
+    pub validators_height: DaBlockHeight,
+    pub finalized_da_height: DaBlockHeight,
+    pub last_committed_finalized_fuel_height: BlockHeight,
+}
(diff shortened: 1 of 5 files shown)


fuel-relayer/src/pending_blocks.rs
```diff
@@ -617,11 +641,27 @@ mod tests {
         let b2 = rng.gen();
 
         let mut blocks = block_commit(1u64.into());
-        let mut db = Box::new(DummyDb::filled());
+        let mut db = Box::new(MockDb::default());
         blocks.handle_block_commit(b1, 2u64.into(), 2, IsReverted::False);
         blocks.handle_block_commit(b2, 3u64.into(), 3, IsReverted::False);
         blocks.handle_block_commit(b2, 3u64.into(), 3, IsReverted::True);
 
+        db.tap_sealed_blocks_mut(|sealed_blocks| {
```
leviathanbeak, line 649: would it make sense to have a helper test function that takes N amount of blocks and creates them instead of repating this process in multiple tests?

fuel-relayer/src/mock_db.rs
```diff
@@ -0,0 +1,255 @@
...
+pub(crate) struct MockDb {
+    pub data: Arc<Mutex<Data>>,
+}
+
+impl MockDb {
+    pub fn tap_sealed_blocks_mut<F, R>(&mut self, func: F) -> R
+    where
+        F: Fn(&mut HashMap<BlockHeight, Arc<SealedFuelBlock>>) -> R,
+    {
+        func(&mut self.data.lock().unwrap().sealed_blocks)
```
leviathanbeak, line 46: I'd usually point too many `unwrap`s but since this is a MockDb, and unwraps are only assuming the mutex is not poisoned it should be fine

Voxelot (approved):

fuel-relayer/src/pending_blocks.rs
bvrooman: Personally, I prefer having the test setup localized to the test itself, even if it means repeated code. Extra helper functions make setup less localized, and may obfuscate test setup (even if it's just by a little bit). The spirit of this PR is to be explicit in the test setup and to increase code locality (by removing the DummyDb and its implicit setup) and I think that such an approach would be counter to that. 
I understand the suggestion, but personally would like to avoid it if I can!

fuel-relayer/src/mock_db.rs
bvrooman: Yes, it's not the most "proper" approach to data access, but we can make that assumption in tests. I do not think MockDb will be a production-ready replacement for RocksDB anytime soon, unfortunately (fortunately?).

Voxelot: Within the impl of the mock db itself, using `lock()/unwrap()` internally is fine. However, I think these tests will likely need to be re-written / redesigned since they are making assertions directly on the internals of the mockdb hashmaps instead of interacting with the relayerdb interface. I think it's fine to defer fixing these tests to a broader relayer cleanup initiative though.

PR #580, opened 2022-08-24
Applied the same formatting for all parts not handled by the `cargo +stable fmt`

xgreenx:
While we are thinking about [using non-standard](https://github.com/FuelLabs/fuel-core/pull/575) `rustfmt.toml`. 
We can apply the same formatting for parts not handled by the default rustfmt.

In most cases, it is sorting of imports, uncollapsing statements, add whitespace for comments=)

### DIFF
--- a/fuel-client/build.rs
+++ b/fuel-client/build.rs
@@ -1,8 +1,10 @@
 use schemafy_lib::{Expander, Schema};
-use std::env;
-use std::fs::{self, File};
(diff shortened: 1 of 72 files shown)


xgreenx:
We agreed on the usage of the nightly fmt. Right now, we need to decide which default value we want to use=)https://github.com/FuelLabs/fuel-core/pull/575

PR #581, opened 2022-08-25
Transaction selection logic for the block producer

Dentosal:
Working towards making #409 mergeable

### DIFF
--- a/fuel-block-producer/src/block_producer/transaction_selector.rs
+++ b/fuel-block-producer/src/block_producer/transaction_selector.rs
@@ -1,3 +1,5 @@
+use std::cmp::Reverse;
+
 use crate::Config;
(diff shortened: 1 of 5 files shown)


fuel-block-producer/src/block_producer/transaction_selector.rs
```diff
@@ -1,41 +1,164 @@
...
+                        Default::default(),
+                        Default::default(),
+                        0,
+                    )
+                    .add_output(Output::Change {
+                        to: Default::default(),
+                        amount: 0,
+                        asset_id: Default::default(),
+                    })
+                    .finalize(),
```
Voxelot, line 84: use finalized_checked() to simplify this conversion: https://github.com/FuelLabs/fuel-tx/blob/master/src/builder.rs#L211

Voxelot (approved):

Issue #568, opened 2022-08-17
Remove owner field from MessageInputs


Voxelot:
ref: https://github.com/FuelLabs/fuel-specs/issues/399

Remove the usage of the `owner` field on transaction input messages. It is being subsumed by `recipient`.

PR #582, opened 2022-08-25
Remove owner from InputMessage

leviathanbeak:
This is my take on #568, from my understanding of the task, I have added the following:

- `recipient` field of `InputMessage` takes value from `owner` 
- `recipient` value is **appended** to `data` field of `InputMessage`

### DIFF
--- a/fuel-client/src/client/schema/tx/tests/transparent_tx.rs
+++ b/fuel-client/src/client/schema/tx/tests/transparent_tx.rs
@@ -198,7 +198,6 @@ pub struct InputMessage {
     recipient: Address,
     amount: U64,
     nonce: U64,
-    owner: Address,
     witness_index: i32,
     data: HexString,
     predicate: HexString,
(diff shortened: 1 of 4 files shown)


ControlCplusControlV (commented):
LGTM, tests updated and `unsafe {}` block has safety comment. Defer to @Voxelot before merging though to ensure this accomplishes the task

fuel-client/src/client/schema/tx/tests/transparent_tx.rs
```diff
@@ -242,26 +241,37 @@ impl TryFrom<Input> for fuel_tx::Input {
                 contract_id: contract.contract.id.into(),
             },
             Input::InputMessage(message) => {
+                let data_len = message.data.0 .0.len();
+
+                let recipient = {
```
pixelcircuits, line 246: There shouldn't be a need for the client to get what was once referred to as "recipient" (the target fuel contract to send the message to). `fuel_tx::Input::MessageSigned` shouldn't have an owner field anymore and the recipient in this case should just be `recipient: message.recipient.into(),` and data should be left like it was before `data: message.data.into()`

fuel-core/src/schema/tx/input.rs
```diff
@@ -218,12 +213,15 @@ impl From<&fuel_tx::Input> for Input {
             } => Input::Message(InputMessage {
                 message_id: MessageId(*message_id),
                 sender: Address(*sender),
-                recipient: Address(*recipient),
+                recipient: Address(*owner),
```
pixelcircuits, line 216: Again, remove `owner` from `fuel_tx::Input::MessageSigned` and `fuel_tx::Input::MessagePredicate` too. It should just be a simple conversion to `Input::Message`

pixelcircuits:
This issue needs to wait for https://github.com/FuelLabs/fuel-tx/issues/179 to finish first

pixelcircuits:
So does fuel-core not check signatures yet? I would expect to see a code change where witness signatures are checked against what we once called "owner" for messages, but is now called "recipient"

Voxelot:
> So does fuel-core not check signatures yet? I would expect to see a code change where witness signatures are checked against what we once called "owner" for messages, but is now called "recipient"

The signature validation logic is located inside of fuel-tx. This needs to be update to use recipient instead of owner here: https://github.com/FuelLabs/fuel-tx/blob/2f46a0b750bed14b8ca26afbed7cbcd2625c0912/src/transaction/validation.rs#L66

fuel-client/src/client/schema/tx/tests/transparent_tx.rs
leviathanbeak: I see, it makes sense, it would've been helpful it this issue was tagged as blocked by `fuel-tx` issue, but I can fix it fast once that is done

fuel-client/src/client/schema/tx/tests/transparent_tx.rs
```diff
@@ -242,26 +241,37 @@ impl TryFrom<Input> for fuel_tx::Input {
...
+                    unsafe {
+                        // SAFETY: last 4 bytes are always recipient data
```
Dentosal, line 248: nit: safety comment should be just before the `unsafe` block, not inside it

Voxelot:
this should probably be closed and redone once the preceeding PRs in fuel-asm, fuel-tx, and fuel-vm are done.

leviathanbeak:
yeah this is obsolete

PR #583, opened 2022-08-25
Refactored usage of storage traits

xgreenx:
Refactored all tables to implement new [storage traits](https://github.com/FuelLabs/fuel-storage/pull/15)(Also refactored all tests). Created types for all tables with descriptions and re-used tables defined in the `fuel-vm`.

Replaced all `<Self as Storage<K, V>>::method` with `self.storage::<Table>().method`.
The same for `MerkleRootStorage` and tests.

Instead of constants for columns, use the `Column` enum to manage tables columns(Use `Column`
everywhere instead of `ColumnId`). In the future, we can use more advanced naming for fields,
instead of "column-{i}" because now we have names of cloumns=)

In some places, removed unnecessary usage of `Vec<u8>`(in `put`, `insert` etc methods).
Removed usage of `Clone` and `Copy` for some `value` in the same methods.

Prepared the code for the next refactoring:
- to simplify the implementation of storage-related traits
- to use only read/write traits in the places where it is really required
- performance improvements to work with types more optimal


fuel-core-interfaces/src/db.rs
```diff
@@ -1,7 +1,91 @@
...
+/// The storage table of coins. Each [`Coin`](crate::model::Coin) is represented by unique `UtxoId`.
+pub struct Coins;
+
+impl Mappable for Coins {
+    type Key = UtxoId;
+    type SetValue = Coin;
+    type GetValue = Self::SetValue;
+}
+
+/// The storage table of bridget from the Ethereum [`Message`](crate::model::Message)s.
```
Voxelot, line 43:
```suggestion
/// The storage table of bridged Ethereum [`Message`](crate::model::Message)s.
```

fuel-core-interfaces/src/model/coin.rs
```diff
@@ -11,6 +11,10 @@ use crate::{
 #[cfg(graphql)]
 use async_graphql::Enum;
 
+/// Represents the user's coin for some asset with `asset_id`.
+/// The `Coin` is either `CoinStatus::Spent` or `CoinStatus::Unspent`. If the coin is unspent,
+/// it can be used as an input to the transaction and can be spent up to the `amount`.
+/// After usage as an input of the transaction, the `Coin` become `CoinStatus::Spent`.
```
Voxelot, line 17:
```suggestion
/// After usage as an input of a transaction, the `Coin` becomes `CoinStatus::Spent`.
```

fuel-core-interfaces/src/db.rs
```diff
@@ -1,7 +1,91 @@
...
+impl Mappable for ValidatorsSet {
+    type Key = ValidatorId;
+    type SetValue = (ValidatorStake, Option<ConsensusId>);
+    type GetValue = Self::SetValue;
+}
+
+/// The storage table of relayer stacking diffs.
+pub struct StackingDiffs;
+
+impl Mappable for StackingDiffs {
```
Voxelot, line 83: stake, not stack.

```suggestion
/// The storage table of relayer staking diffs.
pub struct StakingDiffs;

impl Mappable for StakingDiffs {
```

fuel-core-interfaces/src/utils.rs
```diff
@@ -0,0 +1,18 @@
+pub mod signer {
```
Voxelot, line 1: This change seems pointless and unrelated to storage updates. I don't think we should be randomly stuffing things into `util` modules

fuel-core/src/database/coin.rs
```diff
@@ -122,7 +124,7 @@ impl Database {
         direction: Option<IterDirection>,
     ) -> impl Iterator<Item = Result<UtxoId, Error>> + '_ {
```
Voxelot, line 125: Maybe for a separate PR, seems like we should return the actual `Coin` since we do the work of fetching it from storage to check the spent status anyways:

```suggestion
    ) -> impl Iterator<Item = Result<Coin, Error>> + '_ {
```

xgreenx: It is already done in the https://github.com/FuelLabs/fuel-core/pull/591 =)

fuel-core/src/database.rs
```diff
@@ -280,9 +277,7 @@ impl Default for Database {
         {
             let tmp_dir = TempDir::new().unwrap();
             Self {
-                data: Arc::new(
-                    RocksDb::open(tmp_dir.path(), columns::COLUMN_NUM).unwrap(),
-                ),
+                data: Arc::new(RocksDb::default_open(tmp_dir.path()).unwrap()),
```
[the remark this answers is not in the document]
xgreenx: It is the same as before, but instead of a separate constant, we can use `Column::Count`. To get the number of columns.

fuel-core-interfaces/src/db.rs
```diff
@@ -1,7 +1,91 @@
...
+    type SetValue = [DaBlockHeight];
+    type GetValue = Vec<DaBlockHeight>;
+}
+
+/// The storage table of relayer validators set.
+pub struct ValidatorsSet;
+
+impl Mappable for ValidatorsSet {
+    type Key = ValidatorId;
+    type SetValue = (ValidatorStake, Option<ConsensusId>);
```
vlopes11, line 76: Maybe unrelated to this PR, but why is consensus id optional? Every validator must have a consensus id in order to participate in the validation.

fuel-core/src/database/message.rs
```diff
@@ -12,37 +12,51 @@ use crate::{
...
         value: &Message,
     ) -> Result<Option<Message>, KvStoreError> {
         // insert primary record
         let result =
-            Database::insert(self, key.as_ref(), columns::MESSAGES, value.clone())?;
+            Database::insert(self, key.as_ref(), Column::Messages, value.clone())?;
 
         // insert secondary record by owner
-        Database::insert(
+        let _: Option<bool> = Database::insert(
```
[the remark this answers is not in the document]
xgreenx: Because I changed the `insert` method of the `Database`, it now requires explicit type annotation.

image

fuel-core-interfaces/src/signer.rs
```diff
@@ -3,7 +3,7 @@ use async_trait::async_trait;
 use thiserror::Error;
 
 /// Dummy signer that will be removed in next pull request.
-/// TODO do not use.
+/// TODO: Do not use.
```
vlopes11, line 6: Unrelated to this PR as well, but isn't a bit weird that we have a public trait that should not be used? Shouldn't we make it private or something?

fuel-core-interfaces/src/db.rs
xgreenx: Maybe when we will have many validators in the future, only part of them will participate in the consensus. The other will idle, waiting for their consensus epoch.

fuel-core/src/database/message.rs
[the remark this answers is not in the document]
Voxelot: hmm, would it make sense to have these database methods also use the new mappable types? That way each column is associated with a mappable, and we don't have to painstakingly set these return types in every call this way?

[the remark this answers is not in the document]
xgreenx: We can, but we need to implement `Mappable` for semi-tables(like `(BlockHeight -> BlockHash)`).

Another problem is that `Database` works with other traits, like `Serialize` and `DeserializeOwned`. It is not a problem to add bounds, but all types used in the `Mappable::Key`, `Mappable::SetValue`, and `Mappable::GetValue` should implement them. It can't be true for all types and can cause problems in the future. But seems right now it is not a problem.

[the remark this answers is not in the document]
Voxelot: Hmm yeah on second thought this might be overly constraining, as tables like Metadata don't have a uniform set of types between keys and values.

fuel-core/src/database/storage.rs
```diff
@@ -0,0 +1,48 @@
...
+            Receipt,
+            UtxoId,
+        },
+    },
+    model::FuelBlockDb,
+};
+use fuel_txpool::types::ContractId;
+
+/// The table of blocks generated by Fuels validators.
+/// Right now, we have only that type of block, but we will support others in the future.
```
Voxelot, line 15: > Right now, we have only that type of block, but we will support others in the future.

Not sure what this comment means, are you talking about sealed blocks?

fuel-core/src/database/storage.rs
```diff
@@ -0,0 +1,48 @@
...
+pub struct Receipts;
+
+impl Mappable for Receipts {
+    /// Unique identifier of the receipt.
+    type Key = Bytes32;
+    type SetValue = [Receipt];
+    type GetValue = Vec<Receipt>;
+}
+
+// TODO: Add macro to define all common tables to avoid copy/paste of the code.
```
Voxelot, line 47: Not sure how this would look, since not every implementation is the same (e.g. sometimes updating several columns is needed).

xgreenx: I meant to use macro in cases where it is possible in receipts, contracts, staking diffs, etc.

fuel-core/src/database/storage.rs
xgreenx: We discussed that in the future, we plan to have the blocks from ethereum. But I'm not sure=) I made this assumption based on the current naming `FuelBlock`.

fuel-core-interfaces/src/db.rs
Voxelot: I don't think we should make the ConsensusId optional either. Otherwise it becomes very difficult to vote or participate in any sort of validator set activation changes if the consensus key isn't available.

fuel-core/src/database/storage.rs
Voxelot: we won't be storing actual eth blocks though, just the metadata like validator staking diffs or messages like we already do.

Voxelot (approved):
Overall very nice refactor! I have a few minor outstanding feedback items, but I don't think those are as urgent or blocking as getting fuel-core synced with the VM.

fuel-core/src/database/storage.rs
xgreenx: Then, maybe we can rename `FuelBlock` to `Block`?=) Because it is already in the `fuel-core`, it belongs to Fuel.

Voxelot: Yeah this is probably fine. Separate PR though.

fuel-core-interfaces/src/db.rs
Voxelot: nm this was already optional before this PR. We'll revisit all this once we actually start the PoS milestone.

PR #590, opened 2022-09-04
Connect P2p with TxPool

### LINKED ISSUE #477: Broadcast transactions to peers when they're inserted into txpool via GraphQL API

### LINKED ISSUE #478: Include valid broadcast transactions from peers into the local txpool

ControlCplusControlV:
Connects P2p with TxPool which should allow transactions inserted over the txpool of one node to be broadcasted to the others. Done in collaboration with @bvrooman but this has fallen stale so I had to re-impl over master since merge conflicts and git history had grown too messy

Closes #477 
Closes #478


ControlCplusControlV:
Seems a lot of tests are failing now, do you know why yet @bvrooman ?

bvrooman:
It appears that the difficulty we are encountering is caused by the different configurations created by the combinations of different feature flags, specifically the `"p2p"` and `"relayer"` feature flags. These enable or disable the P2P service and Relayer service respectively. When we disable services and channels based on flags, we must also be cognizant of how the senders and receivers of each channel is affected.  

I'm mapping out the channels in a [draw.io doc here](https://drive.google.com/file/d/1TfzA-7xRE960XL1mWdQEx3dFboUt2n__/view?usp=sharing). This map is not complete, and it may not be completely accurate thus far.

From this map, we can observe that certain channels will have a different set of senders/receivers depending on the services enabled. When running tests with all features enabled, including P2P and Relayer, tests pass. However, we can see that without P2P, some senders are removed from the system and some signals are never passed to receivers. This causes some channels to wait indefinitely causing timeouts.

We need to make sure that tests pass for all combinations of the matrix:

| P2P | Relayer |
|-----|---------|
| ✗ Disabled | ✗ Disabled |
| ✓ Enabled | ✗ Disabled |
| ✗ Disabled | ✓ Enabled |
| ✓ Enabled | ✓ Enabled |

ControlCplusControlV:
Will try and reconfigure so we can get all feature combos working, although realistically we just disable tx gossiping if both aren't enabled. Maybe we could do something like wrapping gossip under a meta feature?

Voxelot:
FYI - I don't have permission to open that draw.io link

I think we should try to find the simplest way to enable this for now without coming up with a perfect solution to the channels issue, since we have a pending initiative to redesign our service arch and reduce our reliance on channels where possible.

bvrooman:
In effect, it seems like the solution is to only enable the `incoming_tx_receiver.recv()` branch of the txpool service context when p2p is enabled. However, macros like `#[cfg(feature = "p2p")]` are not valid syntax once nested inside the `tokio::select!` macro. Right now, I am trying to determine a simple way to disable that branch when p2p is not enabled.

Alternatively, we can just make p2p a default feature, and we can worry about the enabling/disabling of p2p once we go through the rearchitecting of the services.

fuel-core/src/service/modules.rs
```diff
@@ -57,106 +63,121 @@ pub async fn start_modules(config: &Config, database: &Database) -> Result<Modul
...
+            p2p_request_event_receiver,
+            tx_consensus,
+            incoming_tx_sender,
+            block_event_sender,
+        )
+    };
+    #[cfg(not(feature = "p2p"))]
+    {
+        let keep_alive = Box::new(incoming_tx_sender);
+        Box::leak(keep_alive);
```
Voxelot, line 129: Do we also need to leak `block_event_sender` here? I guess the receiver isn't used yet, but it would help future proof the fuel-sync task.

bvrooman: Yea, we can certainly do that for `block_event_sender` and any other p2p channel where the receiver does not have owned senders.

Voxelot:
We should also have an integ test that actually uses p2p to verify the txs can gossip.

fuel-p2p/src/config.rs
```diff
@@ -102,7 +103,7 @@ impl P2PConfig {
             allow_private_addresses: true,
             enable_random_walk: true,
             connection_idle_timeout: Some(Duration::from_secs(120)),
-            topics: vec![],
+            topics: vec![NEW_TX_GOSSIP_TOPIC.into()],
```
leviathanbeak, line 106: I think others should be included by default as well, this was an oversight from my side

bvrooman: Updated

fuel-core/src/service/modules.rs
bvrooman: Updated

fuel-tests/tests/tx_gossip.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    let node_config = create_node_config_from_inputs(tx.inputs());
+    let node_one = FuelService::new_node(node_config).await.unwrap();
+    let client_one = FuelClient::from(node_one.bound_address);
+
+    let node_config = create_node_config_from_inputs(tx.inputs());
+    let node_two = FuelService::new_node(node_config).await.unwrap();
+    let client_two = FuelClient::from(node_two.bound_address);
+
+    tokio::time::sleep(Duration::new(3, 0)).await;
```
bvrooman, line 94: TODO: We can replace the `sleep` calls with polling for peers once the functionality is enabled. This will be tracked in a separate issue here: https://github.com/FuelLabs/fuel-core/issues/649.

fuel-tests/tests/tx_gossip.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    let result = client_one.submit(&tx).await.unwrap();
+
+    let tx = client_one.transaction(&result.0.to_string()).await.unwrap();
+    assert!(tx.is_some());
+
+    tokio::time::sleep(Duration::new(3, 0)).await;
```
bvrooman, line 101: TODO: Same as above; we will replace `sleeps` with polling once that is available.

fuel-txpool/src/service.rs
```diff
@@ -99,10 +161,13 @@ impl ServiceBuilder {
...
-    pub broadcast: broadcast::Sender<TxStatusBroadcast>,
     pub db: Arc<Box<dyn TxPoolDb>>,
-    pub receiver: mpsc::Receiver<TxPoolMpsc>,
-    pub import_block_events: broadcast::Receiver<ImportBlockBroadcast>,
+    pub txpool_receiver: mpsc::Receiver<TxPoolMpsc>,
+    pub tx_status_sender: broadcast::Sender<TxStatusBroadcast>,
+    pub import_block_receiver: broadcast::Receiver<ImportBlockBroadcast>,
+    pub incoming_tx_receiver: broadcast::Receiver<TransactionBroadcast>,
+    #[cfg(feature = "p2p")]
+    pub network_sender: mpsc::Sender<P2pRequestEvent>,
```
xgreenx, line 170: Maybe we can already introduce the traits we discussed on the call in this PR?

You can represent P2P as `Gossiper` and `TransactionProvider` traits. It will aggregate `incoming_tx_receiver` and `network_sender` into one field, making the code cleaner. It seems easy to do because TXPool already knows nothing about  P2P(based on the imports in `Cargo.toml`).

image

fuel-txpool/src/txpool.rs
```diff
@@ -145,11 +150,53 @@ impl TxPool {
...
+        txs: Vec<ArcTx>,
+    ) -> Vec<anyhow::Result<Vec<ArcTx>>> {
+        let mut res = Vec::new();
+        for tx in txs.iter() {
+            let mut pool = txpool.write().await;
+            res.push(pool.insert_inner(tx.clone(), db).await)
+        }
+        for (ret, tx) in res.iter().zip(txs.into_iter()) {
+            match ret {
+                Ok(removed) => {
```
xgreenx, line 168: In the current code, we check that transactions were in the transactions pool, but we don't check that transactions have been committed in the block.

With the introduction of P2P gossiping seems we need to do this check. Overwise, we will gossip about already committed transactions)

xgreenx (commented):
Brief review

fuel-txpool/src/txpool.rs
Voxelot: I think txpool cleanup should be deferred to the block importer/sync task. Without that in place, it doesn't make as much to implement this yet.

fuel-txpool/src/service.rs
Voxelot: I'd prefer to checkpoint this PR as it's been in progress for a long time, and then we can do channel / trait conversion in subsequent PRs.

fuel-txpool/src/service.rs
```diff
@@ -40,36 +48,75 @@ impl Default for ServiceBuilder {
...
+
+    pub fn incoming_tx_receiver(
+        &mut self,
+        incoming_tx_receiver: broadcast::Receiver<TransactionBroadcast>,
+    ) -> &mut Self {
+        self.incoming_tx_receiver = Some(incoming_tx_receiver);
+        self
+    }
+
+    #[cfg(feature = "p2p")]
```
Voxelot, line 106: could we remove this feature flag from txpool + block producer and inject a dummy noop impl of this sender when the central p2p feature is disabled from fuel-core? Otherwise these feature flags will start proliferating all over the place.

fuel-txpool/src/service.rs
```diff
@@ -111,13 +176,37 @@ impl Context {
...
+                    let txpool = txpool.clone();
                     let db = self.db.clone();
+                    let tx_status_sender = self.tx_status_sender.clone();
+
+                    tokio::spawn( async move {
+                        let txpool = txpool.as_ref();
+                        match new_transaction.unwrap() {
+                            TransactionBroadcast::NewTransaction ( tx ) => {
+                                let txs = vec!(Arc::new(tx));
+                                TxPool::insert(txpool, db.as_ref().as_ref(), tx_status_sender, txs).await
```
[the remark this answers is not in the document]
Voxelot: This shouldn't be done unless we change the current gossip logic to only broadcast to immediate peers (depth 1). Right now, any gossiped messages will be propagated to the whole network. We have other P2P tasks to investigate how to prevent nodes from re-gossiping invalid data.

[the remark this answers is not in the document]
Voxelot: https://github.com/FuelLabs/fuel-core/issues/595

fuel-txpool/src/txpool.rs
```diff
@@ -145,11 +150,53 @@ impl TxPool {
...
+                            status: TxStatus::SqueezedOut {
+                                reason: Error::Removed,
+                            },
+                        });
+                    }
+                    let _ = tx_status_sender.send(TxStatusBroadcast {
+                        tx: tx.clone(),
+                        status: TxStatus::Submitted,
+                    });
+                    let _ = network_sender
```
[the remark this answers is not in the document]
Voxelot: Maybe rather than complicating with a closure, this method could just call the one below, iterate over the `Vec<Result<Arc<Transaction>>>` return type and then trigger a broadcast for each successful one.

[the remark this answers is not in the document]
bvrooman: > Maybe rather than complicating with a closure, this method could just call the one below, iterate over the `Vec<Result<Arc<Transaction>>>` return type and then trigger a broadcast for each successful one.

We originally proposed this but we changed it in favour of a more "performant" approach, at the cost of the extra code. But I think this approach is the simplest and cleanest, and I prefer the cleaner approach to the performant approach in this case. 

The draw back is:
- Double the number of iterations
 
But the benefits are:
- No extra method or code duplication
- Easier to read and reason about code

fuel-txpool/src/service.rs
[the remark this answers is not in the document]
bvrooman: Yes, it may not be clear from a cold read, but this branch is executed when new transactions are received from gossip. In this case, we do not want to echo the same transaction back to the network with a broadcast.

fuel-txpool/src/service.rs
bvrooman: Yeah, I can try that out. The feature flag switches really do spread like rabbits once they're in there.

fuel-txpool/src/txpool.rs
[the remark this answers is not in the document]
ControlCplusControlV: We should probably atleast leave a note in the future though in case in future txpool optimizations we need areas to improve

fuel-txpool/src/service.rs
ControlCplusControlV: Should be fixed now

fuel-txpool/src/txpool.rs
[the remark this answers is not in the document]
bvrooman: Draft PR based on this branch: https://github.com/FuelLabs/fuel-core/pull/670

Voxelot (approved):
LGTM! :shipit: 

The service setup w/ channels looks painful, but luckily we have a plan to address this soon.

The PR in draft can come after this one.

Issue #586, opened 2022-08-29
Include messages in coinsToSpend API


Voxelot:
When a user wants to transfer some funds or pay for gas, they need to call the coinsToSpend API to obtain a set of spendable inputs to use. With the addition of messages as a new type of spendable input, the coinsToSpend API should be expanded to also include messages received from L1 in addition to coins.

The current coinsToSpend API uses the random-improve algorithm to minimize utxo dust. While spending messages can't remove dust, there may be cases where using a message instead of a coin can help prevent the creation of dust. Until we have better modeling for this, messages should just be treated interchangeably for coins as an extension of the coin set for both random & largest first based selection.

xgreenx:
Right now, the signature of the `coinsToSpent`:

```rust
struct SpendQueryElementInput {
    /// Asset ID of the coins
    asset_id: AssetId,
    /// Target amount for the query
    amount: U64,
}

type Asset = SpendQueryElementInput;

fn coinsToSpent(
    // Owner of the coins.
    owner: Address, 
    // The list of requested assets` coins with asset ids and the target
    // amount the user wants to reach. We allow several entries with
    // the same asset id.
    assets: Vec<Asset>, 
    // The maximum number of coins that we can return. 
    // It doesn't take into account that for some asset id `X`, 
    // we want to use `N` inputs, and for some asset id `Y`, 
    // we want to use `M` inputs.
    max_coins: Option<u64>,
    // The list of excluded `UtxoId`s during iterating over the coins.
    exclude: Option<Vec<UtxoId>>,
)
// The list of `Coin`s for **all* assets combined together and
// limited by `max_coins`.
-> Result<Vec<Coin>>
```

It is a simplified model for better understanding.

## Return type proposal

With the introduction of `Message`, the query should accept filtering by `MessageId`, and return both `Coin`s and `Message`s.

For return type, we can use `enum`. But how do we want to name the entity that holds `Coin`s and `Message`s? In the PR, I named it `Banknote` as the analog of paper banknotes in real life. But as Brandom notices "Hmm I wouldn't call it banknotes, as there's no bank and they aren't notes".

We can't use `Asset` because it is name for fungible tokens in the network. We can't use `Input` or `Output` because it is reserved for transactions.

Maybe someone has ideas on how better to name it=) As an alternative to `Banknote`, I propose `Currency`.

## Exclude Utxos proposal

[...]

PR #591, opened 2022-09-04
Add support of messages into `coinsToSpend`

xgreenx:
Created a new entity - `Resource`. It is an aggregator for the coins and messages. For more information, check the discussion at the https://github.com/FuelLabs/fuel-core/issues/586

The API looks like:
```rust
struct SpendQueryElementInput {
    /// Asset ID to spend
    asset_id: AssetId,
    /// Target amount for the query
    amount: U64,
    /// The maximum number of resources for selection.
    max: Option<U64>,
}

pub struct ExcludeInput {
    /// Utxos to exclude from the selection.
    utxos: Vec<UtxoId>,
    /// Messages to exclude from the selection.
    messages: Vec<MessageId>,
}

pub enum Resource {
    Coin(Coin),
    Message(Message),
}

fn resourcesToSpend(
    // The `Address` of the resources owner.
    owner: Address, 
    // The list of requested assets` resources with asset ids, `target` amount the user wants
    // to reach, and the `max` number of resources in the selection. Several entries with the
    // same asset id are not allowed.
    query_per_asset: Vec<SpendQueryElementInput>, 
    // The excluded resources from the selection.
    exclude: Option<ExcludeInput>,
)
// The list of spendable resources per asset from the query. The length of the result is
// the same as the length of `query_per_asset`. The ordering of assets and `query_per_asset`
// is the same.
-> Result<Vec<Vec<Resource>>>
```

Added some abstraction to remove logic duplicating. I want to reuse it in the other schemas like `balance`.
Added `schema::Scalar` for `MessageId`.
Added `message_id` for the `schema::Message`.

Closes https://github.com/FuelLabs/fuel-core/issues/586


ControlCplusControlV:
I see a lot of API changes in here, are you sure of how this affects downstream consumers of `fuel-core` yet?

xgreenx:
Do you ask about `coinsToSpent` method signature or changes also in the database?

`coinsToSpent` is an inevitable change because we want to work with different objects(coins, messages). Because it breaks API, we can make more changes to make the API clear. For example, return banknotes per asset, and pass `max_input` per asset.

The changes in the database it forcing people to change how they work with it to have more optimal code.

ControlCplusControlV:
> Do you ask about `coinsToSpent` method signature or changes also in the database?
> 
> `coinsToSpent` is an inevitable change because we want to work with different objects(coins, messages). Because it breaks API, we can make more changes to make the API clear. For example, return banknotes per asset, and pass `max_input` per asset.

Changing to `coinsToSpent` I agree is inevitable for the best, I am more referring to the changes on both endpoint name and in the GQL schema. Any breaking issues there should be identified and a fix prepped there before merging this though to avoid issue

xgreenx:
Before I will update the tests let's discuss the final query signature in the issue=)

fuel-client/src/client/schema/primitives.rs
```diff
@@ -109,8 +109,15 @@ fuel_type_scalar!(AssetId, AssetId);
 fuel_type_scalar!(ContractId, ContractId);
 fuel_type_scalar!(Salt, Salt);
 fuel_type_scalar!(TransactionId, Bytes32);
+
```
Voxelot, line 112: whitespace?

fuel-core/src/database/utils.rs
```diff
@@ -0,0 +1,218 @@
+use crate::{
```
Voxelot, line 1: I'd prefer to avoid a generic `utils` module, as it makes organization less clear and will just keep growing with potentially unrelated things. 

Since this is mostly about coins and messages, maybe rename this utils module into `resources` within the database instead?

fuel-core/src/coin_query.rs
```diff
@@ -1,549 +1,672 @@
...
+pub fn largest_first(
+    query: &AssetQuery,
+) -> Result<Vec<Resource<Coin, Message>>, CoinQueryError> {
+    let mut inputs: Vec<_> = query.unspent_resources().try_collect()?;
+    inputs.sort_by_key(|resource| Reverse(*resource.amount()));
+
+    let mut collected_amount = 0u64;
+    let mut resources = vec![];
+
+    for coin in inputs {
```
Voxelot, line 118:
```suggestion
    for resource in inputs {
```

fuel-core/src/database/utils.rs
```diff
@@ -0,0 +1,218 @@
...
+    borrow::{
+        Borrow,
+        Cow,
+    },
+    collections::HashSet,
+};
+
+/// At least required `target` of the query per asset's `id` with `max` resources.
+#[derive(Clone)]
+pub struct Asset {
```
Voxelot, line 37: This is a selection criteria, not an asset 🤔

fuel-core/src/database/utils.rs
```diff
@@ -0,0 +1,218 @@
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ResourceId {
+    Utxo(UtxoId),
+    Message(MessageId),
+}
+
+/// The primary type of spent or not spent resources(coins, messages, etc). The not spent resource
+/// can be used as a source of information for the creation of the transaction's input.
+#[derive(Debug)]
+pub enum Resource<C, M> {
```
Voxelot, line 175: Since the schema has its own analog to resource, why include these generics here?

fuel-core/src/database/coin.rs
```diff
@@ -90,16 +88,16 @@ impl Storage<UtxoId, Coin> for Database {
 }
 
 impl Database {
-    pub fn owned_coins(
+    pub fn owned_coins_utxos(
```
Voxelot, line 91: why add "utxos"? These aren't utxos unless we filter out the spent coins.

fuel-core/src/database/utils.rs
xgreenx: To avoid unnecessary `Copy` and `Clone`. The query to DB returns `Resource<Cow<Coin>, Cow<Message>>`. But the graphql query returns `Resource<Coin, Message>`.

fuel-core/src/database/coin.rs
xgreenx: What is better `owned_coins_utxo_ids` or `owned_coins_ids`?=)

fuel-core/src/database/utils.rs
```diff
@@ -0,0 +1,218 @@
...
+        Self {
+            id,
+            target,
+            max: max as usize,
+        }
+    }
+}
+
+#[derive(Default)]
+pub struct Exclude {
```
Voxelot, line 54: not db related, should probably live in `resources_to_spend.rs`

fuel-core/src/database/utils.rs
```diff
@@ -0,0 +1,218 @@
...
+                ResourceId::Utxo(utxo) => instance.utxos.insert(utxo),
+                ResourceId::Message(message) => instance.messages.insert(message),
+            };
+        }
+
+        instance
+    }
+}
+
+pub struct AssetQuery<'a> {
```
Voxelot, line 74: not db related, should probably live in `resources_to_spend.rs`

fuel-core/src/coin_query.rs
```diff
@@ -1,549 +1,672 @@
 use crate::{
```
Voxelot, line 1: Should we rename `coin_query` to `resource_query`?

xgreenx: Yes, forgot about it=)

fuel-core/src/database/utils.rs
Voxelot: > Since the schema has its own analog to resource

It looks like the schema doesn't rely on this enum, since it has it's own: https://github.com/FuelLabs/fuel-core/blob/feature/coins-to-spent-messages/fuel-core/src/schema/resource.rs#L52

Couldn't we just simplify this type by making it always a Cow?

xgreenx: No, because `Cow` requires lifetime from DB, and in the `resources_to_spend` we can't return `Vec<Vec<Resource<Cow, Cow>>>`=(

fuel-core/src/database/coin.rs
Voxelot: `owned_coins_ids`

fuel-core/src/state/rocks_db.rs
```diff
@@ -221,6 +221,8 @@ impl KeyValueStore for RocksDb {
             .iterator_cf_opt(&self.cf(column), opts, iter_mode)
             .map(|item| {
                 item.map(|(key, value)| {
+                    // TODO: Avoid copy of the slices into vector.
```
Voxelot, line 224: We could also adjust the API to make use of boxed slices instead of vec 🤔

fuel-core/src/database/utils.rs
Voxelot: Do we really need `Cow`? Or is it always owned once deserialized from the DB anyways? Even though the storage trait uses Cow, I don't think it would ever be borrowed unless we had some kind of zero-copy deserialization going on.

Voxelot:
Done reviewing for now, will wait for the requested changes before approving.

fuel-core/src/database/utils.rs
xgreenx: Yes, if we use RocksDB, in this case, it is useless=) I tried to be consistent with our API. `Database` returns `Cow`, and in some cases (maybe in the future), it can be useful. 

If you think it adds unjustified complexity, I can remove it=)

fuel-core/src/state/rocks_db.rs
xgreenx: We can, the question do we want to add complexity to database API=) But it will force us to work with raw slice where it is possible=)

Voxelot (approved):

fuel-core/src/database/utils.rs
Voxelot: eh it's not a big deal. I just prefer to find ways to hide generics when possible since it can get out of hand sometimes (like in substrate codebases lol)

fuel-core/src/schema/balance.rs
```diff
@@ -76,8 +76,9 @@ impl BalanceQuery {
     ) -> async_graphql::Result<Balance> {
         let db = ctx.data_unchecked::<Database>();
 
+        // TODO: Reuse [`AssetQuery`](crate::database::utils::AssetQuery) with messages
```
leviathanbeak, line 79: nit: would be nice to see Issues created for these `TODO`s (if not already)

leviathanbeak (dismissed):
LGTM, there are handful of TODOs, would be nice that Issues are created for and possibly referenced

ControlCplusControlV (approved):
LGTM, will move new `coinsToSpend` tests over to my rstest PR once merged. We may also want to look at organizing `fuel-tests` in a future issue as it is mostly flat right now and could benefit from tests being organized into folders

Voxelot (dismissed):

Voxelot (approved):

PR #594, opened 2022-09-05
Output Bech32Address for initial accounts

### LINKED ISSUE #593: While printing initial accounts, output `Bech32Address` so that it can be used out-of-box with `forc-{deploy, run}`
In a recent discussion with @adlerjohn over slack, we discussed the possibility of outputting `Bech32Address` as the address for the initial accounts so that they can be directly used in `forc-{deploy, run}`.

kayagokalp:
closes #593.


fuel-core/src/chain_config.rs
```diff
@@ -49,10 +53,14 @@ impl ChainConfig {
             .map(|_| {
                 let secret = fuel_core_interfaces::common::fuel_crypto::SecretKey::random(&mut rng);
                 let address = Address::from(*secret.public_key().hash());
+                let bech32_data = Bytes32::new(*address).to_base32();
```
adlerjohn, line 56: Wait, this is making a bech32 out of the address, not the pubkey. If that's the case (is it?), then we don't actually need this change, as a downstream tool can do this conversion.

That being said, it may be a worthwhile change regardless since all downstream utilities use bech32.

kayagokalp: After investigating how SDK handles this, I came up with this implementation. And yes it is created out of `Address`. I double checked if the output addresses are correct by signing a transaction using initial accounts. Btw to do so, I needed to add an option to `forc-wallet` sign to accept a private key (since initial accounts are not generated from `forc-wallet` and we do not have the phrase for them, we cannot import the wallet. Instead I used the private key printed at the initial output to sign the test transaction). 

This conversion can be done in downstream utilities as you mentioned but outputting this as the default would be good for signing transactions with initial accounts while using local clients. If that is not important we can close this PR too

fuel-core/src/chain_config.rs
```diff
@@ -14,6 +15,9 @@ use serde_with::{serde_as, skip_serializing_none};
 use serialization::{HexNumber, HexType};
 use std::{io::ErrorKind, path::PathBuf, str::FromStr};
 
+// Fuel Network human-readable part for bech32 encoding
+pub const FUEL_BECH32_HRP: &str = "fuel";
```
adlerjohn, line 19: Hmmm. Is there some way to import the SDK instead of duplicating code? If the SDK isn't set up for that, then we can merge as-is then have a follow-up `good first issue` to clean up.

kayagokalp: I think we can import from the SDK as it is currently declared as public but do we want to depend on the SDK from `fuel-core`? If that is acceptable, we will need to declare `fuels-types` as a dependency and after that we can fetch it from there

Voxelot: The bech32 encoding logic for Address should probably reside in fuel-types.

It's also worth noting that none of the fuel-core GraphQL APIs currently support bech32, so hijacking the output format here for the SDK would make them unusable for local client dev just using the graphql API directly.

We should decide whether base16 or bech32 is the standard encoding for addresses, since this is going to cause more confusion for the foreseeable future.

Voxelot: @kayagokalp for now, let's just return both hex and bech32 format for the address until we settle on a standard approach.

Voxelot (approved):

PR #597, opened 2022-09-08
Add a CI check to ensure Cargo.lock file remains up-to-date

mitchmindtree:
This adds a new CI "cargo-verification" check to ensure the `Cargo.lock` file is valid for the current state of all manifests within the workspace. This avoids having `Cargo.lock` files changes leak into unrelated PRs, and allows tools like Nix to always have a valid lock file present at each published release to ensure reproducible builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,6 +94,8 @@ jobs:
             args: --all-targets --all-features
           - command: make
             args: build
+          - command: check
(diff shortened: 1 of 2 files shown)


Dentosal (approved):
Cargo docs specifically say that this is a good idea: https://doc.rust-lang.org/cargo/commands/cargo-check.html#manifest-options

PR #598, opened 2022-09-08
fix prometheus flag

Voxelot:
When moving metrics out into their own crate, a feature flag for integration testing didn't get modified as needed. This was causing the following compilation warning:
```
warning: invalid feature `fuel-core/prometheus` in required-features of target `metrics_test`: feature `prometheus` does not exist in package `fuel-core v0.10.1 (/path/to/fuel-core/fuel-core)`
```

This PR fixes the warning and also restructures the metrics test to share the same test harness as other integ tests to avoid stressing the linker too much with multiple targets.

### DIFF
--- a/fuel-tests/Cargo.toml
+++ b/fuel-tests/Cargo.toml
@@ -20,7 +20,7 @@ harness = true
 [[test]]
 name = "metrics_test"
 path = "tests/metrics.rs"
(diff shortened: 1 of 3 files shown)


ControlCplusControlV (approved):
I didn't catch it my original PR since CI didn't catch it and I didn't notice warnings until after, so will be good to finally have this fixed

ControlCplusControlV (approved):

PR #599, opened 2022-09-09
Block production trigger

Dentosal:
Closes https://github.com/FuelLabs/fuel-core/issues/50

~~Requires #409 to be merged first~~

### DIFF
--- a/fuel-block-producer/src/block_producer/tests.rs
+++ b/fuel-block-producer/src/block_producer/tests.rs
@@ -36,7 +36,7 @@ async fn cant_produce_at_genesis_height() {
     let producer = ctx.producer();
 
     let err = producer
-        .produce_block(0u32.into())
+        .produce_block(0u32.into(), 1_000_000_000)
         .await
         .expect_err("expected failure");
 
@@ -53,7 +53,7 @@ async fn can_produce_initial_block() {
     let ctx = TestContext::default();
     let producer = ctx.producer();
 
-    let result = producer.produce_block(1u32.into()).await;
+    let result = producer.produce_block(1u32.into(), 1_000_000_000).await;
 
     assert!(result.is_ok());
 }
@@ -82,7 +82,9 @@ async fn can_produce_next_block() {
 
     let ctx = TestContext::default_from_db(db);
     let producer = ctx.producer();
-    let result = producer.produce_block(prev_height + 1u32.into()).await;
+    let result = producer
+        .produce_block(prev_height + 1u32.into(), 1_000_000_000)
+        .await;
 
     assert!(result.is_ok());
 }
@@ -94,7 +96,7 @@ async fn cant_produce_if_no_previous_block() {
     let producer = ctx.producer();
 
     let err = producer
-        .produce_block(100u32.into())
+        .produce_block(100u32.into(), 1_000_000_000)
         .await
         .expect_err("expected failure");
 
@@ -139,7 +141,7 @@ async fn cant_produce_if_previous_block_da_height_too_high() {
     let producer = ctx.producer();
 
     let err = producer
-        .produce_block(prev_height + 1u32.into())
+        .produce_block(prev_height + 1u32.into(), 1_000_000_000)
         .await
         .expect_err("expected failure");
 
@@ -170,7 +172,7 @@ async fn production_fails_on_execution_error() {
     let producer = ctx.producer();
 
     let err = producer
-        .produce_block(1u32.into())
+        .produce_block(1u32.into(), 1_000_000_000)
         .await
         .expect_err("expected failure");
 
--- a/fuel-block-producer/src/config.rs
+++ b/fuel-block-producer/src/config.rs
@@ -1,20 +1,6 @@
-use fuel_core_interfaces::common::{
-    fuel_tx::ConsensusParameters,
-    fuel_types::Word,
-};
+use fuel_core_interfaces::common::fuel_tx::ConsensusParameters;
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 pub struct Config {
-    pub max_gas_per_block: Word,
     pub consensus_params: ConsensusParameters,
-}
-
-impl Default for Config {
-    fn default() -> Self {
-        Self {
-            // TODO: pick a reasonable default value based on gas scheduling analysis
-            max_gas_per_block: 10 * ConsensusParameters::DEFAULT.max_gas_per_tx,
-            consensus_params: Default::default(),
-        }
-    }
 }
--- a/fuel-core/src/cli/run.rs
+++ b/fuel-core/src/cli/run.rs
@@ -3,6 +3,7 @@ use crate::{
     FuelService,
 };
 use clap::Parser;
+use fuel_chain_config::ChainConfig;
 use fuel_core::service::{
     Config,
     DbType,
@@ -108,11 +109,14 @@ impl Command {
             }
         };
 
+        let chain_conf: ChainConfig = chain_config.as_str().parse()?;
+        let consensus_params = chain_conf.transaction_parameters;
+
         Ok(Config {
             addr,
             database_path,
             database_type,
-            chain_conf: chain_config.as_str().parse()?,
+            chain_conf,
             utxo_validation,
             manual_blocks_enabled,
             vm: VMConfig {
@@ -124,11 +128,10 @@ impl Command {
             },
             predicates,
             block_importer: Default::default(),
-            block_producer: Default::default(),
+            block_producer: fuel_block_producer::Config { consensus_params },
             block_executor: Default::default(),
             #[cfg(feature = "relayer")]
             relayer: relayer_args.into(),
-            bft: Default::default(),
             sync: Default::default(),
             #[cfg(feature = "p2p")]
             p2p,
--- a/fuel-core/src/service/config.rs
+++ b/fuel-core/src/service/config.rs
@@ -31,7 +31,6 @@ pub struct Config {
     pub block_importer: fuel_block_importer::Config,
     pub block_producer: fuel_block_producer::Config,
     pub block_executor: fuel_block_executor::Config,
-    pub bft: fuel_core_bft::Config,
     pub sync: fuel_sync::Config,
     #[cfg(feature = "relayer")]
     pub relayer: fuel_relayer::Config,
@@ -54,7 +53,6 @@ impl Config {
             block_importer: Default::default(),
             block_producer: Default::default(),
             block_executor: Default::default(),
-            bft: Default::default(),
             sync: Default::default(),
             #[cfg(feature = "relayer")]
             relayer: Default::default(),
--- a/fuel-core/src/service/modules.rs
+++ b/fuel-core/src/service/modules.rs
@@ -1,5 +1,6 @@
 #![allow(clippy::let_unit_value)]
 use crate::{
+    chain_config::BlockProduction,
     database::Database,
     service::Config,
 };
@@ -7,6 +8,7 @@ use anyhow::Result;
 #[cfg(feature = "p2p")]
 use fuel_core_interfaces::p2p::P2pDb;
 use fuel_core_interfaces::{
+    common::prelude::Word,
     block_producer::BlockProducer,
     model::{
         BlockHeight,
@@ -32,7 +34,7 @@ pub struct Modules {
     pub txpool: Arc<fuel_txpool::Service>,
     pub block_importer: Arc<fuel_block_importer::Service>,
     pub block_producer: Arc<dyn BlockProducer>,
-    pub bft: Arc<fuel_core_bft::Service>,
+    pub coordinator: Arc<CoordinatorService>,
     pub sync: Arc<fuel_sync::Service>,
     #[cfg(feature = "relayer")]
     pub relayer: Option<fuel_relayer::RelayerHandle>,
@@ -45,7 +47,7 @@ impl Modules {
         let stops: Vec<JoinHandle<()>> = vec![
             self.txpool.stop().await,
             self.block_importer.stop().await,
-            self.bft.stop().await,
+            self.coordinator.stop().await,
             self.sync.stop().await,
             #[cfg(feature = "p2p")]
             self.network_service.stop().await,
@@ -58,14 +60,39 @@ impl Modules {
     }
 }
 
+pub enum CoordinatorService {
+    Poa(fuel_poa_coordinator::Service),
+    Bft(fuel_core_bft::Service),
+}
+impl CoordinatorService {
+    pub async fn stop(&self) -> Option<tokio::task::JoinHandle<()>> {
+        match self {
+            CoordinatorService::Poa(s) => s.stop().await,
+            CoordinatorService::Bft(s) => s.stop().await,
+        }
+    }
+}
+
 pub async fn start_modules(config: &Config, database: &Database) -> Result<Modules> {
     let db = ();
     // Initialize and bind all components
     let block_importer =
         fuel_block_importer::Service::new(&config.block_importer, db).await?;
     let block_producer = Arc::new(DummyBlockProducer);
-    let bft = fuel_core_bft::Service::new(&config.bft, db).await?;
     let sync = fuel_sync::Service::new(&config.sync).await?;
+
+    let coordinator = match &config.chain_conf.block_production {
+        BlockProduction::ProofOfAuthority { trigger } => CoordinatorService::Poa(
+            fuel_poa_coordinator::Service::new(&fuel_poa_coordinator::Config {
+                trigger: *trigger,
+                block_gas_limit: config.chain_conf.block_gas_limit,
+            }),
+        ),
+        // TODO: enable when btf config is ready to use
+        // CoordinatorConfig::Bft { config } => {
+        //     CoordinatorService::Bft(fuel_core_bft::Service::new(config, db).await?)
+        // }
+    };
 
     #[cfg(feature = "relayer")]
     let relayer = if config.relayer.eth_client.is_some() {
@@ -131,24 +158,43 @@ pub async fn start_modules(config: &Config, database: &Database) -> Result<Modul
 
     txpool_builder.network_sender(p2p_request_event_sender.clone());
 
-    let txpool = txpool_builder.build()?;
-
     // start services
 
     block_importer.start().await;
 
-    bft.start(
-        p2p_request_event_sender.clone(),
-        block_producer.clone(),
-        block_importer.sender().clone(),
-        block_importer.subscribe(),
-    )
-    .await;
+    match &coordinator {
+        CoordinatorService::Poa(poa) => {
+            // TODO: this must be connected to the block import mechanism
+            let (block_import_tx, mut block_import_rx) = broadcast::channel(16);
+            tokio::spawn(async move {
+                // Discard messages for now
+                while block_import_rx.recv().await.is_ok() {}
+            });
+            poa.start(
+                txpool_builder.subscribe(),
+                txpool_builder.sender().clone(),
+                block_import_tx,
+                block_producer.clone(),
+                database.clone(),
+            )
+            .await;
+        }
+        CoordinatorService::Bft(bft) => {
+            bft.start(
+                p2p_request_event_sender.clone(),
+                block_producer.clone(),
+                block_importer.sender().clone(),
+                block_importer.subscribe(),
+            )
+            .await;
+        }
+    }
 
     sync.start(
         block_event_receiver,
         p2p_request_event_sender.clone(),
-        bft.sender().clone(),
+        // TODO: re-introduce this when sync actually depends on the coordinator
+        // bft.sender().clone(),
         block_importer.sender().clone(),
     )
     .await;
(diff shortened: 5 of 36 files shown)


Voxelot:
FYI:

I retargeted this PR to diff against #409 

As far as configuration is concerned, we'll want a way to connect/consolidate with the `ProductionStrategy` chainspec configuration enum: https://github.com/FuelLabs/fuel-core/blob/v0.10.1/fuel-core/src/chain_config.rs#L107

Dentosal:
Ready for the initial round of comments

fuel-block-producer/src/block_producer/transaction_selector.rs
```diff
@@ -0,0 +1,173 @@
+use std::{
```
Voxelot, line 1: This file was moved to the TxPool adapter, as this logic should eventually move to txpool itself. I think this may have re-appeared due to a merge conflict.

bvrooman:
Can we upgrade tokio in a separate PR to reduce some of the noise? It would remove a handful of files from your PR.

fuel-poa-coordinator/src/service.rs
```diff
@@ -0,0 +1,321 @@
...
+                min_block_time,
+                max_tx_idle_time,
+            } => {
+                self.timer
+                    .set_timeout(max_block_time, OnConflict::Min)
+                    .await;
+
+                let block_tx_count =
+                    self.signal_produce_block().await?.transactions.len();
+                let includable_txs = self.txpool_sender.includable().await?.len();
```
bvrooman, line 186: nit: Just for the sake of consistency: 

```suggestion
                let block_tx_count =
                    self.signal_produce_block().await?.transactions.len();
                let includable_txs_count = self.txpool_sender.includable().await?.len();
```

fuel-poa-coordinator/src/service.rs
```diff
@@ -0,0 +1,321 @@
...
+                Trigger::Never | Trigger::Interval { .. } => Ok(()),
+                Trigger::Hybrid {
+                    max_tx_idle_time,
+                    min_block_time,
+                    ..
+                } => {
+                    // Create a new block just to see if we can fit them all in
+                    let block_tx_count =
+                        self.signal_produce_block().await?.transactions.len();
+                    let includable_txs = self.txpool_sender.includable().await?.len();
```
bvrooman, line 225: Same as above:

```suggestion
                    let block_tx_count =
                        self.signal_produce_block().await?.transactions.len();
                    let includable_txs_count = self.txpool_sender.includable().await?.len();
```

fuel-poa-coordinator/src/service.rs
bvrooman: Also, `includable()` returns the sorted txs. If you're only getting the count of the txs in the tx pool, and the order doesn't matter, you could get away with some sort of `size()` call on the `txpool_sender`, but that currently doesn't exist.

I would suggest maybe investigating to see if we can do a similar optimization to get the upcoming size of the block from the block producer without going through all the production steps.

fuel-block-producer/src/block_producer/transaction_selector.rs
```diff
@@ -0,0 +1,173 @@
...
+    // Select all txs that fit into the block, preferring ones with higher gas price.
+    //
+    // Future improvements to this algorithm may take into account the parallel nature of
+    // transactions to maximize throughput.
+    let mut used_block_space: Word = 0;
+
+    // Sort transactions by gas price, highest first
+    includable_txs.sort_by_key(|a| Reverse(a.transaction().gas_price()));
+
+    // Pick as many transactions as we can fit into the block (greedy)
```
ControlCplusControlV, line 25: Should we have some method to seperate `select_transaction` strategies? So either renaming it, or using some shared interface could be helpful. So that block producers can choose between `select_tx_gas_sorted` and `select_mev_tx_priority` etc, etc

fuel-block-producer/src/config.rs
```diff
@@ -1,20 +1,6 @@
...
+#[derive(Clone, Debug, Default)]
 pub struct Config {
-    pub max_gas_per_block: Word,
     pub consensus_params: ConsensusParameters,
 }
-
-impl Default for Config {
-    fn default() -> Self {
-        Self {
-            // TODO: pick a reasonable default value based on gas scheduling analysis
```
ControlCplusControlV, line 15: Was this TODO fixed / survived somewhere else?

fuel-core/src/service/config.rs
```diff
@@ -64,6 +66,19 @@ impl Config {
     }
 }
 
+/// Selects the method for coordinating block
+#[derive(Clone, Debug)]
+pub enum CoordinatorConfig {
+    /// Configurable block production trigger, used for
+    /// testing and proof-of-authority networksc
```
ControlCplusControlV, line 73: [nit] should `networksc` be `networksc`? I learned `iff` was a thing earlier this week so not sure if this is some abbreviation, if it is would be nice to be more clear

fuel-core/src/service/modules.rs
```diff
@@ -109,20 +137,37 @@ pub async fn start_modules(config: &Config, database: &Database) -> Result<Modul
...
-        tx_request_event.clone(),
-        block_producer.clone(),
-        block_importer.sender().clone(),
-        block_importer.subscribe(),
-    )
-    .await;
+    match &coordinator {
+        CoordinatorService::Poa(poa) => {
+            // TODO: this must be connected to the block import mechanism
+            let (block_import_tx, _block_import_rx) = broadcast::channel(16);
```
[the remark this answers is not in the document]
ControlCplusControlV: If not spinning up a separate thread to just receive like

```
tokio::spawn(async move { while (tx_status_receiver.recv().await).is_ok() {} });
```

With a TODO or issue to remove it somewhere

fuel-core/src/service/modules.rs
```diff
@@ -109,20 +137,37 @@ pub async fn start_modules(config: &Config, database: &Database) -> Result<Modul
...
+            .await;
+        }
+    }
 
     sync.start(
         rx_block,
         tx_request_event.clone(),
         relayer_sender,
-        bft.sender().clone(),
+        // bft.sender().clone(),
```
ControlCplusControlV, line 170: Is there a more clear way we can distinguish BFT is disabled because node is currently in PoA configuration

Dentosal:
See https://github.com/FuelLabs/fuel-core/pull/662 for tokio upgrade to simplify this one

fuel-block-producer/src/config.rs
Dentosal: It was moved to be an input argument to the trigger, see fuel-poa-coordinator/src/config.rs

fuel-core/src/service/modules.rs
```diff
@@ -54,19 +60,41 @@ impl Modules {
     }
 }
 
+pub enum CoordinatorService {
+    Poa(fuel_poa_coordinator::Service),
+    Bft(fuel_core_bft::Service),
+}
+impl CoordinatorService {
+    pub async fn stop(&self) -> std::option::Option<tokio::task::JoinHandle<()>> {
```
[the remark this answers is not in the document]
Dentosal: No, my rust-analyzer sometimes adds those without any need to. Removed.

fuel-poa-coordinator/src/service.rs
Dentosal: On thxpool side, we are sending a vector of `Arc`'d transactions and then taking count. The complier might be able to optimize this to just sending then len over.

fuel-block-producer/src/block_producer/transaction_selector.rs
Dentosal: This is outside the scope of this PR. Maybe you could open a new issue if that needs discussion?

fuel-block-producer/src/block_producer/transaction_selector.rs
Dentosal: Removed.

fuel-core/src/service/modules.rs
ControlCplusControlV: Just include in the commeent as to why `bft.sender().clone()` is commented out rather than not outright deleted

Dentosal: Sure

Voxelot:
Looks like we'll need a separate PR for enabling the real block producer in fuel-core and disabling the `executor.submit_txs` flow so that this poa coordinator is used instead.

Voxelot:
Also, have you tried running fuel-core via CLI to see what happens when txs are submitted? I'm not seeing any options on the CLI to configure the block production settings.

Dentosal:
I've added the production settings to ChainConfig. I'm not sure if they should instead be command line parameters for fuel-core instead, but chain config seems potentially better place to keep this.

Voxelot:
ChainConfig does make sense. The reason I was thinking of having CLI options to select the mode is that during development or debugging. It's a fairly common practice in Ethereum to fork the mainnet chain and then put it into instant mode for local testing. 

Given it's not that hard to edit the chainconfig json, I don't mind leaving the settings there for now.

fuel-core/src/cli/run.rs
```diff
@@ -128,7 +130,16 @@ impl Command {
             block_executor: Default::default(),
```
Voxelot, line 130:
```
            block_producer: fuel_block_producer::Config {
                consensus_params: chain_conf.transaction_parameters
            }
```

Dentosal:
This has some weird timeout issues after merging the latest master, as seen by the failed test run. Introduced here in commit 62b2826c2a4af257b84311dbfb3ad2fc48a14b82. I'll try to locate the offending change tomorrow.

fuel-poa-coordinator/src/service.rs
```diff
@@ -0,0 +1,326 @@
...
+        self.init_timers().await;
+        loop {
+            match self.process_next_event().await {
+                Ok(should_continue) => {
+                    if !should_continue {
+                        break
+                    }
+                }
+                Err(err) => {
+                    // TODO: is this the right way to handle an error here?
```
ControlCplusControlV, line 319: Error processing I think should be fleshed out more here. If `process_next_event` errs with a txpool error it might be reasonable to dump the txpool and try to re-populate it, while if there is an error with block production skipping the block or some other recourse like changing block production strategy may be a more desireable outcome. So maybe seperate out the err cases within here to catch both of those, then as a final exhaustive match panic

fuel-poa-coordinator/README.md
```diff
@@ -0,0 +1,3 @@
+# Fuel Core PoA Manager
+
+Coordinates block production in a PoA network. This is a stopgap implementation until `fuel-bft` can be used, i.e. when PoS network is usable.
```
ControlCplusControlV, line 3: Do we want PoS references in READMEs?

ControlCplusControlV (dismissed):
LGTM, don't see any block changes still. Await feedback from the others before merging though

freesig:
I'm getting this error when I try to run the tests in `fuel-poa-coordinator`:
image

fuel-block-producer/src/block_producer/tests.rs
```diff
@@ -53,7 +53,7 @@ async fn can_produce_initial_block() {
     let ctx = TestContext::default();
     let producer = ctx.producer();
 
-    let result = producer.produce_block(1u32.into()).await;
+    let result = producer.produce_block(1u32.into(), 1_000_000_000).await;
```
freesig, line 56: Seems like there should be a new test in that checks what happens if the included txs are > `max_gas`? Or at least showing that > `max_gas` txns never gets included. Or maybe it's just the wrong level to test that?

freesig (changes requested):
I'd like to be able to run the poa-coordinator tests before approving this. It appears they aren't running in CI as CI is passing?

Dentosal:
Waiting for https://github.com/FuelLabs/fuel-core/pull/688

fuel-poa-coordinator/README.md
Voxelot: There could also be scenarios where a PoA configuration is desired in the future for testing or other use-cases. So we should describe what this module accomplishes in a generic sense independently of our roadmap.

fuel-poa-coordinator/src/service.rs
```diff
@@ -0,0 +1,326 @@
...
+                min_block_time,
+                max_tx_idle_time,
+            } => {
+                self.timer
+                    .set_timeout(max_block_time, OnConflict::Min)
+                    .await;
+
+                // TODO: Maybe there could be a way to get the length without having
+                //       to perform full block production computation?
+                let block_tx_count =
```
Voxelot, line 186: Since we are moving the selection of transactions to the txpool, we shouldn't need to use the block producer at all here.

Following from the P&A architecture proposal, the POA module could define it's own trait for getting exactly what it wants from the txpool. Since the POA coordinator only caree about whether total max gas consumable by all the txs in the pool is > the max gas per block, it could have a port like:

```
trait TransactionPool {
    fn get_total_comsumable_gas(&self) -> u64;
}
```

And then use it here to check: `self.txpool.get_total_comsumable_gas() > config.max_gas_per_block()`.

This way, the Poa doesn't risk producing extra or empty blocks for no reason.

fuel-poa-coordinator/src/service.rs
```diff
@@ -0,0 +1,326 @@
...
+                Trigger::Never | Trigger::Interval { .. } => Ok(()),
+                Trigger::Hybrid {
+                    max_tx_idle_time,
+                    min_block_time,
+                    ..
+                } => {
+                    // Create a new block just to see if we can fit them all in
+                    // TODO: Maybe there could be a way to get the length without having
+                    //       to perform full block production computation?
+                    let block_tx_count =
```
Voxelot, line 227: See comment above - also this logical check on the txpool size vs block capacity should be deduplicated into a different function.

fuel-poa-coordinator/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+humantime-serde = "1.1.1"
+parking_lot = "0.12"
+serde = { version = "1.0", features = ["derive"] }
+tokio = { version = "1.21", features = ["full"] }
+tracing = "0.1"
+
+[dev-dependencies]
+async-trait = "0.1"
+fuel-block-producer = { path = "../fuel-block-producer" }
+fuel-core = { path = "../fuel-core" }
```
Voxelot, line 24: We should avoid depending on any other workspace members in these domain crates except for maybe `fuel-core-interfaces`.

Integration tests that fully depend on big dependencies like this should only exist in `fuel-tests`:

- This gives us assurances that the component actually works inside a real instance of fuel-core.
- The harness in fuel-tests is optimized to reduce the number of targets big dependencies like rocksdb must be linked into
- Changes to other crates like block producer or fuel-core won't drastically impact the structure of the domain-focused tests.


From what I can tell, the only reason we need fuel-core is to provide an impl of InterpreterStorage which has a method for returning the block height. However, the POA module shouldn't need to care about any of the other trait requirements of interpreter storage, as it never interacts with an instance of the VM.

The PoA orchestrator should have its own database trait for fetching the block height, and use a mock impl for testing instead of depending on fuel-core's implementation.

fuel-poa-coordinator/src/service.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    pub async fn start<S>(
+        &self,
+        txpool_broadcast: broadcast::Receiver<TxStatusBroadcast>,
+        txpool_sender: TxPoolSender,
+        import_block_events_tx: broadcast::Sender<ImportBlockBroadcast>,
+        block_producer: Arc<dyn BlockProducer>,
+        db: S,
+    ) where
+        S: InterpreterStorage + Send + Clone + 'static,
```
Voxelot, line 68: See comment above: https://github.com/FuelLabs/fuel-core/pull/599/files#r991677314

It's better to create traits with similar getters, than to reuse traits with many other useless things (from a PoA perspective) that make mocking difficult here.

fuel-poa-coordinator/src/service.rs
Dentosal: I'm not exactly sure what do you mean by repopulating the txpool here? I don't think we even do any fallible txpool calls anyway. My reasoning is that if we end up taking the error path here, it must already be a programming error somewhere. I would like to panic early instead of doing this, since it's hightly unlikely that any error can be recovered from.

fuel-block-producer/src/block_producer/tests.rs
Dentosal: I think this is out of scope for this PR, but such tests should be written anyway. I'll create an issue when this PR gets merged.

fuel-poa-coordinator/Cargo.toml
Dentosal: Removed fuel-core dev-dependency and factored block height query to a separate trait.  The block producer dependency is removed as well, by simply copying the imported function as-is, which is actually nice as the test of poa-coordinator shouldn't depend on internals of the block producer.

fuel-poa-coordinator/src/service.rs
Dentosal: Done

fuel-poa-coordinator/src/service.rs
Dentosal: Noted, continuing discussion in the item above

fuel-poa-coordinator/src/service.rs
Dentosal: I now have local version that does exactly this. However, introducing additional channels makes the test timings even harder, and now some of the test cases are flaky, The flakyness is rather hard to work around, for the fix see this tokio issue https://github.com/tokio-rs/tokio/issues/4522  (and also tokio PR 4523, for an earlier attempt to fix this)

Voxelot:
clippy is failing

fuel-poa-coordinator/src/deadline_clock.rs
```diff
@@ -0,0 +1,224 @@
...
+        self.control
+            .send(ControlMessage::Set {
+                deadline,
+                on_conflict,
+            })
+            .await
+            .expect("Deadline clock task has panicked");
+    }
+
+    /// Sets the tiemout, optionally overwriting the existing value
```
Voxelot, line 132: [nit]
```suggestion
    /// Sets the timeout, optionally overwriting the existing value
```

fuel-poa-coordinator/src/deadline_clock.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    /// Waits until the timeout expires. Sleeps forever when not timeout is set.
+    pub async fn wait(&mut self) -> Instant {
+        self.event
+            .recv()
+            .await
+            .expect("Deadline clock task has panicked")
+    }
+
+    /// Sets the tiemout, optionally overwriting the existing value
```
Voxelot, line 121: [nit]

```suggestion
    /// Sets the timeout, optionally overwriting the existing value
```

fuel-core/src/service/modules.rs
```diff
@@ -58,15 +60,40 @@ impl Modules {
...
     let sync = fuel_sync::Service::new(&config.sync).await?;
 
+    let coordinator = match &config.chain_conf.block_production {
+        BlockProduction::ProofOfAuthority { trigger } => CoordinatorService::Poa(
+            fuel_poa_coordinator::Service::new(&fuel_poa_coordinator::Config {
+                trigger: *trigger,
+                block_gas_limit: config.chain_conf.block_gas_limit,
+            }),
+        ),
+        // TODO: enable when btf config is ready to use
```
Voxelot, line 91:
```suggestion
        // TODO: enable when bft config is ready to use
```

fuel-poa-coordinator/src/service.rs
```diff
@@ -0,0 +1,316 @@
...
+                Trigger::Never | Trigger::Interval { .. } => Ok(()),
+                Trigger::Hybrid {
+                    max_tx_idle_time,
+                    min_block_time,
+                    ..
+                } => {
+                    let consumable_gas = self.txpool.total_consumable_gas().await?;
+
+                    // We have at least one transaction, so tx_max_idle_time is the limit
+                    self.timer
```
Voxelot, line 223: [nit] Should we move this to an else clause below and only update the timer if we don't immediately jump into block production?

Dentosal: Makes sense

fuel-poa-coordinator/tests/test_trigger.rs
```diff
@@ -0,0 +1,735 @@
...
+    // Stop
+    let handle = service.stop().await.expect("Get join handle");
+    txpool.stop().await?;
+    handle.await?;
+
+    Ok(())
+}
+
+#[tokio::test(start_paused = true)]
+async fn hybrid_trigger_reacts_correcly_to_full_txpool() -> anyhow::Result<()> {
```
Voxelot, line 680: [nit] 
```suggestion
async fn hybrid_trigger_reacts_correctly_to_full_txpool() -> anyhow::Result<()> {
```

Voxelot (dismissed):
very cool use of tokio time!

the hybrid trigger is the most complicated piece here and also has the least amount of tests. But we can cover this more extensively during the integration with fuel-core.

The only thing that may need to change based on the current arch discussion is that notifying the txpool about the new block should probably be done in the block importer. But since the importer doesn't exist yet, we can defer that.

Voxelot (approved):

fuel-block-producer/src/block_producer/tests.rs
Dentosal: https://github.com/FuelLabs/fuel-core/issues/695

PR #604, opened 2022-09-09
consolidate ci and optimize build commands

Voxelot:
In our CI we had both `cargo make build` and `cargo check --locked` which were doing basically the same things. This PR merges them together using a custom cargo make task. It also uses the cargo make gh action to avoid slower rust based caching.

It also includes fuel-p2p in the normal test pipeline since there hasn't been any flakiness lately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,7 @@ jobs:
           - command: clippy
             args: --all-targets --all-features
           - command: make
-            args: check --locked --workspace
+            args: check --locked
           - command: make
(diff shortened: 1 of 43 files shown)


ellioty (approved):

mitchmindtree (approved):
I've no experience with `cargo make` but that comment is a big help! Interesting stuff, I knew there with issues with cargo's feature unification but didn't realise they were that bad. It'll be especially good to keep this in mind w.r.t. potential feature support in sway/forc.

Anyways, Lgtm!

ellioty (approved):

PR #605, opened 2022-09-09
Reused imports from `fuel-vm` instead of direct import

xgreenx:
Reused import from the `fuel-vm` instead of direct importing into `Cargo.toml`. It should simplify the version bumping. And it is clear now that `fuel-vm` is the required submodule of `fule-core`.

I only did it for crates that imported `fuel-vm` previously.

Used explicit annotation "::fuel_vm::", to simplify changes like this in the future.


fuel-core-interfaces/src/block_importer.rs
```diff
@@ -3,7 +3,7 @@ use super::model::{
     FuelBlock,
     SealedFuelBlock,
 };
-use fuel_types::Bytes32;
+use fuel_vm::fuel_types::Bytes32;
```
Voxelot, line 6: It may look cleaner to use the common re-export instead of importing fuel-vm directly in every circumstance.

```suggestion
use crate::common::fuel_types::Bytes32;
```

ControlCplusControlV (commented):
This has been needed for a while so glad to see, but I would also agree that re-exporting would be cleaner over `::fuel_vm::` syntax

xgreenx:
I agree that better to use `crate::common::` for `fuel-core-interfaces`.
For `fuel-core` we will use `fuel_core_interfaces::common::`.
For `fuel-client` we can use only `fuel_vm::`

fuel-core-interfaces/src/lib.rs
```diff
@@ -10,18 +10,8 @@ pub mod sync;
...
-    #[doc(no_inline)]
-    pub use fuel_storage;
-    #[doc(no_inline)]
-    pub use fuel_tx;
-    #[doc(no_inline)]
-    pub use fuel_types;
     #[doc(no_inline)]
     pub use fuel_vm;
+    #[doc(no_inline)]
+    pub use fuel_vm::*;
```
Voxelot, line 16: do we really want everything from vm included here? This will include a lot more stuff into `common` than before.

xgreenx: Yeah, the question here do we want to not worry about potential future re-exports. Or we want to specify each export manually and update it manually?

Voxelot: Ideally we'd have a standard approach to bundling fuel-* usages (e.g. something like `common`), that way they are easier to re-export downstream. But this doesn't really hurt that much right now.

Voxelot (approved):

fuel-core-interfaces/Cargo.toml
```diff
@@ -15,13 +15,7 @@ anyhow = "1.0"
...
 chrono = { version = "0.4" }
 derive_more = { version = "0.99" }
-fuel-asm = "0.9"
-fuel-crypto = { version = "0.6", default-features = false, features = [ "random" ] }
-fuel-merkle = { version = "0.3" }
-fuel-storage = "0.2"
-fuel-tx = { version = "0.19", default-features = false }
-fuel-types = { version = "0.5", default-features = false }
-fuel-vm = { version = "0.16", default-features = false }
+fuel-vm = { version = "0.16", default-features = false, features = ["random"] }
```
Voxelot, line 18: hmm this will enable lots of random functions we may not want to allow access to in production code 🤔 

As a tech-debt follow-up, we'll need to figure out a different way to solve the need to use random in fuel-crypto, without enabling it in everything else.

ControlCplusControlV (approved):
LGTM

PR #609, opened 2022-09-11
Integrate `rstest` into more tests

ControlCplusControlV:


fuel-tests/tests/coin.rs
```diff
@@ -62,8 +62,12 @@ async fn coin() {
     assert!(coin.is_some());
 }
 
+// Backward fails, this is another issue
```
Voxelot, line 65: do we have an issue opened for this?

fuel-tests/tests/coin.rs
```diff
@@ -219,15 +225,16 @@ async fn only_unspent_coins() {
     assert!(coins
         .results
         .into_iter()
-        .all(|c| c.status == SchemeCoinStatus::Unspent));
+        .all(|c| asset_id == c.asset_id.into()));
```
Voxelot, line 228: The test is called `only_unspent_coins`, and this was the only assertion where the spent status is checked. Why is this check being removed?

fuel-tests/tests/coin.rs
ControlCplusControlV: #610

fuel-tests/tests/coin.rs
ControlCplusControlV: That shouldn't have been removed, I've since added it back in. Thank you for catching it

fuel-tests/tests/coin.rs
```diff
@@ -222,12 +229,13 @@ async fn only_unspent_coins() {
...
 }
 
+#[rstest]
 #[tokio::test]
-async fn coins_to_spend() {
-    let owner = Address::default();
-    let asset_id_a = AssetId::new([1u8; 32]);
-    let asset_id_b = AssetId::new([2u8; 32]);
-
+async fn coins_to_spend(
```
xgreenx, line 234: It will add a headache to my pull request=D Because I split it [into small subtests and added support for messages](https://github.com/FuelLabs/fuel-core/blob/feature/coins-to-spent-messages/fuel-tests/tests/resource.rs#L29)

xgreenx (dismissed):

fuel-tests/tests/coin.rs
Voxelot: If this was majorly redone, we should leave the coins_to_spend test as is in this PR then.

ControlCplusControlV: @xgreenx would you prefer if I keep this test alone then for this PR?

xgreenx: I prefer to avoid conflicts=D But as you wish

ControlCplusControlV:
@xgreenx looks like you avoided the merge conflicts, this should be good to merge now

xgreenx (approved):
I implemented `rstest` for `resources_to_spend`, so PR requires approval from someone else=)

Voxelot (approved):
Seems a little strange to have each resource test as a private fn called from an actual test fn, but each test case is still succinct and clean.

Issue #526, opened 2022-08-02
VM benchmark framework


Voxelot:
- [x] Setup a benchmark script in “fuel-core/fuel-tests/benches” using Criterion for each opcode in isolation, running against a full node on RocksDB
  - [x] Criterion Harness using a sync executor (https://github.com/FuelLabs/fuel-core/pull/616)
  - [x] Opcode Benchmarks
      - For each opcode being tested, repeat inside the script a large number of times to ensure that the vm is being exercised a relatively large amount of time compared to the rest of the node
      - For instructions that can’t be measured in isolation (for example CALL), exclude the benchmarked times of the dependent opcodes used for setup.
      - For opcodes with variable execution time depending on the arguments (e.g. cost of MCP is based on the number of bytes being copied), we should parameterize a range of values to establish a linear fee model (constant ADD, linear MCP/MCL, logarithmic via merklized storage access) 

- [ ] #667
  - Use NOOP as the base unit of cost (1)
  - All other opcodes are priced based on how many times slower they are compared to NOOP
  - Should be a rerunnable script that automatically processes the output from the benchmark harness
- [ ] Target some kind of low-medium cpu ec2 instance type (i.e. the minimum requirements for a validator)
  - Set block size based on target TPS after opcodes are priced

### Future Work

- Consider disabling VM instruction inlining to enable real-time prod monitoring of opcodes using perf
- Investigate how much these relative opcode costs change based on different machine hardware

ControlCplusControlV:
I am assuming we would use `cargo bench` for this, do we also want to modify configs or added a cpu limiting programmer for capped performance on machines to adjust for background usage when benchmarking? (I remember a Go optimization talk on Slack stickers noticeably impacting benchmarking).

We could also clone the sway-applications repo to run that repo's tests against `fuel-core` to get a real world benchmark, then have a seperate set of tests dedicated to trying to find DOS vectors (like 17000 NOPs to see if runtime is mostly in line with gas costs)

Voxelot:
`cargo bench` is typically for micro-benchmarking use-cases. I'm thinking we may need to setup a custom harness with it's own CLI for this 🤔

PR #616, opened 2022-09-12
Add `fuel-benches`

vlopes11:
This commit introduces the benchmarks for fuel-vm. It covers the first phase of #526 and targets individual instructions with the rocksdb/database backend.

This targets gas instrumentalization, so we skip benchmarking for predicate cases.

This uses a dedicated component of fuel-core so it will be easier to isolate and extend this component into block checking, and maybe to add some script to parse the output from criterion.

The benchmark is created from an isolated state of the VM that is ready to execute a given instruction. It is sub-optimal in the case of contract state & call since the call overhead will always be added to any instruction to be checked inside of a contract (i.e. SWW).

As next step, we need to make an API in fuel-vm to allow users to perform the operations of a call without executing the contract code. In other words, just load the contract code, add the call frame to the stack and prepare the registers.

After this phase is complete, we need to benchmark beyond instructions to a block level so the post-execution steps (database commit, etc) is also accounted.


fuel-benches/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+            .with_output(output)
+            .with_prepare_script(prepare_script))
+    }
+
+    pub fn with_db(mut self, db: Database) -> Self {
+        self.db.replace(db);
+        self
+    }
+
+    pub fn with_params(mut self, params: ConsensusParameters) -> Self {
```
Dentosal, line 137: nit: I dislike having setters like this for public fields, especially when there is no need to use builder pattern.

Dentosal (approved):

vlopes11:
Blocked by https://github.com/FuelLabs/fuel-core/pull/583

fuel-benches/src/lib.rs
vlopes11: The idea of these are to avoid having to declare every and each attribute for tests that requires only one attribute (ie instruction).

wdyt?

Dentosal: I feel like I'm missing something here, what's wrong with this?

```rust
let mut vm_bench = VmBench::new(Opcode::NOP);
vm_bench.params = params;
self.gas_price = gas_price;
```

Voxelot: Seems like it's mainly to enable defining benches inline with the criterion run fn's, such as:

```rust
    run_group(
        &mut group,
        "mlog",
        VmBench::new(Opcode::MLOG(0x10, 0x11, 0x12)).with_prepare_script(vec![
            Opcode::MOVI(0x11, 100000),
            Opcode::MOVI(0x12, 27),
        ]),
    );
```

Although the same thing could be accomplished with:

```rust
    run_group(
        &mut group,
        "mlog",
        VmBench {
           prepare_script: vec![
             Opcode::MOVI(0x11, 100000),
             Opcode::MOVI(0x12, 27),
           ])
           ..Self::new(Opcode::MLOG(0x10, 0x11, 0x12))
        }
    );
```

Voxelot (approved):

ControlCplusControlV (approved):
LGTM

PR #617, opened 2022-09-13
calculate message id directly from fuel-tx

### LINKED ISSUE #602: dedupe message id calculation
There is message ID calculation logic located inside of fuel-core that duplicates the logic in fuel-tx.

fuel-core: https://github.com/FuelLabs/fuel-core/blob/v0.10.1/fuel-core-interfaces/src/model/messages.rs#L23
fuel-tx: https://github.com/FuelLabs/fuel-tx/blob/v0.19.0/src/transaction/types/input.rs#L479

fuel-core should rely on the canonical implementation from fuel-tx to avoid the risk of deviating in implementation.

leviathanbeak:
closes #602

### DIFF
--- a/fuel-core-interfaces/src/model/messages.rs
+++ b/fuel-core-interfaces/src/model/messages.rs
@@ -1,12 +1,11 @@
+use fuel_vm::prelude::Input;
+
 use super::BlockHeight;
 use crate::{
-    common::{
-        fuel_crypto::Hasher,
-        fuel_types::{
-            Address,
-            MessageId,
-            Word,
-        },
+    common::fuel_types::{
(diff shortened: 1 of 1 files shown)


fuel-core-interfaces/src/model/messages.rs
```diff
@@ -1,12 +1,11 @@
+use fuel_vm::prelude::Input;
```
xgreenx, line 1: Could you use `crate::common::fuel_vm::prelude::Input`? It will collapse everything into one import

xgreenx (approved):

Voxelot (approved):

PR #624, opened 2022-09-18
 Refactor `FuelBlockHeader::number` to name `da_height` and type `DaBlockHeight` 

### LINKED ISSUE #516: Refactor `FuelBlockHeader::number` to name `da_height` and type `DaBlockHeight`

ControlCplusControlV:
Closes #516 I ended up using u64 because it played much more nicely with the existing changes and I had misinterpreted the issue when I initially asked in that issue.

Migrates `BlockHeight` instances where `DaBlockHeight` should've been used over

### DIFF
--- a/fuel-core-interfaces/src/model/da_block_height.rs
+++ b/fuel-core-interfaces/src/model/da_block_height.rs
@@ -1,0 +1,82 @@
+use derive_more::{
+    Add,
+    Deref,
+    Display,
+    From,
+    Into,
+};
+
+use std::ops::{Add, Sub, Rem};
+
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(
+    Copy,
+    Clone,
+    Debug,
+    Default,
+    PartialEq,
+    PartialOrd,
+    Eq,
+    Add,
+    Ord,
+    Display,
+    Into,
+    From,
+    Deref,
+    Hash,
+)]
+pub struct DaBlockHeight(pub u64);
+
+impl From<DaBlockHeight> for Vec<u8> {
+    fn from(height: DaBlockHeight) -> Self {
+        height.0.to_be_bytes().to_vec()
+    }
+}
+
+impl From<usize> for DaBlockHeight {
+    fn from(n: usize) -> Self {
+        DaBlockHeight(n as u64)
+    }
+}
+
+impl Add<u64> for DaBlockHeight {
+    type Output = Self;
+
+    fn add(self, other: u64) -> Self::Output {
+        Self::from(self.0 + other)
+    }
+}
+
+impl Sub for DaBlockHeight {
+    type Output = Self;
+
+    fn sub(self, other: Self) -> Self::Output {
+        Self::from(self.0 - other.as_u64())
+    }
+}
+
+impl Rem for DaBlockHeight {
+    type Output = Self;
+
+    fn rem(self, other: Self) -> Self::Output {
+        Self::from(self.0 % other.as_u64())
+    }
+}
+
+impl DaBlockHeight {
+    pub fn to_bytes(self) -> [u8; 8] {
+        self.0.to_be_bytes()
+    }
+
+    pub fn to_usize(self) -> usize {
+        self.0 as usize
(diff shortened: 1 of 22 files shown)


ControlCplusControlV:
Also had to move `DaBlockHeight` from a pure `u64` so I can add some implementations and to better mirror `BlockHeight`

bvrooman:
I wanted to pose this as an open question: Can we replace `DaBlockHeight` with `BlockHeight`? I think the way we use `DaBlockHeight` is already parallel to `BlockHeight` - the difference is simply the source of the data (data availability layer vs. Fuel). Separating structs by their data source, rather than by their usage/definition, is not a terribly helpful pattern. 

Before we proceed with this PR, I wanted to solicit some opinions on using `BlockHeight` for this, instead.

ControlCplusControlV:
It could be helpful to have different structs if we are using a variety of DA layers, and it does enforce an explicit difference between DA height and current Fuel Block Height. However I don't know how much value it adds beyond that. Though if we are going to only use `BlockHeight` we should remove the entire `DaBlockHeight` struct

fuel-core-interfaces/src/model/da_block_height.rs
```diff
@@ -0,0 +1,86 @@
+use derive_more::{
+    Add,
+    Deref,
+    Display,
+    From,
+    Into,
+};
+
```
Voxelot, line 8:
```suggestion
```

bvrooman:
> > The difference is simply the source of the data (data availability layer vs. Fuel). Separating structs by their data source, rather than by their usage/definition, is not a terribly helpful pattern.
> 
> The motivation for separate structs is to add more semantic meaning to the values. This is considered best practice in Rust when dealing with unit measurements for example: https://rust-unofficial.github.io/patterns/patterns/behavioural/newtype.html#motivation

I agree that semantic meaning is important, and that having distinctions between concepts like `Miles` and `Kilometres` is superior to using an interchangeable `u64` in both contexts. I am wondering if Fuel block height units and DA block height units are truly semantically different, such as `Miles` and `Kilometres`, or are they really both just `Miles`, so to speak. I can understand the arguments on either side, and I don't have a strong opinion on it. I am fine to proceed with this approach if the general consensus is that these block height units are semantically different.

fuel-core-interfaces/src/model/da_block_height.rs
```diff
@@ -0,0 +1,85 @@
...
+impl Add<u64> for DaBlockHeight {
+    type Output = Self;
+
+    fn add(self, other: u64) -> Self::Output {
+        Self::from(self.0 + other)
+    }
+}
+
+impl Sub for DaBlockHeight {
```
Voxelot, line 53: Sub and Rem should be derivable via [derive_more](https://jeltef.github.io/derive_more/derive_more/)

ControlCplusControlV: Looks like Rem isn't but Sub is

Voxelot:
> or if they are really both just `Miles`, so to speak.

Generally speaking yes, but ultimately it comes down to the type safety and clarity of our interfaces. It ensures we never cross wires and use the incorrect height when it matters.

fuel-core-interfaces/src/model/da_block_height.rs
Voxelot: Got derived Rem working, although it required using `#[rem(forward)]` to avoid breaking other code. Here's the explanation for what `forward` does:

> A simple way of explaining the reasoning behind this difference between Add and Mul deriving, is looking at arithmetic on meters. One meter can be added to one meter, to get two meters. Also, one meter times two would be two meters, but one meter times one meter would be one square meter. As this second case clearly requires more knowledge about the meaning of the type in question deriving for this is not implemented.

> NOTE: In case you don’t want this behaviour you can add #[mul(forward)] in addition to #[derive(Mul)]. This will instead generate a Mul implementation with the same semantics as Add.

https://jeltef.github.io/derive_more/derive_more/mul.html

Voxelot (approved):

bvrooman (approved):

Issue #595, opened 2022-09-07
prevent gossip of invalid txs


Voxelot:
When a node encounters an invalid tx, it shouldn't propagate to the rest of the network.

Voxelot:
utilize [`notify_gossip_transaction_validity`](https://github.com/FuelLabs/fuel-core/blob/master/crates/services/txpool/src/ports.rs#L37)

PR #625, opened 2022-09-18
`fuel-sync` design

xgreenx:
[Rendered](https://github.com/FuelLabs/fuel-core/blob/docs/fuel-sync/docs/fuel-sync-design.md)

I [described the implementation](https://github.com/FuelLabs/fuel-core/blob/docs/fuel-sync/docs/fuel-sync-design.md#the-blocks-synchronization-on-block-header-event), assuming that we can't change the block producer. Or we know the block producer at the height `X` because Ethereum blocks are synced, and we can get information about the producer.

The responsibility from whom and how requests the blocks on the [P2P side](https://github.com/FuelLabs/fuel-core/blob/docs/fuel-sync/docs/fuel-sync-design.md#queuing-of-the-block-header). But it also is possible to move this responsibility into FS, but it requires corresponding modification of traits.


docs/fuel-sync-design.md
```diff
@@ -0,0 +1,345 @@
...
+
+The FS is a connection point between:
+- In the PoA version P2P and BI
+- In the PoS is also additionally connects P2P with CM
+
+There are three possible ways how to implement those relationships:
+
+1. Every service knows about other services related somehow to it. It 
+can be interpreted as: each service struct containing reference to other 
+services inside and using the `async/await` mechanism to interact.
```
vlopes11, line 44: Shouldn't we require `async/await` only for operations that are `I/O`? It might be overkill to have everything async, unless ofc everything is I/O

docs/fuel-sync-design.md
```diff
@@ -0,0 +1,345 @@
...
+1. Every service knows nothing about other services. It can be interpreted
```
vlopes11, line 45: This is ofc desirable, but then we need concrete types to be sent as message (via channels?). We don't know about the internals of the service, but then we need to know about the messages the service is ready to accept.

Or are we going to have a broker in the middle that will do that for us?

docs/fuel-sync-design.md
```diff
@@ -0,0 +1,345 @@
...
+### P2P service requirements
+
+The P2P service should not be aware of FS but still should provide some 
+functionality that the FS can use. As a solution, P2P can implement the 
+`BlocksFetcher` trait.
+
+```rust
+pub enum BlockBroadcast {
+    NewSealedHeaders([Sealed<FuelBlockHeader>]),
+    NewUnsealedHeaders([FuelBlockHeader]),
```
vlopes11, line 71: nit: these channel types must be owned (here its a semi-slice, and not a vec or array)

docs/fuel-sync-design.md
```diff
@@ -0,0 +1,345 @@
...
+    fn gossip(&self, data: &D);
+}
+
+pub trait Punisher<Reason> {
+    /// Anyone can report the validity or invalidity of the data. The `Punisher` decide how to process it.
+    ///
+    /// Example: If the transaction or blocker header is invalid, the punisher can decrease the reputation 
+    /// of the data provider.
+    // TODO: `D` should implement some bound that allows retrieving the data's signer and allowing to broadcast it.
+    fn report<D>(&mut self, data: D, verdict: Verdict<Reason>);
```
vlopes11, line 107: Shouldn't we have a mechanism to enforce the data to identify who is reporting? (ie `AsRef<SomeKey>`)

docs/fuel-sync-design.md
```diff
@@ -0,0 +1,345 @@
...
+}
+
+pub trait BlocksFetcher: Punisher<BlockReason> + Gossiper {
+    /// Returns structure that can be used to subscribe for `BlockBroadcast`.
+    fn sender(&self) -> &Sender<BlockBroadcast>;
```
vlopes11, line 112: Shouldn't we return the sender as owned? I know the user can clone it himself, but I think its always going to be used as owned

docs/fuel-sync-design.md
```diff
@@ -0,0 +1,345 @@
...
+
+    /// Pings the block fetcher to fetch a new latest block header and broadcasts it.
+    fn ping(&mut self);
```
vlopes11, line 115: Maybe we rename it? Ping is traditionally a request with a response (and the response is the indicator that everything is fine). In this case, we are not expecting a response, so we could instead call it `heartbeat`

AFAIU we are expecting the service to take some action in some channel. Do we have a way to enforce that in the signature? Otherwise, a service can implement this trait, and optionally implement it correctly/the intended way

docs/fuel-sync-design.md
```diff
@@ -0,0 +1,345 @@
...
+- Is it a valid header? It should be signed by the expected, for the node, producer/consensus at according height.
+  - No -> `Punisher::report`.
+  - Yes -> Next step.
+  - Don't know. The Ethereum part is not synced yet to this height. If the height is 
+  not far from the local ethereum height(the height returned by the 
+  ethereum node), we ignore this header and do not punish the sender.
+- Is it a new height?
+  - No -> Is it already in the blockchain?
+    - Yes -> Ignore(Maybe report some neutral error to P2P to prevent DDoS)
+    - No -> It is case of the [fork](#fork).
```
vlopes11, line 276: hmm this one is a hard question. I'm not sure we can passively receive forks without being the case of a punishable action. Lets open a thread to discuss this one?

vlopes11 (commented):
looks solid overall; couple of comments

docs/fuel-sync-design.md
xgreenx: Yes, I meant that we could call methods directly. I will rename it to "sync/async".

docs/fuel-sync-design.md
xgreenx: I didn't think about this rule a lot because the usage of channels makes the code more asynchronous but less clear. Also, most services can work synchronously, and channels make the development harder. For example current implementation of the TX pool use channels. To get the list of transactions, we need to write a lot of boilerplate code using `async` where we don't need it.

docs/fuel-sync-design.md
xgreenx: Yes, it is only pseudocode representing the idea of what should be received from the channel.

After implementation, we will use more concrete types.

docs/fuel-sync-design.md
xgreenx: I think it is not essential "which" services report it. More critical "about what" we want to report. If we report about blocks, it can have one score system, if about transaction -> another.

Also, we don't need to know about other services, we only need to know about the data(P2P already knows about data).

docs/fuel-sync-design.md
xgreenx: Returning the reference may be helpful if you want to check something but don't subscribe. Like `is_closed` or something else. On the interface level, we don't need to force the user to actually subscribe=) If he needs it, he will `clone`.

docs/fuel-sync-design.md
xgreenx: Actually, maybe better to remove `ping` at all and add this request to `fetch`.

I didn't get the second part=(

docs/fuel-sync-design.md
xgreenx: I guess you are talking about "Yes -> Ignore(Maybe report some neutral error to P2P to prevent DDoS)".

Yes, we need to punish. Actually, it should be solved on the P2P level. Because P2P should punish nodes that send not requested information.

Each node can report about new height one time, each next time the reputation should decrease(if we didn't request it).

docs/fuel-sync-design.md
bvrooman: A pattern we use now for owned senders is a `subscribe` method. We could have both an owned for creating a subscription and unowned version for read-only inquiries.

xgreenx: We can have both methods, but I want to keep the interface as minimal as possible=) `sender` adds flexibility, but I don't like naming

docs/fuel-sync-design.md
```diff
@@ -0,0 +1,344 @@
...
+have several evolution points).
+
+All names are not final and can be suggested by the reviewers=)
+
+## Relationship with P2P and BI
+
+![img.png](diagrams/fuel-core-design-fuel-sync-part.png)
+
+The FS is a connection point between:
+- In the PoA version P2P and BI
```
Voxelot, line 37: I think we could generalize the connection point between FS and CM regardless of whether we are in PoS mode or not.

The benefit of this approach is that the synchronizer doesn't have to have much knowledge about how to validate the sealed block consensus metadata, and just defers that to whichever configuration the CM is running in (PoA or PoS).

docs/fuel-sync-design.md
```diff
@@ -0,0 +1,344 @@
...
+1. Every service **may** know about descending services but nothing about 
+ascending services. As an example: P2P service knows nothing about its 
+listeners(ascending services), and it only propagates information via 
+channels. But FS knows about P2P(an example of descending service)
+and may request some information directly by calling the sync/async methods. 
+The same applies to BI. FS notifies all subscribers that it synced a valid block(sealed or not sealed).
+
+Those rules are rough, and services can have cases where some functionality 
+requires direct access and some pub-sub model. So it should be decided 
+individually. But after comparison, the third rule covers all our needs
```
Voxelot, line 57: > So it should be decided individually.

I think this is the correct approach, and focuses on using the right tool for each job instead of complicating simple APIs like the txpool with channels just for the sake of consistency.

docs/fuel-sync-design.md
```diff
@@ -0,0 +1,344 @@
...
+    // TODO: ?
+}
+
+pub enum Verdict<Reason> {
+    Positive,
+    Negative(Reason),
+}
+
+pub enum FetchRequest {
+    HeadersByRanges([Range<BlockHeight>]),
```
Voxelot, line 89: What is the usecase for fetching a disjoint set of headers (Ie. `[1..2, 5..10, 20..100]`)? Why not simplify this to

```suggestion
    HeadersByRange(Range<BlockHeight>),
```

docs/fuel-sync-design.md
```diff
@@ -0,0 +1,344 @@
...
+    // TODO: `D` should implement some bound that allows retrieving the data's signer and allowing to broadcast it.
+    fn report<D>(&mut self, data: D, verdict: Verdict<Reason>);
+}
+
+pub trait BlocksFetcher: Punisher<BlockReason> + Gossiper {
+    /// Returns structure that can be used to subscribe for `BlockBroadcast`.
+    fn sender(&self) -> &Sender<BlockBroadcast>;
+
+    /// Fetches the data(somehow, maybe in parallel) and broadcasts ready parts.
+    fn fetch(&mut self, request: FetchRequest);
```
Voxelot, line 117: This may be a good usecase for [FuturesUnordered](https://docs.rs/futures/latest/futures/stream/futures_unordered/struct.FuturesUnordered.html), which allows joining on a set of parallel tasks.

docs/fuel-sync-design.md
Voxelot: Wouldn't this be a `Receiver`? Sender implies that anyone who obtains this could send a broadcast, instead of just listening for broadcasts.

docs/fuel-sync-design.md
```diff
@@ -0,0 +1,344 @@
...
+pub trait BlocksCommitter {
+    /// Returns structure that can be used to subscribe for `CommitBroadcast`.
+    fn sender(&self) -> &Sender<CommitBroadcast>;
+
+    /// Commits the blocks to the database. Executes each block from 
+    /// `blocks`, validates it, and updates the state.
+    /// 
+    /// Return errors if something is wrong during the commit process.
+    /// The state is not updated in that case.
+    fn commit(&mut self, blocks: [FuelBlock]) -> Result<(), CommitError>;
```
Voxelot, line 154: We should pass along the consensus data and require BI to persist it, so that other peers may sync off of this node.

```suggestion
    fn commit(&mut self, blocks: [Sealed<FuelBlock>]) -> Result<(), CommitError>;
```

docs/fuel-sync-design.md
xgreenx: You are right, the idea was to use `Receiver` to be able only to receive information. 

I thought `Sender.send` requires mutability, but it does not. So yea, then better to name the method `subscribe` and return `Receiver<...>`

docs/fuel-sync-design.md
xgreenx: We can=) I see CM as descending service for FS in this case. So FS will execute something like the `check(blocks: &[Sealed<Block>]) -> bool` method and after deciding what to do with blocks(reject or commit).

docs/fuel-sync-design.md
```diff
@@ -0,0 +1,344 @@
...
+    HeadersByRange(Range<BlockHeight>),
+    HeadersByHash([Bytes32]),
+    BlocksByRanges([Range<BlockHeight>]),
+    BlocksByHash([Bytes32]),
+    /// Ask neighbors about the latest block height
+    Heartbeat,
+}
+
+// TODO: `D` should implement some bound that allows to serialize it for gossiping.
+pub trait Gossiper<D> {
```
Voxelot, line 98: I think I'd prefer using a concrete enum over all gossipable data types to a generic param. This makes it very clear what kind of data is actually supported for gossip, and also gives us more compiler guarantees that gossiping is implemented across all of our types via exhaustive matching)

docs/fuel-sync-design.md
```diff
@@ -0,0 +1,348 @@
...
+methods) to the P2P and interacts with it periodically. 
+All data is received from the channel and processed by FS to update the 
+inner state of the synchronization. FS requests data from P2P for synchronization each time when it is required. 
+If some data is invalid or corrupted, the FS reports 
+it to P2P, and P2P can punish the sender of the data by decreasing the 
+reputation(or blacklisting). FS calls `Gossiper::gossiping` at the end of block synchronization.
+
+The same approach can be used between the P2P and TP. 
+`Punisher` trait adds a mechanism for [reputation management](https://github.com/FuelLabs/fuel-core/issues/460).
+`Gossiper` trait solves [gossiping of invalid transactions](https://github.com/FuelLabs/fuel-core/issues/595).
```
leviathanbeak, line 135: I don't see how a trait solves this problem, Gossipsub Messages are already clearly defined in fuel-p2p, the issue #595 that is referenced is more when there is a "valid" type, like a Transaction, but it's invalid in terms of its state on the chain itself.

xgreenx: We have several validity checks:
-  Validate that entity can be properly decoded from the bytes. It should be done on the P2P side.
- Validate that transactions are valid in business logic(all input arguments are unique, all output are valid, use `CheckedTransaction`). It is not P2P responsibility. It seems part of the TXPool.
- Validate that transition already in the TXPool(we don't need to gossip the same transaction). This can be done only by TXPool.

Voxelot: > Validate that transition already in the TXPool(we don't need to gossip the same transaction). This can be done only by TXPool.

It's debatable whether compact blocks over p2p are really useful. In practice, they can slow down consensus if nodes have to query peers for txs they are missing in their txpool. I think we should assume a broadcast of full blocks. Full blocks are also required for syncing old blocks, as the txs will no longer exist in anyone's txpool.

xgreenx: Why should nodes query transactions that missing in node pool?

Voxelot: Ah wait, I misunderstood validating if a transaction already in txpool would be used for compact blocks. Re-read and realized it's just to prevent re-gossiping txs we've already received.

docs/fuel-sync-design.md
```diff
@@ -0,0 +1,355 @@
...
+- For the PoA version: We ignore it and try to find blocks that link with our version of the blockchain. We should report it to the node manager(via logs, or maybe we want to do something more).
+- For the PoS version: We should have rules in the consensus for this case.
+
+#### Queuing of the block header
+
+When FS receives a block header, first FS checks:
+- Is it a valid header? Blocks should be linked, and signatures should match. After, forwards blocks to the CM to verify producer/consensus at the according height.
+  - No -> `Punisher::report`.
+  - Yes -> Next step.
+  - Don't know. The Ethereum part is not synced yet to this height. If the height is
```
Voxelot, line 286: We can defer ethereum height checking to the BI, to consolidate this check between CM->BI and FS->BI.

If we are behind on the ethereum height, the BI should decide whether to wait on a better height or reject the block.

xgreenx: But it means that we should check the consensus details when the ethereum block is available. This check should be done by CM then. CM can wait for a new height from the relayer

Voxelot: Ah good point! Maybe we should check it in a few places then, since it'll up to the CM to decide what details from ethereum it needs (or not). Whereas the block importer needs to provide some guarantees that the messages included in a block are going to exist before passing the block to the executor.

xgreenx: Then it's decided=) `SealValidator::valdiate` is `async`. If there is no Ethereum block, then CM waits for it(In the PoA, it is the default behavior for all block headers, in the PoS, we need to think about cases with fake height). When BI will ask the relayer about the information, it will already be available(because CM awaited it).

With this approach, we have a problem: we can have many async pending requests, and each consumes a memory. Maybe we need to limit this somehow.

Voxelot: The first place we can restrict this is via the active set size, since that's equivalent to the buffer size of blocks we're trying to sync.

The other way to restrict this is to validate blocks with the CM and commit to BI one at a time. This assumes that the CM doesn't wait for the eth height until it's already verified things like the signature first to avoid this becoming an attack vector.

xgreenx: Could you elaborate more on the second solution, please? I didn't get it=(

Voxelot: Ah true, I suppose we could use some heuristic where we check if our relayer node is synced and the da_height is greater than some margin to reject. This way we add some tolerance in case our relayer's eth node claims to be synced when it really isn't.

For example:

```mermaid
flowchart
    a["if block.da_height > relayer.da_height"] -- no --> b["check signature"]
    b -- valid --> s["success"]
    b -- invalid --> f["reject"]
    a -- yes --> c["wait for relayer to sync"]
    c --> d["is block.da_height > relayer.da_height + 10"]
    d -- yes --> f
    d -- no --> a
    
```

xgreenx: Yep, added the final description of what to do in the case when the Relayer is behind.

Voxelot:
Redo to match the current state of sync without reputation. Add reputation in a follow-up.

xgreenx:
A lot of time passed since this draft design=) Closing for now, since we need to create a fresh document because of modifications that we introduced

PR #630, opened 2022-09-20
Add crypto & blockchain to fuel-benches

vlopes11:


ControlCplusControlV:
Is it possible to reduce the amount of files/lines in this PR? I understand many are auto-generated reports, but adding some items to a `.gitignore` could be useful. My browser is lagging try to review all 2100 files and 168,000 lines

fuel-benches/benches/set/alu.rs
```diff
@@ -0,0 +1,247 @@
...
+        &mut group,
+        "not",
+        VmBench::new(Opcode::NOT(0x10, 0x11))
+            .with_prepare_script(vec![Opcode::MOVI(0x11, 100000)]),
+    );
+
+    run_group(
+        &mut group,
+        "or",
+        VmBench::new(Opcode::OR(0x10, 0x11, 0x12)).with_prepare_script(vec![
```
ControlCplusControlV, line 169: I see this script popping up a lot, could these maybe be moved to a single constant or single expression in case it is ever updated?

fuel-benches/benches/set/blockchain.rs
```diff
@@ -0,0 +1,121 @@
...
+
+pub fn run(c: &mut Criterion) {
+    let rng = &mut StdRng::seed_from_u64(2322u64);
+
+    let mut group = c.benchmark_group("blockchain");
+
+    let cases = vec![
+        1, 10, 100, 1_000, 10_000,
+        100_000,
+        // 1_000_000 - I/O error in BAL; investigate
```
ControlCplusControlV, line 19: Should this be added to the issue tracker for `fuel-core`?

fuel-benches/benches/set/crypto.rs
```diff
@@ -0,0 +1,61 @@
...
+            ])
+            .with_data(signature.iter().chain(message.iter()).copied().collect()),
+    );
+
+    run_group(
+        &mut group,
+        "k256",
+        VmBench::new(Opcode::K256(0x10, 0x00, 0x11))
+            .with_prepare_script(vec![
+                Opcode::MOVI(0x10, Bytes32::LEN as Immediate18),
```
ControlCplusControlV, line 52: Same issue on moving this script to a constant or single expression if possible

fuel-benches/benches/set/mem.rs
```diff
@@ -0,0 +1,131 @@
...
+        VmBench::new(Opcode::LW(0x10, REG_ONE, 10)),
+    );
+
+    for i in &linear {
+        run_group(
+            &mut group,
+            format!("mcl ({})", i),
+            VmBench::new(Opcode::MCL(0x10, 0x11)).with_prepare_script(vec![
+                Opcode::MOVI(0x11, *i),
+                Opcode::ALOC(0x11),
```
ControlCplusControlV, line 54: Could be nice to have comments explaining why each script was chosen for each benchmark for those consulting benchmarks and to keep every up to speed with logic used for benchmarking

fuel-benches/benches/set/blockchain.rs
vlopes11: I'm investigating that one to understand if its a bug in the benchmark implementation or something more. The early conclusion is for the latter, but I'm not sure yet

fuel-benches/benches/set/mem.rs
vlopes11: Cool, will add some docs and simplify the code a bit

vlopes11:
@ControlCplusControlV removed the json files. Still have a lot of files, but half the original amount

Voxelot (changes requested):
Hmm - this is quite a bit more generated files than I expected 🤔 

It would be better to figure out a solution using the new gh page automation actions to upload these from ci rather than managing all of this in our git history. This seems like way too much to check in.

vlopes11:
The problem is that `tempdir` implements `Drop` manually, and removes all directories when goes out of scope

https://github.com/rust-lang-deprecated/tempdir/blob/403f918d9f3708ca144c727d20d2c1f8738767fc/src/lib.rs#L349-L356

When the database grows large, some path references are reconstructed so the original tempdir reference path is list - hence, it gets dropped so the database can't find the dir in the disk because it is entirely removed.

We clone every instance of the prepared case before we run the benchmarks so one instance will not affect the other. Instead of making a super hacky way for the reference to outlive this database reorg, the best solution is to clone properly the prepared case by copying all internal data.

Added issue to fix that:
https://github.com/FuelLabs/fuel-core/issues/636

Voxelot:
I got the balance benchmarks working by simply using `Database::default()`, which also uses tempfile under the hood but also wraps the tempdir in an Arc, making it safe to clone the db without prematurely dropping the tempfile.

Voxelot (approved):
approving (assuming we can automerge from master once the rust 1.64 upgrade goes in)

Issue #398, opened 2022-06-06
Relayer - decouple finalized event processing from eth pubsub


Voxelot:
While using the eth pubsub apis for pending events provides useful ux for users tracking deposits or withdrawals, or for validators to track pending block commitments, they should be considered unreliable for sourcing finalized events for consensus critical functions.

This is because ethereum provides zero durability guarantees for their pubsub APIs and no way to replay missed messages. This will be problematic on many levels, for example if there are network disruptions, if the eth client pubsub connection drops, or the eth node reboots / rescales. 

Instead relying on pubsub messages for our source of truth, we should do the following:

1. Make the finalized event handler responsible for its own log/block polling without pubsub, using only finalized blocks it cares about.
2. Pending events will continue to make a best effort to track pending activity, but will wait to be notified by the finalized event handler when an event was processed before removing it.

Inverting the dependency between the pending events and finalized events will simplify the relayer design. It gives us rock-solid guarantees that no log entry was dropped or accidentally processed multiple times in a way that could affect consensus. The handling of pending events can also be greatly relaxed, as pubsub data would be relegated to UX/UI concerns and doesn't need to be perfectly accurate all the time. 

For example, we'd no longer need to be concerned about the precise handling of subscriptions being started or stopped at the perfect time [during initialization](https://github.com/FuelLabs/fuel-core/pull/345#discussion_r889402006), and performing [sanity checks](https://github.com/FuelLabs/fuel-core/blob/feac06c7f883cbba91df6df1dc05520f878a4ebc/fuel-relayer/src/finalization_queue.rs#L259) on all the data we receive from the pubsub system before committing it.

xgreenx:
It is done in the https://github.com/FuelLabs/fuel-core/pull/631

PR #631, opened 2022-09-20
Relayer mvp

freesig:
This is a rewrite of the relayer node to simplify and only download logs. This implements #398


fuel-relayer/flake.nix
```diff
@@ -0,0 +1,64 @@
+{
```
Voxelot, line 1: can we have some tips on how to use this added to fuel-relayer readme.md?

fuel-core/src/database.rs
```diff
@@ -566,5 +566,51 @@ mod relayer {
                     );
                 });
         }
+
+        async fn get_pending_committed_fuel_height(&self) -> Option<BlockHeight> {
```
Voxelot, line 570: Should we just make this method fallible instead of panicking?

fuel-relayer/src/mock_db.rs
```diff
@@ -47,35 +46,21 @@ use fuel_core_interfaces::{
 };
 
 #[derive(Default)]
-pub(crate) struct Data {
+pub struct Data {
     pub messages: HashMap<MessageId, Message>,
-    pub validators: HashMap<ValidatorId, (ValidatorStake, Option<ConsensusId>)>,
-    pub delegator_index: HashMap<Address, Vec<DaBlockHeight>>,
-    pub staking_diffs: BTreeMap<DaBlockHeight, StakingDiff>,
```
Voxelot, line 51:
```suggestion
```

fuel-core/src/database.rs
freesig: I was just copying the behavior of all the other trait methods. I'm not really sure what usecases can actually return an error but if there are some then we should probably change all the functions on this trait to be fallible in a new PR?

fuel-relayer/flake.nix
freesig: Will do, I'm just working on making it a little easier to use

Voxelot:
I'm thinking we shouldn't check in all of this generated code, this is something that could be handled via a build.rs imo

freesig:
> I'm thinking we shouldn't check in all of this generated code, this is something that could be handled via a build.rs or other macro imo

Do you mean the bindings json?

Voxelot:
> Do you mean the bindings json?

Yeah it seems like this is all generated code from the json: [test.rs](https://github.com/FuelLabs/fuel-core/blob/cc60be4aceb2bc471d55938f9d879f833ce5f8c2/test.rs)

fuel-relayer/README.md
```diff
@@ -3,14 +3,8 @@
 
 Functionalities expected from Relayer:
 
-* Validator staking (Handle ValidatorList and use it for fuel-bft):
```
ControlCplusControlV, line 6: Why are so much of the original `fuel-relayer` docs being removed? Would be nice to have them updated where possible. I see a lot of changes also removed a lot of the related `eth` stuff, so having somewhere documenting the current rationale of the `relayer` design and why its currently structured this way would be helpful

fuel-relayer/README.md
```diff
@@ -3,14 +3,8 @@
 
 Functionalities expected from Relayer:
```
ControlCplusControlV, line 5: Why are so much of the original `fuel-relayer` docs being removed? Would be nice to have them updated where possible. I see a lot of changes also removed a lot of the related `eth` stuff, so having somewhere documenting the current rationale of the `relayer` design and why its currently structured this way would be helpful

fuel-relayer/src/config.rs
```diff
@@ -1,163 +1,54 @@
...
+    pub const DEFAULT_DA_FINALIZATION: u64 = 100;
+    pub const DEFAULT_SYNC_MINIMUM_DURATION: Duration = Duration::from_secs(5);
 }
 
 impl Default for Config {
     fn default() -> Self {
         Self {
-            da_finalization: DaBlockHeight::from(64u64),
+            da_finalization: DaBlockHeight::from(Self::DEFAULT_DA_FINALIZATION),
             // Some(String::from("http://localhost:8545"))
```
ControlCplusControlV, line 44: This line should probably be removed or updated since `eth_client` is now a `URL` and not a string

fuel-relayer/src/mock_db.rs
```diff
@@ -1,11 +1,10 @@
+#![allow(missing_docs)]
```
ControlCplusControlV, line 1: I haven't seen this lint tag needed before, is there a reason it hasn't shown before but is now an issue?

fuel-relayer/README.md
```diff
@@ -34,96 +28,25 @@ There are few rules that `da_height` (da as data availability) need to follow an
...
-
-## Block Commitments
-
-When new fuel block is made it needs to be committed to ethereum network.
-For every block that we don't have fuel commit we try to do our best guess and send block that contracts can include, gaps in height will be discarded by contract.
-
-There is timeframe of few eth blocks where current leader can broadcast block to ethereum network. In case that he do it in time he will receive additional reward. In case that leader does not commit fuel block next leader will have ability to do it for him as bundle block commit, in that case bonus reward will go to second leader that has spend additional gas to push this change to ethereum network. This is still WIP discussion.
-
-Github discussion on fuel sync with L1: <https://github.com/FuelLabs/fuel-specs/issues/285>
+## Contract Flake
```
ControlCplusControlV, line 35: Is the Fuel Nix flake also being introduced in this PR? Would it be possible to break that out into a separate PR for cleaner review?

ControlCplusControlV (commented):
Changes look good to me, but since this PR is nearly redesigning the entire relayer crate I think having more documentation would be helpful. Many changes I can see are happening, but having more insight into why they were made and the logic behind them would be helpful. Atleast preserving the same level of documentation previously present would be helpful.

Code style though LGTM except the current clippy errors

fuel-relayer/src/mock_db.rs
freesig: It's because I have `#![deny(missing_docs)]` in the lib.rs but I don't want to force docs on test code.

fuel-relayer/README.md
```diff
@@ -25,105 +18,39 @@ Example of sliding window:
...
-* Solution: introduce `da_height` variable inside fuel block header that will tell at what block we are including validator delegation/withdrawal and token deposits.
+* Solution: introduce `da_height` variable inside fuel block header that will tell at what block we are including token deposits.
 
-There are few rules that `da_height` (da as data availability) need to follow and can be enforced with v2 contract:
+There are few rules that `da_height` need to follow and can be enforced with v2 contract:
 
 * Last block `da_height` needs to be smaller number then current one.
 * `da_height` shouldn't be greater then `best_da_block`-`finalization_slider`
 
 In most cases for fuel block `da_height` can be calculated as `current_da_height-1`
```
freesig, line 28: @Voxelot Is this still relevant / correct? I don't really understand what it means.

fuel-relayer/README.md
```diff
@@ -25,105 +18,39 @@ Example of sliding window:
 ![Sliding Window](../docs/diagrams/fuel_v2_relayer_sliding_window.jpg)
```
freesig, line 18: @Voxelot Should I try and make a new diagram as this one mentions staking etc?

fuel-relayer/README.md
freesig: Ok I've updated the readme

fuel-relayer/README.md
freesig: I have moved the nix stuff too https://github.com/FuelLabs/fuel-core/pull/665

freesig:
@ControlCplusControlV The tests should all be passing now.

fuel-core/src/cli/run/relayer.rs
```diff
@@ -29,42 +22,27 @@ pub struct RelayerArgs {
...
+    pub log_page_size: u64,
 
-    /// number of blocks that will be asked at one time from client, used for initial sync
-    #[clap(long = "relayer-init-sync-step", default_value = "1000")]
-    pub initial_sync_step: usize,
-
-    /// Refresh rate of waiting for eth client to finish its initial sync in seconds.
-    #[clap(long = "relayer-init-sync-refresh", default_value = "10")]
-    pub initial_sync_refresh: u64,
+    #[clap(long = "relayer-min-duration-s", default_value = "5")]
```
Voxelot, line 30: Could you add some rustdoc here explaining what this option is for?

fuel-relayer/README.md
Voxelot: I think this would be more of a concern for the block producer when selecting a da_height. Since we explicitly include a da_height on our block headers, we shouldn't need to make any guesses about the da_height in our code. I vote for just removing this.

fuel-relayer/README.md
```diff
@@ -25,105 +18,20 @@ Example of sliding window:
...
 
 * Problem: How to choose when bridge message event gets enabled for use in fuel, at what exact fuel block does this happen? (Note that we have sliding window)
-* Solution: introduce `da_height` variable inside fuel block header that will tell at what block we are including validator delegation/withdrawal and token deposits.
+* Solution: introduce `da_height` variable inside fuel block header that will tell at what block we are including token deposits.
 
-There are few rules that `da_height` (da as data availability) need to follow and can be enforced with v2 contract:
+There are few rules that `da_height` need to follow and can be enforced with v2 contract:
```
Voxelot, line 23: debatable whether enforcing these rules is relevant to the contracts

fuel-relayer/src/mock_db.rs
```diff
@@ -14,68 +13,40 @@ use std::{
...
-    pub finalized_da_height: DaBlockHeight,
+    pub finalized_da_height: Option<DaBlockHeight>,
     pub last_committed_finalized_fuel_height: BlockHeight,
+    pub pending_committed_fuel_height: Option<BlockHeight>,
 }
 
-#[derive(Default)]
-pub(crate) struct MockDb {
+#[derive(Default, Clone)]
+pub struct MockDb {
```
Dentosal, line 46: Now that this is exported outside the crate, it should have a doc comment describing its usage. Should it be hidden before the `test-helpers` feature? Are breaking changes to MockDb considered semver-breaking? If no, that must be documented as well.

fuel-relayer/README.md
Voxelot: Yeah, we can also just remove the diagram if we think the current approach is simple enough to explain without it.

Dentosal (changes requested):

fuel-relayer/src/mock_db.rs
freesig: The entire module is behind either the `test or test-helpers` feature flag so it's definitely not semvar breaking.

fuel-relayer/tests/integration.rs
```diff
@@ -0,0 +1,116 @@
...
+    eth_node.update_data(|data| data.best_block.number = Some(200.into()));
+    let relayer = RelayerHandle::start_test(eth_node, Box::new(mock_db.clone()), config);
+
+    relayer.await_synced().await.unwrap();
+
+    for msg in expected_messages {
+        assert_eq!(
+            &*StorageInspect::<Messages>::get(&mock_db, msg.id())
+                .unwrap()
+                .unwrap(),
```
xgreenx, line 70:
```suggestion
            &*mock_db
                .storage::<Messages>()
                .get(msg.id())
                .unwrap()
                .unwrap(),
```

fuel-relayer/tests/integration.rs
```diff
@@ -0,0 +1,116 @@
+#![cfg(feature = "test-helpers")]
+
+use fuel_core_interfaces::{
+    common::fuel_storage::StorageInspect,
```
xgreenx, line 4:
```suggestion
    common::fuel_storage::StorageAsRef,
```

fuel-relayer/src/relayer/syncing.rs
```diff
@@ -0,0 +1,70 @@
...
+pub async fn wait_if_eth_syncing<P>(eth_node: &P) -> anyhow::Result<()>
+where
+    P: Middleware<Error = ProviderError> + 'static,
+{
+    let mut count = 0;
+    while get_status(eth_node).await? {
+        count += 1;
+        if count > 12 {
+            tracing::warn!(
+                "relayer has been waiting longer then a minute for the eth node to sync"
```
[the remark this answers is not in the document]
xgreenx: And for example, if the `current_block` didn't change after the `X` synchronization rounds, we will return an error.

fuel-core/src/database/metadata.rs
```diff
@@ -15,6 +15,7 @@ pub(crate) const FINALIZED_DA_HEIGHT_KEY: &[u8] = b"finalized_da_height";
 pub(crate) const VALIDATORS_DA_HEIGHT_KEY: &[u8] = b"current_validator_set";
 pub(crate) const LAST_COMMITTED_FINALIZED_BLOCK_HEIGHT_KEY: &[u8] =
```
xgreenx, line 16: Do we still need these constants and the code associated with them?

fuel-relayer/src/relayer/get_logs.rs
```diff
@@ -0,0 +1,72 @@
...
+pub(crate) fn download_logs<P>(
+    eth_sync_gap: &state::EthSyncGap,
+    contracts: Vec<H160>,
+    eth_node: Arc<P>,
+    page_size: u64,
+) -> impl futures::Stream<Item = Result<Vec<Log>, ProviderError>>
+where
+    P: Middleware<Error = ProviderError> + 'static,
+{
+    // Create a stream of paginated logs.
```
Voxelot, line 17: Might be helpful to have some logging here to track the relayer download progress:

```suggestion
    // Create a stream of paginated logs.
    `debug!("Downloading logs for block range: {}", &eth_sync_gap);`
```

fuel-relayer/src/relayer/state/test.rs
```diff
@@ -0,0 +1,89 @@
...
+    } => None; "ahead so doesn't need to sync"
+)]
+#[test_case(
+    TestDataSource {
+        eth_remote_current: 300,
+        eth_remote_finalization_period: 100,
+        eth_local_finalized: Some(50),
+    } => Some(51..=200); "behind by less so needs to sync"
+)]
+#[tokio::test]
```
Voxelot, line 43: We should probably also test the edge case of what happens when `eth_remote_current < eth_remote_finalization_period`, for new eth nodes (e.g. local dev environment testing)

fuel-core/src/database.rs
```diff
@@ -377,125 +377,17 @@ mod relayer {
...
-                        }
-                        out.push((block, diff))
-                    }
-                    Err(err) => panic!("get_validator_diffs unexpected error:{:?}", err),
-                }
-            }
-            out
-        }
-
-        async fn apply_validator_diffs(
```
Voxelot, line 479: Replying to @xgreenx about old data cleanup:

Yes, we should hunt down the remaining places this unused data is being managed and clean it up.

fuel-relayer/src/relayer/syncing.rs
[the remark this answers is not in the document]
Voxelot: I agree that a minute is too short, syncing a brand new eth node from scratch can even take up to several days!

Although I'd change the log message slightly since the relayer is actually waiting for the eth node to sync before it downloads logs for its own sync process:

```
tracing::info!("Waiting for the ethereum endpoint to finish Syncing: {:?}", status);
```

fuel-core/src/database/metadata.rs
Voxelot: We would have a git history of the block commitment logic, so it's probably better to clean up logic we aren't sure if we'll need. There's a chance we'll come up with a way to never require the relayer to post anything to Ethereum.

fuel-relayer/src/relayer/get_logs.rs
```diff
@@ -0,0 +1,72 @@
...
+/// Write the logs to the database.
+pub(crate) async fn write_logs<S>(
+    database: &mut dyn RelayerDb,
+    logs: S,
+) -> anyhow::Result<()>
+where
+    S: futures::Stream<Item = Result<Vec<Log>, ProviderError>>,
+{
+    tokio::pin!(logs);
+    while let Some(events) = logs.try_next().await? {
```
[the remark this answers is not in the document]
freesig: Yeh I actually intentionally did this because I wasn't sure about adding the complexity on tracking pagination. Maybe this would suite a follow up PR as it would be nice to get it reviewed properly. I did assume that downloading messages was idempotent but it sounds like that's not the case.

fuel-relayer/src/relayer/state/test.rs
```diff
@@ -0,0 +1,89 @@
...
+#[test_case(EthSyncGap::new(0, 1), 2, 0 => (0, 1, false))]
+#[test_case(EthSyncGap::new(0, 1), 2, 1 => (2, 1, true))]
+#[test_case(EthSyncGap::new(0, 1), 2, 2 => (4, 1, true))]
+#[test_case(EthSyncGap::new(0, 3), 2, 0 => (0, 1, false))]
+#[test_case(EthSyncGap::new(0, 3), 2, 1 => (2, 3, false))]
+#[test_case(EthSyncGap::new(0, 3), 2, 2 => (4, 3, true))]
+#[test_case(EthSyncGap::new(1, 8), 3, 0 => (1, 3, false))]
+#[test_case(EthSyncGap::new(1, 8), 3, 1 => (4, 6, false))]
+#[test_case(EthSyncGap::new(1, 8), 3, 2 => (7, 8, false))]
+#[test_case(EthSyncGap::new(1, 8), 3, 3 => (10, 8, true))]
```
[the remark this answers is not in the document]
freesig: So the way the bounds check works is by checking for an empty inclusive range. So `start > end`. This will never actually get used as the page is always checked for empty before getting the bounds. I could return `Option<u64>` from the `oldest` and `latest` methods though, which would maybe be more clear that the page is complete.

fuel-relayer/src/relayer/syncing.rs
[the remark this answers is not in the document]
freesig: Hmm, do we actually imagine we will be querying an eth node that is in the process of initial sync? Basically the relayer can't do anything while the eth node is syncing so a user might be confused why nothing is happening. I can definitely add a configurable reporting period.

fuel-relayer/src/relayer/get_logs.rs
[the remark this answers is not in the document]
Voxelot: I'm fine with deferring this to a follow-up issue, feel free to open new task(s) for these changes to resolve this thread if you'd prefer to address this later.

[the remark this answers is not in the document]
freesig: TODO: Make issue for the second concern.

fuel-relayer/src/relayer/syncing.rs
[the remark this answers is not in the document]
freesig: I have addressed this although I did not return an error if the current block hasn't changed because we don't want to panic the node over this.

fuel-core/src/database.rs
freesig: TODO: Make issue for this

fuel-relayer/src/relayer/get_logs.rs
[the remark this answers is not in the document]
freesig: I've addressed the first concern.

fuel-relayer/src/relayer/syncing.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+    Ok(())
+}
+
+async fn get_status<P>(eth_node: &P) -> anyhow::Result<SyncingStatus>
+where
+    P: Middleware<Error = ProviderError> + 'static,
+{
+    match eth_node.syncing().await {
+        Ok(s) => Ok(s),
```
Voxelot, line 54: [nit] this match could be replaced by using anyhow context instead: https://docs.rs/anyhow/latest/anyhow/trait.Context.html

Voxelot (dismissed):
Approved - please open an issue to fix the message insertion to check existence first though.

Voxelot (approved):

PR #635, opened 2022-09-21
move chain_config to separate crate

### LINKED ISSUE #603: Relocate ChainConfig types to a separate crate within fuel-core
The definition of a chainconfig and associated types currently lives in fuel-core. Since there are many tools that will want to configure the chainspec without depending on the entirety of fuel-core, we should move all chainconfig types to their own crate with minimal dependencies. 

This will prevent further duplication of our types by the SDK:

fuel-core: https://github.com/FuelLabs/fuel-core/blob/v0.10.1/fuel-core/src/chain_config.rs#L142
fuels-rs: https://github.com/FuelLabs/fuels-rs/blob/v0.23.0/packages/fuels-test-helpers/src/node.rs#L50

leviathanbeak:
closes #603 

Mostly copy/pasta apart from one important change, since `ChainConfig` was tightly coupled with `Database` (Db was passed during initialization of ChainConfig), I have created a `ChainConfigDb` trait which I passed instead. And on the `fuel-core/Database` side I have implemented that trait so that it is compatible for `ChainConfig` initialization.

By "implemented" I mean the methods were already previously implemented so I just referenced existing methods in trait implementation.


ControlCplusControlV:
Would it be good to have Relayer and p2p config cli commands moved into a config crate as well. That way all the configuration and parameters to run the network could be in a single place. I'm just curious if this is something we'd want to add into this new crate (which may then require a more general rename).

But I have no problems on the impl, LGTM overall and can approve if there is a reason for against a central config crate with those additional options

leviathanbeak:
@ControlCplusControlV IMO the cli config commands should stay with the CLI app (unless whole CLI gets app moved). These config constructs that I've moved out are mostly that someone (ie SDK) can import them without importing whole of `fuel-core` on the other hand, I don't see that being the case with CLI commands which are just there to be constructed from the cli commands.

fuel-chain-config/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "fuel-chain-config"
+version = "0.1.0"
+edition = "2021"
+
```
Voxelot, line 5: Needs proper package headers like `fuel-core` or `fuel-txpool` in order to be publishable. I.e. `license`, `description`, `version` that matches the other crates etc..

fuel-chain-config/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "fuel-chain-config"
```
Voxelot, line 2: We need to update the "Verify tag version" ci job to check this crate as well

Voxelot:
Missing README.md for new package

Voxelot:
reserved the name on crates.io: https://crates.io/crates/fuel-chain-config

Voxelot:
Maybe we should organize this new crate a bit more and move the different structs/traits into their own submodule files? I'm finding it easy to get lost in this large lib.rs.

leviathanbeak:
@Voxelot good point, I'll do it

leviathanbeak:
I've moved config structs into separate files: `config/{struct_name}.rs` and exported them all

fuel-core/src/lib.rs
```diff
@@ -1,4 +1,3 @@
-pub mod chain_config;
 pub mod database;
```
Voxelot, line 1: maybe we should replace this with the following to avoid breaking imports downstream?
```suggestion
pub use fuel_chain_config as chain_config;
pub mod database;
```

fuel-tests/Cargo.toml
```diff
@@ -19,6 +19,7 @@ harness = true
 [dependencies]
 async-std = "1.12"
 chrono = { version = "0.4", features = ["serde"] }
+fuel-chain-config = { path = "../fuel-chain-config", version = "0.1.0" }
```
Voxelot, line 22: use re-exported config from fuel-core instead of making a new dep here.

fuel-chain-config/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+version = "0.1.0"
```
Voxelot, line 3: version must match the other crates to pass CI verification during publish
```suggestion
version = "0.10.1"
```

PR #637, opened 2022-09-21
using arc'd tmpfile to fix bal(1M) benchmark

Voxelot:
use `Database::default` to get a tmpfile rocksdb db instance. This solves [problems](https://github.com/FuelLabs/fuel-core/pull/630#issuecomment-1254248494) related to prematurely dropping the tempfile by using an Arc.

### DIFF
--- a/fuel-benches/src/lib.rs
+++ b/fuel-benches/src/lib.rs
@@ -1,20 +1,21 @@
-use std::{
-    io,
-    iter,
-};
+pub use fuel_core::database::Database;
 use fuel_core_interfaces::common::fuel_tx::{
     StorageSlot,
     TransactionBuilder,
 };
-pub use fuel_core::database::Database;
 pub use fuel_core_interfaces::common::{
     consts::*,
     prelude::*,
 };
 pub use rand::Rng;
+use std::{
+    io,
+    iter,
+};
 
-fn new_db() -> io::Result<Database> {
-    Ok(Database::default())
+fn new_db() -> Database {
+    // when rocksdb is enabled, this creates a new db instance with a temporary path
+    Database::default()
 }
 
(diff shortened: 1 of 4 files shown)


vlopes11:
If we take that approach, we might create a shared resource that will impact on each iteration of the benchmark.

What if we keep the problem split, first merge #630 and then make another PR to handle this issue? This way, we don't aggregate different responsibilities to the original PR

Voxelot:
> If we take that approach, we might create a shared resource that will impact on each iteration of the benchmark.

This is already happening right now in the current implementation of #630, the only difference this PR introduces is making the tempfile properly clonable like everything else in VmBenchPrepared.

fuel-benches/src/lib.rs
```diff
@@ -1,23 +1,21 @@
...
 pub use rand::Rng;
+use std::{
+    io,
+    iter,
+};
 
-fn new_db() -> io::Result<Database> {
-    TempDir::new().and_then(|t| Database::open(t.as_ref()).map_err(|e| e.into()))
+fn new_db() -> Database {
+    // when rocksdb is enabled, this creates a new db instance with a temporary path
```
ControlCplusControlV, line 17: Should `fuel-core`'s rocksdb feature be enabled by default on this benchmark then? That way no benchmarks are accidentally done on much more performant RAM databases?

fuel-benches/src/lib.rs
```diff
@@ -1,23 +1,21 @@
...
+    Database::default()
```
vlopes11, line 18: Maybe we just delete this `new_db` function and instead `unwrap_or_default`?

vlopes11 (approved):
lgtm, one optional change

Issue #644, opened 2022-09-26
update libp2p to the latest 0.48 version


leviathanbeak:
This will close #530 

Our current version is 0.44, a lot of stuff has changed since then, one of the biggest changes is the removal of `NetworkBehaviourEventProcess` documented at https://github.com/libp2p/rust-libp2p/blob/libp2p-swarm-0.38.0/swarm/CHANGELOG.md#0380

which will warrant changes to our codebase.

PR #661, opened 2022-09-30
update libp2p to v0.48

### LINKED ISSUE #530: RUSTSEC-2022-0040: Multiple soundness issues in `owning_ref`
> Multiple soundness issues in `owning_ref`

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `owning_ref`                      |
| Version             | `0.4.1`                   |
| URL                 | [https://github.com/noamtashma/owning-ref-unsoundness](https://github.com/noamtashma/owning-ref-unsoundness) |
| Date                | 2022-01-26                         |

- `OwningRef::map_with_owner` is [unsound](https://github.com/Kimundi/owning-ref-rs/issues/77) and may result in a use-after-free.
- `OwningRef::map` is [unsound](https://github.com/Kimundi/owning-ref-rs/issues/71) and may result in a use-after-free.
- `OwningRefMut::as_owner` and `OwningRefMut::as_owner_mut` are [unsound](https://github.com/Kimundi/owning-ref-rs/issues/61) and may result in a use-after-free.
- The crate [violates Rust&#39;s aliasing rules](https://github.com/Kimundi/owning-ref-rs/issues/49), which may cause miscompilations on recent compilers that emit the LLVM `noalias` attribute.

No patched versions are available at this time. While a pull request with some fixes is outstanding, the maintainer appears to be unresponsive.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2022-0040.html) for additional details.

leviathanbeak:
Closes #530 #644.

Also this introduces dependency on `protoc` for compiling Protobuf files, previously `prost` library bundled `protoc` within the lib, but they've recently [moved away from that approach](https://github.com/tokio-rs/prost/pull/657).

**Changes**
`libp2p` has made some breaking changes in recent updates, main change was [how you handle behaviour events](https://github.com/libp2p/rust-libp2p/blob/libp2p-swarm-0.38.0/swarm/CHANGELOG.md#0380) - this means that the Behaviour, in our case FuelBehaviour, would be stateless apart from holding other inner behaviours and all the stateful logic would be moved to FuelP2PService that holds the FuelBehaviour.

Usually this would just mean a lot of cut & paste (check the `service.rs` and `behaviour.rs` files), but in our case FuelBehaviour was generic over a NetworkCodec, and leaving it still generic would imply that our FuelBehaviourEvent enum would also need to be generic which made for some unwieldy code. 

I decided to create a wrapper struct
```rust
pub struct BehaviourEventWrapper<Codec: NetworkCodec> {
    event: FuelBehaviourEvent,
    codec: PhantomData<Codec>,
}
```
from this wrapper we would easily get the inner event and handle it in `FuelP2PService`.

Also now `FuelP2PService` needed to become generic over NetworkCodec since it's now the one that handles encoding and decoding of the network messages.


.github/workflows/ci.yml
```diff
@@ -109,7 +109,9 @@ jobs:
       - name: Install Cargo Make
         uses: davidB/rust-cargo-make@v1
         with:
-          version: "0.36.0"
+          version: "0.36.0"
```
ControlCplusControlV, line 112: Ran into an API rate limit on this in the CI, is there any way to get our rate limit expanded?

fuel-p2p/src/service.rs
```diff
@@ -2,63 +2,119 @@ use crate::{
...
 }
 
-#[allow(clippy::large_enum_variant)]
+/// Holds additional Network data for FuelBehavior
+#[derive(Debug)]
+struct NetworkMetadata {
+    gossipsub_topics: GossipsubTopics,
+}
+
+//#[allow(clippy::large_enum_variant)]
```
ControlCplusControlV, line 93:
```suggestion

```

fuel-p2p/src/service.rs
```diff
@@ -141,19 +205,177 @@ impl FuelP2PService {
...
         message: OutboundResponse,
     ) -> Result<(), ResponseError> {
-        self.swarm
-            .behaviour_mut()
-            .send_response_msg(request_id, message)
+        match (
+            self.network_codec.convert_to_intermediate(&message),
+            self.inbound_requests_table.remove(&request_id),
+        ) {
+            (Ok(message), Some(channel)) => {
```
ControlCplusControlV, line 212: Is this purely a libp2p update? It looks like a lot of new logic was introduced here, the added logic and code looks good but just want to make sure that this is indeed added code

fuel-p2p/src/orchestrator.rs
```diff
@@ -93,36 +97,34 @@ impl NetworkOrchestrator {
...
+                                },
+                            }
+                        },
+                        FuelP2PEvent::RequestMessage { request_message, request_id } => {
+                            match request_message {
+                                RequestMessage::RequestBlock(block_height) => {
+                                    let db = self.db.clone();
+                                    let tx_outbound_response = self.tx_outbound_responses.clone();
+
+                                    tokio::spawn(async move {
```
ControlCplusControlV, line 120: [nit] I know it's not the most common occurrence, and this already existed before (so it may warrant an issue if you think it's worth changing) but should this thread have some handler on it? So in the event the node has the chance to gracefully shutdown the peer can get their block rather than timing out

.github/workflows/ci.yml
```diff
@@ -109,7 +109,9 @@ jobs:
       - name: Install Cargo Make
         uses: davidB/rust-cargo-make@v1
         with:
-          version: "0.36.0"
+          version: "0.36.0"      
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
```
ControlCplusControlV, line 114:
```suggestion
        uses: arduino/setup-protoc@v1
         with:
                repo-token: ${{ secrets.GITHUB_TOKEN }}
```

my above suggestion may be a bit weird, but the [Protoc Docs](https://github.com/arduino/setup-protoc/blob/master/README.md) recommends using a token to avoid ratelimiting, which we already have run into on this CI job, so would be nice to have this added

ControlCplusControlV (changes requested):

fuel-p2p/src/orchestrator.rs
freesig: Agree this probably requires an issue but this is not an ideal pattern as the spawned tasks act like an unbounded channel in this case. If Requests come in faster then responses then your upper bound is the computers memory because you have no back pressure.

fuel-p2p/src/discovery/discovery_config.rs
```diff
@@ -102,8 +102,11 @@ impl DiscoveryConfig {
         let memory_store = MemoryStore::new(local_peer_id.to_owned());
         let mut kademlia_config = KademliaConfig::default();
         let network = format!("/fuel/kad/{}/kad/1.0.0", network_name);
-        kademlia_config.set_protocol_name(network.as_bytes().to_vec());
+        let network_names = network.as_bytes().to_vec();
+        kademlia_config
+            .set_protocol_names(std::iter::once(network_names.into()).collect());
```
freesig, line 107:
```suggestion
            .set_protocol_names(vec![network_names.into()]);
```

freesig (changes requested):
Looks good from what I understand. I think the mdns might be the wrong feature though

fuel-p2p/src/service.rs
leviathanbeak: Refer to the (updated) PR description for clarification.

.github/workflows/ci.yml
leviathanbeak: ah that's what I missed, thanks @ControlCplusControlV

fuel-p2p/Cargo.toml
```diff
@@ -18,8 +18,8 @@ fuel-core-interfaces = { path = "../fuel-core-interfaces", features = ["serde"],
 futures = "0.3"
 futures-timer = "3.0"
 ip_network = "0.4"
-libp2p = { version = "0.44", default-features = false, features = [
-    "dns-async-std", "gossipsub", "identify", "kad", "mdns", "mplex", "noise",
+libp2p = { version = "0.48", default-features = false, features = [
+    "dns-async-std", "gossipsub", "identify", "kad", "mdns-async-io", "mplex", "noise",
```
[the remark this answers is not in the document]
leviathanbeak: Not necessarily, we currently use `dns-async`, `tcp-async` and `mdn-async` so should be fine, we might want to move them to tokio but previously it wasn't as stable as the async ones.

fuel-p2p/src/service.rs
leviathanbeak: @xgreenx good point, created a separate issue for it https://github.com/FuelLabs/fuel-core/issues/669

fuel-p2p/src/orchestrator.rs
leviathanbeak: good point, we can do that in another PR, don't want to touch it now since it's not related to this PR + there might be some changes how the orchestrator works because of the design changes we are introducing

.github/workflows/ci.yml
```diff
@@ -109,7 +109,11 @@ jobs:
       - name: Install Cargo Make
         uses: davidB/rust-cargo-make@v1
         with:
-          version: "0.36.0"
+          version: "0.36.0"      
+      - name: Install Protoc
```
Voxelot, line 113: Will this require devs to have protoc installed on their machines to build the repo? If so, that should be added to the CONTRIBUTING.md and README.md steps.

fuel-p2p/src/config.rs
```diff
@@ -122,8 +126,14 @@ pub(crate) async fn build_transport(
...
     let transport = {
-        let tcp = libp2p::tcp::TcpConfig::new().nodelay(true);
-        let ws_tcp = libp2p::websocket::WsConfig::new(tcp.clone()).or_transport(tcp);
+        let generate_tcp_transpot =
+            || TcpTransport::new(GenTcpConfig::new().port_reuse(true).nodelay(true));
+
+        let tcp = generate_tcp_transpot();
+
+        let ws_tcp =
+            libp2p::websocket::WsConfig::new(generate_tcp_transpot()).or_transport(tcp);
```
Voxelot, line 135:
```suggestion
            libp2p::websocket::WsConfig::new(generate_tcp_transport()).or_transport(tcp);
```

fuel-p2p/Cargo.toml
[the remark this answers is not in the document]
freesig: Won't this be running as part of the fuel-core binary though? Which does use tokio. The problem with using the wrong runtime is you end up pulling in the whole async-io runtime which really increases compile times.

Voxelot:
This issue existed before the upgrade, but when I startup fuel-core with p2p enabled:
```
cargo run -p fuel-core --features p2p -- run --network test --enable_mdns
```

and then use `ctrl-c` to exit, I just a bunch of these errors:
```
2022-10-04T03:01:20.376180Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376216Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376254Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376295Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376337Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376349Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376383Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376425Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376437Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376445Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376482Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376495Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376502Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376540Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376579Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376617Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent
2022-10-04T03:01:20.376655Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent

```

.github/workflows/ci.yml
leviathanbeak: yes it will, I will add the steps in the md files

leviathanbeak:
> This issue existed before the upgrade, but when I startup fuel-core with p2p enabled:
> 
> ```
> cargo run -p fuel-core --features p2p -- run --network test --enable_mdns
> ```
> 
> and then use `ctrl-c` to exit, I get a bunch of these errors:
> 
> ```
> 2022-10-04T03:01:20.376180Z  WARN fuel-libp2p: 153: Failed to receive P2PRequestEvent

...
@Voxelot `P2PRequestEvent` is the Event that Network Orchestrator receives from other fuel-core modules.
Meaning, this is not related to p2p itself, but the request is coming from one of the other modules and it seems it's not handled in a best way possible. Network Orchestrator is receiving `None` as a request while shutting down - this would be more of a task for PRs similar to https://github.com/FuelLabs/fuel-core/pull/590

Voxelot (approved):

freesig (approved):
I don't fully understand this code but with that in mind it looks good.

PR #663, opened 2022-10-01
[RFC] fuel-core architecture proposal

Voxelot:
fuel-core has been struggling with the current channel-based architecture. In addition, many of the original interface design choices make less sense now that we are taking a staged release approach with configurable consensus modules (poa vs pos vs rollup mode). It's time to revise our strategy and take a path that will be more stable in the long term and addresses our current pain points.

Main highlights:
 - no channels in public interfaces
 - module-local traits instead of globally shared traits
 - avoidance of duplicate or wasteful block execution
 - better unit testability
 - improved parallelizability of tasks within the client


docs/architecture.md
```diff
@@ -0,0 +1,331 @@
...
+#### Ports: fuel_txpool::ports
+```rust
+trait PeerToPeer {
+    async fn broadcast_transaction(transaction: Transaction) -> Result<()>;
+    fn report_invalid_tx(peer_id: PeerId) -> Result<()>;
+    // handler for gossipped transactions
+    fn on_new_tx(tx: Transaction, peer: PeerId);
+}
+```
+
```
ControlCplusControlV, line 256: Does fuel-txpool need anything else specifying how blocks are produced? I was thinking the other day it might be nice to have some way for block production to be externally delegated to something over a public interface if made accessible. This greatly helps with a PBS or similar structure, however would then require changes in the public interface.

I think it could be very valuable to allow TxPool to delegate a `MakeBlock` function so that custom block building can easily integrate inside the txpool. Although we could also undertake this with a similar internal change of just having some `BlockProduction` Interface, and allowing people to swap in their own custom production methods via that.

I think it's worth weighing the 2 options, as a public interface allows more complex block building (especially if someone has tx's their txpool doesn't know about yet with a txpool aggregator across peergroups) and it also helps with something like PBS, but an inner strategy function avoids needing to have a public interface and is more simplistic.

docs/architecture.md
```diff
@@ -0,0 +1,331 @@
...
+### Executor
+
+#### Ports: fuel_block_executor::ports
+```rust
+trait Database: IntepreterStorage {
+    fn get_coin(id: UtxoId) -> Result<Coin>;
+    fn upsert_coin(coin: Coin) -> Result<()>;
```
ControlCplusControlV, line 274: What exactly does `upsert` mean? Is this to refer to replacing an old UTXO which has now been spent with a new one?

docs/architecture.md
```diff
@@ -0,0 +1,331 @@
...
+}
+```
+
+#### Adapters: fuel_core::service::adapters::relayer
+```rust
+impl fuel_block_importer::ports::Relayer for Service<Relayer> {
+    // insert impl here
+}
+
+impl fuel_block_producer::ports::Relayer for Service<Relayer> {
```
ControlCplusControlV, line 313: Does fuel-block-producer need anything else specifying how blocks are produced? I was thinking the other day it might be nice to have some way for block production to be externally delegated to something over a public interface if made accessible. This greatly helps with a PBS or similar structure, however would then require changes in the public interface.

I think it could be very valuable to allow TxPool to delegate a `MakeBlock` function so that custom block building can easily integrate inside the txpool. Although we could also undertake this with a similar internal change of just having some `BlockProduction` Interface, and allowing people to swap in their own custom production methods via that.

I think it's worth weighing the 2 options, as a public interface allows more complex block building (especially if someone has tx's their txpool doesn't know about yet with a txpool aggregator across peergroups) and it also helps with something like PBS, but an inner strategy function avoids needing to have a public interface and is more simplistic.

freesig:
Overall this looks good. The approach has good abstraction and allows the implementation to specify the details of things like synchronization. Just a few notes:
- I'm not yet convinced that the `Database` trait doesn't need to be `async`. There is still ultimately only a single WAL writer so as far as I can tell if we push data to the db faster than it can flush then we will very quick run out of memory and start blocking on needing to flush. If we don't test this carefully before assuming we can use the database synchronously within async calls then we will ultimately end up blocking the tokio runtime threads and the whole system will grind to a halt. I think the careful approach is to keep the database trait async until we make a workload benchmark that shows this is not a concern. It's always easier to remove async then add it in later.
- A limitation of this design is that the largest level of atomicity is the trait methods. It's not possible to call multiple of these trait methods in a way where they succeed or fail together and writes are made in the same transaction etc. This is because the synchronization will always take place within the methods and locking across multiple methods will result in dead locks / leak out coordination from the abstraction. We should at least ask is there any situation where any of the ports methods will need to be called with the same view of state or commit as part of the same transaction.

docs/architecture.md
xgreenx: What is PBS?=D

docs/architecture.md
```diff
@@ -0,0 +1,331 @@
...
+    fn get_current_height() -> Result<BlockHeight>;
+    // insert consensus information associated with a sealed block that was locally produced
+    fn insert_consensus_data(sealed_block: &SealedFuelBlock) -> Result<()>;
+}
+
+trait PeerToPeer {
+    // broadcast a newly produced block to the network
+    async fn broadcast_new_block(block: SealedFuelBlock) -> Result<()>;
+    // receive a newly produced block from the network
+    async fn on_new_block(block: SealedFuelBlock) -> Result<()>;
```
xgreenx, line 116: I guess you meant the next signature:
```rust
async fn on_new_block() -> Result<SealedFuelBlock>;
```

Why do we need to monitor the block on the CM side, if Synchronizer does the same?

docs/architecture.md
```diff
@@ -0,0 +1,331 @@
...
+```
+
+### Synchronizer
+
+#### Ports: fuel_sync::ports
+```rust
+trait PeerToPeer {
+    async fn fetch_best_peer_height() -> Result<(PeerId, BlockHeight)>;
+    async fn fetch_block_headers(query: Range<BlockHeight>) -> Result<Vec<(PeerId, SealedFuelBlockHeader)>>;
+    async fn fetch_block_data(id: block_id) -> Result<Vec<(PeerId, FuelBlockBody)>>;
```
xgreenx, line 196:
```suggestion
pub struct Entity<Data> {
    pub data: Data,
    pub sender: PeerId,
};

trait PeerToPeer {
    async fn fetch_best_network_block_header() -> Result<Entity<SealedFuelBlockHeader>>;
    async fn fetch_blocks(query: Range<BlockHeight>) -> Result<Entity<Vec<SealedFuelBlock>>>;
```

docs/architecture.md
```diff
@@ -0,0 +1,331 @@
...
+dependencies, and also allows for lightweight rapid developer iteration within a domain module. It also ensures
+the purity of the domain modules, preventing implementation details of modules from contaminating each other. Adapters
+are intended to restrict the viral spread of any ugly complexity that arises when integrating modules. Decisions
+such as whether to use channels, networking, IPC, or direct calls between domain modules can be decided in the adapter
+without any effect on the domain logic. They should avoid implementing any business rules, and only contain integration
+glue logic between systems.
+
+Special care was also taken in this design approach to keep implementation details regarding synchronization primitives
+or proprietary async runtime requirements out of any public interfaces (No channels!).
+
```
xgreenx, line 29: It will be good to add a description that all primary implementation of those traits should be done in the corresponding services and reused in the `fuel-core`. Because our main goal is that everyone can take any of our services(separate crate), and it should provide most of the required functionality.

Example:

If we need the list of all consensus participants from the database, it should already have a public API that returns it.

Or

If we need to implement `broadcast_transaction`, P2P should provide some method `broadcast`, and we can reuse it in the `fuel-core`.

Because implementation of all business logic in the `fuel-core` will make our crate system useless then=)

docs/architecture.md
```diff
@@ -0,0 +1,331 @@
...
+        end
+        subgraph ES[Executor Service]
+            ESD{{Domain}}
+            BPEXE[Block Producer Adapter]
+            BIEXE[Block Importer Adapter]
+        end
+        BID & BPD --> ES & RS & TXS
+        SSD --> POAS & P2PS & BIS
+        TXD --> P2PS
+        POAD --> P2PS & RS & BPS & BIS
```
xgreenx, line 77: https://github.com/FuelLabs/fuel-core/pull/666

xgreenx:
> I'm not yet convinced that the Database trait doesn't need to be async. There is still ultimately only a single WAL writer so as far as I can tell if we push data to the db faster than it can flush then we will very quick run out of memory and start blocking on needing to flush. If we don't test this carefully before assuming we can use the database synchronously within async calls then we will ultimately end up blocking the tokio runtime threads and the whole system will grind to a halt. I think the careful approach is to keep the database trait async until we make a workload benchmark that shows this is not a concern. It's always easier to remove async then add it in later.

All methods in the `RocksDB` are sync, so adding async to them does nothing=) We have only one write operation in the blockchain during the `commit` of the block. All other operations are read.

Voxelot:
> * A limitation of this design is that the largest level of atomicity is the trait methods. It's not possible to call multiple of these trait methods in a way where they succeed or fail together and writes are made in the same transaction etc. This is because the synchronization will always take place within the methods and locking across multiple methods will result in dead locks / leak out coordination from the abstraction. We should at least ask is there any situation where any of the ports methods will need to be called with the same view of state or commit as part of the same transaction.

A lot of this design is relying on the assumption that we will have instant finality to avoid versioning issues or potential reorgs. The biggest risk imo is splitting writes for block state diffs vs sealed consensus metadata. I see two potential ways to address this issue:
- Globally managed transactions behind the scenes (interior to the DB impl) where we can start and stop transactions, without the domain modules being aware.
- Domain methods accept a reference to an instance of a db transaction on a per call basis

bvrooman:
Overall, the proposal looks strong. I like the rigorous separation of concerns and loose coupling. The interfaces and flows look well defined.

Is there a concept of nested transactions? The idea with nested transactions is that you can start a transaction at a high level, and lower level services running adjacently or in parallel can start additional transactions within. If any of the nested transactions fail, the entire hierarchy of transactions fails too. This provides inter-service atomicity. In order to avoid leaking knowledge of the ongoing transactions across services, this would need to be some form of database primitive, where the database itself tracks ongoing transactions.

docs/architecture.md
Dentosal: Upsert means update/insert, meaning that the value is inserted if it doesn't exist, and updated if it does.

freesig:
> * Globally managed transactions behind the scenes (interior to the DB impl) where we can start and stop transactions, without the domain modules being aware.
> 
> * Domain methods accept a reference to an instance of a db transaction on a per call basis

Not a huge fan of the first option because that's basically implicit global state which is kind of asking for trouble.

The second option is better but note I wasn't just referring to the database. It's any internal synchronization, such as lock or channels or whatever. The second option also leaks the internal sync abstraction which we don't really want. 

I think another approach could be that anytime this need arises we simply add another method to the `Port` trait that combines both of the others within a lock on the impl. 
This won't work if we ever need to tie together calls across `Ports` but maybe if that need ever arises we could create a `Port` for a pair of modules?

docs/architecture.md
Voxelot: > Although we could also undertake this with a similar internal change of just having some BlockProduction Interface, and allowing people to swap in their own custom production methods via that.

This is the way to go. The consensus module defines it's own port for block production, and the adapter for it could use our defacto fuel_block_producer implementation or swap it out with something else if wanted.

If it's just a matter of overriding the selection algorithm of transactions, that could also be swapped out by the block producer's adapter for the txpool.

docs/poa/flows.md
```diff
@@ -0,0 +1,107 @@
...
+    D-->>BI: 
+    BI->>+TX: drop_committed_txs
+    TX-->>-BI: 
+    BI-->>-S: 
+    
+```
+
+## PoA Gossip-Sync Flow
+
+When a non-producer is synced to the PoA node, bypass Synchronizer and react to block broadcasts.
```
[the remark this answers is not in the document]
Voxelot: My thinking was that sync would be a polling mechanism for already sealed blocks, whereas gossipsync via the CM would be push-based. 

The motivation for this because when we move to PoS, it will then require additional back and forth between synchronizer and CM to convert an unsealed block into a sealed block, pending enough votes. Since block broadcast messages are very specific to a consensus algorithm, I think it will be easiest to let the CM decide on what to do with them.

The downside is if the synchronizer starts polling for the same block the CM is already in the process of sealing/finalizing. This could be mitigated via some global "is syncing" toggle switch like ethereum and most other chains use.

docs/architecture.md
Voxelot: ah right, this should be impl of the domain, not the port's adapter.

The intention was to have a callback-style interface, which I guess would really look like this:

```rust
async fn on_new_block<F: Fn(SealedFuelBlock)>(block_callback: F);
```

But this may not have the best semantics either, for instance, there's no good reason to be able to register multiple callbacks from the CM perspective.

Perhaps a stream would be a more straightforward interface to consume:

```rust
fn broadcast_blocks() -> impl Stream<SealedFuelBlock>;
```

docs/architecture.md
```diff
@@ -0,0 +1,345 @@
...
+
+### Transaction Pool
+
+#### Ports: fuel_txpool::ports
+```rust
+trait PeerToPeer {
+    async fn broadcast_transaction(transaction: Transaction) -> Result<()>;
+    fn report_invalid_tx(peer_id: PeerId) -> Result<()>;
+    // handler for gossipped transactions
+    fn on_new_tx(tx: Transaction, peer: PeerId);
```
Voxelot, line 267:
```suggestion
    fn received_network_transactions() -> impl Stream<(Transaction, PeerId)>;
```

freesig:
> All methods in the `RocksDB` are sync, so adding async to them does nothing=) We have only one write operation in the blockchain during the `commit` of the block. All other operations are read.

Yeh you would need to put the `RocksDB` within a `spawn_blocking` call and then await that to actually make it async. The reasoning for adding async now is because it would be easier to remove it if it turns out we don't need it later.

Yes there's one write per block but you also have to consider any other service that is writing to the database (like the relayer writes logs) because we are sharing a single database between all the services.

docs/architecture.md
ControlCplusControlV: > What is PBS?=D

PBS is Proposer Builder Seperation, so the idea that people like Flashbots and Manifold finance can build blocks for example that extract maximal MEV, then they can send these blocks to validators, who propose the blocks. Seperating Validators from Block Builders, it's the logical conclusion of Ethereum's MEV state, so I thought it would be worth considering in our new design, and allow for block building to be done separate from proposal

xgreenx:
> Yes there's one write per block but you also have to consider any other service that is writing to the database (like the relayer writes logs) because we are sharing a single database between all the services.

We have one `Database` instance that is used by different services, it is true. But each service has its tables, so write operations don't overlap.

> Yeh you would need to put the RocksDB within a spawn_blocking call and then await that to actually make it async. The reasoning for adding async now is because it would be easier to remove it if it turns out we don't need it later.


For me seems easier to add `async` in the future than remove it=D The database is a fundamental thing and is used in all services. The introduction of `async` means that all methods that use database also should be async -> most of all methods in the code will be async.

If all methods are `async`, the compiler will not force you to use sync methods -> if the method is sync, and you want to use `async` inside(the compiler will not allow it), you need to think twice do you really need `async` there. So usage of sync methods now will prevent you from spawning `async` methods in the code.

docs/architecture.md
xgreenx: In the traits, we can't return `impl`, so it will be `Box<dyn Stream<...>>`.

Also, `Entity` from the suggestion for FS can be a global type, and we can use it here `Stream<Entity<Transaction>>`=)

xgreenx: And another question, do we want to return `Stream` or use `async next_new_tx() -> Result<...>`?

Because we need to use the same pattern in all services=)

docs/poa/flows.md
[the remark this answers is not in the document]
xgreenx: P2P forwards all consensus information to a separate channel/stream to CM. 
- If you are a producer, you collect votes. After the collection of votes for the block, you will commit it. After committing, you will broadcast it to the network.
- If you are not a producer in the PoS but part of a council, you verify the block and send a vote to P2P. You ignore votes.
- If you are not a producer or part of the council, you ignore all information.

FS of each node will receive broadcasted block and will handle synchronization=) 

As we discussed, we can always cache the state of the executed pending block, to speed up the commit. But we don't need to collect votes in the CM(it is possible in the future as part of optimization).

docs/architecture.md
Voxelot: Good call on still needing `BoxStream` (wen GAT?). 

`async next_data_item() -> Result<..>` is also a good option and less complicated. We can just leave it up to each domain-specific adapter to figure out the details of how to reliably provide this data to the domain (e.g. cloning tokio pubsub receivers). This way, the adapter won't be forced to always create a stream if the underlying data source is some other kind of custom future.

The main reason to stick with a stream in the domain interface is that it's the most conventional way to represent an async future that will yield multiple values. However, from a consumer perspective this just comes down to having a `stream.next().await` instead of a `future.await`. Having our own async `port.next_something().await` methods seems fairly similar semantically to calling next on a stream and erases some of the pains of dealing with streams directly.

docs/architecture.md
Voxelot: added the following guidelines about adapters:

- Adapters implement Ports. They are for integrating with the "real-world" dependencies like using other domain modules,
   implementing serialization logic, or calling a database.
 - Adapters should either reside in the top-level crate ("fuel-core") or in their own crates. This prevents the domains
   from being polluted by dependencies with large compile times or unstable interfaces. This allows for libraries like 
   networking, third-party SDKs, or database drivers to be swapped out without breaking our core domain specific logic.

freesig:
> We have one `Database` instance that is used by different services, it is true. But each service has its tables, so write operations don't overlap.

As far as I can tell this is incorrect. There's still only one writer for the `WAL` file and that is still going to be a bottleneck even if services are in different tables.

docs/architecture.md
```diff
@@ -0,0 +1,465 @@
...
+    type GossipedBlock: NetworkData<SealedFuelBlock>;
+    
+    async fn fetch_best_network_block_header() -> Result<Self::SealedHeaderResponse>;
+    async fn fetch_blocks(query: Range<BlockHeight>) -> Result<Self::BlockResponse>;
+    // punish the sender for providing an invalid block header
+    fn report_invalid_block_header(invalid_header: &Self::SealedHeaderResponse) -> Result<()>;
+    // punish the sender for providing a set of blocks that aren't valid
+    fn report_invalid_blocks(invalid_blocks: &Self::BlockResponse) -> Result<()>;
+    // await a newly produced block from the network (similar to stream.next())
+    async fn next_gossiped_block() -> Result<Self::GossipedBlock>;
```
xgreenx, line 232: Do we really want to gossip about the whole `FuelBlock`? And handle it in another channel?

Other peers can gossip about the new height via `fetch_best_network_block_header` and then we can use default logic to process it.

PoA peer -> produce block -> commit block locally -> gossip about new height -> `fetch_best_network_block_header` on non-PoA node -> `fetch_blocks` with new height -> await new block and start process it.

docs/architecture.md
```diff
@@ -0,0 +1,465 @@
...
+    end
+```
+
+### (PoA) Consensus Module
+
+#### Ports: fuel_poa_consensus::ports
+```rust
+trait BlockProducer {
+    // used by the PoA node to produce a new block and immediately save the resultant block state
+    async fn produce_and_store_block(height: BlockHeight) -> Result<FuelBlock>;
```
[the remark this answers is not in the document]
Voxelot: yeah I was thinking that would be more efficient from the producer perspective to avoid handling state_diffs or recomputing the block again in validation mode. That's why I was considering making the PoA module do the block broadcast instead of block importer.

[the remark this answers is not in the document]
Voxelot: Maybe these actions could be consolidated from POA to block importer by having a separate block importer method that just commits to the SealedBlockHeaders without executing the block.

docs/architecture.md
Voxelot: Ah true - the peer might not care about the gossiped message and so it would be wasteful to include the full block.

The only place where we might want to consider broadcasting the full block is during consensus (i.e. POS), to cut down on latency.

docs/architecture.md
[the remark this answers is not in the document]
xgreenx: We will face problems with state diff later when we will implement PoS. Because The block should not be committed while there are not enough votes.

So for example `async fn produce(height: BlockHeight) -> Result<(FuelBlock, StateDiff)>;` may be helpful for CM. CM calls it, receives the block, and after voting gathering, calls `commit(block, Some(state_diff))`(or a separate method).

So in the final design, we will use something like this. For now, we can do double-work and optimize it later.

docs/architecture.md
xgreenx: Yep, but for PoS, we will have a separate channel/stream to receive only blocks/votes related for consensus. FS should not be aware of this data=)

Voxelot: Ah so you're saying we don't need gossip at all and can just poll for headers?

Voxelot: Yeah the full block for consensus would be a separate kind of message than what the synchronizer would subscribe to.

docs/poa/flows.md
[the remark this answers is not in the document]
Voxelot: Updated to use the synchronizer for finalized block gossip

docs/architecture.md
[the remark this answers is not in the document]
xgreenx: > My original goal was to allow the system to work optimally without being adversely affected by uncertainty related to components we haven't built out or designed yet. Personally, I'd prefer some kind of approach that is explicit about what is going on, rather than relying on some magic executor cache to fix things in the future for block production.

I mentioned PoS only to show that we will have this problem in the future=) My main idea is not to have several places where we can commit the block because, in this case, we need to broadcast from both places. `BlockProducer` can use `BlockCommiter` inside, but I don't see how we can do that without caching/metadata.

> Having the block producer return a StateDiff could help, but if we had a way to manage database transactions across services we could bundle other things like consensus metadata as part of one atomic commit


I forgot about the pattern with metadata=D Yea, we can use it. But then we need to have different metadata like `TrustedMetadata` and `Untrusted`, because we previously discussed that other peers could share metadata about execution to speed up the sync process of other nodes. All metadata received from the network is untrusted. All metadata received from the `BlockProducer` is trusted

docs/architecture.md
xgreenx: > Ah so you're saying we don't need gossip at all and can just poll for headers?

Yep, non-PoA peers will receive a block from the synchronizer by default synchronization flow=) In this case we don't need to write any additional logic

Voxelot: So in this case we just remove the PoA sealed block broadcast altogether? I suppose the liveness shouldn't be too terrible assuming our block times are higher than the default synchronizer polling interval.

xgreenx: Yep, block header polling/gossiping should be enough for syncing=)

Voxelot:
> > * Globally managed transactions behind the scenes (interior to the DB impl) where we can start and stop transactions, without the domain modules being aware.
> > * Domain methods accept a reference to an instance of a db transaction on a per call basis
> 
> Not a huge fan of the first option because that's basically implicit global state which is kind of asking for trouble.
> 
> The second option is better but note I wasn't just referring to the database. It's any internal synchronization, such as lock or channels or whatever. The second option also leaks the internal sync abstraction which we don't really want.
> 
> I think another approach could be that anytime this need arises we simply add another method to the `Port` trait that combines both of the others within a lock on the impl. This won't work if we ever need to tie together calls across `Ports` but maybe if that need ever arises we could create a `Port` for a pair of modules?

Composing independent actions with a new port make sense, as the adapter impl could handle things like starting/stopping db transactions or utilizing other concurrency primitives without bubbling those implementation concerns into the business logic. It's also more practical than trying to adopt some top-level concurrency management architecture like actors or STM.

docs/architecture.md
Voxelot: Given this simplification, how about we continue letting block production save to db and the BI commit (which validates blocks) also save to db? The only overlap between POA and BI at that point would be persisting the consensus metadata and the system would be more efficient.

When we implement PoS later, it would have to define it's own block producer port which could return a state_diff instead of immediately persisting. This way we keep the PoA mode efficient, without trying to plan too far ahead and by attempting to align with interfaces/usage patterns only needed for PoS.

xgreenx: I meant we don't need to broadcast the whole block, only the header=D It decreases latency for the network to sync=) Also, reduce the number of polling=D

But yea, we can to not broadcast it to the network=) But still, we need to notify the transaction pool, and Block Importer is a good place for it=)

Voxelot: Sorry for the misunderstanding, broadcasting the finalized block headers makes sense.

Voxelot: ok updated to gossip headers instead of full blocks, keeping the gossipsync flow intact.

also updated the poa to use the imported for committing the finalized state of the block to avoid multiple committers.

xgreenx (approved):

bvrooman:
The proposal looks good to me!

docs/architecture.md
```diff
@@ -0,0 +1,473 @@
...
+
+#### Ports: fuel_sync::ports
+
+```rust
+trait PeerToPeer {
+    type SealedHeaderResponse: NetworkData<SealedFuelBlockHeader>;
+    type BlockResponse: NetworkData<Vec<SealedFuelBlock>>;
+    type GossipedBlockHeader: NetworkData<SealedFuelBlockHeader>;
+    
+    async fn fetch_best_network_block_header() -> Result<Self::SealedHeaderResponse>;
```
freesig, line 233: Is this always guaranteed to return a response or should it be `Option<Result<Self::SealedHeaderResponse>>`?

docs/architecture.md
```diff
@@ -0,0 +1,473 @@
...
+    async fn next_gossiped_block_header() -> Result<Self::SealedFuelBlockHeader>;
+    // notify the p2p network whether to continue gossiping this message to others or
+    // punish the peer that sent it
+    fn notify_gossip_block_validity(message: &Self::GossipedBlockHeader, validity: GossipValidity);
+}
+
+// Generic wrapper for data received from the network
+pub trait NetworkData<D> {
+    // Transfer ownership of the provided data without consuming other message metadata
+    fn take_data(&mut self) -> Option<D>;
```
freesig, line 249: What does this option represent? Is this saying that all network data is optional. If so that answers my above question.

docs/architecture.md
```diff
@@ -0,0 +1,473 @@
...
+#### Adapters: fuel_core::service::adapters::fuel_sync
+_N/A_
+
+### P2P
+
+#### Ports: fuel_p2p::ports
+```rust
+trait Database {
+    // used when other peers are syncing
+    fn get_sealed_block(height: BlockHeight) -> Result<SealedFuelBlock>;
```
freesig, line 286: This seems like it should be optional right? What if a height is requested that doesn't exist yet or is that an error case?

docs/architecture.md
```diff
@@ -0,0 +1,473 @@
...
+```
+
+### Transaction Pool
+
+#### Ports: fuel_txpool::ports
+```rust
+trait PeerToPeer {
+    type GossipedTransaction: NetworkData<Transaction>;
+    // Gossip broadcast a transaction inserted via API.
+    async fn broadcast_transaction(transaction: Transaction);
```
freesig, line 313: What does this async actually wait for? There is no error here so it can't really wait for the broadcast to send or be received (which is failable). Also similar concerns to my above question on broadcast.

docs/architecture.md
```diff
@@ -0,0 +1,473 @@
...
+    // Await the next transaction from network gossip (similar to stream.next()).
+    async fn next_gossiped_transaction(&self) -> Self::GossipedTransaction;
```
freesig, line 315: Won't this need an error case?

freesig: Or maybe that's handled in the `NetworkData` trait?

docs/architecture.md
```diff
@@ -0,0 +1,473 @@
...
+trait Database {
+    // get the current fuel chain height
+    fn get_current_height() -> Result<BlockHeight>;
+    // insert consensus information associated with a sealed block that was locally produced
+    fn insert_consensus_data(sealed_block: &SealedFuelBlock) -> Result<()>;
+}
+
+trait PeerToPeer {
+    // broadcast a newly produced block to the network
+    async fn broadcast_new_block(block: SealedFuelBlock) -> Result<()>;
```
freesig, line 150: Might be worth just adding a comment as to what this function actually awaits. Broadcasts are traditionally best effort but I think the options are:
1. Await a successful connection to N peers.
2. Await a successful response "received" response from N peers.
3. Await on a permit to send a broadcast (to avoid overloading the background broadcast task).
4. Don't await anything (probably not recommended as this acts like an unbounded channel then).
It might depend on the under lying lib on what we can await. IMHO it's best to go option `3.` as `1.` or `2.` can create make this function very slow to return (imagine talking to a peer on a crappy network connection) and it might not be in a part of the system where you want that.

docs/architecture.md
freesig: Also does this timeout / if so do we want to retry (so it would need to be an error case).

docs/architecture.md
```diff
@@ -0,0 +1,473 @@
...
+    // insert impl here
+}
+```
+
+### Executor
+
+#### Ports: fuel_block_executor::ports
+```rust
+trait Database: IntepreterStorage {
+    fn get_coin(id: UtxoId) -> Result<Coin>;
```
freesig, line 361: Should this be optional?

docs/architecture.md
```diff
@@ -0,0 +1,473 @@
...
+    fn upsert_coin(coin: Coin) -> Result<()>;
+    fn get_message(message_id: MessageId) -> Result<Message>;
```
freesig, line 363: optional?

docs/architecture.md
```diff
@@ -0,0 +1,473 @@
...
+    fn mark_message_spent(message_id: MessageId) -> Result<()>;
+    fn get_contract_utxo_id(contract_id: ContractId) -> Result<UtxoId>;
+    fn set_contract_utxo_id(contract_id: ContractId, utxo_id: UtxoId) -> Result<()>;
+    fn get_block(height: BlockHeight) -> Result<FuelBlock>;
```
freesig, line 367: optional?

docs/architecture.md
```diff
@@ -0,0 +1,473 @@
...
+    // insert impl here
+}
+```
+
+### Relayer
+
+#### Ports: fuel_relayer::ports
+```rust
+trait Database: StorageMutate<Messages, Error = KvStoreError> {
+    fn get_current_sealed_block(&self) -> Result<SealedFuelBlock>;
```
freesig, line 395: Will there always be a current sealed block?

docs/architecture.md
```diff
@@ -0,0 +1,473 @@
...
+    fn get_da_synced_height(&self) -> Result<DaBlockHeight>;
+    fn set_da_synced_height(&self, height: DaBlockHeight) -> Result<Option<DaBlockHeight>>;
```
freesig, line 397: What does the return type mean here?

freesig (commented):
Lots of questions around function signatures. Might just be my lack of knowledge showing but happy with everything else / will approve

docs/architecture.md
Voxelot: This should be an Option if you're not connected to any peers

docs/architecture.md
Voxelot: The intention of this type is to hide p2p metadata like network message ids or peer ids needed for reporting back message validity. I could get rid of the option and just have an `AsRef<D>` to access the message payload, but that would force consumers to always have to clone the data.

docs/architecture.md
Voxelot: true, this probably doesn't need to be async

docs/architecture.md
freesig: Is it also possible that none of the peers have any sealed blocks?

docs/architecture.md
freesig: I think I'm still missing something. If it can be `AsRef<D>` why can't it just be `-> D`?

docs/architecture.md
freesig: Maybe only because of overloading the broadcast task, see my other comment https://github.com/FuelLabs/fuel-core/pull/663#discussion_r995207775

docs/architecture.md
Voxelot: Possibly? I was thinking this could be consumed in a similar manner to [steam.next()](https://docs.rs/futures/latest/futures/stream/trait.Stream.html#tymethod.poll_next). Since this should never close while the service is running, I didn't think it'd need an option either.

docs/architecture.md
Voxelot: I think we can actually remove this broadcast, we will use the broadcast from the block importer for PoA instead.

docs/architecture.md
Voxelot: sure, that should also be a `None`

docs/architecture.md
freesig: Yep so it's basically wait forever until the next transaction arrives. Maybe we could log errors internally or something.

docs/architecture.md
freesig: I think the same question applies to that broadcast as well.

docs/architecture.md
Voxelot: It was just a way to be able to move the data out of this wrapper without cloning, but maybe there are better ways.

Ultimately this isn't intended to be the exact design as added in the disclaimer above. I think it will be more clear during implementation.

docs/architecture.md
Voxelot: libp2p actually uses sync interface to put a broadcast request onto the queue. There's likely some parameters within the library we have to tune to address the issues you raised, but that doesn't need to percolate up through these interfaces luckily.

docs/architecture.md
Voxelot: maybe previously saved value? probably not needed

docs/architecture.md
freesig: It would probably be good to have a comment to say what it means when this function returns though as it affects how it should be used.

Voxelot: hmm looking closer, it actually looks libp2p directly does the full broadcast operation (along with all the encoding and networking) from this sync call. Our p2p service has a separately spawned task that manages these broadcast requests, so that will likely require async at this level.

docs/architecture.md
Voxelot: no, but we can just remove this for now

Voxelot:
@freesig just pushed a set of updates, lmk if this addresses all your outstanding questions

docs/architecture.md
freesig: Yeh probably to prevent spawning too many tasks

docs/architecture.md
```diff
@@ -0,0 +1,473 @@
...
+    fn drop_txs(txs: Vec<TxId>) -> Result<()>;
+}
+
+trait Database {
+    // insert consensus information associated with a sealed block
+    fn insert_consensus_data(sealed_block: &SealedFuelBlock) -> Result<()>;
+}
+
+trait PeerToPeer {
+   // broadcast the finalized header to peers who may need to sync
```
freesig, line 207:
```suggestion
   // broadcast the finalized header to peers who may need to sync.
   // Total amount of broadcasts is bounded and this function may await if the queue is full.
   // Successful result does not guarantee the header has been delivered to any peers. 
```

docs/architecture.md
```diff
@@ -0,0 +1,473 @@
...
+}
+```
+
+### Transaction Pool
+
+#### Ports: fuel_txpool::ports
+```rust
+trait PeerToPeer {
+    type GossipedTransaction: NetworkData<Transaction>;
+    // Gossip broadcast a transaction inserted via API.
```
freesig, line 312:
```suggestion
    // Gossip broadcast a transaction inserted via API.
    // Total amount of broadcasts is bounded and this function may await if the queue is full.
   // Successful result does not guarantee the transaction has been delivered to any peers. 
```

freesig (approved):
Just some comment suggestions otherwise all looks good. Nice job on this PR!

Voxelot:
Thanks for all the feedback, glad to see this merged!

PR #665, opened 2022-10-03
Nix flake to generate contract abi bindings

freesig:


Voxelot (approved):
Nice 👌🏻

ControlCplusControlV:
Hmmm, tx-gossiping tests seem to be running into some issue 🤔, don't think its related specifically to this PR but doing some investigating to see

freesig:
@Voxelot Can I get another review when you get a chance

Voxelot (approved):

ControlCplusControlV (commented):
LGTM, I tried replicating on a fresh install and ran into some issues so it would be nice to maybe expand this this in the future for first time users of nix rather than just experienced users.

From a fresh install I tried the commands and regularly ran into use '--extra-experimental-features nix-command` and use '--extra-experimental-features flakes` being required. Then a github permissions issue on `fuel-v2-contracts` (presumably because I haven't setup git with nix right)

freesig:
> LGTM, I tried replicating on a fresh install and ran into some issues so it would be nice to maybe expand this this in the future for first time users of nix rather than just experienced users.
> 
> From a fresh install I tried the commands and regularly ran into use '--extra-experimental-features nix-command` and use '--extra-experimental-features flakes` being required. Then a github permissions issue on `fuel-v2-contracts` (presumably because I haven't setup git with nix right)

Yeh I think the `fuel-v2-contracts` is because the repo is private so you need to have git ssh setup on your machine. I will open an issue to update the comments for this and the experimental features flag.

PR #670, opened 2022-10-03
refactor: Broadcast tx insert to p2p 

bvrooman:
- Removes the `insert_with_broadcast` method
- Replaces call to `insert_with_broadcast` with sequential `insert` followed by broadcast for each inserted tx

This approach reduces some code complexity at the cost of second iteration over the transactions. See the discussion in the parent PR here https://github.com/FuelLabs/fuel-core/pull/590#discussion_r984933174.


ControlCplusControlV:
Checks are failing? I actually would prefer to hold off on this since the txpool is particularly performance sensitive and the double iteration over tx's. It might also be useful later to separate `insert` from `insert_from_broadcast` for when more complex p2p tx broadcasting is introduced (such as broadcasting beyond a peer-depth of 1)

Voxelot:
> I actually would prefer to hold off on this since the txpool is particularly performance sensitive and the double iteration over tx's. It might also be useful later to separate `insert` from `insert_from_broadcast` for when more complex p2p tx broadcasting is introduced (such as broadcasting beyond a peer-depth of 1) but I understand if readability is more important at this time

We should generally try to avoid code duplication, as it makes future refactors more difficult - there might not be any immediate errors or tests that fail if the shared behavior changes in one and not the other. If we really want to optimize out an extra for-loop, there are other ways to do that without maintaining duplicate code.

Voxelot (approved):

PR #683, opened 2022-10-06
Prevent benchmarks from running during standard CI tests

Voxelot:
When VM benchmarks were added to fuel-core, they unintentionally started being included in the standard test pass in all CI runs due to `cargo test --all-targets`, increasing the times a lot. This PR restricts our cargo test parameters so that benchmarks aren't run on every CI workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,9 +95,9 @@ jobs:
           - command: make
             args: check --locked
           - command: test
-            args: --all-targets --all-features --workspace
+            args: --tests --all-features --workspace
           - command: test
-            args: --all-targets --no-default-features --workspace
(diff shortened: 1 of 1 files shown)


ControlCplusControlV (approved):
Had actually been noticing this when running so some tests so will be nice to have. Benchmark snapshots might be nice to have for each merged PR down in the future but I doubt are necessary yet

Voxelot:
> Had actually been noticing this when running so some tests so will be nice to have. Benchmark snapshots might be nice to have for each merged PR down in the future but I doubt are necessary yet

Agreed. Benchmarks would be better as a nightly or weekly CI job, ideally including a step to upload the HTML results in some easily viewable way on github.

Issue #682, opened 2022-10-06
Clean up unused data related to the old relayer


freesig:
Commit 64717897d23f49e1e119d1f2a6e921cafedc4b9a removes several database keys along with the changes to the super traits on the relayer database trait. There is now unused code in fuel core that needs to be found and removed.
- Related: https://github.com/FuelLabs/fuel-core/pull/631#discussion_r988276754

PR #689, opened 2022-10-11
Remove unused types

Voxelot:
closes: #682

Remove a bunch of unused data types that no longer match up with our current spec or roadmap.


freesig (approved):
Nice

ControlCplusControlV (approved):
LGTM, do we have any CI checks for unused types??

Voxelot:
> LGTM, do we have any CI checks for unused types??

Cargo will already provide warnings for unused types that it can detect which block CI. However, if any of those types are marked `pub`, it's impossible for cargo to know whether the type is used or not.

PR #690, opened 2022-10-11
Fix CI check against the lock file being up to date

Voxelot:
I noticed that using cargo make was fixing the lockfile before we had a chance to verify the checked in version by using the `--locked` flag in our CI process. This update prevents the cargo make plugin from running any cargo commands before our task.

This may impact the nightly manifest used by fuel.nix, as our lockfile on master fell out of date temporarily because I accidentally broke this check when moving to cargo make cc @mitchmindtree


Voxelot:
Note: the first failing CI pass on this PR demonstrates the `--locked` check is now working again

ControlCplusControlV (approved):
LGTM

PR #691, opened 2022-10-11
Enable withdrawals

### LINKED ISSUE #567: Add GraphQL API for getting merkle proofs of output messages
In order for integration tests to test messages sent from Fuel to the base chain (Ethereum) there needs to be an endpoint in the GraphQL API that can return a valid merkle proof to relay an output message after it has been committed. A lot of the objects and patterns with this API will be similar to what gets defined for [this issue](https://github.com/FuelLabs/fuel-core/issues/566), although this new merkle proof API will likely need to come first.

There might be some overlap with [this issue](https://github.com/FuelLabs/fuel-core/issues/528) as well.

Here is a rough outline of what this could look like:
```
/*==Objects==*/
MessageProof {
  messageId: MessageId
  root: U256
  key: U256
  numLeaves: U256
  data: [U256]
}

/*==Queries==*/
Query messagesToProve
  arguments
    messageIds: [MessageId]
  fields
    nodes: [MessageProof]

```

freesig:
resolves #567


fuel-core/src/lib.rs
```diff
@@ -2,6 +2,7 @@ pub use fuel_chain_config as chain_config;
 pub mod database;
 pub mod executor;
 pub mod model;
+mod query;
```
ControlCplusControlV, line 5: Why is query located here? Wouldn't it be more suited to be directly in `/schema` since all of its associated functions are a part of a specific query?

fuel-core/src/query.rs
```diff
@@ -0,0 +1,183 @@
...
+    model::{
+        FuelBlockDb,
+        OutputProof,
+    },
+};
+
+use crate::tx_pool::TransactionStatus;
+
+#[cfg(test)]
+mod test;
```
ControlCplusControlV, line 18: Why does query have its own mod with a specific test file? I would rather unit tests be attached directly to the associated functions than breaking out into a `test.rs`

ControlCplusControlV (dismissed):

fuel-client/src/client.rs
```diff
@@ -534,6 +537,24 @@ impl FuelClient {
 
         Ok(messages)
     }
+
+    /// Request a merkle proof of an output message.
+    pub async fn output_proof(
```
Voxelot, line 542: Maybe we should call this `message_output_proof`? As we may have other kinds of proofs in the future.

fuel-core/src/lib.rs
Voxelot: I have been discussing with @freesig and @xgreenx about moving the business logic of these queries out of the graphql handlers so that they can be unit tested and also allow us to support alternate entrypoints other than graphql in the future. Then the graphql handlers will have minimal logic to only accepting parameters and unwrapping types before calling our internal Rust API.

fuel-core/src/query.rs
Voxelot: Structurally it's basically the same as if the tests were in this file, since they need to be enclosed in a test module anyways. Personally, I prefer splitting modules into smaller files to reduce the amount of scrolling when working on something and make it easier to split related code across multiple windows.

fuel-tests/tests/messages.rs
```diff
@@ -207,3 +232,108 @@ async fn messages_empty_results_for_owner_with_no_messages(
...
+    let result = client
+        .output_proof(
+            transaction_id.to_string().as_str(),
+            message_id.to_string().as_str(),
+        )
+        .await
+        .unwrap()
+        .unwrap();
+
+    let mut tree = fuel_gql_client::fuel_merkle::binary::in_memory::MerkleTree::new();
```
Voxelot, line 312: Interesting use of client deps to verify that they are in sync with the rest of fuel-core 🤔

fuel-core/src/query.rs
```diff
@@ -0,0 +1,183 @@
...
+            }
+            Err(e) => {
+                // Boxing is required because of the different iterator types
+                // returned depending on the error case.
+                let iter: Box<dyn Iterator<Item = _>> = Box::new(std::iter::once(Err(e)));
+                iter
+            }
+        })
+        // Take all message ids up to and including the
+        // requested id.
```
Voxelot, line 118: Wouldn't ending the iteration here cause this merkle tree to have a different root than the block? We need to add all message ids in the block to the tree to regenerate the same root as the block header, and then get a proof to the matching root.

fuel-core/src/query.rs
freesig: Yeh definitely a fan of breaking into a file. Keeping the tests inline only really works with very small tests.

fuel-core/src/lib.rs
freesig: I'm happy to move it down now

fuel-tests/tests/messages.rs
freesig: That's the only way to get the dep in this crate

ControlCplusControlV (commented):
Test is failing but other than that didn't notice anything

fuel-core/src/query/output_proof.rs
```diff
@@ -0,0 +1,170 @@
...
+        if message_id == id {
+            // Save the index of this message to use as the proof index.
+            proof_index = Some(index);
+        }
+
+        // Build the merkle tree.
+        tree.push(id.as_ref());
+    }
+
+    // If we found the proof then return the proof.
```
Voxelot, line 133:
```suggestion
    // If we found the proof index then return the proof.
```

fuel-core/src/query/message_proof/test.rs
```diff
@@ -0,0 +1,151 @@
...
+        .with(eq(Bytes32::default()))
+        .returning(|_| Ok(Some(FuelBlockDb::default())));
+
+    let p = message_proof(&data, transaction_id, MESSAGE_ID)
+        .await
+        .unwrap()
+        .unwrap();
+    assert_eq!(p.message.id(), MESSAGE_ID);
+    assert_eq!(p.signature, fuel_crypto::Signature::default());
+    assert_eq!(p.block.id(), FuelBlockDb::default().id());
```
Voxelot, line 150: shouldn't the block id be different than the default one since it should hash the message root?

fuel-tests/tests/messages.rs
```diff
@@ -207,3 +232,108 @@ async fn messages_empty_results_for_owner_with_no_messages(
 
     assert_eq!(result.results.len(), 0);
 }
+
+#[tokio::test]
+async fn can_get_message_proof() {
+    let transaction_id: TransactionId = Bytes32::default().into();
+    let message_id: fuel_core::schema::scalars::MessageId =
+        fuel_gql_client::fuel_types::MessageId::default().into();
```
Voxelot, line 240: maybe we should use a non-default value for the message ids?

fuel-tests/tests/messages.rs
```diff
@@ -207,3 +232,108 @@ async fn messages_empty_results_for_owner_with_no_messages(
...
+        .await
+        .unwrap()
+        .unwrap();
+
+    let mut tree = fuel_gql_client::fuel_merkle::binary::in_memory::MerkleTree::new();
+    tree.push(fuel_gql_client::fuel_types::MessageId::from(message_id).as_ref());
+    let (expected_root, expected_set) = tree.prove(0).unwrap();
+    assert_eq!(
+        *fuel_gql_client::fuel_types::Bytes32::from(result.proof_root),
+        expected_root
```
Voxelot, line 317: should we test that the proof root also matches the block message root?

freesig: But that would only test we actually passed in the same block on line `286` which has nothing to do with what we are testing.

fuel-client/src/client/schema/message.rs
```diff
@@ -82,16 +82,22 @@ pub struct MessageProofQuery {
...
+    /// The messages sender address.
+    pub sender: Address,
+    /// The messages recipient address.
+    pub recipient: Address,
+    /// The nonce from the message.
+    pub nonce: Bytes32,
+    /// The amount from the message.
+    pub amount: U64,
+    /// The data from the message.
+    pub data: Vec<i32>,
```
Voxelot, line 100: We typically use hex encoding for byte arrays:

```suggestion
    pub data: HexString
```

fuel-tests/tests/messages.rs
```diff
@@ -207,3 +207,145 @@ async fn messages_empty_results_for_owner_with_no_messages(
...
+    .collect::<Vec<u8>>()
+    .into();
+
+    let salt = Salt::zeroed();
+    let contract = Contract::from(bytecode.as_ref());
+    let root = contract.root();
+    let state_root = Contract::initial_state_root(std::iter::empty());
+    let id = contract.id(&salt, &root, &state_root);
+    let output = Output::contract_created(id.clone(), state_root.clone());
+    let txn = TransactionBuilder::create(bytecode, Salt::zeroed(), Default::default())
```
Voxelot, line 237:
```suggestion
    let txn = TransactionBuilder::create(bytecode, salt, Default::default())
```

fuel-tests/tests/messages.rs
```diff
@@ -207,3 +207,145 @@ async fn messages_empty_results_for_owner_with_no_messages(
...
+
+    let config = Config::local_node();
+
+    let transaction_id = script.id();
+
+    // setup server & client
+    let srv = FuelService::new_node(config).await.unwrap();
+    let client = FuelClient::from(srv.bound_address);
+
+    client.submit(&txn).await.unwrap();
```
Voxelot, line 296:
```suggestion
    client.submit(&deploy_txn).await.unwrap();
```

fuel-core/src/schema/message.rs
```diff
@@ -179,4 +196,129 @@ impl MessageQuery {
...
+            .map(|block| block.into_owned().transactions)
+            .unwrap_or_else(|| Vec::with_capacity(0)))
+    }
+
+    fn signature(
+        &self,
+        _block_id: &fuel_core_interfaces::common::prelude::Bytes32,
+    ) -> Result<Option<fuel_core_interfaces::common::fuel_crypto::Signature>, KvStoreError>
+    {
+        Ok(Some(Signature::default()))
```
Voxelot, line 311: this shouldn't need to be default anymore

fuel-tests/tests/messages.rs
```diff
@@ -217,3 +218,303 @@ async fn messages_empty_results_for_owner_with_no_messages(
...
+    // TODO: Verify using merkle tree verify().
+    true
+}
+
+fn verify_signature(
+    _block_id: fuel_gql_client::prelude::Bytes32,
+    _signature: fuel_gql_client::client::schema::Signature,
+) -> bool {
+    // TODO: Verify using Signature verify() once we actually start signing headers.
+    true
```
Voxelot, line 519: this should be able to do a real signature verification now

fuel-tests/tests/messages.rs
```diff
@@ -217,3 +218,303 @@ async fn messages_empty_results_for_owner_with_no_messages(
...
+        .collect();
+
+        // Call contract script.
+        let script = vec![
+            // Save the ptr to the script data to register 16.
+            // This will be used to read the contract id + two
+            // empty params. So 32 + 8 + 8.
+            Opcode::gtf(0x10, 0x00, GTFArgs::ScriptData),
+            // // Set register 17 to zeros which is the base asset id.
+            // Opcode::MOVI(0x11, 1_000),
```
Voxelot, line 320:
```suggestion
```

fuel-tests/tests/messages.rs
```diff
@@ -217,3 +218,303 @@ async fn messages_empty_results_for_owner_with_no_messages(
...
+        .unwrap()
+        .unwrap();
+
+        tokio::time::timeout(
+            std::time::Duration::from_secs(5),
+            client.await_transaction_commit(&call_id.to_string()),
+        )
+        .await
+        .unwrap()
+        .unwrap();
```
Voxelot, line 406:
```suggestion
        let deploy_id = client.submit(&contract_deploy).await.unwrap();
        tokio::time::timeout(
            std::time::Duration::from_secs(5),
            client.await_transaction_commit(&deploy_id.to_string()),
        )
        .await
        .unwrap()
        .unwrap();

        // Call the contract.
        let call_id = client.submit(&script).await.unwrap();
        tokio::time::timeout(
            std::time::Duration::from_secs(5),
            client.await_transaction_commit(&call_id.to_string()),
        )
        .await
        .unwrap()
        .unwrap();
```

Voxelot (approved):

Issue #673, opened 2022-10-04
SealedFuelBlock DB model


Voxelot:
- Make sure this is forward-compatible with multi-producer and PoS systems
- We need at least: The signature, related nonce if any, and information on what node signed the block
- Determine related database columns

freesig:
Note, I moved the `Decide fields that are to be added to a SealedFuelBlock` #674

PR #692, opened 2022-10-11
Reconcile Fuel Block Header

### LINKED ISSUE #674: Reconcile FuelBlockHeader with specs
Make fuel block header match the specs so that they can be authenticated on ethereum for messaging.

Some things to change:

- remove producer field
- add message output root & actually set it after block execution

freesig:
resolves #674


fuel-core/src/database.rs
```diff
@@ -301,10 +301,11 @@ impl InterpreterStorage for Database {
 
     fn timestamp(&self, height: u32) -> Result<Word, Self::DataError> {
         let id = self.block_hash(height)?;
-        let block = self.storage::<FuelBlocks>().get(&id)?.unwrap_or_default();
+        // let block = self.storage::<FuelBlocks>().get(&id)?.unwrap_or_default();
```
freesig, line 304: Not sure what to do here?

fuel-core/src/database.rs
```diff
@@ -392,7 +391,8 @@ mod relayer {
             _height: BlockHeight,
         ) -> Option<Arc<SealedFuelBlock>> {
             // TODO
-            Some(Arc::new(SealedFuelBlock::default()))
+            // Some(Arc::new(SealedFuelBlock::default()))
```
freesig, line 394: Also blocked by #673

fuel-core/src/executor.rs
```diff
@@ -102,34 +100,42 @@ impl Executor {
...
-        let pre_exec_block_id = match mode {
-            ExecutionMode::Production => Default::default(),
-            ExecutionMode::Validation => block.id(),
-        };
+    ) -> Result<FuelBlock, Error> {
+        // // Compute the block id before execution, if mode is set to production just use zeroed id.
+        // let pre_exec_block_id = match mode {
+        //     ExecutionMode::Production => Default::default(),
+        //     ExecutionMode::Validation => block.id(),
+        // };
```
freesig, line 138: This doesn't make sense anymore because a partial block doesn't have an id.

fuel-core/src/executor.rs
```diff
@@ -234,8 +240,8 @@ impl Executor {
             // TODO: reference the bytes directly from VM memory to save serialization. This isn't
             //       possible atm because the change output values are set on the tx instance in the vm
             //       and not also on the in-memory representation of the tx.
-            let tx_bytes = vm_result.tx().clone().to_bytes();
-            txs_merkle.push(&tx_bytes);
+            // let tx_bytes = vm_result.tx().clone().to_bytes();
+            // txs_merkle.push(&tx_bytes);
```
freesig, line 244: This was actually different to the specs as it's meant to be a merkle tree of transaction ids and not just the bytes. Also now this is done in the `generate()` function.

fuel-core/src/executor.rs
```diff
@@ -299,46 +305,52 @@ impl Executor {
...
+        // match mode {
+        //     ExecutionMode::Production => {
+        //         block.header.transactions_root = txs_root;
+        //     }
+        //     ExecutionMode::Validation => {
+        //         if block.header.transactions_root != txs_root {
+        //             return Err(Error::InvalidTransactionRoot)
+        //         }
+        //     }
+        // }
```
freesig, line 342: This no longer makes sense as the root is always generated.

fuel-core/src/executor.rs
```diff
@@ -299,46 +305,52 @@ impl Executor {
...
-            // In theory this shouldn't happen since any deviance in the block should've already
-            // been checked by now.
-            return Err(Error::InvalidBlockId)
-        }
+        // // check if block id doesn't match proposed block id
+        // if mode == ExecutionMode::Validation && pre_exec_block_id != finalized_block_id {
+        //     // In theory this shouldn't happen since any deviance in the block should've already
+        //     // been checked by now.
+        //     return Err(Error::InvalidBlockId)
+        // }
```
freesig, line 353: Again the partial block has no id to compare to.

fuel-core/src/schema/chain.rs
```diff
@@ -93,7 +93,8 @@ impl ChainInfo {
         let db = ctx.data_unchecked::<Database>().clone();
         let height = db.get_block_height()?.unwrap_or_default();
         let id = db.get_block_id(height)?.unwrap_or_default();
-        let block = db.storage::<FuelBlocks>().get(&id)?.unwrap_or_default();
+        // let block = db.storage::<FuelBlocks>().get(&id)?.unwrap_or_default();
+        let block = db.storage::<FuelBlocks>().get(&id)?.unwrap();
```
freesig, line 97: Also not sure what to do here

fuel-core/src/database.rs
Voxelot: Hmm yeah there is some edge cases that need to be handled here:

- If we request `block_hash(0)`, it will just return a default Bytes32::zeroed(). Since we temporarily store the currently executing block in the db at this same hash while the txs are being mutated, this will always return the current time. This is likely incorrect, as `timestamp(0)` should probably reflect the time of the genesis block (which would have to come from the chain_config instead of the DB).
- If we request the time of some block in the future that doesn't exist yet, (e.g.`timestamp(10)` when current height is 1), this will also return a zeroed block id which would refer to the currently executing block's timestamp. However, according to the specs, requesting the time of a block in the future should cause a [panic]

(https://github.com/FuelLabs/fuel-specs/blob/master/specs/vm/instruction_set.md#time-timstamp-at-height):
  > Panic if:
  >  - $rB is greater than the current block height.
  - Would need to check if this path is also triggered when the time is requested for the current block height, depending on whether or not the height index is updated to point current_height -> Bytes32::zeroed() during block execution.
- If we call these methods outside of the VM / block execution context, then both of these cases would return a default timestamp rather than the currently executing block, since there will be no block with a hash of Bytes32::zeroed() in the DB. Moving this to unwrap() would cause a panic, when we should either use genesis block time if height is 0, or a well-formed error if height > current height. The only case where a panic would make sense is when 0 < height <= current_block_height, since the block should definitely exist.

My suggestion would be to leave the behavior as it was, since addressing these edge cases seems orthogonal to updating the block header definition. We can file a bug to investigate and fix any behavioral discrepancies in a future PR.

fuel-core/src/database.rs
```diff
@@ -316,10 +317,8 @@ impl InterpreterStorage for Database {
     }
 
     fn coinbase(&self) -> Result<Address, Error> {
-        let height = self.get_block_height()?.unwrap_or_default();
-        let id = self.block_hash(height.into())?;
-        let block = self.storage::<FuelBlocks>().get(&id)?.unwrap_or_default();
-        Ok(block.header.producer)
+        // FIXME: Get producer address from block signature.
```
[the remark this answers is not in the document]
Voxelot: It's fine to leave this as default for this PR since that was the same behavior as before.

fuel-core/src/database.rs
Voxelot: Although this is no longer needed by the relayer correct?

fuel-core/src/executor.rs
Voxelot: Hmm actually I think it might've been correct to use the mutated tx bytes from the VM. This is because this is our only way to obtain a commitment to the resultant state of executing the tx, whereas the transaction ID doesn't include any of the malleable fields. Without a commitment to the actual transaction bytes, things like change amounts calculated by the VM wouldn't be provable between nodes or on Ethereum.

fuel-core/src/executor.rs
Voxelot: Whoops meant to post my other comment here instead

https://github.com/FuelLabs/fuel-core/pull/692/files#r993910492

fuel-core/src/database.rs
freesig: Ok I will file an issue for this

Voxelot:
To summarize our call earlier:

- Instead of ExecutionMode, the executor will accept a fuel block containing either a partial header or a full header.
  - If we only have a partial header, the executor will run in "production" mode and compute the full header values
    - we may need to spoof a full header into the current execution context in order for VM scripts to able to retrieve things like block height or timestamp from the partial header.
  - If we receive a full header, the executor will run in "validation" mode and compare the computed header with the provided full header.
 - We should still support transmitting full block headers over the network for light clients and other sync optimizations

fuel-core/src/executor.rs
freesig: Ok added this test back.

fuel-core/src/executor.rs
```diff
@@ -103,51 +115,136 @@ impl Executor {
...
+            if block.header().transactions_root != pre_exec_txs_root {
+                return Err(Error::InvalidTransactionRoot)
+            }
+        }
+
+        let finalized_block_id = block.id();
+
+        debug!(
+            "Block {:#x} fees: {}",
+            pre_exec_block_id.unwrap_or_default(),
```
Voxelot, line 183: Should this use the finalized id if the pre-exec block ID is none (e.g. production mode of partial header)?

```suggestion
            pre_exec_block_id.unwrap_or(finalized_block_id),
```

fuel-core/src/executor.rs
```diff
@@ -229,27 +326,22 @@ impl Executor {
...
-
-            match mode {
-                ExecutionMode::Validation => {
+            execution_data.coinbase = execution_data
+                .coinbase
+                .checked_add(tx_fee)
+                .ok_or(Error::FeeOverflow)?;
+
+            // Check or set the executed transaction.
+            match execution_kind.check_or_set(tx) {
```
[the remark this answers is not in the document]
freesig: Yeh I was trying to somehow represent that we only need mutable access on production but this was the best I could come up with. I'm happy to remove it though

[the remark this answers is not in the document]
freesig: Removed this

fuel-core-interfaces/src/model/block.rs
```diff
@@ -14,136 +12,510 @@ use crate::{
...
+    }
 }
 
-/// This structure is created as placeholder for future usage.
 #[derive(Clone, Debug, Default)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-pub struct FuelBlockConsensus {}
+/// The consensus related data that doesn't live on the
+/// header.
+pub struct FuelBlockConsensus {
```
Voxelot, line 510: What about making this an enum for backward compatibility as we add new consensus modes?

freesig: Do you mean backward compatibility with production data that's living on other older fuel nodes?
We would have to make it an `#[non_exhaustive]` enum for it to be backwards compatible. Do we want to do this now for POA when we will be the only node running and can just migrate the database?

fuel-core-interfaces/src/executor.rs
```diff
@@ -13,26 +13,50 @@ use crate::{
...
+/// Validation starts with a full [`FuelBlock`].
+pub type ExecutionBlock = ExecutionTypes<PartialFuelBlock, FuelBlock>;
+
+/// Execution wrapper with only a single type.
+pub type ExecutionType<T> = ExecutionTypes<T, T>;
+
+/// Execution wrapper for setting or checking a type.
+/// Production sets the mutable reference.
+/// Validation checks the reference.
+pub type ExecutionCheckOrSet<'a, T> = ExecutionTypes<&'a mut T, &'a T>;
```
Voxelot, line 36: unused

fuel-core-interfaces/src/model/block.rs
freesig: TODO: Make trait for behavior that must be consistent across all variants of this enum

freesig: Ok I added the enum but the trait doesn't make sense because we can't implement it in the interfaces module

fuel-core/src/database.rs
```diff
@@ -323,10 +329,23 @@ impl InterpreterStorage for Database {
...
+                // FIXME: Get producer address from block signature.
+                // block_id -> Signature
+                let signature = Signature::default();
+                let message = unsafe {
+                    fuel_core_interfaces::common::fuel_crypto::Message::from_bytes_unchecked(
+                block.header.hash().into(),
+            )
+                };
+                let address = signature.recover(&message).unwrap_or_default();
+                Ok((*address.hash()).into())
```
Voxelot, line 346:
```suggestion
                // TODO: throw an error if public key isn't recoverable
                //  when implementing signing (https://github.com/FuelLabs/fuel-core/issues/668)
                let public_key = signature.recover(&message).unwrap_or_default();
                Ok(crate::fuel_tx::Input::owner(&public_key))
```

fuel-core/src/database.rs
```diff
@@ -323,10 +329,23 @@ impl InterpreterStorage for Database {
...
+        match block.consensus_type() {
+            ConsensusType::PoA => {
+                // FIXME: Get producer address from block signature.
+                // block_id -> Signature
+                let signature = Signature::default();
+                let message = unsafe {
+                    fuel_core_interfaces::common::fuel_crypto::Message::from_bytes_unchecked(
+                block.header.hash().into(),
+            )
+                };
```
Voxelot, line 344: Formatting + use cached block id instead of recalculating it every time

```suggestion
                let message = unsafe {
                    fuel_core_interfaces::common::fuel_crypto::Message::from_bytes_unchecked(
                    block.header.id().into(),
                    )
                };
```

fuel-core/src/database/block.rs
```diff
@@ -95,6 +96,26 @@ impl Database {
         Ok(id)
     }
 
+    /// Get the current block at the head of the chain.
+    pub fn get_current_block(&self) -> Result<Option<Cow<FuelBlockDb>>, Error> {
+        let block_entry: Option<(Vec<u8>, Bytes32)> = self
```
Voxelot, line 101: [nit] Maybe factor this common logic with `get_block_height(&self)`?

fuel-core/src/database.rs
freesig: It just cam from the database though so it will never be set and using the cached one will clone it.

fuel-core/src/executor.rs
```diff
@@ -592,11 +647,13 @@ impl Executor {
...
-        tx: &mut Transaction,
-        mode: &ExecutionMode,
+        tx: &mut ExecutionType<&mut Transaction>,
         db: &Database,
     ) -> Result<(), Error> {
-        if let Transaction::Script { inputs, .. } = tx {
+        // Get the kind of execution.
+        let execution_kind = tx.to_kind();
+
+        if let Transaction::Script { inputs, .. } = tx.deref_mut() {
```
Voxelot, line 656: this ExecutionType wrapper could also be removed since the tx is always mutable and only the kind is really needed.

freesig: I went with a different approach which I think actually captures the fact that the inputs are only mutable in production.

Voxelot (approved):

PR #694, opened 2022-10-12
Add gossipsub message validation

### LINKED ISSUE #679: enable message validation on gossipsub
enable it within `GossipsubConfig` 
and then use `gossipsub`s `report_message_validation_result()` on:
- messages decoded
- messages checked for validity (is tx valid etc)

leviathanbeak:
Closes #679 

In this PR we introduce message validation for our gossipsub messages.

The flow goes like this:
- on new GossipsubMessage received
- check if the format is correct
- if the format is correct, the message is propagated to the modules interested in it
- there it's checked if is valid
- once checked for validity, the module responds with the message validity
- and that is reported to the Gossipsub protocol

In this PR I have also removed unnecessary loop from p2p service's `next_event`.


fuel-core-interfaces/src/p2p.rs
```diff
@@ -8,23 +8,54 @@ use crate::{
     model::ConsensusVote,
 };
 use async_trait::async_trait;
-use std::sync::Arc;
+use serde::{
```
Voxelot, line 11: serde is an optional dependency, and usage needs to be under a `#cfg(feature = "serde")` attribute.

fuel-core-interfaces/src/p2p.rs
```diff
@@ -8,23 +8,54 @@ use crate::{
...
     NewVote(ConsensusVote),
 }
 
+#[derive(Debug, Clone)]
 pub enum BlockBroadcast {
     /// fuel block without consensus data
     NewBlock(FuelBlock),
 }
 
+#[derive(Debug)]
```
Voxelot, line 37: might as well add `Copy, Eq, PartialEq` for convenience?

fuel-p2p/src/orchestrator.rs
```diff
@@ -159,6 +179,42 @@ impl NetworkOrchestrator {
...
+        match p2p_service.report_message_validation_result(&msg_id, &peer_id, acceptance)
+        {
+            Ok(true) => {
+                info!(target: "fuel-libp2p", "Sent a report for MessageId: {} from PeerId: {}", msg_id, peer_id);
+            }
+            Ok(false) => {
+                warn!(target: "fuel-libp2p", "Message with MessageId: {} not found in the Gossipsub Message Cache", msg_id);
+            }
+            Err(e) => {
+                warn!(target: "fuel-libp2p", "Failed to publish Message with MessageId: {} with Error: {:?}", msg_id, e);
```
Voxelot, line 210:
```suggestion
                error!(target: "fuel-libp2p", "Failed to publish Message with MessageId: {} with Error: {:?}", msg_id, e);
```

fuel-p2p/src/orchestrator.rs
```diff
@@ -159,6 +179,42 @@ impl NetworkOrchestrator {
...
+            }
+        }
+    } else {
+        warn!(target: "fuel-libp2p", "Failed to read PeerId from received GossipsubMessageId: {}", msg_id);
```
Voxelot, line 214:
```suggestion
        error!(target: "fuel-libp2p", "Failed to read PeerId from received GossipsubMessageId: {}", msg_id);
```

fuel-p2p/src/orchestrator.rs
```diff
@@ -159,6 +179,42 @@ impl NetworkOrchestrator {
...
+        let acceptance = match acceptance {
+            GossipsubMessageAcceptance::Accept => MessageAcceptance::Accept,
+            GossipsubMessageAcceptance::Reject => MessageAcceptance::Reject,
+            GossipsubMessageAcceptance::Ignore => MessageAcceptance::Ignore,
+        };
+
+        match p2p_service.report_message_validation_result(&msg_id, &peer_id, acceptance)
+        {
+            Ok(true) => {
+                info!(target: "fuel-libp2p", "Sent a report for MessageId: {} from PeerId: {}", msg_id, peer_id);
```
Voxelot, line 204: This should probably be debug or trace level in the normal case as it could flood the logs. Maybe a `warn!` if we are rejecting / ignoring the message?

```suggestion
                debug!(target: "fuel-libp2p", "Sent a report for MessageId: {} from PeerId: {}", msg_id, peer_id);
```

fuel-p2p/src/service.rs
```diff
@@ -239,16 +243,28 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
...
+    /// Handles P2P Events.
+    /// Returns only events that are of interest to the Network Orchestrator.
+    pub async fn next_event(&mut self) -> Option<FuelP2PEvent> {
+        if let SwarmEvent::Behaviour(fuel_behaviour) = self.swarm.select_next_some().await
+        {
+            return self.handle_behaviour_event(fuel_behaviour)
         }
+
+        None
     }
```
Voxelot, line 268: [nit] unnessecary return

```suggestion
            self.handle_behaviour_event(fuel_behaviour)
        } else {
            None
        }
    }
```

fuel-p2p/src/gossipsub/builder.rs
```diff
@@ -35,11 +33,12 @@ pub fn build_gossipsub(local_key: &Keypair, p2p_config: &P2PConfig) -> Gossipsub
         .mesh_n_high(p2p_config.max_mesh_size)
         .message_id_fn(gossip_message_id)
         .fast_message_id_fn(fast_gossip_message_id)
+        .validate_messages()
```
Voxelot, line 36: Should we expose any settings to fine-tune the size and TTL of the message validation memcache?

fuel-txpool/src/service.rs
```diff
@@ -179,11 +181,9 @@ impl Context {
...
                     tokio::spawn( async move {
                         let txpool = txpool.as_ref();
-                        match new_transaction.unwrap() {
-                            TransactionBroadcast::NewTransaction ( tx ) => {
-                                let txs = vec!(Arc::new(tx));
-                                TxPool::insert(txpool, db.as_ref().as_ref(), tx_status_sender, txs).await
-                            }
+                        if let GossipData { data: Some(TransactionBroadcast::NewTransaction ( tx )), .. } =  new_transaction.unwrap() {
+                            let txs = vec!(Arc::new(tx));
+                            TxPool::insert(txpool, db.as_ref().as_ref(), tx_status_sender, txs).await;
```
Voxelot, line 186: Should we implement the message validation response here? Or add a todo for this later?

fuel-p2p/src/gossipsub/builder.rs
```diff
@@ -1,31 +1,29 @@
...
+    Sha224,
     Sha256,
 };
 
 use crate::config::P2PConfig;
 
-pub fn build_gossipsub(local_key: &Keypair, p2p_config: &P2PConfig) -> Gossipsub {
+pub fn build_gossipsub(p2p_config: &P2PConfig) -> Gossipsub {
     let gossip_message_id = move |message: &GossipsubMessage| {
         MessageId::from(&Sha256::digest(&message.data)[..20])
```
Voxelot, line 22: We should hash the entire message for uniqueness, otherwise we may drop messages that are actually different from each other.

fuel-p2p/src/gossipsub/builder.rs
```diff
@@ -1,31 +1,29 @@
...
     };
 
     let fast_gossip_message_id = move |message: &RawGossipsubMessage| {
-        FastMessageId::from(&Sha256::digest(&message.data)[..8])
+        FastMessageId::from(&Sha224::digest(&message.data)[..16])
```
Voxelot, line 26: Since these algorithms will just pad the input out to some multiple of 256 or 224bits this is still doing roughly the same amount of work.

From reading the through the libp2p code, it seems that the fast message ID needs to be unique as well. This is because it creates a 1:1 mapping of fast message ID's to regular message IDs, so if there is a collision between fast message ID's, it will get marked as a duplicate even if the regular message ID is different. It seems the only difference for fast message ids is that they are computed before the message is decompressed, allowing for duplicate checks to skip message decoding / decompression. However, I think we should still probably hash the entire compressed message with something like a sha256 since the uniqueness matters here.

fuel-p2p/src/gossipsub/builder.rs
leviathanbeak: I'm thinking I expose `GossipsubConfig` struct from libp2p, that way we can set all the fields we want, instead of exposing only few ones and hiding the rest

leviathanbeak: Instead of multiple gossipsub related fields in P2pConfig - I have exposed single `gossipsub_config: GossipsubConfig` field instead. I have also exposed `GossipsubConfigBuilder` to make it easy to set it up.

fuel-txpool/src/service.rs
leviathanbeak: I will create an issue for it

leviathanbeak: https://github.com/FuelLabs/fuel-core/issues/704

fuel-core-interfaces/src/p2p.rs
```diff
@@ -40,9 +72,48 @@ pub enum P2pRequestEvent {
...
+            data: Some(data),
+            peer_id: peer_id.into(),
+            message_id: message_id.into(),
+        }
+    }
+}
+
+pub trait NetworkData<T>: Debug + Send {
+    fn take_data(&mut self) -> Option<T>;
+}
```
freesig, line 109:
```suggestion
pub trait NetworkData<T>: Debug + Send {
    type Metadata;
    fn take_data(self) -> (Self::Metadata, T);
}
```
I'm thinking it would be better to actually split the type so that we don't have to handle the option?

fuel-txpool/src/service.rs
```diff
@@ -179,11 +181,9 @@ impl Context {
 
                     tokio::spawn( async move {
                         let txpool = txpool.as_ref();
-                        match new_transaction.unwrap() {
-                            TransactionBroadcast::NewTransaction ( tx ) => {
-                                let txs = vec!(Arc::new(tx));
-                                TxPool::insert(txpool, db.as_ref().as_ref(), tx_status_sender, txs).await
-                            }
+                        if let GossipData { data: Some(TransactionBroadcast::NewTransaction ( tx )), .. } =  new_transaction.unwrap() {
```
freesig, line 184: This is the unwrap that I mean

freesig (changes requested):
Overall looks good. I wonder about the `NetworkData` trait though. Also should there be a test of this new behavior?

fuel-txpool/src/service.rs
leviathanbeak: that `unwrap` is from `Result<TransactionGossipData>` not from the `Option` if that's what you meant

fuel-core-interfaces/src/p2p.rs
Voxelot: hmm maybe we can defer this to a followup task? Ideally in a way that wouldn't require users of this trait to know or care about the concrete type for metadata.

Voxelot (approved):

fuel-core-interfaces/src/p2p.rs
freesig: I did raise this concern in the architecture PR but was told it was an implementation detail. Happy to do a follow up though

Voxelot: yeah I think it's just a matter of having a fully fleshed out proposal of this actually working

Voxelot (approved):

Issue #677, opened 2022-10-04
Integrate block trigger with block producer, and remove graphql driven production


Voxelot:
Eliminate the need for `Executor::submit_txs` using the real block production trigger https://github.com/FuelLabs/fuel-core/pull/599 instead. Transactions must be removed from the txpool when blocks are produced as well.

Debugging and dry run functionaly should be implemented for this system as well, but that can be deferred to a later issue.

Dentosal:
@Voxelot  Should the dry-run functionality still run a single one-transaction block and ignore the block production trigger system completely?  Or should it create a new block with currently available transactions and dry-run them all?

PR #698, opened 2022-10-13
Set eth block height on new blocks

Voxelot:
The da_height header field was being set to default, which would prevent downloaded msgs from eth from being spent. This updates the currently used block generation method to always set the current relayer synced eth height.

Also added an integ test that downloaded messages from eth are spendable on fuel.

This is a bandaid solution until #677 is implemented.

### DIFF
--- a/fuel-core/src/executor.rs
+++ b/fuel-core/src/executor.rs
@@ -60,6 +60,7 @@ use fuel_core_interfaces::{
         FuelBlockHeader,
         Message,
     },
+    relayer::RelayerDb,
 };
 use fuel_storage::{
     StorageAsMut,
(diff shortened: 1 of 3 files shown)


freesig (approved):
Kinda hard to tell what actually changed as it looks like most of the changes are just moving a function but looks fine to me.

fuel-core/src/executor.rs
```diff
@@ -103,13 +104,15 @@ impl Executor {
         // setup and execute block
         let current_height = db.get_block_height()?.unwrap_or_default();
         let current_hash = db.get_block_id(current_height)?.unwrap_or_default();
+        let da_height = db.get_finalized_da_height().await.unwrap_or_default();
```
freesig, line 107: Does default actually make sense here? Like this is saying that the da chain has really finalized block zero no matter what. Is that actually true?

freesig (changes requested):
Actually just noticed something

freesig (dismissed):

freesig (approved):

PR #712, opened 2022-10-17
Integrate block production and the new trigger system

### LINKED ISSUE #677: Integrate block trigger with block producer, and remove graphql driven production

Dentosal:
Closes #677


fuel-core/src/database.rs
```diff
@@ -332,6 +335,20 @@ impl InterpreterStorage for Database {
...
+impl BlockProducerDatabase for Database {
+    fn get_block(
+        &self,
+        fuel_height: BlockHeight,
+    ) -> anyhow::Result<Option<Cow<FuelBlockDb>>> {
+        // let v = self.get(fuel_height.into(), Column::FuelBlocks);
+        // let v = v?;
+
+        let id = self.block_hash(fuel_height.into())?;
+        let block = self.storage::<FuelBlocks>().get(&id)?.unwrap_or_default();
```
Voxelot, line 347: Since the return type is Option<Cow<Block>>, we should return None instead of using `unwrap_or_default()`

fuel-core/src/schema/tx.rs
```diff
@@ -324,6 +324,8 @@ impl TxMutation {
         // for read-only calls.
         utxo_validation: Option<bool>,
     ) -> async_graphql::Result<Vec<receipt::Receipt>> {
+        todo!("Migrate dry-run to the new trigger architecture");
```
Voxelot, line 327: Since this method only returns receipts, I don't think it needs to interact with the orchestrator since this doesn't impact regular block production.

I can make a PR to add this functionality.

Voxelot:
Although I could approve and merge this right now, since both @Dentosal and I worked on this PR we should get some other reviews.

xgreenx (approved):

xgreenx (commented):
The change looks good!=)
I'm not sure why we need this line:

```rust

        // cleanup unfinalized headers (block height + time + producer)
        block_db_transaction
            .deref_mut()
            .storage::<FuelBlocks>()
            .remove(&Bytes32::zeroed())?;
```

And maybe we need to rename `produce_block` into `produce_block_and_commit` or `produce_block_and_execute` to make it clear.

Do we want to move the commit of the block into a separate step in the future?

xgreenx (approved):

Voxelot:
> The change looks good!=) I'm not sure why we need this line:
> 
> ```rust
>         // cleanup unfinalized headers (block height + time + producer)
>         block_db_transaction
>             .deref_mut()
>             .storage::<FuelBlocks>()
>             .remove(&Bytes32::zeroed())?;
> ```
> 
> And maybe we need to rename `produce_block` into `produce_block_and_commit` or `produce_block_and_execute` to make it clear.
> 
> Do we want to move the commit of the block into a separate step in the future?

Yeah this is kind of ugly, but we need a way to reference the current block during execution for some of the VM opcodes to be able to fetch the current timestamp etc. Since the block id isn't known until execution has completed, it is temporarily stored with a block id of 0x00 for the duration of the execution and then unset at the end.

PR #715, opened 2022-10-20
Multi tx blocks

Voxelot:
This adds support for dry_run as well as eliminates the gql based usage of "submit_txs", using the channels instead.

There's a few new test failures due to transactions being committed asynchronously, so I've been working on updating the tests to use a new fuel-client method for awaiting the tx status to change. However, there's still a few strange failure cases to chase down.


Voxelot:
got most of the tests working except for the transaction pool, as they relied on invalid txs in a lot places which aren't compatible with the new validation performed on insert.

Dentosal (approved):
The changeset is quite large, and some test cases are still broken. I'm deferring the actual review to the parent PR, but there are no obvious issues with this approach. It makes no sense to continue work on top of the parent without merging this first, so I'm simply doing so.

Issue #513, opened 2022-07-29
Incorporate CheckedTransaction structure into txpool


Voxelot:
Now that fuel-tx supports the immutable [CheckedTransaction](https://github.com/FuelLabs/fuel-tx/blob/master/src/checked_transaction.rs) pattern for handling verified data, the txpool should fully leverage this new pattern to streamline tx validation across the system.

Replace all Arc<Transaction> references with Arc<CheckedTransaction>, so that the block producer is assured that it is only including fully valid and ready to execute transactions.

Additionally, the txpool should verify any predicates of CheckedTransactions before inserting them.

PR #702, opened 2022-10-14
Make dry_run return transparent Receipt

Salka1988:
blocked by https://github.com/FuelLabs/fuel-vm/pull/233

PR #716, opened 2022-10-20
Refactoring of getters/setters/offset and adapt `fuel-core`

xgreenx:
Better to merge https://github.com/FuelLabs/fuel-core/pull/702 first before merging this change because the new version of `fuel-vm` also has other changes.

Applied changes from new `fuel-tx` and `fuel-vm`:
- Usage of `Checked<Tx>` type instead of `CheckedTransaction`. 
- Usage of corresponding renamed methods.
- Replaced the work with public fields by getters and setters.
- Moved common logic(for `Create` and `Mint`) into generic methods.
- Replaced `Interpreter<Database>` with `Interpreter<Database, Tx>` and `Interpreter<Database, Script>` accordingly to the logic around(See `ConcreteStorage`).

Added a new `PoolTransaction` type that describes transactions stored and accepted by the `TxPool`. It is prepared to support `Mint` transactions that should be excluded from `TxPool`. Another solved TODO by this change - store only checked transactions in the `TxPool` and pass them into `Producer`(because `PoolTransaction` stores only checked transactions inside).

The `TransactionBuilder` from `fuel-vm` and `fuel-tx` now generate transaction of exact type as `Script` or `Create` instead of `Transaction`. Almost all tests in the `fuel-core` prefer to work with the `Transaction` type, so there is a lot of conversion `Script`/`Create` into `Transaction`.

Close https://github.com/FuelLabs/fuel-core/issues/513


fuel-client/src/client/schema/tx/tests/transparent_receipt.rs
```diff
@@ -173,6 +173,7 @@ impl TryFrom<Receipt> for fuel_vm::prelude::Receipt {
                     .is
                     .ok_or_else(|| MissingField("is".to_string()))?
                     .into(),
+                contract_id: None,
```
xgreenx, line 176: It is implemented properly in https://github.com/FuelLabs/fuel-core/pull/702

fuel-block-producer/src/adapters/transaction_selector.rs
```diff
@@ -99,25 +100,23 @@ mod tests {
...
-                .finalize_checked_without_signature(
+                .finalize_checked_basic(
                     0,
                     &ConsensusParameters {
                         gas_price_factor: 1,
                         ..ConsensusParameters::default()
                     },
                 )
             })
+            .map(Into::into)
```
leviathanbeak, line 111: nit: these are 3 `map()`s in a row, why not `Arc::new(checked_tx.into())` in the first map instead

fuel-block-producer/src/adapters.rs
```diff
@@ -3,44 +3,28 @@ use crate::{
...
     pub sender: Sender,
-    pub consensus_params: ConsensusParameters,
 }
 
 #[async_trait::async_trait]
 impl TxPool for TxPoolAdapter {
     async fn get_includable_txs(
         &self,
-        block_height: BlockHeight,
+        _block_height: BlockHeight,
```
leviathanbeak, line 23: what about block height, is it needed or ?

xgreenx: It is not used right now, but it should be passed into `self.sender.includable` in the future to include only transactions valid for `block_height` .

fuel-block-producer/src/block_producer.rs
```diff
@@ -144,24 +143,19 @@ impl Trait for Producer {
...
         } + 1u64.into();
         let checked = if self.config.utxo_validation {
-            CheckedTransaction::check(
-                transaction,
-                height.into(),
-                &self.config.consensus_params,
-            )?
+            let (transaction, _) = transaction
+                .into_checked(height.into(), &self.config.consensus_params)?
+                .into();
```
leviathanbeak, line 148: Not sure how others feel, but for me this is a bit unclear, I have no idea what does `into()` return here by just looking at the code

fuel-client/src/client/schema/tx/tests/transparent_tx.rs
```diff
@@ -92,101 +96,98 @@ impl TryFrom<Transaction> for fuel_vm::prelude::Transaction {
...
+                tx.witnesses.into_iter().map(|w| w.0 .0.into()).collect(),
+            );
+            create.into()
+        };
+
+        // This `match` block is added here to enforce compilation error if a new variant
+        // is added into the `fuel_tx::Transaction` enum.
+        //
+        // If you face a compilation error, please update the code above and add a new variant below.
+        match tx {
```
leviathanbeak, line 185: nice

fuel-poa-coordinator/tests/test_trigger.rs
```diff
@@ -114,15 +101,16 @@ impl BlockProducer for MockBlockProducer {
     }
 }
 
+// TODO: The same code is in the `adapters::transaction_selector::select_transactions`
```
leviathanbeak, line 104: what about this todo, should we remove this function in favor of the one in adapters or vice versa ?
not clear from the comment

leviathanbeak (approved):
left few minor comments, looks solid

fuel-tests/tests/example_tx.json
```diff
@@ -1,5 +1,6 @@
 {
     "Script": {
+        "discriminant": "Script",
```
Voxelot, line 3: should this be removed?

fuel-core-interfaces/src/txpool.rs
```diff
@@ -34,13 +49,226 @@ use derive_more::{
...
+// TODO: After removing of `utxo_validation` from TxPool, use either `Fully` or `Partially`
+//  and remove usage of `Either` type.
+#[derive(Debug, Eq, PartialEq)]
+pub enum Either<Tx: IntoChecked>
+where
+    Tx::Metadata: Debug + Eq,
+{
+    Partially(Checked<Tx, Partially>),
+    Fully(Checked<Tx, Fully>),
+}
```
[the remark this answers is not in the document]
xgreenx: Removed

fuel-tests/tests/example_tx.json
xgreenx: Yep, the `fuel-tx` was not updated to the latest version=)

fuel-poa-coordinator/tests/test_trigger.rs
xgreenx: It is a problem that I can't say the right place with the current codebase=D
`TxPool` should do sorting inside, and the block producer should expect sorted transactions from it. This TODO highlights that we need to implement this logic to remove this TODO=)

fuel-block-producer/src/block_producer.rs
xgreenx: It is cast back to `Transaction`.  Actually, we can remove this check here, because it is checked during execution inside. 

image

I guess we do this check here because in the future, we want to remove checking it from the executor. Am I right? Or we can remove it?=D @Voxelot

fuel-poa-coordinator/tests/test_trigger.rs
```diff
@@ -101,7 +101,8 @@ impl BlockProducer for MockBlockProducer {
     }
 }
 
-// TODO: The same code is in the `adapters::transaction_selector::select_transactions`
+// TODO: The same code is in the `adapters::transaction_selector::select_transactions`. We need
```
leviathanbeak, line 104: "...we need `to` move..."

Voxelot:
should we just combine this PR with the mint PR now? Seems like we won't be able to really merge them separately since the types were released together in fuel-tx/fuel-vm.

fuel-block-producer/src/block_producer.rs
Voxelot: Ah yeah this seems redundant.

fuel-core-interfaces/src/txpool.rs
```diff
@@ -34,13 +49,226 @@ use derive_more::{
...
+
+/// Transaction used by the transaction pool.
+#[derive(Debug, Eq, PartialEq)]
+pub enum PoolTransaction {
```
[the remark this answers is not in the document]
Voxelot: resolves: https://github.com/FuelLabs/fuel-core/issues/513

fuel-poa-coordinator/tests/test_trigger.rs
```diff
@@ -130,15 +117,17 @@ impl BlockProducer for MockBlockProducer {
     }
 }
 
+// TODO: The same code is in the `adapters::transaction_selector::select_transactions`. We need
+//  to move transaction selection logic into `TxPool` to avoid duplication of the code in tests.
```
Voxelot, line 121: It may make more sense to abstract these fns as a separate module, since these tests use a mocked tx pool and shouldn't take a direct dep on the txpool crate.

Voxelot (approved):
for such a drastic internal structure change to the txs, the breaking impact here actually seems quite minimal. Nice work!

freesig (approved):
Nice, will be good to get this merged asap

Issue #680, opened 2022-10-06
Add (optional) metrics to gossipsub


leviathanbeak:
On initialization Gossipsub supports receiving [Registry struct](https://docs.rs/prometheus/latest/prometheus/struct.Registry.html) for collecting metrics.
This should be optionally supported, added to the p2p config and gossipsub builder function.

ControlCplusControlV:
Added in #717

Issue #483, opened 2022-07-11
Develop `fuel-metrics` crate


ControlCplusControlV:
Develop a `fuel-metrics` crate to handle tracking and metrics for each associated fuel service. This involves breaking out the existing fuel-core database metrics into this new crate. This also means

**Fuel-p2p**
- [ ] - Track maybe amount of p2p messages
- [ ] - Unique peers connected to
- [ ] - Track Transactions from peers

**Fuel-txpool**
- [x] - Average and Median Gas Price of a Tx  
- [ ] - Number of Pending Transactions
- [ ] - Average Lifespan of a transaction
- [x] - Average Gas of a Transaction
- [x] - Median Transaction size

Voxelot:
Some additional metrics that would be useful

Block Stats (executor):
- [ ] histogram: block execution duration
- [ ] histogram: gas per block
- [ ] histogram: fee/coinbase per block
- [ ] histogram: txs per block
- [ ] histogram: I/O (db) time per block

ControlCplusControlV:
Related #680

Voxelot:
@ControlCplusControlV since this is marked as done, we should split the remaining metrics that aren't being collected yet into follow-up tasks.

ControlCplusControlV:
#807 #808

PR #717, opened 2022-10-22
Introduce txpool and p2p metrics

ControlCplusControlV:


ControlCplusControlV:
I changed the test to just scan for numbers rather than using manual checks for each metrics, as now at 40 items I felt it would be too exhaustive to manually check each one.

I would prefer if someone like @leviathanbeak could look over the `gossip_sub` metrics config as I wasn't sure what to put with it. 

This PR reworks how metrics are done, while still using a global static variable to point to them, they are instead configured through the proffered library of `prometheus_client`, where the registries and their accompany counters are held. This requires the registry be held behind a Read/Write, although this should mostly be safe as the same registry will likely not have multiple counters attempted to registered onto it at once.

This close #680 and fixes part of #483, the original metrics do need to reworked into this new system as well in a subsequent PR. Also future work should hopefully break metrics back into a feature rather than just as a config option inside of each crate. Some of the design things are kind of weird so please suggest any changes/fixes and I will make them, but wanted to get an initial version to start tracking these metrics so that collection can begin while they are being improved

leviathanbeak:
@ControlCplusControlV `gossipsub/builder.rs` file has been "removed" in favor of `config.rs`

ControlCplusControlV:
Oh ok, will refactor using that then

ControlCplusControlV:
> @ControlCplusControlV `gossipsub/builder.rs` file has been "removed" in favor of `config.rs`

Should use config.rs now and be integrated properly there

fuel-metrics/src/txpool_metrics.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+lazy_static! {
+    pub static ref TXPOOL_METRICS: RwLock<TxPoolMetrics> = {
+        let registry = TxPoolMetrics::default();
+        let metrics = init(registry);
+
+        RwLock::new(metrics)
+    };
+}
```
freesig, line 59: Global locks are almost always a bad idea. Is there a way we can make this not a global?

freesig (changes requested):

fuel-metrics/src/txpool_metrics.rs
freesig: I'd suggest just passing around the pool as:
```rust
struct TxPoolMetrics(Arc<RwLock<TxPoolMetricsInner>>);
```
Also avoid leaking the lock out to anything that uses the `TxPoolMetrics` and instead use methods that can keep the lock internal.

freesig: Also given that the majority will be writing it's probably faster to use a `Mutex` then a `RwLock`

ControlCplusControlV: Yeah it had initially started as just a way to prevent threads from both registering at the same time so I used a `RwLock`, would a `Mutex` actually end up being better if the lock is just moved down to the Registry (since registration occurs once, but a reading on `/metrics` occurs often?) It is unfortunately necessary to have metrics in this weird global pattern, since plumbing them through in each instance pollutes a lot of function parameters, and becomes even more difficult with feature flags

freesig: If it's just about initializing once then maybe something like a `Once` could work and then you don't need the lock? I honestly think the plumbing is worth it. Globals cause so many issues. The sway repo is having a really hard time with them currently.

ControlCplusControlV: I hadn't thought of `Once`, pinging @Voxelot since we also weighed the cost of global variables. I think it'd be interesting, but there is the cost that the counters need to be in scope of whatever they are measuring + the gql endpoint which can sometimes be a bit awkward. Counter's shouldn't need a lock though so using `Once` on Registries to attach counters could end up working out

Voxelot: Once sounds interesting, I just don't have much exp with it yet.

As far as the globals... It's really egregious to pass these metrics around to where they are needed. For example in fuel-p2p, it requires plumbing through 3+ different p2p tasks / service objects before it gets to where it's needed.

I agree that globals are generally bad, but another way to look at it is to compare updating metrics to emitting log events. It needs to be done everywhere and in many places, and if a logging library didn't use thread-locals and required every function that wanted to log something to pass in a logger context I doubt anyone would use that library.

freesig: I can see your point but in that case we should be leveraging something like `tracing` and just write a subscriber to our backend. They have done all the work to make a global safe and fast for this sort of thing. If this doesn't actually require a lock then it might be ok but even things like incorrectly setup atomics can greatly slow down an application if used everywhere.

Voxelot: `tracing` isn't able to do metrics. There may be some crates like opentelemetry which attempt to do something similar, but it's not well maintained or up-to-date with the best prometheus libs.

freesig: tracing can have any subscriber you want so it's very easy to create a metrics backend

freesig: If we really need this to be a global then lets get rid of the locks

freesig: Also keep in mind this will be global across all tests or multiple fuel-core instance running under the same process

ControlCplusControlV: Locks are there for protection of the registry during registration and with gossipsub since mutable access is required during registration

Voxelot: Writing the metrics as a tracing subscriber sounds interesting 🤔 One issue is that some of our deps like libp2p require direct access to a prometheus registry, so that may be more difficult to fit into the tracing model.

I'd prefer to have at least rudimentary telemetry in the near term even if that involves globals, while we come up with a more ideal solution.

ControlCplusControlV: Talked with @freesig and I am going ot move the Registries into a `Once` cell, then expose that through global functions in `fuel-metrics`. `Once_cell` should be able to support the lazy initialization too of attaching counters to the registries, and since Counters are safe (Underlying type is in an `Arc`) we don't need a lock on those

fuel-metrics/src/service.rs
```diff
@@ -0,0 +1,30 @@
...
+use std::vec;
+
+pub fn encode_metrics_response() -> impl IntoResponse {
+    let mut encoded = vec![];
+
+    if let Option::Some(value) = P2P_METRICS.gossip_sub_registry.get() {
+        encode(&mut encoded, value).unwrap();
+    }
+
+    encode(&mut encoded, &P2P_METRICS.peer_metrics).unwrap();
```
freesig, line 22: Can these unwraps fail?

fuel-p2p/src/gossipsub/config.rs
```diff
@@ -51,15 +53,42 @@ pub(crate) fn default_gossipsub_config() -> GossipsubConfig {
...
+            .with_peer_score(PeerScoreParams::default(), PeerScoreThresholds::default())
+            .expect("gossipsub initialized with peer score");
+
+        gossipsub
+    } else {
+        let mut gossipsub = Gossipsub::new(
+            MessageAuthenticity::Signed(p2p_config.local_keypair.clone()),
+            p2p_config.gossipsub_config.clone(),
+        )
+        .expect("gossipsub initialized");
```
freesig, line 86: Is panicking here what we want?

freesig (dismissed):
Nice one, no locks! Only a few questions around the `unwrap` / `expect`s

fuel-txpool/src/txpool.rs
```diff
@@ -205,6 +207,12 @@ impl TxPool {
             Transaction::Create(create) => create.price(),
             Transaction::Mint(_) => unreachable!(),
         };
+        if self.config.metrics {
+            // Gas Price metrics are recorded here to avoid double matching for
+            // every single transaction, but also means metrics aren't collected on gas
+            // price if there is no minimum gas price
+            TXPOOL_METRICS.gas_price_histogram.observe(price as f64);
```
xgreenx, line 214: Do we want to record data for potentially invalid transactions? Maybe better do it only for committed transactions(during committing the block)? Or have two metrics for invalid and valid transactions=)

fuel-metrics/src/txpool_metrics.rs
```diff
@@ -0,0 +1,59 @@
+use lazy_static::lazy_static;
+pub use prometheus_client::metrics::histogram::Histogram;
+use prometheus_client::registry::Registry;
+use std::{
+    boxed::Box,
+    default::Default,
+};
+
+pub struct TxPoolMetrics {
```
xgreenx, line 9: Could you add a description for each field? Maybe copy/paste from the `init` function or add a comment like "For more details, check the `init` function." (the same for all metrics)

fuel-metrics/src/p2p_metrics.rs
```diff
@@ -0,0 +1,48 @@
...
+    metrics::counter::Counter,
+    registry::Registry,
+};
+use std::{
+    boxed::Box,
+    default::Default,
+};
+
+pub struct P2PMetrics {
+    pub gossip_sub_registry: OnceBox<Registry>,
```
xgreenx, line 13: I'm not sure it is the right pattern to have a separate `Registry` inside each `{}Metrics` structure. It looks redundant because the `{}Metrics` already contains all fields.

Maybe it is better to have one `Registry` that is used by each service. This service can be used to encode all metrics at once.

In this case, we don't need many static metrics for each service, we have one. For each service(crate), we can define its metric inside and add the `init` method and `getters` via traits.

Like:

```rust
pub trait TXPoolMetrics {
    fn init(&mut self); // Maybe return an error, like duplicate in registry?

    fn tx_pool(&self) -> TxPool;
}
```

We don't need to pass metrics into each service constructor in this case. Because the register is already static, you need to call `init` only if the `metrics` feature is enabled. Also, you don't need a workaround with `gossip_sub_registry`.

fuel-metrics/src/p2p_metrics.rs
```diff
@@ -0,0 +1,48 @@
...
+
+        Self {
+            gossip_sub_registry: OnceBox::new(),
+            peer_metrics,
+            unique_peers,
+        }
+    }
+}
+
+pub fn init(mut metrics: P2PMetrics) -> P2PMetrics {
```
xgreenx, line 32: Seems it can be easy part of `P2PMetrics` as a consturctor=)

fuel-metrics/src/service.rs
```diff
@@ -0,0 +1,30 @@
...
+    encode(&mut encoded, &TXPOOL_METRICS.registry).unwrap();
```
xgreenx, line 23: This line and the line below are the same

xgreenx (changes requested):
Forgot to mark=)

fuel-metrics/src/service.rs
ControlCplusControlV: Had missed this, will fix

fuel-txpool/src/txpool.rs
ControlCplusControlV: There reason I did it here was to avoid checking price twice on each transaction, I can move it down to just before it is actually inserted into the txpool. I do think it might be worth exploring metrics for invalid tx's though for insights like if the pool is being spammed with a bunch of low gas txs so I might tackle that in a subsequent PR

fuel-metrics/src/p2p_metrics.rs
xgreenx: It seems the private `new` method feets better than the `Default` because we expect only one instance of the metrics.

ControlCplusControlV: Good idea, will change initialization from `default` to `new`

fuel-txpool/src/txpool.rs
ControlCplusControlV: moved this

fuel-p2p/src/gossipsub/config.rs
ControlCplusControlV: [This `.expect` is mirrored on the non-metrics version](https://github.com/FuelLabs/fuel-core/blob/master/fuel-p2p/src/gossipsub/config.rs#L58), so I'm carrying out over into initialization with metrics. However, since it happens in the existing non-metrics initialization I wouldn't expect the added one to introduce any additional problems.

fuel-metrics/src/service.rs
ControlCplusControlV: They may actually fail if there is some corruption at the `Registry` level, so I've adjusted this to now return an error code if a Registry fails to be encoded

fuel-metrics/src/p2p_metrics.rs
xgreenx: @FuelLabs/client What do you think about the idea with one global `Registry`?

freesig: This won't work because you can't have `&mut self` in a global without using locks.

xgreenx: We need `&mut self` only during initialization once so `RefCell` works for us=)

freesig: Do you mean `OnceCell`? I don't think a `RefCell` can be used in a global because they aren't `Sync`.

xgreenx: Oh, wait, we don't need to use any cell, we need call `init` during `lazy_static`=)

ControlCplusControlV: I've removed `init` and added it to `::new` method

xgreenx: I meant for the idea with traits, we can use the `init` method in the `lazy_static` to init metrics for each service. In this case, we don't need to worry about `&mut self` because it is used only during lazy initialization.

freesig: @ControlCplusControlV didn't you need the `&mut self` to pass to libp2p?

ControlCplusControlV: Yes the `&mut self` is needed to initialize the gossipsub with metrics

xgreenx: If you initialize all metrics in one `lazy_static` call, you don't need `&mut self` outside of that lazy initialization

ControlCplusControlV: I've generally done that where possible, however `gossip_sub` metrics require the `&mut self` to be passed across into `fuel-p2p` as documented [here](https://docs.rs/libp2p-gossipsub/latest/libp2p_gossipsub/struct.Gossipsub.html#method.new_with_metrics) with the initialization of the p2p gossipsub service

xgreenx: Oh, I see. We need it here=)
image

Then yeah, `OnceCell` can be our choice=)

ControlCplusControlV: Global metrics are required because it is the lesser of evils when it comes to plumbing metrics. To pass instances of counters and registries would be nightmarish in the amount of functions and service handlers it would touch, especially if this gets abstracted to a compile time flag later.

Since without global metrics we end up with scenarios where `Counters` need to be plumbed all the way to down to the basic `write` and `get` calls of the Database (which would pollute a lot of function signatures)

freesig (approved):
all looks good, nice job

fuel-metrics/src/core_metrics.rs
```diff
@@ -1,89 +1,73 @@
...
+            bytes_read: bytes_read_histogram,
+            bytes_written: bytes_written_histogram,
+        }
+    }
 }
 
-/// CoreMetrics tracks metrics for most of fuel-core, wrapping core metrics similar to
-/// DatabaseMetrics, kept seperate from GQLMetrics which holds more specific endpoint metrics
-#[derive(Clone, Debug)]
-pub struct CoreMetrics {
```
xgreenx, line 53: Hmmm, why did we remove these metrics?

xgreenx: I see that we don't use them, but maybe it is time to start=D Or we can create an issue with a list of all metrics that we want to have and put this metric there(to not forget). I think it may be useful

ControlCplusControlV: Yeah I can break out an issue, we have this a bit in #483 which should have everything, but if you have any suggestions please add there

xgreenx: Oh, I forgot to check the main issue. It's already there=) Nevermind

fuel-core/src/cli/run/p2p.rs
```diff
@@ -113,6 +113,10 @@ pub struct P2pArgs {
     /// Choose how long RequestResponse protocol connections will live if idle
     #[clap(long = "connection_keep_alive", default_value = "20")]
     pub connection_keep_alive: u64,
+
+    /// Enable p2p metrics or not
+    #[clap(long = "metrics", default_value = "false")]
```
Voxelot, line 118: it might be better to just have a global metrics flag, otherwise we'll need to enable metrics on a per service basis:
`--p2p-metrics --txpool-metrics --blockproducer-metrics` etc.

```suggestion
    #[clap(long = "p2p-metrics", default_value = "false")]
```

Voxelot: this name conflicts with the main metrics flag. Repro:

```
cargo run -p fuel-core --features p2p -- run --help

thread 'main' panicked at 'Command run: Argument names must be unique, but 'metrics' is in use by more than one argument or group',
```

Voxelot (approved):

Issue #668, opened 2022-10-03
PoA Block Signing


Voxelot:
Follow-up to #599 

Sign blocks, save consensus metadata, and broadcast to peers

Need to spec SealedFuelBlock and database columns for block consensus metadata.

Voxelot:
breaking sealedfuelblock into separate task #673

PR #718, opened 2022-10-25
PoA Block Signing

Voxelot:
Initial implementation of PoA block signing & persistence of fuel consensus data.

Todo:

- [ ] update devops configs to accept a consensus secret (deferred to followup PR)
- [x] (deferrable) safe method to convert block id's into messages for signatures (may require a BlockId newtype)
- [x] (deferrable) update fuel-crypto to support zeroize on secret types

In the future, we should use the keystore traits provided by fuel-crypto to allow for a pluggable signing backend.

closes: #668
closes: #673

### DIFF
--- a/fuel-core-interfaces/src/model/block.rs
+++ b/fuel-core-interfaces/src/model/block.rs
@@ -25,9 +25,31 @@ use std::fmt;
 
 #[cfg(any(test, feature = "test-helpers"))]
 use chrono::TimeZone;
+use derive_more::{
+    Display,
+    From,
+    FromStr,
+    LowerHex,
+    UpperHex,
+};
 
 /// A cryptographically secure hash, identifying a block.
-#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
+#[derive(
+    Clone,
+    Copy,
+    Debug,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    Default,
+    FromStr,
+    From,
+    LowerHex,
+    UpperHex,
+    Display,
+)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(transparent))]
 #[repr(transparent)]
@@ -40,25 +62,6 @@ impl BlockId {
         unsafe { fuel_crypto::Message::from_bytes_unchecked(*self.0) }
         // Without this, the signature would be using a hash of the id making it more
         // difficult to verify.
-    }
-}
-
-impl From<Bytes32> for BlockId {
-    fn from(bytes: Bytes32) -> Self {
-        Self(bytes)
-    }
-}
-
-#[allow(clippy::from_over_into)] // Avoids circular dependency
-impl Into<Bytes32> for BlockId {
-    fn into(self) -> Bytes32 {
-        self.0
-    }
-}
-
-impl fmt::LowerHex for BlockId {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        fmt::LowerHex::fmt(&self.0, f)
     }
 }
 
--- a/fuel-core-interfaces/src/model/block.rs
+++ b/fuel-core-interfaces/src/model/block.rs
@@ -16,13 +16,51 @@ use chrono::{
 };
 use core::ops::Deref;
 use fuel_vm::{
+    fuel_crypto,
     fuel_merkle,
     fuel_types::MessageId,
     prelude::Signature,
 };
+use std::fmt;
 
 #[cfg(any(test, feature = "test-helpers"))]
 use chrono::TimeZone;
+
+/// A cryptographically secure hash, identifying a block.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(transparent))]
+#[repr(transparent)]
+pub struct BlockId(Bytes32);
+
+impl BlockId {
+    /// Converts the hash into a message having the same bytes.
+    pub fn into_message(self) -> fuel_crypto::Message {
+        // This is safe because BlockId is a cryptographically secure hash.
+        unsafe { fuel_crypto::Message::from_bytes_unchecked(*self.0) }
+        // Without this, the signature would be using a hash of the id making it more
+        // difficult to verify.
+    }
+}
+
+impl From<Bytes32> for BlockId {
+    fn from(bytes: Bytes32) -> Self {
+        Self(bytes)
+    }
+}
+
+#[allow(clippy::from_over_into)] // Avoids circular dependency
+impl Into<Bytes32> for BlockId {
+    fn into(self) -> Bytes32 {
+        self.0
+    }
+}
+
+impl fmt::LowerHex for BlockId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::LowerHex::fmt(&self.0, f)
+    }
+}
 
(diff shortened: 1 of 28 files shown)


fuel-core/src/cli/run.rs
```diff
@@ -148,3 +162,23 @@ pub async fn exec(command: Command) -> anyhow::Result<()> {
 
     Ok(())
 }
+
+// Attempt to load the consensus key from cli arg first, otherwise check the env.
+fn load_consensus_key(
+    cli_arg: Option<String>,
+) -> anyhow::Result<Option<Secret<[u8; 32]>>> {
+    let secret_string = if let Some(cli_arg) = cli_arg {
+        warn!("Consensus key configured insecurely using cli args. Consider setting the {} env var instead.", CONSENSUS_KEY_ENV);
```
Dentosal, line 171: Should we simply not support this if it's insecure?

Voxelot: It's useful for dev testing when we do multinode setups

Dentosal:
Zeroize support in fuel-crypto has been merged. A non-breaking release of fuel-crypto should be doable, so it requires no changes here.

xgreenx:
Also, maybe we can clean up "TODO: Do not use." and related code=)

Voxelot:
Will defer devops configuration of the secret key to a follow-up PR pending some investigation.

fuel-core-interfaces/src/model/block.rs
```diff
@@ -16,14 +16,52 @@ use chrono::{
...
+        Self(bytes)
+    }
+}
+
+#[allow(clippy::from_over_into)] // Avoids circular dependency
+impl Into<Bytes32> for BlockId {
+    fn into(self) -> Bytes32 {
+        self.0
+    }
+}
```
freesig, line 57:
```suggestion
impl From<BlockId> for Bytes32 {
    fn from(block: BlockId) -> Self {
        block.0
    }
}
```
You should be able to still impl this as from here.

fuel-core/src/database/sealed_block.rs
```diff
@@ -0,0 +1,100 @@
...
+        &mut self,
+        key: &Bytes32,
+    ) -> Result<Option<FuelBlockConsensus>, KvStoreError> {
+        Database::remove(self, key.as_ref(), Column::FuelBlockConsensus)
+            .map_err(Into::into)
+    }
+}
+
+impl Database {
+    pub fn get_sealed_block(
```
freesig, line 61: Should we check for mismatches here? Like the header having a different consensus type then the `FuelBlockConsensus`?

fuel-core/src/service/config.rs
```diff
@@ -75,3 +85,11 @@ pub enum DbType {
     InMemory,
     RocksDb,
 }
+
+/// A default secret key to use for testing purposes only
+pub fn default_consensus_dev_key() -> SecretKey {
```
freesig, line 90: Should this be under `test-helpers` flag to avoid production use?

fuel-core/src/database.rs
```diff
@@ -297,10 +307,28 @@ impl Default for Database {
...
+            .contains_key(&block_id.into())?
+        {
+            return Err(anyhow!("block {:x} is already sealed", &block_id))
+        }
+
+        self.storage::<SealedBlockConsensus>()
+            .insert(&block_id.into(), &consensus)
+            .map(|_| ())
+            .map_err(Into::into)
+    }
```
freesig, line 331: This is a race condition because they are not in the same database transaction. Does it actually matter if the block is sealed twice?

fuel-core/src/database.rs
```diff
@@ -339,11 +367,7 @@ impl InterpreterStorage for Database {
...
                 let signature = Signature::default();
-                let message = unsafe {
-                    fuel_core_interfaces::common::fuel_crypto::Message::from_bytes_unchecked(
-                    block.header.id().into(),
-                )
-                };
+                let message = block.header.id().into_message();
                 // TODO: throw an error if public key isn't recoverable
                 //  when implementing signing (https://github.com/FuelLabs/fuel-core/issues/668)
                 let public_key = signature.recover(&message).unwrap_or_default();
```
freesig, line 373: This should not be `unwrap_or_default` anymore.

freesig (commented):
Mostly looks good, just a few suggestions

fuel-core/src/service/config.rs
Voxelot: It may be useful for testing nodes from the CLI when users don't really care about having a real signing key.

fuel-core/src/database.rs
Voxelot: Hmm yeah it probably doesn't matter if its inserted twice. Will need to come up with a way to manage db transactions across service calls in the future though (maybe task local storage or similar).

fuel-core-interfaces/src/model/block.rs
```diff
@@ -598,60 +705,19 @@ impl SealedFuelBlock {
...
-    fn default() -> Self {
-        Self {
-            time: Utc.timestamp(0, 0),
-            height: BlockHeight::default(),
-            prev_root: Bytes32::default(),
-            generated: Default::default(),
-        }
-    }
-}
+impl Deref for SealedFuelBlockHeader {
```
bvrooman, line 717: I would think that providing deref behaviour would mean using the type as a pointer, so you can use pointer dereferencing syntax, e.g. 
```rust
...
let value = *v;
```
Is `SealedFuelBlockHeader` being used as a pointer type? 
From an API standpoint, this adds an implicit deref operation that returns the header; are there any design concerns there?

fuel-core/src/database.rs
```diff
@@ -443,9 +467,15 @@ mod relayer {
 
         async fn get_sealed_block(
             &self,
-            _height: BlockHeight,
+            height: BlockHeight,
         ) -> Option<Arc<SealedFuelBlock>> {
-            Some(Arc::new(SealedFuelBlock::fix_me_default_block()))
+            let block_id = self
+                .get_block_id(height)
+                .unwrap_or_else(|_| panic!("nonexistent block height {}", height))?;
```
bvrooman, line 474: Can you simply use `expect` here?
```suggestion
let block_id = self
                .get_block_id(height)
                .expect("nonexistent block height {}", height))?;
```

fuel-core/src/database/sealed_block.rs
Voxelot: we could, although there's not really much the application can do if it gets into that state 🤔

fuel-core-interfaces/src/model/block.rs
Voxelot: replaced with derive_more

fuel-core/src/database.rs
Voxelot: fixed to use the correct signature and public key

fuel-core-interfaces/src/model/block.rs
Voxelot: We probably abuse deref in many places to convert types when we probably should be using conversion traits like AsRef instead or simply use getters. Probably due to wanting concise `*` syntax.

I'll look into an alternative approach.

Voxelot: removed deref

freesig: Yeh technically we should probably be using `as_ref()` but it's more of a pain.

freesig (approved):
Looks good

fuel-core-interfaces/src/model/block.rs
bvrooman: Interesting read! Thanks for the context. It appears to be a convenient pattern but I would agree with the document you posted that it is a surprising mechanism. But it's good to know how and why this comes about.

fuel-core/src/database/sealed_block.rs
Voxelot: I think we should defer this checking until we have more than one kind of consensus variant, as rust won't let me even write the error case right now.

PR #719, opened 2022-10-25
Create a BlockId type for safe conversion to Message

Dentosal:
It's hard to find the right boundary for the newtype. It cannot be created like types in the fuel-types crate, since the construction shouldn't be public. This is the minimal amount of changes to introduce the type, and it's converted to Bytes32 for interoperating with database and GraphQL functions. It might make sense to switch the database to use the BlockId type as well.

Requested in https://github.com/FuelLabs/fuel-core/pull/718

### DIFF
--- a/fuel-core-interfaces/Cargo.toml
+++ b/fuel-core-interfaces/Cargo.toml
@@ -15,6 +15,7 @@ anyhow = "1.0"
 async-trait = "0.1"
 chrono = { version = "0.4" }
 derive_more = { version = "0.99" }
+fuel-crypto = { version = "0.6", default-features = false }
 fuel-vm = { version = "0.18", default-features = false, features = ["random"] }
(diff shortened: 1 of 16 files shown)


fuel-core-interfaces/Cargo.toml
```diff
@@ -15,6 +15,7 @@ anyhow = "1.0"
 async-trait = "0.1"
 chrono = { version = "0.4" }
 derive_more = { version = "0.99" }
+fuel-crypto = { version = "0.6", default-features = false }
```
Voxelot, line 18: This is actually re-exported via fuel-vm -> fuel-tx -> fuel-crypto

Dentosal: Fixed, rust-analyzer didn't find it.

Issue #640, opened 2022-09-23
VM time interface impl uses seconds instead of milliseconds


Voxelot:
When fetching the time of a block, the current implementation gets the block time in seconds since Jan 1, 1970. It should be in milliseconds.

This is due to using `.timestamp()` instead of `.timestamp_millis()` here:
```
    fn timestamp(&self, height: u32) -> Result<Word, Self::DataError> {
        let id = self.block_hash(height)?;
        let block = self.storage::<FuelBlocks>().get(&id)?.unwrap_or_default();
        block
            .headers
            .time
            .timestamp()
            .try_into()
            .map_err(|e| Self::DataError::DatabaseError(Box::new(e)))
    }
```
https://github.com/FuelLabs/fuel-core/blob/v0.10.1/fuel-core/src/database.rs#L273

PR #720, opened 2022-10-25
Use milliseconds for DB interpreter storage timestamp

bvrooman:
Related issues:
- Closes https://github.com/FuelLabs/fuel-core/issues/640

### DIFF
--- a/fuel-core/src/executor.rs
+++ b/fuel-core/src/executor.rs
@@ -2137,13 +2137,13 @@ mod tests {
 
         // setup block
         let block_height = rng.gen_range(5u32..1000u32);
-        let time = rng.gen_range(1u32..u32::MAX);
+        let time = Utc.timestamp(rng.gen_range(1u32..u32::MAX) as i64, 0);
 
         let block = PartialFuelBlock {
             header: PartialFuelBlockHeader {
                 consensus: FuelConsensusHeader {
                     height: block_height.into(),
-                    time: Utc.timestamp(time as i64, 0),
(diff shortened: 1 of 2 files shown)


fuel-core/src/executor.rs
```diff
@@ -2189,6 +2189,6 @@ mod tests {
             .unwrap()
             .unwrap();
 
-        assert_eq!(time as u64, receipts[0].val().unwrap());
+        assert_eq!(time.timestamp_millis() as Word, receipts[0].val().unwrap());
```
bvrooman, line 2192: This simultaneously tests that:
- The timestamp is equal to the receipt value
- The receipt value is returned in milliseconds

It may be worth separating these into two asserts, i.e.

```
let value = receipts[0].val().unwrap();
assert_eq!(time, Utc.timestamp_millis(value as i64));
assert_eq!(time.timestamp_millis() as Word, value);
```

Voxelot (approved):

ControlCplusControlV (approved):

xgreenx:
Maybe better to create our inner type that works with milliseconds to ensure we've updated all places?

Voxelot:
> Maybe better to create our inner type that works with milliseconds to ensure we've updated all places?

opened a tech debt issue for this

bvrooman:
Oops, sorry @xgreenx, I didn't see your comment until I merged. Thanks @Voxelot for the extra ticket. I will look into the idea now.

PR #725, opened 2022-10-26
Add support for custom timestamps

### LINKED ISSUE #639: Support custom timestamps in produce_blocks API
SDK users would like the ability to set custom timestamps for block production when writing smart contract integration tests involving the `TIME` opcode.

The current signature for producing blocks is: 
```
impl BlockMutation {
    async fn produce_blocks(
        &self,
        ctx: &Context<'_>,
        blocks_to_produce: U64,
    ) -> async_graphql::Result<U64> {
```

To support custom timestamps, this could be changed to something like:
```
#[derive(InputObject)]
struct TimeParameters {
    /// The time to set on the first block
    start_time: U64,
    /// The time interval between subsequent blocks
    block_time_interval: U64,
}

impl BlockMutation {
    async fn produce_blocks(
        &self,
        ctx: &Context<'_>,
        blocks_to_produce: U64,
        time: Option<TimeParameters>,
    ) -> async_graphql::Result<U64> {
```

cc @digorithm @mohammadfawaz

MujkicA:
Closes #639

This PR adds `TimeParameters` to the produce_block mutation, as described in the issue. The provided time is checked to confirm that it is set after the latest block time, and that the addition of `block_time_interval` does not lead to an overflow.
Three tests are added to verify the normal behaviour + the two error cases.

### DIFF
--- a/fuel-core/src/schema/block.rs
+++ b/fuel-core/src/schema/block.rs
@@ -28,10 +28,12 @@ use async_graphql::{
         EmptyFields,
     },
     Context,
+    InputObject,
     Object,
 };
 use chrono::{
     DateTime,
+    NaiveDateTime,
     Utc,
 };
 use fuel_core_interfaces::{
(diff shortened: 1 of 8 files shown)


fuel-core/src/schema/block.rs
```diff
@@ -392,6 +405,67 @@ impl BlockMutation {
     }
 }
 
+async fn get_time_closure(
```
Voxelot, line 408: looks like this method and the following ones don't actually need to be async?

Voxelot (approved):

Issue #650, opened 2022-09-28
Support `Mint` for coinbase 


Voxelot:
related to: https://github.com/FuelLabs/fuel-specs/issues/286, https://github.com/FuelLabs/fuel-tx/issues/193, https://github.com/FuelLabs/fuel-vm/issues/225

The block producer needs to include a single Mint transaction in each block to capture the fees of the block, with the owner set to a configured wallet address (does not need to match the PoA consensus key).

When executing a block in production mode, the executor will set the correct output amount of the mint transaction to match the coinbase award amount of the block. 

When executing a block in validation mode, the executor will verify that:
 - No more than 1 mint transaction exists in the block
 - The mint only includes a single output
 - The single mint output is of `AssetId::BASE`
 - The amount of the mint output is equivalent to the expected coinbase amount from the gas fees

The coinbase mint transaction does not incur any gas fees itself.

PR #726, opened 2022-10-26
Implementation of the coinbase logic

xgreenx:
Implementation of the coinbase logic as described in the issue.

Close https://github.com/FuelLabs/fuel-core/issues/650


fuel-block-producer/src/block_producer.rs
```diff
@@ -146,13 +146,17 @@ impl Trait for Producer {
...
-            .next()
-            .ok_or_else(|| anyhow!("Expected at least one set of receipts"))
+            .await?
+            .into_iter()
+            .flatten()
+            .collect();
+        if res.is_empty() {
+            return Err(anyhow!("Expected at least one set of receipts"))
+        }
+        Ok(res)
```
xgreenx, line 159: Return all receipts instead of only from the first transaction.

fuel-core/src/executor.rs
```diff
@@ -327,16 +346,90 @@ impl Executor {
...
+            coinbase_tx.outputs(),
+        )?;
+        execution_data.tx_status.insert(
+            0,
+            (
+                coinbase_id,
+                TransactionStatus::Success {
+                    block_id: Default::default(),
+                    time: *block.header.time(),
+                    result: None,
```
xgreenx, line 383: The execution result is not available for `Mint` and `Create` transactions, so I make it `Option`.

fuel-core/src/cli/run.rs
```diff
@@ -145,6 +159,18 @@ impl Command {
...
+        let coinbase_recipient = if let Some(coinbase_recipient) = coinbase_recipient {
+            Address::from_str(coinbase_recipient.as_str()).map_err(|err| anyhow!(err))?
+        } else {
+            let consensus_key = consensus_key
+                .as_ref()
+                .cloned()
+                .unwrap_or_else(|| Secret::new(SecretKeyWrapper::default()));
+
+            let sk = consensus_key.expose_secret().deref();
+            Address::from(*sk.public_key().hash())
```
xgreenx, line 171: Is that the right way to retrieve `Address` from `SecretKey`?

Voxelot: it is, but I should've probably used a helper fn from fuel-tx instead.

xgreenx: What helper do you mean?

Voxelot (dismissed):

fuel-core/src/cli/run.rs
Voxelot: Input::owner(&publick_key)

freesig (dismissed):
Looks great. Only feedback would be I would love some code comments. It makes reviewing and refactoring a lot simpler.

fuel-core/src/executor.rs
```diff
@@ -327,14 +354,105 @@ impl Executor {
...
+        if let Some(Output::Coin {
+            asset_id, amount, ..
+        }) = checked_mint.transaction().outputs().first()
+        {
+            if asset_id != &AssetId::BASE {
+                return Err(Error::CoinbaseOutputIsInvalid)
+            }
+
+            if let Some(expected_amount) = expected_amount {
+                if expected_amount != *amount {
```
Voxelot, line 446: Maybe we shouldn't force the block producer to claim the full coinbase amount? They could have the option to burn the gas fees instead.

```suggestion
                if expected_amount < *amount {
```

Voxelot (approved):

fuel-core/src/executor.rs
xgreenx: Hmm, I don't see how the block producer can use it to cheat, so we can allow it=)

With `!=`, we can know how much gas was spent, but with `<`, we can burn=) I'm okay with both variants, so you can commit your suggestion if you wish and merge PR(all tests are green, I checked locally)=)

Voxelot: We can defer this for now

PR #729, opened 2022-10-27
Fixing `CB` command for coinbase

xgreenx:
Bug: `InterpreterStorage::coinbase` expects that the coinbase transaction for the current block is available in the storage.
image

But it was not. Added temporary coinbase into the storage and after replaced with actual.
Added unit test for this case.

### DIFF
--- a/fuel-core/src/executor.rs
+++ b/fuel-core/src/executor.rs
@@ -290,6 +290,18 @@ impl Executor {
                 &block.to_partial_db_block(),
             )?;
 
+        // Temporary insert coinbase into the storage because `InterpreterStorage::coinbase`
+        // gets coinbase transaction from the storage during execution of `Script`.
+        let coinbase_id = coinbase_tx.id();
+        if block_db_transaction
+            .deref_mut()
+            .storage::<Transactions>()
+            .insert(&coinbase_id, &coinbase_tx.clone().into())?
+            .is_some()
+        {
+            return Err(Error::TransactionIdCollision(coinbase_id))
+        }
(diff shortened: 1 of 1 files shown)


leviathanbeak:
code seems fine, feels a bit hacky, is there anything we can do on VM side to avoid this or this _needs_ to be done?

xgreenx:
I think we can create `VMDatabase` that will store all information required for VM. In this case we don't need to put temporary block and coinbase into the storage

xgreenx:
I tried to implement this change, but it requires modification in many places. Also, `dapp.rs` requires creation of a fake block... 

Maybe better to create a separate storage for metadata used during execution and use it to get information about time, block height, and coinbase. And not push block with zero id

Voxelot:
> > I think we can create VMDatabase that will store all information required for VM. In this case we don't need to put temporary block and coinbase into the storage
> 
> I tried to implement this change, but it requires modification in many places. Also, `dapp.rs` requires creation of a fake block...
> 
> Maybe better to create a separate storage for metadata used during execution and use it to get information about time, block height, and coinbase. And not push block with zero id

I agree this would be preferable, just a big pita of boilerplate to write for delegating the InterpreterStorage methods from VmDatabase to the real db. [ambassador](https://crates.io/crates/ambassador) seems interesting, but it requires modifications to the trait and doesn't allow you to delegate any trait you want.

xgreenx:
Closed in favor https://github.com/FuelLabs/fuel-core/pull/731/files

PR #731, opened 2022-10-27
VmDatabase for proper transaction execution

Voxelot:
related to #730
fixes: https://github.com/FuelLabs/fuel-core/pull/729

Instead of creating a temporary block during execution for certain opcodes, this PR adds a new database wrapper that implements the InterpreterStorage trait. This wrapper can be initialized directly with the values the VM needs to retrieve, so that we don't have to simulate useless parts of the unfinalized block or coinbase tx to satisfy the VM storage API.

### DIFF
--- a/fuel-core/src/executor.rs
+++ b/fuel-core/src/executor.rs
@@ -282,21 +282,6 @@ impl Executor {
             }
         };
 
-        // Temporary insert coinbase into the storage because `InterpreterStorage::coinbase`
-        // gets coinbase transaction from the storage during execution of `Script`.
-        //
-        // # Dev-note: in production mode, the `id` of the temporary coinbase is not the same as
-        // the `id ` of the final coinbase.
-        let coinbase_id = coinbase_tx.id();
-        if block_db_transaction
-            .deref_mut()
-            .storage::<Transactions>()
-            .insert(&coinbase_id, &coinbase_tx.clone().into())?
-            .is_some()
-        {
-            return Err(Error::TransactionIdCollision(coinbase_id))
-        }
-
         // Skip `coinbase` from execution.
         let tx_iter = block.transactions.iter_mut().enumerate().skip(1);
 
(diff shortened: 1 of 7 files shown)


fuel-core/src/executor.rs
```diff
@@ -281,14 +282,20 @@ impl Executor {
             }
         };
 
-        // Insert the current headers (including time, block height into the db tx)
-        block_db_transaction
+        // Temporary insert coinbase into the storage because `InterpreterStorage::coinbase`
```
xgreenx, line 285: I guess we can remove this code

fuel-core/src/executor.rs
```diff
@@ -357,6 +364,13 @@ impl Executor {
             };
         }
 
+        // Remove temporary added coinbase with `coinbase_id`.
```
xgreenx, line 367: And this

fuel-core/src/executor.rs
```diff
@@ -410,14 +417,10 @@ impl Executor {
                 },
             ),
         );
-        if block_db_transaction
+        block_db_transaction
```
xgreenx, line 420: And let's return this part back=)

Voxelot:
thanks for the coinbase cleanup!

xgreenx (approved):

xgreenx (approved):

fuel-core/src/executor.rs
```diff
@@ -1516,6 +1507,35 @@ mod tests {
             assert_eq!(validated_block.transactions(), produced_txs);
         }
 
+        #[tokio::test]
+        async fn execute_cb_command() {
```
bvrooman, line 1511: It looks like the test is just executing the script containing the CB opcode and expecting that the execution doesn't error. Would it be prudent to test that we get the expected return value? Perhaps a test that asserts an error when it's expected?

Voxelot: Ah yeah I was thinking of adding that as well, it would require a more advanced script though.

PR #734, opened 2022-10-27
Add metrics config option

ControlCplusControlV:
Adds a metrics config option so future `metrics` prs don't require a breaking change

### DIFF
--- a/fuel-core/src/cli/run.rs
+++ b/fuel-core/src/cli/run.rs
@@ -112,7 +112,7 @@ pub struct Command {
     #[clap(flatten)]
     pub p2p_args: p2p::P2pArgs,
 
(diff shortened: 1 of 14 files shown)


fuel-core/src/cli/run.rs
```diff
@@ -110,6 +111,9 @@ pub struct Command {
     #[cfg(feature = "p2p")]
     #[clap(flatten)]
     pub p2p_args: p2p::P2pArgs,
+
+    #[clap(default_value = "false")]
```
Voxelot, line 115:
```suggestion
    #[clap(long = "metrics")]
```

Voxelot (approved):

PR #737, opened 2022-10-28
Bugfix: dry run doesn't work for `Create`

xgreenx:
We don't generate receipts for `Create` transactions, so "Expected at least one set of receipts" is relevant only for `Script`.


Dentosal (approved):
LGTM

Voxelot (approved):

PR #741, opened 2022-10-30
[WIP] feat: Persistent Block BMT Data

bvrooman:


fuel-core/src/database.rs
```diff
@@ -361,22 +370,35 @@ impl ExecutorDatabase for Database {
...
+        };
+
+        // Insert the data into the database
+        tree.push(data)?;
+
+        metadata.leaves_count += 1;
+        metadata.root = tree.root()?.into();
+
+        self.storage::<FuelBlockMerkleMetadata>()
+            .insert(&"FuelBlocks".to_string(), &metadata)?;
```
Voxelot, line 401: "FuelBlocks" should be a const

fuel-core/src/database/block.rs
```diff
@@ -120,39 +125,68 @@ impl StorageMutate<FuelBlockIds> for Database {
     }
 }
 
-impl StorageInspect<FuelBlockMerkleData> for Database {
+impl StorageInspect<fuel_merkle::binary::NodesTable> for Database {
```
Voxelot, line 128: Hmm, won't this prevent us from having any other kinds of persistent bmt in the future other than blocks? We should have a unique table struct def per tree.

bvrooman: Yes. This is, unfortunately, a bit of a problematic design: the `fuel-merkle` trees are generic over a `StorageType` type, but `StorageType` is currently hardcoded to a specific table, i.e.:
```rust
impl<StorageType, StorageError> MerkleTree<StorageType>
where
    StorageType: StorageMutate<NodesTable, Error = StorageError>,
 ...
 ```
Right now, I am cutting corners to get the functionality fleshed out in this WIP branch. But this must be addressed in order for a real integration to be possible. I am looking into how I can make these trees generic on the table as well. I will derive a solution, but it is possible this it be a breaking change to `fuel-merkle`; if so, I will shelve this work during the code freeze in favour of higher priority tasks.

bvrooman:
Closing as stale. I am in the process of a new PR to achieve persistent BMT data.

Issue #738, opened 2022-10-28
The proper handling of invalid transactions during block execution


xgreenx:
<img width="1291" alt="image" src="https://user-images.githubusercontent.com/18346821/198714051-0c0eb5cb-adbb-4747-9b11-a8d56a5c41ab.png">

Any invalid transaction in the `TxPool` ruins block production. We need to handle it for stable production.
It seems the best way is to exclude failed transactions from the block and continue block generation(Without interruption). As the result of block execution, we can return generated block, and skipped transactions. `TxPool` can validate skipped transactions and clean them up or include block production again.

Voxelot:
We also should figure out why invalid txs were able to make it to the block producer in the place and update the txpool validation accordingly.

xgreenx:
> We also should figure out why invalid txs were able to make it to the block producer in the place and update the txpool validation accordingly.

The cause was a misconfiguration https://github.com/FuelLabs/fuel-core/pull/743

PR #746, opened 2022-11-02
Added skipping of invalid transactions during block production

xgreenx:
Added skipping of invalid transactions during block production with unit tests.

Close https://github.com/FuelLabs/fuel-core/issues/738

### DIFF
--- a/fuel-core/src/executor.rs
+++ b/fuel-core/src/executor.rs
@@ -1487,9 +1487,10 @@ mod tests {
             )
             .unwrap()
             .total();
+            let invalid_duplicate_tx = script.clone().into();
 
             let mut block = FuelBlock::default();
-            *block.transactions_mut() = vec![script.into()];
+            *block.transactions_mut() = vec![script.into(), invalid_duplicate_tx];
 
             let ExecutionResult {
                 block,
@@ -1499,7 +1500,7 @@ mod tests {
                 .await
                 .unwrap();
 
-            assert!(skipped_transactions.is_empty());
+            assert_eq!(skipped_transactions.len(), 1);
             assert_eq!(block.transactions().len(), 2);
             assert!(block.transactions()[0].as_mint().is_some());
             assert_eq!(
@@ -1513,6 +1514,7 @@ mod tests {
             }) = block.transactions()[0].as_mint().unwrap().outputs().first()
             {
                 assert_eq!(asset_id, &AssetId::BASE);
+                assert!(expected_fee_amount > 0);
                 assert_eq!(*amount, expected_fee_amount);
                 assert_eq!(to, &recipient);
             } else {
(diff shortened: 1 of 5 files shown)


fuel-core/src/executor.rs
```diff
@@ -267,86 +283,54 @@ impl Executor {
...
-                    // All other `Mint` transaction is not allowed.
-                    let (mint, _): (Mint, _) = mint.into();
-                    return Err(Error::NotSupportedTransaction(Box::new(mint.into())))
+                    ExecutionKind::Production => {
+                        // If, during block production, we get an invalid transaction, remove it
+                        // from the block and continue block creation. An invalid transaction
+                        // means that the caller didn't validate it first, so maybe something
+                        // is wrong with validation rules in the `TxPool`(or in another place that
+                        // should validate it). Or we forgot to clean up some dependent/conflict
+                        // transactions. But it definitely means that something goes wrong, and
```
Voxelot, line 322:
```suggestion
                        // transactions. But it definitely means that something went wrong, and
```

fuel-core/src/executor.rs
```diff
@@ -267,86 +283,54 @@ impl Executor {
...
             }
         };
-
-        // Skip `coinbase` from execution.
-        let tx_iter = block.transactions.iter_mut().enumerate().skip(1);
+        // Skip the coinbase transaction.
+        let mut tx_index = 1;
 
         // Execute each transaction.
-        for (idx, tx) in tx_iter {
```
Voxelot, line 289: why do we need to change this to a while loop with manually managed index incrementing? Is it just for skipping the first tx?

fuel-core/src/executor.rs
```diff
@@ -370,6 +354,76 @@ impl Executor {
...
+                    execution_data,
+                    tx_db_transaction,
+                    execution_kind,
+                )?
+            }
+            CheckedTransaction::Mint(mint) => {
+                // Right now, we only support `Mint` transactions for coinbase,
+                // which are processed separately as a first transaction.
+                //
+                // All other `Mint` transaction is not allowed.
```
Voxelot, line 418:
```suggestion
                // All other `Mint` transactions are not allowed.
```

fuel-core/src/executor.rs
xgreenx: It was done to be able to use `swap_remove` below 
image


With an iterator, it is not possible. Alternative to the `while` loop, I can transform transactions into `into_iterator`, filter them, and after collect them into vector and set to the block.

freesig: You could use [retain_mut](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain_mut)

freesig (changes requested):
It would be good to see this done without indices. Even if you get it correct it will be very easy for someone else to get it wrong in a refactor

fuel-core/src/executor.rs
Voxelot: Hmm I don't think we'd want to swap_remove, as wouldn't that change the sorted order of the transactions and potentially impact the validity of the block?

For example, let's say you have 3 transactions (tx1, tx2, tx3):
- let tx3 depend on the output of tx2
- let tx1 be a faulty tx that needs to be skipped

If tx1 is swap_removed, the vec of transactions will now be [tx3, tx2]. Worse, tx3 wont even be processed at all.

I made a simple playground to demonstrate how this can lead to pretty severely incorrect results on a larger set of txs where half of the txs are invalid: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=50c1cc2edb68d931ebd99aa3e2ce58bc

I'd rather focus of the code be on correctness and clarity, vs optimizing array access and modification.

As a remedy, I'd prefer to lean more heavily on iterator style operations like [FilterMap](https://doc.rust-lang.org/std/iter/struct.FilterMap.html) as opposed to imperatively mucking with the memory.

Voxelot: Rust is also specially tuned to speed up iterators vs loops. It allows the compiler to do more inlining and other optimizations that aren't possible with side-effectful imperative looping styles: https://doc.rust-lang.org/book/ch13-04-performance.html

Voxelot: retain_mut also looks like a good option

Voxelot: Also, I don't think we'd want to increment the index if the tx was skipped. As the index is used for txpointers and other indexes where we want to record the total location of a tx within a block. If there are skipped txs, these indexes will point to incorrect locations in the block.

xgreenx: In while we don't increment it for skipped transacitons=)

Voxelot: ah right - the `continue`

Voxelot: The changes look better, although since the original bug wasn't caught in the original unit tests I think we should add new cases to cover the issue I uncovered above.

fuel-core/src/executor.rs
```diff
@@ -267,87 +284,61 @@ impl Executor {
...
-                    let (mint, _): (Mint, _) = mint.into();
-                    return Err(Error::NotSupportedTransaction(Box::new(mint.into())))
+                if let Err(err) = tx_db_transaction.commit() {
+                    return Some(Err(err.into()))
                 }
-            };
-        }
+                tx_index += 1;
+                Some(Ok(tx))
+            })
```
freesig, line 340:
```suggestion
            })
            .map(|tx| {
               if tx.is_ok() {
                  tx_index += 1
                }
                tx
              })
```
If you do this and remove the above increments then you can avoid incrementing the index in multiple places

freesig (changes requested):

fuel-core/src/executor.rs
xgreenx: I need `tx_index` to execute transaction

freesig: Yeh that's ok as you are taking a copy into the above closure

freesig: And this will mean the skipped transactions are already filtered out

xgreenx: I don't need a copy with the value `0` in the closure above=) I need to update it after each valid transaction=) I can't have at the same time two closures: first to read `tx_index` and second to write into `tx_index`.

fuel-core/src/executor.rs
xgreenx: Done

fuel-core/src/executor.rs
freesig: Yeh I see because of the double borrow issue. You might need to do something like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ad2c53b5399b7479e8ab7d39860f58ce)

xgreenx: I already do that=)

freesig: I mean make a function for what is within the filter and then check if it's `Some(Ok())` before incrementing

xgreenx: I got it, but it is still the same because I have only one place with `tx_index += 1` after `execute_transaction` at the end of the closure.

Added one more inner closure to finish this thread=)

Voxelot (approved):

PR #751, opened 2022-11-03
Added removal of skipped transactions from `TxPool`

xgreenx:
Added removal of skipped transactions from `TxPool` on the `PoACoordinator` level instead of the `BlockProducer` level.
Added unit test for this case to check that skipped transactions are propagated to `TxPool`.
Renamed `produce_block` into `produce_and_execute_block` to clarify current behavior.

### DIFF
--- a/fuel-poa-coordinator/src/service.rs
+++ b/fuel-poa-coordinator/src/service.rs
@@ -14,6 +14,7 @@ use fuel_core_interfaces::{
     block_importer::ImportBlockBroadcast,
     block_producer::BlockProducer,
     common::{
+        fuel_tx::UniqueIdentifier,
         prelude::{
             Signature,
             Word,
@@ -23,6 +24,7 @@ use fuel_core_interfaces::{
             Secret,
         },
     },
+    executor::ExecutionResult,
     model::{
         BlockHeight,
         FuelBlock,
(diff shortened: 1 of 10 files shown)


Voxelot (approved):

fuel-poa-coordinator/src/service.rs
```diff
@@ -174,11 +176,30 @@ where
...
+            block,
+            skipped_transactions,
+        } = self.signal_produce_block().await?;
 
         // sign the block and seal it
         self.seal_block(&block)?;
 
+        let mut tx_ids_to_remove = Vec::with_capacity(skipped_transactions.len());
+        for (tx, err) in skipped_transactions {
+            error!(
```
freesig, line 189: These should probably not be `error` level as this isn't an error with the software. Maybe `info` level?

freesig:
So for anyone else who finds this a bit confusing the context is that the executor will not be deciding when a block is committed in the future, the consensus coordinator will.

fuel-poa-coordinator/src/service.rs
xgreenx: If everything works correctly, we shouldn't have cases with skipped transactions at all. It is why I marked it `error`=)

PR #753, opened 2022-11-03
Remove requirement for GraphQL hex scalars to start with 0x

Voxelot:
We've been getting a lot of graphql errors in our environments due requests not including a leading `0x` prefix on hex string data. This is likely due to a change in fuel-types serde that removed this 0x by default. Even though all the SDK tests work fine with things the way they are, this PR removes the strict requirement on API requests to reduce the likelihood of failed requests that should otherwise pass.

image


bvrooman (commented):
I would suggest adding two small tests: one to check that you can create a scalar from a hex input with the `0x` prefix and another test to check that you can create a string from a hex input without the prefix.

bvrooman (approved):
👍

PR #755, opened 2022-11-04
More integration tests for PoA block triggering behavior

Dentosal:
This PR adds some integration tests for PoA block triggering. Timing-aware testing is currently blocked by Tokio issue 4522, since we cannot actually stop time in the tests. These should still be an improvement compared to the current testing situation. I'm hoping to make a PR to Tokio regarding the issue at some point, or alternatively add a way to toggle trigger on and off for testing.


Voxelot:
Seems like some of the tests are stalling @Dentosal  Maybe there's a channel not getting closed somewhere?

Dentosal:
I'm unable to reproduce stalling failing tests locally, neither on macos or linux. I'm trying to find out what is actually causing this. However, the tests are written in a way that some rare error conditions might lead to the test stalling indefinitely, so I'll rewrite those parts first.

Voxelot:
For me it seems the timers are triggering blocks very quickly, after 2-5s of running the test has already produced thousands of (mostly) empty blocks. So the query to fetch 1024 blocks is already full on the first request that sets `count_start`. It seems like we need a way to pause the timer for block production until after all the txs are submitted to the pool.

xgreenx (approved):
I guess https://github.com/FuelLabs/fuel-core/pull/797 fixed an issue with CI, so we can merge it=)

Dentosal:
The changes seem to actually fix the issues. I'd still like to test the timings separately, but at least this PR can be merged now.

Issue #721, opened 2022-10-25
Create a millisecond newtype wrapper for handling time


Voxelot:
Maybe better to create our inner type that works with milliseconds to ensure we've updated all places?

_Originally posted by @xgreenx in https://github.com/FuelLabs/fuel-core/issues/720#issuecomment-1291219604_

Voxelot:
closing as we should be using tai64 instead of ms

Issue #384, opened 2022-06-03
BlockHeader timestamp field


rakita:
discussion: https://github.com/FuelLabs/fuel-core/pull/345#discussion_r882809858

tldr: Replace `chrono` with `time` or `u64` inside BlockHeader.

PR #758, opened 2022-11-05
Change timestamps from chrono to tai64

Voxelot:
We were incorrectly using unix time in blocks when we should be using tai64. This was causing issues for the solidity proofs when trying to verify the block hash and signature.

This is a breaking change that will require a new breaking release of fuel-core (0.14). However, this upgrade should be much easier than 0.13 and will be backward compatible with forc 0.30.

### DIFF
--- a/fuel-client/src/client/schema/tx.rs
+++ b/fuel-client/src/client/schema/tx.rs
@@ -8,6 +8,7 @@ use crate::client::{
         HexString,
         PageInfo,
         TransactionId,
+        U64,
     },
     types::TransactionResponse,
     PageDirection,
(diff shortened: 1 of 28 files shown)


fuel-client/src/client/schema/tx.rs
```diff
@@ -175,22 +176,22 @@ pub enum TransactionStatus {
 #[derive(cynic::QueryFragment, Debug)]
 #[cynic(schema_path = "./assets/schema.sdl")]
 pub struct SubmittedStatus {
-    pub time: super::DateTime,
+    pub time: U64,
```
ControlCplusControlV, line 179: It might be nice to keep a wrapper time around time so that is explicitly clear that the `u64` time is in TAI format and not unix time as commonly seen with `u64` time

ControlCplusControlV (approved):
Don't see anything beyond suggestion of keeping a DateTime wrapper to document format, but that's not as pressing. Everything else LGTM code style and in local testing

xgreenx:
Seems it closes https://github.com/FuelLabs/fuel-core/issues/721

ControlCplusControlV:
@Voxelot it actually seems you should remove the [async graphql feature flags](https://github.com/FuelLabs/fuel-core/blob/master/fuel-core/Cargo.toml#L23) then this technically closes #115 #115 #384

Voxelot:
added graphql scalar

pixelcircuits (approved):

Issue #752, opened 2022-11-03
use semaphore or restricted size threadpool for dry_run block execution


Voxelot:
By default, the block executor runs directly on the calling thread. While this may be appropriate to ensure lively block production, dry_run calls triggered by API users should be rate limited and processed on blocking threads to prevent the tokio runtime from stalling or becoming unresponsive.

As a start, we could configure tokio with a limited size blocking threadpool with long keep-alive (to reduce overhead). Each dry_run request should use this size restricted threadpool to offload work from the tokio runtime and limit the number of concurrent execution requests. This will allow other APIs to remain responsive and also prevent block production from being stalled by the API.

cc @freesig

freesig:
You probably don't want a long keep alive that is too long because if there is nothing happening you want to close the threads. You want enough so that the threads stay open between constant dry run calls. A semaphore would be useful if we want to limit the dry runs more then the available worker threads (because we want to keep a few for other uses) but if this is not the case we can just let tokio limit the worker threads. `spawn_blocking` will yield if there are no available worker threads.

I also think we should do some basic perf with a few workloads and generate some flamegraphs / look at what is being saturated (IO or cpu or network etc) so we have some idea of what we might be facing. 
@Voxelot did make a good point though that it could really be transaction dependent as a script / contract could make lots of db calls vs another that has no db calls and is all signing / hashing.

PR #761, opened 2022-11-05
Use spawn_blocking for dry-run requests

Voxelot:
fixes: #752

### Moves dry_run to a tokio::spawn_blocking call

Since the executor is mostly blocking code, issuing too many dry-run requests could impact more critical functions of the node (such as block production) if executed directly on the main worker pool.

### Limits the number of concurrent dry_run requests by the CPU count
This is because in the worst case we will be CPU bound. While some txs may be state-access heavy and I/O bound, other txs could be CPU bound by using lots of crypto ops like hashing and ecrecover. To ensure we don't grind the system to a halt with hundreds of CPU intensive blocking threads, semaphore permits are used to rate limit the number of dry_run requests.

### Removes async from the executor
Previously the executor methods were marked as async, even though there were no async calls. This posed a risk when calling the executor within spawn_blocking, as spawn_blocking is a sync context and must use `block_on` to call any futures. Using `block_on` is risky because it can easily lead to deadlocks when using async code. While block_on  likely would've been safe to use in this one case, it could be the source of difficult-to-debug issues if more asynchronous code is added to the executor in the future.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2246,6 +2246,7 @@ dependencies = [
  "itertools",
  "lazy_static",
  "mockall",
+ "num_cpus",
  "rand 0.8.5",
  "rocksdb",
  "serde",
--- a/fuel-block-producer/src/block_producer.rs
+++ b/fuel-block-producer/src/block_producer.rs
@@ -42,7 +42,14 @@ use fuel_core_interfaces::{
         PartialFuelBlockHeader,
     },
 };
-use tokio::sync::Mutex;
+use std::sync::Arc;
+use tokio::{
+    sync::{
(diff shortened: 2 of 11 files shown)


xgreenx (approved):
The change looks clean. I like the reduction of `async` in our code=D Now, `Executor` acts as expected.

We can add a unit test where the semaphore's value is `2`. We run  4 `dry_run`:
- First `dry_run` runs forever
- Second `dry_run` immediately executed
- Third `dry_run`  runs forever
- Fourth `dry_run` is blocked because the limit is reached(test can wait for timeout).

But maybe it is overkill, up to you=)

PR #767, opened 2022-11-09
fix bug where txs wouldn't get squeezed if coin state already existed

Voxelot:
Found a bug where submitting more than two txs that conflict on the same input would stop squeezing the lower priced txs after the first squeeze.

This would ultimately lead to the tx with the lower price getting skipped by the executor, making this only a minor annoyance. However it causes wasted computation during block production.

### DIFF
--- a/fuel-txpool/src/containers/dependency.rs
+++ b/fuel-txpool/src/containers/dependency.rs
@@ -399,7 +399,7 @@ impl Dependency {
                         *utxo_id,
                         CoinState {
                             is_spend_by: Some(tx.id() as TxId),
-                            depth: 0,
+                            depth: max_depth - 1,
                         },
                     );
                     // yey we got our coin
--- a/fuel-txpool/src/containers/dependency.rs
+++ b/fuel-txpool/src/containers/dependency.rs
@@ -392,17 +392,16 @@ impl Dependency {
 
                             Self::check_if_coin_input_can_spend_db_coin(&coin, input)?;
                         }
-
                         max_depth = core::cmp::max(1, max_depth);
-                        db_coins.insert(
-                            *utxo_id,
-                            CoinState {
-                                is_spend_by: Some(tx.id() as TxId),
-                                depth: 0,
-                            },
-                        );
                     }
-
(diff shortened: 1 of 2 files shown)


Dentosal (dismissed):
LGTM. Maybe adding bug description to the test case would be useful, but OTOH this PR can be found from the git history anyway

freesig (dismissed):
Nice!

fuel-txpool/src/containers/dependency.rs
```diff
@@ -392,17 +392,16 @@ impl Dependency {
...
-                            *utxo_id,
-                            CoinState {
-                                is_spend_by: Some(tx.id() as TxId),
-                                depth: 0,
-                            },
-                        );
                     }
-
+                    // mark this coin as spent by the current tx
+                    db_coins.insert(
```
xgreenx, line 398: You will overwrite state, also in the case if `state.is_spend_by` is `None`. 

You only need to do that if you really have a conflict

xgreenx (dismissed):

fuel-txpool/src/containers/dependency.rs
Voxelot: shouldn't we still mark the coin as spent by this tx even if it was previously none?

For example, if I have a tx2 that depends on a tx1 output, that output should now be marked as spent by tx2 even without any conflicts.

fuel-txpool/src/containers/dependency.rs
```diff
@@ -392,17 +392,16 @@ impl Dependency {
...
+                        *utxo_id,
+                        CoinState {
+                            is_spend_by: Some(tx.id() as TxId),
+                            depth: 0,
```
xgreenx, line 402:
```suggestion
                            depth: max_depth - 1,
```

fuel-txpool/src/containers/dependency.rs
xgreenx: I already dismissed my review=)

Voxelot:
looks like the tests are passing now

freesig (approved):

PR #770, opened 2022-11-11
Expose signature of the block, if any in GraphQL

xgreenx:
Previously the producer of the block was a part of the header. But after [change](https://github.com/FuelLabs/fuel-core/pull/692/files#diff-ff6b04ec42abf1fb1b859e16b4bca4cdf83a2e62ee229e60a69b652c03ad0bfc) it is not. Added exposing of all information required to get it


fuel-core/src/schema/block.rs
```diff
@@ -79,6 +84,19 @@ impl Block {
         &self.header
     }
 
+    async fn signature(&self, ctx: &Context<'_>) -> Option<Signature> {
```
Voxelot, line 87: I think we should have something more generic than signature here, as PoS will have multiple signatures etc. Can we reuse the fuel block consensus type instead?

xgreenx: Of course, we can=) I did it as a signature because it is easy to use. The question do we want to bring the complexity now or later=)

xgreenx: After a private discussion, we can add complexity now and add an additional `block_producer` method to the `Block`.

fuel-client/src/client/schema/block.rs
```diff
@@ -72,6 +73,7 @@ pub struct BlockEdge {
 pub struct Block {
     pub id: BlockId,
     pub header: Header,
+    pub consensus: Option<Consensus>,
```
ControlCplusControlV, line 76: Is there a reason an Option was chosen for Consensus instead of just having a default enum type in `Consensus` like `LocalAuthority` or similar?

fuel-client/src/client/schema/block.rs
```diff
@@ -120,6 +122,33 @@ pub struct Header {
...
+
+#[derive(cynic::QueryFragment, Debug)]
+#[cynic(schema_path = "./assets/schema.sdl")]
+pub struct PoAConsensus {
+    pub signature: Signature,
+}
+
+impl Block {
+    /// Returns the block producer public key, if any.
+    pub fn block_producer(&self) -> Option<fuel_vm::fuel_crypto::PublicKey> {
```
ControlCplusControlV, line 139: [nit] is there any reason for using fully qualified imports here instead importing at the top of the file then referencing the type?

xgreenx: `PublicKey` is used only once in this function, importing it for the whole file is not practical. Plus, we don't have any types of `fuel-crypto`, so it will remain imports clean.

fuel-client/src/client/schema/block.rs
xgreenx: I thought that we can have not finished blocks without seal information. But if we need to have this kind of block in the future, I think we will use a separate type and endpoint. Replaced with `Consensus`

Cargo.toml
```diff
@@ -19,6 +19,7 @@ members = [
   "fuel-tests",
   "fuel-txpool",
   "xtask",
+  "version-compatibility",
```
xgreenx, line 22: @Voxelot wrote in the slack:
> Im thinking we should test if client will still work against 0.14.1. Since it's a new field it'd be nice if an older version of the server just ignored the request, but we should double check that this won't suddenly break the SDK etc

So I added tests to verify version compatibility between `0.14.1` and `0.14.2`

version-compatibility/0.14.1-0.14.2/fuel-core:0.14.1-client:0.14.2/test.rs
```diff
@@ -0,0 +1,56 @@
...
+
+        let transaction_response = client
+            .transaction(&format!("{:#x}", tx.id()))
+            .await
+            .unwrap();
+
+        if let TransactionStatus::Success { block_id, .. } =
+            transaction_response.unwrap().status
+        {
+            // TODO: We need to decide is that breaking change or not. For now, the test fails,
```
xgreenx, line 48: If `fuel-client` is `0.14.2` but `fuel-core` is `0.14.1` the test fails with "Unknown field 'consensus'". 
We have 4 variants on how to fix that=)

1. Make a breaking change `0.15.0`.
1. Extract `consensus` from `Block` into a separate method and on the client side try to fetch that part if it is possible in a separate request.
1. Do two requests, one with a new version of the block, if it fails, send it with the old version=)
1. Add a new method to the `Block` in `fuel-client` to get the `consensus` field by passing `FuelClient`=)

Voxelot: Option #1 would be the cleanest, as fuel-client 0.15.0 would still be compatible with fuel-core 0.14.1.

Voxelot (approved):

PR #774, opened 2022-11-15
Updated the `cynic` to `2.2.0`

xgreenx:
It requires https://github.com/obmarg/cynic/pull/579

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,3 +26,6 @@ members = [
 codegen-units = 1
 lto = "fat"
 panic = "abort"
+
+[patch.crates-io]
+cynic = { path = "../cynic/cynic" }
--- a/fuel-client/Cargo.toml
+++ b/fuel-client/Cargo.toml
@@ -17,7 +17,7 @@ path = "src/main.rs"
 [dependencies]
 anyhow = "1.0"
 clap = { version = "3.1", features = ["derive"] }
-cynic = { version = "1.0", features = ["surf"] }
+cynic = { version = "2", features = ["http-surf"] }
 derive_more = { version = "0.99" }
 eventsource-client = "0.10.2"
 fuel-vm = { version = "0.22", features = ["serde"] }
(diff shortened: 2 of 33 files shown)


Cargo.toml
```diff
@@ -26,3 +26,6 @@ members = [
 codegen-units = 1
 lto = "fat"
 panic = "abort"
+
+[patch.crates-io]
+cynic = { git = "https://github.com/xgreenx/cynic", branch = "feagure/expose-steaming-operation" }
```
ControlCplusControlV, line 31: Looks like your [PR has been merged](https://github.com/obmarg/cynic/commit/aff0aed74cfd21c87dcbde63351a862285106e9c) and we can switch back to master awaiting release

fuel-client/src/client/schema/resource.rs
```diff
@@ -46,18 +46,21 @@ pub struct SpendQueryElementInput {
 pub enum Resource {
     Coin(Coin),
     Message(Message),
+    #[cynic(fallback)]
```
ControlCplusControlV, line 49: If cynic is now requiring a fallback type on enums, do we want to use `Unknown`, or is there another way annotate it as an error value that should never realistically be populated? `Unkown` though also works though, but wondering if there is some way to show that some of these Unknowns should not be used

fuel-client/Cargo.toml
```diff
@@ -17,14 +17,15 @@ path = "src/main.rs"
 [dependencies]
 anyhow = "1.0"
 clap = { version = "3.1", features = ["derive"] }
-cynic = { version = "1.0", features = ["surf"] }
+cynic = { version = "2", features = ["http-surf"] }
```
ControlCplusControlV, line 20: @Dentosal 's comments on the initial issue mentioned we might want to wait a week or 2 for initial bugfixes, did you encounter any weirdness when migrating to the new version?

Also any reason for the downgrade in version precision from x.x to x?

fuel-client/src/client/schema/resource.rs
xgreenx: [It](https://cynic-rs.dev/derives/inline-fragments.html#fallbacks) says that it is only one way=(

fuel-client/Cargo.toml
xgreenx: "1.0" and "1" are the same=)
image

leviathanbeak (approved):
Overall a welcome improvement, even snapshots look nicer :)

I don't think we need to wait few weeks if all works on our side, there will always be bugs, just let's keep it up to date as we go

Issue #382, opened 2022-06-02
Refactor all graphql pagination using iter_all into common code


Voxelot:
Currently each paginated api which uses the database iter_all is reimplementing lots of similar pieces of code. This make consistency in pagination behavior inconsistent, and harder to test. 

The interface should accept all the common pagination params (first, last, before, after), the database column to iterator over, and accept a closure to map db results into grapqhl schema types.

ControlCplusControlV:
Should also note that these tests should utilize the new [rstest](https://docs.rs/rstest/latest/rstest/) which is currently(?) used on the `contract.rs` test

xgreenx:
Fixed in the https://github.com/FuelLabs/fuel-core/pull/834

PR #775, opened 2022-11-15
Return error for no argument pagination

### LINKED ISSUE #747: Return error for no option pagination arguments passed into queries
If no pagination arguments are passed into a query it currently returns an empty list.  We should instead return an error.  See [here](https://github.com/FuelLabs/fuel-core/pull/508) for a similar pr.

ControlCplusControlV:
Fixes #747 https://github.com/FuelLabs/fuel-core/issues/382

Technically breaking but there shouldn't(?) be anything downstream which relies on getting nothing. This behavior is more of a bug than a feature/interface at this point


fuel-core/src/schema.rs
```diff
@@ -42,6 +42,19 @@ pub struct Subscription(tx::TxStatusSubscription);
 
 pub type CoreSchema = Schema<Query, Mutation, Subscription>;
 
+fn check_pagination_parameters<T>(
```
leviathanbeak, line 45: Can the name be more descriptive + potentially a comment above the function?
I know that it `checks` parameters but for what?, their validity, their combo-validity, etc?

leviathanbeak:
looks good, will approve after @bvrooman 's suggestion is addressed (and potentially mine if @ControlCplusControlV  agrees with it)

fuel-core/src/schema.rs
```diff
@@ -42,6 +42,19 @@ pub struct Subscription(tx::TxStatusSubscription);
...
+    first: &Option<i32>,
+    after: &Option<T>,
+    last: &Option<i32>,
+    before: &Option<T>,
+) -> bool {
+    // Condition check will short-circuit so all validation logic is combined
+    first.is_some() && before.is_some()
```
xgreenx, line 52: What do you think about this variant? It seems more readable to me.

```rust
match (first, before, last, after) {
    (Some(_), Some(_), _, _)
    | (_, _, Some(_), Some(_))
    | (None, Some(_), _, Some(_))
    | (None, None, None, None) => true,
    (_, _, _, _) => false,
}
```

Also, do we need to allow `(None, Some(before), Some(last), None)` and `(Some(first), None, None, Some(after))`?

fuel-core/src/schema/balance.rs
```diff
@@ -118,6 +120,10 @@ impl BalanceQuery {
         before: Option<String>,
     ) -> async_graphql::Result<Connection<AssetId, Balance, EmptyFields, EmptyFields>>
     {
+        if check_pagination_parameters(&first, &after, &last, &before) {
```
xgreenx, line 123: We have this check "(first.is_some() && before.is_some())" in 7 places, but you only updated it in 4
image

Maybe better to create a new function, `query` or `query_local`, and use it everywhere with this check inside=)

xgreenx (changes requested):

fuel-core/src/schema.rs
ControlCplusControlV: Yes I can change that to be more descriptive

ControlCplusControlV:
> looks good, will approve after @bvrooman 's suggestion is addressed (and potentially mine if @ControlCplusControlV agrees with it)

I do, with some of the changes surrounding moving out fuel-client I had been dragging my feet a bit, since `fuel-client` doesn't support empty pagination queries  and so this requires a new cynic version to test

fuel-core/src/schema/coin.rs
```diff
@@ -111,105 +110,39 @@ impl CoinQuery {
...
-                        .try_collect()?;
-
-                    // filter coins by asset ID
-                    let mut coins = coins;
-                    if let Some(asset_id) = filter.asset_id {
-                        coins.retain(|coin| coin.1.asset_id == asset_id.0);
-                    }
-
-                    // filter coins by status
-                    coins.retain(|coin| coin.1.status == CoinStatusModel::Unspent);
```
xgreenx, line 197: I removed this filter because, on the SDK side, we use it to get all coins(spent and unspent).

Voxelot (approved):

xgreenx (approved):

Voxelot:
Looks good but I'd like to get 0.15.1 released with the sequential slot access before merging this

Issue #85, opened 2021-12-09
Sentry Node


Voxelot:
In order to secure our network in sidechain mode, we'll need to operate sentries that monitor the consensus behaviors of participants for any foul play (e.g. vote withholding or voting for invalid blocks). This will also need to page the security team when problems are discovered.

Dentosal:
Related https://github.com/FuelLabs/fuel-core/issues/647 and https://github.com/FuelLabs/fuel-core/issues/645

PR #779, opened 2022-11-16
add reserved nodes

### LINKED ISSUE #645: Add reserved nodes configuration for P2P
For sentry-mode configuration, we will need our implementation of LibP2P behaviors to support the concept of "Reserved Nodes".

Reserved nodes should be a configurable list of Peer Ids that the node must always try to maintain a connection to. Reserved nodes do not count against the configured peer input/output limits.

There should also be a boolean option to prevent a node from connecting to any discovered peers, except those in the reserved nodes set.

In sentry mode:

- The POA node will configure the Peer Id's of all Sentries as its set of reserved peers. The POA node must also disable connections to any peers not in the reserved set.
- Each sentry node will configure the POA Node as the only reserved node Peer Id. In contrast to the POA node, it must allow connections to anonymously discovered peers in order to allow the PoA node to gossip with the outside world.

See Substrate for reference: https://github.com/paritytech/substrate/blob/0ba251c9388452c879bfcca425ada66f1f9bc802/client/peerset/src/lib.rs#L259

leviathanbeak:
closes #645 

The approach here was to create a wrapper over the Noise protocol authentication.
In the Noise protocol 2 peers, that are about to establish a connection, exchange their Identities,

In our wrapper, once we receive the remote's identity we check it against the `reserved_nodes` HashSet, if it's part of the given set the connection is allowed and established, if not it is rejected.

This check is only performed when `reserved_nodes_only_mode` is enabled, if not, the check is not done, that is, the Noise Protocol is used without the wrapper, but they do return the same value for the `protocol_info`, meaning, out there in the wild, they are the same protocol.


ControlCplusControlV (dismissed):
LGTM, good to see reserved nodes added. It would be nice maybe in a subsequent PR to add my pre-configurated config's so that Sentry/Guarded Nodes could be setup. Testing is minimal but sufficient, and can be expanded as we expand the GQL endpoints to make that easier with p2p

Also does this solve #85 ? Since a Sentry node is a node that is the `reserved_node` of another node?

fuel-p2p/src/config.rs
```diff
@@ -167,10 +195,121 @@ pub(crate) fn build_transport(local_keypair: Keypair) -> Boxed<(PeerId, StreamMu
...
+impl<P, C: Zeroize, R> NoiseWithReservedNodes<P, C, R> {
+    fn new(
+        noise_authenticated: NoiseAuthenticated<P, C, R>,
+        reserved_nodes: &[Multiaddr],
+    ) -> Self {
+        Self {
+            noise_authenticated,
+            reserved_nodes: reserved_nodes
+                .iter()
+                .map(|address| PeerId::try_from_multiaddr(address).unwrap())
```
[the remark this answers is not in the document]
leviathanbeak: added it, but in general this is only used on starting the node, so if it's incorrect it will only panic on start, it's not used after initialization

leviathanbeak:
I have refactored the sentry test a bit. Now it waits for one of the reserved nodes to connect to multiple non-reserved nodes, while always checking that 2 guarded nodes are only connected to their respective reserved nodes.

fuel-core/src/cli/run/p2p.rs
```diff
@@ -40,10 +40,19 @@ pub struct P2pArgs {
...
     pub bootstrap_nodes: Vec<Multiaddr>,
 
+    /// Addresses of the reserved nodes
+    /// They should contain PeerId within their `Multiaddr`
+    #[clap(long = "reserved_nodes")]
+    pub reserved_nodes: Vec<Multiaddr>,
+
+    /// Set to true if you only want to connect to the reserved nodes
+    #[clap(long = "reserved_nodes_only_mode")]
+    pub reserved_nodes_only_mode: bool,
```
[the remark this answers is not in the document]
leviathanbeak: I updated the description a bit, but don't want to be too much specific because there are use cases for it outside of PoA nodes, and the reasoning is mainly straightforward -> you don't want it to be connected to any other node (for whatever reason)

ControlCplusControlV (approved):
LGTM

bvrooman (approved):
I'm not familiar with the Noise protocol, but overall LGTM

fuel-p2p/src/discovery/discovery_config.rs
```diff
@@ -117,16 +137,31 @@ impl DiscoveryConfig {
...
         }
 
         if let Err(e) = kademlia.bootstrap() {
             warn!("Kademlia bootstrap failed: {}", e);
         }
 
-        let next_kad_random_walk = if self.with_random_walk {
-            Some(Delay::new(Duration::new(0, 0)))
+        let next_kad_random_walk = if self.with_random_walk && !reserved_nodes_only_mode {
+            Some(Delay::new(Duration::from_secs(5)))
```
freesig, line 164: Might be worth making this a const and having a comment on why it's 5 seconds?

leviathanbeak: good point, this was unrelated to the issue itself, I just noticed that random walk was set to 0 will update it with a configurable value

Voxelot:
Looks like one of the tests is now timing out:

```
test service::tests::sentry_nodes_working has been running for over 60 seconds
```

Maybe this is related to the recent libp2p upgrades? 

cc @freesig

freesig:
Hmm not sure. The select in that test is pretty complex, are exit conditions always guaranteed to be reached?

leviathanbeak:
@Voxelot @freesig yeah I saw it last night as well, the check is pretty straightforward, 2 guarded nodes need to connect to their specified reserved nodes, and then we check if 1 non-guarded node has discovered and connected to multiple other nodes in the network, I will try to decrease the number of desired connections to the non-guarded node. Locally the test completes pretty fast. I will also add a timeout ~5mins for example and check at that point if the guarded nodes have connected to their reserved nodes, if yes the test can be considered as success, if not then we need to fail it.

leviathanbeak:
ok, the only thing I changed in the exit condition was instead of checking that certain amount of peers _is_ connected at a given time -> to that a certain amount of peers has successfully connected at any point (so some connections might've dropped but they did establish connection at one point). The test is now a bit relaxed and it passes in the CI.

The main point of the test is to ensure that only reserved peers establish connection to their guarded nodes,
once they all establish - the exit condition is not that important, but we do want to make sure that other non-guarded nodes establish connection with different peers in the network (apart from their neighbouring sentry nodes)

fuel-p2p/src/service.rs
```diff
@@ -551,6 +550,128 @@ mod tests {
...
+        let mut p2p_config = P2PConfig::default_with_network("sentry_nodes_working");
+        // enable mdns for faster discovery of nodes
+        p2p_config.enable_mdns = true;
+
+        // let's increase the ideal mesh number
+        // this still should not affect the nodes that are enabling `reserved_nodes_only_mode`
+        let gossipsub_builder = default_gossipsub_builder()
+            .mesh_n(double_reserved_nodes_size)
+            .mesh_n_low(double_reserved_nodes_size - 2)
+            .mesh_n_high(double_reserved_nodes_size + 4)
```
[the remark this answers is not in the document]
Voxelot: We should have a comment explaining why we need the mesh size to be larger than the reserved node size, so that we don't have people making up their own random reasons about the significance of these magic numbers when refactoring or updating these tests in the future.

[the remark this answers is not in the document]
leviathanbeak: I wrote a more detailed comment, plus I made mesh lows and highs same as the ideal mesh size to avoid any confusion.

fuel-p2p/src/service.rs
```diff
@@ -554,6 +553,135 @@ mod tests {
...
+            .mesh_n_high(double_reserved_nodes_size)
+            .build()
+            .expect("valid gossipsub configuration");
+
+        p2p_config.gossipsub_config = gossipsub_config;
+
+        let build_sentry_nodes = || async {
+            let instantiate_nodes: Vec<_> = (0..reserved_nodes_size)
+                .map(|_| async {
+                    let mut node = build_fuel_p2p_service(p2p_config.clone());
```
Voxelot, line 584: We should also set the guarded node as reserved for the sentries, so that the don't accept too many other random peers and drop their connection to the guarded node they're assigned to.

fuel-core/src/cli/run/p2p.rs
```diff
@@ -147,17 +155,26 @@ impl From<P2pArgs> for anyhow::Result<P2PConfig> {
             }
         };
 
+        // Reserved nodes do not count against the configured peer input/output limits.
```
Voxelot, line 158: We should test this for sentries, which have a reserved node but can also connect to other non-reserved peers.

fuel-core/src/cli/run/p2p.rs
```diff
@@ -147,17 +155,26 @@ impl From<P2pArgs> for anyhow::Result<P2PConfig> {
...
+        let reserved_nodes_count = args.reserved_nodes.len();
+
         let gossipsub_config = default_gossipsub_builder()
-            .mesh_n(args.ideal_mesh_size)
-            .mesh_n_low(args.min_mesh_size)
-            .mesh_n_high(args.max_mesh_size)
+            .mesh_n(args.ideal_mesh_size + reserved_nodes_count)
```
Voxelot, line 162: How do we prevent non-reserved peers from using up the extra connection slots added by `reserved_nodes_count`? Will this be addressed in a follow-up PR to ensure we continuously attempting to connect to reserved peer?

leviathanbeak: yes that will be done with the incoming PR that will handle Peer Connections of reserved and non-reserved nodes

fuel-p2p/src/service.rs
leviathanbeak: In order to achieve this, I needed to refactor a bit the way I initialize p2p services within tests, previously I didn't know a nodes' address until it started,  only then I could pass its address to another node's bootstrap/reserved nodes list.

And since here we wanted that both sentry nodes and guarded nodes have each other address specified on start, we had a chicken and egg problem.

I created `NodeData` struct to deal with this issue, and consequently I refactored other tests to use this approach - it was a bit cleaner instead of .awaiting for node's address every time.

Voxelot (approved):

PR #783, opened 2022-11-18
add public address for local node to be reachable for the nodes in the network

### LINKED ISSUE #648: Add PublicAddr option to p2p config
Add a `--public-addr` configuration setting to P2P. 

When broadcasting connection info to peers, the public addr setting should be included in the set of "external addresses" to allow for inbound connections from peer discovery via load balancer.

Specifically, the public addr should be loaded into the swarm behavior via this API: https://docs.rs/libp2p-swarm/latest/libp2p_swarm/struct.Swarm.html#method.add_external_address

leviathanbeak:
closes #648

### DIFF
--- a/fuel-core/src/cli/run/p2p.rs
+++ b/fuel-core/src/cli/run/p2p.rs
@@ -31,6 +31,10 @@ pub struct P2pArgs {
     #[clap(long = "address")]
     pub address: Option<IpAddr>,
 
+    /// Optional address of your local node made reachable for other nodes in the network.
+    #[clap(long = "public-address")]
+    pub public_address: Option<Multiaddr>,
+
     /// p2p network's TCP Port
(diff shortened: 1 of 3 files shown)


bvrooman (approved):

ControlCplusControlV (approved):

xgreenx (approved):
Hmm, I'm not sure that `Multiaddr` supports parsing. But we already use it for `bootstrap_nodes`, so maybe it is okay=)

But did someone try to use `bootstrap_nodes` via cli?=D Maybe we need to add unit tests for cli arguments in the future

leviathanbeak:
> Hmm, I'm not sure that `Multiaddr` supports parsing. 

@xgreenx of course it does, in the end it's just a parsed string

ControlCplusControlV (approved):

Issue #596, opened 2022-09-07
fuel client connection staleness


Voxelot:
After running the faucet for over 70 days, the connection to fuel-core went stale and the SDK was unable to communicate with the node. While a simple restart of the pod fixed the problem, we should figure out why the fuel-gql-client didn't refresh the connection on its own.

Voxelot:
related: #766

PR #786, opened 2022-11-24
Use `http-reqwest` instead of `http-surf`

xgreenx:
The bug was reported by the users. 

Brandon in slack:
> It seems like this is potentially an issue between h1-client (used in surf) and nginx: https://github.com/http-rs/async-h1/issues/108#issue-615126521

 I was able to reproduce the bug the problem is `async-h1` used by `surf`. The usage of `reqwest` fixes the problem. 

Close https://github.com/FuelLabs/fuel-core/issues/596


bvrooman (approved):
Version compatibility tests are nice 👍

ControlCplusControlV (approved):
Works for me, surprised changes were this small

Issue #772, opened 2022-11-14
Move writing transaction status db writing into txpool and add full status to subscription


freesig:
_Originally posted by @freesig in https://github.com/FuelLabs/fuel-core/pull/748#discussion_r1017330305_

xgreenx:
Nope

Issue #627, opened 2022-09-19
Merkle root for Contract state


bvrooman:
- [ ] Contract state is accumulated in a persistent sparse Merkle tree (SMT)
- [ ] The `contractStateRoot` is calculated from the SMT using all contract state

TODO: How is the contract state Merkle root used?

ControlCplusControlV:
@bvrooman main use of a contract state root us will be with fraud proofs, if you have the state root for a contract you can prove the values in specific storage slots before executing an opcode on L1. AFAIK it also helps with some light client things

Issue #626, opened 2022-09-19
Merkle root for Contract balances


bvrooman:
- [ ] Contract balances are accumulated in a persistent sparse Merkle tree (SMT)
- [ ] The `contractBalancesRoot` is calculated from the SMT using all contract balances

TODO: How is the contract balance Merkle root used?

PR #789, opened 2022-11-25
Implemented update of `balance_root` and `state_root` for inputs and outputs

xgreenx:
The change actualizes the `balance_root` and `state_root` for inputs before execution and outputs after execution. It only preparation of the code to do those updates with the current dummy algorithm. The correct algorithm for state calculation should be implemented in https://github.com/FuelLabs/fuel-core/issues/627 and https://github.com/FuelLabs/fuel-core/issues/626.

Also, it is preparation for updating `tx_pointer`. Because it requires an additional field in the database, it should be done in a separate PR.

Minor changes:
- Created `refs` folder in the `fuel-executor` with a wrapper to minimize the amount of code.
- Added `tx_status` to the `ExecutionResult`, preparation for the https://github.com/FuelLabs/fuel-core/issues/772 and more advanced testing.
- Moved `tx_pool::TransactionStatus` to `fuel-core-interfaces`.

CC @bvrooman This change should answer your question in the related issues.


fuel-core/src/executor.rs
```diff
@@ -897,67 +903,175 @@ impl Executor {
...
-                        ..
-                    } => Some((utxo_id, contract_id)),
-                    _ => None,
-                });
-                for (utxo_id, contract_id) in iter {
-                    if *utxo_id != expected_utxo_id(contract_id)? {
-                        return Err(Error::InvalidTransactionOutcome {
-                            transaction_id: tx.id(),
-                        })
+    // TODO: Maybe we need move it to `fuel-vm`? O_o Because other `Outputs` are processed there
```
xgreenx, line 1003: I'm unsure about this part. `fuel-vm` already updates outputs for all utxos. Maybe better to move it there or extract all updates here.

fuel-core/src/executor.rs
```diff
@@ -2449,6 +2574,281 @@ mod tests {
...
+            .build()
+            .transaction()
+            .clone();
+        if let Some(Output::Coin { to, .. }) = foreign_transfer
+            .as_script_mut()
+            .unwrap()
+            .outputs_mut()
+            .last_mut()
+        {
+            *to = Address::try_from(contract_id.as_ref()).unwrap();
```
xgreenx, line 2817: @Voxelot I added this test but not sure what our expectation is. Maybe we want to reject transactions that try sending tokens to the contract via outputs(not the script body).

fuel-block-executor/src/refs/contract.rs
```diff
@@ -0,0 +1,75 @@
...
+        ContractsLatestUtxo,
+        ContractsState,
+    },
+};
+use std::borrow::Cow;
+
+/// The wrapper around `contract_id` to simplify work with `Contract` in the database.
+pub struct ContractRef<'a, Database> {
+    database: Database,
+    contract_id: &'a ContractId,
```
Voxelot, line 25: do we really need to lifetime reference this? the ContractId implements copy and should be fairly cheap to own here 🤔

fuel-core/src/executor.rs
Voxelot: fuel-vm could set the balance and state roots, but it doesn't manage utxo id's or utxo state directly. This is mainly to enable the client to do bulk optimization for utxo lookups and validation across many txs at once in parallel.

fuel-block-executor/src/refs/contract.rs
xgreenx: I added this type to simplify usage without affecting performance. Because the type is public and used inside of another crate, the compiler will not optimize usage of this struct, and it will add a smaaaaaaaaaaaaaall overhead(to copy 32 bytes)=)

If this lifetime causes pain to your eyes, I will remove it=D

fuel-core/src/executor.rs
```diff
@@ -897,67 +903,175 @@ impl Executor {
...
                 maybe_utxo_id
-                    .ok_or(Error::ContractUtxoMissing(*contract_id))?
-                    .into_owned()
+                    .ok_or_else(|| Error::ContractUtxoMissing(*contract.contract_id()))?
             } else {
-                maybe_utxo_id.unwrap_or_default().into_owned()
+                maybe_utxo_id.unwrap_or_default()
             };
             Result::<_, Error>::Ok(expected_utxo_id)
         };
```
bvrooman, line 926: Instead of a lambda, can `expected_utxo_id` be replaced by a getter on the `ContractRef` interface? If that is practical, I think it would make the readability a bit better and consistent:

```rust
let mut contract = ContractRef::new(&mut *db, contract_id);
*utxo_id = contract.utxo_id()?;
*balance_root = contract.balance_root()?;
*state_root = contract.state_root()?;
```

bvrooman (approved):

fuel-block-executor/src/refs/contract.rs
bvrooman: While copying does incur a small overhead, dereferencing is also not free from a cache standpoint. I also think the simpler struct definition we get may be worth the cost of copying the Bytes32 .

fuel-core/src/executor.rs
```diff
@@ -2449,6 +2574,281 @@ mod tests {
         assert_eq!(coin.status, CoinStatus::Spent);
     }
 
+    #[test]
```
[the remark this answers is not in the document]
xgreenx: We will, but in a separate PR. `executor.rs` should live in the `fuel-executor`, and during this refactoring, we will definitely split the logic across several files and the same with tests.

fuel-core/src/executor.rs
xgreenx: I tried to avoid any business logic inside of those wrappers=)

fuel-block-executor/src/refs/contract.rs
xgreenx: I don't know why we are afraid of lifetimes=) But if two people don't like it, I will replace it with four copies.

fuel-core/src/executor.rs
bvrooman: Good point. I am not strongly opinionated either way. As usual with software, there are pros and cons to either approach!

bvrooman (approved):

fuel-block-executor/src/refs/contract.rs
Voxelot: I know it's more efficient to use the lifetimes, but my preference is to try and keep the code as simple as possible until it's proven to be a bottleneck.

bvrooman: Rather than a question of lifetimes, I think (if I'm understanding Brandon's original comment correctly) it's more of a question about when to use references vs. when to use copies. Personally, I think IDs (similar to ints) are something you should be able to copy without much fear of overhead. But I can also understand that in a performance-critical application, avoid copies is a general goal (but there is the potential cost of accumulating cache misses if this is in any sort of loop).

fuel-core/src/executor.rs
```diff
@@ -897,67 +903,168 @@ impl Executor {
...
+                            *balance_root = contract.balance_root()?;
+                            *state_root = contract.state_root()?;
+                            // TODO: Also calculate `tx_pointer` based on utxo's pointer.
+                        }
+                        _ => {}
+                    }
                 }
             }
             // Needed to convince the compiler that tx is taken by ref here
             #[allow(clippy::needless_borrow)]
```
Voxelot, line 946: Does this clippy annotation still apply?

fuel-core/src/executor.rs
```diff
@@ -603,29 +598,38 @@ impl Executor {
...
         // persist any outputs
-        self.persist_outputs(
+        self.persist_output_utxos(
             *header.height(),
             &tx_id,
             tx_db_transaction.deref_mut(),
             vm_result.tx().inputs(),
             vm_result.tx().outputs(),
         )?;
+        self.compute_outputs(
```
Voxelot, line 617: It looks like we're persisting the outputs before we compute their state/balance roots etc. If we had tests to check that the balance / state roots on contract outputs are set and persisted correctly, this would've been caught.

xgreenx: We have a test=) `persist_output_utxos` persist only outputs related to utxo=)

fuel-core/src/executor.rs
xgreenx: No, we can remove

fuel-core/src/executor.rs
```diff
@@ -2449,6 +2567,281 @@ mod tests {
...
+            },
+            transactions: vec![create.into(), modify_balance_and_state_tx.into()],
+        };
+
+        let ExecutionResult {
+            block, tx_status, ..
+        } = executor.execute(ExecutionBlock::Production(block)).unwrap();
+
+        let empty_state = Bytes32::from(*empty_sum());
+        let executed_tx = block.transactions()[2].as_script().unwrap();
```
Voxelot, line 2768: we should also be testing:

1. that the tx in the db matches the tx in the block
2. that utxos in the db match the expected state, since those are separate from the output types on the tx

related: https://github.com/FuelLabs/fuel-core/pull/789/files#r1040374277

fuel-core/src/executor.rs
Voxelot: Hmm yeah, we could do this but it would add extra overhead to every transfer to check if the output address is a contract id.

Voxelot (dismissed):
blocking the previous approval due to bug in utxo persistence

Voxelot:
Ahh that's right, these output calculations aren't persisted to the utxo state set. They are just metadata on the tx itself. Thanks for clarifying.

Voxelot (approved):

fuel-core-interfaces/src/executor.rs
```diff
@@ -50,6 +51,14 @@ pub struct ExecutionResult {
     /// The list of skipped transactions with corresponding errors. Those transactions were
     /// not included in the block and didn't affect the state of the blockchain.
     pub skipped_transactions: Vec<(Transaction, Error)>,
+    /// The status of the transactions execution included into the block.
```
ControlCplusControlV, line 54: [nit] I think this sentence might need to be adjusted along the lines of "The status of executed transactions included in the block"

ControlCplusControlV (approved):
LGTM

PR #796, opened 2022-11-28
Avoid contract clone

freesig:
Avoids cloning and serializing the contract code by inserting raw contract bytes (padded) directly into the database so it can be copied directly into vm memory. Ready for review but marked as draft so this is not merged until dependency is.
Depends on https://github.com/FuelLabs/fuel-vm/pull/272 merging first

### DIFF
--- a/fuel-core/src/state/in_memory/transaction.rs
+++ b/fuel-core/src/state/in_memory/transaction.rs
@@ -1,10 +1,7 @@
 use crate::{
     database::Column,
     state::{
-        in_memory::{
-            column_key,
-            memory_store::MemoryStore,
-        },
+        in_memory::memory_store::MemoryStore,
         BatchOperations,
         DataSource,
         IterDirection,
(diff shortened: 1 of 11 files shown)


fuel-core/src/state/in_memory/transaction.rs
```diff
@@ -30,44 +27,47 @@ use std::{
...
         }
     }
 
     pub fn commit(&self) -> crate::state::Result<()> {
-        self.data_source.batch_write(
-            &mut self
-                .changes
-                .lock()
+        let mut iter = self.changes.iter().flat_map(|lock| {
+            lock.lock()
```
Voxelot, line 54: hmm 🤔 

Shouldn't we lock all the columns first before draining anything? Otherwise there could still be writes happening to the unlocked columns during a commit.

Voxelot:
replaced by #994

Issue #791, opened 2022-11-26
Race conditions between read and write into database


xgreenx:
## Failing of API requests

Due to race conditions, some API requests may fail. As an example, it is any API related to `FuelBlock`. The `FuelBlock` has a `consensus` field that lazy-loaded from the `SealedBlockConsensus` table.

If we will put a pause between the execution of the block(committing `FuelBlock` into DB) and sealing block(committing `Consensus` into DB) here:

<img width="1081" alt="image" src="https://user-images.githubusercontent.com/18346821/204070594-b555dcf3-ffdb-421d-8823-288cbd0da29c.png">

It will reproduce the failed execution of the [CI](https://github.com/FuelLabs/fuel-core/actions/runs/3547606191/attempts/1). 

## Main problem

Race conditions affect not only API queries but also the tx pool(and maybe other places, we need to think about all services).
A not-bad solution may be moving modification of the database into one place(like CM/BI) and locking the database for the bulk update.

With https://github.com/FuelLabs/fuel-core/issues/782 we will remove this problem for API but still may have it for simple RPC queries.

Voxelot:
Oops didn't mean to close this issue, i wrote that this issue was "partially fixed" and github automatically closed this issue because of that.

PR #797, opened 2022-11-29
Avoid triggering empty block production

Voxelot:
Partially fixes: https://github.com/FuelLabs/fuel-core/issues/791

When a lot of txs are inserted into the txpool simultaneously, each tx creates a new "submitted" event which activates the PoA coordinator. The issue is that even if all the txs are included in the first block, each event was still triggering a block to be produced in instant mode (even though the pool was already empty). This PR updates the PoA Service to ignore txpool events if the pool is already empty, which also avoids triggering block production for other events like "SqueezedOut" or "Completed" which shouldn't impact block production signals.

Ideally, the PoA service just uses a notifier of newly submitted txs. That way we don't have to exhaust a full queue of tx status notifications which don't signal any new or meaningful data. It would also avoid extra processing of status updates like when a tx is removed from the txpool due to being included in a block (triggering an echo effect of block production).


xgreenx (approved):

fuel-poa-coordinator/src/service.rs
```diff
@@ -525,10 +541,130 @@ mod test {
...
+
+        // simulate some txpool events to see if any block production is erroneously triggered
+        txpool_tx.send(TxStatus::Submitted).unwrap();
+        txpool_tx.send(TxStatus::Completed).unwrap();
+        txpool_tx
+            .send(TxStatus::SqueezedOut { reason: NoMetadata })
+            .unwrap();
+
+        // wait max_tx_idle_time - causes block production to occur if
+        // consumable gas > 0 is not checked.
```
Voxelot, line 661:
```suggestion
        // pending txs > 0 is not checked.
```

Voxelot:
apart from some minor nits your changes look good to me @xgreenx, thanks for the assist.

freesig (approved):
Nice work. Yeh this makes sense and might also explain some things we saw in the block explorer

PR #799, opened 2022-11-29
Pin libp2p

freesig:
- Pin all libp2p crates
- Update to latest libp2p including deprecated changes
- Update ethers because the old version was in conflict with libp2p


fuel-p2p/src/discovery/mdns.rs
```diff
@@ -1,8 +1,8 @@
 use libp2p::{
     mdns::{
-        MdnsConfig,
-        MdnsEvent,
-        TokioMdns,
+        tokio::Behaviour,
```
leviathanbeak, line 3: I think it would more readable that it's imported as `TokioMdns`, so that we know which behaviour it is (Mdns) and which implementation it uses (Tokio). We should use that approach throught libp2p imports, for behaviours and events, it makes more sense especially when you import multiple Events/Behaviours within a single file.

fuel-p2p/src/discovery/mdns.rs
```diff
@@ -1,8 +1,8 @@
...
+        Config,
+        Event,
```
leviathanbeak, line 5: as `MdnsEvent`

fuel-p2p/src/discovery.rs
```diff
@@ -3,23 +3,18 @@ use futures::FutureExt;
...
+    core::connection::ConnectionId,
     kad::{
         handler::KademliaHandlerProto,
         store::MemoryStore,
         Kademlia,
         KademliaEvent,
         QueryId,
     },
-    mdns::MdnsEvent,
+    mdns::Event,
```
leviathanbeak, line 14: as `MdnsEvent`

fuel-p2p/src/service.rs
```diff
@@ -124,11 +124,15 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
         // configure and build P2P Service
         let transport = build_transport(config.local_keypair.clone());
         let behaviour = FuelBehaviour::new(&config, codec.clone());
-        let mut swarm = SwarmBuilder::new(transport, behaviour, local_peer_id)
-            .executor(Box::new(|fut| {
+        let mut swarm = SwarmBuilder::with_executor(
```
leviathanbeak, line 127: I think we can use here `with_tokio_executor`

leviathanbeak:
PR looks solid, I'd prefer that libp2p's imports are named accordingly, everything else is fine. Thanks for doing the update and removing deprecated methods, I am willing to merge this first and consequently adjust my PRs after it.

Voxelot:
One thing to factor in, our downstream repos like sdk and forc currently aren't updated to Rust 1.64. This may be a non-trivial task as IIRC sway/forc had tons of clippy issues with Rust 1.64.

fuel-p2p/Cargo.toml
```diff
@@ -14,15 +14,49 @@ description = "Fuel client networking"
...
+libp2p-kad = "=0.42.0"
+libp2p-mdns = "=0.42.0"
+libp2p-metrics = "=0.11.0"
+libp2p-mplex = "=0.38.0"
+libp2p-noise = "=0.41.0"
+libp2p-ping = "=0.41.0"
+libp2p-quic = "=0.7.0-alpha"
+libp2p-request-response = "=0.23.0"
+libp2p-swarm = "=0.41.0"
+libp2p-swarm-derive = "=0.30.2"
```
Voxelot, line 54: this version has been yanked: https://crates.io/crates/libp2p-swarm-derive/0.30.2

fuel-p2p/src/discovery/mdns.rs
freesig: Do you mean like `tokio::Behaviour as TokioMdns`?

fuel-p2p/Cargo.toml
Voxelot: as discussed over slack, let's keep the yanked version until libp2p 0.50 is patched to work with swarm-derive 0.31

Voxelot (approved):

PR #800, opened 2022-11-29
Use ascending order for blocks

### LINKED ISSUE #795: Use ascending order for blocks to match other paginated responses
blocks are returned in descending order unlike most other paginated resources. This is at odds with the graphql api specs, which expect responses to be in ascending order. This is causing issues with start and end cursors on paginated responses to be set incorrectly, and likely other issues as well.

deekerno:
Closes #795.

## Changelog
- Use ascending order for `blocks()` query

## Testing Plan
Adjusted `PageDirection::Forward` assertion to match against the first five elements of the iterator. Tests should pass.

### DIFF
--- a/fuel-core/src/schema/block.rs
+++ b/fuel-core/src/schema/block.rs
@@ -347,10 +347,7 @@ where
             true
         })
         .take(records_to_fetch);
-    let mut blocks: Vec<(BlockHeight, fuel_types::Bytes32)> = blocks.try_collect()?;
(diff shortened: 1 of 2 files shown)


Voxelot:
It seems like reversing to maintain ascending order in other resources also breaks the start and end cursors in other apis as well 🤔

Voxelot (approved):
thanks for the quick fix!

PR #805, opened 2022-11-30
remove unused crates

freesig:


fuel-chain-config/Cargo.toml
```diff
@@ -19,7 +19,6 @@ fuel-poa-coordinator = { path = "../fuel-poa-coordinator", version = "0.14.1" }
 hex = { version = "0.4", features = ["serde"] }
 itertools = "0.10"
 rand = "0.8"
-ron = "0.8"
```
Voxelot, line 22: idk how this even got here 😅

Voxelot (approved):

Voxelot:
yay less dep hell

ControlCplusControlV:
Is there not a CI job we could make to check for unused deps?

deekerno:
> Is there not a CI job we could make to check for unused deps? Using something in the line of [unused-deps](https://github.com/est31/cargo-udeps)

On the indexer, we use the following:
```
      - name: Check Unused Deps
        uses: actions-rs/cargo@v1
        with:
          command: udeps
          args: --locked --all-targets --all-features
```
It does require nightly though.

ControlCplusControlV:
We used to use `nightly` for format so this might be worth trying to see, since IIRC you can use nightly toolchain for just the tools

PR #806, opened 2022-11-30
Fuel benches

freesig:
Depends on https://github.com/FuelLabs/fuel-vm/pull/279


benches/src/lib.rs
```diff
@@ -86,12 +91,16 @@ impl VmBench {
 
     pub fn new(instruction: Opcode) -> Self {
         Self {
-            params: ConsensusParameters::default(),
+            params: ConsensusParameters {
+                max_gas_per_tx: u64::MAX - 1000,
```
xgreenx, line 95:
```suggestion
                max_gas_per_tx: LARGE_GAS_LIMIT + 1,
```

benches/src/bin/collect.rs
```diff
@@ -0,0 +1,741 @@
...
+        assert_eq!(state.ids.len(), 3);
+        assert!(state.ids.contains_key("mcp/10000"));
+        assert!(state.ids.contains_key("mcp/100000"));
+        assert!(state.ids.contains_key("mcpi/10000"));
+    }
+
+    #[test]
+    fn to_yaml() {
+        let input =
+            std::fs::read_to_string(concat!(env!("CARGO_MANIFEST_DIR"), "/test.json"))
```
xgreenx, line 662: You forgot to commit `test.json`

benches/README.md
```diff
@@ -0,0 +1,53 @@
...
+call:
+  base: 311
+  dep_per_unit: 14
+```
+There are multiple output options. See `collect --help` for more information.
+
+### Generating the `fuel-vm` `default-gas-costs.rs`
+Do the same as above except run the following `collect` command.
+`cargo run -p fuel-core-benches --bin collect --release -- --input bench.json -f rust --output default-gas-costs.rs`
+This will generate a `default-gas-costs.rs` file in the current directory like:
```
ControlCplusControlV, line 49: Example code should probably be filled out

benches/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Benchmarking a specific function
+First build the benchmark with debug symbols:
+`CARGO_PROFILE_RELEASE_DEBUG=true cargo-criterion -p fuel-core-benches --no-run`
+Then you need to find the actual test binary that was built:
+`ls target/release/deps/vm* -lat | head -1`
+Should give you something like:
+`target/release/deps/vm-a17190f2ca5e7169`
+Then you can run the benchmark with the `flamegraph` command (runs the bench for 10 seconds):
+`flamegraph target/release/deps/vm-a17190f2ca5e7169 --bench bench_name --profile-time 10`
```
Voxelot, line 23:
```suggestion
`flamegraph -- target/release/deps/vm-a17190f2ca5e7169 --bench bench_name --profile-time 10`
```

benches/README.md
```diff
@@ -0,0 +1,53 @@
...
+For example if you want to profile the `swwq` op you would do (it's regex):
+`flamegraph target/release/deps/vm-a17190f2ca5e7169 --bench '^swwq/swwq$' --profile-time 10`
```
Voxelot, line 25:
```suggestion
`flamegraph -- target/release/deps/vm-a17190f2ca5e7169 --bench '^swwq/swwq$' --profile-time 10`
```

benches/README.md
```diff
@@ -0,0 +1,53 @@
...
+Or for a dependent bench you would do:
+`flamegraph target/release/deps/vm-a17190f2ca5e7169 --bench '^srwq/100$' --profile-time 10`
```
Voxelot, line 27:
```suggestion
`flamegraph -- target/release/deps/vm-a17190f2ca5e7169 --bench '^srwq/100$' --profile-time 10`
```

benches/README.md
freesig: What do you mean?

benches/README.md
freesig: Hmm this is not needed for me. Maybe you have a different version of flamegraph?

benches/benches/set/blockchain.rs
```diff
@@ -108,17 +243,258 @@ pub fn run(c: &mut Criterion) {
...
+            Opcode::MULI(0x15, 0x15, 32),
+            Opcode::ADDI(0x15, 0x15, 1),
+            Opcode::ALOC(0x15),
+            Opcode::ADDI(0x14, REG_HP, 1),
+        ];
+        let mut bench = VmBench::contract(rng, Opcode::SRWQ(0x14, 0x11, 0x27, 0x16))
+            .expect("failed to prepare contract")
+            .with_post_call(post_call)
+            .with_prepare_db(move |mut db| {
+                db.merkle_contract_state_insert(&contract, &key, &key)?;
```
Voxelot, line 483:
```suggestion
                let values = vec![key; i as usize];
                db.merkle_contract_state_insert_range(&contract, &key, &values)?;
```

benches/README.md
ControlCplusControlV: README has an empty block where the `default-gas-costs.rs` example should go

Voxelot: > This will generate a `default-gas-costs.rs` file in the current directory like:
> ```rust
> ```

benches/README.md
Voxelot: strange, maybe it'd be better to use cargo commands to choose the target and bench since flamegraph supports criterion?

either way this is a minor thing that doesn't need to be addressed now.

benches/benches/set/blockchain.rs
Voxelot: we weren't actually reading more than a single slot before, and the benchmark dependent param was just testing how fast it could translate unset slots into zeroed out vm memory.

benches/Cargo.toml
```diff
@@ -1,18 +1,24 @@
...
-version = "0.0.0"
 edition = "2021"
 license = "BUSL-1.1"
+name = "fuel-core-benches"
 publish = false
+version = "0.0.0"
 
 [dependencies]
-fuel-core = { path = "../crates/fuel-core" }
+clap = { version = "4.0.17", features = ["derive"] }
```
Dentosal, line 9: We likely shouldn't depend on patch versions here

freesig: It's not pinned to `4.0.17` it just means `4.0.17` and above.

Voxelot: We conventionally try to avoid setting patch version requirements, although it's less of a problem for binary targets. The main issue is that if libraries have different patch requirements it can cause headaches if they aren't compatible with each other when combined downstream.

Voxelot:
Pushed a couple of fixes and updated the default gas cost here: https://github.com/FuelLabs/fuel-vm/pull/317

I noticed some pretty big variance (approx 100 noops) between runs for some opcodes like ECR. Maybe we need to sample a larger dataset, but I'm reasonably satisfied with the current state.

xgreenx (approved):

Issue #802, opened 2022-11-29
fuel-gql-client cli binary broken


Voxelot:
In the move from surf to reqwest, we introduced a hidden requirement that the client must be used from within a tokio runtime. Investigate if there is a way to avoid the bloat of tokio and have a simple async client, if this doesn't seem feasible then we need to add a `#[tokio::main]` to the cli.

Voxelot:
Yeah we might need to use cynic manually instead of as an http client middleware, and use ureq instead. This would probably be for the best as depending on cynic to configure our http client causes a lot of issues with OpenSSL etc in the past.

xgreenx:
I tried to [implement manual usage](https://github.com/FuelLabs/fuel-core/pull/810) of `ureq`, it works with real-life applications, but all unit tests are broken because `ureq` doesn't support async and blocks the main thread(the server that should receive requests is in the same thread).

PR #810, opened 2022-12-03
Usage of `ureq` instead of `reqwest`

xgreenx:
Implementation of the https://github.com/FuelLabs/fuel-core/issues/802#issuecomment-1335547702. It works with real-life applications, but tests are broken=(


xgreenx:
It was only an example. The corresponding issue is solved so close it

PR #811, opened 2022-12-03
Use `tokio::main` for `fuel-gql-client` cli

xgreenx:
Close https://github.com/FuelLabs/fuel-core/issues/802 based on https://github.com/FuelLabs/fuel-core/issues/802#issuecomment-1336182360

### DIFF
--- a/fuel-client/Cargo.toml
+++ b/fuel-client/Cargo.toml
@@ -13,6 +13,7 @@ description = "Tx client and schema specification."
 [[bin]]
 name = "fuel-gql-cli"
 path = "src/main.rs"
+required-features = ["build-binary"]
 
 [dependencies]
 anyhow = "1.0"
@@ -29,7 +30,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0", features = ["raw_value"] }
 tai64 = {version = "4.0", features = ["serde"] }
 thiserror = "1.0"
-tokio = { version = "1.21", features = ["macros"] }
+tokio = { version = "1.21", features = ["macros"], optional = true }
 
 [dev-dependencies]
 insta = "1.8"
@@ -39,5 +40,6 @@ schemafy_lib = { version = "0.5", optional = true }
 serde_json = { version = "1.0", features = ["raw_value"], optional = true }
 
 [features]
+build-binary = ["tokio"]
 test-helpers = []
 dap = ["schemafy_lib", "serde_json"]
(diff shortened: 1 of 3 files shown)


fuel-client/Cargo.toml
```diff
@@ -29,6 +29,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0", features = ["raw_value"] }
 tai64 = {version = "4.0", features = ["serde"] }
 thiserror = "1.0"
+tokio = { version = "1.21", features = ["macros", "rt-multi-thread"] }
```
ControlCplusControlV, line 32:
```suggestion
tokio = { version = "1.21", features = ["macros"] }
```

fuel-client/src/main.rs
```diff
@@ -64,6 +64,7 @@ impl CliArgs {
     }
 }
 
-fn main() {
+#[tokio::main]
```
ControlCplusControlV, line 67:
```suggestion
#[tokio::main(flavor="current_thread")]
```

ControlCplusControlV (changes requested):
Tested with some basic commands on `beta-2`
Such as
```sh
./fuel-gql-cli --endpoint http://node-beta-2.fuel.network transaction receipts 0x555c0e13daa880ec4ce315fe38bd86b4f54bbbd4f439451ca772fd89ae17ba68
```
and it returns back results successfully.

Looking at the discussion, it seems `tokio::main` is unavoidable, maybe this can be revisited as we continue to refactor our async code (I know there are a lot of async functions that don't need to be still left) and gql client. This also brings up that I would suggest adding `cli` tests back into the pipeline, because while they do slow down CI, as people start using our cli more and more it provides critical coverage so we know we aren't breaking things (but that is a separate issue).

This is where my suggestions come in though, if we use a single_thread runtime we get async compatibility but can cut out compiling an entire multi-threaded runtime for the client

fuel-client/src/main.rs
```diff
@@ -64,6 +64,7 @@ impl CliArgs {
...
+async fn main() {
     futures::executor::block_on(CliArgs::parse().exec());
```
ControlCplusControlV, line 69:
```suggestion
    CliArgs::parse().exec().await;
```

Is also needed for my changes to work

ControlCplusControlV (approved):
LGTM, and with these changes adding `tokio` as a dependency should also address @Voxelot 's concerns of weight (although wait to hear back ofc before merging)

Voxelot:
Yeah my first concern was weight, the other being locked into tokio for sdks/user-facing libraries. Ideally async interfaces would work with whatever async runtime they use in their app. Since there's no great way around the second concern for now, I think this is a practical compromise for now.

xgreenx:
To reduce the size, added `tokio` only for binary.

ControlCplusControlV (approved):

Voxelot (approved):

Issue #809, opened 2022-12-02
Ports and Adapters Architecture Migration


Voxelot:
Align fuel core with the new [architecture](https://github.com/FuelLabs/fuel-core/blob/master/docs/architecture.md) design to avoid snowballing our tech-debt before working on any major new features.

- [x] Extract all traits (ports) out of fuel-core-interfaces and into their respective crates (under a ports module).
- [x] Remove fuel-* prefix from project folders, since it doesn't impact the published crate names
- [x] Move plain rust common domain types (such as blocks/sealed blocks etc) into a new `fuel-core-types` crate to remove the need for fuel-core-interfaces.
- [x] Refactor txpool service to not use both channels and locks
- [x] Eliminate all usage of channels in public interfaces between services and use async-traits instead (see [architecture doc](https://github.com/FuelLabs/fuel-core/blob/master/docs/architecture.md) for proposed replacements).

Issue #464, opened 2022-07-06
Move db from `fuel-core` to `fuel-database`


rakita:
Mostly refactoring and cleanup. move all database related items to separate crate `fuel-database`.

Voxelot:
deprioritizing this as we want to avoid:

service module crates taking a direct dependency on database / rocksdb, as that will drastically impact our unit testing performance
a large amount of churn while there are other critical tasks in flight
loose design / boundaries, the db-related logic in fuel-core could likely be split into several different crates more design and justification is necessary here.

Issue #793, opened 2022-11-28
Trying to deploy to beta-2 testnet with https returns an invalid url scheme


kayagokalp:
I am not sure if this is related to the deployment configuration or related to here but trying to deploy a contract to the beta-2 test net with `forc deploy --url https://node-beta-2.fuel.network/graphql --gas-price 1` returns graphql error: invalid URL, scheme is not http. Switching https://node-beta-2.fuel.network/graphql to http://node-beta-2.fuel.network/graphql fixes this.

ControlCplusControlV:
```
Error: Custom { kind: Other, error: "Graphql error: HttpStream(hyper::Error(Connect, \"invalid URL, scheme is not http\"))" }
```

Full error message, seems [hyper_tls](https://docs.rs/hyper-tls/latest/hyper_tls/) may be needed somewhere

ControlCplusControlV:
So I've traced the dep. We use `cynic` in `fuel-client` which in the deployed (but not current build) uses `surf` which then has an upstream dep on `hyper` because [cynic doesn't use rust-ls](https://github.com/obmarg/cynic/blob/master/cynic/Cargo.toml#L19) I think? Continuing to debug but posting update here

Issue #792, opened 2022-11-26
Impossible to get genesis block


xgreenx:
After https://github.com/FuelLabs/fuel-core/pull/770, it is impossible to get the genesis block, because we create a fake empty block, but do not process the `consensus` getter for this height.

We can fix it fast in several ways:
1. Return a signature created from the zeroed secret key for a fake block.
1. Add a new type into the `FuelBlockConsensus` enum - `Genesis`. So the sealed information for the genesis block is `FuelBlockConsensus::Genesis`(I prefer this way).

Or we can fix the problem for this part of the code too:

<img width="1275" alt="image" src="https://user-images.githubusercontent.com/18346821/204071149-53354fb5-2c34-4afe-be62-eabdcbaa2702.png">

We can create and commit a genesis block into the database. It can still use the new `FuelBlockConsensus::Genesis` as sign of sealing.

If we select a more complex way, we need to decide on the following:
- Do we want to include invalid transactions into the genesis block with valid outputs and execute this block(via `fuel-block-executor`) with `utxo_validation = false`? In this case, we will update the state the same way as during block execution.
- Or only commit the new state and set only the message root(as a field of the genesis block header) with zero transactions? It is a custom implementation(we already do that) that requires maintenance.

Voxelot:
I prefer option #2 as well. But if we want to retain backwards compat with fuel-gql-client v0.14, option #1 seems like the safest option in the near term. Once our downstream deps are using cynic 2.0+, we won't have to worry about new enum variants causing massive breaking changes.

xgreenx:
We added `Block.consensus` field for the upcoming release, so we still can change the output=)

Voxelot:
ah true! let's go with option 2 then before 0.15 is released :)

PR #812, opened 2022-12-04
Extracted `Database` into a separate `fuel-database` crate

xgreenx:
## Overview

It is pure extraction of the `Database` and all related stuff(tables definition, `DatabaseTransaction`, other types) to its own crate. There are no additional changes in the business logic except the small one highlighted in the comment. Also, the change follows the ideology described below and moves the implementation of the DB trait defined by each service into the service.

I accidentally started this change because I needed database primitives to have cross-service database transaction(I meant  `DatabaseTransaction` with some additional traits). 

We already tried to do that in the past during this issue https://github.com/FuelLabs/fuel-core/issues/464. But it was not a good time for this kind of change. With upcoming [architecture refactoring](https://github.com/FuelLabs/fuel-core/issues/809), we need this kind of change if we want finally remove `fuel-core-interfaces`.

In the linked issue, there were some concerns:

> service module crates taking a direct dependency on database / rocksdb, as that will drastically impact our unit testing performance

`rocksdb` is an optional dependency of the `fuel-database` disabled by default. All services will use it only for database primitives/types. Only `fuel-core` will enable it, and only `fuel-core` will decide in runtime which database type to use(In memory/RocksDB/New one). 

All services will use an in-memory database for the unit tests by default. `fuel-database`(in most cases) and integration tests are responsible for testing of how RocksDB works.

> a large amount of churn while there are other critical tasks in flight

Seems it is a good time for this change if we want to remove `fuel-core-interfaces`=) Also, it can remove `MockDB` in each service(and supporting of it). 

> loose design / boundaries, the db-related logic in fuel-core could likely be split into several different crates. More design and justification is necessary here.

`fuel-database` only implements the basic functionality of storage and common business logic(like get block, transactions, receipts), and defines primitives and abstraction. The `fuel-database` as upcoming `fuel-core-types` is a low-level library used by each service. The logic of each service belongs to the service. So each service defines its database trait with all required functionality. The service should implement this trait for `Database` directly or by a generic implementation(by using some traits, for example, `KeyValueStore`). 

The services still want to work with the trait defined by them because it allows them to not depend on `Database` and allows downstream crates to provide their "Database" with their own implementation of traits(at the same time `fuel-core` will use a `Database` as the primary type).

## Follow-ups

This change only moves the code we had before without introducing a new one. But we can make more changes later:

- With a low-level database crate, we can remove the usage of `MockDB` and use the `Database` directly.
- Maybe we want to implement all basic functionality for a generic type that implements `KeyValueStore` instead of `Database`.
- Maybe we want to use a generic database instead of `DataSource` inside services. It will allow having more complex traits with generics(with `dyn` keyword the trait should be [safe](https://doc.rust-lang.org/reference/items/traits.html#object-safety)).
- Change the structure of the `fuel-database` crate because it is the same as it was in the `fuel-core`.
- Review traits and types, maybe change naming, add comments, maybe improve our abstraction.
- Improve our errors during work with the database(because we use several there).


fuel-core/src/service/modules.rs
```diff
@@ -292,12 +295,16 @@ impl BlockProducerRelayer for MaybeRelayerAdapter {
...
-            .get_finalized_da_height()
-            .await
-            .unwrap_or_default())
+            Ok(self
+                .database
+                .get_finalized_da_height()
+                .await
+                .unwrap_or_default())
+        }
+        #[cfg(not(feature = "relayer"))]
```
xgreenx, line 305: Here I added a new logic(to return a default value)=( Because the `RelayerDB` trait now lives inside of `fuel-relayer`(instead of `fuel-core-interfaces`) I can't use the `get_finalized_da_height` if `relayer` feature is disabled.

fuel-database/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+pub mod rocks_db;
+pub mod tables;
+pub mod transactional;
+
+/// Database tables column ids.
+#[repr(u32)]
+#[derive(
+    Copy, Clone, Debug, strum_macros::EnumCount, PartialEq, Eq, enum_iterator::Sequence,
+)]
+pub enum Column {
```
xgreenx, line 38: Moving this stuff will affect https://github.com/FuelLabs/fuel-core/pull/796 =( But I will help with it

fuel-database/src/error.rs
```diff
@@ -0,0 +1,132 @@
+use crate::InterpreterError;
+use std::io::ErrorKind;
+use thiserror::Error;
+
+// TODO: Use either `KvStoreError` or `Error`
```
ControlCplusControlV, line 5: As a library I believe the correct way would be hierarchical errors, so external facing functions all return a generic `DatabaseError`, which is itself a wrapper for various errors like `IoError` (which may occur writing data) or `InitializationError` (an error wrapper for all initialization related errors) 

Looks like this has been done to some degree, but would be good to finalize an external error enum before merging

fuel-database/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+strum = "0.24"
+strum_macros = "0.24"
+tempfile = "3.3"
+thiserror = "1.0"
+
+[dev-dependencies]
+rand = "0.8"
+
+[features]
+default = []
```
ControlCplusControlV, line 34: Default features on `fuel-core` include both `metrics` and `rocksdb`, wouldn't this then be a behavior change?

fuel-poa-coordinator/src/db.rs
```diff
@@ -0,0 +1,38 @@
...
+        BlockId,
+        FuelBlockConsensus,
+    },
+};
+use fuel_database::tables::SealedBlockConsensus;
+
+pub trait BlockDb: Send + Sync {
+    fn block_height(&self) -> Result<BlockHeight>;
+
+    // Returns error if already sealed
```
ControlCplusControlV, line 15:
```suggestion
    /// Returns error if already sealed
```
Doc comment

fuel-relayer/src/db.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// Set the last fuel block height that was published to the da layer.
+    async fn set_last_published_fuel_height(&self, block_height: BlockHeight);
+}
+
+#[async_trait::async_trait]
+impl RelayerDb for Database {
+    async fn insert_message(&mut self, message: &CheckedMessage) {
+        let _ = self
+            .storage::<Messages>()
+            .insert(message.id(), message.as_ref());
```
ControlCplusControlV, line 56: Doesn't a result need to be handled here?

ControlCplusControlV (changes requested):
Some initial changes / questions, sorry if I overlooked something, this is a lot to digest

Also I do understand why we need this change, but at the very least shouldn't this wait until the issues mentioned on Slack (#792 , #710 , #793 , #802 ) be fixed before merging?

Also would be nice to add this to #809 since it has to do with upcoming refactors

fuel-database/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+    }
+}
+
+pub trait KeyValueStore {
+    fn get(&self, key: &[u8], column: Column) -> Result<Option<Vec<u8>>>;
+    fn put(&self, key: &[u8], column: Column, value: Vec<u8>) -> Result<Option<Vec<u8>>>;
+    fn delete(&self, key: &[u8], column: Column) -> Result<Option<Vec<u8>>>;
+    fn exists(&self, key: &[u8], column: Column) -> Result<bool>;
+    // TODO: Use `Option<&[u8]>` instead of `Option<Vec<u8>>`. Also decide, do we really need usage
+    //  of `Option`? If `len` is zero it is the same as `None`. Apply the same change for all upper
```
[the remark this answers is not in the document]
xgreenx: I only moved the code=)

BTW, it is handled in the https://github.com/FuelLabs/fuel-core/pull/643

fuel-database/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+}
+
+impl Default for IterDirection {
+    fn default() -> Self {
+        Self::Forward
+    }
+}
+
+pub trait KeyValueStore {
+    fn get(&self, key: &[u8], column: Column) -> Result<Option<Vec<u8>>>;
```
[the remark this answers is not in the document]
xgreenx: Which TODO do you mean?

fuel-database/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+}
+
+pub trait BatchOperations: KeyValueStore {
+    fn batch_write(
+        &self,
+        entries: &mut dyn Iterator<Item = WriteOperation>,
+    ) -> Result<()> {
+        for entry in entries {
+            match entry {
+                // TODO: error handling
```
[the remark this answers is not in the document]
xgreenx: I only moved the code, it was here before=)

fuel-database/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+pub trait KeyValueStore {
+    fn get(&self, key: &[u8], column: Column) -> Result<Option<Vec<u8>>>;
+    fn put(&self, key: &[u8], column: Column, value: Vec<u8>) -> Result<Option<Vec<u8>>>;
+    fn delete(&self, key: &[u8], column: Column) -> Result<Option<Vec<u8>>>;
+    fn exists(&self, key: &[u8], column: Column) -> Result<bool>;
+    // TODO: Use `Option<&[u8]>` instead of `Option<Vec<u8>>`. Also decide, do we really need usage
+    //  of `Option`? If `len` is zero it is the same as `None`. Apply the same change for all upper
+    //  functions.
+    //  https://github.com/FuelLabs/fuel-core/issues/622
+    fn iter_all(
```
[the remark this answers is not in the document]
xgreenx: What do you mean?

fuel-database/src/error.rs
```diff
@@ -0,0 +1,132 @@
...
+
+impl From<KvStoreError> for std::io::Error {
+    fn from(e: KvStoreError) -> Self {
+        std::io::Error::new(ErrorKind::Other, e)
+    }
+}
+
+impl From<Error> for InterpreterError {
+    fn from(e: Error) -> Self {
+        InterpreterError::Io(std::io::Error::new(std::io::ErrorKind::Other, e))
```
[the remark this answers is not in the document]
xgreenx: It was before here, but okay)

fuel-database/src/error.rs
xgreenx: I agree, but this PR is simple movement stuff from one place to another. The issue with errors is mentioned in the `follow-ups` section in the description=)

fuel-database/Cargo.toml
xgreenx: No, because `fuel-core` still enables those features by default=)

fuel-database/src/tables/resource.rs
```diff
@@ -212,6 +214,8 @@ impl<'a> AssetQuery<'a> {
     }
 }
 
+// TODO: Move definition of `ResourceId` and `Resource` to `fuel-core-interfaces`
```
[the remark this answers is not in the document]
xgreenx: It will be done in the follow-up PRs mentioned in the description. I will create them after the merge of this PR and all comments. If we forget about it, it is not critical todo to forget =) It means that no one uses resources.

fuel-p2p/Cargo.toml
```diff
@@ -17,6 +17,7 @@ bincode = "1.3"
 fuel-core-interfaces = { path = "../fuel-core-interfaces", features = [
     "serde",
 ], version = "0.14.1" }
+fuel-database = { path = "../fuel-database", version = "0.14.1", default-features = false }
```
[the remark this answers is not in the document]
xgreenx: It is a guard for the future to not use unexpected default features. Right now, `fuel-database` hasn't default features.

fuel-relayer/src/db.rs
xgreenx: That code was before=)
image


Of course we need to handle it and return a result, but it is not part of this PR=)

xgreenx (commented):
> but at the very least shouldn't this wait until the issues mentioned on Slack 

I tried to fix https://github.com/FuelLabs/fuel-core/issues/791. But seems it is not listed, so yeah, we can wait for a new release first if we don't need it=)

fuel-database/src/lib.rs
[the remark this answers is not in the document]
ControlCplusControlV: There was a TODO I saw but was on an `impl` not trait definition so unrealted, my bad

fuel-database/src/error.rs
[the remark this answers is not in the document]
ControlCplusControlV: Thanks, sorry this just got to be me before and now there is a chance to fix it

fuel-database/src/lib.rs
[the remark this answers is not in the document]
ControlCplusControlV: Since you're moving the code and this isn't a refactor this is fine, but `iter_all` logic is definitely weird right now in the database

fuel-database/src/lib.rs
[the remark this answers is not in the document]
ControlCplusControlV: Ok, as long as its marked down somewhere

Voxelot:
> rocksdb is an optional dependency of the fuel-database disabled by default. All services will use it only for database primitives/types. Only fuel-core will enable it, and only fuel-core will decide in runtime which database type to use(In memory/RocksDB/New one).

My preference would be to avoid taking a dep on rocksdb at all, making the database crate mostly pure / interfaces. This would likely involve the adding the `KeyValueStore` & `BatchOperations` traits as ports to fuel-database.

fuel-relayer/src/db.rs
Voxelot: Didn't you say you were going to fix this already? 😆 

https://github.com/FuelLabs/fuel-core/issues/563#issuecomment-1216085668

fuel-database/src/lib.rs
Voxelot: This is an example of why I've been hesitant towards creating a mega-db crate. This shared set of columns seemingly breaks our attempt of having a clean separation of concerns. While implementing mappables and service db ports from the crate perspective is nice, we still end up with a cross-service dependency through these shared columns. Any service that needs to change columns will cascade into potentially breaking adapter changes in all the other crates (even if they had no real logical changes).

While this may not seem like a big issue now when compared to the benefits of sharing the database when considering the executor migration or cross-service, we now open the possibility for implementers of new crates to just start throwing what they need into the shared database crate instead of using their own local ports without really being aware of any difference.

By allowing fuel-core to be responsible for aggregating the database ports from services and having full control over how they are implemented, it resolves the following situations:
 - Allowing custom impls / fine tuning of database traits in the context of a fuel-core node. E.g. using separate instances of rocksdb for merkle tree updates vs plain contract state for performance.
 - Prevent devs from arbitrarily updating a shared database crate whenever they need something custom for their use-case, causing unintended side-affects for other projects using our service crates in different contexts (e.g. putting our block executor into a parachain or cosmos app).

I get how some set of shared data interfaces would be beneficial for things like the executor migration or cross-service txs. An alternative approach to solving these problems without fuel-database or fuel-core-interfaces could be:

- Create an alternative to `DatabaseTransaction` that is entirely trait based. The new trait should have similar AsRef/Deref functionality as before allowing for casts to the db port of each respective service using it.
  - This could be in a new crate as long as it's only highly generic database traits that don't know anything about the underlying services or what their requirements/mappables/columns are. 
- Then services like the executor would just use its own db port trait which defines the mappings it needs, and also use the new trait for database transactions.

This way fuel-core has ultimate control over how db calls get implemented and services don't share any dependency on each other's database requirements.

freesig: I’ve read through this a few times and it’s not really landing. Can you provide a small example of what a trait based transaction could like like?

freesig:
I'm not really convinced we need this change. Are there any other projects that actually currently want to use `fuel-database`?

fuel-database/src/lib.rs
freesig: Is the idea to have some kind of custom functions for each request on a service? So instead of using columns and keys we have a getter function for every piece of data we need?

xgreenx: ## Purpose of this PR

I want to highlight one more time that this change only moves stuff to make the review as simple as possible. Real refactoring to be a crate of database primitives will be done in follow-ups. There are two variants of how to create this crate:
- Extract the whole logic into one crate and break it down(to downstream crates) with new changes after.
- Define new primitives in this new crate. After, we will start refactoring existing logic to use those primitives.

I selected the first approach because it allows refactoring with the right primitives that will be easily renamed/replaced later. And it will parallelize refactoring(unblock in some cases because anyone can contribute to the `fuel-database` crate) to use ports/adaptors and new database primitives. 

> This could be in a new crate as long as it's only composed of highly generic database traits that don't know anything about the underlying services or what their requirements/mappables/columns are. 

You already mentioned a new crate where we can store database primitives. It is this crate but designed for `fuel-core` purposes. I will describe "`fuel-core` purposes" below.

## `fuel-core` purposes

The approach with ports and adaptors is good and will provide more flexibility for third parties who want to use our crates.

But at the same time, it means we move most of the logic into `fuel-core`. How to work with the storage is also a business logic of the service. Yes, it is not the responsibility of the service to know about the `RocksDB` or `InMemory` database, but it is the responsibility to update several storage tables, like `Coins` and `Messages`. 

We will face this problem while moving the `executor.rs` into the `fuel-executor` crate. [`ContractRef`](https://github.com/FuelLabs/fuel-core/blob/bad93d1e626aae61cfe9a8ede873f026508cc34b/fuel-block-executor/src/refs/contract.rs#L24) is an example of simplification development and aggregation of the business logic of the service. Moving this into `fuel-core` also transits the business logic of the service there.

My idea is to follow the ports/adapters approach but simultaneously provide a default implementation of the database port for `fuel-core`. Because our development is oriented toward `fuel-core` and support of modularity is an additional feature.

For third parties, it means that they have more ways of customizability:
1. They can reuse our `Database`(`RocksDB` or `InMemory`) and the default implementation of the database port then they have the same logic.
1. They can use their custom "database" but our default implementation of the port.
1. They can implement database port with their own logic.

Without it, they can only `1.` and `3.`.

At the same time, for us, we store the database business logic of the service inside of the service.

It is why it is database primitives for `fuel-core` purposes=)

## Other thoughts after the discussion above

Initially, I wanted to call this crate `fuel-storage`, but we already have the same crate. So I called it the `fuel-database`. But after some thought, it is the wrong naming. The proper name is `fuel-core-storage`.

The `fuel-storage` is a library for database primitives used by `fuel-core` and `fuel-vm`. Because `fuel-vm` is independent(doesn't know about `fuel-core`), it defines its database port. But also defines its `Mappable` types and provides a default implementation for this port.

The same with `fuel-core`. We use primitives from `fuel-storage` and `fuel-vm-storage` to define the required `Mappable` types for the default storage implementation around an unknown `Database`. So `fuel-core-storage` seems a good name, and we can use `fuel-database` for the crate with `Database` implementation(instead of storing it inside of the `fuel-core`).

The `fuel-core-storage` provides storage traits, mappable types, and other traits/types like `Transactional`/`Transaction<Database>`. It will be used by all services.

The `fuel-database` crate will provide an implementation of the `RocksDB`, `InMemory`, `Transactional<InMemory, RocksDB>` databases. It is used only by `fuel-core` in the `[dependencies]` section, and used by all service in the `[dev-dependencies]`

## Answers to questions

> Create an alternative to DatabaseTransaction that is entirely trait based. The new trait should have similar AsRef/Deref functionality as before allowing for casts to the db port of each respective service using it.

> I’ve read through this a few times and it’s not really landing. Can you provide a small example of what a trait based transaction could like like?

We discussed that yesterday on the call=) It is what I plan to implement [here](https://github.com/FuelLabs/fuel-core/compare/master...bugfix/race-condition-seal-block?expand=1) for the `poa-coordinator` service.
 
The idea is: 
Because the `DatabaseTransaction` structure is only known on the `fuel-core` level, we can't use it in the `fuel-block-producer`. But we can use a generic `DbTransaction: Transactional + AsRef<BlockProducerDB> + AsMut<BlockProducerDB>`, that implements transactional functionality(like `commit`) and also return a `Database` instance that implements the database port, defined by the by `fuel-block-producer` crate. 

With generics, it doesn't look lovely like:
```rust
pub trait Transactional {
    fn commit(self) -> anyhow::Result<()>;
}

// It is a database port defined by the `poa-coordiantor`.
pub trait BlockDb: Send + Sync {
    fn block_height(&self) -> Result<BlockHeight>;

    fn seal_block(
        &mut self,
        block_id: BlockId,
        consensus: FuelBlockConsensus,
    ) -> Result<()>;
}

#[async_trait::async_trait]
pub trait BlockProducer<DbTransaction, Database>: Send + Sync
where
    DbTransaction: Transactional + AsRef<Database> + AsMut<Database>,
    Database: BlockDb,
{
    // TODO: Right now production and execution of the block is one step, but in the future,
    //  `produce_block` should only produce a block without affecting the blockchain state.
    async fn produce_and_execute_block(
        &self,
        height: BlockHeight,
        max_gas: Word,
    ) -> anyhow::Result<UncommittedResult<DbTransaction>>;

    async fn dry_run(
        &self,
        transaction: Transaction,
        height: Option<BlockHeight>,
        utxo_validation: Option<bool>,
    ) -> anyhow::Result<Vec<Receipt>>;
}
```

You can see that `BlockProducer<DbTransaction, Database>` will cause the propagation of these two generics inside of `PoACoordinator`.

With `DatabaseTransaction<Database>` inside of `fuel-core-storage`, it would be `BlockProducer<Database>` it looks more friendly. Also, we plan to replace `dyn BlockDB` with `Database` in the future, so it is expected to be generic around `Database`=)

For now, instead of ugly `BlockProducer<DbTransaction, Database>`, we will use dynamic resolution:
```rust
pub trait Transactional {
    fn commit(self) -> anyhow::Result<()>;
}

// It is a database port defined by the `poa-coordiantor`.
pub trait BlockDb: Send + Sync {
    fn block_height(&self) -> Result<BlockHeight>;

    fn seal_block(
        &mut self,
        block_id: BlockId,
        consensus: FuelBlockConsensus,
    ) -> Result<()>;
}

pub trait DatabaseTransaction: Transactional + AsRef<&dyn BlockDb> + AsMut<&mut dyn BlockDb> {}

#[async_trait::async_trait]
pub trait BlockProducer: Send + Sync {
    // TODO: Right now production and execution of the block is one step, but in the future,
    //  `produce_block` should only produce a block without affecting the blockchain state.
    async fn produce_and_execute_block(
        &self,
        height: BlockHeight,
        max_gas: Word,
    ) -> anyhow::Result<UncommittedResult<Arc<dyn DatabaseTransaction>>>;

    async fn dry_run(
        &self,
        transaction: Transaction,
        height: Option<BlockHeight>,
        utxo_validation: Option<bool>,
    ) -> anyhow::Result<Vec<Receipt>>;
}
```

But this approach creates more `Arc` and `dyn` in our code, but doesn't require any ugly/not expected changes=)

Voxelot: > The fuel-core-storage provides storage traits, mappable types, and other traits/types like Transactional/Transaction<Database>. It will be used by all services.
> 
> The fuel-database crate will provide an implementation of the RocksDB, InMemory, Transactional<InMemory, RocksDB> databases. It is used only by fuel-core in the [dependencies] section, and used by all service in the [dev-dependencies]
> 

Splitting into a fuel-core-storage vs fuel-database, where all the heavy deps and implementation logic of the database is separate, mostly resolves my concerns. As fuel-core-storage would just be high-level traits.

> You can see that BlockProducer<DbTransaction, Database> will cause the propagation of these two generics inside of PoACoordinator.

We definitely want to avoid propagating service-specific db traits between services, as it pollutes the domain with unrelated constraints.

xgreenx:
The PR is too stale, so create an issue to implement it later https://github.com/FuelLabs/fuel-core/issues/970

PR #814, opened 2022-12-05
Move block producer requirements into corresponding crates

xgreenx:
I reused the pattern of the adapter for the `block-producer` and moved block-producer requirements into corresponding crates.
It is part of the change to return an `ExecutionResult` with a database transaction inside. The rest of the change will be in a separate PR.


fuel-block-producer/src/block_producer.rs
```diff
@@ -55,6 +47,27 @@ use tracing::debug;
 #[cfg(test)]
 mod tests;
 
+#[derive(Error, Debug)]
+pub enum Error {
```
ControlCplusControlV, line 51: For `error` interfaces do we want a name of Error across every service with type being the hint to where it's from, or should we do `<ServiceName>Error`. I think it's worth considering the latter because when debugging no matter how many weird string conversions or formatting takes place atleast the source is still preserved

fuel-core/src/service/modules.rs
```diff
@@ -301,3 +305,31 @@ impl BlockProducerRelayer for MaybeRelayerAdapter {
...
+            .produce_and_execute_block(height, max_gas)
+            .await
+    }
+
+    async fn dry_run(
+        &self,
+        transaction: Transaction,
+        height: Option<BlockHeight>,
+        utxo_validation: Option<bool>,
+    ) -> anyhow::Result<Vec<Receipt>> {
```
ControlCplusControlV, line 330: [nit] any reason to just not import Result?

fuel-poa-coordinator/src/service.rs
```diff
@@ -56,6 +59,24 @@ use tracing::{
     warn,
 };
 
+#[async_trait::async_trait]
+pub trait BlockProducer: Send + Sync {
+    // TODO: Right now production and execution of the block is one step, but in the future,
```
ControlCplusControlV, line 64: Since we are refactoring right now, would it be worth considering building out this code path right now to make it easier for future milestones. `produce_and_execute_block` is relatively simple and we could make a `produce` function which returns a block, then just call an `execute` function right away, which preserves functionality but avoids a future breaking interface change on the `BlockProducer`

ControlCplusControlV (commented):
Wait for discussion for others, I had a couple questions, but nothing overall. Main concern is this + db causing large churn before other more pressing issues (like the ones mentioned on slack) can be fixed

fuel-poa-coordinator/src/service.rs
xgreenx: It is not refactoring mentioned in the https://github.com/FuelLabs/fuel-core/issues/809. It is simply moving from one place to another to unblock the development for me and don't create a mess in one PR=)

The comment describes the problem bad=) We can't produce the block without the execution of transactions(otherwise, it is a useless block). The comment was meant to be, "The current code produces, executes, and commits the block. But it should only produce and execute".

It is that I'm trying to fix=) This "refactoring" for [this](https://github.com/FuelLabs/fuel-core/compare/feature/block-producer-adapter...bugfix/race-condition-seal-block?expand=1) change. But without https://github.com/FuelLabs/fuel-core/pull/812 it looks ugly.

fuel-block-producer/src/block_producer.rs
ControlCplusControlV: Yes that's correct, I see what you mean. My response had been to the effect that this can be tricky with serialization and stuff (as that Service module) isn't always preserved, but since we already do this for naming and other things this is fine

fuel-block-producer/src/block_producer.rs
```diff
@@ -55,6 +47,27 @@ use tracing::debug;
...
+    MissingBlock(BlockHeight),
+    #[error("Best finalized da_height {best} is behind previous block da_height {previous_block}")]
+    InvalidDaFinalizationState {
+        best: DaBlockHeight,
+        previous_block: DaBlockHeight,
+    },
+}
+
+#[async_trait::async_trait]
+pub trait Relayer: Sync + Send {
```
Voxelot, line 66: could we move this to a `ports` module?

fuel-core/src/service/modules.rs
```diff
@@ -301,3 +305,31 @@ impl BlockProducerRelayer for MaybeRelayerAdapter {
             .unwrap_or_default())
     }
 }
+
+struct PoACoordinatorAdapter {
```
Voxelot, line 309: Could we move this to an adapters module, e.g. similar to this? https://github.com/FuelLabs/fuel-core/blob/master/docs/architecture.md#adapters-fuel_coreserviceadapterspoa_consensus

Voxelot (approved):
overall looks good to me, just a couple nits about using consistent ports and adapters nomenclature for organizing these traits.

Voxelot (approved):

PR #817, opened 2022-12-05
Added support of genesis block and genesis consensus field

xgreenx:
Close https://github.com/FuelLabs/fuel-core/issues/792.

The change inserts the genesis block into the database and the corresponding `FuelBlockConsensus::Genesis` seal information to define the initial state of the blockchain and the consensus for future blocks.

We also need to add rules for the genesis into the specification(it is in progress).


fuel-block-producer/src/block_producer.rs
```diff
@@ -194,17 +194,11 @@ impl Producer {
     }
 
     fn previous_block_info(&self, height: BlockHeight) -> Result<PreviousBlockInfo> {
+        // TODO: It is not guaranteed that the genesis height is `0` height. Update the code to
```
Voxelot, line 197: ahh good catch, since we allow setting any arbitrary starting block height in the chainspec.

fuel-block-producer/src/block_producer/tests.rs
```diff
@@ -52,8 +52,18 @@ async fn cant_produce_at_genesis_height() {
 
 #[tokio::test]
 async fn can_produce_initial_block() {
-    // simple happy path for producing the first block
-    let ctx = TestContext::default();
+    // setup dummy genesis block
+    let genesis_height = 0u32.into();
```
Voxelot, line 56: should we make setting a default genesis block part of the TestContext init? Seems like this is going to turn into lots of extra boilerplate unrelated to the actual test logic if we have to include this every time.

fuel-chain-config/src/config/chain.rs
```diff
@@ -30,6 +30,8 @@ use super::{
 pub const LOCAL_TESTNET: &str = "local_testnet";
 pub const TESTNET_INITIAL_BALANCE: u64 = 10_000_000;
 
+// TODO: Remove not consensus/network fields from `ChainConfig` or create a new config only
```
Voxelot, line 33: Which fields are you referring to here?

fuel-client/assets/schema.sdl
```diff
@@ -263,6 +263,26 @@ type FailureStatus {
...
+	"""
+	coinsHash: Bytes32!
+	"""
+	The Binary Merkle Tree hash of state, balances, contracts code hash of each contract.
+	"""
+	contractsHash: Bytes32!
+	"""
+	The Binary Merkle Tree hash of all genesis messages.
+	"""
+	messagesHash: Bytes32!
```
Voxelot, line 283: Hash -> Root

fuel-chain-config/src/config/chain.rs
xgreenx: It was about `block_production` and `chain_name`. But maybe `chain_name` may live here, but I can't say that is a "required" field for us=D

The `block_production` describes how to behave and is more related to `fuel-block-producer` because of `fuel_poa_coordinator::Trigger::Never`(In the network the only part of nodes uses `Never`). Maybe we need another type to describe production rules for `ChainConfig`.

Voxelot: In the case of networks with multiple producers, they should have the same block production trigger config imo.

xgreenx: Yes, but if you are sentry node, your `ChainConfig` will be different=)

fuel-core/src/service/genesis.rs
```diff
@@ -30,126 +36,295 @@ use fuel_core_interfaces::{
...
 
         Ok(())
     }
 
+    pub fn add_genesis_block(config: &Config, database: &mut Database) -> Result<()> {
+        // Initialize the chain id and height.
+        database.init(&config.chain_conf)?;
+
+        let chain_config_hash = config.chain_conf.clone().root()?.into();
+        let coins_hash =
```
Voxelot, line 177: Hmm I'm a little torn on this approach of reading a partially setup DB to generate the genesis block.

- On one hand, deriving the genesis block fields from the database ensures that it matches what is actually contained in the state. 
- On the other hand, it seems like deriving the genesis block field could be stateless op done without a db by just inspecting the `chain_conf` values.

The main risk I see with the current approach is that in order to generate a genesis block for comparison or other tools, it requires a fully clean db initialized in a very specific way. This could hamstring our options in the future such as patching existing environments with a missing genesis block, as the DB would already be contaminated with lots of state after genesis. 

It seems like we either run the risk that the genesis block contains more than just genesis state in some unforeseen future usages of the API. Or we run the risk that the computed genesis block doesn't actually align to the initial DB state, however this issue seems less likely to me. IMO it seems like being able to derive the genesis block from the `chain_conf` without a pre-populated DB would be a safer alternative.

fuel-chain-config/src/config/chain.rs
Voxelot: Ok but lets say in a multi-producer network, we want to use this field to configure the initial set of validator keys. It shouldn't change whether you are in sentry mode or not. To me that seems like production behavior could be based on whether you actually have a local private key configured that matches the configured validator set?

fuel-core-interfaces/src/model/block.rs
```diff
@@ -598,11 +598,11 @@ pub struct Genesis {
...
     pub chain_config_hash: Bytes32,
-    /// The Binary Merkle Tree hash of all genesis coins.
+    /// The Binary Merkle Tree root of all genesis coins.
     pub coins_hash: Bytes32,
-    /// The Binary Merkle Tree hash of state, balances, contracts code hash of each contract.
+    /// The Binary Merkle Tree root of state, balances, contracts code hash of each contract.
     pub contracts_hash: Bytes32,
-    /// The Binary Merkle Tree hash of all genesis messages.
+    /// The Binary Merkle Tree root of all genesis messages.
     pub messages_hash: Bytes32,
```
Voxelot, line 606: hash -> root

fuel-core-interfaces/src/model/block.rs
```diff
@@ -598,11 +598,11 @@ pub struct Genesis {
     /// The chain configs define what consensus type to use, what settlement layer to use,
     /// rules of block validity, etc.
     pub chain_config_hash: Bytes32,
-    /// The Binary Merkle Tree hash of all genesis coins.
+    /// The Binary Merkle Tree root of all genesis coins.
     pub coins_hash: Bytes32,
-    /// The Binary Merkle Tree hash of state, balances, contracts code hash of each contract.
+    /// The Binary Merkle Tree root of state, balances, contracts code hash of each contract.
     pub contracts_hash: Bytes32,
```
Voxelot, line 604: this can probably stay as hash since it's a hash of all the roots

fuel-core-interfaces/src/model/block.rs
```diff
@@ -616,18 +588,39 @@ impl From<FuelBlock> for PartialFuelBlock {
...
+/// the hash on the initial config of the network that defines the consensus rules for following
+/// blocks.
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Genesis {
+    /// The chain configs define what consensus type to use, what settlement layer to use,
+    /// rules of block validity, etc.
+    pub chain_config_hash: Bytes32,
+    /// The Binary Merkle Tree root of all genesis coins.
+    pub coins_hash: Bytes32,
```
Voxelot, line 602: hash->root: this is actually a root and not just a hash

fuel-core/src/service/genesis.rs
xgreenx: `Coin`s and `Message`s hashes are based only on data from the `ChainConfig`, so a dirty database will not affect it.
`Contract`s state assumes that the contract didn't exist before, so the state for it also should be clean.

The hashing takes into account only entities added during this initialisation, so dirty database will not break it

fuel-core-interfaces/src/model/block.rs
xgreenx: It is root of all roots=)

fuel-core/src/service/genesis.rs
xgreenx: I added check that inserted values are new

Voxelot: This should suffice for now, especially considering the extra work to impl merklizing using only chain_conf data instead of just using the merkle storage interfaces. Also contract UTXO ID's may be tricky.

fuel-client/assets/schema.sdl
```diff
@@ -263,6 +263,26 @@ type FailureStatus {
 }
 
 
+type Genesis {
```
Voxelot, line 266: Seems like we also need to add this to the cynic enums so that it doesn't use the fallback variant?

Voxelot:
We should have integ tests to verify that fetching the genesis block works as expected, and the fields are set correctly. 

Ideally we also have a test to verify that this resolves the issue for https://github.com/FuelLabs/fuels-rs/pull/710

fuel-client/assets/schema.sdl
xgreenx: Yep

fuel-chain-config/src/config/chain.rs
xgreenx: Yes, but my concern is that right now the type used for `block_production` is not right. All nodes in the network should have the same `ChainConfig` regardless of the role of the node. This type fits very well for `ProducerConfig`, but for `ChainConfig` we need another

xgreenx:
> Ideally we also have a test to verify that this resolves the issue for

Good call, it didn't work because we didn't seal during manual production=) Fixed and added a test. Now it works for `fuels-rs`=)

fuel-core/src/schema/block.rs
```diff
@@ -73,9 +77,23 @@ pub struct Header(pub(crate) FuelBlockHeader);
 
 #[derive(Union)]
 pub enum Consensus {
+    Genesis(Genesis),
     PoA(PoAConsensus),
 }
 
+#[derive(SimpleObject)]
+pub struct Genesis {
```
Voxelot, line 85: [nit] this should probably exist in some kind of separate types module

Voxelot: nm - we don't have one on the server side of graphql

Voxelot (approved):

PR #819, opened 2022-12-07
Bugfix: Race conditions between read and write into database

xgreenx:
Close https://github.com/FuelLabs/fuel-core/issues/791.

This change would be nice to do after https://github.com/FuelLabs/fuel-core/pull/812 with `Transaction<Database>` structure from `fuel-core-storage`. But while it is not merged, it is a variant of implementation without a known type. A workaround is to define a `DatabaseTransaction<Database>` trait and use the `Box<dyn DatabaseTransaction<Database>>` type instead.

The main idea of the change is to return a `UncommittedResult` from `Executor` and commit it after all updates of the `Database` at the end of the block production loop of `PoACoordinator`.


fuel-core/src/database.rs
```diff
@@ -54,6 +52,8 @@ use std::{
```
Voxelot, line 52:
```suggestion
```

these new imports should probably be sorted in with other `fuel_*` imports

fuel-core/src/schema/block.rs
```diff
@@ -436,9 +436,11 @@ impl BlockMutation {
                 vec![],
             );
 
-            let ExecutionResult { block, .. } =
-                executor.execute(ExecutionBlock::Production(block))?;
-            seal_block(&config.consensus_key, &block, &mut executor.database)?;
+            let (ExecutionResult { block, .. }, mut db_transaction) = executor
+                .execute_without_commit(ExecutionBlock::Production(block))?
+                .into();
+            seal_block(&config.consensus_key, &block, db_transaction.database_mut())?;
```
Voxelot, line 442: I wonder if there's a way we could reuse the CM to handle the production / sealing of the block, that way we don't need to be concerned with ensuring the blocks are committed with the expected sealed metadata here 🤔

xgreenx: Yep, I will add TODO for that

fuel-block-producer/src/ports.rs
```diff
@@ -21,3 +44,34 @@ pub trait Relayer: Sync + Send {
     /// Get the best finalized height from the DA layer
     async fn get_best_finalized_da_height(&self) -> anyhow::Result<DaBlockHeight>;
 }
+
+// TODO: Replace by the analog from the `fuel-core-storage`.
+pub type DBTransaction<Database> = Box<dyn DatabaseTransaction<Database>>;
+
+pub trait Executor<Database: ?Sized>: Sync + Send {
+    /// Executes the block and commits the result of the execution into the inner `Database`.
+    fn execute_and_commit(
```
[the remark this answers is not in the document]
xgreenx: You are right. Let's keep only the required methods(I removed it)

Voxelot (approved):
This is a good incremental step towards a fuel-core-storage crate, while also addressing our immediate issues with the block fetching race condition 👍🏻 

Even though the architecture doc laid out a case against having globally shared traits, I think it's justifiable in certain circumstances with traits like `Transactional`. This is because it's very  narrow in scope and doesn't bleed any irrelevant concerns across domains (e.g. changes to the TxPool db trait are still isolated from other domain crates).

Voxelot:
Would be nice if we had a way to test that this solves the race condition, but a clean solution for doing so eludes me 🤔

PR #821, opened 2022-12-08
fix: Use HttpsConnector for fuel-client's GraphQL subscriptions

bvrooman:
Related issues:
- Closes https://github.com/FuelLabs/fuel-core/issues/793

Currently, `fuel-client` makes calls to the GraphQL subscription endpoint using an `eventsource` client (built on Hyper) with an HTTP-only connector. This connector forbids HTTPs connections and throws an `INVALID_NOT_HTTP` error when it encounters one. When deploying a contract to a node deployed to an HTTPs URL, the connector throws this error. 

I was able to reproduce this error by doing the following:
- Registering the domain brandonvrooman.com 
- Getting an SSL cert for a subdomain (node.brandonvrooman.com) from LetsEncrypt and adding it to my local keychain
- Adding the url (node.brandonvrooman.com) to my `/etc/hosts` and pointing it to localhost
- Configuring NGINX to point this url (node.brandonvrooman.com) to my local node running on localhost:4000 using SSL
- Deploying a contract to https://node.brandonvrooman.com

With this fix, I was able to test that I can deploy to http and https at the respective urls:
- http://localhost:4000
- https://node.brandonvrooman.com

### DIFF
--- a/fuel-client/src/client.rs
+++ b/fuel-client/src/client.rs
@@ -9,6 +9,7 @@ use cynic::{
     QueryBuilder,
     StreamingOperation,
 };
+use eventsource_client::HttpsConnector;
 use fuel_vm::prelude::*;
 use futures::StreamExt;
 use itertools::Itertools;
@@ -192,7 +193,7 @@ impl FuelClient {
                     format!("Failed to add header to client {:?}", e),
                 )
             })?
-            .build_http();
+            .build_with_conn(HttpsConnector::with_native_roots());
 
         let mut last = None;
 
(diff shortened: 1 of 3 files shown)


fuel-client/src/client.rs
```diff
@@ -192,7 +193,7 @@ impl FuelClient {
                     format!("Failed to add header to client {:?}", e),
                 )
             })?
-            .build_http();
+            .build_with_conn(HttpsConnector::with_native_roots());
```
Voxelot, line 196: we should use `with_webpki_roots` instead to avoid depending on openssl_probe

ControlCplusControlV (approved):

PR #827, opened 2022-12-10
unify chain id generation and usage

### LINKED ISSUE #824: Unify `chain_id` generation and usage across services
https://github.com/FuelLabs/fuel-core/pull/785#discussion_r1043812652
https://github.com/FuelLabs/fuel-core/pull/785#discussion_r1043813806

> Also, these root functions should exist in the chain-config module so that they can be used without fuel-core. As fuel-p2p is in it's own crate and won't be able to use the genesis module directly.

Means also need to move all stuff regarding the genesis root calculation into `fuel-chain-config`, because it is where it belongs. Also, `Genesis` should be a part of the `ChainConfig` and `initial_state` field only as metadata without any impact on the root calculation.

leviathanbeak:
closes #824 

I renamed `Merklization` trait to `GenesisCommitment` and moved it to `chain-config` module.
Moved the implementations of the trait to their respective files.

Included few extra fields in `ChainConfig` hash calculation.
Implemented `GenesisCommitment` for `Genesis` within `chain-config` module.

Made a wrapper struct Checksum([u8; 32]).

### DIFF
--- a/fuel-chain-config/src/config/chain.rs
+++ b/fuel-chain-config/src/config/chain.rs
@@ -136,7 +136,6 @@ impl GenesisCommitment for ChainConfig {
             .chain(self.block_gas_limit.to_be_bytes())
             .chain(self.transaction_parameters.root()?)
             .chain(self.chain_name.as_bytes())
-            .chain(bincode::serialize(&self.initial_state)?)
             .finalize();
 
         Ok(config_hash)
--- a/fuel-chain-config/src/config/chain.rs
+++ b/fuel-chain-config/src/config/chain.rs
@@ -4,6 +4,10 @@ use fuel_core_interfaces::common::{
         Address,
         AssetId,
     },
+    prelude::{
+        Hasher,
+        MerkleRoot,
+    },
 };
 use itertools::Itertools;
 use rand::{
@@ -21,6 +25,8 @@ use std::{
     path::PathBuf,
     str::FromStr,
 };
+
+use crate::GenesisCommitment;
 
 use super::{
     coin::CoinConfig,
@@ -120,6 +126,31 @@ impl FromStr for ChainConfig {
                 })
             }
         }
+    }
+}
+
+impl GenesisCommitment for ChainConfig {
+    fn root(&mut self) -> anyhow::Result<MerkleRoot> {
+        // TODO: Hash settlement configuration, consensus block production
+        let config_hash = *Hasher::default()
+            .chain(self.block_gas_limit.to_be_bytes())
+            .chain(self.transaction_parameters.root()?)
+            .chain(self.chain_name.as_bytes())
+            .chain(bincode::serialize(&self.initial_state)?)
+            .finalize();
+
+        Ok(config_hash)
+    }
+}
+
+impl GenesisCommitment for ConsensusParameters {
+    fn root(&mut self) -> anyhow::Result<MerkleRoot> {
+        // TODO: Define hash algorithm for `ConsensusParameters`
+        let params_hash = Hasher::default()
+            .chain(bincode::serialize(&self)?)
+            .finalize();
+
+        Ok(params_hash.into())
     }
 }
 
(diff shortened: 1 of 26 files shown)


fuel-chain-config/src/config/chain.rs
```diff
@@ -123,6 +129,31 @@ impl FromStr for ChainConfig {
...
 }
 
+impl GenesisCommitment for ChainConfig {
+    fn root(&mut self) -> anyhow::Result<MerkleRoot> {
+        // TODO: Hash settlement configuration, consensus block production
+        let config_hash = *Hasher::default()
+            .chain(self.block_gas_limit.to_be_bytes())
+            .chain(self.transaction_parameters.root()?)
+            .chain(self.chain_name.as_bytes())
+            .chain(bincode::serialize(&self.initial_state)?)
```
xgreenx, line 139: The root/hash of the initial state is included in the `Genesis` struct, so we don’t need it here:)

fuel-chain-config/src/genesis.rs
```diff
@@ -0,0 +1,25 @@
...
+
+pub trait GenesisCommitment {
+    /// Calculates the merkle root of the state of the entity.
+    fn root(&mut self) -> anyhow::Result<MerkleRoot>;
+}
+
+impl GenesisCommitment for Genesis {
+    fn root(&mut self) -> anyhow::Result<MerkleRoot> {
+        let genesis_hash = *Hasher::default()
+            .chain(self.coins_root)
```
xgreenx, line 17: Let’s sort them in the same order as in the struct definition

fuel-chain-config/src/config/chain.rs
leviathanbeak: I see, I misunderstood @Voxelot's [comment](https://github.com/FuelLabs/fuel-core/issues/824#issuecomment-1344801975) - you mean we will add it as part of the Gensis in another PR?

fuel-core/src/cli/run/p2p.rs
```diff
@@ -133,10 +136,11 @@ pub struct P2pArgs {
...
 impl P2pArgs {
-    pub fn into_config(self, chain_config: &ChainConfig) -> anyhow::Result<P2PConfig> {
-        let checksum = *fuel_crypto::Hasher::default()
-            .chain(bincode::serialize(chain_config)?)
-            .finalize();
+    pub fn into_config(
+        self,
+        chain_config: &mut ChainConfig,
+    ) -> anyhow::Result<P2PConfig> {
+        let checksum = chain_config.root()?.into();
```
xgreenx, line 143: `Genesis::root` contains all information

fuel-chain-config/src/config/chain.rs
xgreenx: I meant `Genesis` already has `chain_config_hash` and roots of all states, the hash of `Genesis` covers all our needs:) We can use it as unique chain identifier, aka checksum

fuel-core/src/cli/run/p2p.rs
xgreenx: We can solve it in this https://github.com/FuelLabs/fuel-core/pull/828 way, for example

xgreenx (approved):

fuel-core/src/database/block.rs
```diff
@@ -139,6 +139,22 @@ impl Database {
...
+        self.iter_all(
+            Column::FuelBlockIds,
+            None,
+            None,
+            Some(IterDirection::Forward),
+        )
+        .next()
+        .ok_or(not_found!("Genesis block height"))?
+        .map(|(height, id): (Vec<u8>, Bytes32)| {
+            // safety: we know that all block heights are stored with the correct amount of bytes
```
Dentosal, line 152: nit: Maybe replace this comment with just using `.expect(...)` instead of `.unwrap()`

xgreenx (approved):

PR #828, opened 2022-12-10
Use `Genesis::root` as a chain id

xgreenx:
https://github.com/FuelLabs/fuel-core/pull/827#discussion_r1045087825

Also:
- Renamed `P2pArgs` -> `P2PArgs`
- Renamed `P2PConfig.local_keypair` -> `P2PConfig.keypair`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2498,6 +2498,7 @@ dependencies = [
  "async-trait",
  "bincode",
  "ctor",
+ "fuel-chain-config",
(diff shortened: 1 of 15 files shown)


leviathanbeak (approved):
I like this approach of p2p config lazy initialization since it's possible that some other things could also be initialized lazily.

I would've preferred that other refactors (renaming) were done separately so that it was a bit easier to review but no biggie =)

PR #818, opened 2022-12-06
State Read/Write/Clear range implementation

mitch-fuel:
Signed-off-by: Mitch Martin <mitch.martin@fuel.sh>

PR #830, opened 2022-12-13
Simplified the code of range operations

xgreenx:
Suggestions for https://github.com/FuelLabs/fuel-core/pull/818

### DIFF
--- a/fuel-core/src/database/vm_database.rs
+++ b/fuel-core/src/database/vm_database.rs
@@ -298,6 +298,21 @@ mod tests {
         Bytes32::from(bytes)
     }
 
+    fn setup_value(
+        db: &VmDatabase,
+        contract_id: ContractId,
+        start_key: U256,
+        i: usize,
+        value: &Bytes32,
+    ) {
+        let key = start_key.add(i);
+        let key = u256_to_bytes32(key);
+        let multi_key = MultiKey::new(&(contract_id.as_ref(), key.as_ref()));
+        db.database
+            .insert::<_, _, Bytes32>(&multi_key, Column::ContractsState, value)
+            .unwrap();
+    }
+
     #[test]
     fn read_single_value() {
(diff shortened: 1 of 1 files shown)


fuel-core/src/database/vm_database.rs
```diff
@@ -162,79 +169,46 @@ impl InterpreterStorage for VmDatabase {
         start_key: &Bytes32,
         range: Word,
     ) -> Result<Vec<Option<Cow<Bytes32>>>, Self::DataError> {
+        // TODO: Optimization: Iterate only over `range` elements.
         let mut iterator = self.database.iter_all::<Vec<u8>, Bytes32>(
```
xgreenx, line 173: `iter_all` returns only entries with the same `prefix` by default. We don't need to check it

fuel-core/src/database/vm_database.rs
```diff
@@ -285,28 +256,25 @@ impl InterpreterStorage for VmDatabase {
 
         let mut current_key = U256::from_big_endian(start_key.as_ref());
 
-        let transaction = self.database.transaction();
```
xgreenx, line 288: We use `VMDatabase` only with `DatabaseTransaction` right now. Also, the chance that we will use it directly with a storage database in the future is low.

Voxelot (approved):
Nice improvements! This is way less complex now.

fuel-core/src/database/vm_database.rs
Voxelot: Makes sense, it's basically already batched most of the time even if not explicitly handled here.

fuel-core/src/database/vm_database.rs
Voxelot: True, this should be covered by tests at the DB level against whatever backends it supports since the interface should guarantee the prefix is always correct.

PR #837, opened 2022-12-14
refactoring orchestrator

leviathanbeak:
This is a work in progress, the code compiles but there is some commented out code,
some of it will be removed some of it will be re-activated once other features are added.

I have started with refactoring p2p's orchestrator and network service, to provide better API without channels.
Also I have added `PeerToPeer` port to `fuel-txpool` so that `P2pAdapter` can implement it within `adapters.rs`.

I will continue cleaning this up, and finishing these two methods from `PeerToPeer` trait/port:
```rust
async fn next_gossiped_transaction(&self) -> Self::GossipedTransaction {
        todo!()
    }

    fn notify_gossip_transaction_validity(
        &self,
        message: &Self::GossipedTransaction,
        validity: fuel_txpool::ports::GossipValidity,
    ) {
        todo!()
    }
```

after that I will move to creating other p2p related ports in other modules, and implementing it in the above adapter.


Voxelot:
@leviathanbeak should we just close this PR now? Not sure if this is still relevant anymore.

leviathanbeak:
Yeah, I think this got merged via some other intermediate branch :)

PR #849, opened 2022-12-16
remove channels from our interface

leviathanbeak:
This branch has kept up to date with `teamwork/europa-16-december` but it also introduces changes of removing (some of) the channels in some of the modules/services.

Mainly I have created `PeerToPeer` port for `txpool` and consequently an Adapter for it.
I have also created a simple `Database` port for p2p. Updated p2p's Orchestrator and Network Service to match the new interface.

Warning: 
The code compiles, but there are few `todos` and warnings of unused code which will be used once other things are updated.

Test in `producer` and `txpool` currently fail, due to the change of the api a bit, so that would be a good place to start for refactoring, basically instead of providing a p2p channel now you need to provide a p2p adapter.


Dentosal (approved):
LGTM

PR #850, opened 2022-12-16
Binary folder and cleaning up dependencies

xgreenx:
Moved `fuel-core` binary into `bin/fuel-core`. It is still named `fuel-core`, but package is `fuel-core-bin`.

Added `#![deny(unused_crate_dependencies)]` almost in all crates. Sometimes it doesn't work for `dev-dependencies` so not all crates are covered. Cleaned up imports from unused crates.


Voxelot (approved):

xgreenx (commented):
Approve=)

PR #858, opened 2022-12-20
Refactoring of the `graphql-api` service to use ports for `Database`

ControlCplusControlV:
Refs https://github.com/FuelLabs/fuel-core/issues/809
Closes: https://github.com/FuelLabs/fuel-core/issues/907

It is the refactoring for the `graphql-api` service to use ports instead of direct work with `Database`. The service defines minimal database functionality and builds its functionality on top of it to process GrapQL requests.

Right now, this functionality is represented by `{}QueryContext` structures with the public methods. It is done to make the migration easier.
But later, we will need to move this functionality into corresponding traits and implement it directly for the `DatabaseTemp` type. Like:

```rust
impl {}QueryContext for DatabaseTemp {...}

or

impl<D: DatabasePort> {}QueryContext for D {...}
```

With traits, we can mock them and improve our tests.
The PR contains the following changes:

- Renamed `Transactional` into `Transaction`.
- Added `Transactional` trait with `transaction` method.
- Used `BlockId` instead of `Bytes32` in `Database`. 
- Created ports for `block.rs`, `tx.rs`, `message.rs`, `balance.rs`, `coin.rs` and updated corresponding schema. 
- Added and used in ports `BoxedIter` instead of associated types(In most cases they were `Box<dyn>` too).
- Almost removed usage of fuel_core::database from schema. Only `dap.rs` uses it right now for `VMDatabase`.
- Introduced `CompressedCoin` and `Coin`(contains `UtxoId`). `CompressedCoin` is used during work with the database, `Coin` in external API.
- Moved resources into `fuel-core-type` and simplified it by removing generics.
- Tried to re-work `dap.rs` but the change is big and also related to `fuel-core-executor`, so will do it in a separate PR.
- Updated `Transactional` to support `Transactional<dyn Database>`

Before we can extract `grapgql-api` into its crate, we need:
- Add ports for other services.
- Add iteration abstraction into `fuel-core-storage`.
- Solve the problem with `DatabaseVM`(The solution should also fit for `fuel-core-executor`).


crates/fuel-core/src/query/block.rs
```diff
@@ -0,0 +1,64 @@
...
+    ) -> std::result::Result<
+        StorageResult<(BlockHeader, Vec<fuel_core_types::fuel_types::Bytes32>)>,
+        anyhow::Error,
+    > {
+        let db = self.0;
+
+        let mut block = db
+            .storage::<FuelBlocks>()
+            .get(&id)?
+            .ok_or_else(|| anyhow!("Block height non-existent"))?
```
xgreenx, line 39: You can use `fuel_core_storage::not_found` here and return `fuel_core_storage::Error`

crates/fuel-core/src/query/block.rs
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    fn block_id(
+        &self,
+        height: u64,
+    ) -> std::result::Result<fuel_core_types::fuel_types::Bytes32, anyhow::Error> {
+        let db = self.0;
+        let id = db
+            .get_block_id(height.try_into()?)?
+            .ok_or_else(|| anyhow!("Block height non-existent"))?;
```
xgreenx, line 60: The same here

crates/fuel-core/src/query/chain.rs
```diff
@@ -0,0 +1,60 @@
...
+    fn name(&self) -> StorageResult<String>;
+    fn latest_block(&self) -> StorageResult<(BlockHeader, Vec<Bytes32>)>;
+}
+
+pub struct ChainQueryContext<'a>(pub &'a Database);
+
+impl ChainQueryData for ChainQueryContext<'_> {
+    fn latest_block(
+        &self,
+    ) -> fuel_core_storage::Result<(
```
xgreenx, line 26: Let's use the `StorageResult` alias for it(as we do in other places).

crates/fuel-core/src/query/coin.rs
```diff
@@ -0,0 +1,39 @@
...
+        let block = db
+            .storage::<Coins>()
+            .get(&utxo_id)?
+            .map(|coin| coin.clone().into_owned());
+
+        Ok(block)
+    }
+}
+
+// FilterMap<Map<IntoIter<UtxoId>, |UtxoId| -> Result<(UtxoId, Coin), Error>>, |Result<(UtxoId, Coin), Error>| -> Option<Result<(UtxoId, Coin), Error>>>
```
xgreenx, line 39: Do we need it here?

crates/fuel-core/src/schema/block.rs
```diff
@@ -221,19 +223,26 @@ impl BlockQuery {
             (Some(id), None) => id.into(),
             (None, Some(height)) => {
                 let height: u64 = height.into();
-                db.get_block_id(height.try_into()?)?
-                    .ok_or("Block height non-existent")?
+                let data = BlockQueryContext(ctx.data_unchecked());
```
xgreenx, line 226: I guess you can create it earlier and reuse bellow without double allocation

crates/fuel-core/src/schema/block.rs
```diff
@@ -221,19 +223,26 @@ impl BlockQuery {
...
+                data.block_id(height).map_err(|_| {
```
xgreenx, line 227: Hmm, `Into::into` should be enough here.

crates/fuel-core/src/schema/block.rs
```diff
@@ -221,19 +223,26 @@ impl BlockQuery {
...
 
-        let block = db
-            .storage::<FuelBlocks>()
-            .get(&id)?
-            .map(|b| Block::from(b.into_owned()));
-        Ok(block)
+        let data = BlockQueryContext(ctx.data_unchecked());
+        let block = data.block(id)??;
+
+        Ok(Some(Block {
```
xgreenx, line 242: I guess `block.into` should be enough. If not, we can add this kind of convertion=)

crates/fuel-core/src/schema/block.rs
```diff
@@ -221,19 +223,26 @@ impl BlockQuery {
...
+            header: crate::schema::block::Header(block.0),
+            transactions: block.1,
+        }))
     }
 
     async fn blocks(
```
xgreenx, line 248: Do you plan to move functionality below in a separate PR?

crates/fuel-core/src/schema/chain.rs
```diff
@@ -81,22 +77,20 @@ impl ConsensusParameters {
...
-        let block = db
-            .storage::<FuelBlocks>()
-            .get(&id)?
-            .ok_or(not_found!(FuelBlocks))?;
-        Ok(Block::from(block.into_owned()))
+        let data = ChainQueryContext(ctx.data_unchecked());
+
+        let latest_block = data.latest_block()?;
+
+        Ok(Block {
```
xgreenx, line 90: I guess `latest_block.into` should be enough

crates/fuel-core/src/schema/coin.rs
```diff
@@ -90,13 +94,12 @@ impl CoinQuery {
...
-        let db = ctx.data_unchecked::<Database>().clone();
-        let block = db
-            .storage::<Coins>()
-            .get(&utxo_id)?
-            .map(|coin| Coin(utxo_id, coin.into_owned()));
-        Ok(block)
+        let data = CoinQueryContext(ctx.data_unchecked());
+        let coin = data.coin(utxo_id.0)?;
+
+        let coin = coin.map(|coin| Coin(utxo_id.0, coin));
```
xgreenx, line 100: Maybe better to implement `From<fuel_core_types::entities::coin::Coin>` and use `into`.

crates/fuel-core/src/query/coin.rs
ControlCplusControlV: No, was testing pagination queries and comment carried over, sorry

crates/fuel-core/src/schema/block.rs
ControlCplusControlV: `Into` is implemented for this tuple variant so we need to use a literal

crates/fuel-core/src/schema/block.rs
ControlCplusControlV: Yes, was moving over just as a part of a general refactors, but paginated queries will be handled later, they just have some issues with the closure which is taking more time

.github/workflows/ci.yml
```diff
@@ -296,6 +296,11 @@ jobs:
           - os: macos-latest
             platform: darwin-arm
             target: aarch64-apple-darwin
+          - os: ubuntu-latest
```
Voxelot, line 299: Maybe this CI change should be in a separate PR? Seems unrelated to GQL changes and should be tested out separately.

Voxelot: For example, does the default `mingw-w64` image come with protoc and rocksdb installed? Have you tried running the cross commands locally with this windows setup?

ControlCplusControlV: Sorry, must've slipped in when toying around with cross-compilation yesterday. I had change branches on my terminal but VScode was still open so I think it must've picked up the changes on the wrong branch

Voxelot:
This is getting to be a very large PR, what's left todo on it? It may be good to work on getting this merged before more refactoring.

xgreenx:
It is in the stage when almost all work with `Database` is replaced with ports(except `VMDatabase`). I also vote to merge it=)

crates/fuel-core/src/query/message_proof.rs
```diff
@@ -72,65 +174,53 @@ pub async fn message_proof(
...
-
-    // Exit if the status doesn't exist or is not ready.
-    let block_id = match block_id {
-        Some(b) => b,
-        None => return Ok(None),
+    let block_id = match data
+        .transaction_status(&transaction_id)
+        .into_api_result::<TransactionStatus, StorageError>()?
+    {
+        Some(TransactionStatus::Success { block_id, .. }) => block_id,
```
xgreenx, line 181: I want to highlight that here I changed the business logic, now we don't allow verification for failed transactions.

xgreenx (dismissed):

ControlCplusControlV:
I have changes fixing the `TempDatabase` vs `Database` issue, so I would prefer to wait until that is in before merging. Awaiting some work around the debugger before committing (since its a shared PR), but discussing with @Voxelot soon

xgreenx (dismissed):

ControlCplusControlV:
PR should be good for review now that dap is separated out and instances of `TempDatabase` have been renamed

crates/fuel-core/src/graphql_api/ports.rs
```diff
@@ -0,0 +1,146 @@
...
+    + DatabaseMessages
+    + DatabaseCoins
+    + DatabaseContracts
+    + DatabaseChain
+{
+}
+
+/// The database port expected by GraphQL API service.
+#[cfg(test)]
+pub trait DatabasePort:
```
Voxelot, line 58: [nit] why not just use define a trait extension in tests?

crates/fuel-core/src/schema/chain.rs
```diff
@@ -83,22 +77,15 @@ impl ConsensusParameters {
...
-            .get(&id)?
-            .ok_or(not_found!(FuelBlocks))?;
-        Ok(Block::from(block.into_owned()))
+        let query = BlockQueryContext(ctx.data_unchecked());
+
+        let latest_block = query.latest_block()?.into();
+        Ok(latest_block)
     }
 
     async fn base_chain_height(&self) -> U64 {
```
Voxelot, line 91: just realized we should probably update this to fetch da_height from the relayer

crates/fuel-core/src/query/message_proof.rs
Voxelot: This is probably ok since messages are only set by the VM currently. However there's nothing specifically preventing failed scripts from having a valid message output in the future.

crates/fuel-core/src/database/relayer.rs
```diff
@@ -21,7 +21,7 @@ use fuel_core_types::{
 
 #[async_trait::async_trait]
 impl RelayerDb for Database {
-    async fn insert_message(
```
Voxelot, line 24: 👍🏻

Voxelot (approved):

PR #860, opened 2022-12-21
Unify the work of runnable services

xgreenx:
Created a `fuel-core-services` crate to have common stuff of services.
Added initial implementation of `ServiceRunner`.
Used `ServiceRunner` to implement `fuel_core_poa::Service`. Removed `fuel-core-bft` from `fuel-core` because we don't use it now.


crates/services/src/service.rs
```diff
@@ -0,0 +1,244 @@
...
+        // If the state after update is not `Started` then return to stop the service.
+        if !state.borrow().started() {
+            return
+        }
+
+        // We can panic here, because it is inside of the task.
+        service
+            .initialize()
+            .await
+            .expect("The initialization of the service field.");
```
Voxelot, line 215:
```suggestion
            .expect("The initialization of the service failed.");
```

xgreenx (commented):
Approve

Voxelot (approved):

PR #862, opened 2022-12-21
Fixed Request Body Limit Size

ControlCplusControlV:
Fixed a common error of 
```
thread 'test' panicked at 'called `Result::unwrap()` on an `Err` value: ProviderError("Server returned 413: Failed to buffer the request body: length limit exceeded")', 
```

Removed the default request body limit, and instead pushed it to the nginx configuration to handle request body size limiting. Local nodes will now allow unbounded size for API requests, which should now be handled at a proxy level

Based off of refactor branch because changes aren't needed in prod (where it's fine if limit persists), but can still trickle down through tooling so developers are unblocked locally.


ControlCplusControlV:
Closed in favor of #865 which performs the proper checks

PR #864, opened 2022-12-22
Fix Windows Compilation

### LINKED ISSUE #463: Windows Support
Support Windows eventually, not a priority atm. Maybe add some CI Job so that compatibility is ensured. At the time of writing this is blocked by in graph_api.rs](https://github.com/FuelLabs/fuel-core/blob/master/fuel-core/src/service/graph_api.rs#L25). But this is very bottom of priority list so really whenever anyhow has free time

ControlCplusControlV:
Windows would just not compile because a missed import

Closes #463 

Clearer Protoc instructions are still needed but now there is nothing technically blocking it. Although I would actually be in favor of just distributing a windows binary in the future and telling people to vendor it


Voxelot (approved):
nice find

PR #869, opened 2022-12-23
Fix Request Body Size Limit

ControlCplusControlV:


Voxelot (approved):
verified the new body size limit works as expected against the preview environment

ControlCplusControlV:
Auto-merging once CI finishes as I've also verified request body limit is disabled locally on the binary

PR #875, opened 2023-01-04
Refactor of `GraphQL` service and `FuelService` to use `ServiceRunner`

xgreenx:
The final change of https://github.com/FuelLabs/fuel-core/pull/860 epic. Ref https://github.com/FuelLabs/fuel-core/issues/809

Reworked `RunnableService` to be `RunnableService` and `RunnableTask`. `RunnableService::initialize` replaced with the `RunnableService::into_task` method. `into_task` returns a runnable task that implements the `RunnableTask` trait. `into_task` may return another type after initialization.

Updated all services to implement new traits. Implemented GraphQL service via `ServiceRunner`. Extracted the graph QL logic into a separate module(preparation to move this service into its own crate). 

Re-used`ServiceRunner` for `FuelService`. Replaced `Modules` with `SharedState` and `SubServices`.

Added a new `Starting` state of the service lifecycle. Added functions to allow to await `Started` or `Stop` state.


crates/fuel-core/src/graph-api/mod.rs
```diff
@@ -0,0 +1,21 @@
+use fuel_core_types::{
```
Voxelot, line 1: no mod.rs, please keep consistent with rust 2018 modules.

Voxelot: I'd prefer to keep the module style consistent for now, and maybe add a note to move this file to lib.rs in a separate commit to avoid losing the history.

crates/fuel-core/src/service.rs
```diff
@@ -1,56 +1,78 @@
...
+        let bound_address = runner.shared.graph_ql.bound_address;
+        Ok(FuelService {
+            bound_address,
+            shared,
+            runner,
+        })
+    }
+
+    /// Creates and starts fuel node instance from service config
+    #[tracing::instrument(skip(config))]
```
Voxelot, line 74: we should probably move this tracing annotation to the above `new` method so that tracing is always activated regardless of which entrypoint is used to start the svc.

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -133,22 +117,51 @@ pub async fn start_modules(
...
+        network: network.shared.clone(),
+        #[cfg(feature = "relayer")]
+        relayer: relayer.as_ref().map(|r| r.shared.clone()),
+        graph_ql: graph_ql.shared.clone(),
+    };
+
+    #[allow(unused_mut)]
+    // `FuelService` starts and shutdowns all sub-services in the `services` order
+    let mut services: SubServices = vec![
+        // GraphQL should be shutdown first, so let's start it first.
```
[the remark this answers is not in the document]
Voxelot: ah nm just checked the into_task impl

crates/services/p2p/src/service.rs
```diff
@@ -111,20 +113,28 @@ impl<D> Task<D> {
...
+    type Task = Task<D>;
 
     fn shared_data(&self) -> Self::SharedData {
         self.shared.clone()
     }
 
-    async fn initialize(&mut self) -> anyhow::Result<()> {
-        self.p2p_service.start()
+    async fn into_task(mut self, _: &StateWatcher) -> anyhow::Result<Self::Task> {
+        self.p2p_service.start()?;
```
Voxelot, line 128: does p2p monitor state watcher for the shutdown signal?

Voxelot (approved):

crates/services/p2p/src/service.rs
Voxelot: nvm this is handled at the service runner level

PR #876, opened 2023-01-04
add peer manager

leviathanbeak:
This `PeerManager` replaces existing `PeerInfo` network behaviour.

In this first implementation, `PeerManager` maintains the connections of reserved nodes, and tries to re-connect to them if the disconnect happens. Also it always keeps extra connection slots for reserved peers if they are not available at the moment.
If another non-reserved peer tries to take that reserved slot - it will be immediately disconnected.


crates/services/p2p/src/discovery.rs
```diff
@@ -146,9 +149,6 @@ impl NetworkBehaviour for DiscoveryBehaviour {
             }) => {
                 if *remaining_established == 0 {
                     self.connected_peers.remove(peer_id);
-                    self.events
-                        .push_back(DiscoveryEvent::Disconnected(*peer_id));
```
Voxelot, line 150: why do we no longer need to notify the swarm about disconnected peers?

crates/services/p2p/src/peer_manager.rs
```diff
@@ -302,6 +300,18 @@ impl NetworkBehaviour for PeerInfoBehaviour {
         cx: &mut Context<'_>,
         params: &mut impl PollParameters,
     ) -> Poll<NetworkBehaviourAction<Self::OutEvent, Self::ConnectionHandler>> {
+        if let Some(event) = self.peer_manager.pending_events.pop_front() {
+            return Poll::Ready(NetworkBehaviourAction::GenerateEvent(event))
+        }
+
+        if self.health_check.poll_tick(cx).is_ready() {
+            if let Some(peer_id) = self.peer_manager.get_disconnected_reserved_peer() {
+                return Poll::Ready(NetworkBehaviourAction::GenerateEvent(
```
Voxelot, line 309: Consider adding DEBUG level logging here to indicate reconnect attempts to reserved nodes

crates/services/p2p/src/peer_manager.rs
```diff
@@ -302,6 +300,18 @@ impl NetworkBehaviour for PeerInfoBehaviour {
         cx: &mut Context<'_>,
         params: &mut impl PollParameters,
     ) -> Poll<NetworkBehaviourAction<Self::OutEvent, Self::ConnectionHandler>> {
+        if let Some(event) = self.peer_manager.pending_events.pop_front() {
+            return Poll::Ready(NetworkBehaviourAction::GenerateEvent(event))
+        }
+
+        if self.health_check.poll_tick(cx).is_ready() {
+            if let Some(peer_id) = self.peer_manager.get_disconnected_reserved_peer() {
```
Voxelot, line 308: Should we push Reconnect events for all disconnected reserved nodes? Otherwise it seems like it could take a while to fully reconnect if we only reconnect to one node per heartbeat interval.

crates/services/p2p/src/p2p_service.rs
```diff
@@ -219,19 +212,22 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
...
+                let mut peers = self.get_peers_ids();
+                let peers_count = self.swarm.behaviour().total_peers_connected();
+
+                if peers_count == 0 {
                     return Err(RequestError::NoPeersConnected)
                 }
-                let rand_index = rand::thread_rng().gen_range(0..connected_peers.len());
-                *connected_peers.keys().nth(rand_index).unwrap()
+
+                let rand_index = rand::thread_rng().gen_range(0..peers_count);
```
Voxelot, line 222: consider using [SliceRandom](https://docs.rs/rand/latest/rand/seq/trait.SliceRandom.html)

freesig: I think maybe [IteratorRandom](https://docs.rs/rand/latest/rand/seq/trait.IteratorRandom.html) would make more sense as the index is just used in an iterator below.

crates/services/p2p/src/peer_manager.rs
leviathanbeak: done

crates/services/p2p/src/peer_manager.rs
leviathanbeak: good point, I updated the code so that it collects all disconnected peers

crates/services/p2p/src/p2p_service.rs
```diff
@@ -635,6 +679,180 @@ mod tests {
...
+                            // 0. This is the initial connection with the reserved node
+                            if !previously_connected {
+                                // 1. Disconnect and ban the reserved node
+                                // this is done so that other nodes in the network try to take the slot of the reserved node
+                                let _ = sentry_node.swarm.disconnect_peer_id(reserved_node_peer_id);
+                                sentry_node.swarm.ban_peer_id(reserved_node_peer_id);
+                                previously_connected = true;
+                            } else {
+                                // 3. We have successfully re-connected.
+                                // Multiple other nodes in the meantime have tried to establish the connection
```
Voxelot, line 751: It seems like there could be lots of race conditions that don't exercise this behavior. For instance, the node may unban and reconnect to the reserved node before the other nodes get a chance to connect. Seems like it would be simpler to only start the reserved node once the we've connected to the other peers first.

crates/services/p2p/src/peer_manager.rs
```diff
@@ -453,3 +475,217 @@ impl NetworkBehaviour for PeerInfoBehaviour {
...
+            let non_reserved_peers_connected =
+                self.connected_peers.len() - self.reserved_peers_connected_count();
+
+            // check if there is no more space for non-resereved peers
+            if non_reserved_peers_connected >= self.non_reserved_peers_allowed {
+                // todo/potential improvement: once `Peer Reputation` is implemented we could check if there are peers
+                // with poor reputation and disconnect them instead?
+
+                // Too many peers already connected, disconnect the Peer
+                self.pending_events.push_back(PeerInfoEvent::TooManyPeers {
```
Voxelot, line 570: Could this be done on handshake rather than after they've (temporarily) consumed a connection slot? If peers are repeatedly reconnecting and consuming available connections, this could explain why there's a big variance in running the reconnect test as it basically has to bruteforce connections until a reserved node gets lucky.

crates/services/p2p/src/peer_manager.rs
```diff
@@ -453,3 +475,217 @@ impl NetworkBehaviour for PeerInfoBehaviour {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn get_random_peers(size: usize) -> Vec<PeerId> {
+        (0..size).map(|_| PeerId::random()).collect()
+    }
+
+    #[test]
+    fn test_peer_manager_struct() {
```
Voxelot, line 616: This test should probably be split into smaller units, it's not immediately obvious what being covered here or what the intention of the test is without reading through each assertion.

crates/services/p2p/src/p2p_service.rs
```diff
@@ -133,8 +131,33 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
...
+        };
+
+        let max_established_incoming = {
+            if config.reserved_nodes_only_mode {
+                // If this is a guarded node,
+                // it should not receive any incoming connection requests.
+                // Rather, it will send outgoing connection requests to its reserved nodes
+                0
+            } else {
+                total_connections / 2
```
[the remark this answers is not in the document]
leviathanbeak: Not necessarily, this defines the number of max established "incoming" connections, I arbitrarily set it to be 50% of total connections allowed, so 50:50 between incoming and outgoing connections. But we can have this configurable in p2p config if we want, plus maybe we should add config options for max pending connections, for both outgoing and incoming.

crates/services/p2p/src/p2p_service.rs
leviathanbeak: yes that would be even better approach, I will update it

crates/services/p2p/src/peer_manager.rs
leviathanbeak: Yes, that is very trye, it's just basically waiting for it's turn to connect and stay connected, while other surplus nodes get disconnected. I will experiment a bit with handshake, maybe we could hold some reference to current connection state in that protocol and based on that complete the handshake or reject the connection.

As for the cooldown, I'm not aware there's such a thing, we could have a wrapper over ban/unban if needed

crates/services/p2p/src/peer_manager.rs
leviathanbeak: separated into 3 different tests

crates/services/p2p/src/p2p_service.rs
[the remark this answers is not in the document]
Voxelot: sounds good, created a ticket for pending connections config

crates/services/p2p/src/peer_manager.rs
```diff
@@ -0,0 +1,815 @@
...
+    LatestPing(Duration),
+}
+
+/// Manages Peers and their events
+#[derive(Debug, Default, Clone)]
+struct PeerManager {
+    pending_events: VecDeque<PeerInfoEvent>,
+    non_reserved_connected_peers: HashMap<PeerId, PeerInfo>,
+    reserved_connected_peers: HashMap<PeerId, PeerInfo>,
+    reserved_peers: HashSet<PeerId>,
```
Voxelot, line 515: we should add a counter for the number of reserved peers in our prometheus metrics

cc @ControlCplusControlV

Voxelot (approved):

PR #889, opened 2023-01-05
fuel-core-sync

### LINKED ISSUE #880: [Sync] Create sync service wrapper for sync tasks
The sync service will be responsible for managing two sub-tasks (sync and import). Setup a [RunnableService](https://github.com/FuelLabs/fuel-core/blob/master/crates/services/src/service.rs#L38) in fuel-core-sync that handles the startup and shutdown of both the sync and import tasks.

### LINKED ISSUE #881: [Sync] Sync task
The sync task receives information about the block height of peers via a stream from P2P. It primarily filters out previously seen block heights and then updates a mutexed shared state with the import task. It will never generate back pressure because it can always update the state and notify import without awaiting.

```mermaid
flowchart LR
	H{Heatbeat} --> |BlockHeight|S
	S --> |height > best_height|F[stream::filter ]
	F -.->|Update| ST{{State}}
	F -.->|Notify| Import{Import}
```

## Shared State:

```rust
/// State that is shared between the sync and import tasks
struct State {
	/// The highest seen height so far.
  best_seen_height: Option<Height>,
	/// The range of heights that are currently being imported.
  in_flight_heights: Option<Range<Height>>,
}

// Empty state means we have not observed any heights and
// we have not imported any headers.
State {
	best_seen_height: None,
	in_flight_height: None,
}

// There is a new height that has been observed but not imported.
State {
	best_seen_height: Some(5),
	in_flight_height: None,
}

// Currently importing headers 0..=5.
State {
	best_seen_height: Some(5),
	in_flight_height: Some(0..=5),
}

// Have imported headers 0..=3 but still need to import 4..=5.
State {
	best_seen_height: Some(5),
	in_flight_height: Some(3..=5),
}

// A newer height has been seen while the stream is importing headers.
State {
	best_seen_height: Some(7),
	in_flight_height: Some(4..=5),
}

// All caught up with the current best seen.
State {
	best_seen_height: Some(7),
	in_flight_height: Some(7..=7),
}
```

## Ports

```rust
trait P2P {
    /// Get a stream of block heights reported by peers
    async fn block_heights(&self) -> BoxStream<BlockHeight>;
}
```

### LINKED ISSUE #882: [Sync] Import task
The import task will check for work to do. 

- If there is none it will wait to be notified about new work.
- If there is some it will run the import stream potentially getting back pressure from sections of the stream. In other words the import task will not take on new work until the current work is complete

## Task Loop
```mermaid
flowchart TD
	I(Import Loop) -.-> |Check if work to do| ST{{State}}
	ST -.-> |Yes| I
	ST -.-> |No| N
	I --> |range: i..=j| IS[import stream]
	IS --> N[wait to be notified]
	N --> I
```
## Import Stream

- [ ] get sealed header associated with a block height
- [ ] validate the sealed metadata with the consensus module
- [ ] fetch all transactions associated with the block header
- [ ] send full sealed block to the importer

If any of these stages fail, drop the stream and reset the shared state.

### Stream example diagram

```mermaid
	
flowchart TD
	R[range_iterator] --> |sub range m..=n|GH[get_header_range]
	GH --> |get_header future| BP1
	subgraph BP1[Back pressure if MAX_GET_HEADER_REQUESTS is reached ]
		B1[stream::buffered MAX_GET_HEADER_REQUESTS]
	end
	B1 --> FL[stream::flattern]
	FL --> |Header| GT[get_transactions]
	GT --> |get_transactions future| BP2
	subgraph BP2[Back pressure if MAX_GET_TXNS_REQUESTS is reached ]
		B2[stream::buffered MAX_GET_TXNS_REQUESTS]
	end
	B2 --> |Block| BP3
	subgraph BP3[Back pressure while execute is running]
		SB[tokio::spawn_blocking]
		SB --> E[Execute]
	end
```

## Ports
APIs that the import task will need from other services

```rust
trait P2P {
    /// Get the sealed block associated with the provided block height
    async fn get_sealed_block_header(height: BlockHeight) -> Result<(Sealed<Header>, PeerId)>;
    /// Get all transactions associated with a block id from a specific peer
    async fn get_block_transactions(block_id: BlockId, peer_id: PeerId) -> Result<Vec<Transaction>>;
}

trait Importer {
    /// Execute the provided block in verification mode and permanently commit the state if it's valid
    async fn execute_and_commit(&self, block: SealedBlock) -> Result<()>;
}

trait Consensus {
    /// Given a previous block header, validate the next header follows the consensus rules
    async fn validate_next_header(previous_header: Sealed<Header>, new_header: Sealed<Header>);
}

```

freesig:
- Closes #881 
- Closes #882 
- Closes #880


crates/services/sync/src/import.rs
```diff
@@ -0,0 +1,331 @@
...
+}
+
+/// Long running task that imports blocks from the network and executes them.
+pub(super) async fn import<P, E, C>(
+    state: SharedMutex<State>,
+    notify: Arc<Notify>,
+    params: Params,
+    ports: Ports<P, E, C>,
+    shutdown: Shutdown,
+    // Needed for testing. Would be nice to find a better way to do this.
```
xgreenx, line 67: If you had the `run` function with one iteration of the importer service, you could easily get the result after its execution and compare them ;)

crates/services/sync/src/service.rs
```diff
@@ -1,61 +1,232 @@
...
+        // Check if the state of the task runner is still running.
+        if !watcher.borrow().started() {
+            // The task runner has stopped so we kill the sync and import tasks.
+            self.kill_switch.kill_all();
+            if let Some(sync_task) = self.sync_task.take() {
+                let _ = sync_task.await;
+            }
+            if let Some(import_task) = self.import_task.take() {
+                let _ = import_task.await;
+            }
```
xgreenx, line 145: Could you create a `shutdown` method in the `RunnableTask` that we discussed and do it there, please? In this case, you don't need to `!watcher.borrow().started()`

crates/services/sync/src/service.rs
```diff
@@ -1,61 +1,232 @@
...
-        if join.is_none() {
-            *join = Some(tokio::spawn(async {}));
+        // Check we still have tasks to run otherwise return false.
+        match (&mut self.sync_task, &mut self.import_task) {
+            (Some(sync_task), Some(import_task)) => {
+                // Select over the task runner state, the sync task and the import task.
+                // Match on either the sync task or the import task completing.
+                // If the task runner state changes then we check it below.
+                if let Either::Right((join_handles, _)) =
+                    select(watcher.changed().boxed(), select(sync_task, import_task))
```
xgreenx, line 96:
```suggestion
                    select(watcher.while_started().boxed(), select(sync_task, import_task))
```

xgreenx (commented):
The change looks solid, so many tests=D I didn't review all files, will continue later. I left several questions and suggestion

crates/services/sync/src/service.rs
```diff
@@ -1,61 +1,232 @@
...
+                        .await
+                {
+                    match join_handles {
+                        // Sync task has completed.
+                        Either::Left((sync_result, import_task)) => match sync_result {
+                            // Sync task has panicked.
+                            Err(e) if e.is_panic() => {
+                                // Kill the import task and resume the panic.
+                                self.kill_switch.kill_all();
+                                let _ = import_task.await;
```
ControlCplusControlV, line 106: The main reason for handling these results I'd assume is that the task is being killed anyway? Is there any chance for a messy shutdown with errors being reason for panic here?

crates/services/sync/src/state.rs
```diff
@@ -0,0 +1,212 @@
...
+            // Currently processing a range and recording a commit.
+            Status::Processing(range) => match height.cmp(range.end()) {
+                // The commit is less than the end of the range, so the range
+                // is still being processed.
+                std::cmp::Ordering::Less => {
+                    Some(Status::Processing(height.saturating_add(1)..=*range.end()))
+                }
+                // The commit is equal or greater than the end of the range,
+                // so the range is fully committed.
+                std::cmp::Ordering::Equal | std::cmp::Ordering::Greater => {
```
ControlCplusControlV, line 81: Importing `std::cmp::Ordering` would make this cleaner since it is used multiple times

ControlCplusControlV (commented):
Read through changes but didn't have too much to say from my end, so defer to others as well. But looks good beyond Green's suggestions

crates/services/src/lib.rs
```diff
@@ -19,13 +19,27 @@ pub mod stream {
         core::pin::Pin<Box<dyn Stream<Item = T> + Send + Sync + 'static>>;
 }
 
+/// The source of some network data.
+pub struct SourcePeer<T> {
```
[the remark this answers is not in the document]
leviathanbeak: I have created something similar in my hearbeat PR, although this one is more generic, I could use it in my PR as well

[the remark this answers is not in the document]
leviathanbeak: but agree, it should be moved there

crates/services/src/lib.rs
```diff
@@ -19,13 +19,27 @@ pub mod stream {
...
+    /// The source of the data.
+    pub peer_id: PeerId,
+    /// The data.
+    pub data: T,
+}
+
+/// Placeholder for a peer id
+pub struct PeerId;
```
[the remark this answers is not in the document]
freesig: Yeh that is basically why I left this here as a placeholder because I wasn't sure what it should be. @leviathanbeak Do you know how long the bytes is for the PeerId? I could change this to `pub struct PeerId([u8; 32])` or whatever the length is.

crates/services/src/lib.rs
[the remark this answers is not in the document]
freesig: Will do

crates/services/sync/src/import.rs
freesig: But then there's no way to wait on the notify.

xgreenx: First, you can send the notification and after, you can do `run.await`

crates/services/sync/src/service.rs
freesig: Yeh I thought about that but you can't really combine two panics (well not that I know of) so we don't really have a choice

crates/services/sync/src/service.rs
freesig: If we don't check the watcher here then there's no way to shutdown the tasks based on the state changing to stopped though. Maybe I don't understand what you mean?

xgreenx: It is already checked in the in the `ServiceRunner` 
image

crates/services/sync/src/service.rs
```diff
@@ -1,61 +1,232 @@
...
+{
+    const NAME: &'static str = "fuel-core-sync";
+
+    type SharedData = ();
+
+    type Task = Task;
+
+    fn shared_data(&self) -> Self::SharedData {}
+
+    async fn into_task(self, _: &StateWatcher) -> anyhow::Result<Self::Task> {
```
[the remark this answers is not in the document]
freesig: I disagree that it is simpler. Using the service means I have no control over how it is implemented and just need to hope it fits both tasks. The tasks have unique requirements like awaiting on a notify and the sync task being unable to return an error which is not possible to control with the RunnableService trait. I get that we want to standadize the launching of each service but that should not extend into prescibing how services manage their internal spawned tasks.

crates/types/src/blockchain/block.rs
```diff
@@ -80,6 +80,23 @@ impl Block<Transaction> {
...
     }
 
+    /// Try creating a new full fuel block from a [`BlockHeader`] and
+    /// **previously executed** transactions.
+    /// This will fail if the transactions don't match the header.
+    pub fn try_from_executed(
+        header: BlockHeader,
+        mut transactions: Vec<Transaction>,
+    ) -> Option<Self> {
+        let transaction_ids: Vec<_> =
```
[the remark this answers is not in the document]
freesig: Actually the `txRoot` is the merkle tree of the transaction bytes **not** the transaction ids. The reason is because the id is the same regardless of whether the transaction is executed or not but the merkle root needs to be based on executed transactions. This is a good reason why we should have a seperate type for executed vs zeroed transactions. @Voxelot the [spec](https://fuellabs.github.io/fuel-specs/master/protocol/block_header.html) doesn't make this clear.

crates/types/src/blockchain/header.rs
```diff
@@ -178,6 +178,14 @@ impl BlockHeader {
             self.hash()
         }
     }
+
+    /// Validate the transactions match the header.
+    pub fn validate_transactions(&self, transactions: &[Vec<u8>]) -> bool {
```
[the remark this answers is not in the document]
freesig: See my response to the previous comment

crates/services/sync/src/service.rs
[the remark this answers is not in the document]
xgreenx: > like awaiting on a notify 

You can do it as you did in the current implementation. It is exactly the body of the `run` method:

image

> the sync task being unable to return an error

But the service will log the error, you don't need to return it.

crates/types/src/blockchain/block.rs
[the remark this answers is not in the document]
xgreenx: While we are awaiting a response from @Voxelot .

I made that conclusion based on the name of the variables=) If we really need to use `tx_bytes` instead of `tx_id`(transaction id is a `hash` of the `to_bytes`), then we need to update the naming. Because `transaction_ids` means `identifiers of transactions`.

crates/services/src/lib.rs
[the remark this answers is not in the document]
freesig: Looks like it's variable so will make this a vec

crates/services/sync/src/import.rs
freesig: I don't understand?

crates/services/sync/src/service.rs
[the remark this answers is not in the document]
freesig: I don't understand, you're saying to await the notify in run?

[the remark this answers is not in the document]
xgreenx: Yep, as you already do that, it would be notification from the parent sync service

crates/services/sync/src/state.rs
```diff
@@ -0,0 +1,212 @@
...
+
+/// If a commit is made to a height that is
+/// below the existing committed height this is
+/// new evidence and we check if there is a gap between
+/// the existing committed height and the new commit.
+///
+/// This case should not occur but because we must handle
+/// it then the most resilient way is to assume that we
+/// should re-process the gap.
+fn commit_creates_processing(
```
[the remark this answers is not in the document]
freesig: So this is a bit of a hard one. Really by design we can't ever have this happen because the commits are run in order and there's no state that can lead to re-running a commit. However I can't handle this case as an error at the stream level because there's nothing you can do so for the sake of completeness I needed to do something here. If we get a commit that is lower then the current commit then something is wrong and my way of looking at this is we can't trust the higher commit value so it should reprocess the headers in between. But you could make an argument either way.

crates/types/src/blockchain/block.rs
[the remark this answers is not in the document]
freesig: @Voxelot Can you comment on this?

crates/services/sync/src/import.rs
```diff
@@ -0,0 +1,363 @@
...
+        } = &self;
+        // Request up to `max_get_header_requests` headers from the network.
+        get_header_range_buffered(range.clone(), params, p2p.clone())
+        .map({
+            let p2p = p2p.clone();
+            let consensus_port = consensus.clone();
+            move |result| {
+                let p2p = p2p.clone();
+                let consensus_port = consensus_port.clone();
+                async move {
```
xgreenx, line 156: Hmm, it seems we can create this function one time during the creation of the task and use it here. Instead of creating it on each iteration.

But it will affect your current code, so I guess it is enough for now to create it at the begin of the function.

crates/services/sync/src/import.rs
```diff
@@ -0,0 +1,363 @@
...
+        .scan_none_or_err()
+        // Continue the stream until the shutdown signal is received.
+        .take_until({
+            let mut s = shutdown.clone();
+            async move { s.while_started().await }
+        })
+        .then({
+            let state = state.clone();
+            let executor = executor.clone();
+            move |block| {
```
xgreenx, line 205: The same here, we can create it only one time

crates/services/sync/src/import/back_pressure_tests.rs
```diff
@@ -0,0 +1,273 @@
...
+    Input {
+        executes: Duration::from_millis(10),
+        ..Default::default()
+    },
+    State::new(None, 50),
+    Config{
+        max_get_header_requests: 10,
+        max_get_txns_requests: 10,
+    }
+    => is less_or_equal_than Count{ headers: 10, transactions: 10, consensus_calls: 10, executes: 1, blocks: 21 }
```
xgreenx, line 100: Hmm, how does comparation work? It will compare each field and only if all fields `<=` then `true`?

crates/services/sync/src/import.rs
```diff
@@ -0,0 +1,363 @@
...
+            params,
+            p2p,
+            executor,
+            consensus,
+        }
+    }
+}
+impl<P, E, C> Import<P, E, C>
+where
+    P: PeerToPeerPort + Send + Sync + 'static,
```
xgreenx, line 97: [nit] I guess you can make `Send + Sync` default bound for all ports like `PeerToPeerPort: Send + Sync`

crates/services/sync/src/service.rs
```diff
@@ -1,61 +1,198 @@
...
+    current_fuel_block_height: BlockHeight,
+    p2p: P,
+    executor: E,
+    consensus: C,
+    params: Config,
+) -> anyhow::Result<ServiceRunner<SyncTask<P, E, C>>>
+where
+    P: ports::PeerToPeerPort + Send + Sync + 'static,
+    E: ports::BlockImporterPort + Send + Sync + 'static,
+    C: ports::ConsensusPort + Send + Sync + 'static,
```
xgreenx, line 49: Hmmm, do we need those constraints here? Maybe we can remove them like:
```suggestion
where
    P: ports::PeerToPeerPort,
```

xgreenx: And for types down, we can remove all constraints, because we don't need them during creation of tasks=)

crates/types/src/blockchain/block.rs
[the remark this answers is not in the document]
Voxelot: This code looks correct to me, we do need the tx bytes for the txroot. Although I agree that maybe a different name from transaction ids makes sense. Especially since this is just bytes that haven't had any additional hashing or anything else done to them, so maybe `transction_ids` -> `tx_bytes`?

crates/types/src/blockchain/header.rs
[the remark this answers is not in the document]
Voxelot: could also call this param `transaction_bytes`

crates/types/src/services/p2p.rs
```diff
@@ -43,6 +43,19 @@ pub type TransactionGossipData = GossipData<Transaction>;
...
+pub struct SourcePeer<T> {
+    /// The source of the data.
+    pub peer_id: PeerId,
+    /// The data.
+    pub data: T,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+/// Opaque peer identifier.
+pub struct PeerId(Vec<u8>);
```
Voxelot, line 57: We should make a tech-debt cleanup task for updating the other gossip data types here to use this new PeerId type.

crates/services/sync/src/import/back_pressure_tests.rs
Voxelot: I'm assuming it just uses the existing derivation of `Ord`/`PartialOrd` for the `Count` type

Voxelot (approved):

crates/services/sync/src/service.rs
freesig: @xgreenx we actually do because you can't add extra constraints to the `run` on `RunnableTask` so we are forced to put them on the type.

crates/services/sync/src/import.rs
freesig: Is that really a concern though? Very likely this just gets optimized away by the compiler.

freesig: Also we can't do that because this takes the closure owned so you would need to clone the closure anyway.

crates/services/sync/src/import/back_pressure_tests.rs
freesig: Yeh exactly. It's just checking the tests never exceed those maximum values

crates/services/sync/src/import.rs
freesig: There's no reason for these traits to be send and sync it's only in this usecase that it's required.

crates/services/sync/src/service.rs
xgreenx: You can add constraints on the `impl` section itself, we did that for all other services.

PR #894, opened 2023-01-10
fuel authentication with connection state

leviathanbeak:
following @Voxelot 's [comment](https://github.com/FuelLabs/fuel-core/pull/876#discussion_r1065400628) this is a prototype of how would we move checking of peer slot availability during handshake instead of after the connection is established.

The approach is that we share a state between PeerManager and Fuel Authenticator (that does the handshake), we update the state only on PeerManager side, but we read the state from Fuel Authenticator every time we want to establish a connection with a peer

### DIFF
--- a/crates/services/p2p/src/config/connection_tracker.rs
+++ b/crates/services/p2p/src/config/connection_tracker.rs
@@ -1,0 +1,50 @@
+use super::{
+    fuel_authenticated::Approver,
+    peer_ids_set_from,
+};
+use crate::peer_manager::ConnectionState;
+use libp2p::{
+    Multiaddr,
+    PeerId,
+};
+use std::{
+    collections::HashSet,
+    sync::{
+        Arc,
+        RwLock,
+    },
+};
+
+/// A `ConnectionTracker` allows either Reserved Peers or other peers if there is an available slot.
+/// It is synced with `PeerManager` which keeps track of the `ConnectionState`.
+#[derive(Debug, Clone)]
+pub(crate) struct ConnectionTracker {
+    reserved_nodes: HashSet<PeerId>,
+    connection_state: Arc<RwLock<ConnectionState>>,
+}
+
+impl ConnectionTracker {
+    pub(crate) fn new(
+        reserved_nodes: &[Multiaddr],
+        connection_state: Arc<RwLock<ConnectionState>>,
+    ) -> Self {
+        Self {
+            reserved_nodes: peer_ids_set_from(reserved_nodes),
+            connection_state,
+        }
+    }
+}
+
+impl Approver for ConnectionTracker {
+    fn allow_peer(&self, peer_id: &PeerId) -> bool {
+        if self.reserved_nodes.contains(peer_id) {
+            return true
+        }
+
+        if let Ok(connection_state) = self.connection_state.read() {
+            return connection_state.available_slot()
+        }
+
+        false
+    }
+}
--- a/crates/services/p2p/src/behavior.rs
+++ b/crates/services/p2p/src/behavior.rs
@@ -1,3 +1,8 @@
+use std::sync::{
+    Arc,
+    RwLock,
+};
+
 use crate::{
     codecs::NetworkCodec,
     config::Config,
@@ -11,6 +16,7 @@ use crate::{
         topics::GossipTopic,
     },
     peer_manager::{
+        ConnectionState,
         PeerInfo,
         PeerInfoEvent,
         PeerManagerBehaviour,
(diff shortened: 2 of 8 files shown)


leviathanbeak:
I have created `ConnectionState` as a struct instead of a simple `bool` so that the intent is more clearer and in the future we might add more state to it.
Having said that, maybe using `AtomicBool` instead wouldn't be a bad idea?

I have used `RwLock` from std lib, it's blocking, maybe we should use something else? tokio has one as well, but that would turn a lot of non-async code into async one instead.

I have still kept all the logic of checking for available slots in the Peer Manager and disconnecting extra peers*. 
Because even with this `FuelAuthenticated` approach sometimes we get extra peers connected in spite of `allow_peer()` returning false! - this would necessitate more debugging with libp2p not sure do we have the time to dig in now?

* - this might happen because in that test `max_peers_connected_works` I dial > 100 peers and I want to connect to only 5, checking swarm's internal connection count _sometimes_ I see a number > 5, but it's immediately decremented to 5, because `PeerManager` reports the peer and disconnects

Voxelot:
Generally if we have an opportunity to use atomics instead of locks, that would be preferable. 

>  tokio has one as well, but that would turn a lot of non-async code into async one instead.

Tokio docs specifically say that you should only use their async locks as a last resort, I don't think have any good reason to use them.

> sometimes the value read from allow_peer() returns true since there are a lot of nodes reading the value and the writer does not get the chance to immediately update it to false

Using a mutex or atomic instead of rwlock might make this more fair, but ultimately not solvable without a more advanced solution. The issue is that there is some delay between checking the state during authentication vs updating it afer a peer as connected. One way to work around could be to increase the connection count atomically during authentication to prevent too many auth requests, and then if the connection fails or gets dropped then lower the count again. 

However, this race condition is only a small temporary issue that exists right when the node hits the peer connection limit. Although this slight edge could be a DOS vector to stall connections to reserved nodes if a malicious actor intentionally kept dropping peer connections and reconnecting right at the limit, causing slots to be temporarily consumed by the malicious nodes.

crates/services/p2p/src/config/connection_tracker.rs
```diff
@@ -0,0 +1,50 @@
...
+        RwLock,
+    },
+};
+
+/// A `ConnectionTracker` allows either Reserved Peers or other peers if there is an available slot.
+/// It is synced with `PeerManager` which keeps track of the `ConnectionState`.
+#[derive(Debug, Clone)]
+pub(crate) struct ConnectionTracker {
+    reserved_nodes: HashSet<PeerId>,
+    connection_state: Arc<RwLock<ConnectionState>>,
```
xgreenx, line 23: Why do we want to track it on the transport level instead of behavior?

Maybe it would be simpler to handle it inside of `on_connection_handler_event` of `PeerManagerBehaviour`? In this case, we don't need any locks

xgreenx: I see that we do it because of https://github.com/FuelLabs/fuel-core/pull/876#discussion_r1065400628. But does it really consume a slot between our check in the `FuelNodeAuthentication` and when we can process it here?
image

Voxelot: In reserved nodes only mode, there's no good reason to allow connections from any non-reserved peers. Consider it as a DOS mitigation vector, because otherwise a sybil network could target validator nodes with lots of connections and interfere with the connection limit and potentially interfere with consensus / liveness.

For regular nodes that maintain connections with both reserved and unreserved peer ids, deferring the peer disconnection to an async event could still allow for DOS issues. For instance a sybil attacker could saturate the connections of a sentry node and occupy connections slots that should be used for the reserved node, causing it to desync.

By tracking this at the transport level, we minimize the number of resources an attacker can consume. Even in the normal mode of operation, waiting for connections to be established and then asynchronously dropping them later could cause scenarios like what I described above.

xgreenx: Yes, I understand that. My question was, what exactly will we save by this change=)

I found an answer. We will skip [this](https://github.com/libp2p/rust-libp2p/blob/5782a96af2241f068ee0cb4a99503db5e2941035/swarm/src/connection/pool.rs#L622) step, For us it saves:
- We don't increase connection counters. So it fixes the problem described by you. We can always consider having unlimited connections, but it is a bad idea based on the points below=)
- We don't spawn a new `new_for_established_connection` task(the major saving) with loop inside to process data.
- We don't create a new `ConnectionEstablished` event - minus one additional iteration in the swarm.

So yeah, it makes sense.

Hmm, why the `substrate` doesn't do that=) They process peers in the behavior and allow `10000` connections.

Voxelot: Maybe a good question for tomaka :)

Voxelot: If I had to guess, I could see how setting a very high connection threshold and then maintaining a smaller set of peers in network behavior may help with things like peer discovery if the bootstrap nodes are saturated. In that case it would only make sense to have transport auth for nodes that are configured as reserved-only. I would have to take a look to see if allowing more connections than peers would be enough to allow for peer discovery though.

Voxelot (approved):
Im good with moving forward with this as-is. We can optimize the connections handling more in the future.

Issue #883, opened 2023-01-05
[Importer] Block import service impl


Voxelot:
The importer service will be responsible for committing locally produced blocks or executing and committing synced blocks to state. Setup a [RunnableService](https://github.com/FuelLabs/fuel-core/blob/master/crates/services/src/service.rs#L38) in fuel-core-importer that wraps the executor and executes blocks.

## Ports
```rust
trait Executor {
    /// validate and commit the provided block
    fn execute_and_commit(&self, sealed_block: SealedBlock) -> Result<()>;
    /// immediately commit the execution result of a previously validated block
    fn commit_produced_block(&self, uncommitted_block: UncommittedResult) -> Result<()>;
}
```

PR #903, opened 2023-01-13
Implementation of the block importer

xgreenx:
- Closes: https://github.com/FuelLabs/fuel-core/issues/883

What is done:

- Added the main logic of the block importer.
- Added `debug_assert_eq` to verify that the block header id is an identifier of the block. And that it uses the last hash of the application header.

No unit tests right now.


crates/services/consensus_module/src/block_verifier.rs
```diff
@@ -0,0 +1,185 @@
...
+            "The genesis `da_height` is not as expected"
+        );
+        ensure!(
+            expected_genesis_height == actual_genesis_height,
+            "The genesis height is not as expected"
+        );
+        Ok(())
+    }
+
+    fn verify_poa_block_fields(&self, block: &Block) -> anyhow::Result<()> {
```
Voxelot, line 95: todo: this kind of logic should be owned by the respective CM impl (e.g. done in PoA).

crates/fuel-core/src/database/storage.rs
```diff
@@ -0,0 +1,96 @@
...
+};
+use fuel_core_types::blockchain::primitives::BlockHeight;
+use serde::{
+    de::DeserializeOwned,
+    Serialize,
+};
+use std::borrow::Cow;
+
+/// The trait shows that the database table has a corresponding Merkle root table.
+trait HasMerkleRoots: Mappable {
```
freesig, line 21: What are these private traits for? I don't understand how they are related to the block importer? Was this left here back accident?

crates/services/importer/src/importer.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// the first is done. It uses the `tokio::sync::Mutex`, so the parent task will wake up
+    /// when it is ready.
+    pub async fn commit_result<EDatabase>(
+        &self,
+        result: UncommittedResult<StorageTransaction<EDatabase>>,
+    ) -> anyhow::Result<()>
+    where
+        EDatabase: ExecutorDatabase,
+    {
+        let _guard = self.guard.lock().await;
```
freesig, line 88: This is only protecting the database from being committed to. If we allow blocks to execute in any order then try to force them in order on commit then the data they are executing on will be stale. Also this creates a race condition where imported headers error because they both pass validation but then try to commit at the same height

crates/services/importer/src/importer.rs
```diff
@@ -0,0 +1,241 @@
...
+};
+use std::sync::Arc;
+use tokio::sync::broadcast;
+
+pub struct Importer<D, E, V> {
+    database: D,
+    executor: E,
+    verifier: V,
+    broadcast: broadcast::Sender<Arc<ImportResult>>,
+    guard: tokio::sync::Mutex<()>,
```
freesig, line 45: There's no need to lock the execution. The call to `execute_and_commit` should have a semaphore with a single permit so that it can only be called once.

crates/services/importer/src/importer.rs
```diff
@@ -0,0 +1,241 @@
...
+    pub async fn execute_and_commit(
+        &self,
+        sealed_block: SealedBlock,
+    ) -> anyhow::Result<()> {
+        let consensus = sealed_block.consensus;
+        let block = sealed_block.entity;
+        let sealed_block_id = block.id();
+
+        let result_of_verification =
+            self.verifier.verify_block_fields(&consensus, &block).await;
```
freesig, line 182: This call should not be async.

crates/services/importer/src/importer.rs
```diff
@@ -0,0 +1,241 @@
...
+        let (
+            ExecutionResult {
+                block,
+                skipped_transactions,
+                tx_status,
+            },
+            db_tx,
+        ) = self
+            .executor
+            .execute_without_commit(ExecutionBlock::Validation(block))?
```
freesig, line 197: This function is running synchronously within an async context which we don't want.

crates/services/importer/src/importer.rs
```diff
@@ -0,0 +1,241 @@
...
+    IDatabase: ImporterDatabase,
+    E: Executor,
+    V: BlockVerifier,
+{
+    /// The method validates the `Block` fields, executes the `Block`, and commits it with data from
+    /// `SealedBlock`. More about the commit process in [`Importer::commit_result`].
+    ///
+    /// It validates only the `Block` execution rules and the `Block` fields' validity.
+    /// The validity of the `SealedBlock` and seal information is not the concern of this function.
+    pub async fn execute_and_commit(
```
freesig, line 173: This function should not be async as all it needs to do is execute (Which can only happen on the single execution thread), validate (which is just cpu work) and database writing which is sync.

crates/services/importer/src/importer.rs
```diff
@@ -0,0 +1,241 @@
...
+    ///
+    /// It doesn't do any checks regarding block validity(execution, fields, signing, etc.).
+    /// It only checks the validity of the database.
+    ///
+    /// After the commit into the database notifies about a new imported block.
+    ///
+    /// Only one commit may be in progress at the time. All other calls will be suspended until
+    /// the first is done. It uses the `tokio::sync::Mutex`, so the parent task will wake up
+    /// when it is ready.
+    pub async fn commit_result<EDatabase>(
```
freesig, line 81: This function should also be sync

freesig (changes requested):
This is a very large PR, it's a little hard to tell what is related to the block importer. Are there other changes that got pulled in by accident?
I think we need to rework the async / locking within the Importer as well. Maybe we can pair on this tomorrow?

crates/fuel-core/src/database/storage.rs
xgreenx: Yep, I forgot to remove it, thank you

crates/services/importer/src/importer.rs
xgreenx: If you try to execute new blocks in any order, you will face an error that some data is unavailable. By locking here, I tried to highlight that we are really worrying about the "commit" part.
So it should not be possible, but you are right about stale resources, let's lock on `execute_and_commit` too.

crates/services/importer/src/importer.rs
xgreenx: Yep, thank you, forgot to update after I realize that we don't need a relayer to verify fields.

crates/services/importer/src/importer.rs
xgreenx: I thought to add [spawn_blocking](https://docs.rs/tokio/0.2.25/tokio/task/fn.spawn_blocking.html) later here, but let it would the responsibility of the caller.

crates/services/importer/src/importer.rs
xgreenx: I planned to have `Mutex`/`Semaphore` from `tokio`(async lock) and [spawn_blocking](https://docs.rs/tokio/0.2.25/tokio/task/fn.spawn_blocking.html) for executing/committing(to not handle that on the caller's side + in the future executor will be async for parallel execution). But yeah, let's minimize async for as possible and add it later.

Because right now, either `PoA` or `fuel-core-sync` will import blocks, we shouldn't have cases where we really need to await lock. So, for now, we will have this lock=) 

image

xgreenx:
I extracted the `BlockVerifier` logic into https://github.com/FuelLabs/fuel-core/pull/916.
And the final connection of all parts together with CM in the https://github.com/FuelLabs/fuel-core/pull/915.

crates/services/importer/src/importer.rs
```diff
@@ -0,0 +1,339 @@
...
+        let root = ephemeral_merkle_root(
+            vec![*block.header().prev_root(), block_id.into()].iter(),
+        );
+        db_after_execution
+            .insert_block_header_merkle_root(&expected_next_height, &root)?
+            .should_be_unique(&expected_next_height)?;
+
+        db_tx.commit()?;
+
+        let _ = self.broadcast.send(Arc::new(result));
```
freesig, line 232: Who is waiting on the other end of this channel? If this fails to send is the result lost or can it be retrieved from the db?

freesig (dismissed):
Awesome, Nice work on the tests

crates/services/importer/src/importer.rs
xgreenx: There is `TxPool` right now, but after maybe it would be one more - GraphQL service.

The `SealedBlock` is available from the database; transaction status is available in the current code, but we will remove it in the future because it's not part of the blockchain. So if the receiver is slow, it could lose some data. It is why I set it for now as `1 << 10`, but maybe we need to increase it in the future or make sure that receivers are not slow=)

crates/types/src/services/executor.rs
```diff
@@ -310,8 +262,6 @@ pub enum Error {
     Backtrace(Box<Backtrace>),
     #[error("Transaction doesn't match expected result: {transaction_id:#x}")]
     InvalidTransactionOutcome { transaction_id: Bytes32 },
-    #[error("Transaction root is invalid")]
-    InvalidTransactionRoot,
```
Voxelot, line 314: Are we removing this since the headers are guaranteed to be correct by construction?

crates/services/importer/src/config.rs
```diff
@@ -1,4 +1,14 @@
...
+#[derive(Debug, Clone)]
 pub struct Config {
+    pub max_block_notify_buffer: usize,
     pub metrics: bool,
 }
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            max_block_notify_buffer: 1 << 10,
```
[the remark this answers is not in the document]
Voxelot: I think the only slow importers would likely be graphql subscriptions if we decide to expose this channel there. In which case if the are more than 1k blocks behind it's probably a stale connection and dropping the messages is fine.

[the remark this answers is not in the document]
Voxelot: Also as long as the consumers only really care about being notified about latest blocks instead of expecting this to be a persistent and guaranteed delivery message system, we could probably even drop this lower.

crates/services/importer/src/importer.rs
```diff
@@ -0,0 +1,342 @@
...
+
+impl<D, E, V> Importer<D, E, V> {
+    pub fn new(config: Config, database: D, executor: E, verifier: V) -> Self {
+        let (broadcast, _) = broadcast::channel(config.max_block_notify_buffer);
+        Self {
+            database,
+            executor,
+            verifier,
+            broadcast,
+            guard: tokio::sync::Semaphore::new(1),
```
Voxelot, line 108: Any reason to use semaphore instead of tokio mutex? It's basically the same thing but a little bit simpler https://docs.rs/tokio/latest/src/tokio/sync/mutex.rs.html#131

crates/services/importer/src/importer.rs
```diff
@@ -0,0 +1,342 @@
...
+                    "The semaphore was acquired before. It is a problem \
+                    because the current architecture doesn't expect that."
+                );
+                Err(Error::SemaphoreError(err))
+            }
+        }
+    }
+}
+
+impl<IDatabase, E, V> Importer<IDatabase, E, V>
```
Voxelot, line 131: `IDatabase` smells like C# 😂

tests/tests/blocks.rs
```diff
@@ -41,23 +29,15 @@ use std::ops::Deref;
...
     // setup server & client
-    let srv = FuelService::from_database(db, Config::local_node())
+    let srv = FuelService::from_database(Database::default(), Config::local_node())
         .await
         .unwrap();
     let client = FuelClient::from(srv.bound_address);
 
     // run test
-    let id_bytes: Bytes32 = id.into();
+    let block = client.block_by_height(0).await.unwrap().unwrap();
```
Voxelot, line 39: isn't this now basically the same as the next test "`get_genesis_block`"?

crates/types/src/services/executor.rs
xgreenx: Yes, because we construct the block during validation, we will hash all fields and compare the `BlockId` of all fields.

crates/services/importer/src/importer.rs
xgreenx: Initially, it was `Mutex`, but Tommi suggested using `Semaphore`, and I agreed because there are two reasons:
- `Mutex` uses `Semaphore` under the hood
- With `Mutex`, we need to specify the type, and it looks like `Mutex<()>`, when with `Semaphore` we don't need that. Plus, `Mutex` is about access to resources, while `Semaphore` is about signaling mechanisms. Here we use it to signal that some process is already in progress.

tests/tests/blocks.rs
xgreenx: Partially, it checks the `block` endpoint, which can receive the block while the `get_genesis_block` checks that we can get the genesis, but you are right, they are almost the same. Updated them to show the difference more.

Voxelot (approved):

PR #909, opened 2023-01-16
Prepare GraphQL For Future Extraction

ControlCplusControlV:
I've added ports for both `TxPool`, `Executor`,  and `BlockProducer`. Also doing some final cleanup, so that the GraphQL API can be extracted from `fuel-core` soon in a future PR, but goal is to fix most issues in the crate in this PR to keep diff clean in a future migration (after beta-3).

WIP


crates/fuel-core/src/graphql_api/ports.rs
```diff
@@ -130,3 +149,38 @@ pub trait DatabaseChain {
 
     fn base_chain_height(&self) -> StorageResult<DaBlockHeight>;
 }
+
+pub trait FindTx {
```
xgreenx, line 153: You don't need to split each method into its trait. One trait with methods will be enough.

crates/fuel-core/src/graphql_api/ports.rs
```diff
@@ -130,3 +149,38 @@ pub trait DatabaseChain {
...
+    fn find_one(&self, id: TxId) -> Option<TxInfo>;
+}
+
+pub trait InsertTx {
+    fn insert(&self, txs: Vec<Arc<Transaction>>) -> Vec<anyhow::Result<InsertionResult>>;
+}
+
+pub trait TxSubscription {
+    fn tx_update_subscribe(&self) -> tokio::sync::broadcast::Receiver<TxUpdate>;
```
xgreenx, line 162: Instead of `Receiver` please use `fuel_core_services::stream::BoxStream`=)

crates/fuel-core/src/graphql_api/ports.rs
xgreenx: The same goes for other traits, we don't need to split them=) We did that for Database because we built the new functionality on top of it.

crates/fuel-core/src/graphql_api/ports.rs
```diff
@@ -130,3 +149,38 @@ pub trait DatabaseChain {
...
+        &self,
+        transaction: Transaction,
+        height: Option<BlockHeight>,
+        utxo_validation: Option<bool>,
+    ) -> Result<Vec<Receipt>>;
+}
+
+pub trait BlockProducerPort: Send + Sync + DryRunExecution {}
+
+pub trait ExecuteWithoutCommit {
```
xgreenx, line 179: `Database` should be generic of the trait

crates/fuel-core/src/graphql_api/service.rs
```diff
@@ -56,26 +60,24 @@ use std::{
         TcpListener,
     },
     pin::Pin,
-    sync::Arc,
 };
 use tokio_stream::StreamExt;
 use tower_http::{
     set_header::SetResponseHeaderLayer,
     trace::TraceLayer,
 };
```
xgreenx, line 69: Could you connect imports please?=)

crates/fuel-core/src/graphql_api/ports.rs
ControlCplusControlV: Changed

crates/fuel-core/src/graphql_api/ports.rs
ControlCplusControlV: I added it as a generic, but I'm not sure what else you wanted this to entail, so you could please let me know if I didn't everything related you wanted?

crates/fuel-core/src/graphql_api/service.rs
ControlCplusControlV: Imports are as connected as they can be right now, `cargo fmt` should catch most of them too because we are on nightly

ControlCplusControlV:
Awaiting #915 and #916 which will cause breaking changes that this will need to rebase from

ControlCplusControlV:
Blocked by #916 and Blocked by #915

crates/fuel-core/src/graphql_api/ports.rs
freesig: Yeh I agree and ideally these should be named to indicate which adapter they will match to. So this should probably just be the `TxPoolPort`

crates/fuel-core/src/service/adapters/graphql_api.rs
```diff
@@ -175,3 +204,62 @@ impl DatabaseChain for Database {
...
+    fn insert(&self, txs: Vec<Arc<Transaction>>) -> Vec<anyhow::Result<InsertionResult>> {
+        self.shared.insert(txs)
+    }
+}
+
+impl TxSubscription for TxPoolService {
+    fn tx_update_subscribe(
+        &self,
+    ) -> fuel_core_services::stream::BoxStream<
+        Result<fuel_core_txpool::service::TxUpdate, BroadcastStreamRecvError>,
```
freesig, line 218: One thing to keep in mind here is that you can get [lagged errors](https://docs.rs/tokio/latest/tokio/sync/broadcast/error/enum.RecvError.html#variant.Lagged) which you probably don't want to pass on to the consumer. You could filter them out and convert to an anyhow error

freesig (changes requested):

crates/fuel-core/src/graphql_api/ports.rs
ControlCplusControlV: Ok, will merge all of the traits into one adapter then

crates/fuel-core/src/service/adapters/graphql_api.rs
ControlCplusControlV: This returns `BroadcastStreamRecvError`which only contains lagged out errors, so would you prefer I filter out errors entirely and just return a TxUpdate?

crates/fuel-core/src/graphql_api/service.rs
```diff
@@ -69,14 +72,13 @@ use tower_http::{
 pub type Service = fuel_core_services::ServiceRunner<NotInitializedTask>;
 
 pub type Database = Box<dyn DatabasePort>;
+// TODO: When the port for `Executor` will exist we need to replace it with `Box<dyn ExecutorPort>
+pub type Executor = Box<dyn ExecutorPort<Database>>;
```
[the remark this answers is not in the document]
ControlCplusControlV: Removed

crates/fuel-core/src/service/adapters/graphql_api.rs
freesig: Yeh I think it's best to filter out lagged errors as there's nothing the caller can really do about them

ControlCplusControlV: Actually there is, looking through the upstream changes this would force [transaction_status_change](https://github.com/FuelLabs/fuel-core/blob/controlc/gql_extraction/crates/fuel-core/src/query/subscriptions.rs#L29) to accept a `TxUpdate` Non-Result (as `BroadcastStreamRecvError` only has `LaggedOut`), and `transaction_status_change` will check the [database if the status lagged out](https://github.com/FuelLabs/fuel-core/blob/controlc/gql_extraction/crates/fuel-core/src/query/subscriptions.rs#L89) to make sure the update wasn't missed

Voxelot (approved):

crates/fuel-core/src/service/adapters/graphql_api.rs
freesig: Ah nice catch

freesig (approved):
Nice work

PR #910, opened 2023-01-17
restore original db in vm for benchmarking

Voxelot:
Fixes an issue where each benchmark run was nesting the interpreter storage transaction from a previous run into a new transaction.

### DIFF
--- a/benches/benches/vm.rs
+++ b/benches/benches/vm.rs
@@ -27,12 +27,15 @@ where
                 diff,
             } = &mut i;
             let code = OpcodeRepr::from_u8(instruction.op());
+            let original_db = vm.as_mut().database_mut().clone();
             let mut db_txn = {
                 let db = vm.as_mut().database_mut();
                 let db_txn = db.transaction();
+                // update vm database in-place to use transaction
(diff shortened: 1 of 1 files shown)


benches/benches/vm.rs
```diff
@@ -27,12 +27,15 @@ where
                 diff,
             } = &mut i;
             let code = OpcodeRepr::from_u8(instruction.op());
+            let original_db = vm.as_mut().database_mut().clone();
```
xgreenx, line 30: Hmm, why do we want to work with database transaction here at all?

Voxelot: To more accurately replicate the overhead the executor would actually have. Since executor will never use rocksdb directly without a view layer.

xgreenx: But we can use `DatabaseTransaction` in the tests by default, then=)

Voxelot: The issue is that we are trying to reuse the same vm instance across multiple benchmarks by using state diffs to undo the previous test run. This means we have to update the db on the vm instead of making a new vm each time.

freesig (approved):

PR #911, opened 2023-01-17
heartbeat protocol

### LINKED ISSUE #879: [P2P] Block height inclusion in ping protocol
To avoid busy polling of block heights from peers, include the current block height as part of the ping protocol. The block height of each connected peer should be available in the peer info table.

Add an API that returns either a broadcast or stream of all block heights received by ping.

leviathanbeak:
Closes #879 

This PR introduces `heartbeat` protocol, which replaces existing `ping` protocol.

On a high level, P2PService now exposes 2 APIs:
- `update_block_height(BlockHeight)` - this method informs p2p service of the latest locally known block height (more on it soon)
- `subscribe_block_height() -> broadcast::Receiver<BlockHeightHeartbeatData>` - caller of this method subscribes to getting latest block heights from connected peers.

The approach here was following:
- on each new Peer connection, we start a Heartbeat (`struct HeartbeatHandler`), beating to some configured interval, on each beat it first requests from p2p Service the latest known block height and then it shares that height with the connected peer.
- p2p service needs to be up to date with latest known block height so that it can share the info with Hearbeat

small todos (shouldn't stop you from reviewing):
- expose `HeartbeatConfig` for configuration, currently it's hardcoded
- (maybe) document the flow of `HeartbeatHandler` and its `OutboundState`


crates/services/p2p/src/peer_manager.rs
```diff
@@ -497,13 +500,13 @@ impl NetworkBehaviour for PeerManagerBehaviour {
 pub struct PeerInfo {
     pub peer_addresses: HashSet<Multiaddr>,
     pub client_version: Option<String>,
-    pub latest_ping: Option<Duration>,
```
Voxelot, line 500: would it be useful to track both block height and duration since last heartbeat?

crates/services/p2p/src/service.rs
```diff
@@ -260,6 +278,13 @@ impl SharedState {
         Ok(())
     }
 
+    pub fn update_block_height(&self, block_height: BlockHeight) -> anyhow::Result<()> {
```
Voxelot, line 281: We'll need to update the p2p adapter during the integration with importer to have it update the block height. Alternately, we could just add a db port for latest block height in p2p and fetch it on-demand instead of notifying the p2p service of block height changes.

Voxelot (dismissed):

crates/services/p2p/src/service.rs
xgreenx: Hmm, previously, we discussed that the block importer notifies all listeners, and p2p is one of the listeners. It means that we need to pass a `BoxStream`=)

With latest height we will notify neighbors as soon as possible

crates/services/p2p/src/heartbeat/handler.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    fn connection_keep_alive(&self) -> KeepAlive {
+        // Heartbeat protocol wants to keep the connection alive
+        KeepAlive::Yes
+    }
+
+    fn on_behaviour_event(&mut self, event: Self::InEvent) {
+        let HeartbeatInEvent::LatestBlock(block_height) = event;
+
+        match self.outbound.take() {
```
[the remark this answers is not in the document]
Voxelot: @leviathanbeak did the previous ping protocol do anything to mitigate ping spam?

Voxelot:
> With latest height we will notify neighbors as soon as possible

I was thinking we could just pull from the DB, but this approach is nice since it reduces the amount of I/O required for heartbeats.

crates/types/src/services/p2p.rs
```diff
@@ -69,3 +70,11 @@ impl<T: Debug + Send + 'static> NetworkData<T> for GossipData<T> {
         self.data.take()
     }
 }
+/// Used for relying latest `BlockHeight` info from connected peers
+#[derive(Debug, Clone)]
+pub struct BlockHeightHeartbeatData {
+    /// PeerId as bytes
+    pub peer_id: Vec<u8>,
```
freesig, line 77: We can't send the peer ids in the heartbeat because imagine the scenario where the sync look has commit header `4` then get's the heartbeat for `8`. What peer id will be associated with `5, 6, 7`?

crates/services/p2p/src/service.rs
freesig: Do you mean the fuel sync not the importer? This should be a stream that is pushed to by p2p and is pulled from the sync

crates/services/p2p/src/service.rs
```diff
@@ -260,6 +278,13 @@ impl SharedState {
...
+        self.request_sender
+            .try_send(TaskRequest::UpdateWithBlockHeight(block_height))?;
```
freesig, line 283: No reason to return this error if it will just be ignored

crates/services/p2p/src/service.rs
```diff
@@ -186,6 +194,14 @@ where
             }
             p2p_event = self.p2p_service.next_event() => {
                 match p2p_event {
+                    Some(FuelP2PEvent::PeerInfoUpdated { peer_id, block_height }) => {
+                        let block_height_data = BlockHeightHeartbeatData {
+                            peer_id: peer_id.into(),
+                            block_height,
+                        };
+
+                        let _ = self.shared.block_height_broadcast.send(block_height_data);
```
freesig, line 203: This is an example of where a height and therefor peers can be missed. Just send height only.

crates/services/p2p/src/heartbeat.rs
```diff
@@ -0,0 +1,126 @@
...
+                        connection_id,
+                        in_event: HeartbeatInEvent::LatestBlock(
+                            self.current_block_height,
+                        ),
+                    })
+            }
+        }
+    }
+
+    fn poll(
```
freesig, line 115: Doesn't this lead to the poll function hot looping because the context is never told to sleep?

freesig (changes requested):
Looks good, need to update the interface though to not send peer data. Also we need functionality to request sealed header and request transaction from the network

crates/services/p2p/src/peer_manager.rs
leviathanbeak: yeah, especially if we gonna protect it against ddos

crates/services/p2p/src/service.rs
leviathanbeak: yeah you're right @xgreenx this was just an initial iteration, I'll refactor it to BoxStream :)

crates/services/p2p/src/heartbeat/handler.rs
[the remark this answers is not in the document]
leviathanbeak: @Voxelot no it did not,

@xgreenx I will add timed buffer on receive heartbeats and if it overflows we will disconnect the peer

leviathanbeak:
> Also we need functionality to request sealed header and request transaction from the network
@freesig there's another issue for that, it's not part of the hearbeat

crates/services/p2p/src/service.rs
xgreenx: > Do you mean the fuel sync not the importer? This should be a stream that is pushed to by p2p and is pulled from the sync

I meant block importer=) This function is used to notify P2P about new heights, not the `fuel-core-sync`. For `fuel-core-sync` we have `subscribe_block_height` below=)

crates/services/p2p/src/heartbeat.rs
leviathanbeak: it doesn't, the context is passed optionally, this `poll()` is just a part of the main `poll()` for all behaviours, so if this poll returns Pending it won't be immediately called again, the context will be passed to other behaviours to do their `poll`ing

crates/services/p2p/Cargo.toml
```diff
@@ -58,6 +58,7 @@ rand = "0.8"
 serde = { version = "1.0", features = ["derive"] }
 sha2 = "0.10"
 tokio = { version = "1.21", features = ["sync"] }
+tokio-stream = "0.1"
```
freesig, line 61: You can use `futures::stream::StreamExt` instead of adding a new dependency

freesig (changes requested):

crates/services/p2p/src/peer_manager.rs
leviathanbeak: done

crates/services/p2p/Cargo.toml
leviathanbeak: good one, thanks

crates/services/p2p/src/service.rs
xgreenx: I didn't get you=) The code sends the height and only **one** peer id corresponding to the received heights.

crates/services/p2p/src/heartbeat/handler.rs
```diff
@@ -0,0 +1,330 @@
...
+                    }
+
+                    break
+                }
+                Some(OutboundState::SendingBlockHeight(mut outbound_block_height)) => {
+                    match outbound_block_height.poll_unpin(cx) {
+                        Poll::Pending => {
+                            if self.timer.poll_unpin(cx).is_ready() {
+                                // Time for successfull send expired!
+                                self.failure_count += 1;
```
Voxelot, line 204: Do we need to reset the timer here? Otherwise it seems like this could get polled 5x in a row immediately once the timer expires, which is effectively the same as having a failure count of 1.

crates/services/p2p/src/heartbeat/handler.rs
```diff
@@ -0,0 +1,330 @@
...
+                        Poll::Ready(Ok(stream)) => {
+                            // reset failure count
+                            self.failure_count = 0;
+                            // start new idle timeout until next request & send
+                            self.timer.reset(self.config.idle_timeout);
+                            self.outbound = Some(OutboundState::Idle(stream));
+                        }
+                        Poll::Ready(Err(_)) => {
+                            self.failure_count += 1;
+                            debug!(target: "fuel-libp2p", "Sending Heartbeat failed, this is {} time it failed with this connection", self.failure_count);
```
Voxelot, line 222: Consider using [ordinal](https://docs.rs/ordinal/latest/ordinal/) spelling or something else like:

```suggestion
                            debug!(target: "fuel-libp2p", "Sending Heartbeat failed, failure {}/{} for this connection", self.failure_count,  self.config.max_failures);
```

crates/services/p2p/src/service.rs
Voxelot: I think this is leaking implementation details of how fuel-sync currently works into the p2p design. From p2p perspective it just reports which data it sees from which peers, it's up to other services to decide if the peer info is useful or not.

Voxelot (dismissed):
minor feedback on the failure counting timer, but seems good overall.

crates/services/p2p/src/heartbeat/handler.rs
leviathanbeak: We don't have to, since we have _taken_ the `Option<OutboundState>`, let's say it was
`Some(OutboundState::SendingBlockHeight)`
we check for the `timer` it expired -> we increment the failure count,
next `poll` or `loop` iteration if previous was expired, now the `Option<Outbound>` will be `None`
so it won't hit the timer check again.

The timer is restarted once we move to `OutboundState::SendingBlockHeight` again, [here](https://github.com/FuelLabs/fuel-core/pull/911/files#diff-81c845fee50fec4b1c2cab23105e502b561fec4db6dab63ea158f4d89679c2edR136).

freesig (approved):
All looks good. A few suggestions.
One thing that's a little confusing is I thought we were pigging backing onto the existing libp2p heartbeat but this seems to be adding a new event or am I misunderstanding something? @Voxelot

freesig (approved):

Voxelot (approved):

leviathanbeak:
> One thing that's a little confusing is I thought we were pigging backing onto the existing libp2p heartbeat but this seems to be adding a new event or am I misunderstanding something?

We removed Ping protocol in favor of Heartbeat, it does the similar thing just with our own data instead of the random bytes sent over the wire. Ping Protocol does not support custom data.

PR #912, opened 2023-01-17
Hotfix v0.15.3

Voxelot:
* do not merge *


Dentosal (commented):
Do we really have to patch-bumb every unmodified crate as well?

Voxelot:
> Do we really have to patch-bumb every unmodified crate as well?

In theory no, but our publishing verification won't allow any crates in the workspace to be on separate versions since it would be easy to forget a bump somewhere and only discover the issue after the release has already been tagged.

PR #913, opened 2023-01-18
[WIP] feat: Persistent BMT data for blocks

bvrooman:


xgreenx:
Hmm, it seems my [PR](https://github.com/FuelLabs/fuel-core/pull/915) overlaps with yours partially. I'm also touching on this part because it is part of the block verification. I guess we will land a block importer before this PR, so maybe it makes sense to rebase on mine to avoid double work and reuse some types.

BTW, I guess you need to update the `fuel-merkle` to use the latest `fuel-storage`, but in this case, you also will depend on https://github.com/FuelLabs/fuel-core/pull/806

Voxelot:
Looks like we'll need to do a big merge cleanup to address all the shuffling around that occurred during the block sync feature.

cc @bvrooman

bvrooman:
Sounds good. I will start on that.

crates/fuel-core/src/database/block.rs
```diff
@@ -198,3 +308,71 @@ impl Database {
...
+        Table: Mappable<Key = u64, SetValue = Primitive, GetValue = Primitive>,
+        Self: StorageInspect<Table, Error = StorageError>,
+    {
+        let storage = self.borrow();
+        let tree = MerkleTree::load(storage, version).unwrap();
+        Ok(tree)
+    }
+}
+
+const FUEL_BLOCK_MERKLE_METADATA_KEY: &str = "FuelBlocks";
```
Voxelot, line 378: might be good to move this const to the top of the file

Voxelot:
closing in favor of #935

Issue #887, opened 2023-01-05
Importer/CM integration


Voxelot:
- [ ] Implement importer related ports from the PoA service for the importer adapter in fuel-core.
- [ ] Provide the importer adapter to the PoA service during initialization in FuelService.

Voxelot:
depends on: #883, #884

Issue #884, opened 2023-01-05
[CM] defer commitment of executed blocks to the importer


Voxelot:
Rather than committing locally produced blocks directly to the DB, the consensus module should defer commitment of block state to the importer.

## Ports
```rust
trait Importer {
    async fn import_block(&self, result: UncommittedResult) -> Result<()>;
}

```

Voxelot:
fixed by: #903

PR #915, opened 2023-01-18
Connect the block importer with outside world

xgreenx:
- Closes https://github.com/FuelLabs/fuel-core/issues/887
- Closes https://github.com/FuelLabs/fuel-core/issues/884

What is done:

- Removed `CHAIN_HEIGHT_KEY`, because now we have a genesis block to track this information.
- Unified all database API that works with `BlockHeight` to work with references. Previously in some places, it was the reference, in some not.
- Added `manual_produce_block` into `PoA` and used it in the `GraphQL` to have the one source of the block production. There are no unit tests(help is appreciated).
- Connected `PoA` and `BlockImporter` via ports. Added `BlockImporterAdapter`.
- Added the `FuelBlockRoots` column with the corresponding table. It is used to track BMT MMR roots for fuel blocks. The implementation adds new `DatabaseColumn` and `IntoDatabaseKey` traits to automate the storage trait implementation for some tables.
- Moved implementation of database ports into the corresponding file in the `fuel_core::service::adapter`.
- Removed the `Database` generic from `PoA` at all. Now we must pass the `last_height` while creating the `Task`. Updated tests to work with it.
- Block producer now accepts the `block_time` along with `block_height`.


crates/fuel-core/src/database/block.rs
```diff
@@ -114,9 +110,9 @@ impl Database {
         Ok(block.header().time().to_owned())
     }
 
-    pub fn get_block_id(&self, height: BlockHeight) -> StorageResult<Option<BlockId>> {
-        Database::get(self, &height.to_bytes()[..], Column::FuelBlockIds)
-            .map_err(Into::into)
+    pub fn get_block_id(&self, height: &BlockHeight) -> StorageResult<Option<BlockId>> {
+        let key = height.database_key();
+        Database::get(self, key.as_ref(), Column::FuelBlockIds).map_err(Into::into)
```
ControlCplusControlV, line 115:
```suggestion
        Database::get(self, height.database_key();.as_ref(), Column::FuelBlockIds).map_err(Into::into)
```
Is a bit more compact but just as readable imo

crates/fuel-core/src/database/block.rs
```diff
@@ -85,15 +86,10 @@ impl StorageMutate<FuelBlocks> for Database {
...
-        let id = match id {
-            Some((id, _)) => Some(id),
-            None => self.get_starting_chain_height()?,
-        };
-
-        Ok(id)
+    pub fn latest_height(&self) -> StorageResult<BlockHeight> {
+        self.ids_of_latest_block()?
+            .map(|(height, _)| height)
+            .ok_or(not_found!("BlockHeight"))
```
ControlCplusControlV, line 92: I'm not too familiar, but what is the main reason for a string literal of "BlockHeight," rather than passing the `BlockHeight` type specifically?

.github/workflows/ci.yml
```diff
@@ -95,6 +95,8 @@ jobs:
         include:
           - command: clippy
             args: --all-targets --all-features
+          - command: check
+            args: --all-targets
```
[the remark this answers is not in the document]
freesig: Yep it catches running `cargo check --all-targets` at the workspace root which nothing else does. It has a different feature set the `all-features` or running the command at each crate level like the make command does.

crates/fuel-core/src/database.rs
```diff
@@ -99,6 +69,7 @@ mod state;
 pub mod balances;
 pub mod metadata;
 // TODO: Rename in a separate PR into `transaction`
```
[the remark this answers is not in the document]
freesig: What do you mean?

crates/fuel-core/src/database/block.rs
freesig: Maybe you meant `height.database_key().as_ref()`?

crates/fuel-core/src/database/block.rs
freesig: Yeh also should this be an option not an error?

crates/chain-config/src/config/state.rs
```diff
@@ -57,5 +57,5 @@ pub trait ChainConfigDb {
     /// Returns *all* unspent message configs available in the database.
     fn get_message_config(&self) -> StorageResult<Option<Vec<MessageConfig>>>;
     /// Returns the last available block height.
-    fn get_block_height(&self) -> StorageResult<Option<BlockHeight>>;
+    fn get_block_height(&self) -> StorageResult<BlockHeight>;
```
freesig, line 60: Why is the option removed? If the chain is empty won't this be none?

crates/fuel-core/src/database/storage.rs
```diff
@@ -0,0 +1,90 @@
...
+use std::borrow::Cow;
+
+impl DatabaseColumn for FuelBlockRoots {
+    fn column() -> Column {
+        Column::BlockHeaderMerkle
+    }
+}
+
+/// The table has a corresponding column in the database.
+trait DatabaseColumn {
```
freesig, line 27: @Voxelot Is this the pattern you were talking about in last weeks sync?

crates/fuel-core/src/database/storage.rs
```diff
@@ -0,0 +1,90 @@
...
+    fn remove(&mut self, key: &T::Key) -> StorageResult<Option<T::GetValue>> {
+        let key = key.database_key();
+        Database::remove(self, key.as_ref(), T::column()).map_err(Into::into)
+    }
+}
+
+// TODO: Implement this trait for all keys and use `type Type = MultiKey` for tuples.
+//  -> After replace all common implementation with blanket, if possible.
+/// Some keys requires pre-processing that could change their type.
+pub trait IntoDatabaseKey {
```
freesig, line 76: This should be `ToDatabaseKey` because it doesn't consume self.

crates/fuel-core/src/service/adapters/producer.rs
```diff
@@ -85,3 +106,23 @@ impl fuel_core_producer::ports::Relayer for MaybeRelayerAdapter {
         }
     }
 }
+
+impl fuel_core_producer::ports::BlockProducerDatabase for Database {
+    fn get_block(&self, height: &BlockHeight) -> StorageResult<Cow<CompressedBlock>> {
+        let id = self.get_block_id(height)?.ok_or(not_found!("BlockId"))?;
+        self.storage::<FuelBlocks>()
+            .get(&id)?
+            .ok_or(not_found!(FuelBlocks))
```
freesig, line 115: This should be an option not an error

crates/fuel-core/src/service/adapters/producer.rs
```diff
@@ -85,3 +106,23 @@ impl fuel_core_producer::ports::Relayer for MaybeRelayerAdapter {
...
+    }
+
+    fn block_header_merkle_root(&self, height: &BlockHeight) -> StorageResult<Bytes32> {
+        self.storage::<FuelBlockRoots>()
+            .get(height)?
+            .ok_or(not_found!(FuelBlocks))
```
freesig, line 121: Also should be an option not an error

freesig (changes requested):
Just a few issues around errors instead of options

crates/fuel-core/src/service/genesis.rs
```diff
@@ -57,30 +58,34 @@ use fuel_core_types::{
...
+    services::block_importer::{
+        ImportResult,
+        UncommittedResult as UncommittedImportResult,
+    },
 };
 use itertools::Itertools;
 
 /// Loads state from the chain config into database
-pub(crate) fn initialize_state(
+pub(crate) fn maybe_initialize_state(
```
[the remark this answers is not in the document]
xgreenx: Returning the result means nothing=) The `Err` only shows that we got an error during initialization, while `Ok` maybe be "We changed something" or "We changed nothing"

crates/fuel-core/src/service/genesis.rs
```diff
@@ -102,6 +107,7 @@ fn add_genesis_block(config: &Config, database: &mut Database) -> anyhow::Result
     let block = Block::new(
         PartialBlockHeader {
             application: ApplicationHeader::<Empty> {
+                // TODO: Set `da_height` based on the chain config.
```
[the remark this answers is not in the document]
xgreenx: The setting of the `da_height` here also requires supporting it in the `ChainConfig` and in the relayer("To sync blocks from events > than `da_height`"). It is not the scope of the current change=)

crates/fuel-core/src/database/block.rs
xgreenx: > I'm not too familiar, but what is the main reason for a string literal of "BlockHeight," rather than passing the BlockHeight type specifically?

Because we don't have support for not `Mappable` types=) 
image

> Yeh also should this be an option not an error?

`NonFound` error is an `Option`, I will describe the reason below=)

crates/chain-config/src/config/state.rs
xgreenx: All chains have a genesis block now, so it should not be `None` + we have `NotFound` to represent `None` case.

crates/fuel-core/src/graphql_api/service.rs
```diff
@@ -67,11 +69,10 @@ use tower_http::{
 pub type Service = fuel_core_services::ServiceRunner<NotInitializedTask>;
 
 pub type Database = Box<dyn DatabasePort>;
-// TODO: When the port for `Executor` will exist we need to replace it with `Box<dyn ExecutorPort>
-pub type Executor = crate::service::adapters::ExecutorAdapter;
+pub type ConsensusModule = Box<dyn ConsensusModulePort>;
 // TODO: When the port of BlockProducer will exist we need to replace it with
 //  `Box<dyn BlockProducerPort>
-pub type BlockProducer = Arc<fuel_core_producer::Producer<crate::database::Database>>;
+pub type BlockProducer = crate::service::adapters::BlockProducerAdapter;
```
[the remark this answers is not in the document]
xgreenx: To use `Box` here requires the definition of the port. I don't want to add a new code and mix the work that is already handled by a mentioned PR=)

crates/fuel-core/src/service/adapters/producer.rs
xgreenx: image

We don't process the case of the `Option`. We replace it with the same error on the caller's side. There is not sense to return an `Option` if we don't plan to process it in a different way.

crates/fuel-core/src/service/adapters/producer.rs
xgreenx: The same as for `get_block`, we don't process an option on the caller's side

.github/workflows/ci.yml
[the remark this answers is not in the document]
xgreenx: Actually, I guess we can remove it, because `cargo test --all-features --workspace` do the same)

crates/fuel-core/src/service/adapters/producer.rs
freesig: I'm sort of ok with this but is does feel like an odd pattern. I guess it's not so bad because the trait is only used by a single crate.

freesig (dismissed):
Ok looks good

crates/fuel-core/src/database/block.rs
ControlCplusControlV: I did

crates/fuel-core/src/database.rs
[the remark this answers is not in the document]
ControlCplusControlV: Comment should be above the `mod transactional`, not really important but wanted to mention

crates/fuel-core/src/service/genesis.rs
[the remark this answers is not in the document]
ControlCplusControlV: Oh I see, so it only `maybe` will have an impact

crates/fuel-core/src/graphql_api/service.rs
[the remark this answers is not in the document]
ControlCplusControlV: Ok, will refactor that PR off of this one when this gets merged then

.github/workflows/ci.yml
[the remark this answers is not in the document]
ControlCplusControlV: Would prefer to remove it then since CI times are already pretty long

ControlCplusControlV:
Bumping the CI check issue, would prefer to remove that before merging so CI times don't get too long if `--workspace` is already covering that case

freesig (approved):

PR #916, opened 2023-01-18
External block verifier

xgreenx:
Refs: https://github.com/FuelLabs/fuel-core/issues/883

It is the implementation of the block verifier. Right now it provides only one function to verify the block's fields. But in the future, we can add verification of the seals(it is why it holds a relayer) and reuse this verifier in the `fuel-core-sync`.

Along with the verifier, I changed the `fuel_core_poa::Config` and `ChainConfig` to have rules for the block's time of the network and separate rules for block production.

- Added `NodeRole::Producer` and `NodeRole::Validator`. Based on the role, we will create a `fuel_core_poa::Config`.
- Removed `Never` from `BlockProduction` from `ChainConfig` because it is the block production information of the `PoA`.

No unit tests right now.


crates/services/consensus_module/poa/src/verifier.rs
```diff
@@ -0,0 +1,140 @@
...
+                    );
+                }
+                PoABlockProduction::Hybrid {
+                    min_block_time,
+                    max_block_time,
+                    ..
+                } => {
+                    let previous_block_time = Tai64N::from(prev_header.time());
+                    // The block should be in the range
+                    // [min..max + (max - min)/2]
```
freesig, line 105: I don't understand why this is true?

freesig (approved):
Looks good, just one question around the time bounds.

crates/services/consensus_module/poa/src/verifier.rs
xgreenx: Because it is not true=) Created an issue to discuss it https://github.com/FuelLabs/fuel-core/issues/918

I already disabled this check in the production for now.

Dentosal (approved):
LGTM; Helpful comments as well

crates/fuel-core/src/service/config.rs
```diff
@@ -63,13 +77,48 @@ impl Config {
...
+            block_importer: Default::default(),
             #[cfg(feature = "relayer")]
             relayer: Default::default(),
             #[cfg(feature = "p2p")]
             p2p: P2PConfig::<NotInitialized>::default("test_network"),
             consensus_key: Some(Secret::new(default_consensus_dev_key().into())),
         }
     }
+
+    pub fn poa_config(&self) -> anyhow::Result<fuel_core_poa::Config> {
```
Voxelot, line 89: Would using `impl From<Config> for fuel_core_poa::Config` instead be more idiomatic rust?

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -75,6 +75,9 @@ pub struct Task<D, T, B> {
     /// a bit, but doesn't cause any other issues.
     pub(crate) last_block_created: Instant,
     pub(crate) trigger: Trigger,
+    // TODO: Consider that the creation of the block takes some time, and maybe we need to
```
Voxelot, line 78: 👍🏻

crates/services/consensus_module/poa/src/verifier.rs
```diff
@@ -0,0 +1,144 @@
...
+mod tests;
+
+/// The config of the block verifier.
+pub struct Config {
+    /// If the manual block is enabled, skip verification of some fields.
+    pub enabled_manual_blocks: bool,
+    /// This configuration exists because right now consensus module uses the `Tai64::now`
+    /// during the generation of the block. That means if the node(block producer) was
+    /// offline for a while, it would not use the expected block time after
+    /// restarting(it should generate blocks with old timestamps).
```
Voxelot, line 25: > (it should generate blocks with old timestamps)

Don't follow why we'd want to forge stale timestamps for blocks. Using older timestamps may make the verification logic easier to implement, but the time should always aim to be as current/correct as possible.

crates/services/consensus_module/poa/src/verifier.rs
```diff
@@ -0,0 +1,144 @@
...
+                        "The `time` of the next block can't be lower"
+                    );
+                }
+                PoABlockProduction::Interval {
+                    block_time: average_block_time,
+                } => {
+                    let previous_block_time = Tai64N::from(prev_header.time());
+
+                    // If the `average_block_time` is 15 seconds, the block should be in the range
+                    // [15..15 + 15/2] -> [15..23]
```
Voxelot, line 86: I think we should only have a min bound, there's no reason to be strict about the max time between blocks.

Voxelot:
@xgreenx and I were discussing the issue of whether or not we should be validating things like min or max block times in poa. 

Ultimately his conclusion was that if something is part of the chainconfig, it should be part of the consensus validation in some way. My conclusion was that we shouldn't add all these verification around timestamps for poa as it could make the system brittle and cause syncing to stall.

So we decided that poa trigger configs should be local parameters that only apply to the authority node, and not included in the chain config. Eventually as our cli args grow in complexity, we could extend clap to support loading from a config file instead which will make it easier to manage.

Green already did the work of stripping the trigger params out of chainconfig, tomorrow I will improve the cli interface a bit using clap argument groups.

crates/chain-config/Cargo.toml
```diff
@@ -14,13 +14,13 @@ description = "Fuel Chain config types"
...
 bech32 = "0.9.0"
 bincode = "1.3"
-fuel-core-poa = { path = "../services/consensus_module/poa", version = "0.15.1" }
 fuel-core-storage = { path = "../storage", version = "0.15.1" }
 fuel-core-types = { path = "../types", version = "0.15.1", features = [
     "serde",
     "random",
 ] }
 hex = { version = "0.4", features = ["serde"] }
+humantime-serde = "1.1.1"
```
ControlCplusControlV, line 23: Is this actually used anywhere? I removed the dep and ran `ci_checks.sh` locally and everything seemed to pass for me

xgreenx: Nope, it is not used, I don't know why we didn't get a warning=D

xgreenx: Oh, I know=D 
image

xgreenx:
Approved!=)

crates/services/consensus_module/poa/src/verifier.rs
```diff
@@ -0,0 +1,78 @@
...
+        height != 0u32.into(),
+        "The PoA block can't have the zero height"
+    );
+
+    let prev_height = height - 1u32.into();
+    let prev_root = database.block_header_merkle_root(&prev_height)?;
+    let header = block.header();
+    ensure!(
+        header.prev_root() == &prev_root,
+        "The genesis time should be unix epoch time"
```
Voxelot, line 47: Unsure how this check relates to epoch time 🤔

Voxelot, line 47:
```suggestion
        "Previous root of the next block should match the previous block root"
```

Voxelot: seems like a copy-paste bug from genesis verification, applied a suggestion to fix it.

Voxelot (approved):

Issue #877, opened 2023-01-05
[P2P] GetSealedBlockHeader


Voxelot:
Add P2P functionality to request a sealed block header from any peer that has sufficient block height. It should consult peer info to randomly select a peer that has reached a sufficient block height. It should also return the associated peer id that provided the header.

Should aim to satisfy the following trait example:

```rust
trait P2P {
    async fn get_sealed_block_header(&self, height: BlockHeight) -> Result<(Sealed<Header>, PeerId)>;
}
```

Voxelot:
depends on #879

PR #919, opened 2023-01-19
request response for get transactions & sealed header from p2p network

### LINKED ISSUE #878: [P2P] GetBlockTransactions
Add P2P functionality to request all transactions associated with a block from a specified peer.

Should aim to satisfy the following trait example:

```rust
trait P2P {
    async fn get_block_transactions(&self, block_id: BlockId, peer_id: PeerId) -> Result<Vec<Transaction>>;
}
```

leviathanbeak:
closes #878 #877 

The work has been done on 2 (now merged) branches, by @freesig and myself.


crates/services/p2p/src/service.rs
```diff
@@ -246,6 +263,23 @@ where
                                     );
                                 });
                             }
+                            RequestMessage::Transactions(block_id) => {
+                                let db = self.db.clone();
+                                let request_sender = self.shared.request_sender.clone();
+
+                                tokio::spawn(async move {
```
freesig, line 270: Just noting here and above are spawns that are not ever joined. This is equivalent to an unbounded channel. There's nothing stopping someone spamming (or even legitimately requesting data) until the machine runs out of memory. I really think we want to maintain backpressure in all our production code.

Voxelot: Let's avoid these spawns for now if possible to avoid backpressure issues. Improving concurrency can come later.

leviathanbeak: this has been handled in @freesig's PR to this branch, but we need a long term better solution

leviathanbeak:
Note: the PR is not _that big_, it contains small refactors and better naming apart from the needed changes

crates/services/p2p/src/service.rs
```diff
@@ -299,11 +375,55 @@ impl SharedState {
...
+            .map_err(|e| anyhow!("{}", e))
+    }
+
+    pub async fn get_transactions_from_peer(
+        &self,
+        peer_id: Vec<u8>,
+        block_id: BlockId,
+    ) -> anyhow::Result<Option<Vec<Transaction>>> {
+        let (sender, receiver) = oneshot::channel();
+        let from_peer = PeerId::from_bytes(&peer_id).expect("Valid PeeeId");
```
ControlCplusControlV, line 419:
```suggestion
        let from_peer = PeerId::from_bytes(&peer_id).expect("Valid PeerId");
```

freesig (changes requested):

crates/services/p2p/src/p2p_service.rs
```diff
@@ -1275,7 +1287,7 @@ mod tests {
             tokio::select! {
                 message_sent = rx_test_end.recv() => {
                     // we received a signal to end the test
-                    assert_eq!(message_sent, Some(true), "Received wrong block height!");
+                    assert_eq!(message_sent, Some(true), "Receuved incorrect or missing missing messsage");
```
ControlCplusControlV, line 1290: [nit]
```suggestion
                    assert_eq!(message_sent, Some(true), "Received incorrect or missing missing messsage");
```

Assertion seems to be failing here

crates/fuel-core/src/service/adapters/p2p.rs
```diff
@@ -19,6 +26,20 @@ impl P2pDb for Database {
...
+    ) -> StorageResult<Option<SealedBlockHeader>> {
+        self.get_sealed_header(height).await
+    }
+
+    async fn get_transactions(
+        &self,
+        block_id: BlockId,
+    ) -> StorageResult<Option<Vec<Transaction>>> {
+        self.get_transactions(block_id).await
+    }
```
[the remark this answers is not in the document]
leviathanbeak: yeah lol, I was thinking of referencing `Database` methods and used `self` just like `fn get_sealed_block()`, we need to rename those

crates/services/p2p/src/p2p_service.rs
leviathanbeak: there is something wrong with `Request::Transactions` test others are passing, if no one solves it by tomorrow I will try it.

It seems the request doesn't even go through, the node gets disconnected every time I re-ran it with Transactions, other requests pass which is (slightly) weird

crates/fuel-core/src/service/adapters/p2p.rs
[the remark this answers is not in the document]
freesig: We can just fully qualified calls. Having a look at it now

freesig (approved):

PR #920, opened 2023-01-20
Freesig/get sealed header

freesig:
fixes: https://github.com/FuelLabs/fuel-core/issues/877


leviathanbeak:
love the changes, there's only one thing different than other two existing request responses, my two responses only return if the result exists so Result<T>, yours is Result<Option<T>>, but the way the `send_response_msg` is set up currently it would never return None, since

```rust
if message.is_none() {
            self.inbound_requests_table.remove(&request_id);
            return Ok(())
        }
```

so the requestee would get an error, either Ok(T) or Err(e) the request_id gets cleaned at
```rust
self.outbound_requests_table.remove(&request_id),
```
but I have just noticed on Err(e) we do not inform the oneshot `channel` about the error.

So we need to do 2 things:
- unify our API and logic, do we send `None` responses over the network or not
- on `Error` inform the channel item that it failed (I will do this now and push to this branch)

crates/services/p2p/src/service.rs
```diff
@@ -243,53 +263,71 @@ where
...
             latest_block_height = self.next_block_height.next() => {
                 if let Some(latest_block_height) = latest_block_height {
                     let _ = self.p2p_service.update_block_height(latest_block_height);
                 }
+                None
             }
+        };
+
+        if let Some(f) = to_send {
+            let msg = f.await;
```
xgreenx, line 329: It means that we will scan the database during the current iteration. The idea of the `spawn` was to do it asynchronously because it could take some time=)

I agree that the previous solution was not good(because we didn't control spawned tasks). But it allowed us to process each iteration fast(all other operations in this `run` method are very fast).

Because it is testnet, I'm okay if we do them synchronously than spawning new tasks. But in the future, maybe we need to consider another solution.

Regarding the current solution, all methods to the database are synchronous, so we don't need `async` closures and `Option<BoxedFuture>`. We can interact with the database directly from the method.

leviathanbeak: ah I missed this, that the tasks are not spawned anymore, I think we should definitely spawn them with some mechanism that will throttle new requests and spawns if there are a lot of them, plus once we implement some ddos measures across whole p2p network like limiting number of requests from a single peer in a short span of time etc will help in managing these tasks

Voxelot: We should probably use a semaphore restricted `spawn_blocking` pool similar to dry_run for offloading these kinds of workloads. Doing it synchronously is probably safer for now so that we don't end up with an unbounded queue of requests.

freesig: Spawning tasks has an overhead and is not cheap. We can do things to make this run concurrently by buffering futures and we can even spawn a limited number of tasks to churn through the concurrent futures in parallel but it wouldn't make any sense to spawn more then the number of threads on the machine because otherwise you are just wasting overhead.

I feel like the main point here is being missed. It's not so much about where we run the futures but more about applying back pressure to the run function when we are filling up. We don't want to pull in more new messages from the network if we are already at capacity.

The problem needs to be broken down into parts:
1. Concurrency: This means multiple futures being polled in the same task. (Like a select or a stream of futures with buffered). This has the lowest overhead for context switching.
2. Parallelism: Basically throwing more threads at the problem. Yes tasks are cheaper then os threads but a cpu core can only poll one task at any given time. Each cpu is **concurrent** across a set of tasks (and the tasks can be moved to cpus under less pressue). But there's nothing to gain by spawning heaps of tasks (especially not a task per network request). You just loose time to the overhead. In previous projects I have seen this exact situation grind networking to a crawl. I think honestly the best approach to spawning is to not do it and start with concurrency. Then benchmark and incrementally add it. You will be surprised how it doesn't actually make things faster. Spawn overhead is measure in milliseconds where as your db read might complete in micro or even nano seconds.
3. Backpressure: IMHO this is the most important part to get right around networking. If we pull in more work then the system can currently process then you are saturating the system and performance gets much much worse then a queue that is less then 100% utilized (in fact you usually want to aim for much less then 100% utilization). Everywhere in our system we **must** have bounded queues and the part that feeds the queue **must** slow down if the part that is process the queue is not keeping up. This is an incredibly hard problem to test because it requires actually running at scale / full load tests. Also the problems show up in extremely weird long tail latency type bugs. We will save our selves so much future pain if we design our system with proper back pressure.

Sorry for the rant but this is the type of thing I have seen ruin a lot of projects (because it usually doesn't show up until production and is super hard to debug). I also think it's a lot easier to do from the start of a project then trying to retroactively add it in later. It isn't something we should push off just because we are only building for a test net.

Voxelot: > I feel like the main point here is being missed. It's not so much about where we run the futures but more about applying back pressure to the run function when we are filling up

Is using a semaphore like I suggested not an acceptable way to apply backpressure? I proposed that we use a similar solution to dry_run which limits the number of permits to the cpu core count and we discussed at length as being a good approach in the past.

I agree that spawning extra tasks because there's already too many requests is not a great idea. But I don't think your point isn't being conveyed, as Elvis was also mentioning mechanisms like:

> plus once we implement some ddos measures across whole p2p network like limiting number of requests from a single peer in a short span of time etc will help

freesig: Yep a semaphore and spawn_blocking also works. I wasn't saying we couldn't solve this issue, more trying to give an argument for this becoming one of the criteria we have for accepting pull requests. This just seemed like a good example to base it on.

freesig: Also worth noting that limiting number of connections per peer doesn't help with the number of requests a peer can give or the number of peers you can have. It's usually simpler to limit the max resource usage of the actual pipeline then rely on limiting the public api usage. Although some combination of both is probably going to work best.

leviathanbeak: > Also worth noting that limiting number of connections per peer doesn't help with the number of requests a peer can give or the number of peers you can have.

I meant limiting number of requests per peer, not (only) connections.

But yeah as @Voxelot we're on the same page, ie we're gonna use the best possible solution, by "spawning tasks" comment I was thinking something along `FuturesUnordered` approach, had some ideas...but if there's a better,  more efficient idea, no one's gonna go against it

freesig:
> love the changes, there's only one thing different than other two existing request responses, my two responses only return if the result exists so `Result<T>` , yours is `Result<Option<T>>`, but the way the `send_response_msg` is set up currently it would never return None, since
> 
> ```rust
> if message.is_none() {
>             self.inbound_requests_table.remove(&request_id);
>             return Ok(())
>         }
> ```
> 
> so the requestee would either get `Ok(t)` or `Err(e)` and the `request_id` gets cleaned at
> 
> ```rust
> self.outbound_requests_table.remove(&request_id),
> ```
> 
> So we need to unify our API and logic, do we send `None` responses over the network or not

If you don't send a `None` won't the requester need to wait for a timeout?

leviathanbeak:
> If you don't send a `None` won't the requester need to wait for a timeout?
yes, that's what I realized when writing the comment, still wrote and waited for your feedback first, 
sure thing I will update it on my side, I will also merge this with my branch and then do the work there

PR #921, opened 2023-01-20
Return back the "Output Bech32Address for initial accounts" change.

xgreenx:
[This change](https://github.com/FuelLabs/fuel-core/pull/594) was done at the same time as moving this file into another place, so during conflict resolution, we missed it.

Closes https://github.com/FuelLabs/fuel-core/issues/593


kayagokalp (approved):
Thanks! LGTM

ControlCplusControlV (approved):
Tested locally with `cargo run -- run --db-type in-memory` bech32 is outputted as expected, LGTM

Voxelot (approved):

PR #924, opened 2023-01-23
Integrate fuel-core-sync with importer and p2p

### LINKED ISSUE #885: Sync/P2P Integration
- [ ] Implement P2P related ports from fuel-core-sync for the P2P adapter in fuel-core
- [ ] Provide the P2P adapter to the sync service during initialization in `FuelService` startup

### LINKED ISSUE #886: Sync/Importer Integration
- [ ]  Implement (fuel-sync) Importer related ports for the Block Import adapter in fuel-core.
- [ ]  Provide the import adapter to the sync service during initialization in FuelService startup.

### LINKED ISSUE #901: Sync/CM Integration
Implement (fuel-sync) Consensus Module related ports for the ConsensusModule adapter in fuel-core.

### LINKED ISSUE #902: Add sealed block header verification api to PoA Coordinator
Implement an API in the PoACoordinator module to verify a sealed block header during block sync.

- check authority signature

freesig:
Adds the p2p adapter for fuel-sync 

closes #885
closes #886 
closes #901 
closes #902


crates/fuel-core/src/service/adapters/sync.rs
```diff
@@ -0,0 +1,70 @@
...
+            BlockHeight,
+            BlockId,
+        },
+        SealedBlockHeader,
+    },
+    fuel_tx::Transaction,
+    services::p2p::SourcePeer,
+};
+
+#[cfg(not(feature = "p2p"))]
```
ControlCplusControlV, line 16: Is there a reason to explicitly implement this when `p2p` is disabled rather than just using something like `automock`?

freesig: automock is a testing only framework. It should never be used in production code

ControlCplusControlV: I had assumed this impl would mainly be for testing, but would it actually see use? Or rather does there need to be an implementation of `PeerToPeerPort` if `p2p` is disabled, or is this being chosen over attaching feature flags to every function which would call this port?

freesig: Basically you can't use sync without p2p however there's nothing in fuel sync that depends on p2p so it would be a waste to put it behind a feature flag. Maybe this isn't needed though as we could make the actual sync service optional and only set it when the p2p flag is set

freesig: Yeh just looked at this and it would require propagating the feature flag to a lot more places. Maybe it's cleaner as is?

ControlCplusControlV: Making sync optional depending on p2p is preferable for me since it would also reduce compile times for forc testing if fuel-sync doesn't need to be compiled to run local testnets (and forms code clarity perspective) , but I could see that being a separate issue and pr if it's too much. Do you want to do it in this PR, I can approve if not?

freesig: I don't think it should be an optional dependency. I meant option type like how the maybe relayer works

Voxelot:
Will this PR also cover signature verification during sync? i.e. https://github.com/FuelLabs/fuel-core/issues/902

freesig:
> Will this PR also cover signature verification during sync? i.e. #902

Yep just added it

crates/services/sync/src/import/tests.rs
```diff
@@ -592,6 +592,9 @@ impl DefaultMocks for MockBlockImporterPort {
 pub(crate) fn empty_header(h: BlockHeight) -> SourcePeer<SealedBlockHeader> {
     let mut header = BlockHeader::default();
     header.consensus.height = h;
+    let mut transaction_tree =
+        fuel_core_types::fuel_merkle::binary::in_memory::MerkleTree::new();
+    header.application.generated.transactions_root = transaction_tree.root().into();
```
Voxelot, line 597: cc @bvrooman for some reason the root returned here doesn't match the root in the data-driven test suite

tests/tests/sync.rs
```diff
@@ -0,0 +1,111 @@
+use std::sync::Arc;
+
```
Voxelot, line 2:
```suggestion
```

Voxelot (approved):

tests/tests/sync.rs
Voxelot: not worth re-running the CI over

crates/services/sync/src/import/tests.rs
Voxelot: What we actually get: `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`

What is in the test specs for bmt with zero leaves: `ce04537ec4b4b84cf02fa5c50c7810e2f6592e3b00de7423d7c3e7e7d8689a67`

bvrooman: Discussed on Slack: `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` is actually the correct root in this case - it is the sha256 hash of the empty string.

PR #925, opened 2023-01-23
Support Configuring Clap via ENV vars

Voxelot:
Updates to make devops easier

- Support configuration via env vars instead of cli args
- allow libp2p keypair to have a secret passed in directly without a file (i.e. env var)
- upgrade all CLI's to clap 0.4.1


bin/fuel-core/src/cli/run.rs
```diff
@@ -60,52 +60,59 @@ mod relayer;
 /// Run the Fuel client node locally.
 #[derive(Debug, Clone, Parser)]
 pub struct Command {
-    #[clap(long = "ip", default_value = "127.0.0.1", value_parser)]
+    #[clap(long = "ip", default_value = "127.0.0.1", value_parser, env)]
```
ControlCplusControlV, line 63: Could you specify `env = PORT`, and use all capitals as convention for environment variables?

ControlCplusControlV (commented):
LGTM to be overall, just some nits on naming

bin/fuel-core/src/cli/run.rs
Voxelot: this is IP not port, and the default naming for envs is SCREAMING_SNAKE_CASE:

> env [= <str>]: [Arg::env](https://docs.rs/clap/latest/clap/builder/struct.Arg.html#method.env) (needs [env feature](https://docs.rs/clap/latest/clap/_features/index.html) enabled)
When not present: no env set
Without <str>: defaults to the case-converted field name
https://docs.rs/clap/latest/clap/_derive/index.html#arg-attributes

> rename_all_env = <string_literal>: Override default field name case conversion for env variables for [Arg::env](https://docs.rs/clap/latest/clap/builder/struct.Arg.html#method.env)
When not present: "SCREAMING_SNAKE_CASE"
Available values: "camelCase", "kebab-case", "PascalCase", "SCREAMING_SNAKE_CASE", "snake_case", "lower", "UPPER", "verbatim"
https://docs.rs/clap/latest/clap/_derive/index.html#command-attributes


You can also confirm the casing of the variable names by running `cargo run --bin fuel-core -- run --help`:
```
 fuel-core git:(Voxelot/clap-env) cargo run --bin fuel-core -- run --help
    Finished dev [unoptimized + debuginfo] target(s) in 0.23s
     Running `target/debug/fuel-core run --help`
Run the Fuel client node locally

Usage: fuel-core run [OPTIONS]

Options:
      --ip <IP>
          [env: IP=]
          [default: 127.0.0.1]

      --port <PORT>
          [env: PORT=]
          [default: 4000]

      --db-path <DB_PATH>
          [env: DB_PATH=]
          [default: /Users/voxelot/.fuel/db]

      --db-type <DATABASE_TYPE>
          [env: DATABASE_TYPE=]
          [default: rocks-db]
          [possible values: in-memory, rocks-db]
```

Voxelot:
Note this will help unblock the work that @uberscott is doing on the operator side for sentry nodes

bin/fuel-core/src/cli/run.rs
ControlCplusControlV: Oh this also explains why I was also having trouble testing configuring via env locally

ControlCplusControlV (approved):
LGTM, testing locally and could configure things like port via env variables

PR #926, opened 2023-01-24
Use postcard instead of bincode for serialization

### LINKED ISSUE #776: Investigate alternatives to bincode serialization
Maybe instead of `serde` and `bincode`, we could use something like [`rkyv`](https://docs.rs/rkyv/latest/rkyv/)? In any case, `bincode` format isn't stable across Rust versions and should not be used for persistent storage.

### LINKED ISSUE #905: Replace p2p bincode codec with a more stable serialization format
P2P currently uses bincode for messaging due to simplicity, however it is not a safe format to use for networked services as the format could change due to simply upgrading rust.

Protobuf is likely the most standard/straightforward option, but we can consider alternatives such as cbor/msgpack as well.

Dentosal:
[Postcard](https://github.com/jamesmunns/postcard) is a drop-in replacement for bincode. It has [almost identical performance characteristics, but it's bit a more comact](https://github.com/djkoloski/rust_serialization_benchmark) and [stable across Rust versions](https://github.com/jamesmunns/postcard#format-stability). It recently reached 1.0 and is considered rather stable.

This PR simply replaces all invocations to bincode with their postcard equivalents. Closes #905 and closes #776

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -425,6 +425,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic-polyfill"
+version = "0.1.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e3ff7eb3f316534d83a8a2c3d1674ace8a5a71198eba31e2e2b597833f699b28"
+dependencies = [
+ "critical-section",
+]
+
+[[package]]
 name = "atomic-waker"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -985,6 +994,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "cobs"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "67ba02a97a2bd10f4b59b25c7973101c79642302776489e030cd13cdab09ed15"
+
+[[package]]
 name = "coins-bip32"
 version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1204,6 +1219,12 @@ dependencies = [
  "cast",
  "itertools",
 ]
+
+[[package]]
+name = "critical-section"
+version = "1.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6548a0ad5d2549e111e1f6a11a6c2e2d00ce6a3dafe22948d67c2b443f775e52"
 
 [[package]]
 name = "crossbeam-channel"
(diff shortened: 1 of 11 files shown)


ControlCplusControlV:
Is there a reason you chose Postcard? I know we discussed `rkyv` a while back for performance reasons, and because its been battle tested with Solana

Dentosal:
The reason for use `postcard` is that it works as a drop-in replacement and fullfills our needs. While `rkyv` is a bit faster than `postcard`, it sometimes produces significantly larger output. I'm going to investigate `rkyv` later this week, but this PR provides a simple way to fix all known issues about our current code, i.e. data format non-stability across Rust versions.

leviathanbeak (changes requested):
In p2p, `postcard` codec is directly implemented within `bincode` codec, it should be a separate implementation (I don't mind renaming existing file) but all references to `bincode` should be removed

freesig:
> In p2p, `postcard` codec is directly implemented within `bincode` codec, it should be a separate implementation (I don't mind renaming existing file) but all references to `bincode` should be removed/replaced

Yeh I wonder how this will affect things like https://github.com/FuelLabs/fuel-core/blob/4c945a91f2472d0cd44b82f5afb0a738e3bb9522/crates/services/p2p/src/codecs/bincode.rs#L289

freesig:
For example https://github.com/FuelLabs/fuel-core/blob/4c945a91f2472d0cd44b82f5afb0a738e3bb9522/crates/services/p2p/src/request_response/messages.rs#L28
is probably no longer correct?

leviathanbeak:
> For example
> 
> https://github.com/FuelLabs/fuel-core/blob/4c945a91f2472d0cd44b82f5afb0a738e3bb9522/crates/services/p2p/src/request_response/messages.rs#L28
> 
> 
> is probably no longer correct?

@freesig why not? Does `postcard` mess up the size?
As far as I see it, that piece of code should not change irrespective of the serializer, it's just the size of the biggest enum variant.

Dentosal:
> As far as I see it, that piece of code should not change irrespective of the serializer, it's just the size of the biggest enum variant.

Postcard is able to reduce the size, likely by encoding the enum tag as a single-byte varint instead of u32. I've updated the check to use `<` instead of equality.

leviathanbeak:
> Postcard is able to reduce the size, likely by encoding the enum tag as a single-byte varint instead of u32. I've updated the check to use `<` instead of equality.

that's actually good news, maybe still having `<=` cuz even if it's equal it should be fine and the test should convey that imo,
as for the rest of the code I will review it tomorrow morning

Voxelot (dismissed):

Voxelot (approved):

leviathanbeak (approved):
🥇

PR #928, opened 2023-01-24
Fixed some [nit]s after review the PR for `fuel-core-sync` integration

xgreenx:
Because I didn't have a chance to review the PR in time, I did a review after=)

### DIFF
--- a/crates/fuel-core/src/service/adapters.rs
+++ b/crates/fuel-core/src/service/adapters.rs
@@ -8,14 +8,14 @@ use std::sync::Arc;
 
 pub mod block_importer;
 pub mod consensus_module;
+pub mod executor;
 pub mod graphql_api;
 #[cfg(feature = "p2p")]
 pub mod p2p;
 pub mod producer;
-pub mod txpool;
-
 #[cfg(feature = "p2p")]
 pub mod sync;
+pub mod txpool;
 
 #[derive(Clone)]
 pub struct PoAAdapter {
@@ -60,7 +60,6 @@ pub struct BlockProducerAdapter {
 pub struct BlockImporterAdapter {
     pub block_importer:
         Arc<fuel_core_importer::Importer<Database, ExecutorAdapter, VerifierAdapter>>,
-    execution_semaphore: Arc<tokio::sync::Semaphore>,
 }
 
 #[cfg(feature = "p2p")]
--- a/crates/services/consensus_module/poa/src/verifier.rs
+++ b/crates/services/consensus_module/poa/src/verifier.rs
@@ -49,7 +49,7 @@ pub fn verify_consensus(
     }
 }
 
-pub fn verify_poa_block_fields<D: Database>(
+pub fn verify_block_fields<D: Database>(
     config: &Config,
     database: &D,
     block: &Block,
--- a/crates/services/consensus_module/poa/src/verifier.rs
+++ b/crates/services/consensus_module/poa/src/verifier.rs
@@ -4,9 +4,9 @@ use fuel_core_storage::Result as StorageResult;
 use fuel_core_types::{
     blockchain::{
         block::Block,
+        consensus::poa::PoAConsensus,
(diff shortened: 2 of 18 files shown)


crates/types/src/blockchain/header.rs
```diff
@@ -265,10 +267,13 @@ impl PartialBlockHeader {
     }
 }
 
-fn generate_txns_root(transactions: &[Vec<u8>]) -> Bytes32 {
+fn generate_txns_root(transactions: &[Transaction]) -> Bytes32 {
+    // TODO: The `to_bytes` requires mutability(but it is problem of the API).
```
xgreenx, line 271: The serialization should not require `&mut self`. It is a problem of our current approach with the `io::Read` trait for serialization. Later we will rework it, so for now we can clone it here.

https://github.com/FuelLabs/fuel-tx/pull/190 fixes the problem, so maybe we need to consider finishing this PR.

crates/fuel-core/src/service/adapters.rs
```diff
@@ -60,7 +60,6 @@ pub struct BlockProducerAdapter {
 pub struct BlockImporterAdapter {
     pub block_importer:
         Arc<fuel_core_importer::Importer<Database, ExecutorAdapter, VerifierAdapter>>,
-    execution_semaphore: Arc<tokio::sync::Semaphore>,
```
xgreenx, line 63: I removed the semaphore from here for now because `Importer` already returns an error if the import process is in progress. 

We don't plan to import several blocks simultaneously. I think receiving an error is better than hiding it under the semaphore.

crates/services/p2p/src/peer_manager.rs
```diff
@@ -672,11 +672,14 @@ impl PeerManager {
...
         let mut range = rand::thread_rng();
+        // TODO: Optimize the selection of the peer.
+        //  We can store pair `(peer id, height)` for all nodes(reserved and not) in the
+        //  https://docs.rs/sorted-vec/latest/sorted_vec/struct.SortedVec.html
         self.non_reserved_connected_peers
             .iter()
             .chain(self.reserved_connected_peers.iter())
             .filter(|(_, peer_info)| {
-                peer_info.heartbeat_data.block_height == Some(*height)
+                peer_info.heartbeat_data.block_height >= Some(*height)
```
Voxelot, line 682: Good catch

crates/fuel-core/src/service/adapters.rs
Voxelot: Would this be a problem in local testing if manual-block-production is enabled? I.e. maybe they have interval mode set for block time but manually trigger some blocks and there's a conflict.

crates/services/consensus_module/poa/src/verifier.rs
```diff
@@ -31,28 +31,21 @@ pub trait Database {
     fn block_header_merkle_root(&self, height: &BlockHeight) -> StorageResult<Bytes32>;
 }
 
+// TODO: Make this function `async` and await the synchronization with the relayer.
 pub fn verify_consensus(
```
Voxelot, line 35: should we rename this to `verify_poa_consensus`?

Voxelot (dismissed):

crates/fuel-core/src/service/adapters.rs
xgreenx: If they face an error in the tests, it may be better to update the test=D Because if you produce more blocks during manual production than you asked, it may cause wrong test results.

crates/services/consensus_module/poa/src/verifier.rs
xgreenx: It lives inside of `PoA`, so we don't need to add this prefix. We need to rename `verify_poa_block_fields` into `verify_block_fields`

bvrooman (approved):

Voxelot (approved):

Issue #777, opened 2022-11-15
Observability for p2p


freesig:
- [x] Make is easy to run tracing in tests without recompiling. Done in #932 
- [x] Log block id / height when importer gets new block
- [x] Log incoming block id and block height at block sync.
- [ ] Log transaction id at tx pool on submit.
- [ ] Log transaction id at tx pool on receiver.
- [x] Log next_event as this is very useful for debugging

PR #933, opened 2023-01-25
Tracing on fuel sync

freesig:
Mostly finishes #777 
But I can't remember what these meant:

> Log transaction id at tx pool on submit.
Log transaction id at tx pool on receiver.


Voxelot:
> Log transaction id at tx pool on submit.
When a tx is submitted via the graphql api and either successfully enters or gets rejected by the txpool we should log something.

> Log transaction id at tx pool on receiver.
When a tx is received via libp2p gossip and either successfully enters or gets rejected by the txpool we should log something.

Voxelot:
I think we should have more info-level logging around sync progress, as it is fairly typical for most blockchains to always log their sync progress and is usually one of the first things to check when examining the logs of a blockchain node.

crates/services/sync/src/import.rs
```diff
@@ -302,6 +315,72 @@ fn validate_header_height(
...
+where
+    E: BlockImporterPort + Send + Sync + 'static,
+{
+    // Execute and commit the block.
+    let height = *block.entity.header().height();
+    let r = executor.execute_and_commit(block).await;
+
+    // If the block executed successfully, mark it as committed.
+    if r.is_ok() {
+        state.apply(|s| s.commit(*height))
```
Voxelot, line 377: we should info-level log here to indicate sync progress

Voxelot: Although it seems like maybe this should be logged already via `tracing::instrument`? I'm not seeing this fn in the logs when I enable `FUEL_TRACE=1`

Voxelot: was able to get instrument to show up using:

```
            "1" | "true" | "on" => {
                let _ = tracing_subscriber::FmtSubscriber::builder()
                    .with_env_filter(EnvFilter::from_default_env())
                    .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
                    .try_init();
            }
```

However it is _very_ noisy, so I still think more info! events could be useful.

freesig: It's logged in the state.rs but it's debug level. I can bump it up though

crates/services/sync/src/state.rs
```diff
@@ -178,8 +190,14 @@ impl State {
     }
 
     fn apply_status(&mut self, status: Option<Status>) {
-        if let Some(s) = status {
-            self.status = s;
+        match status {
+            Some(s) => {
+                tracing::debug!("Status change from: {:?}, to: {:?}", self.status, s);
```
freesig, line 195: @Voxelot This is actually where the event is logged but it will capture the spans from parent `instrument` functions. Would you like these to be info?

Voxelot: yeah that would be helpful, as well as info-level confirmation of each block imported if the range is < 100 blocks. If we're syncing a large range of thousands of blocks it would be too noisy to log each block individually.

freesig: I mean this does mean the block is confirmed imported because this call only happens post execute_and_commit.

freesig: I don't think we should disable logging programmatically because it's very hard to do and also someone might need to know what when wrong even post 100 blocks. If it's too noisy that's what filtering is for.

Voxelot: Ah true, if you're syncing small ranges <100 this should get logged often enough to show progress. However, even showing the latest block id that was committed from the range would be a useful improvement.

For very large ranges it would still be hard to tell how well your node is even progressing or if it stalled for some reason since this wouldn't get logged very often.

freesig: It does show the block id though. It's just captured in the span above this not the event. So it will print out an event that inner state has changed that is within the span showing for which block.
As for adjusting log levels I'd strongly suggest for debugging the best approach is always going to be capture everything then filter through the logs after. Cutting out information for convenience means that you will potentially miss the cause of an issue. It's also possible to dynamically change the log filter without restarting the server but I think capturing and replaying / filtering is always better.

freesig: Ok with the latest push info level trace gives:
image

freesig: vs debug level 
image

freesig:
> > Log transaction id at tx pool on submit.
> 
> When a tx is submitted via the graphql api and either successfully enters the txpool or gets rejected we should log something.
> 
> > Log transaction id at tx pool on receiver.
> 
> When a tx is received via libp2p gossip and either successfully enters the txpool or gets rejected we should log something.

Ok these should be addressed now

crates/services/sync/src/state.rs
Voxelot: Nice this looks pretty good!

As far as logging in large ranges vs just enabling more debug levels, this is more of a UX issue than a debugging issue. Node operators are accustomed to seeing regular log updates when syncing a node that is far behind the network so they can see how well it's progressing. If nothing is updated in the logs for a long time they may think the sync isn't working even when it is.

freesig: I'm not sure I understand, why would sync be making progress but the logs not update?

Voxelot: let's say the range is 0..=10_000_000, if we only log when the state changes from processing to committed then it might be a long time for logs to come in. Although I guess this is mitigated by the state changing as new block heights are discovered. We can fix this later if it looks like an issue.

freesig: It logs on each block in that range

PR #935, opened 2023-01-25
Alternative implementation: Persistent BMT data for blocks

xgreenx:
It is an alternative implementation of the https://github.com/FuelLabs/fuel-core/pull/913 with minimum changes. The main change is the same, but other changes are minimal.


xgreenx:
@Voxelot I used `master` as the base because it shows the final diff(without overriding changes in the parent PR).
Plus, we don't need to merge this -> parent -> master=)

crates/fuel-core/src/database/storage.rs
```diff
@@ -3,32 +3,104 @@ use crate::database::{
...
+    type Key = u64;
+    type OwnedKey = Self::Key;
+    type Value = binary::Primitive;
+    type OwnedValue = Self::Value;
+}
+
+/// The metadata table for [`FuelBlockMerkleData`] table.
+pub struct FuelBlockMerkleMetadata;
+
+impl Mappable for FuelBlockMerkleMetadata {
```
Voxelot, line 73: Is the idea to keep these mappables here instead of fuel-core-storage because they aren't needed anywhere else other than fuel-core right now?

crates/services/importer/src/importer/test.rs
```diff
@@ -236,43 +220,37 @@ where
 #[test_case(
     genesis(0),
     underlying_db(not_found),
-    executor_db(ok(0), ok(None), ok(None), 1)
+    executor_db(ok(0), ok(None), 1)
```
Voxelot, line 223: test_case's should be documented with human readable descriptions, ie.

`#[test_case(...; "successfully imports genesis block when database is empty")`

Voxelot: created issue #940 to address this as a followup.

Voxelot (approved):

bvrooman (approved):
Reapproving after dismissing @Voxelot's approval

Voxelot (approved):

PR #936, opened 2023-01-25
add message validation for gossipsub messages from txpool

### LINKED ISSUE #595: prevent gossip of invalid txs

leviathanbeak:
Closes #595 

For gossipsub messages, we need to validate **all** the messages.
Messages will not be propagated unless a _timely_ validation is reported.

I have updated the gossipsub tests, now instead of just checking that Node A sends a gossipsub message to Node B,
we also check that Node C gets the propagated message from Node B.
Of course if the validation is positive, if the message is rejected Node C should not receive it.


crates/services/txpool/src/service.rs
```diff
@@ -136,12 +138,29 @@ where
...
+                    let tx = new_transaction.take_data().unwrap();
                     let txs = vec!(Arc::new(tx));
-                    self.shared.txpool.lock().insert(
+
+                    let mut result = self.shared.txpool.lock().insert(
                         &self.shared.tx_status_sender,
                         &txs
                     );
+
+                    if let Some(acceptance) = match result.pop() {
```
xgreenx, line 152: I would prefer to add `expect` here because we expect that `result.len() == 1`, in this case you can use `if else` statement instead of `match`

crates/services/txpool/src/service.rs
```diff
@@ -136,12 +138,29 @@ where
                 should_continue = false;
             }
             new_transaction = self.gossiped_tx_stream.next() => {
-                if let Some(GossipData { data: Some(tx), .. }) = new_transaction {
+                if let Some(GossipData { data: Some(_), .. }) = &new_transaction {
```
xgreenx, line 141: What was the problem with the previous code `let Some(GossipData { data: Some(tx), .. }) = new_transaction`?

Seems it does the same in a strange way=D

crates/services/p2p/src/p2p_service.rs
```diff
@@ -1275,11 +1327,33 @@ mod tests {
                         let broadcast_request = broadcast_request.clone();
                         matches!(node_b.publish_message(broadcast_request), Err(PublishError::Duplicate));
 
-                        break
+                        match acceptance {
+                            GossipsubMessageAcceptance::Reject | GossipsubMessageAcceptance::Ignore => {
+                                break
```
xgreenx, line 1332: Seems we expect panic here

crates/services/p2p/src/p2p_service.rs
```diff
@@ -1241,7 +1288,12 @@ mod tests {
                     tracing::info!("Node A Event: {:?}", node_a_event);
                 },
                 node_b_event = node_b.next_event() => {
-                    if let Some(FuelP2PEvent::GossipsubMessage { topic_hash, message, .. }) = node_b_event.clone() {
+                    if let Some(FuelP2PEvent::GossipsubMessage { topic_hash, message, message_id, peer_id }) = node_b_event.clone() {
+                        // Message Validation must be reported
+                        // If it's `Accept`, Node B will propagate the message to Node C
+                        // If it's `Ignore` or `Reject`, Node C should not receive anything
+                        let msg_acceptance = to_message_acceptance(&acceptance);
+                        let _ = node_b.report_message_validation_result(&message_id, &peer_id, msg_acceptance);
```
xgreenx, line 1296: Hmm, I guess it is better to test it on the `service.rs` layer than on `p2p_service.rs`. In this case, you don't need to simulate the logic of the service and we will have a test to verify that service is gossiping data.

crates/services/txpool/src/service.rs
Voxelot: alternately, maybe just adding a helper fn to convert any kind of `Ok` or `Err` into an `Accept` or `Reject` would be enough to clean up this code without risking any runtime panic.

crates/services/txpool/src/service.rs
Voxelot: It's because we want to use `take_data` to move the tx data out of gossip without cloning. It might be better to just expose the peer id in the interfaces so we don't need to worry about passing the GossipData back and forth between p2p and services receiving messages.

crates/services/p2p/src/p2p_service.rs
Voxelot: Maybe we shouldn't even break here? If we want to test that node c doesn't receive ignored messages then we should leave node b operating long enough for it to potentially send a message it shouldn't have.

The tricky part would be that we'd need some kind of timeout to determine when it's safe to assume node b hasn't sent a bad message and we can end the test 🤔

leviathanbeak: I'm with @Voxelot but yeah deciding when to exit is a bit tricky, can think of something maybe with timeouts, or sending another message...

crates/services/p2p/src/p2p_service.rs
leviathanbeak: I don't get the benefit, these are unit tests for `p2p_service` and we should have as many tests as possible for all kinds of scenarios where there multiple p2p services

crates/services/txpool/src/service.rs
leviathanbeak: yeah it's cuz of ownership, if I didn't check it with a reference it would've been consumed and later unusable for reporting validation, having said that the interface does leave a lot to be desired so we should improve it.

crates/services/p2p/src/p2p_service.rs
leviathanbeak: > Seems we expect panic here

I don't get it, why panic, the test is a success at that point (if we're looking at the same thing)

crates/services/p2p/src/p2p_service.rs
xgreenx: I meant we still need to test that the `service.rs` gossips transactions. Because `p2p_service.rs` is part of the `service.rs`, instead of unit testing it here, maybe it makes more sense to test it in the `service.rs`. In this case, you don't need to simulate the sending of `acceptance`.

Previously this test only checked the `publish_message` method. Now it checks `publish_message` and `report_message_validation_result`. Maybe we can test `report_message_validation_result` in a more simple setup?

crates/services/p2p/src/p2p_service.rs
xgreenx: > I don't get it, why panic, the test is a success at that point (if we're looking at the same thing)

The comment below says "// the `exit` should happen in Node C". So I assume that `break` in the node C should end the test=)

crates/services/txpool/src/service.rs
xgreenx: https://github.com/FuelLabs/fuel-core/pull/942

crates/services/p2p/src/p2p_service.rs
leviathanbeak: yes that is correct, but that's just a comment to describe the flow of the test, it should not panic when it hits the line of the comment, cuz it needs to hit that line, ie the message validity is in `Accept` state and that is fine

xgreenx: But why do we need two `break`?

leviathanbeak: depends on type of test, you can pass Ignore/Reject or Accept case, depending which one happens, the test should either exit in Node B or Node C.

That's just a reusable function  for tests like
```
#[tokio::test]
    #[instrument]
    async fn gossipsub_broadcast_block_with_accept() {
        gossipsub_broadcast(
            GossipsubBroadcastRequest::NewBlock(Arc::new(Block::default())),
            GossipsubMessageAcceptance::Accept,
        )
        .await;
    }

    #[tokio::test]
    #[instrument]
    async fn gossipsub_broadcast_block_with_ignore() {
        gossipsub_broadcast(
            GossipsubBroadcastRequest::NewBlock(Arc::new(Block::default())),
            GossipsubMessageAcceptance::Ignore,
        )
        .await;
    }
```

crates/services/p2p/src/p2p_service.rs
leviathanbeak: The test confirms 2 things:
- publish message works (A sends to B, B receives it)
- a) message propagation works on `Accept` report (B receives the message and sends it to C)
- b) message is not propagated to C if the report is not `Accept`

In order to test either a) or b) we need to `publish_message` anyway hence we test both things in the same test(s).
`report_message_validation_result` is useless without actual publishing.

`service.rs` should test it's own methods, channels etc, but p2p_service should be tested in isolation that it works as a unit regardless of its wrapper (`service.rs`).

crates/services/p2p/src/p2p_service.rs
Voxelot: lets defer the timeout issue to a followup tech-debt item

Voxelot (approved):

crates/services/txpool/src/service.rs
```diff
@@ -140,16 +142,33 @@ where
                 should_continue = false;
             }
             new_transaction = self.gossiped_tx_stream.next() => {
-                if let Some(GossipData { data: Some(tx), .. }) = new_transaction {
+                if let Some(GossipData { data: Some(tx), message_id, peer_id }) = new_transaction {
```
Voxelot, line 145: we could get rid of the option around data now cc @xgreenx

PR #942, opened 2023-01-26
Addressed the comment in the PR

xgreenx:
Addressed https://github.com/FuelLabs/fuel-core/pull/936#discussion_r1087182354

### DIFF
--- a/crates/services/p2p/src/service.rs
+++ b/crates/services/p2p/src/service.rs
@@ -46,6 +46,7 @@ use fuel_core_types::{
         BlockHeightHeartbeatData,
         GossipData,
         GossipsubMessageAcceptance,
+        GossipsubMessageInfo,
         TransactionGossipData,
     },
 };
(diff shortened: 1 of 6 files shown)


crates/services/p2p/src/service.rs
```diff
@@ -312,19 +313,14 @@ pub struct SharedState {
...
-    where
-        GossipsubMessageInfo: From<&'a T>,
-    {
-        let msg_info = message.into();
-
+    ) -> anyhow::Result<()> {
         self.request_sender
             .try_send(TaskRequest::RespondWithGossipsubMessageReport((
-                msg_info, acceptance,
+                message, acceptance,
```
leviathanbeak, line 323: I'd prefer this is called `message_info` rather than `message` since it's only info/metadata of the message not the message itself

crates/services/txpool/src/service.rs
```diff
@@ -158,7 +155,11 @@ where
                         }
                         _ => None
                     } {
-                        let _ = self.shared.p2p.notify_gossip_transaction_validity(&new_transaction, acceptance);
+                        let message = GossipsubMessageInfo {
```
leviathanbeak, line 158: here as well

leviathanbeak (commented):
this is the straightforward way to solve the interface problem, seems fine for current use cases, not sure do we become less flexible without `take_data` approach, but can't think of any case right now

PR #960, opened 2023-01-27
Fixed binary publishing for `arm64`

xgreenx:


deployment/scripts/chainspec/dev_chainspec.json
```diff
@@ -1,10 +1,5 @@
 {
   "chain_name": "Testnet",
-  "block_production": {
```
ControlCplusControlV, line 3: Why is deployment chainspec getting changed in a publishing fix?

Voxelot: This section has been replaced by a "consensus" portion. We no longer use the chainspec for poa trigger configuration since verifiers can't check things like whether or not the poa produced the block instantly. Chainspec should only contain fields necessary for verification of blocks, not general node-specific parameters that can't be checked by anyone else.

Voxelot (approved):
nice work on testing and fixing this!

PR #962, opened 2023-01-28
Voxelot/add p2p to keygen

Voxelot:
enables generating from libp2p keys + peer ids using the fuel-core-keygen CLI


leviathanbeak (approved):
:shipit:

PR #963, opened 2023-01-28
Update to Rust 1.67 and FIx Clippy Lints

ControlCplusControlV:


ControlCplusControlV:
Mainly thinking of updating so soon because of the [fuel-vm pr](https://github.com/FuelLabs/fuel-vm/pull/332) to use new stabilized std lib function, so would be good to ensure everything is on the same version

leviathanbeak (approved):
clippy ftw

ControlCplusControlV:
Do we need another review before merge? I know updating rust versions can sometimes break things so a bit hesitant to merge but if its fine I can

freesig (approved):
I think it's fine to merge

xgreenx (approved):

Voxelot:
We should also update the rust version in our docker image build as well

PR #954, opened 2023-01-27
Adding helm charts for sentry support

rfuelsh:
Refs: https://github.com/FuelLabs/fuel-core/issues/481 https://github.com/FuelLabs/fuel-core/issues/647

PR #964, opened 2023-01-30
Add block production options to Helm

Voxelot:


xgreenx:
Close it, because the same is implemented in the https://github.com/FuelLabs/fuel-core/pull/954.
Moved `1.67` chef into https://github.com/FuelLabs/fuel-core/pull/965

Issue #681, opened 2022-10-06
Idempotent message insertion


freesig:
Another separate concern is that the way we are inserting messsage in our DB could overwrite the spent status if the message was already downloaded on a previous attempt, allowing for double spend. For example, if the write log fails after partially saving a page and then retries again sometime later, a user could have spent one of the messages that made it into the DB (changing its state). When the relayer retries and overwrites the message, the spent status field would be reset, making the message elidgable for spending again. 

We should check if the message ID exists before saving it to avoid issues with this method not being idempotent. While the previous fix reduces the window of time that this could happen, it's still a risk even if we only have a lag-1 page retry. 

In the future, it might be worth tracking the spent status of messages in a separate column to be able to safely prune the data of spent messages from state without risking double-spends in case the relayer da_height is reset for some reason.

_Originally posted by @Voxelot in https://github.com/FuelLabs/fuel-core/pull/631#discussion_r988247445_

xgreenx:
Related https://github.com/FuelLabs/fuel-core/issues/781

PR #971, opened 2023-02-01
Relayer syncing can't double spend

### LINKED ISSUE #781: Test relayer can't double spend
Check the relayer can't download same message twice and set from spent to unspent

freesig:
closes #781 #681 #977


crates/fuel-core/src/executor.rs
```diff
@@ -737,8 +761,12 @@ impl Executor {
                 Input::Contract { .. } => {}
                 Input::MessageSigned { message_id, .. }
                 | Input::MessagePredicate { message_id, .. } => {
-                    if let Some(message) = db.storage::<Messages>().get(message_id)? {
-                        if message.fuel_block_spend.is_some() {
+                    if let Some(message) = self
+                        .relayer
+                        .get_message(message_id, &block_da_height)
```
xgreenx, line 766: I guess it is handled by `message.da_height > block_da_height` and you don't need to pass `block_da_height`

crates/fuel-core/src/executor.rs
```diff
@@ -737,8 +761,12 @@ impl Executor {
...
+                        .map_err(|e| ExecutorError::StorageError(e.into()))?
+                    {
+                        if db.message_spent(message_id)? {
```
xgreenx, line 769: Could we rename it to `is_message_spent`?

crates/fuel-core/src/executor.rs
```diff
@@ -737,8 +761,12 @@ impl Executor {
                 Input::Contract { .. } => {}
                 Input::MessageSigned { message_id, .. }
                 | Input::MessagePredicate { message_id, .. } => {
-                    if let Some(message) = db.storage::<Messages>().get(message_id)? {
-                        if message.fuel_block_spend.is_some() {
+                    if let Some(message) = self
```
xgreenx, line 764: We also need to verify that all fields of the message from the predicate are the same in the relayer.

It seems we had a bug previously, so we can create a separate issue for that=)

crates/fuel-core/src/executor.rs
```diff
@@ -809,7 +837,6 @@ impl Executor {
         &self,
         tx: &Tx,
         db: &mut Database,
-        block_height: BlockHeight,
```
xgreenx, line 812: We need to update GraphQL API to return messages with `Spent` status as we do it for `Coin`s.

crates/fuel-core/src/query/balance/asset_query.rs
```diff
@@ -140,12 +140,20 @@ impl<'a> AssetsQuery<'a> {
...
-            .filter_ok(|message| {
-                if let Resource::Message(message) = message {
-                    message.fuel_block_spend.is_none()
-                } else {
-                    true
+            .filter_map(|message| match message {
+                Ok(message) => {
+                    if let Resource::Message(m) = &message {
+                        // Only return messages that have not been spent.
+                        match self.database.message_spent(&m.id()) {
```
xgreenx, line 147: We should already fetch messages with this information. Could you add a new type, `CompressedMessage`, please?

`CompressedMessage` will be a message only with fields that are required by the blockchain. `Message` will be the entity used by GraphQL with `CoinStatus`. We already do the same for `Coin` and `CompressedCoin`.

crates/fuel-core/src/service/adapters/executor.rs
```diff
@@ -46,3 +51,38 @@ impl ExecutorAdapter {
...
+        use fuel_core_storage::{
+            tables::Messages,
+            StorageAsRef,
+        };
+        use std::borrow::Cow;
+        #[cfg(feature = "relayer")]
+        {
+            match self.relayer_synced.as_ref() {
+                Some(sync) => sync.get_message(id, da_height),
+                None => Ok(self
```
xgreenx, line 70: We should request messages only from relayer If it is enabled.

crates/fuel-core/src/service/adapters/graphql_api.rs
```diff
@@ -117,6 +118,11 @@ impl DatabaseMessages for Database {
             .into_boxed()
     }
 }
+impl DatabaseSpentMessages for Database {
```
xgreenx, line 121: We don't need an additional method. `owned_message_ids` and `all_messages` should return messages with this information.

crates/fuel-core/src/service/adapters/producer.rs
```diff
@@ -76,21 +78,27 @@ impl fuel_core_producer::ports::Executor<Database> for ExecutorAdapter {
...
+                sync.await_at_least_synced(height).await?;
             }
 
             Ok(self.database.get_finalized_da_height().unwrap_or_default())
         }
         #[cfg(not(feature = "relayer"))]
         {
-            Ok(Default::default())
+            anyhow::ensure!(
+                **da_height == 0,
```
xgreenx, line 97:
```suggestion
                **height == 0,
```

crates/fuel-core/src/service/adapters/sync.rs
```diff
@@ -83,3 +86,35 @@ impl ConsensusPort for VerifierAdapter {
...
+impl RelayerPort for MaybeRelayerAdapter {
+    async fn await_until_if_in_range(
+        &self,
+        da_height: &DaBlockHeight,
+        max_da_lag: &DaBlockHeight,
+    ) -> anyhow::Result<()> {
+        #[cfg(feature = "relayer")]
+        {
+            use fuel_core_relayer::ports::RelayerDb;
+            let current_height = self.database.get_finalized_da_height()?;
```
xgreenx, line 100: You should get this height from the `relayer`

crates/services/relayer/src/ports.rs
```diff
@@ -1,27 +1,133 @@
...
-    /// Add bridge message to database. Messages are not revertible.
-    fn insert_message(
-        &mut self,
-        message: &CheckedMessage,
-    ) -> StorageResult<Option<Message>>;
+    /// Add bridge messages to database. Messages are not revertible.
+    /// Must set the maximum da height for these messages in
+    /// the same write transaction.
+    /// Must only set a new da height if it is greater than the current.
+    fn insert_messages(&mut self, messages: &[CheckedMessage]) -> StorageResult<()>;
```
xgreenx, line 30: We don't need to require the calculation of max height by the database based on messages. We need to pass the height of the finalization from the relayer.

Like "Hey, database, please save the final state of the synchronization with those messages at X height".

crates/services/relayer/src/ports.rs
```diff
@@ -1,27 +1,133 @@
...
+                .insert(DA_HEIGHT_KEY, height)?;
+        }
+    }
+    Ok(())
+}
+
+/// Metadata for relayer.
+pub struct RelayerMetadata;
+impl Mappable for RelayerMetadata {
+    type Key = [u8];
```
xgreenx, line 124: Why don't you use `()`?

crates/services/relayer/src/ports.rs
```diff
@@ -1,27 +1,133 @@
...
+        .map(|cow| cow.as_u64());
+    match current {
+        Some(current) => {
+            if **height > current {
+                s.storage::<RelayerMetadata>()
+                    .insert(DA_HEIGHT_KEY, height)?;
+            }
+        }
+        None => {
+            s.storage::<RelayerMetadata>()
```
xgreenx, line 114: I guess it should be an error because we should set metadata before the start of the relayer.

crates/services/relayer/src/ports.rs
```diff
@@ -1,27 +1,133 @@
...
+where
+    Storage: StorageMutate<RelayerMetadata, Error = StorageError>,
+{
+    let current = (&s)
+        .storage::<RelayerMetadata>()
+        .get(DA_HEIGHT_KEY)?
+        .map(|cow| cow.as_u64());
+    match current {
+        Some(current) => {
+            if **height > current {
```
xgreenx, line 108: Hmm, maybe we want to return an error here, if it is not monotonical

crates/services/sync/src/import.rs
```diff
@@ -186,6 +215,9 @@ where
                         return Ok(None)
                     }
 
+                    // Wait for the da to be at least the da height on the header.
+                    await_at_least_synced(relayer.as_ref(), &header.entity.da_height, &params.max_da_lag, params.max_wait_time).await?;
```
xgreenx, line 219: You need to do it before `check_sealed_header`.

Also, I guess it is the responsibility of the `BlockVerifier` to await `Relaeyr`.
You don't need to pass `Relayer` to the `fuel-core-sync`. The method `check_sealed_header` should be async and here you need to call it with a timeout.

crates/storage/src/tables.rs
```diff
@@ -100,6 +100,16 @@ impl Mappable for Messages {
     type OwnedValue = Message;
 }
 
+/// The storage table of spent [`Message`](crate::model::Message)s.
```
xgreenx, line 103:
```suggestion
/// The storage table indicates was the [`Message`](crate::model::Message) spent or not.
```

crates/storage/src/test_helpers.rs
```diff
@@ -25,14 +31,152 @@ impl AsMut<EmptyStorage> for EmptyStorage {
     }
 }
 
-impl Transactional<EmptyStorage> for EmptyStorage {
-    fn transaction(&self) -> StorageTransaction<EmptyStorage> {
+impl Transactional for EmptyStorage {
+    type Storage = EmptyStorage;
```
xgreenx, line 35:
```suggestion
    type Storage = EmptyStorage;
    
```

crates/storage/src/transactional.rs
```diff
@@ -2,10 +2,13 @@
 
 use crate::Result as StorageResult;
 
+#[cfg_attr(feature = "test-helpers", mockall::automock(type Storage = ();))]
```
xgreenx, line 5: `EmptyStorage` exists for this purpose=)

xgreenx (changes requested):

crates/fuel-core/src/executor.rs
```diff
@@ -109,13 +107,19 @@ use tracing::{
...
 /// In production mode, block fields like transaction commitments are set based on the executed txs.
 /// In validation mode, the processed block commitments are compared with the proposed block.
 #[derive(Clone, Debug)]
-pub struct Executor {
+pub struct Executor<R>
+where
+    R: RelayerPort + Clone,
+{
     pub database: Database,
+    pub relayer: R,
```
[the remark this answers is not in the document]
freesig: Why is this needed? It breaks a lot of code to do that instead of just putting it into the struct.

crates/fuel-core/src/executor.rs
freesig: Yes but this is more robust because it doesn't rely on the caller to do the check (even though they do)

crates/services/relayer/src/ports.rs
freesig: Why would we error? If we are resyncing this could be true right?

crates/services/relayer/src/ports.rs
freesig: I think this is the function used to do that though

crates/services/sync/src/import.rs
freesig: Also if we really need to move it down to verify then we can do that in a follow up because it will also break the producer which isn't great

xgreenx: To verify the signature, we need access to the relayer and logs.

`BlockProducer` doesn't use `BlockVerifeir`, so it should not affect him

freesig: Why do we need logs to verify the signature?

crates/fuel-core/src/schema/message.rs
```diff
@@ -52,10 +53,13 @@ impl Message {
...
         self.0.da_height.as_u64().into()
     }
+}
 
-    async fn fuel_block_spend(&self) -> Option<U64> {
-        self.0.fuel_block_spend.map(|v| v.into())
-    }
+#[derive(Enum, Copy, Clone, Eq, PartialEq)]
+#[graphql(remote = "entities::message::MessageStatus")]
+pub enum MessageStatus {
```
xgreenx, line 60: You can reuse `CoinStatus` because during the next iteration we will move it into coin=)

crates/fuel-core/src/service/adapters/executor.rs
```diff
@@ -46,3 +51,26 @@ impl ExecutorAdapter {
 impl ContractStorageTrait for Database {
     type InnerError = StorageError;
 }
+
+impl crate::executor::RelayerPort for MaybeRelayerAdapter {
+    fn get_message(
+        &self,
+        id: &fuel_core_types::fuel_types::MessageId,
+        da_height: &DaBlockHeight,
+    ) -> anyhow::Result<Option<fuel_core_types::entities::message::CompressedMessage>>
```
xgreenx, line 60: It is a too long line, so why not import them directly?=)

crates/fuel-core/src/service/adapters/executor.rs
```diff
@@ -46,3 +51,26 @@ impl ExecutorAdapter {
...
+        #[cfg(feature = "relayer")]
+        {
+            match self.relayer_synced.as_ref() {
+                Some(sync) => sync.get_message(id, da_height),
+                None => Ok(None),
+            }
+        }
+        #[cfg(not(feature = "relayer"))]
+        {
+            core::mem::drop(id);
```
xgreenx, line 71: Hmm, maybe it is better to add `_` or `allow(...)`

crates/fuel-core/src/service/adapters/sync.rs
```diff
@@ -83,3 +86,34 @@ impl ConsensusPort for VerifierAdapter {
...
+        &self,
+        da_height: &DaBlockHeight,
+        max_da_lag: &DaBlockHeight,
+    ) -> anyhow::Result<()> {
+        #[cfg(feature = "relayer")]
+        {
+            if let Some(sync) = self.relayer_synced.as_ref() {
+                let current_height = sync.get_finalized_da_height()?;
+                anyhow::ensure!(
+                    current_height.abs_diff(**da_height) <= **max_da_lag,
```
xgreenx, line 102: I would say that `da_height - current_height <= max_da_lah` is required, but if `current_height > da_height`, we can instantly return without any error.

crates/services/relayer/src/mock_db.rs
```diff
@@ -37,27 +40,42 @@ pub struct MockDb {
 }
 
 impl MockDb {
```
xgreenx, line 42: Hmm, maybe we want to replace it with `mockall`?
It's up to use, but it seems it would reduce the amount of code for you=)

crates/services/relayer/src/ports.rs
xgreenx: Oh yeah, you use it in the `set_finalized_da_height_to_at_least`

crates/services/relayer/src/ports.rs
xgreenx: Is not the purpose of this function to return an error if the height is not monotonic? I guess with the error, we can be sure that each commit is monatomic.

If we do re-sync of the relayer, I guess we will clean up the state of the relayer first and start re-sync after.

crates/services/relayer/src/ports.rs
```diff
@@ -1,27 +1,142 @@
...
+    type Key = Self::OwnedKey;
+    type OwnedKey = ();
+    type Value = Self::OwnedValue;
+    type OwnedValue = DaBlockHeight;
+}
+
+/// Key for da height.
+/// If the relayer metadata ever contains more than one key, this should be
+/// changed from a unit value.
+const DA_HEIGHT_KEY: () = ();
```
xgreenx, line 142:
```suggestion
const METADATA_KEY: () = ();
```

crates/services/relayer/src/service.rs
```diff
@@ -207,11 +211,54 @@ impl SharedState {
...
+    /// and is <= the given height.
+    pub fn get_message(
+        &self,
+        id: &MessageId,
+        da_height: &DaBlockHeight,
+    ) -> anyhow::Result<Option<CompressedMessage>>
+    where
+        D: StorageInspect<Messages, Error = fuel_core_storage::Error>,
+    {
+        Ok(self
```
xgreenx, line 246: Maybe you want to put it also into the `RelayerDB` trait?

Voxelot:
Seems like these tests are stalling out our CI time limit:
```
test messages::can_submit_genesis_message has been running for over 60 seconds
test relayer::messages_are_spendable_after_relayer_is_synced has been running for over 60 seconds
```

crates/services/relayer/src/service.rs
```diff
@@ -148,17 +149,20 @@ where
...
+    type SharedData = SharedState<D>;
     type Task = Task<P, D>;
 
     fn shared_data(&self) -> Self::SharedData {
         let synced = self.synced.subscribe();
 
-        SharedState { synced }
+        SharedState {
+            synced,
+            database: self.database.clone(),
```
xgreenx, line 164: I think it is better to put `Database` into `Arc`. It is used everywhere only to read except `insert_messages`(but you use db transaction inside, so it also could be read).

crates/services/sync/src/import.rs
xgreenx: Because the `Relayer` is our source of the truth about block producers(But it depends on our key rotation strategy).

crates/storage/src/transactional.rs
```diff
@@ -2,10 +2,13 @@
 
 use crate::Result as StorageResult;
 
+#[cfg_attr(feature = "test-helpers", mockall::automock(type Storage = crate::test_helpers::EmptyStorage;))]
 /// The types is transactional and may create `StorageTransaction`.
-pub trait Transactional<Storage: ?Sized> {
+pub trait Transactional {
+    /// The storage used when creating the transaction.
+    type Storage: ?Sized;
```
xgreenx, line 9:
```suggestion
    type Storage: ?Sized;

```

crates/types/src/entities/message.rs
```diff
@@ -37,11 +35,20 @@ pub struct Message {
...
     pub da_height: DaBlockHeight,
-    /// When the message was spent in the Fuel chain
-    pub fuel_block_spend: Option<BlockHeight>, // TODO: get rid of this
 }
 
-impl Message {
+/// Message send from Da layer to fuel by bridge
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct Message {
```
xgreenx, line 43: I know that it is annoying, but could you duplicate all fields of the `CompressedMessage` here?

We want to make the life of the SDK users as easer as possible.

crates/types/src/entities/message.rs
```diff
@@ -37,11 +35,20 @@ pub struct Message {
...
+    #[cfg_attr(feature = "serde", serde(flatten))]
+    /// The message content relevant to the blockchain.
+    pub message: CompressedMessage,
+    /// Whether a message has been spent or not
+    pub status: MessageStatus,
```
xgreenx, line 48: We can use `CoinStatus` here for now. We plan to use it in the future for `MessageCoin`s

crates/services/sync/src/import.rs
freesig: Ah I see. Ok will make this change next week. I still think the signature check should be before awaiting the relayer though. Even if it's all done within the verifier. We should at least check the signature is correct for the header so we aren't awaiting for a da_height that no one has signed.

crates/fuel-core/src/schema/message.rs
Voxelot: not true, there will be DepositCoins and regular old messages. Not all messages will become coins, just the ones that only include gas.

crates/services/relayer/src/mock_db.rs
freesig: Yeh probably but this is unrelated to this PR

crates/services/relayer/src/service.rs
freesig: I don't see why? It's only used here.

crates/services/relayer/src/service.rs
freesig: It's up the the generic provider to provide a Clonable database and they all are clone

crates/types/src/entities/message.rs
freesig: It's flatten on serde though so it is reproduced. The SDK users will consume this as a flattened struct.

xgreenx: SDK users will use `fuel-core-types`, they don’t need to create their version of this structure

freesig: Ok can do

freesig:
I think all pr suggestions have been met. If tests actually pass we should be good

crates/client/src/client.rs
```diff
@@ -473,7 +476,9 @@ impl FuelClient {
         use cynic::SubscriptionBuilder;
         let s = schema::tx::StatusChangeSubscription::build(TxIdArgs { id: id.parse()? });
 
+        tracing::debug!("subscribing");
```
Voxelot, line 479: should we include the tx id here or is that covered by the instrumentation span?

freesig: It's included in the span above

crates/fuel-core/src/service/adapters/block_importer.rs
```diff
@@ -75,6 +79,37 @@ impl BlockVerifier for VerifierAdapter {
...
+#[async_trait::async_trait]
+impl RelayerPort for MaybeRelayerAdapter {
+    async fn await_until_if_in_range(
+        &self,
+        da_height: &DaBlockHeight,
+        max_da_lag: &DaBlockHeight,
+    ) -> anyhow::Result<()> {
+        #[cfg(feature = "relayer")]
+        {
+            if let Some(sync) = self.relayer_synced.as_ref() {
```
Voxelot, line 91: What if the relayer is still syncing for a new node? Maybe we should wait for the relayer to finish syncing with eth before checking da_height against max_da_lag?

Voxelot: I guess if this fails, the stream will just start over. However, if we don't handle this case here it could be harder to determine whether the advertised da_height was malicious from the peer reputation perspective.

freesig: The relayer is always kind of syncing so it's kind of hard to say if it's doing an initial sync. Even if we did know that I don't know what we should do differently?

Voxelot (approved):

Voxelot (approved):

PR #972, opened 2023-02-01
conditionally set consensus secret

Voxelot:


Voxelot:
this seems to always enable the secret even when it shouldn't, not going to block beta3 launch on this though since it's just a testnet.

Voxelot:
closing for now, will address this issue closer to mainnet based on the progress of the operator

Issue #967, opened 2023-01-31
Unite test failed: The node should only connect to the specified reserved nodes!


xgreenx:
Investigate fail of the [`sentry_nodes_working`](https://github.com/FuelLabs/fuel-core/actions/runs/4051236761/jobs/6969363155) unit test

xgreenx:
https://github.com/FuelLabs/fuel-core/actions/runs/4055609356/jobs/6978991218

PR #974, opened 2023-02-02
Fixed bug with disconnection and `max_non_reserved_peers`

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/967

### DIFF
--- a/crates/services/p2p/src/p2p_service.rs
+++ b/crates/services/p2p/src/p2p_service.rs
@@ -461,16 +461,9 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
                     }
                     return Some(FuelP2PEvent::PeerDisconnected(peer_id))
                 }
-                PeerInfoEvent::TooManyPeers {
-                    peer_to_disconnect,
-                    peer_to_connect,
-                } => {
+                PeerInfoEvent::TooManyPeers { peer_to_disconnect } => {
                     // disconnect the surplus peer
                     let _ = self.swarm.disconnect_peer_id(peer_to_disconnect);
-                    // reconnect the reserved peer
-                    if let Some(peer_id) = peer_to_connect {
-                        let _ = self.swarm.dial(peer_id);
-                    }
                 }
             },
             FuelBehaviourEvent::RequestResponse(req_res_event) => match req_res_event {
@@ -876,7 +869,7 @@ mod tests {
         let jh = tokio::spawn(async move {
             while rx.try_recv().is_err() {
                 futures::stream::iter(node_services.iter_mut())
-                    .for_each_concurrent(10, |node| async move {
+                    .for_each_concurrent(20, |node| async move {
                         node.next_event().await;
                     })
                     .await;
--- a/crates/services/p2p/src/peer_manager.rs
+++ b/crates/services/p2p/src/peer_manager.rs
@@ -74,7 +74,6 @@ pub enum PeerInfoEvent {
     },
     TooManyPeers {
         peer_to_disconnect: PeerId,
-        peer_to_connect: Option<PeerId>,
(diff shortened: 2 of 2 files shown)


crates/services/p2p/src/p2p_service.rs
```diff
@@ -461,16 +461,9 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
...
                     return Some(FuelP2PEvent::PeerDisconnected(peer_id))
                 }
-                PeerInfoEvent::TooManyPeers {
-                    peer_to_disconnect,
-                    peer_to_connect,
-                } => {
+                PeerInfoEvent::TooManyPeers { peer_to_disconnect } => {
                     // disconnect the surplus peer
                     let _ = self.swarm.disconnect_peer_id(peer_to_disconnect);
-                    // reconnect the reserved peer
```
xgreenx, line 470: We already have a `health` timer that connects to disconnected reserved peers. We don't need to duplicate this work here=)

crates/services/p2p/src/peer_manager.rs
```diff
@@ -199,13 +198,13 @@ impl NetworkBehaviour for PeerManagerBehaviour {
...
 
-                let addresses = self.addresses_of_peer(&peer_id);
-                self.insert_peer_addresses(&peer_id, addresses);
-
                 if other_established == 0 {
                     // this is the first connection to a given Peer
                     self.peer_manager.handle_initial_connection(peer_id);
                 }
+
+                let addresses = self.addresses_of_peer(&peer_id);
```
xgreenx, line 206: First, we need to insert the peer into the table in the `handle_initial_connection`, and after we can add peer addresses=)

crates/services/p2p/src/peer_manager.rs
```diff
@@ -611,10 +603,9 @@ impl PeerManager {
         if !self.reserved_peers.contains(&peer_id) {
             // check if all the slots are already taken
             if non_reserved_peers_connected >= self.max_non_reserved_peers {
-                // Too many peers already connected, disconnect the Peer
-                self.pending_events.push_back(PeerInfoEvent::TooManyPeers {
+                // Too many peers already connected, disconnect the Peer with the first priority.
+                self.pending_events.push_front(PeerInfoEvent::TooManyPeers {
```
xgreenx, line 607: I think we need to disconnect from peers in the first place. Otherwise, we will have a pending task that will receive incoming messages that we plan to drop.

crates/services/p2p/src/peer_manager.rs
```diff
@@ -645,28 +636,33 @@ impl PeerManager {
     fn handle_peer_disconnect(&mut self, peer_id: PeerId) {
         // try immediate reconnect if it's a reserved peer
         let is_reserved = self.reserved_peers.contains(&peer_id);
+        let is_removed;
```
xgreenx, line 639: In the `handle_initial_connection` we don't connect to the peer -> it absents in our `self.non_reserved_connected_peers` map.

We had a bug here that we will `allow_new_peers` even if the peer was not connected -> we allowed new connections when we had `max_non_reserved_peers`.

crates/services/p2p/src/p2p_service.rs
```diff
@@ -876,7 +869,7 @@ mod tests {
         let jh = tokio::spawn(async move {
             while rx.try_recv().is_err() {
                 futures::stream::iter(node_services.iter_mut())
-                    .for_each_concurrent(10, |node| async move {
+                    .for_each_concurrent(20, |node| async move {
```
freesig, line 872: Should this at least be a const value?

freesig (approved):
Nice work

Issue #622, opened 2022-09-14
Some improvements during work with the database


xgreenx:
Use `Option<&[u8]>` instead of `Option<Vec<u8>>` in the `iter_all` of `fuel_core::state::KeyValueStore`. Also decide, do we really need usage of `Option`? If `len` is zero it is the same as `None`. Apply the same change for all upper functions.

<img width="759" alt="image" src="https://user-images.githubusercontent.com/18346821/190230186-4b00ecd8-56cc-4b67-a997-61bda212a0e7.png">

Also there in the implementation of the iterator, we can avoid the copy of the slices into vectors. Also think about [comment](https://github.com/FuelLabs/fuel-core/pull/591#discussion_r970116644).

<img width="793" alt="image" src="https://user-images.githubusercontent.com/18346821/190230122-eed1758c-75f1-4312-9771-2da7a55f28fd.png">

Minor improvement in `Database::insert`:
<img width="904" alt="image" src="https://user-images.githubusercontent.com/18346821/190273055-c90112e0-7388-4a9c-ae82-c22296614fbb.png">

To use the same naming across the project in `Database::exists`:
<img width="903" alt="image" src="https://user-images.githubusercontent.com/18346821/190273090-697510e9-29fb-4a49-bf7f-02bdade96415.png">

Voxelot:
>  If len is zero it is the same as None

Not necessarily, there is value in being able to distinguish between empty values and non-existent ones.

xgreenx:
@Voxelot What do you think about renaming all `Database`'s methods by underscore? To highlight that they are internal methods, and you need to work with `StorageMutate` and `StorageInspect`. 

- `Database::insert` -> `Database::_insert`
- `Database::remove` -> `Database::_remove`
- `Database::get` -> `Database::_get`
- `Database::exists` -> `Database::_contains_key`
- `Database::iter_all` -> `Database::_iter_all`

AliJafriETH:
Hey @xgreenx & @Voxelot, is the issue still open. i'd like to work on it :))

xgreenx:
There is in progress PR https://github.com/FuelLabs/fuel-core/pull/643 that is blocked by refactoring https://github.com/FuelLabs/fuel-core/issues/809. The change is almost ready, so it is better to select another issue=D

PR #976, opened 2023-02-03
Some improvements during work with the database

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/622

Each point of the issue is made as a separate commit, so it may be simpler to review each commit.

### DIFF
--- a/crates/fuel-core/src/database.rs
+++ b/crates/fuel-core/src/database.rs
@@ -177,7 +177,7 @@ impl Database {
         }
     }
 
-    fn _insert<K: AsRef<[u8]>, V: Serialize, R: DeserializeOwned>(
+    fn insert<K: AsRef<[u8]>, V: Serialize, R: DeserializeOwned>(
         &self,
         key: K,
         column: Column,
@@ -197,7 +197,7 @@ impl Database {
         }
     }
 
-    fn _remove<V: DeserializeOwned>(
+    fn remove<V: DeserializeOwned>(
         &self,
         key: &[u8],
         column: Column,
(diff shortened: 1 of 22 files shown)


crates/fuel-core/src/database.rs
```diff
@@ -177,19 +177,16 @@ impl Database {
         }
     }
 
-    // TODO: Get `K` and `V` by reference to force compilation error for the current
-    //  code(we have many `Copy`).
-    //  https://github.com/FuelLabs/fuel-core/issues/622
-    fn insert<K: AsRef<[u8]>, V: Serialize, R: DeserializeOwned>(
+    fn _insert<K: AsRef<[u8]>, V: Serialize, R: DeserializeOwned>(
```
freesig, line 180: This style of naming is a little hard to understand. Maybe instead of `_` we should use `raw_`. So `raw_insert` etc.?

freesig (changes requested):

crates/fuel-core/src/database.rs
Voxelot: Usually underscores are meant to denote unused code. Since these fn's are already private idk why we need to underscore them?

xgreenx: I thought that we agreed on it [here](https://github.com/FuelLabs/fuel-core/issues/622#issuecomment-1247361991). But it is not true, I didn't get any feedback, I don't know why I thought that=D Reverted it back=)

Voxelot (approved):
nice work, will be interesting to see how much this improves perf by re-benchmarking after we get through the rest of these todos

crates/fuel-core/src/database.rs
freesig: Sorry I didn't see that issue :)

PR #982, opened 2023-02-03
replace usage of futures-timer to avoid extra helper threads

Voxelot:
futures-timer was being used in p2p, however it makes use of helper threads which are unnecessary since we use a tokio runtime.


bvrooman (approved):
LGTM. Were you able to manually verify that there were the expected number of threads now?

PR #983, opened 2023-02-04
parallelize check & test for workspace members

Voxelot:


Voxelot:
seems like there's too much overhead from github CI when creating hundreds of parallel jobs to see any performance improvements. We'll need to investigate other alternatives hyper-parallelization to improve our pipeline (e.g. find a balance by only parallelizing the slowest parts)

PR #984, opened 2023-02-06
fix(client): feature-gate tx subscription for WASM

Br1ght0ne:
To allow `fuels-types` to use `fuel-core-client` on WASM (and unblock the Indexer team), usage of `eventsource-client` (which uses TCP, sockets etc.) needs to be feature-gated while I'm working on bringing the subscription functionality to WASM.

This should be merged before #446, as it still needs to be completed and is of much lesser priority.


deekerno (dismissed):
LGTM.

Voxelot (approved):

PR #987, opened 2023-02-06
Avoid feature unification during testing of --no-default-features

Voxelot:
our --no-default-features testing wasn't working as expected due to feature unification. This change ensures each crate is tested in isolation with no extra features enabled.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,8 +99,9 @@ jobs:
             args: check --locked
           - command: test
             args: --all-features --workspace
-          - command: test
-            args: --no-default-features --workspace
+          # use cargo make test to avoid workspace feature unification
+          - command: make
+            args: test --no-default-features --workspace
           - command: test
             args: --manifest-path version-compatibility/Cargo.toml --workspace
     # disallow any job that takes longer than 45 minutes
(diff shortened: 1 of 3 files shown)


Voxelot:
Hmm this is too slow (which is probably why we didn't do it before). Will investigate how we can speed this up.

.github/workflows/ci.yml
```diff
@@ -99,8 +99,9 @@ jobs:
             args: check --locked
           - command: test
             args: --all-features --workspace
-          - command: test
```
freesig, line 102: We need to at least add a cargo check at the workspace level otherwise running `cargo check` in the repo can break (which breaks IDEs and is just bad practice)

freesig (changes requested):

.github/workflows/ci.yml
Voxelot: this only affected test, we have a `cargo check` job 4 lines above this.

freesig (dismissed):

freesig (approved):

Voxelot:
Still pretty slow after fixing the --workspace flag. Takes 30m to build and test each crate individually. Might be worth doing the slowest crates in a parallel job.

freesig (approved):

Voxelot:
looks like after warming up the cache, the build time is now the same as the normal test job with all features enabled.

PR #988, opened 2023-02-06
Ignore unpublish changes on standard publish verification

Voxelot:
Ignore unpublished changes when simulating publishing on master / PR CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,22 +140,22 @@ jobs:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_NOTIFY_BUILD }}
           RUSTFLAGS: -D warnings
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -155,11 +155,12 @@ jobs:
         with:
           dry-run: true
           check-repo: false
+          ignore-unpublished-changes: true
 
   verifications-complete:
     needs:
       - cargo-verifications
-      - publish-crates-check
```
freesig, line 162: Should this be a comment or deleted?

Voxelot: ah good catch, adding it back. 

I originally set out to disable this job, but then realized it might work with `ignore-unpublished-changes: true` set.

freesig (approved):

ControlCplusControlV (approved):

PR #990, opened 2023-02-06
Stop the `ServiceRunner` if it is not used

xgreenx:


crates/services/src/service.rs
```diff
@@ -101,16 +101,25 @@ pub trait RunnableTask: Send {
...
     pub shared: S::SharedData,
     state: Shared<watch::Sender<State>>,
 }
 
+impl<S> Drop for ServiceRunner<S>
+where
+    S: RunnableService + 'static,
+{
+    fn drop(&mut self) {
+        self.stop();
```
Voxelot, line 116: wouldn't this only help if the service doesn't own the join handles of any sub-tasks it spawns? Or will dropping join handles still allow tspawned tasks to continue executing?

Voxelot: > A JoinHandle detaches the associated task when it is dropped, which means that there is no longer any handle to the task, and no way to join on it.

https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html

Looks like the spawned tasks just get orphaned and continue to run if the owner drops the join handles.

Voxelot (dismissed):

crates/services/src/service.rs
xgreenx: Yep

Voxelot: hmm looks like this caused a bunch of test failures

xgreenx: Yep because we create a clone of `TxPool` service during each GraphQL request with transactions=)))))

freesig: Yeh probably needs to be within an Arc?

Voxelot: I'm thinking we probably should allow clonable objects to drop inner resources, to avoid any confusion in the future about which resources are really cloneable. Something like an Arc makes sense.

crates/services/src/service.rs
```diff
@@ -101,16 +101,25 @@ pub trait RunnableTask: Send {
...
+    }
+}
+
 impl<S> Clone for ServiceRunner<S>
 where
-    S: RunnableService,
+    S: RunnableService + 'static,
 {
     fn clone(&self) -> Self {
```
Voxelot, line 124: We should either remove clone or make drop only stop the service when the last cloned instance of the service runner drops.

Voxelot (approved):

PR #991, opened 2023-02-06
Don't capture signals within lib

### LINKED ISSUE #980: Tests leave zombie process
When shutting down our integration tests with `ctrl-c` there is a zomebie process left behind

freesig:
closes #980

### DIFF
--- a/crates/services/src/service.rs
+++ b/crates/services/src/service.rs
@@ -218,7 +218,7 @@ where
     }
 }
 
-#[tracing::instrument(skip_all)]
+#[tracing::instrument(skip_all, fields(service = S::NAME))]
 /// Initialize the background loop as a spawned task.
 fn initialize_loop<S>(service: S) -> Shared<watch::Sender<State>>
 where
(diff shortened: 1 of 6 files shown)


crates/services/src/service.rs
```diff
@@ -218,7 +218,7 @@ where
     }
 }
 
-#[tracing::instrument(skip_all)]
+#[tracing::instrument(skip_all, fields(service = S::NAME))]
```
Voxelot, line 221: This is cool! Didn't realize this was an available option to annotate all the logs from a struct instance.

tests/tests/messages.rs
```diff
@@ -70,7 +70,10 @@ async fn can_submit_genesis_message() {
     assert!(
         matches!(status, TransactionStatus::Success { .. }),
         "expected success, received {status:?}",
-    )
+    );
+    loop {
+        tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
+    }
```
Voxelot, line 76:
```suggestion
```

This should be removed. I presume it was added to repro the issue locally?

freesig: ah yeh whoops

Voxelot (approved):

xgreenx (approved):

PR #1000, opened 2023-02-08
avoid compiling usued deps when --no-default-features is used

Voxelot:
speedup builds using `--no-default-features` by disabling extra deps that aren't used


freesig (approved):
I'm ok with this if ci must build benches with no default features. Could also just skip benches for that ci run

ControlCplusControlV (approved):

Voxelot:
> I'm ok with this if ci must build benches with no default features. Could also just skip benches for that ci run

I tried adding 

```
env = { "CARGO_MAKE_WORKSPACE_SKIP_MEMBERS" = ["fuel-core-benches"] }
```

to the Makefile.toml for the test task but it seemed to run anyways.

Issue #482, opened 2022-07-11
Acceptance testing harness that runs against fuel-core CD environment


Voxelot:
Every PR to fuel-core automatically gets deployed to the dev environment in AWS. Acceptance Tests (ATs) are CI tests we run after a deployment to perform basic functional verification that the system is operating as intended.

Reference Material: https://www.slideshare.net/AgileIndia2019/acceptance-testing-for-continuous-delivery-by-dave-farley-at-agileindia2019

MitchTurner:
Related to #1153

MitchTurner:
What about running the sdk tests regularly?

PR #1003, opened 2023-02-09
barebones e2e test client

Voxelot:
This adds a new publishable binary to fuel-core that allows for e2e testing against deployed environments. Ideally tests added here should be relatively lightweight, tolerate concurrency and make minimal assumptions about the state of the network before executing.

Currently the test only performs a basic transfer, but the suite should be expanded to cover other common use-cases like contract deployments.

The main binary uses libtest-mimic to make it compatible with most cargo test runner plugins (e.g. nextest) or common IDE patterns for debugging tests.

It is driven by a configuration tomlfile which allows endpoints and wallet secrets to be configured based on the environment being tested. This is described more in the README.md.

helps unblock https://github.com/FuelLabs/fuel-core/issues/482 and https://github.com/FuelLabs/fuel-k8s-operator/issues/26


leviathanbeak (approved):

Voxelot:
Added a multinode version of the suite integration tests, which assigns a different node to each wallet.

bin/e2e-test-client/src/test_context.rs
```diff
@@ -0,0 +1,191 @@
...
+            {
+                return Ok(true)
+            }
+        }
+
+        Ok(false)
+    }
+
+    /// Transfers coins from this wallet to another
+    pub async fn transfer(
```
xgreenx, line 120: Hmm, I hope we will not bring the logic of the SDK into our e2e tests in the future=)

bin/e2e-test-client/tests/integration_tests.rs
```diff
@@ -0,0 +1,122 @@
...
+    };
+
+    config.wallet_a.endpoint = Some(producer.node.bound_address.to_string());
+    config.wallet_b.endpoint = Some(validator.node.bound_address.to_string());
+
+    // save config file
+    let config = save_config_file(config);
+
+    // execute suite
+    execute_suite(config.path)
```
xgreenx, line 77: Hmm, nice trick to run the e2e test for local environment because `producer` and `validator` are not destroyed at this point yet=D

xgreenx (approved):

PR #1004, opened 2023-02-09
Bugfix: set prefix extractor for all columns that need it

Voxelot:
- Fixes column settings for all APIs that seek using a prefix
- Enables rocksdb in integration_tests when using default features
- Disables backwards pagination for APIs where a prefix is used, as rocksdb doesn't support it

### DIFF
--- a/crates/fuel-core/src/state/rocks_db.rs
+++ b/crates/fuel-core/src/state/rocks_db.rs
@@ -126,8 +126,13 @@ impl RocksDb {
         opts.create_if_missing(true);
         opts.set_compression_type(DBCompressionType::Lz4);
 
+        // All double-keys should be configured here
         match column {
-            Column::OwnedCoins | Column::TransactionsByOwnerBlockIdx => {
+            Column::OwnedCoins
+            | Column::TransactionsByOwnerBlockIdx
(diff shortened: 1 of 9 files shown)


crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -126,8 +126,13 @@ impl RocksDb {
         opts.create_if_missing(true);
         opts.set_compression_type(DBCompressionType::Lz4);
 
+        // All double-keys should be configured here
         match column {
-            Column::OwnedCoins | Column::TransactionsByOwnerBlockIdx => {
+            Column::OwnedCoins
```
bvrooman, line 131: 👍

crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -126,8 +126,13 @@ impl RocksDb {
...
+            | Column::TransactionsByOwnerBlockIdx
+            | Column::OwnedMessageIds
+            | Column::ContractsAssets
+            | Column::ContractsState => {
```
bvrooman, line 135: 👍

bvrooman (approved):

Issue #943, opened 2023-01-26
Peer Reputation for syncing


Voxelot:
The plan is to first come up with an approach how to handle Peer Reputation.
That includes, but not limited to:
- ban & its consequences (potential holes in the approach, different attacks etc)
- peer scoring (both positive and negative?)
- defining implementation details

Once we've made a decision on the approach, and defined implementation details then we can start working on it.

| Offense | Affects Internal Node? | Severity | Explanation | Completing PR |
| ---------- | ------------------------------ | ----------- | --------------- | -------------------- |
| Bad Block Height | No | High | Peering is useless if the values provided don't match what is advertised by peer | https://github.com/FuelLabs/fuel-core/pull/1331
| Bad Header | No | High |  "  |  https://github.com/FuelLabs/fuel-core/pull/1331
| Bad Transactions | No | High |  "  | https://github.com/FuelLabs/fuel-core/pull/1331
| Bad Block | No | High |  "  | https://github.com/FuelLabs/fuel-core/pull/1331
| Heartbeats Stopped | No | Medium |  This is concerning but doesn't guarantee the peer is unreachable. It's best to know that peer at least claims to be available  | https://github.com/FuelLabs/fuel-core/pull/1356 |
| Slow Heartbeats | No | Low |  This is helpful in selecting a peer that is _most_ available, but shouldn't be a reason to ban a member unless it gets really bad. | https://github.com/FuelLabs/fuel-core/pull/1356 |
| Incorrect Response Data | No | High | We should avoid peers that give us bad data. There aren't really situations where a friendly node would be sending bad data on accident, and if they were we might want to cut the connection anyway |  |
| Too Many Requests | No | Medium | This is a serious attack vector, but it's not obvious how we might define "too much". It also depends on the validity of the txs. Too many alwasy can cause problems, but if they are all valid, then its not as punishable | |
| Weak Connection | No | Low | Unclear what this means. This might overlap with "Slow Heartbeats" slightly| |
| Request Timeout | Almost No | Medum | This is more likely to happen with a friendly peer than getting bad data, but in some ways is more harmful as it necessarily takes longer and is just as detrimental to the performance of a node. | https://github.com/FuelLabs/fuel-core/issues/1346 |
| Gossiping bad transactions | No | Low | Malicious peer broadcasts invalid transactions to the network. | |

Rewardable Actions:
- [x] Valid Response (At least partially with https://github.com/FuelLabs/fuel-core/pull/1331)
[...]

PR #1007, opened 2023-02-10
peer reputation framework

leviathanbeak:
This is the initial setup for PeerReputation #943 

The approach was to provide a simple framework where if some of the scoring logic changes,
the API should remain the same, only changes should be done within `types/p2p/peer_reputation`.

Todo:
- [ ] implement unban peers after certain period (?)
- [ ] integrate gossipsub scoring (?)
- [ ] decide what to do with reserved peers and their reputation
- [ ] libp2p can internally also:
     - [ ] increase reputation
     - [ ] decrease reputation

### DIFF
--- a/crates/services/p2p/src/behavior.rs
+++ b/crates/services/p2p/src/behavior.rs
@@ -26,7 +26,10 @@ use crate::{
         RequestMessage,
     },
 };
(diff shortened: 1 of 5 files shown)


leviathanbeak:
closing in favor of #1028

PR #1008, opened 2023-02-10
feat: Contract state and assets merkle data storage

bvrooman:
Related PRs:
- Closes https://github.com/FuelLabs/fuel-core/issues/626
- Closes https://github.com/FuelLabs/fuel-core/issues/627

This PR calculates SMT roots for a contract's state and balances. Both states and balances are modified via random-access inserts and deletes, so roots must be calculated as an SMT on top of a sparse key-value dataset. 

Additionally, this PR modifies the root calculations and related tests to use the correct empty root, the 32 byte 0, rather than the hash of the empty string. The 32 byte 0 is the default for SMTs while the hash of the empty string is the default for BMTs.


crates/fuel-core/src/executor.rs
```diff
@@ -2626,7 +2625,7 @@ mod tests {
     fn contracts_balance_and_state_roots_no_modifications_updated() {
         // Values in inputs and outputs are random. If the execution of the transaction successful,
         // it should actualize them to use a valid the balance and state roots. Because it is not
-        // changes, the balance the root should be default - `fuel_merkle::common::empty_sum_sha256()`.
```
bvrooman, line 2629: This was actually incorrect: the spec states that this should be a 32 byte 0 for any tree with height greater than 0. See https://github.com/celestiaorg/celestia-specs/blob/master/src/specs/data_structures.md#sparse-merkle-tree.

Voxelot:
Is this still WIP?

bvrooman:
> Is this still WIP (title of the PR)?

I am going to add some tests, but the implementation is ready for a first round of review. 
I will convert it back to draft while I add the tests. In the meantime, if anyone has comments on the implementation code, it is completely welcome!

crates/fuel-core/src/database/balances.rs
```diff
@@ -48,37 +59,101 @@ impl StorageMutate<ContractsAssets> for Database {
...
+                // The tree is empty
+                MerkleTree::new(storage)
+            } else {
+                // Load the tree saved in metadata
+                MerkleTree::load(storage, &root)
+                    .map_err(|err| StorageError::Other(err.into()))?
+            }
+        };
+
+        // Update the key-value dataset. The key is the contract id and the
```
xgreenx, line 89: Hmm, is that true? The value is not 32 bytes, it is `Word` - 8 bytes

crates/fuel-core/src/database/balances.rs
```diff
@@ -48,37 +59,101 @@ impl StorageMutate<ContractsAssets> for Database {
         key: &<ContractsAssets as Mappable>::Key,
         value: &<ContractsAssets as Mappable>::Value,
     ) -> Result<Option<<ContractsAssets as Mappable>::OwnedValue>, Self::Error> {
-        Database::insert(self, key.as_ref(), Column::ContractsAssets, value)
-            .map_err(Into::into)
+        let prev = Database::insert(self, key.as_ref(), Column::ContractsAssets, value)
+            .map_err(Into::into);
+
+        // Get latest metadata entry
+        let prev_metadata = self
```
xgreenx, line 66: Why do we need `iter_all` if we have only one entry of the `SparseMerkleMetadata` in the database for the `ContractsAssetsMerkleMetadata` table?

I think you need to use `self.storage::<ContractsAssetsMerkleMetadata>().get(key.contract_id())`. Because now it seems you take the state of any last contract=)

crates/fuel-core/src/database/balances.rs
```diff
@@ -48,37 +59,101 @@ impl StorageMutate<ContractsAssets> for Database {
...
     fn remove(
         &mut self,
         key: &<ContractsAssets as Mappable>::Key,
     ) -> Result<Option<<ContractsAssets as Mappable>::OwnedValue>, Self::Error> {
-        Database::remove(self, key.as_ref(), Column::ContractsAssets).map_err(Into::into)
+        let prev = Database::remove(self, key.as_ref(), Column::ContractsAssets)
+            .map_err(Into::into);
+
+        // Get latest metadata entry
+        let prev_metadata = self
```
xgreenx, line 111: The same comments from the `insert` method here=)

crates/fuel-core/src/database/balances.rs
```diff
@@ -48,37 +59,101 @@ impl StorageMutate<ContractsAssets> for Database {
...
-                (&key[..parent.len()] == parent.as_ref()).then_some((key, value))
-            })
-            .sorted_by_key(|t| t.0)
-            .map(|(_, value)| value.to_be_bytes());
-
-        Ok(crypto::ephemeral_merkle_root(root).into())
+        let metadata = self
+            .storage::<ContractsAssetsMerkleMetadata>()
+            .get(parent)?;
+        let root = metadata
```
xgreenx, line 153: Why in some places we use `[0; 32]` and in some `in_memory::MerkleTree::new().root()`? Maybe the implementation of the `Default` trait for `SparseMerkleMetadata` should use `in_memory::MerkleTree::new().root()` for the `root` field?

crates/fuel-core/src/database/state.rs
```diff
@@ -50,37 +59,99 @@ impl StorageMutate<ContractsState> for Database {
         key: &<ContractsState as Mappable>::Key,
         value: &<ContractsState as Mappable>::Value,
     ) -> Result<Option<<ContractsState as Mappable>::OwnedValue>, Self::Error> {
-        Database::insert(self, key.as_ref(), Column::ContractsState, value)
-            .map_err(Into::into)
+        let prev = Database::insert(self, key.as_ref(), Column::ContractsState, value)
+            .map_err(Into::into);
+
+        // Get latest metadata entry
+        let prev_metadata = self
```
xgreenx, line 66: The same here. You need to modify the specific metadata for the contract, not the last entry in the database=)

crates/fuel-core/src/database/state.rs
```diff
@@ -50,37 +59,99 @@ impl StorageMutate<ContractsState> for Database {
...
     fn remove(
         &mut self,
         key: &<ContractsState as Mappable>::Key,
     ) -> Result<Option<<ContractsState as Mappable>::OwnedValue>, Self::Error> {
-        Database::remove(self, key.as_ref(), Column::ContractsState).map_err(Into::into)
+        let prev = Database::remove(self, key.as_ref(), Column::ContractsState)
+            .map_err(Into::into);
+
+        // Get latest metadata entry
+        let prev_metadata = self
```
xgreenx, line 111: The same here=)

crates/fuel-core/src/database/balances.rs
bvrooman: Yep, thanks for the catch!

crates/fuel-core/src/database/balances.rs
bvrooman: Yes, I think you're right! I will make that change.

crates/fuel-core/src/database/balances.rs
bvrooman: The `Default` trait for `SparseMerkleMetadata` actually does use an empty SMT to get the root: https://github.com/FuelLabs/fuel-core/pull/1008/files#diff-7919c97bb9270707dfee96e03589ae3bd0474612f23f12b938a91ba95d81cd01R64. 

In some places, I just used `[0; 32]`, as a shortcut. I could use another empty SMT for the root instead, if that would be more clear.

crates/fuel-core/src/database/balances.rs
```diff
@@ -48,37 +59,91 @@ impl StorageMutate<ContractsAssets> for Database {
...
+        // Get latest metadata entry for this contract id
+        let prev_metadata = self
+            .storage::<ContractsAssetsMerkleMetadata>()
+            .get(key.contract_id())?
+            .unwrap_or_default();
+
+        let root = prev_metadata.root;
+        let mut tree: MerkleTree<ContractsAssetsMerkleData, _> = {
+            let storage = self.borrow_mut();
+            if root == [0; 32] {
```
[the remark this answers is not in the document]
bvrooman: It sounds like you're suggesting that we update the functionality of `MerkleTree::load` to something like:
- If the root is the empty root, return a new tree (without loading anything)
- Otherwise, load the existing tree from the root

That could work! I will create a PR for that in fuel-vm to see how that looks. 

However, I would say that this is not an internal detail because this is mandated by the spec: https://github.com/celestiaorg/celestia-specs/blob/master/src/specs/data_structures.md#sparse-merkle-tree. All spec-compliant Merkle trees have this quality.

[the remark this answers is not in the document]
xgreenx: I meant from the `fuel-core` perspective it is internal details of the `fuel-merkle`=)

crates/fuel-core/src/database/balances.rs
```diff
@@ -48,37 +59,91 @@ impl StorageMutate<ContractsAssets> for Database {
...
+            .map_err(Into::into);
+
+        // Get latest metadata entry for this contract id
+        let prev_metadata = self
+            .storage::<ContractsAssetsMerkleMetadata>()
+            .get(key.contract_id())?
+            .unwrap_or_default();
+
+        let root = prev_metadata.root;
+        if root == [0; 32] {
```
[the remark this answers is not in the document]
bvrooman: It sounds like you're suggesting that we check the presence of the returned option, rather than the root. That could work, but there are two concerns that come to mind:
1. When we delete the key from the loaded Merkle tree below, we would have to check the root to see if it is empty. If it is empty, we must delete the Merkle metadata record for this contract. Otherwise, this check returns `Some(...)` rather than `None` in this case.
2. Comparing the root against the empty root is already specification compliant. According to the spec, that is a how we know that the Merkle tree is empty. Checking the DB return value will work too, but I don't know if it's actually simpler in this case.

crates/fuel-core/src/database/balances.rs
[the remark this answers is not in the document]
bvrooman: PR up: https://github.com/FuelLabs/fuel-vm/pull/379. 
This will simplify the above code to something like:
```
let root = prev_metadata.root;
let storage = self.borrow_mut();
let mut tree: MerkleTree<ContractsAssetsMerkleData, _> = MerkleTree::load(storage, root)...
```

crates/fuel-core/src/executor.rs
```diff
@@ -2754,7 +2753,7 @@ mod tests {
             .unwrap();
 
         // Assert the balance and state roots should be the same before and after execution.
-        let empty_state = Bytes32::from(*empty_sum_sha256());
+        let empty_state = [0; 32].into();
```
xgreenx, line 2756: [nit] We can use `MerkleTree::empty_root` instead of `[0; 32].into()`

xgreenx (approved):

crates/fuel-core/src/executor.rs
bvrooman: Unfortunately, we still need to provide the generic arguments to `MerkleTree`, so it turns into something like `MerkleTree<Table, StorageType<Table>>::empty_root()`, which is kinda ugly. That's why I've been using the empty in-memory trees for empty roots for now. But with the recent `fuel-vm` PR, we will be able to just use `sparse::empty_root()`.

xgreenx (approved):
[nit] You also can add a unit test to check that we remove empty metadata=)

Issue #985, opened 2023-02-06
Allow query module to be used directly without any GraphQL


Voxelot:
Currently the query module is private, we should make it possible to use these query apis directly on a FuelService without needing to enable the graphql api.

![image](https://user-images.githubusercontent.com/794823/217043535-e3cc4c1f-4782-4662-8f20-bc63c23207fd.png)

PR #1012, opened 2023-02-14
Prepare GraphQL Crate for extraction

ControlCplusControlV:
This PR works to simplify the `query` interfaces by extracting the logic from all `*contexts` into traits, then refactors all queries to fetch this information via type objects rather than directly through the database, which will allow a clean removal of the graphql crate, where the traits will move to, while the implementations for `Database` remain within `fuel-core`.

There are some small pointer annoyances such as `borrowed_box`, but this is unavoidable because `Database` is passed into the graphql context as a `Box`, and so whenever its accessed the pointer to it by default is `&Box<T>`, and so any function which requires erasing `Box<DatabasePort>` (which is aliased to `Database` with the graphql crate) to something like `Box<dyn MessageProofData>` will inevitable create a borrowed box and so clippy will complain, but this can't really be avoided due to the graphql structure


ControlCplusControlV:
Closes #985

crates/fuel-core/src/graphql_api/service.rs
```diff
@@ -141,20 +142,24 @@ impl RunnableTask for Task {
     }
 }
 
+// Need a seperate Data Object for each Query endpoint, cannot be avoided
 pub fn new_service(
     config: Config,
-    database: Database,
     schema: CoreSchemaBuilder,
-    producer: BlockProducer,
+    query_data: Box<dyn QueryData>,
```
xgreenx, line 149: We don't need to pass any `Box<dyn QueryData>`. The `Database` itself already should implement `QueryData` traits because of your blanket implementation.

And seems `QueryData` should be a super trait for `DatabasePort`.

crates/fuel-core/src/query.rs
```diff
@@ -7,6 +8,21 @@ mod message;
...
+    + BlockQueryData
+    + ChainQueryData
+    + CoinQueryData
+    + ContractQueryData
+    + MessageQueryData
+    + MessageProofData
+    + TransactionQueryData
+{
+}
+impl<D: DatabasePort + 'static> QueryData for D {}
```
xgreenx, line 24:
```suggestion
impl<D> QueryData for D 
where D:
     Send
     + Sync
     + BalanceQueryData
     + BlockQueryData
     + ChainQueryData
     + CoinQueryData
     + ContractQueryData
     + MessageQueryData
     + MessageProofData
     + TransactionQueryData
     {}
```

I guess it should be something like this

crates/fuel-core/src/query/balance/asset_query.rs
```diff
@@ -95,10 +99,9 @@ impl<'a> AssetsQuery<'a> {
     /// # Note: The resources of different type are not grouped by the `asset_id`.
     // TODO: Optimize this by creating an index
     //  https://github.com/FuelLabs/fuel-core/issues/588
-    pub fn unspent_resources(
-        &self,
-    ) -> impl Iterator<Item = StorageResult<Resource>> + '_ {
-        let coins_iter = CoinQueryContext(self.database)
+    pub fn unspent_resources(&self) -> BoxedIter<StorageResult<Resource>> {
```
xgreenx, line 102: The output type should be `impl Iterator<Item = StorageResult<Resource>>`.

```
             .into_boxed()
             .collect::<Vec<_>>()
             .into_iter()
             .into_boxed()
```

Should definitely be re-worked=) I guess the previous implementation should work for you

crates/fuel-core/src/query/contract.rs
```diff
@@ -21,26 +23,35 @@ use fuel_core_types::{
     services::graphql_api::ContractBalance,
 };
 
-pub struct ContractQueryContext<'a>(pub &'a Database);
+pub trait ContractQueryData: Send + Sync {
```
xgreenx, line 26: Could you add `\n` after each method please?=)

crates/fuel-core/src/query/contract.rs
```diff
@@ -81,13 +88,13 @@ impl ContractQueryContext<'_> {
...
+    fn contract_balances(
         &self,
         contract_id: ContractId,
         start_asset: Option<AssetId>,
         direction: IterDirection,
-    ) -> impl Iterator<Item = StorageResult<ContractBalance>> + '_ {
-        self.0
-            .contract_balances(contract_id, start_asset, direction)
+    ) -> BoxedIter<StorageResult<ContractBalance>> {
+        self.contract_balances(contract_id, start_asset, direction)
```
xgreenx, line 97: You don't need to wrap it `into_boxed` because it is already wrapped=) 
image

Could you review all other places with `into_boxed` to avoid double wrapping, please?=)

crates/fuel-core/src/query/message.rs
```diff
@@ -52,65 +56,75 @@ use std::borrow::Cow;
...
 
-pub struct MessageQueryContext<'a>(pub &'a Database);
+pub trait MessageQueryData: Send + Sync {
+    fn message(&self, message_id: &MessageId) -> StorageResult<Message>;
+    fn owned_message_ids(
+        &self,
+        owner: &Address,
+        start_message_id: Option<MessageId>,
+        direction: IterDirection,
+    ) -> BoxedIter<StorageResult<MessageId>>;
```
xgreenx, line 66:
```suggestion
    ) -> BoxedIter<StorageResult<MessageId>>;
    
```

And for other methods too

crates/fuel-core/src/query/message.rs
```diff
@@ -120,54 +134,61 @@ pub trait MessageProofData {
...
     }
 }
 
 /// Generate an output proof.
 // TODO: Do we want to return `Option` here?
-pub fn message_proof<Data: MessageProofData>(
-    data: &Data,
+#[allow(clippy::borrowed_box)]
+pub fn message_proof<'a, T: MessageProofData + ?Sized>(
+    data: &Box<T>,
```
xgreenx, line 185:
```suggestion
pub fn message_proof<'a, T: MessageProofData + ?Sized>(
    data: &T,
```

crates/fuel-core/src/query/message.rs
```diff
@@ -120,54 +134,61 @@ pub trait MessageProofData {
...
         match consensus {
             // TODO: https://github.com/FuelLabs/fuel-core/issues/816
             Consensus::Genesis(_) => Ok(Default::default()),
             Consensus::PoA(c) => Ok(c.signature),
         }
     }
 
-    fn block(&self, block_id: &BlockId) -> StorageResult<CompressedBlock> {
-        BlockQueryContext(self.0).block(block_id)
+    fn msg_receipts(&self, transaction_id: &TxId) -> StorageResult<Vec<Receipt>> {
```
xgreenx, line 168: We don't need to duplicate functionality from `TransactionQueryData`, you simply can remove those function and use analogs from `TransactionQueryData`

crates/fuel-core/src/query/tx.rs
```diff
@@ -22,43 +24,49 @@ use fuel_core_types::{
     services::txpool::TransactionStatus,
 };
 
-pub struct TransactionQueryContext<'a>(pub &'a Database);
+pub trait TransactionQueryData: Send + Sync {
+    fn transaction(&self, tx_id: &TxId) -> StorageResult<Transaction>;
```
xgreenx, line 28:
```suggestion
    fn transaction(&self, tx_id: &TxId) -> StorageResult<Transaction>;
    
```

And other methods

crates/fuel-core/src/resource_query.rs
```diff
@@ -93,6 +93,9 @@ impl SpendQuery {
 
     /// Return [`AssetQuery`]s.
     pub fn asset_queries<'a>(&'a self, db: &'a Database) -> Vec<AssetQuery<'a>> {
+        let db = db.to_owned();
+        let db = db.as_ref();
+
```
xgreenx, line 98:
```suggestion
```

crates/fuel-core/src/resource_query.rs
```diff
@@ -305,16 +304,18 @@ mod tests {
...
-                                    (*resource.asset_id(), *resource.amount())
-                                })
-                                .collect()
-                        },
-                    )
+                    largest_first(&AssetQuery::new(
+                        owner,
+                        asset,
+                        None,
+                        db.to_owned().as_ref(),
```
xgreenx, line 311:
```suggestion
                        db,
```

crates/fuel-core/src/schema.rs
```diff
@@ -1,4 +1,6 @@
-use crate::state::IterDirection;
+#![allow(clippy::borrowed_box)]
+// Boxes are used to house type objects, but when passed through the context they become borrowed
+// so avoiding this lint just results in double references, which is the same as a box
```
xgreenx, line 3:
```suggestion
```

crates/fuel-core/src/schema/balance.rs
```diff
@@ -52,7 +52,7 @@ impl BalanceQuery {
         #[graphql(desc = "address of the owner")] owner: Address,
         #[graphql(desc = "asset_id of the coin")] asset_id: AssetId,
     ) -> async_graphql::Result<Balance> {
-        let data = BalanceQueryContext(ctx.data_unchecked());
+        let data: &Box<dyn QueryData> = ctx.data_unchecked();
```
xgreenx, line 55:
```suggestion
        let data: &Database = ctx.data_unchecked();
```

crates/fuel-core/src/schema/balance.rs
```diff
@@ -74,7 +74,7 @@ impl BalanceQuery {
         if before.is_some() || after.is_some() {
             return Err(anyhow!("pagination is not yet supported").into())
         }
-        let query = BalanceQueryContext(ctx.data_unchecked());
+        let query: &Box<dyn QueryData> = ctx.data_unchecked();
```
xgreenx, line 77:
```suggestion
        let query: &Database = ctx.data_unchecked();
```

crates/fuel-core/src/schema/block.rs
```diff
@@ -89,7 +89,7 @@ impl Block {
     }
 
     async fn consensus(&self, ctx: &Context<'_>) -> async_graphql::Result<Consensus> {
-        let query = BlockQueryContext(ctx.data_unchecked());
+        let query: &Box<dyn QueryData> = ctx.data_unchecked();
```
xgreenx, line 92:
```suggestion
        let query: &Database = ctx.data_unchecked();
```

And the same in other places=)

crates/fuel-core/src/schema/tx.rs
```diff
@@ -262,12 +263,28 @@ impl TxStatusSubscription {
...
+
 #[async_trait::async_trait]
 impl<'a> TxnStatusChangeState for StreamState<'a> {
     async fn get_tx_status(
         &self,
         id: fuel_types::Bytes32,
     ) -> StorageResult<Option<TransactionStatus>> {
-        types::get_tx_status(id, &TransactionQueryContext(self.db), self.txpool).await
+        let query = self.db;
+        match query
```
xgreenx, line 276: Hmm, the previous code with `get_tx_status` was better=) Could we return it back?

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -158,7 +159,8 @@ pub fn init_sub_services(
         config.chain_conf.gas_costs.clone(),
     )
     .data(database.clone());
-    let gql_database = Box::new(database.clone());
+
+    let query_data: Box<dyn QueryData> = Box::new(database.clone());
```
xgreenx, line 177: Yeah, we definitely don't need a separate `query_data` type=) Because it is `Database`

crates/storage/Cargo.toml
```diff
@@ -21,6 +21,7 @@ anyhow = { workspace = true }
 fuel-core-types = { workspace = true }
 fuel-vm-private = { workspace = true }
 mockall = { workspace = true, optional = true }
+rocksdb = { version = "0.19", default-features = false }
```
xgreenx, line 24: Only `fuel-core` should depend on the `rocksdb`. It is better to do this mapping in the `fuel-core`

xgreenx (dismissed):

crates/fuel-core/src/query/balance/asset_query.rs
ControlCplusControlV: The output type of `impl Iterator<Item = StorageResult<Resource>>` can't be used as a return type in a trait definition, and referencing the iterator type directly is nearly impossible because it ends up being long enough that the compiler will dump its type name into a `.txt` file when compiling, is there another way around this?

crates/fuel-core/src/query/contract.rs
ControlCplusControlV: Added :)

crates/fuel-core/src/query/balance/asset_query.rs
xgreenx: It is not a trait, it is a public method=)

ControlCplusControlV: Oh I had this confused from some of the other `impl Iterator`'s which I changed, thanks for fixing this!

crates/fuel-core/src/resource_query.rs
ControlCplusControlV: fixed

crates/fuel-core/src/schema/balance.rs
ControlCplusControlV: Fixed

crates/fuel-core/src/query/message.rs
ControlCplusControlV: Resolved by moving directly to a GQL`Database` reference, sized issues among other things stopped me from doing just `&T`

crates/fuel-core/src/query/message.rs
ControlCplusControlV: It breaks a test which depends on `mockall` to gain access to these functions, but will refactor those tests

crates/storage/Cargo.toml
ControlCplusControlV: Made a wrapper type and fixed it

crates/fuel-core/src/graphql_api/service.rs
ControlCplusControlV: Removed QueryData

crates/fuel-core/src/schema/tx.rs
ControlCplusControlV: moved back

deekerno (commented):
LGTM, but I'll defer to reviewers with more experience in this codebase.

crates/fuel-core/src/graphql_api.rs
```diff
@@ -12,6 +12,9 @@ use std::net::SocketAddr;
 pub mod ports;
 pub mod service;
 
+// #[cfg(test)]
+// pub mod test_helpers;
```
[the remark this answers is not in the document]
ControlCplusControlV: Fixed, good catch

crates/fuel-core/src/schema/tx.rs
```diff
@@ -267,6 +269,7 @@ impl<'a> TxnStatusChangeState for StreamState<'a> {
         &self,
         id: fuel_types::Bytes32,
     ) -> StorageResult<Option<TransactionStatus>> {
-        types::get_tx_status(id, &TransactionQueryContext(self.db), self.txpool).await
+        let query = self.db;
+        types::get_tx_status(id, query, self.txpool).await
```
xgreenx, line 273:
```suggestion
        types::get_tx_status(id, self.db, self.txpool).await
```

crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -217,15 +218,17 @@ impl KeyValueStore for RocksDb {
             }
             (Some(prefix), None) => {
                 // start iterating in a certain direction within the keyspace
-                let iter_mode = IteratorMode::From(prefix, direction.into());
+                let iter_mode =
+                    IteratorMode::From(prefix, DatabaseIterDirection(direction).into());
```
xgreenx, line 222: If you use `DatabaseIterDirection` only here, it is better not to create this type and replace converting by some funciton=)

crates/fuel-core/src/schema/tx/types.rs
```diff
@@ -396,16 +400,27 @@ impl Transaction {
         ctx: &Context<'_>,
     ) -> async_graphql::Result<Option<TransactionStatus>> {
         let id = self.0.id();
-        let query = TransactionQueryContext(ctx.data_unchecked());
+        let query: &Database = ctx.data_unchecked();
         let txpool = ctx.data_unchecked::<TxPool>();
-        get_tx_status(id, &query, txpool).await.map_err(Into::into)
+        match query
```
xgreenx, line 405: Why did you extract this functionality form the `get_tx_status` method?

ControlCplusControlV: There was a brief lifetime error where it wouldn't pass in that I fixed but forgot to change this back

crates/fuel-core/src/query/balance.rs
```diff
@@ -23,15 +27,24 @@ use std::{
...
-impl BalanceQueryContext<'_> {
-    pub fn balance(
+    fn balances(
+        &self,
+        owner: Address,
+        direction: IterDirection,
+    ) -> BoxedIter<StorageResult<AddressBalance>>;
+}
+
+impl BalanceQueryData for Box<dyn DatabasePort> {
```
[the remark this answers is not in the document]
ControlCplusControlV: This is unfortunately blocked by
```
        let amount = AssetQuery::new(
            &owner,
            &AssetSpendTarget::new(asset_id, u64::MAX, u64::MAX),
            None,
            self,
        )
 ```
 
 Just a couple lines below, as `self` cannot be passed alone because it is explicitly not sized, and even when trying to pass it as a reference I kept running into multiple sizing issues, and a Box pointer was the most simple way around this, if you do have a way that works locally feel free to commit it directly but I tried to get this to work both now and when writing it, and the sizing constraint blocks it from working

xgreenx (approved):

PR #1015, opened 2023-02-14
Fix database version serialization

Voxelot:
The database version was being double initialized between `Database::init` and `RocksDB`. However, the `Database` was using postcard, and `RocksDB` was using manual byte serialization. This was causing version conflicts when restarting a fuel-core node off of an existing database, as the serialization methods were mismatched.

This PR resolves the issue by ensuring the database metadata is only managed by the database itself, and not by RocksDb.

This is a blocker for beta3.

### DIFF
--- a/crates/fuel-core/src/database/metadata.rs
+++ b/crates/fuel-core/src/database/metadata.rs
@@ -10,22 +10,35 @@ pub(crate) const DB_VERSION_KEY: &[u8] = b"version";
 pub(crate) const CHAIN_NAME_KEY: &[u8] = b"chain_name";
 
 /// Can be used to perform migrations in the future.
-pub(crate) const DB_VERSION: u32 = 0;
+pub(crate) const DB_VERSION: u32 = 0x00;
 
 impl Database {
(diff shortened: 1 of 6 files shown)


crates/fuel-core/src/database/metadata.rs
```diff
@@ -10,22 +10,35 @@ pub(crate) const DB_VERSION_KEY: &[u8] = b"version";
...
+                        Err(DatabaseError::ChainAlreadyInitialized)
+                    } else {
+                        Ok(())
+                    }
+                })?;
+        }
 
-        let _: Option<u32> =
-            self.insert(DB_VERSION_KEY, Column::Metadata, &DB_VERSION)?;
+        // Ensure the database version is correct
```
freesig, line 30: Guessing no race condition is possible here?

freesig (approved):
Looks good

crates/fuel-core/src/database/metadata.rs
Voxelot: there shouldn't be, as the database uses a lockfile.

Issue #949, opened 2023-01-26
Graceful Shutdown for all services


Voxelot:
Add the support of graceful shutdown and panic handling for all services.

- [ ] https://github.com/FuelLabs/fuel-core/issues/897
- [ ] Other services

Issue #897, opened 2023-01-11
Support graceful shutdown for `RelayerService`


xgreenx:
<img width="1216" alt="image" src="https://user-images.githubusercontent.com/18346821/211699209-8c7d6080-26f4-4290-a897-e5c54b2dcc3f.png">

Without graceful shutdown support, waiting for the service to stop during the synchronization process will take a lot of time.
During this task, also add cover it with tests.

PR #1018, opened 2023-02-15
Graceful Shutdown for all services

xgreenx:
- Closes https://github.com/FuelLabs/fuel-core/issues/949
- Closes https://github.com/FuelLabs/fuel-core/issues/897

Removed second spawned task for `fuel_core_services::service::run` function.

Instead, wrapped it into `catch_unwind`.
Added a `shutdown` method for the task that allows doing some cleanup for methods. For, almost, all services, it does nothing, but can be reused in the future. Added handling of the shutdown for `Relayer` to act immediately instead of waiting for the synchronization's end. Improved handling of the stop signal for `fuel-core-sync`.

### DIFF
--- a/crates/services/src/service.rs
+++ b/crates/services/src/service.rs
@@ -3,10 +3,8 @@ use crate::state::{
     StateWatcher,
 };
 use anyhow::anyhow;
-use tokio::{
-    sync::watch,
-    task::JoinHandle,
-};
+use futures::FutureExt;
+use tokio::sync::watch;
 use tracing::Instrument;
 
 /// Alias for Arc<T>
@@ -53,6 +51,9 @@ pub trait Service {
 
     /// The current state of the service (i.e. `Started`, `Stopped`, etc..)
     fn state(&self) -> State;
+
+    /// Returns the state watcher of the service.
+    fn state_watcher(&self) -> StateWatcher;
 }
 
 /// Trait used by `ServiceRunner` to encapsulate the business logic tasks for a service.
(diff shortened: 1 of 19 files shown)


crates/services/sync/src/service/tests.rs
```diff
@@ -60,12 +61,15 @@ async fn test_new_service() {
...
     );
+    let mut last_value = 0;
     while let Some(h) = rx.recv().await {
+        last_value = h;
         if h == 16 {
             break
         }
     }
 
+    assert_eq!(last_value, 16);
```
freesig, line 72: The loop only breaks on 16 so this is always true

crates/services/src/service.rs
```diff
@@ -228,12 +236,13 @@ where
     tokio::task::spawn(
         async move {
             tracing::debug!("running");
-            let join_handler = run(service, stop_sender.clone());
+            let run = std::panic::AssertUnwindSafe(run(service, stop_sender.clone()));
             tracing::debug!("awaiting run");
-            let result = join_handler.await;
+            let result = run.catch_unwind().await;
```
freesig, line 241: Doesn't this panic need to be resumed?

freesig (changes requested):

crates/services/sync/src/service/tests.rs
xgreenx: Previously the loop didn't work=) It was `None` because the import task failed at the start.

freesig: ah I see

crates/services/src/service.rs
xgreenx: It is useless because we have an anonymous task. But we can do that. Why not=)

Voxelot (approved):

crates/services/src/service.rs
freesig: It just means drops are called for the task

freesig (approved):

PR #1025, opened 2023-02-16
Add RUST_LOG option to deployment charts & increase max_transmit_size

Voxelot:
- add rust_log option to deployment charts
- increase default max_transmit_size to match our load balancer limit


xgreenx (approved):

Voxelot:
verified this setting works:
image

PR #1027, opened 2023-02-17
Disable pagination args for balances

Voxelot:
Throw an error when attempting to use args like `before` and `after` for balance queries since they aren't used and are not currently supported.

Also updated the integ test for balances to setup assets for multiple owners to ensure the query is able to precisely select a particular owner.


ControlCplusControlV:
Shouldn't cause too many merge conflicts with my PRs landing this weekend

ControlCplusControlV (approved):

PR #1028, opened 2023-02-17
peer reputation with separate app & gossipsub score

leviathanbeak:
The approach is mostly detailed in [Notion](https://www.notion.so/fuellabs/New-Proposal-96da9de6609b4faf94c30ae6bd355bc4).

These are the first steps of our Peer Reputation approach, there are certain _todos_ left, could be added later on or if needed I can add them with this PR.

Todo:
- [x] peer reputation decay with time
- [ ] implement unban peers after certain period
- [x] reserved peers should not be banned/scored ? - this is kind of "implemented" by only updating the score of non-reserved peers

Ref https://github.com/FuelLabs/fuel-core/issues/943


leviathanbeak:
Note: the scores/config/params for both, gossipsub and app scoring are **not** finalized, this just the approach, the barebones for our Peer Reputation. We can discuss and (later?) adjust all the parameters however we decide it's for the best

ControlCplusControlV:
Can review once CI is passing again

leviathanbeak:
Added peer reputation decay + some refactors regarding the flow of peer reporting

crates/services/p2p/src/peer_manager.rs
```diff
@@ -160,6 +186,41 @@ impl PeerManagerBehaviour {
...
+            .peer_manager
+            .non_reserved_connected_peers
+            .contains_key(peer_id)
+            && gossip_score
+                < self
+                    .peer_manager
+                    .peer_score_config
+                    .get_min_gossipsub_score()
+        {
+            self.peer_manager
```
xgreenx, line 219: After today's discussion, we agreed that it is worth considering moving the `PeerManager` logic above to the `P2PService` level or upper to disconnect immediately. Also, it will reduce the exact method duplication on each level(like triple `report_peer` and so on).

crates/services/p2p/src/peer_manager.rs
```diff
@@ -553,6 +633,41 @@ impl PeerManager {
         }
     }
 
+    fn batch_update_score(&mut self, score: PeerScore) {
+        for (peer_id, peer_info) in &mut self.non_reserved_connected_peers {
+            let new_score = peer_info.score + score;
```
xgreenx, line 638: Instead of `peer_info.score + score` maybe it worth to do something like `peer_info.score * DECAY_FACTOR` where `DECAY_FACTOR < 1`. In this case, it will be not possible to disconnect during decaying=)

crates/services/p2p/src/service.rs
```diff
@@ -222,6 +232,9 @@ where
                     Some(TaskRequest::RespondWithGossipsubMessageReport((message, acceptance))) => {
                         report_message(&mut self.p2p_service, message, acceptance);
                     }
+                    Some(TaskRequest::RespondWithPeerReport { peer_id, peer_score, reporting_service }) => {
```
xgreenx, line 235: It would be cool to make gossiping score independent. The current implementation updates the score and decides whether we want to propagate the transaction through gossiping or not by this report. But it would be nice to make those two processes independent because the gossiping score should be related only to the validity of messages and meshing.

crates/types/src/services/p2p/peer_reputation.rs
```diff
@@ -0,0 +1,56 @@
+/// PeerScore type used for Peer Reputation
+pub type PeerScore = f64;
+
+/// Minimum allowed peer score before peer is banned
+pub const MIN_PEER_SCORE: PeerScore = 0.0;
+/// Default value for peer score
+pub const DEFAULT_PEER_SCORE: PeerScore = 50.0;
```
xgreenx, line 7: I think it would be better to use negative values for bad reputation and `0` as a default score for all.

crates/types/src/services/p2p/peer_reputation.rs
```diff
@@ -0,0 +1,56 @@
+/// PeerScore type used for Peer Reputation
+pub type PeerScore = f64;
```
xgreenx, line 2: It would be nice to have its own type for the application score and maybe `i32` or `i64` would be enough for us=)

The `PeerScore` is good naming for the application score and gossiping score. But if we make the gossiping score independent, then we need only `ApplicationScore`

crates/services/p2p/src/service.rs
leviathanbeak: `RespondWithGossipsubMessageReport` deals with gossipsub,
`RespondWithPeerReport ` deals with application specific score (ie everything else except gossipsub)

crates/types/src/services/p2p/peer_reputation.rs
leviathanbeak: Sure, I will update it, but either way we will have to do another deep dive on scoring soon.

crates/types/src/services/p2p/peer_reputation.rs
leviathanbeak: good point, `AppScore` is clearer name. I left `f64` type since gossipsub uses the same, and we can decay the score by ```current_score * 0.9``` as we discussed

crates/services/p2p/src/peer_report.rs
```diff
@@ -0,0 +1,452 @@
...
+        match event {
+            FromSwarm::ConnectionEstablished(connection_established) => {
+                let ConnectionEstablished {
+                    peer_id,
+                    other_established,
+                    ..
+                } = connection_established;
+
+                self.heartbeat
+                    .on_swarm_event(FromSwarm::ConnectionEstablished(
```
freesig, line 156: Most of these functions are just forwarding events. A helper function could reduce duplication

crates/services/p2p/src/peer_report.rs
```diff
@@ -0,0 +1,452 @@
...
+                                agent_version,
+                                addresses: listen_addrs,
+                            };
+
+                            return Poll::Ready(NetworkBehaviourAction::GenerateEvent(
+                                event,
+                            ))
+                        }
+                        IdentifyEvent::Error { peer_id, error } => {
+                            debug!(target: "fuel-p2p", "Identification with peer {:?} failed => {}", peer_id, error)
```
freesig, line 413: It's not a good idea to set the target to the wrong thing. The target should be this module not `fuel-core`

crates/services/p2p/src/peer_report.rs
leviathanbeak: All of these events need to be extracted separately and passed down to two different behaviours.

I don't see a way to reduce "duplication" since all events are different, this is caused because parent `ConnectionHandler` is different than the children behaviours'. So it first needs to be extracted and sent down separately to the children.

crates/services/p2p/src/peer_manager.rs
```diff
@@ -553,115 +86,127 @@ impl PeerManager {
...
+            peer.score = new_score;
+
+            let should_ban = new_score < self.peer_score_config.min_score_allowed;
+
+            Some(PeerScoreUpdated {
+                should_ban,
+                score: new_score,
+            })
+        } else {
+            log_missing_peer(&peer_id);
```
Voxelot, line 158: This log message might be confusing if the peer is connected as a reserved peer.

crates/services/p2p/src/peer_report.rs
```diff
@@ -0,0 +1,452 @@
...
+};
+
+use tracing::debug;
+
+/// Maximum amount of peer's addresses that we are ready to store per peer
+const MAX_IDENTIFY_ADDRESSES: usize = 10;
+const HEALTH_CHECK_INTERVAL_IN_SECONDS: u64 = 10;
+const REPUTATION_DECAY_INTERVAL_IN_SECONDS: u64 = 1;
+
+/// Events emitted by PeerInfoBehaviour
```
Voxelot, line 59:
```suggestion
/// Events emitted by PeerReportBehavior
```

crates/services/p2p/src/peer_report.rs
```diff
@@ -0,0 +1,452 @@
...
+        block_height: BlockHeight,
+    },
+    /// Informs p2p service / PeerManager to check health of reserved nodes' connections
+    CheckReservedNodesHealth,
+    /// Informs p2p service / PeerManager to perform reputation decay of connected nodes
+    PerformDecay,
+}
+
+// `Behaviour` that reports events about peers
+pub struct PeerReportBehaviour {
```
Voxelot, line 86: This seems a little confusing to call this peer report behavior, as it implies this service handles reputation. However, it seems like it handles nearly all the info for each peer.

crates/services/p2p/src/peer_report.rs
Voxelot: Made a draft PR here which extracts these conversions out into separate functions: https://github.com/FuelLabs/fuel-core/pull/1109

For the events that are simply pass-throughs, I wonder if the helper methods like `.map_handler` and `.map_out` could be used 🤔

leviathanbeak: That looks good, but I understood that freesig meant de-duplicating forwarding of events in `on_swarm_event`

crates/services/p2p/src/peer_manager.rs
leviathanbeak: If you meant specifically in the context of `update_peer_score_with()`, that method works only with `non_reserved_connected_peers` HashMap so no reserved peer will be logged from here.

If you meant `log_missing_peer` in general, the (reserved) peer could still not be connected at a given point when an action was tried, so it logs that the peer is missing/disconnected.

crates/services/p2p/src/peer_report.rs
leviathanbeak: Good point, I mostly aimed at "peer reporting" as in report peer events, ie disconnected, connected, identified etc. I am definitely open to a better name

crates/services/p2p/src/peer_report.rs
leviathanbeak: @Voxelot here's my take at the `on_swarm_event` de-dup

crates/services/p2p/src/service.rs
```diff
@@ -425,6 +434,32 @@ impl SharedState {
...
+
+    pub fn report_peer<T: PeerReport>(
+        &self,
+        peer_id: FuelPeerId,
+        peer_report: T,
+        reporting_service: &'static str,
+    ) -> anyhow::Result<()> {
+        match Vec::from(peer_id).try_into() {
+            Ok(peer_id) => {
+                let peer_score = peer_report.get_score_from_report();
```
xgreenx, line 446: I guess it should be `app_score` otherwise, it confuses=)

crates/services/p2p/src/service.rs
```diff
@@ -425,6 +434,32 @@ impl SharedState {
...
+
+                self.request_sender
+                    .try_send(TaskRequest::RespondWithPeerReport {
```
xgreenx, line 449: What do you think about renaming `RespondWithPeerReport` -> `ApplicationReport`?

crates/services/p2p/src/peer_manager.rs
```diff
@@ -553,115 +86,127 @@ impl PeerManager {
...
+    pub fn batch_update_score_with_decay(&mut self) {
+        for peer_info in self.non_reserved_connected_peers.values_mut() {
+            peer_info.score *= DECAY_APP_SCORE;
+        }
+    }
+
+    pub fn update_peer_score_with(
+        &mut self,
+        peer_id: PeerId,
+        score: AppScore,
```
xgreenx, line 144: Why do we update peer score(the name of the method) with `AppScore`?=) I think all external methods should work with app score because peer score is calculated internally in the behavior

crates/services/p2p/src/service.rs
leviathanbeak: I'd prefer keep `RespondWith` and as discussed, the report is about the peer,  `ApplicationReport` would suggest otherwise

crates/services/p2p/src/p2p_service.rs
```diff
@@ -311,21 +344,53 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
...
+    /// If application peer score is below allowed threshold
+    /// the peer is banend
+    pub fn report_peer(
+        &mut self,
+        peer_id: PeerId,
+        app_score: AppScore,
+        reporting_service: &str,
+    ) {
+        if let Some(PeerScoreUpdated { should_ban, score }) =
+            self.peer_manager.update_app_score(peer_id, app_score)
```
xgreenx, line 386: I think we don't need to hesitate and don't pass the `&mut self.swarm` into `update_app_score`. If you pass it there, you can remove `PeerScoreUpdated` and simplify the logic. It is also applyable to other places that return `should_ban`

leviathanbeak: resolved with #1118

Voxelot (approved):

PR #1029, opened 2023-02-17
Add e2e test to check that Bob can send money back to Alice

xgreenx:
Added `transfer_back` test to check that Bob can receive money back from Alice.
Subscribed for `TxPool` instead of block importer.


Voxelot (approved):

Voxelot:
looks like we have a clippy error

Voxelot (approved):

PR #1030, opened 2023-02-17
Predicate error handling

Voxelot:
- Added tests to ensure errors like `PredicateExhaustedGas` are correctly returned by the txpool
- Removed duplicate predicate verification
- Removed async from tests that didn't need it
- Removed tests related to predicates being enabled or disabled since they are always enabled.

### DIFF
--- a/crates/services/txpool/src/txpool.rs
+++ b/crates/services/txpool/src/txpool.rs
@@ -9,7 +9,6 @@ use crate::{
     Config,
     Error,
 };
-use anyhow::anyhow;
 use fuel_core_metrics::txpool_metrics::TXPOOL_METRICS;
 use fuel_core_types::{
     blockchain::SealedBlock,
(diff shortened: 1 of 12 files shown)


crates/services/txpool/src/txpool.rs
```diff
@@ -122,14 +121,6 @@ where
             .into())
         }
 
-        // verify predicates
-        if !tx.check_predicates(
```
xgreenx, line 126: Oh, yeah, not we verify predicates in the `into_checked`

crates/services/txpool/src/txpool/tests.rs
```diff
@@ -42,8 +50,8 @@ use std::{
 
 const GAS_LIMIT: Word = 1000;
 
-#[tokio::test]
-async fn insert_simple_tx_succeeds() {
+#[test]
+fn insert_simple_tx_succeeds() {
```
xgreenx, line 54: Nice, we can remove it because we did the refactoring in the `TxPool`. Previouslt `insert_inner` was `async`=)

xgreenx (approved):

ControlCplusControlV (approved):

PR #1033, opened 2023-02-20
Pruning of the transaction by TTL

xgreenx:
Added basic pruning of the transaction by the TTL.

It is done periodically for constant period of time. We can do it often(based on the remaining time to the next timeout), but I tried to avoid to often locking.

### DIFF
--- a/crates/services/txpool/Cargo.toml
+++ b/crates/services/txpool/Cargo.toml
@@ -29,6 +29,10 @@ fuel-core-txpool = { path = "", features = ["test-helpers"] }
 itertools = { workspace = true }
 mockall = { workspace = true }
 rstest = "0.15"
+tokio = { workspace = true, features = [
+    "sync",
+    "test-util",
+] }
 
 [features]
 test-helpers = [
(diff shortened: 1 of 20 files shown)


crates/services/txpool/Cargo.toml
```diff
@@ -29,6 +29,10 @@ fuel-core-txpool = { path = "", features = ["test-helpers"] }
 itertools = { workspace = true }
 mockall = { workspace = true }
 rstest = "0.15"
+tokio = { workspace = true, features = [
+    "sync",
+    "test-util",
```
Voxelot, line 34: this feature should probably be only enabled in dev-deps right?

crates/types/src/services/txpool.rs
```diff
@@ -302,6 +267,8 @@ pub enum Error {
     // small todo for now it can pass but in future we should include better messages
     #[error("Transaction removed.")]
     Removed,
+    #[error("Transaction removed because of the TTL.")]
```
Voxelot, line 270:
```suggestion
    #[error("Transaction expired because it exceeded the configured time to live `tx-pool-ttl`.")]
```

Voxelot (approved):
Nice usage of tokio time to make testing easier!

crates/services/txpool/Cargo.toml
Voxelot: nvm it already is

Voxelot (approved):

PR #1050, opened 2023-02-27
Several new configurations for the `rustfmt.toml`

xgreenx:
Added several new options for the `rustfmt.toml`, because I'm trying to keep the code in this style without them. Why not to automate them=D

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,5 +1,6 @@
 max_width = 90 # changed
 comment_width = 90 # changed
+wrap_comments = true # changed
 normalize_comments = true # changed
 imports_layout = "Vertical" # changed
 imports_granularity = "Crate" # changed
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,4 +1,5 @@
 max_width = 90 # changed
+comment_width = 90 # changed
 normalize_comments = true # changed
 imports_layout = "Vertical" # changed
 imports_granularity = "Crate" # changed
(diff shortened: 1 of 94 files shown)


.rustfmt.toml
```diff
@@ -1,8 +1,12 @@
...
 normalize_comments = true # changed
 imports_layout = "Vertical" # changed
 imports_granularity = "Crate" # changed
 trailing_semicolon = false # changed
 edition = "2021" # changed
 use_try_shorthand = true # changed
-use_field_init_shorthand = true # changed
\ No newline at end of file
+use_field_init_shorthand = true # changed
+reorder_impl_items = true # changed
```
xgreenx, line 11: Ohh, wait, it "Reorder impl items. type and const are put first, then macros and methods.".

I thought it would do the same order as in the trait definition. Then maybe it is better to remove it=) But if someone likes it, we can keep it=)

.rustfmt.toml
```diff
@@ -1,8 +1,13 @@
...
+normalize_doc_attributes = true # changed
+format_macro_matchers = true # changed
```
ControlCplusControlV, line 13: Not looking to start a war, but may I suggest
```suggestion
format_macro_matchers = true # changed
hard_tabs = true
```

ControlCplusControlV: It makes keyboard navigation a bit easier and allows easier tuning in the IDE of spacing

xgreenx: I don't know the preferences of others, but I like four spaces more=) It looks better in GitHub than tabs

PR #1051, opened 2023-03-03
WIP: multi-arch build

tirkesi:
I was able to build an arm64 image using docker buildx with the following steps locally while in the root of the fuel-core project

https://cloud.google.com/kubernetes-engine/docs/how-to/build-multi-arch-for-arm

docker buildx create --name mybuilder --use --bootstrap

docker buildx build --platform linux/arm64 -t fuel-core:buildx-latest . -f deployment/Dockerfile --load

```
docker images
REPOSITORY                                TAG               IMAGE ID       CREATED          SIZE
fuel-core                                 buildx-latest     61af27bfabe2   27 minutes ago   99.6MB
```

I still have some testing to do to make sure this works.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -215,20 +215,24 @@ jobs:
           flavor: |
             latest=${{ github.ref == 'refs/heads/master' }}
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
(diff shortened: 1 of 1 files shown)


tirkesi:
There's a memory issue with buildx.

Need to look into rust cross compile.

Issue #544, opened 2022-08-10
Set TxPointer field while producing blocks


Voxelot:
The TxPointer field on Coin and Contract input types needs to be set to the correct block height and idx that corresponds to the utxo id. 

In block validation mode, the TxPointer field should be verified against the expected value.

This logic would happen inside of the executor.

PR #1054, opened 2023-03-08
Set tx pointer during block production

Voxelot:
closes: [#544](https://github.com/FuelLabs/fuel-core/issues/544)

- adds logic to the executor for setting / verifying tx_pointers
- adds tx pointer to chain config for coins and contracts
- adds utxo id to chain config for contracts


crates/storage/src/tables.rs
```diff
@@ -54,7 +54,7 @@ impl Mappable for ContractsLatestUtxo {
     type OwnedKey = ContractId;
     /// The latest UTXO id.
     type Value = Self::OwnedValue;
-    type OwnedValue = UtxoId;
+    type OwnedValue = (UtxoId, fuel_core_types::fuel_tx::TxPointer);
```
xgreenx, line 57: Let's replace it with a new type like `ContractsBalance` or `ContractsCoin` or `ContractsState`

crates/storage/src/tables.rs
```diff
@@ -54,7 +54,7 @@ impl Mappable for ContractsLatestUtxo {
     type OwnedKey = ContractId;
     /// The latest UTXO id.
```
xgreenx, line 55: Seems, it is not actual anymore=)

crates/types/src/entities/coin.rs
```diff
@@ -30,8 +30,10 @@ pub struct Coin {
     pub maturity: BlockHeight,
     /// Whether a coin has been spent or not
     pub status: CoinStatus,
-    /// Which block this coin was created in
+    /// The height of the block this coin was created in
```
xgreenx, line 33: Why we don't want to use `TxPointer` here? Because it will require modification of a lot of code?

crates/fuel-core/src/service/genesis.rs
```diff
@@ -222,6 +224,37 @@ fn init_contracts(
...
+                                        .into_iter(),
+                                )
+                                .collect_vec()
+                                .as_slice(),
+                        )
+                        .expect("Incorrect genesis transaction id byte length"),
+                        generated_output_index as u8,
+                    )
+                };
+                let tx_pointer = if let (Some(block_height), Some(tx_idx)) = (
```
xgreenx, line 249: I'm not sure how we plan to use `TxPointer` in the fork, because it is hard to verify transactions from the previous version of the network. But I guess it is okay to set it=)

Maybe we need to check that the `TxPointer.block_height <= Genesis.block_height`, otherwise it could create situations when coins from the future=)

crates/fuel-core/src/executor.rs
```diff
@@ -979,21 +962,47 @@ where
...
+                            ref mut tx_pointer,
                             ref contract_id,
                             ..
                         } => {
                             let mut contract = ContractRef::new(&mut *db, *contract_id);
-                            *utxo_id =
+                            (*utxo_id, *tx_pointer) =
                                 contract.validated_utxo(self.config.utxo_validation)?;
                             *balance_root = contract.balance_root()?;
                             *state_root = contract.state_root()?;
```
xgreenx, line 1008: Please feel free to remove the comment below this line(I can't leave a comment there).

crates/fuel-core/src/executor.rs
```diff
@@ -908,29 +903,17 @@ where
...
-                        },
+                    let mut coin = self.get_coin_or_default(
+                        db,
+                        *utxo_id,
+                        *owner,
+                        *amount,
+                        *asset_id,
+                        (*maturity).into(),
                     )?;
+                    coin.status = CoinStatus::Spent;
```
xgreenx, line 914: It is not related to this PR, but maybe we want to do the same as we did for the message - a separate column to track whether is coin spent or not.

crates/chain-config/src/serialization.rs
```diff
@@ -59,12 +89,99 @@ impl<'de> DeserializeAs<'de, Word> for HexNumber {
     }
 }
 
+impl<'de> DeserializeAs<'de, u32> for HexNumber {
```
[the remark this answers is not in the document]
xgreenx: And the same for `SerializeAs<u8>`

crates/fuel-core/src/service/genesis.rs
Voxelot: > I'm not sure how we plan to use TxPointer in the fork, because it is hard to verify transactions from the previous version of the network. But I guess it is okay to set it=)

The txs might not be available on the hard-fork upgrade chain, but assuming we have DA for blocks they would still be useful for fraudproofs of older blocks.

crates/fuel-core/src/executor.rs
Voxelot: done

crates/chain-config/src/serialization.rs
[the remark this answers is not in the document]
Voxelot: done

crates/storage/src/tables.rs
Voxelot: done

crates/storage/src/tables.rs
Voxelot: removed

crates/types/src/entities/coin.rs
Voxelot: switched to tx_pointer

xgreenx (approved):

Issue #1053, opened 2023-03-07
Prune spent coins from OwnedCoins column


Voxelot:
The explorer doesn't need to query previously spent coins by owner, so we should prune spent coins from this index to speed up coin selection.

xgreenx:
But we support the `coins` request that returns spent and unspent coins. It means we do not need to support this endpoint.

Voxelot:
Agreed, we should only return unspent coins from this endpoint.

Voxelot:
Actually we already filter that API to only return unspent coins in beta2.

xgreenx:
Yeah, I changed the behavior to return all coins for beta3 because SDK expected all coins.

I think we need to rename it to `unspent_coins` and prune also spent messages.

Voxelot:
We should probably also make this as a patch for beta3 to avoid perf issues there.

PR #1055, opened 2023-03-08
Prune owned coin idx when inputs are spent

Voxelot:
fixes: #1053

Remove owned coin index entries when coins are spent to improve the performance of random-improve based coin selection.

Already released v0.17.4 hotfix from this branch, ready to merge into mainline now.


crates/fuel-core/src/executor.rs
```diff
@@ -931,6 +933,10 @@ where
                             block_created,
                         },
                     )?;
+                    // Remove spent coin from owned index
+                    // TODO: avoid updating coin owners index when updating the coin status above
+                    db.storage::<OwnedCoins>()
```
xgreenx, line 938: I can't find a code where we insert values into the `OwnedCoins` table.

xgreenx: Oh, we already had this column. Could you update this code as well, please?

image

xgreenx (commented):
Could you also update all internal functions to indicate that we return `Unspent` coins?
Like `unspent_owned_coins_ids` instead of `owned_coins_ids`.

Also, in the code, we will fetch coins and try to filter them based on the status, but now we can do it without filtering.

In the main PR for the next release(0.18.0), we also need to update the endpoint to `unspent_coins` instead of `coins`. 

P.S. We also need to inform SDK and the application team because it could affect them.

crates/fuel-core/src/executor.rs
```diff
@@ -881,60 +878,23 @@ where
...
-                    )?;
+                Input::CoinSigned { utxo_id, .. }
+                | Input::CoinPredicate { utxo_id, .. } => {
+                    // prune utxo from db
+                    db.storage::<Coins>().remove(utxo_id)?;
                 }
                 Input::MessageSigned { message_id, .. }
                 | Input::MessagePredicate { message_id, .. } => {
+                    // mark message id as spent
                     db.storage::<SpentMessages>().insert(message_id, &())?;
```
xgreenx, line 895: Could you double-check here that the return type is not `Some`? I know that we do this check at the beginning of the execution(that the message is not spent), but it is cheap also to check it here to be sure(maybe someone in the future will remove the check)=)

crates/storage/src/tables.rs
```diff
@@ -90,6 +91,18 @@ impl Mappable for Coins {
     type OwnedValue = CompressedCoin;
 }
 
+/// The storage table of owned coin ids. Maps addresses to owned coins.
+pub struct OwnedCoins;
```
xgreenx, line 95: We can define this table in the `fuel-core` because it is used only internally.

xgreenx (commented):
We can remove the `status` field from `CompressedCoin` now=)

xgreenx (dismissed):
Removing this field will break the database for beta 3=) Let's remove it in the main PR

xgreenx (approved):

crates/fuel-core/src/executor.rs
```diff
@@ -893,37 +894,22 @@ where
...
-                } => {
-                    let mut coin = self.get_coin_or_default(
-                        db,
-                        *utxo_id,
-                        *owner,
-                        *amount,
-                        *asset_id,
-                        (*maturity).into(),
-                    )?;
-                    coin.status = CoinStatus::Spent;
```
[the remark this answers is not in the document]
xgreenx: And in some places, we still try to fetch `MessageStatus`, but we know that all messages from our database are unspent. Could you also update those places, please?

Voxelot:
Removed spent status from messages and coins. I'm hesitant to rename all the apis to `unspent_`, since in most utxo based chains it's expected that any utxos returned from the state set are unspent.

xgreenx:
After removing of `Spent` status setting, we don't need to rename with `unspent_` prefix=)

crates/client/assets/schema.sdl
```diff
@@ -111,10 +111,6 @@ type Coin {
 	assetId: AssetId!
 	maturity: U64!
 	"""
-	The spendable status of the coin
-	"""
-	status: CoinStatus!
```
xgreenx, line 116: From an API perspective, we still need to return the status of the coins and messages. Because we plan to move this API to the indexer and they can fetch this information. It is why I said to remove it only from "Compressed" types=)

Voxelot: Hmm maybe, although no one was using this part of the API and so it's possible the indexer may not need to track spent coins either. I'd rather rip the bandaid fully off now, and add the status back later if there's an actual requirement for it.

xgreenx (approved):

PR #1056, opened 2023-03-08
Move out `peer_manager` from behaviour up to `p2p_service`

leviathanbeak:
In this _PR to PR_ I moved out our inner `peer_manager` logic from `PeerInfoBehaviour` to top of the p2p service, so that behaviour is not tasked with handling logic and decisions about peers, it just reports about peer events.

So I renamed previously named `PeerManagerBehaviour` to `PeerReportBehaviour`
and left `PeerManager` struct inside `peer_manager.rs` consequently this messes up the **diffs be warned**.

Although moving `peer_manager` up the stack cleaned up a lot of code and was a good decision, someone still had to poll `decay` and `health_check` intervals to inform p2p_service/peer_manager to perform the mentioned tasks, I had left those events inside `PeerReportBehaviour` as for now at least, until maybe we refactor some other parts and make it clearer where/who should perform the polling.


crates/services/p2p/src/peer_report.rs
```diff
@@ -0,0 +1,452 @@
...
+            self.heartbeat.new_handler(),
+            self.identify.new_handler(),
+        )
+    }
+
+    fn addresses_of_peer(&mut self, peer_id: &PeerId) -> Vec<Multiaddr> {
+        self.identify.addresses_of_peer(peer_id)
+    }
+
+    fn on_swarm_event(&mut self, event: FromSwarm<Self::ConnectionHandler>) {
```
xgreenx, line 146: Cool, now this method and the file itself contain only propagation events logic=)

I don't know when `on_swarm_event` is called. Maybe in the future, we also will be able to move the handling of some events directly into `PeerManager`, but it is already good for now=)

xgreenx (commented):
Thank you for extracting of the `PeerManager` logic from behavior!=)

crates/services/p2p/src/p2p_service.rs
```diff
@@ -324,14 +337,19 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
...
-    ) -> Result<bool, PublishError> {
-        let result = self.swarm.behaviour_mut().report_message_validation_result(
-            msg_id,
-            &propagation_source,
-            acceptance,
-        );
-
-        result
+    ) {
+        if let Some(peer_score) = self
```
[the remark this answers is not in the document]
leviathanbeak: I have decoupled gossipsub score from peer_score_config (ie application score config).

crates/services/p2p/src/p2p_service.rs
```diff
@@ -445,47 +459,96 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
                 }
             }
 
-            FuelBehaviourEvent::PeerInfo(peer_info_event) => match peer_info_event {
-                PeerInfoEvent::PeerIdentified { peer_id, addresses } => {
+            FuelBehaviourEvent::PeerReport(peer_info_event) => match peer_info_event {
```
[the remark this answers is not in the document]
leviathanbeak: I don't think that would be good idea since `PeerReport` is not just related to `PeerManager` other behaviours might use it, for example `Discovery` behaviour that needs `addresses` for unrouteable peers.

Doing it that way would seem that PeerManager is the only part that is interested in the events.
Having said that, I see your point, too many things were exposed from `PeerManager` so I have refactored this part, exposed cleaner methods from `PeerManager` and now they handle events in which they are interested in.

crates/services/p2p/src/peer_report.rs
leviathanbeak: Yes things are much nicer in `PeerReport`.

> Maybe in the future, we also will be able to move the handling of some events directly into `PeerManager`, but it is already good for now=)

If we did this we would end up with the previous issue we just had, where `PeerManager` would become a Behaviour and in order to talk to p2p_service it would need to constantly push events up to the p2p_service (ie `PeerManagerEvent::BanPeer(peerId)`)

`on_swarm_event` comes from the swarm -> behaviour -> p2p_service

leviathanbeak:
Note, I have refactored some of the p2p events handling, and consequently I realized Discovery events need to be refactored/removed but I will do that in another PR don't want to clutter this one with those changes

crates/services/p2p/src/behavior.rs
```diff
@@ -15,22 +10,17 @@ use crate::{
...
     request_response::messages::{
         NetworkResponse,
         RequestMessage,
     },
 };
-use fuel_core_types::{
-    blockchain::primitives::BlockHeight,
-    services::p2p::peer_reputation::PeerScore,
-};
+
```
Voxelot, line 22:
```suggestion
```

crates/services/p2p/src/behavior.rs
```diff
@@ -189,45 +165,33 @@ impl<Codec: NetworkCodec> FuelBehaviour<Codec> {
...
+                debug!(target: "fuel-p2p", "Sent a report for MessageId: {} from PeerId: {}", msg_id, propagation_source);
+                if should_check_score {
+                    return self.gossipsub.peer_score(propagation_source)
+                }
+            }
+            Ok(false) => {
+                warn!(target: "fuel-p2p", "Message with MessageId: {} not found in the Gossipsub Message Cache", msg_id);
+            }
+            Err(e) => {
+                error!(target: "fuel-p2p", "Failed to publish Message with MessageId: {} with Error: {:?}", msg_id, e);
```
Voxelot, line 186:
```suggestion
                error!(target: "fuel-p2p", "Failed to report Message with MessageId: {} with Error: {:?}", msg_id, e);
```

crates/services/p2p/src/peer_report.rs
Voxelot: It does seem strange to have all this logic unrelated to reputation included here (e.g. notifying swarm about incoming connections). I'm fine with it for now though if we open up a ticket to revisit this.

crates/services/p2p/src/p2p_service.rs
```diff
@@ -380,113 +429,122 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
...
+    pub fn peer_manager(&self) -> &PeerManager {
+        &self.peer_manager
     }
 
     fn handle_behaviour_event(
         &mut self,
         event: FuelBehaviourEvent,
     ) -> Option<FuelP2PEvent> {
         match event {
-            FuelBehaviourEvent::Discovery(discovery_event) => {
```
Voxelot, line 392: Why did we remove the discovery event?

Voxelot (approved):

PR #1059, opened 2023-03-11
Take into account the previous block timestamp during block production

xgreenx:
Updated the logic of the `PoA` to take into account the `last_timestamp` during block production. If the user manually specified the block with the future timestamp, the block production will use a new timestamp as a base to produce blocks.

Changed the API of the `produce_blocks` endpoint to accept only the start time and the number of blocks(The user can't specify the interval between blocks, it is defined by the `Trigger::Interval` mode).

Allowed enabling of the `manual_blocks_enabled` for the `Trigger::Interval` production mode.

### DIFF
--- a/crates/fuel-core/src/schema/block.rs
+++ b/crates/fuel-core/src/schema/block.rs
@@ -29,7 +29,6 @@ use async_graphql::{
         EmptyFields,
     },
     Context,
-    InputObject,
     Object,
     SimpleObject,
     Union,
@@ -277,21 +276,13 @@ where
 #[derive(Default)]
 pub struct BlockMutation;
 
-#[derive(InputObject)]
-struct TimeParameters {
-    /// The time to set on the first block
-    start_time: U64,
-    /// The time interval between subsequent blocks
-    block_time_interval: U64,
-}
-
 #[Object]
 impl BlockMutation {
     async fn produce_blocks(
         &self,
         ctx: &Context<'_>,
+        start_timestamp: Option<U64>,
         blocks_to_produce: U64,
-        time: Option<TimeParameters>,
     ) -> async_graphql::Result<U64> {
         let query = BlockQueryContext(ctx.data_unchecked());
         let consensus_module = ctx.data_unchecked::<ConsensusModule>();
@@ -303,80 +294,19 @@ impl BlockMutation {
             )
         }
 
-        let latest_block = query.latest_block()?;
-        let block_times = get_time_closure(&latest_block, time, blocks_to_produce.0)?;
-        consensus_module.manual_produce_block(block_times).await?;
+        let start_time = start_timestamp
+            .map(|timestamp| Tai64(timestamp.into()))
+            .unwrap_or(Tai64::now());
+        let blocks_to_produce: u64 = blocks_to_produce.into();
+        consensus_module
+            .manual_produce_block(start_time, blocks_to_produce as u32)
+            .await?;
 
         query
             .latest_block_height()
             .map(Into::into)
             .map_err(Into::into)
     }
-}
-
-fn get_time_closure(
-    latest_block: &CompressedBlock,
-    time_parameters: Option<TimeParameters>,
-    blocks_to_produce: u64,
-) -> anyhow::Result<Vec<Option<Tai64>>> {
-    if let Some(params) = time_parameters {
-        let start_time = params.start_time.into();
-        check_start_after_latest_block(latest_block, start_time)?;
-        check_block_time_overflow(&params, blocks_to_produce)?;
-        let interval: u64 = params.block_time_interval.into();
-
-        let vec = (0..blocks_to_produce)
-            .into_iter()
-            .map(|idx| {
-                let (timestamp, _) = params
-                    .start_time
-                    .0
-                    .overflowing_add(interval.overflowing_mul(idx).0);
-                Some(Tai64(timestamp))
-            })
-            .collect();
-        return Ok(vec)
-    };
-
-    Ok(vec![None; blocks_to_produce as usize])
-}
-
-fn check_start_after_latest_block(
-    latest_block: &CompressedBlock,
-    start_time: u64,
-) -> anyhow::Result<()> {
-    let current_height = *latest_block.header().height();
-
-    if current_height.as_usize() == 0 {
-        return Ok(())
-    }
-
-    let latest_time = latest_block.header().time();
-    if latest_time.0 > start_time {
-        return Err(anyhow!(
-            "The start time must be set after the latest block time: {:?}",
-            latest_time
-        ))
-    }
-
-    Ok(())
-}
-
-fn check_block_time_overflow(
-    params: &TimeParameters,
-    blocks_to_produce: u64,
-) -> anyhow::Result<()> {
-    let (final_offset, overflow_mul) = params
-        .block_time_interval
-        .0
-        .overflowing_mul(blocks_to_produce);
-    let (_, overflow_add) = params.start_time.0.overflowing_add(final_offset);
-
-    if overflow_mul || overflow_add {
-        return Err(anyhow!("The provided time parameters lead to an overflow"))
-    };
-
-    Ok(())
 }
 
 impl From<CompressedBlock> for Block {
--- a/crates/fuel-core/src/schema/block.rs
+++ b/crates/fuel-core/src/schema/block.rs
@@ -278,6 +278,10 @@ pub struct BlockMutation;
 
 #[Object]
 impl BlockMutation {
+    /// Sequentially produces `blocks_to_produce` blocks. The first block starts with
+    /// `start_timestamp`. If the block production in the [`crate::service::Config`] is
+    /// `Trigger::Interval { block_time }`, produces blocks with `block_time ` intervals between
+    /// them. The `start_timestamp` is the timestamp in seconds.
     async fn produce_blocks(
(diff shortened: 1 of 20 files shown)


nfurfaro:
One thing that's unclear to me as a user: Will I still be able to simulate the passing of a set amount of time if I'm not able to specify the interval between blocks ?

iqdecay:
> One thing that's unclear to me as a user: Will I still be able to simulate the passing of a set amount of time if I'm not able to specify the interval between blocks ?

Same question from the SDK team, this is a rather important feature for us.

xgreenx:
You can specify the start time of the next block and how many you want to generate blocks via the GraphQL endpoint. If the production mode in the node config is `Tigger::Interval { block_time }`(you can specify it during the start of the node or in the `Config`), the next blocks will be generated with the `block_time` intervals between them.

[Here](https://github.com/FuelLabs/fuel-core/pull/1059/files#diff-b43da8aec184b70c3e3ac74b3f156eb2142bf651b724ccfb3d7dc9e6d7d70b01) is an example of how to use it in the unit tests.

ControlCplusControlV:
Can we please wait for #1012 to go in first?

crates/fuel-core/src/schema/block.rs
```diff
@@ -277,21 +276,13 @@ where
...
-    /// The time interval between subsequent blocks
-    block_time_interval: U64,
-}
-
 #[Object]
 impl BlockMutation {
     async fn produce_blocks(
         &self,
         ctx: &Context<'_>,
+        start_timestamp: Option<U64>,
```
ControlCplusControlV, line 284: It would be nice to have a doc comment indicating start timestamp is in Tai64 format, as it's not as common and may cause some confusion if information is not preserved upstream

crates/services/consensus_module/poa/src/service_test/manually_produce_tests.rs
```diff
@@ -42,7 +58,7 @@ async fn can_manually_produce_block(
     ctx_builder.with_txpool(txpool);
 
     let mut importer = MockBlockImporter::default();
-    let (tx, mut rx) = tokio::sync::mpsc::channel(times.len() + 1);
+    let (tx, mut rx) = tokio::sync::mpsc::channel(times.len());
```
ControlCplusControlV, line 61: Why did channel length decrease here?

ControlCplusControlV (approved):
Looks good to me, I know a couple sdk users have been asking for node time travel so hopefully this make it easier, but please wait to merge until my GraphQL refactor lands as I am hoping to minimize merge conflicts and it has been taking me a while to get that in.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -158,6 +176,29 @@ where
...
+                }
+                Trigger::Hybrid { min_block_time, .. } => {
+                    increase_time(self.last_timestamp, min_block_time)
+                }
+            },
+            RequestType::Trigger => {
+                if now > self.last_timestamp {
+                    Ok(now)
+                } else {
+                    let duration = Instant::now().duration_since(self.last_block_created);
```
Dentosal, line 196: nit: you can use `self.last_block_created.elapsed()` instead

crates/services/consensus_module/poa/src/service_test/manually_produce_tests.rs
xgreenx: Because there is no reason to keep it +1 =D

leviathanbeak (dismissed):

crates/fuel-core/src/schema/block.rs
```diff
@@ -283,21 +282,17 @@ where
...
 #[Object]
 impl BlockMutation {
+    /// Sequentially produces `blocks_to_produce` blocks. The first block starts with
+    /// `start_timestamp`. If the block production in the [`crate::service::Config`] is
+    /// `Trigger::Interval { block_time }`, produces blocks with `block_time ` intervals between
+    /// them. The `start_timestamp` is the timestamp in seconds.
     async fn produce_blocks(
         &self,
         ctx: &Context<'_>,
+        start_timestamp: Option<U64>,
```
Voxelot, line 294:
```suggestion
        start_timestamp: Option<Tai64Timestamp>,
```

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -47,12 +47,13 @@ impl PoAAdapter {
 impl ConsensusModulePort for PoAAdapter {
     async fn manual_produce_block(
```
Voxelot, line 48:
```suggestion
    async fn manually_produce_blocks(
```

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -47,12 +47,13 @@ impl PoAAdapter {
...
         &self,
-        block_times: Vec<Option<Tai64>>,
+        block_time: Tai64,
+        number_of_blocks: u32,
     ) -> anyhow::Result<()> {
         self.shared_state
             .as_ref()
             .ok_or(anyhow!("The block production is disabled"))?
-            .manually_produce_block(block_times)
+            .manually_produce_block(block_time, number_of_blocks)
```
Voxelot, line 56:
```suggestion
            .manually_produce_blocks(block_time, number_of_blocks)
```

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -89,7 +90,7 @@ impl fuel_core_poa::ports::BlockProducer for BlockProducerAdapter {
     async fn produce_and_execute_block(
         &self,
         height: BlockHeight,
-        block_time: Option<Tai64>,
+        block_time: Tai64,
```
Voxelot, line 93:
```suggestion
        start_time: Tai64,
```

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -131,14 +145,17 @@ where
...
+        last_block: &BlockHeader,
         config: Config,
         txpool: T,
         block_producer: B,
         block_importer: I,
     ) -> Self {
         let tx_status_update_stream = txpool.transaction_status_events();
         let (request_sender, request_receiver) = mpsc::channel(100);
+        let last_timestamp = last_block.time();
+        let duration = Duration::from_secs(Tai64::now().0 - last_timestamp.0);
```
Voxelot, line 157: What happens if the service is initialized after some manual blocks were already produced with a timestamp > now?

xgreenx: Overflow=) Replaced with `saturating_sub`

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -158,6 +177,29 @@ where
...
+                }
+                Trigger::Hybrid { min_block_time, .. } => {
+                    increase_time(self.last_timestamp, min_block_time)
+                }
+            },
+            RequestType::Trigger => {
+                if now > self.last_timestamp {
+                    Ok(now)
+                } else {
+                    let duration = self.last_block_created.elapsed();
```
Voxelot, line 197: hmm should we still preserve the interval time period between blocks if manual mode is enabled? Otherwise if this gets triggered during a test it will likely start creating blocks at the same height for a while.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -75,22 +79,34 @@ pub struct SharedState {
...
+                },
+                sender,
+            )))
             .await?;
         receiver.await?
     }
 }
 
+struct ManualProduction {
+    pub start_time: Tai64,
```
Voxelot, line 101: Should this be optional as well?

crates/services/txpool/src/service.rs
```diff
@@ -66,8 +67,10 @@ impl TxStatusChange {
         }
     }
 
-    pub fn send_complete(&self, id: Bytes32) {
-        tracing::info!("Transaction {id} successfully included in a block");
+    pub fn send_complete(&self, id: Bytes32, block_height: &BlockHeight) {
+        tracing::info!(
+            "Transaction {id} successfully included in a block {block_height}"
```
Voxelot, line 72:
```suggestion
            "Transaction {id} successfully included in block {block_height}"
```

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -281,9 +345,8 @@ where
                         .await;
                 }
             }
-            (Trigger::Interval { .. }, RequestType::Manual)
-            | (Trigger::Hybrid { .. }, RequestType::Manual) => {
-                unreachable!("Trigger types interval and hybrid cannot be used with manual. This is enforced during config validation")
+            (Trigger::Hybrid { .. }, RequestType::Manual) => {
+                unreachable!("Trigger types hybrid cannot be used with manual. This is enforced during config validation")
```
Voxelot, line 349: Is this still the case though? Seems like `next_time(..)` would just use the `min_block_time` as the interval.

xgreenx: Yeah, we can remove it, but I didn't do that to minimize the number of cases that we need to support=D

Voxelot (approved):

ControlCplusControlV (approved):

Issue #1061, opened 2023-03-13
Large contracts fail to deploy over p2p


Voxelot:
Elix finance reported that their contract deployment failed on beta3. The contract bytecode is roughly 3mb. 

We recently increased the max p2p message payload size to 7mb, but that may need to be revisited to look for any other p2p message size limitations preventing 3mb from being broadcast.

PR #1062, opened 2023-03-14
Support large contract deployments over p2p

Voxelot:
fixes: https://github.com/FuelLabs/fuel-core/issues/1061

Increases the max_block_size to match the size limit of transactions

### DIFF
--- a/crates/services/p2p/src/config.rs
+++ b/crates/services/p2p/src/config.rs
@@ -275,8 +275,7 @@ pub(crate) fn build_transport(
     };
 
     let multiplex_config = {
-        let mut mplex_config = mplex::MplexConfig::default();
-        mplex_config.set_max_buffer_size(usize::MAX);
+        let mplex_config = mplex::MplexConfig::default();
 
         let mut yamux_config = yamux::YamuxConfig::default();
         yamux_config.set_max_buffer_size(MAX_RESPONSE_SIZE);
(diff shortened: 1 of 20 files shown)


crates/services/p2p/src/service.rs
```diff
@@ -181,6 +181,7 @@ where
                         let broadcast = GossipsubBroadcastRequest::NewTx(transaction);
                         let result = self.p2p_service.publish_message(broadcast);
                         if let Err(e) = result {
+                            eprintln!("{e:?}");
```
xgreenx, line 184:
```suggestion
```

xgreenx (approved):

xgreenx (dismissed):

xgreenx (dismissed):

crates/services/p2p/src/config.rs
```diff
@@ -272,10 +275,11 @@ pub(crate) fn build_transport(
     };
 
     let multiplex_config = {
-        let mut mplex_config = mplex::MplexConfig::new();
-        mplex_config.set_max_buffer_size(MAX_NUM_OF_FRAMES_BUFFERED);
+        let mut mplex_config = mplex::MplexConfig::default();
+        mplex_config.set_max_buffer_size(usize::MAX);
```
freesig, line 279: Is this really meant to be usize::MAX?

Voxelot: We followed the settings parity uses here in production. In a separate PR we may consider just removing mplex altogether since it's not really used.

freesig: Will this lead to an unbounded buffer though (well bounded by the system resources)?

Voxelot: Since it still works without this setting I'll just remove it.

freesig (approved):

Issue #588, opened 2022-09-01
Proposal to move out GraphQL API from `fuel-core` to `fuel-indexer`


xgreenx:
# Overview
The proposal was discussed with the `fuel-indexer` team and accepted. You can find the proposal's description at the end of this description. The beginning of its description tracks the progress of the implementation.

--- 

# The progress of the implementation

- [x] Migrate all internal graphql query resolver db logic into a plain rust api for the FuelService. The graphql resolvers should just be a minimal wrapper around our Rust APIs and perform minimal logic. 
   - [x] https://github.com/FuelLabs/fuel-core/pull/858 
   - [x] https://github.com/FuelLabs/fuel-core/pull/909
- [x] #1069
   - [x] https://github.com/FuelLabs/fuel-core/pull/875
   - [ ] Extract related folders and field into a separate crate.
- [x] Clean up the `fuel-core-client` by removing any graphql/cynic types in the publicly facing client interface and replacing them with the plain and ergonomic domain types in `fuel-core-types` https://github.com/FuelLabs/fuel-core/issues/378
- [ ] Implement simple rpc-based API for `fuel-core` with minimal functionality.
- [ ] Re-use the `fuel-core-graphql-api` in the `fuel-indexer`.
- [ ] Move  `fuel-core-client` into `fuel-indexer` repository.
- [ ] Update all downstream crates to work with `fuel-indexer`.

## Unresolved questions
- Do we need to move debugger API into `fuel-indexer`?

## Child tickets:

- https://github.com/FuelLabs/fuel-core/issues/1127
- https://github.com/FuelLabs/fuel-core/issues/1128
- https://github.com/FuelLabs/fuel-core/issues/1126
- https://github.com/FuelLabs/fuel-core/issues/1122
- https://github.com/FuelLabs/fuel-core/issues/1123
- https://github.com/FuelLabs/fuel-core/issues/1124
- https://github.com/FuelLabs/fuel-core/issues/1125
- https://github.com/FuelLabs/fuel-core/issues/1129
- https://github.com/FuelLabs/fuel-core/issues/1130

--- 
# The initial proposal

## Summary

The `fuel-core` implements CRUD functionality, interaction with the user, indexation, subscription on events(not implemented yet), and implementation of some specific queries. All this API functionality is not **required** for the blockchain.

This issue is a place where we can discuss which parts of the API functionality we want to move out and what not. Where do we want to implement moved-out functionality, how, and who?

[...]

PR #1066, opened 2023-03-17
[Draft] Extract graphql into its own crate

ControlCplusControlV:
Opening for transparency and early feedback

TO-DO
- [ ] - re-export `async-graphql` 
- [ ] - find a way to merge `dap::Query` back into main query
- [ ] - fix tests
- [ ] - move dap behind debugger feature flag


ControlCplusControlV:
closes #588

freesig (changes requested):
This moves the query files into the graphql crate but those files are needed in fuel core. The idea was that the query module doesn't depend on graphql. If these files are moved out of fuel core we have no way of using fuel-core as a library

ControlCplusControlV:
> This moves the query files into the graphql crate but those files are needed in fuel core. The idea was that the query module doesn't depend on graphql. If these files are moved out of fuel core we have no way of using fuel-core as a library

The traits cannot be in `fuel-core` though because it would create a circular dependency between `fuel-core` and `fuel-graphql`, would moving them into `fuel-core-types` be better so they can still act as an interface?

Issue #946, opened 2023-01-26
Retryable Messages


Voxelot:
Implementation of new spec changes here: https://github.com/FuelLabs/fuel-specs/pull/443

The primary purpose of these changes is to protect messages from being lost if a transaction is reverted. In our current model, messages contain both valuable metadata as well as gas for paying L2 fees.

In the new model, messages will be split into two types:
1. Messages: Metadata only - retryable (may still include an eth amount but it cannot be spent on gas)
2. DepositCoin: Gas only - not retryable

The main changes to fuel-core/fuel-vm involves:

 

- [ ] Introduction of a new `DepositCoin` transaction input type for gas-only bridge messages. This new tx input type functions similarly to regular coin inputs, but instead of being dependent on the output of a previous tx, it depends on gas-only message that has been relayed.
- [ ] Messages with metadata no longer contribute to the available gas.
- [ ] The resourcesToSpend API can be changed back to coinsToSpend with the following modifications:
    - No longer returns messages that have any metadata
    - Will now need to select `DepositCoin`'s in addition to regular coins
- [ ] Messages aren't marked as spent unless included in a successfully executed script tx. No other tx types may use message inputs (they may use DepositCoin's though)
- [ ] Exclude gas from messages in the available free balances for a tx.
- [ ] Include gas from DepositCoin's in the available free balances for a tx.
- [ ] Remove `Output::Message` from `fuel-tx`.
- [ ] Renamed fields in the block header for messages according to the spec.
- [ ] Remove `MessageId` from all places and replace it with the `Message.nonce`, because it has the same meaning.
- [ ] In the `fuel-vm` allow `recipient` to be zero in the `Receipt::Message`.

PR #1067, opened 2023-03-17
Retrayable messages fuel-core part

xgreenx:
Fixes https://github.com/FuelLabs/fuel-core/issues/946 

This PR also uses the latest `fuel-vm` with https://github.com/FuelLabs/fuel-vm/pull/373.

- Updated the code to use named input types.
- Removed the usage of the `MessageId` from all places except the `BlockHeader` and related to the block stuff.
- Used `Nonce` as a primary key for messages instead of `MessageId`. Because of it, we don't need `CheckedMessage` anymore. Also, several tests failed because the nonce was the same.
- Updated the `Nonce` from `u64` to 32 bytes. Updated the ABI to use `uint256` instead of `uint64` - https://github.com/FuelLabs/fuel-v2-contracts/pull/35
- Renamed `Resources` into `Coins`. Renamed all comments and GraphQL API. Moved `coin.rs` and `resource.rs` into one `coins.rs`.
- Fixed some tests for https://github.com/FuelLabs/fuel-vm/pull/377 not to increase the `HP`.
- Removed usage of the `Output::Message`.
- Renamed `message_output_count` - > `message_receipt_count` and `message_output_root` -> `message_receipt_root`

And the main feature of why this change was created(implemented by several lines): 
`MessageData` aren't marked as spent unless included in a successfully executed script tx


crates/client/src/client/schema/coins.rs
```diff
@@ -132,6 +136,98 @@ pub struct CoinIdFragment {
...
+    pub amount: U64,
+    pub sender: Address,
+    pub recipient: Address,
+    pub nonce: Nonce,
+    pub da_height: U64,
+}
+
+#[derive(cynic::InlineFragments, Debug)]
+#[cynic(schema_path = "./assets/schema.sdl")]
+pub enum Coins {
```
Voxelot, line 180: Maybe `CoinType` since `Coins` is plural and implies multiple coins?

```suggestion
pub enum CoinType {
```

Voxelot (approved):

Issue #1046, opened 2023-02-24
Update withdrawal proof API to support proving from a block header committed to L1


Voxelot:
### Periodic Commits

In mainnet, an external service will periodically commit block ids to ethereum. This new approach approach will be future compatible with alternate modes of consensus in the future (e.g. PoS) while also improving security. For example, it will be much more difficult for a compromised sequencer key to be used for equivocating or fraudulently producing withdrawal proofs.

This new design will require users to select from an existing finalized commit on ethereum which will be used to create merkle proof from that commit to the block. The commit block height needs to be greater than the block header containing the message being withdrawn, however it should be chosen as close a possible to reduce the merkle proof size.

### Changes to API inputs
```rust
async fn message_proof(
        &self,
        ctx: &Context<'_>,
        transaction_id: TransactionId,
        message_id: MessageId,
        commit_block_id: BlockId,
    ) -> async_graphql::Result<Option<MessageProof>>
```

### Changes to API response:
```rust

/// Type containing merkle proof data
pub struct MerkleProof {
    /// The proof set
    pub proof_set: Vec<Bytes32>,
    /// The proof index
    pub proof_index: u64,
}

/// Proves to da layer that this message was included in a Fuel block
pub struct MessageProof {
    /// Proof that message is contained within the provided block header
    pub message_proof: MerkleProof,
    /// Proof that the provided block header is contained within the blockchain history
    pub block_proof: MerkleProof,
    /// The previous fuel block header that contains the message. message block height <=
    /// commit block height.
    pub message_block_header: BlockHeader,
    /// The consensus header associated with the finalized commit being used
    /// as the root of the block proof
    pub commit_block_header: ConsensusHeader<GeneratedConsensusFields>,
    /// The messages sender address.
    pub sender: Address,
    /// The messages recipient address.
    pub recipient: Address,
    /// The nonce from the message.
    pub nonce: Bytes32,
    /// The amount from the message.
    pub amount: Word,
    /// The data from the message.
    pub data: Vec<u8>,
}
```

The block proof will need to generate a binary merkle proof using the commit_block_id as the root, to prove the existence of the message block header. This will utilize the `FuelBlockMerkleData` table.

cc: @pixelcircuits

PR #1071, opened 2023-03-22
Update withdrawal proof API to support proving from a block header committed to L1

xgreenx:
Close https://github.com/FuelLabs/fuel-core/issues/1046

Updated the code to use `fuel-vm 0.29`


crates/client/src/client/schema/message.rs
```diff
@@ -108,6 +125,10 @@ pub struct MessageProofArgs {
     pub transaction_id: TransactionId,
     /// Message id of the output message that requires a proof.
     pub message_id: MessageId,
+    /// The block id of the commitment block.
+    pub commit_block_id: Option<BlockId>,
```
Voxelot, line 129: maybe add a comment explaining why this might be optional?

crates/fuel-core/src/query/message.rs
```diff
@@ -167,52 +152,108 @@ pub fn message_proof<T: MessageProofData + ?Sized>(
...
+        };
+
+    if message_block_header.message_receipt_root != root {
+        // This is bad as it means there's a bug in our prove code.
+        tracing::error!(
+            "block header {:?} root doesn't match generated proof root {:?}",
+            message_block_header,
+            root
+        );
+        return Ok(None)
```
Voxelot, line 199: Should we just return an actual error here?

crates/fuel-core/src/query/message.rs
```diff
@@ -167,52 +152,108 @@ pub fn message_proof<T: MessageProofData + ?Sized>(
...
+    let commit_block_header = match database
+        .block(&commit_block_id)
+        .into_api_result::<CompressedBlock, StorageError>()?
+    {
+        Some(t) => t.into_inner().0,
+        None => return Ok(None),
+    };
+
+    let verifiable_commit_block_height = *commit_block_header.height() - 1u32.into();
+    let block_proof = database.block_history_prove(
```
Voxelot, line 212: Maybe this should be consistent with `message_receipts_proof`?

```suggestion
    let block_proof = database.block_history_proof(
```

crates/fuel-core/src/query/message.rs
```diff
@@ -167,52 +152,108 @@ pub fn message_proof<T: MessageProofData + ?Sized>(
...
 
     // Build the merkle proof from the above iterator.
-    let mut tree = fuel_merkle::binary::in_memory::MerkleTree::new();
+    let mut tree = MerkleTree::new();
 
     let mut proof_index = None;
 
-    for (index, id) in leaves {
+    for (index, id) in leaves.enumerate() {
         // Check id this is the message.
```
Voxelot, line 253:
```suggestion
        // Check if this is the message id being proved
```

crates/fuel-core/src/query/message.rs
```diff
@@ -167,52 +152,108 @@ pub fn message_proof<T: MessageProofData + ?Sized>(
...
+        sender,
+        recipient,
+        nonce,
+        amount,
+        data,
+    }))
+}
+
+fn message_receipts_proof<T: MessageProofData + ?Sized>(
+    data: &T,
```
Voxelot, line 231: Might be better to call this database/datastore, as data sounds like it could be anything.

```suggestion
    database: &T,
```

crates/fuel-core/src/query/message.rs
```diff
@@ -135,30 +132,18 @@ impl<D: DatabasePort + ?Sized> MessageProofData for D {
...
-    }
 }
 
 /// Generate an output proof.
 // TODO: Do we want to return `Option` here?
 pub fn message_proof<T: MessageProofData + ?Sized>(
-    data: &T,
+    database: &T,
     transaction_id: Bytes32,
     message_id: MessageId,
```
Voxelot, line 142: should we change this to nonce?

crates/fuel-core/src/schema/message.rs
```diff
@@ -115,33 +116,67 @@ impl MessageQuery {
...
     ) -> async_graphql::Result<Option<MessageProof>> {
         let data: &Database = ctx.data_unchecked();
+        let block_id = match (commit_block_id, commit_block_height) {
+            (Some(commit_block_id), None) => commit_block_id.0.into(),
+            (None, Some(commit_block_height)) => {
+                let block_height = commit_block_height.0.into();
+                data.block_id(&block_height)?
+            }
+            _ => Err(anyhow::anyhow!(
+                "Support either `commit_block_id` or `commit_block_height`"
```
Voxelot, line 130:
```suggestion
                "Either `commit_block_id` or `commit_block_height` must be provided exclusively"
```

crates/fuel-core/src/query/message.rs
xgreenx: `Nonce` or `MessageId` for us doesn't make any difference but may affect the end user. I left it `MessageId` because it is part of the verification process. Also, tests on our side and, I guess, on the contracts side, use `MessageId`=)

crates/fuel-core/src/query/message.rs
```diff
@@ -167,52 +152,107 @@ pub fn message_proof<T: MessageProofData + ?Sized>(
...
+    }))
+}
+
+fn message_receipts_proof<T: MessageProofData + ?Sized>(
+    database: &T,
+    message_id: MessageId,
+    message_block_txs: &[Bytes32],
+) -> StorageResult<Option<(Bytes32, MerkleProof)>> {
+    // Get the message receipts from the block.
+    let leaves: Vec<Vec<Receipt>> = message_block_txs
```
bvrooman, line 235: I'm not very familiar with how receipts are created, but is it possible to have storage-backed Merkle trees for receipts, and append to the Merkle tree whenever we create new receipts? It looks like we might be able to use per-message BMTs (keyed by `message_id`)?

crates/fuel-core/src/query/message.rs
```diff
@@ -167,52 +152,107 @@ pub fn message_proof<T: MessageProofData + ?Sized>(
...
+        commit_block_header,
+        sender,
+        recipient,
+        nonce,
+        amount,
+        data,
+    }))
+}
+
+fn message_receipts_proof<T: MessageProofData + ?Sized>(
```
bvrooman, line 229: What do you think about separating this call into two calls: one to get the proof set, and a second to get the root?
A while back, we were thinking about modifying the underlying Merkle tree `prove` call to return only the proof set, and call `root` separately when we need the root for verification (https://github.com/FuelLabs/fuel-merkle/issues/126)

crates/fuel-core/src/query/message.rs
```diff
@@ -167,52 +152,107 @@ pub fn message_proof<T: MessageProofData + ?Sized>(
...
+        None => return Ok(None),
+    };
+
+    let (root, message_proof) =
+        match message_receipts_proof(database, message_id, &message_block_txs)? {
+            Some(proof) => proof,
+            None => return Ok(None),
+        };
+
+    if message_block_header.message_receipt_root != root {
```
bvrooman, line 192: Do we know if this can actually happen? Maybe we can omit this if we have good test coverage to make sure our proofs are generated correctly and that our block production works correctly. Then, we also don't need the root here.

crates/fuel-core/src/query/message.rs
xgreenx: Usually, we don't have too many messages in the block, so it is faster to calculate that in runtime than have access to the storage, I guess. 

But if we have a lot of messages, yeah, optimization like this would not be bad. Bad it's more about GraphQL performance, and maybe should be solved on the indexer side

crates/fuel-core/src/service/adapters/graphql_api.rs
```diff
@@ -223,6 +236,40 @@ impl TxPoolPort for TxPoolAdapter {
     }
 }
 
+impl DatabaseMessageProof for Database {
+    fn block_history_proof(
+        &self,
+        message_block_height: &BlockHeight,
+        commit_block_height: &BlockHeight,
+    ) -> StorageResult<MerkleProof> {
+        if message_block_height > commit_block_height {
```
Voxelot, line 245: This seems like a lot of business logic that is being wrapped inside of an adapter. Ideally, this logic would exist in a more unit testable domain context such as one of the following:


- A query context for merkle proofs that just fulfills this particular API. The ports would need the lower-level functions like fetching merkle data that do minimal logic.
- Implemented on the fuel-core database itself

xgreenx: > A query context for merkle proofs that just fulfills this particular API. The ports would need the lower-level functions like fetching merkle data that do minimal logic.

It is not actually a lot of business logic because it only takes data from the table and passes it into `MerkleTree`. I put it here because I don't want to expose the `FuelBlockMerkleMetadata` and `FuelBlockMerkleData` tables outside the `fuel-core`(The `MerkleTree` requires explicit work with `FuelBlockMerkleData`). 

Also, it aligns with the current code for blocks. It is an internal detail of how the merklezation of the database works and we can change it at any time.

> Implemented on the fuel-core database itself

I will move it to `Database` itself=)

Voxelot (dismissed):

bvrooman (dismissed):

Voxelot (approved):

PR #1072, opened 2023-03-22
Predicate Estimation `fuel-core` part

ghost:
The PR adds support for predicate estimation from [`fuel-vm`](https://github.com/FuelLabs/fuel-vm/pull/386).

The change adds:
- All predicate inputs now accept the `predicate_gas_used` field.
- New API endpoint `estimate_predicates` that estimates predicates and returns the transaction with updated predicates.


bin/fuel-core-client/src/main.rs
```diff
@@ -15,6 +15,8 @@ enum TransactionCommands {
     /// Submit a JSON encoded transaction for inclusion in a block
     Submit { tx: String },
     /// Submit a JSON encoded transaction for a dry-run execution
+    Estimate { tx: String },
```
Voxelot, line 18:
```suggestion
    EstimatePredicates { tx: String },
```

we don't want to confuse people with thinking this is a general estimation API like what the SDK does.

bin/fuel-core-client/src/main.rs
```diff
@@ -15,6 +15,8 @@ enum TransactionCommands {
     /// Submit a JSON encoded transaction for inclusion in a block
     Submit { tx: String },
     /// Submit a JSON encoded transaction for a dry-run execution
```
Voxelot, line 17: This doesn't use dry-run at all right?

crates/fuel-core/src/schema/tx/input.rs
```diff
@@ -174,6 +184,7 @@ impl From<&fuel_tx::Input> for Input {
                 tx_pointer: TxPointer(*tx_pointer),
                 witness_index: *witness_index,
                 maturity: (*maturity).into(),
+                predicate_gas_used: 0.into(),
```
Voxelot, line 187: Should this use the value from the input?

Voxelot (approved):

Voxelot (approved):
approving - but I don't think this change is mergable until the VM is released.

xgreenx:
Close in favor of the https://github.com/FuelLabs/fuel-core/pull/1195

PR #1080, opened 2023-03-28
Nginx tuneup

Voxelot:
- enables buffering of proxy responses
- increases client and response buffer sizes to 1mb to avoid frequent tmp file access / disk IO
- increase timeout from a default of 1m to 1h to avoid 504 timeouts

### DIFF
--- a/deployment/ingress/eks/fuel-core-ingress.yaml
+++ b/deployment/ingress/eks/fuel-core-ingress.yaml
@@ -12,9 +12,12 @@ metadata:
     nginx.ingress.kubernetes.io/affinity-mode: persistent
     nginx.ingress.kubernetes.io/session-cookie-expires: "172800"
     nginx.ingress.kubernetes.io/session-cookie-max-age: "172800"
-    nginx.org/proxy-connect-timeout: "3600s"
-    nginx.org/proxy-read-timeout: "3600s"
-    nginx.org/proxy-send-timeout: "3600s"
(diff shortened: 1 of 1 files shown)


deployment/ingress/eks/fuel-core-ingress.yaml
```diff
@@ -12,6 +12,12 @@ metadata:
     nginx.ingress.kubernetes.io/affinity-mode: persistent
     nginx.ingress.kubernetes.io/session-cookie-expires: "172800"
     nginx.ingress.kubernetes.io/session-cookie-max-age: "172800"
+    nginx.ingress.kubernetes.io/proxy-connect-timeout: "3600s"
+    nginx.ingress.kubernetes.io/proxy-read-timeout: "3600s"
```
xgreenx, line 16: Hmm, I'm afraid that it could cause a huge queue of requests. I think it worth to try it, but we need to monitor the queue size somehow=)

xgreenx (approved):

deployment/ingress/eks/fuel-core-ingress.yaml
Voxelot: it looks like we should be able to monitor this in prometheus, using the nginx exporter plugin: https://github.com/nginxinc/nginx-prometheus-exporter

Will make a ticket for this in the devops backlog

Voxelot: https://github.com/FuelLabs/infrastructure/issues/99

PR #1090, opened 2023-03-30
add prometheus extension for graphql tracing

### LINKED ISSUE #1036: Create extension for async-graphql to collect APM stats in prometheus
Follow the async-graphql extension for [opentelemetry](https://github.com/async-graphql/async-graphql/blob/master/src/extensions/opentelemetry.rs) as an example to create a prometheus extension.

It should have the following metrics on a per grapqhl query basis:
- bucket response times into a histogram
- track the number of requests per second via gauges

leviathanbeak:
closes #1036 

on metrics side I created GrapqhlMetrics with a Counter (requests per second) and Histogram (response times),
and on Grapqhl side I created a PrometheusExtension implementing `Extension` trait from async-grapqhl.

### DIFF
--- a/crates/fuel-core/src/graphql_api.rs
+++ b/crates/fuel-core/src/graphql_api.rs
@@ -10,6 +10,7 @@ use fuel_core_types::{
 use std::net::SocketAddr;
 
 pub mod ports;
+#[cfg(feature = "metrics")]
 pub(crate) mod prometheus;
 pub mod service;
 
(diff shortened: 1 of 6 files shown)


Voxelot:
Could we utilize info from the ctx and dynamically register different metrics for each query & mutation? This way we know how long coins_to_select is taking vs dry_run.

leviathanbeak:
> Could we utilize info from the ctx and dynamically register different metrics for each query & mutation? This way we know how long coins_to_select is taking vs dry_run.

@Voxelot good idea, will check it out

leviathanbeak:
Closing in favor of #1113

PR #1093, opened 2023-03-30
Restrict concurrency of GraphQL API

Voxelot:
apply a concurrency limit to graphql api to reserve at least one free core. This should help alleviate CPU overloading issues.

I didn't apply a restriction to the subscriptions route, since they're mostly async friendly.


crates/fuel-core/src/graphql_api/service.rs
```diff
@@ -164,7 +166,20 @@ pub fn new_service(
...
+        .route(
+            "/graphql",
+            post(graphql_handler)
+                .route_layer({
+                    let num_cpu = num_cpus::get();
+                    // reserve at least one free cpu core to avoid overloading the async runtime
+                    // with blocking work. If there's only 1 core, ensure we don't go below 1.
+                    let concurrency = max(1, num_cpu.checked_sub(1).unwrap_or_default());
+                    ServiceBuilder::new()
+                        .concurrency_limit(concurrency)
```
xgreenx, line 178: I’m afraid that it will block almost all requests if we have `NUM_CPU - 1` heavy dry runs=D

Also, maybe let’s make it configurable. The main idea with it is to remove a lot of simultaneous requests that access the database. I think we can increase concurrency to 100 to unblock queries. And do some experiments with this value=)

xgreenx: But with a limit of `100`, we can have a situation where all requests are `dry_run` and `93` of them will wait for semaphore blocking all other queries. It would be nice to have a separate limit `100` for common queries and `7` for `dry-run`s=)

freesig: Yeh this doesn't make sense. Axum is not using a thread per connection. They are async tasks

Voxelot:
closing as I think we need to address this issue a bit differently

PR #1096, opened 2023-04-02
Release 0.17.8

Voxelot:
Yet another hotfix.. In my local testing it seems like jemalloc has less overhead when initializing vm instances (which we do frequently). It should also do a better job of reclaiming memory.

The release also disables `dap` endpoints for the `production`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,6 +93,8 @@ jobs:
             args: test --no-default-features
           - command: test
             args: --manifest-path version-compatibility/Cargo.toml --workspace
+          - command: build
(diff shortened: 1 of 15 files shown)


xgreenx (approved):
I tested the change in the DEV net. Sentry-1(blue) had 0.17.8(with this change) and Sentry-3(red) had 0.17.7(without this change).

I spammed each node two times with one million of dry_runs (simple transfer transactions). After each spam, I did a small pause to allow the node to rest.

The jemalloc shows much better memory management. It used less memory during each run, after the end of each spamming attack the memory almost normalized.

image

image

Voxelot:
This has been released already, and the changes have been ported to land on master in #1102

PR #1107, opened 2023-04-06
use chainid for transactions and predicates

Voxelot:
Adapts fuel-core to changes in https://github.com/FuelLabs/fuel-vm/pull/406

Routing the consensus parameters everywhere is a little painful, since we need to fetch the transaction ID so often. I've cleaned this up a little bit by introducing a new infallible api in [fuel-vm](https://github.com/FuelLabs/fuel-vm/pull/419) to fetch the transaction id when it's known to be computed already. 

This could be cleaned up even further by using Checked\<Tx\> in more places, such as using `pub type CheckedBlock = Block<CheckedTransaction>` in the executor and consensus modules. But this would involve larger structural changes and optimizations to avoid rechecking transactions that have already been checked in the tx pool during sync, and should be done in a separate PR.

### DIFF
--- a/crates/fuel-core/src/executor.rs
+++ b/crates/fuel-core/src/executor.rs
@@ -71,6 +71,7 @@ use fuel_core_types::{
         Receipt,
         Transaction,
         TransactionFee,
+        TxId,
         TxPointer,
         UniqueIdentifier,
         UtxoId,
@@ -593,9 +594,7 @@ where
             &self.config.chain_conf.transaction_parameters,
         )?;
 
-        let tx_id = checked_tx
-            .transaction()
-            .id(&self.config.chain_conf.transaction_parameters);
+        let tx_id = checked_tx.id();
         let min_fee = checked_tx.metadata().min_fee();
         let max_fee = checked_tx.metadata().max_fee();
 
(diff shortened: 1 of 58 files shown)


crates/fuel-core/src/executor.rs
```diff
@@ -309,7 +310,12 @@ where
         block_db_transaction
             .deref_mut()
             .storage::<FuelBlocks>()
-            .insert(&finalized_block_id, &result.block.compress())?;
+            .insert(
+                &finalized_block_id,
+                &result
+                    .block
+                    .compress(&self.config.chain_conf.transaction_parameters),
```
xgreenx, line 317: It seems we need to rename `transaction_parameters` into `consensus_parameters`

crates/fuel-core/src/executor.rs
```diff
@@ -1053,7 +1064,10 @@ where
                             )?;
                             if tx_pointer != &coin.tx_pointer {
                                 return Err(ExecutorError::InvalidTransactionOutcome {
-                                    transaction_id: tx.id(),
+                                    transaction_id: tx.id(&self
```
xgreenx, line 1067: Maybe we can calculate it one time at the begin of the `execute_create_or_script` method and pass it everywhere?

xgreenx (approved):

crates/fuel-core/src/executor.rs
Voxelot: Can do as a follow up

Issue #823, opened 2022-12-09
Properly handle the closing of the subscription


xgreenx:
## Overview

The source is a pull request https://github.com/FuelLabs/fuel-core/pull/822.

While we support only one subscription(for transaction status), it is not a critical issue. Also, after https://github.com/FuelLabs/fuel-core/issues/588, it would be the responsibility of the indexer to provide subscription functionality. But if we will start adding more subscriptions on the `fuel-core` side before moving to the indexer, we need to resolve this issue.

## Issue

[`eventsource-client`](https://crates.io/crates/eventsource-client) library used for subscription(and seems [Server-Sent Events](https://html.spec.whatwg.org/multipage/server-sent-events.html) protocol) doesn't provide a built-in way to gracefully terminate a stream from the server side.

## Solution

Create a generic type `Event<Data>` and wrap all subscription types into this type. This type should answer the question, "Is that the last event?". Based on the answer, the server should close the connection. The client should use the same type to receive events and do the same check and close the connection if it is the last event.

It will prevent on the client side several useless iterations like reconnection and error logging and will provide a generic workflow over each subscription to not have this issue in the future.

PR #1108, opened 2023-04-06
Make transaction status stream work

freesig:
This PR refactors the `transaction_status_change` function to improve its readability and efficiency. The main changes are:

1. Changed input stream type from `BoxStream<'a, Result<TxUpdate, BroadcastStreamRecvError>>` to `BoxStream<'a, TxStatusMessage>`.
2. Changed return type from `impl Stream<Item = anyhow::Result<TransactionStatus>> + 'a` to `impl Stream<Item = anyhow::Result<ApiTxStatus>> + 'a`.
3. Removed unnecessary async/await and error mapping for the initial database status check.
4. Replaced the complex unfold logic with a simpler chain of the initial database check and stream, followed by a take_until to handle stream termination.
5. Replaced the nested match structure with a single match block to improve readability.
6. Simplified the stream termination logic by closing the stream if s new transaction status is not Submitted.

Additionally, this PR introduces property testing to ensure the refactored `transaction_status_change` function behaves correctly under various input conditions. This will help identify potential edge cases, increase the test coverage, and provide greater confidence in the robustness of the changes made.


Voxelot:
We'll likely also want to consider this issue as well, because without a terminal message indicating the closing of the stream it makes error handling on the client side more difficult: https://github.com/FuelLabs/fuel-core/issues/823

crates/fuel-core/src/service/query.rs
```diff
@@ -0,0 +1,88 @@
...
+use crate::{
+    query::transaction_status_change,
+    schema::tx::types::TransactionStatus,
+};
+
+use super::*;
+
+impl FuelService {
+    /// Submit a transaction to the txpool.
+    pub fn submit(&self, mut tx: Transaction) -> anyhow::Result<InsertionResult> {
```
Voxelot, line 28: 👍🏻 Putting the rust-level APIs here will be helpful for migrating to other RPCs and also for those using fuel-core as a library (indexer)

crates/services/txpool/src/service/update_sender.rs
```diff
@@ -0,0 +1,362 @@
...
+        C: CreateChannel,
+    {
+        // Remove closed senders from the list.
+        // Careful not to hold the lock while awaiting.
+        {
+            remove_closed(&mut self.senders.lock());
+        }
+
+        // Acquire a permit from the semaphore asynchronously.
+        let permit = Arc::clone(&self.permits).acquire().await;
```
Voxelot, line 244: Could this cause messages to get missed if the creation of the sender is awaiting a permit?

mitch-fuel: are you suggesting a message wouldn't be sent in this scenario?

mitch-fuel: or that the transaction could complete before it is possible to finalize the subscription?

mitch-fuel: we could require subscribing before the transaction is submitted given that the id is known but this could be abused. another option would be holding an LRU of recent tx that can still be subscribed for, but that supposes timing requirements

freesig: Yes it will cause messages to be missed but we can't garauntee any sort of message delivery anyway

xgreenx (approved):
@freesig  I made small changes around the main logic. Could you check it, please? If everything okay, I think we can merge it=)

crates/fuel-core/src/service/adapters/txpool.rs
```diff
@@ -147,7 +148,10 @@ impl fuel_core_txpool::ports::TxPoolDb for Database {
     fn transaction_status(
         &self,
         tx_id: &fuel_core_types::fuel_types::Bytes32,
-    ) -> StorageResult<Option<fuel_core_types::services::txpool::TransactionStatus>> {
-        Ok(self.get_tx_status(tx_id)?)
+    ) -> StorageResult<fuel_core_types::services::txpool::TransactionStatus> {
```
freesig, line 151: Kinda weird to return an error for not found

freesig (commented):
looks good

PR #1118, opened 2023-04-12
pass swarm to manager

leviathanbeak:
Deals with https://github.com/FuelLabs/fuel-core/pull/1028#discussion_r1163074564

Pass the Punisher to PeerManager so it can deal with the banishment internally :)

### DIFF
--- a/crates/services/p2p/src/peer_manager.rs
+++ b/crates/services/p2p/src/peer_manager.rs
@@ -25,13 +25,10 @@ use std::{
     time::Duration,
 };
 use tokio::time::Instant;
-use tracing::debug;
-
-#[derive(Debug, Clone, Copy)]
-pub struct PeerScoreUpdated {
-    pub should_ban: bool,
-    pub score: AppScore,
-}
+use tracing::{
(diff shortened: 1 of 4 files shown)


crates/services/p2p/src/peer_manager.rs
```diff
@@ -86,8 +83,17 @@ impl PeerManager {
         }
     }
 
-    pub fn is_reserved_peer(&self, peer_id: &PeerId) -> bool {
-        self.reserved_peers.contains(peer_id)
+    pub fn handle_gossip_score_update<T: Punisher>(
```
mitch-fuel, line 86: maybe consider changing the function name to match the pattern of `update_app_score`

leviathanbeak: app score gets updated by PeerManager, but gossipsub does not, this one just handles the post-update consequences

mitch-fuel (approved):
I think this looks great, addresses all things discussed.

xgreenx (approved):

PR #1119, opened 2023-04-12
Release/v0.17.10

xgreenx:
It is the change only for beta 3. We don't have this problem in the `master` branch.
It removes the calculation/verification of the Merkle root for the `state_root` and `balance_root` fields in inputs and outputs.

More about it in the [slack](https://fuellabs.slack.com/archives/C04EWEF3XEZ/p1681594984254819?thread_ts=1681222434.771829&cid=C04EWEF3XEZ).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2483,7 +2483,7 @@ dependencies = [
 
 [[package]]
 name = "fuel-core"
-version = "0.17.9"
+version = "0.17.10"
 dependencies = [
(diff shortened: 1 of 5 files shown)


bvrooman (approved):
LGTM

Voxelot (approved):

xgreenx:
We've released it. We don't need to duplicate the change for the master because we don't have this problem

PR #1140, opened 2023-04-19
Add pyroscope profiling

Voxelot:
This PR adds pyroscope to fuel-core-bin. Pyroscope is a platform that allows for continuous profiling, enabling real-time flamegraphs in deployed environments.

This is still in draft, and requires additional testing in the dev environment.

image


deployment/Dockerfile
```diff
@@ -19,6 +19,7 @@ RUN cargo chef prepare --recipe-path recipe.json
 
 FROM chef as builder
 ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
+ENV CARGO_PROFILE_RELEASE_DEBUG=true
```
xgreenx, line 22: Hmm, maybe we need to create two images. One with debug information and one without. Or keep with debug info until mainnet and split into two images before mainnat. But in this case, we need to create an issue not to forget to split it.

Voxelot: If we do split debug info, the compilation time for docker images wont really change. It would only allow users to opt-out of a larger docker image, which I don't think is a huge deal?

Voxelot:
This is somewhat blocked on https://github.com/pyroscope-io/pyroscope-rs/pull/99 to avoid a dependency on installing openssl.

PR #1141, opened 2023-04-19
Rocksdb database optimisations

xgreenx:
- Replaced combined `column_key` with an array of hash maps.
- Added RocksDB LRU cache for uncompressed rows. It speeds up the loading of objects from the database two times. 
image
image

- Optimized the deserialization of the contracts via the `OptimizedContract` wrapper type. Instead of 53% now it takes only 10% of the time.
image

- Replaced all values `Vec<u8>` with `Arc<Vec<u8>>` to avoid cloning because we use it only for deserialization, and we can't consume the actual `Vec<u8>`. It is helpful for `MemoryStore` and `MemoryTransactionView`.
- Added a new e2e test for the contract with a large state. 
- Updated the `snapshot`command to do a snapshot only of the contract's state.
- Updated the `MemoryStore` to use `BTreeMap` instead of `HashMap`. It significantly improved the performance of the `iter_all` without `RocksDB`. Without this optimization, a new e2e failed for `--no-default-features` build with a timeout(but worked with `RocksDB`).


crates/fuel-core/src/state.rs
```diff
@@ -68,25 +65,12 @@ pub trait BatchOperations: KeyValueStore {
 
 #[derive(Debug)]
 pub enum WriteOperation {
-    Insert(Vec<u8>, Column, Vec<u8>),
-    Remove(Vec<u8>, Column),
+    Insert(Value),
```
Voxelot, line 68: Do we still need to have all these changes to the write operation given that we are now using the built-in row-cache for rocksdb?

crates/chain-config/src/snapshots/fuel_core_chain_config__config__tests__snapshot_simple_coin_state.snap
```diff
@@ -26,7 +26,7 @@ expression: json
     "max_gas_per_tx": 100000000,
     "max_script_length": 1048576,
     "max_script_data_length": 1048576,
-    "max_storage_slots": 255,
+    "max_storage_slots": 131072,
```
Voxelot, line 29: we probably shouldn't be changing the max storage slots for the default config?

crates/fuel-core/src/database.rs
```diff
@@ -76,50 +77,60 @@ pub mod vm_database;
 )]
 pub enum Column {
     /// The column id of metadata about the blockchain
-    Metadata = 0,
+    Metadata,
```
Voxelot, line 80: one advantage of assigning integers here, is that we can remove columns in the future without causing a lot of problems on existing datasets.

crates/chain-config/src/snapshots/fuel_core_chain_config__config__tests__snapshot_simple_coin_state.snap
xgreenx: Without it, a new e2e test will fail=) I can change it only for the integration test in the e2e module, but maybe it is also will be useful for benches in the future, so I decided to change the default values

crates/fuel-core/src/state.rs
```diff
@@ -1,9 +1,6 @@
-use crate::{
-    database::{
-        Column,
-        Result as DatabaseResult,
-    },
-    state::in_memory::transaction::MemoryTransactionView,
+use crate::database::{
+    Column,
```
Voxelot, line 2: This state module was suppose to be something that could be extracted from the main fuel-core crate without any knowledge of its' data model. Why do we need to include the columns here?

crates/fuel-core/src/database.rs
xgreenx: We can handle it later in the `as_usize` function, if we need this functionality

bin/fuel-core/src/cli/snapshot.rs
```diff
@@ -1,24 +1,44 @@
...
+#[derive(Debug, Clone, Subcommand)]
+pub enum SubCommands {
+    /// Creates a snapshot of the entire database and produces a chain config.
+    #[command(arg_required_else_help = true)]
+    Everything {
+        /// Specify either an alias to a built-in configuration or filepath to a JSON file.
+        #[clap(name = "CHAIN_CONFIG", long = "chain", default_value = "local_testnet")]
+        chain_config: String,
+    },
+    /// Creates a config for teh contract.
```
Voxelot, line 35:
```suggestion
    /// Creates a config for the contract.
```

crates/chain-config/src/config/chain.rs
```diff
@@ -119,6 +119,10 @@ impl ChainConfig {
                 coins: Some(initial_coins),
                 ..StateConfig::default()
             }),
+            transaction_parameters: ConsensusParameters {
+                max_storage_slots: 1 << 17, // 131072
```
Voxelot, line 123: This should be overidden in the integ tests, not in the main default config

crates/fuel-core/src/state.rs
xgreenx: It was here before=)

crates/chain-config/src/snapshots/fuel_core_chain_config__config__tests__snapshot_simple_coin_state.snap
Voxelot: maybe we should make a new default for these kinds of benchmarking / integ tests that doesn't apply to general usage

crates/fuel-core/src/state.rs
Voxelot: I see, nm then. But maybe something to resolve in the future.

freesig:
Hmm the diff on this PR is 120K lines. Is something wrong with the diff?

bin/e2e-test-client/tests/integration_tests.rs
```diff
@@ -82,7 +82,9 @@ async fn execute_suite(config_path: String) {
 }
 
 async fn setup_local_node() -> FuelService {
-    FuelService::new_node(Config::local_node()).await.unwrap()
+    let mut config = Config::local_node();
+    config.chain_conf.transaction_parameters.max_storage_slots = 1 << 17; // 131072
```
freesig, line 86: Can you add a comment on why this is set so high?

bin/e2e-test-client/src/tests/script.rs
```diff
@@ -46,14 +54,51 @@ pub async fn dry_run(ctx: &TestContext) -> Result<(), Failed> {
     )
     .await??;
 
+    _dry_runs(ctx, &transaction, 1000).await
+}
+
+// Maybe deploy a contract with large state and execute teh script
```
freesig, line 60:
```suggestion
// Maybe deploy a contract with large state and execute the script
```

crates/fuel-core/src/database.rs
```diff
@@ -76,50 +77,60 @@ pub mod vm_database;
 )]
 pub enum Column {
```
freesig, line 78:
```suggestion
/// # Warning this enum is order dependent and changing the variant order will break the database.
pub enum Column {
```

crates/fuel-core/src/database.rs
freesig: Honestly don't love this as it makes the enum order dependent

crates/fuel-core/src/database.rs
```diff
@@ -167,10 +178,17 @@ unsafe impl Send for Database {}
...
+    pub fn new(data_source: DataSource) -> Self {
+        Self {
+            data: data_source,
+            _drop: Default::default(),
+        }
+    }
+
     #[cfg(feature = "rocksdb")]
-    pub fn open(path: &Path) -> DatabaseResult<Self> {
+    pub fn open(path: &Path, capacity: Option<usize>) -> DatabaseResult<Self> {
```
freesig, line 189:
```suggestion
    pub fn open(path: &Path, capacity: impl Into<Option<usize>>) -> DatabaseResult<Self> {
```
Then the api user can pass in `open(path, 1000)` or `open(path, None)`

crates/fuel-core/src/state.rs
```diff
@@ -12,18 +9,18 @@ use std::{
...
 
 pub type DataSource = Arc<dyn TransactableStorage>;
-pub type ColumnId = u32;
-pub type KVItem = DatabaseResult<(Vec<u8>, Vec<u8>)>;
+pub type Value = Arc<Vec<u8>>;
+pub type KVItem = DatabaseResult<(Vec<u8>, Value)>;
 
 pub trait KeyValueStore {
-    fn get(&self, key: &[u8], column: Column) -> DatabaseResult<Option<Vec<u8>>>;
+    fn get(&self, key: &[u8], column: Column) -> DatabaseResult<Option<Value>>;
```
freesig, line 16: One downside to forcing this to be inside an arc is that the caller cannot modify the `Vec<u8>` if they need to. Rocks does not return an `Arc` and if we move to a rocks transaction then it also won't have an `Arc` so it's probably best to not force the `Arc` on the caller.

crates/fuel-core/src/database/balances.rs
```diff
@@ -61,24 +56,8 @@ impl StorageMutate<ContractsAssets> for Database {
...
-            .iter()
-            .filter_map(|(key, value)| {
-                (&key[..parent.len()] == parent.as_ref()).then_some((key, value))
-            })
-            .sorted_by_key(|t| t.0)
-            .map(|(_, value)| value.to_be_bytes());
-
-        Ok(crypto::ephemeral_merkle_root(root).into())
+    fn root(&self, _: &ContractId) -> Result<MerkleRoot, Self::Error> {
+        Ok(*empty_sum_sha256())
```
freesig, line 60: Why are we doing this again?

crates/fuel-core/src/database/contracts.rs
```diff
@@ -16,19 +16,114 @@ use fuel_core_storage::{
...
+                E: serde::de::Error,
+            {
+                Ok(v.to_vec())
+            }
+        }
+
+        let bytes = deserializer.deserialize_bytes(BytesVisitor {})?;
+        Ok(Self { bytes })
+    }
+}
```
freesig, line 76: Can this be replace by [serde_bytes](https://docs.rs/serde_bytes/latest/serde_bytes/)?

crates/fuel-core/src/database/contracts.rs
```diff
@@ -16,19 +16,114 @@ use fuel_core_storage::{
...
+    ) -> Result<Option<<ContractsRawCode as Mappable>::OwnedValue>, Self::Error> {
+        Database::insert::<_, _, OptimizedContract>(
+            self,
+            key.as_ref(),
+            Column::ContractsRawCode,
+            &value,
+        )
+        .map(|option| option.map(Into::into))
+        .map_err(Into::into)
+    }
```
freesig, line 114: On master this insert adds the contract in a way that it can be loaded directly into vm memory and executed. Is this also the case here?

crates/fuel-core/src/state/in_memory/memory_store.rs
```diff
@@ -4,35 +4,28 @@ use crate::{
...
     sync::Mutex,
 };
 
 #[derive(Default, Debug)]
 pub struct MemoryStore {
-    // TODO: Remove `Mutex` and usage of the `column_key`.
-    // TODO: Use `BTreeMap`.
-    inner: Mutex<HashMap<Vec<u8>, Vec<u8>>>,
+    // TODO: Remove `Mutex`.
+    inner: [Mutex<BTreeMap<Vec<u8>, Value>>; Column::COUNT],
```
freesig, line 28: Can't this Mutex just be removed? This type is not clone so it's not being shared anyway

crates/fuel-core/src/state.rs
xgreenx: The caller is deserializer, so we don't modify it and only use it in read only mode

crates/fuel-core/src/database/balances.rs
xgreenx: To be sure that we don't calculate root in any case for the beta 3=)

crates/fuel-core/src/state/in_memory/memory_store.rs
xgreenx: Because it requires modification of the `KeyValueStore` trait to use `&mut self`, but it requires a huge refactoring

crates/fuel-core/src/database/contracts.rs
xgreenx: Yes, it is what we want to do, and also improve the performance of the `contracts` endpoint

crates/fuel-core/src/database/contracts.rs
xgreenx: We can, I only want to have fewer dependencies and the code that do exactly what we want. But we plan to remove it in the master, so I'm okay to use `serder_bytes`

PR #1144, opened 2023-04-20
feat: Client primitives

bvrooman:
Related issues:
- #1121 

This PR introduces a module of Rust types used by the `Client` in the `fuel-core-client` crate. These types are referred to as client types and they are used to compose the `Client` interface. All function signatures in the Client interface now use only client types and Rust primitives. All references to GraphQL schema types are removed from the interface's function signatures. 

Specifically, this PR introduces the following client types:
- `Balance`
- `Block` (as well as its constituent parts `Header`, `Consensus`, `Genesis`, and `PoAConsensus`)
- `ChainInfo`
- `Coin`, `MessageCoin`, and `CoinType`
- `ConsensusParameters`
- `Contract` and `ContractBalance`
- `MerkleProof`
- `Message` and `MessageProof`
- `NodeInfo` 

While GraphQL types are used inside the client methods, all types returned by the GraphQL server are transformed back into client types so that the `Client` returns only these client types. This is enabled by defining a number of `From` traits, that transform the GraphQL schema type into the client type.


crates/client/src/client/types/translation.rs
```diff
@@ -0,0 +1,191 @@
...
+            message_receipt_root: message_receipt_root.into(),
+            height: value.height.0.into(),
+            prev_root: prev_root.into(),
+            time: value.time.0.into(),
+            application_hash,
+        }
+    }
+}
+
+impl From<schema::block::Consensus> for Consensus {
```
freesig, line 49: I'd probably just put these translation types into the mod that defines the type they convert into. In this case the `block` mod

crates/client/src/client/hex_formatted.rs
```diff
@@ -0,0 +1,79 @@
...
+
+#[derive(Debug, Clone, Copy, Default, PartialEq)]
+pub struct HexFormatted<T: Debug + Clone + Default + PartialEq>(pub T);
+
+impl<T: LowerHex + Debug + Clone + Default + PartialEq> Serialize for HexFormatted<T> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.serialize_str(format!("{:#x}", self.0).as_str())
```
freesig, line 32: This formats to hex when it is serialized but the type indicates it's already hex formated. If this is intentional `pub T` should at least be `T`

bvrooman: I reverted these changes now that I am not redefining any primitives. Instead, I am keep the existing `HexFormatted` struct in `schema.rs`, but that has the same questionable design. I can make your suggested change in this PR or I can look deeper into that another time.

crates/client/src/client/types.rs
```diff
@@ -1,3 +1,73 @@
+pub mod balance;
```
freesig, line 1: Maybe move mod below use

crates/client/src/client/types.rs
```diff
@@ -1,3 +1,73 @@
...
+    };
+
+    pub type BlockId = Bytes32;
+    pub type Hash = Bytes32;
+    pub type HexString = Vec<u8>;
+    pub type MerkleRoot = Bytes32;
+    pub type TransactionId = Bytes32;
+
+    #[derive(Debug)]
+    pub struct Tai64Timestamp(pub Tai64);
```
freesig, line 39: Probably don't need this wrapper type

freesig (dismissed):
All lokks good, a few suggesstions. Nice work.

xgreenx (approved):

crates/client/src/client/types/message.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// The messages sender address.
+    pub sender: Address,
+    /// The messages recipient address.
+    pub recipient: Address,
+    /// The nonce from the message.
+    pub nonce: Nonce,
+    /// The amount from the message.
+    pub amount: u64,
+    /// The data from the message.
+    pub data: HexString,
```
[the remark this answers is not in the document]
bvrooman: As discussed in sync today, I changed the alias to be called `Bytes` to avoid the confusion around `HexString`.

xgreenx (approved):

PR #1145, opened 2023-04-20
[Do not merge] Add logging of the long transaction

xgreenx:


xgreenx:
it was temporary PR to trigger teh docket image build

PR #1148, opened 2023-04-21
[Do not merge] Use `tokio_rayon::spawn` instead of `tokio::task::spawn_blocking`

xgreenx:


xgreenx:
I didn't see a lot of difference with vs `spawn_block_fifo`, but let's keep `fifo` because it is more fair

PR #1150, opened 2023-04-22
Updated `PrometheusExtension` to not use lock

xgreenx:


xgreenx:
Included into https://github.com/FuelLabs/fuel-core/pull/1151

PR #1151, opened 2023-04-22
Release v0.17.11 final patch

xgreenx:
- Added a new e2e test that allows dry run any hex-encoded transaction.
- Simplified the `PrometheusExtension`.
- Use a new `fuel-vm 0.26.3` with disabled logging for `instruction` function.
- Use `tokio_rayon::spawn_fifo` to run dry runs instead of manual semaphore and `tokio::spawn_blocking`.
- Removed 5 second delay for dry runs caused by the relayer.


crates/fuel-core/src/graphql_api/prometheus.rs
```diff
@@ -45,41 +35,33 @@ impl Extension for PrometheusExtInner {
     ) -> Response {
         let start_time = Instant::now();
         let result = next.run(ctx).await;
-
-        let op_name = self.operation_name.read().await;
-        if let Some(op) = &*op_name {
-            GRAPHQL_METRICS.graphql_observe(op, start_time.elapsed().as_secs_f64());
-        }
+        let seconds = start_time.elapsed().as_secs_f64();
+        GRAPHQL_METRICS.graphql_observe("request", seconds);
```
Voxelot, line 39: do we still need this if the resolve method is what actually logs the ops?

crates/services/producer/src/block_producer.rs
```diff
@@ -171,23 +160,12 @@ where
         height: BlockHeight,
         block_time: Option<Tai64>,
     ) -> anyhow::Result<PartialBlockHeader> {
-        let previous_block_info = self.previous_block_info(height)?;
-        let new_da_height = self
-            .select_new_da_height(previous_block_info.da_height)
-            .await?;
+        let mut block_header = self._new_header(height, block_time)?;
+        let new_da_height = self.select_new_da_height(block_header.da_height).await?;
```
Voxelot, line 164: So the delay for dry runs was caused by awaiting on the relayer?

crates/fuel-core/src/graphql_api/prometheus.rs
xgreenx: I prefer to keep it because, in this case, we can see how long someone has waited for some request. Because one request may have 100+ resolves. And based on information from it we can start future investigations

crates/services/producer/src/block_producer.rs
xgreenx: Yes, it caused 5 seconds delays and queuing requests. After the import of the eth block, it seems that the `synced` became `None` and waited for an actualization after `sync_minimum_duration`.

image

I'm not sure how it is better to fix it because we may need to really request some blocks again that we could miss while we loaded the previous one. So I fixed that on the `dry_run` side

Voxelot: It should return whatever the last synced state was, and only block temporarily while the state is being updated.

Voxelot: we might need to add a separate cached state that is always available to poll, and just uses AtomicU64 to update the block height. We should really only wait if the relayer is really far behind the da layer.

xgreenx: https://github.com/FuelLabs/fuel-core/pull/1151/commits/46853f3af368df2b504c93395a747c24923bcd31

freesig (approved):
I think we should fix the relayer issue in a way where it records the synced height always without setting it to none but we can do that in another PR on master

PR #1152, opened 2023-04-24
Update cargo audit and add to commit

freesig:
Updates the audit to the latest version and also adds it to PRs


freesig:
I also removed honeycomb so this can actually pass

freesig:
hmm this workflow seems to just fail randomly

Voxelot:
Looks like this action isn't compatible with sparse registries yet, it's likely importing an outdated version of cargo:

```
Caused by:
  usage of sparse registries requires `-Z sparse-registry`
```

crates/fuel-core/Cargo.toml
```diff
@@ -39,7 +39,7 @@ fuel-core-types = { workspace = true, features = [
 ] }
 futures = { workspace = true }
 hex = { version = "0.4", features = ["serde"] }
-hyper = { version = "0.14" }
+hyper = { version = "0.14.26" }
```
xgreenx, line 42: You can move it on workspace level and reuse in all places `{ workspace = true }`

crates/fuel-core/src/graphql_api/service.rs
```diff
@@ -158,20 +155,14 @@ pub fn new_service(
...
         .data(txpool)
         .data(producer)
         .data(consensus_module);
     // use honeycomb tracing wrapper if api key is configured
-    let builder = if honeycomb_enabled {
-        builder.extension(HoneyTrace)
-    } else {
-        builder.extension(Tracing)
-    };
+    let builder = builder.extension(Tracing);
```
xgreenx, line 165: We can remove tracing at all because we don't use it in the testnet

bin/fuel-core/src/cli/run.rs
```diff
@@ -158,9 +158,6 @@ pub struct Command {
 
     #[clap(long = "tx-pool-ttl", default_value = "5m", env)]
     pub tx_pool_ttl: humantime::Duration,
-
```
xgreenx, line 161: Could you also update helm chart in `deployment/charts/templates/fuel-core-deploy.yaml`, please?

crates/client/src/client.rs
```diff
@@ -247,8 +247,13 @@ impl FuelClient {
                 })?;
         }
 
-        let client =
-            client_builder.build_with_conn(es::HttpsConnector::with_webpki_roots());
+        let client = client_builder.build_with_conn(
+            hyper_rustls::HttpsConnectorBuilder::new()
+                .with_webpki_roots()
+                .https_only()
```
xgreenx, line 253: Why do we want to use only `https`? I think it is better to allow all [`https_or_http`](https://docs.rs/hyper-rustls/latest/hyper_rustls/struct.HttpsConnectorBuilder.html#method.https_or_http)

It is useful when you need to connect locally via `127.0.0.1`

Voxelot (approved):

crates/fuel-core/src/graphql_api/service.rs
Voxelot: It's still useful to setup the spans used to wrap events logged within the apis

freesig:
> Looks like this action isn't compatible with sparse registries yet, it's likely importing an outdated version of cargo:
> 
> ```
> Caused by:
>   usage of sparse registries requires `-Z sparse-registry`
> ```

How can we fix this? Do we need to fork the action?

xgreenx (approved):

PR #1156, opened 2023-04-25
Accurate Transaction Verification Costing

freesig:
# Improved Transaction Costing Specification
[Related #4](https://github.com/FuelLabs/fuel-vm/issues/415)

## Objective
Create a more accurate transaction costing model by identifying and addressing areas where costs are not adequately captured, thus minimizing risks such as liveness failures, unsynchronizable nodes, and unanticipated DA expenses.


freesig:
It's much easier if I don't need to worry about the correct combinations of inputs x outputs but then I need to know when a bench fails that should be correct. It would be nice to print something out on failure with the input so this is clear

xgreenx:
We close this PR for now, but the work done here is useful in the future to finish https://github.com/FuelLabs/fuel-core/issues/1402.

PR #1157, opened 2023-04-25
Release 0.17.11

xgreenx:
- Replaced combined `column_key` with an array of hash maps.
- Added RocksDB LRU cache for uncompressed rows. It speeds up the loading of objects from the database two times.
image image

- Optimized the deserialization of the contracts via the `OptimizedContract` wrapper type. Instead of 53% now it takes only 10% of the time.
image

- Replaced all values `Vec<u8>` with `Arc<Vec<u8>>` to avoid cloning because we use it only for deserialization, and we can't consume the actual `Vec<u8>`. It is helpful for `MemoryStore` and `MemoryTransactionView`.
- Added a new e2e test for the contract with a large state. 
- Updated the `snapshot`command to do a snapshot only of the contract's state.
- Updated the `MemoryStore` to use `BTreeMap` instead of `HashMap`. It significantly improved the performance of the `iter_all` without `RocksDB`. Without this optimization, a new e2e failed for `--no-default-features` build with a timeout(but worked with `RocksDB`).
- Added a new e2e test that allows dry run any hex-encoded transaction.
- Simplified the `PrometheusExtension`.
- Use a new `fuel-vm 0.26.3` with disabled logging for `instruction` function.
- Use `tokio_rayon::spawn_fifo` to run dry runs instead of manual semaphore and `tokio::spawn_blocking`.
- Removed 5 second delay for dry runs caused by the relayer.


freesig:
I'm confused, didn't I already review this?

xgreenx:
Yes, you did=) But it was 2 separate branches, now it is merged into one final release branch.

Release is created https://github.com/FuelLabs/fuel-core/releases/tag/v0.17.11

Issue #1137, opened 2023-04-17
Database caching for uncompressed data


Voxelot:

Voxelot:
use rocksdb lru cache

Issue #1143, opened 2023-04-20
The `state_root` calculation for the `Create` transaction takes an eternity


xgreenx:
If the user provided a vast number of storage slots(in the `Create` transaction) during the contract creation, it takes a lot of time to calculate the `state_root`.

We currently use a default value `255` for the `ConsensusParameters::max_storage_slots`, so it is not critical. But if we want to support duplication of the contract with their state, we need to speed up this process and increase the default `max_storage_slots`.

freesig:
- [ ] Benchmark and flamegraph fuel-core with "large" state
- [ ] Benchmark and flamegraph fuel-merkle with "large" state
- [ ] Compare difference

bvrooman:
Our current approach is based on the idea that we can implement a new algorithm that takes advantage of having all the leaves of the Sparse Merkle tree up front. Theoretically, this knowledge can be used to reduce or eliminate redundant calculations when compared to sequentially updating the SMT for each leaf, where we recalculate the root at each update operation.

This means:
- Implementing a new algorithm that uses all leaves at once, rather than sequentially
- Proving the implementation is sound and correct, and produces the same roots when compared to sequential updates
- Establishing a baseline by benchmarking the performance of sequential updates
- Benchmarking the new algorithm and demonstrating that this new algorithm has superior performance (most importantly with large inputs) over the baseline

PR #1158, opened 2023-04-25
Applying optimizations from `0.17.11` release

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1137

The PR duplicates most of the changes from the https://github.com/FuelLabs/fuel-core/pull/1157 but for the `master` branch.
It adds two e2e tests, but they fail because of new changes in the `fuel-vm`(the change to the `PC` register). It is not critical, but replacing them with the correct scripts later would be nice.
- One test uploads the contract with a huge large state and executes it. The test takes a lot of time to execute and highlights that we have performance problems with the state calculation: https://github.com/FuelLabs/fuel-core/issues/1143
- Another test is more of a base for spamming the testnets. We can put any hex-encoded transaction and dry run it on the network. It contains a simple transfer transaction right now.

The change improves the `fuel-core snapshot` command to accept the arguments, and you can either snap the whole state of the blockchain via the `everyone` command, or the state only of the contract. In the follow-up PRs we can snapping of the coins and messages too.

Optimizations:
- Use a lru cache to store the uncompressed data in the DB. The size is configurable via CLI `max-database-cache-size`.
- Added a `DryRun` execution type that skips the coinbase `Mint` transaction. In the future, we can skip some non-actual steps too.
- Updated the `PrometheusExtension` to not use the lock.
- Removed usage of the tracing extension in the graph QL because we have `PrometheusExtension` to gather the statistic about requests. https://github.com/FuelLabs/fuel-core/pull/1152 removes the honeycomb, so I didn't duplicate it here.
- Replaced the `tokio::task::spawn_blocking` + `Semaphore` by `tokio_rayon::spawn_fifo`.
- Replaced the concatenated `column_key` with the array for separate hash maps per column in the `MemoryStore` and `MemoryTransactionView`.
- Replaced the `Vec<u8>` in the KV API by the `Arc<Vec<u8>>`, because in most cases, we use it only for deserialization, and the memory database can continue to store the type. It helps reduce the number of clones.
- Used `BTreeMap` in the `MemoryStore`, because it allows to have a more optimized `iter_all` method. Without this change, a new e2e test fails with a timeout.
- Dry run doesn't wait for relayer anymore.


freesig:
This seems like the same PR that I already reviewed. If so maybe it can just be merged?

xgreenx:
It is almost the same, but tuned for the `master` version=) The difference only in `KeyValueStore`.

If you think that it is ready to land `master` then approve it=D

crates/fuel-core/src/executor.rs
```diff
@@ -430,12 +437,19 @@ where
...
         block.transactions.append(&mut filtered_transactions);
 
-        coinbase_tx = self.check_coinbase(
-            block_height,
-            coinbase_tx,
-            Some(execution_data.coinbase),
-        )?;
-        self.apply_coinbase(coinbase_tx, block, execution_data, block_db_transaction)?;
+        if execution_kind != ExecutionKind::DryRun {
+            coinbase_tx = self.check_coinbase(
```
Voxelot, line 441: Do we still need to do this check in production mode?

crates/fuel-core/src/executor.rs
```diff
@@ -1821,6 +1837,39 @@ mod tests {
...
+
+            config.chain_conf.transaction_parameters.gas_price_factor = gas_price_factor;
+
+            let producer = Executor::test(Default::default(), config);
+
+            let mut block = Block::default();
+            *block.transactions_mut() = vec![script.into()];
+
+            let ExecutionResult { block, .. } = producer
+                .execute_and_commit(ExecutionBlock::DryRun(block.into()))
```
Voxelot, line 1867: should we add any special logic here to ensure that dry run's won't actually commit? Or even better, enforce it at a type level?

crates/services/producer/src/block_producer.rs
```diff
@@ -171,23 +160,12 @@ where
         height: BlockHeight,
         block_time: Tai64,
     ) -> anyhow::Result<PartialBlockHeader> {
-        let previous_block_info = self.previous_block_info(height)?;
-        let new_da_height = self
-            .select_new_da_height(previous_block_info.da_height)
-            .await?;
+        let mut block_header = self._new_header(height, block_time)?;
+        let new_da_height = self.select_new_da_height(block_header.da_height).await?;
```
Voxelot, line 164: maybe add some comments around this part of the code, to make it more clear that dry runs should reuse a previous da_height while new blocks should fetch da_height from the relayer. Since this was a big performance issue in the past, I think it's worth making the distinction very clear and explain why.

crates/fuel-core/src/executor.rs
xgreenx: We need only to apply coinbase, but not to check. But while the `check_coinbase` function does simple check, I think it is better to keep it to be sure that everything is okay(and makes the code consistance)=)

crates/fuel-core/src/executor.rs
xgreenx: We use `execute_and_commit` only for tests, so I marked it with the corresponding compilation feature. Inside the method will panic if someone passes the `ExecutionBlock ::DryRun`

Voxelot (approved):

PR #1159, opened 2023-04-25
add task params

leviathanbeak:
Adding optional Task parameters when initializing `Task:RunnableTask` from `RunnableService`.

### DIFF
--- a/crates/services/src/service.rs
+++ b/crates/services/src/service.rs
@@ -71,6 +71,9 @@ pub trait RunnableService: Send {
     /// The initialized runnable task type.
     type Task: RunnableTask;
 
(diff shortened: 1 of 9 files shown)


crates/services/src/service.rs
```diff
@@ -71,14 +71,21 @@ pub trait RunnableService: Send {
     /// The initialized runnable task type.
     type Task: RunnableTask;
 
+    /// Optional parameters used to when initializing into task.
+    type TaskParams: Send + Default;
```
xgreenx, line 75: You don't need to require `Default` here=) Only in the `impl` section

Issue #1120, opened 2023-04-12
Block Production should wait until the node is synced


Voxelot:
In a data-recovery scenario, the sequencer needs to wait until it is synced with it's peers before producing any blocks.

freesig:
Must connect to M of N trusted reserved nodes

PR #1169, opened 2023-05-02
sync with peers before producing blocks

leviathanbeak:
Closes https://github.com/FuelLabs/fuel-core/issues/1120

After refactoring and enabling `TaskParams` for `Services` in order for them to be turned into `Task` with a `TaskParam`,
I created SyncPort, and added it as a `TaskParam` for `poa`'s Service.

The current implementation is that the `SyncAdapter` only checks for 60 seconds if a new block was imported, if not, it considers the PoA node to be synced.

Improvement on top of this would be - waiting first to connect with N amount of Reserved Peers.

### DIFF
--- a/crates/fuel-core/src/service/adapters.rs
+++ b/crates/fuel-core/src/service/adapters.rs
@@ -4,7 +4,9 @@ use crate::{
 };
 use fuel_core_consensus_module::block_verifier::Verifier;
 use fuel_core_txpool::service::SharedState as TxPoolSharedState;
+use fuel_core_types::services::block_importer::ImportResult;
 use std::sync::Arc;
+use tokio::sync::broadcast::Receiver;
 
 pub mod block_importer;
 pub mod consensus_module;
@@ -85,5 +87,15 @@ impl P2PAdapter {
 impl P2PAdapter {
     pub fn new() -> Self {
         Default::default()
+    }
+}
+
+pub struct SyncAdapter {
+    block_rx: Receiver<Arc<ImportResult>>,
+}
+
+impl SyncAdapter {
+    pub fn new(block_rx: Receiver<Arc<ImportResult>>) -> Self {
+        Self { block_rx }
     }
 }
--- a/crates/fuel-core/src/service/adapters/consensus_module/poa.rs
+++ b/crates/fuel-core/src/service/adapters/consensus_module/poa.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use crate::{
     database::Database,
     fuel_core_graphql_api::ports::ConsensusModulePort,
@@ -12,6 +14,7 @@ use anyhow::anyhow;
 use fuel_core_poa::{
     ports::{
         BlockImporter,
+        SyncPort,
         TransactionPool,
     },
     service::SharedState,
@@ -105,5 +108,28 @@ impl BlockImporter for BlockImporterAdapter {
         self.block_importer
             .commit_result(result)
             .map_err(Into::into)
+    }
+}
+
+const TIME_UNTIL_SYNCED: Duration = Duration::from_secs(60);
+
+#[async_trait::async_trait]
+impl SyncPort for crate::service::adapters::SyncAdapter {
+    async fn sync_with_peers(&mut self) -> anyhow::Result<()> {
+        // todo: connect to N amount of peers first!
+
+        'outer: loop {
+            tokio::select! {
+                _ = self.block_rx.recv() => {
+                    // keep receiving them blocks
+                }
+                _ = tokio::time::sleep(TIME_UNTIL_SYNCED) => {
+                    // time expired we are synced!
+                    break 'outer;
+                }
+            }
+        }
+
+        Ok(())
     }
 }
--- a/crates/services/consensus_module/poa/src/service.rs
+++ b/crates/services/consensus_module/poa/src/service.rs
@@ -6,6 +6,7 @@ use crate::{
     ports::{
         BlockImporter,
         BlockProducer,
+        SyncPort,
         TransactionPool,
     },
     Config,
@@ -428,7 +429,7 @@ where
 
     type SharedData = SharedState;
     type Task = Task<T, B, I>;
-    type TaskParams = ();
+    type TaskParams = Box<dyn SyncPort>;
 
     fn shared_data(&self) -> Self::SharedData {
         self.shared_state.clone()
@@ -437,8 +438,10 @@ where
     async fn into_task(
         self,
         _: &StateWatcher,
-        _: Self::TaskParams,
+        mut syncer: Self::TaskParams,
     ) -> anyhow::Result<Self::Task> {
+        let _ = syncer.as_mut().sync_with_peers().await;
+
         match self.trigger {
             Trigger::Never | Trigger::Instant => {}
             Trigger::Interval { block_time } => {
@@ -517,19 +520,17 @@ pub fn new_service<D, T, B, I>(
     txpool: T,
     block_producer: B,
     block_importer: I,
+    sync: Box<dyn SyncPort>,
 ) -> Service<T, B, I>
 where
     T: TransactionPool + 'static,
     B: BlockProducer<Database = D> + 'static,
     I: BlockImporter<Database = D> + 'static,
 {
-    Service::new(Task::new(
-        last_block,
-        config,
-        txpool,
-        block_producer,
-        block_importer,
-    ))
+    Service::new_with_params(
+        Task::new(last_block, config, txpool, block_producer, block_importer),
+        sync,
+    )
 }
 
 fn seal_block(
--- a/crates/fuel-core/src/service/sub_services.rs
+++ b/crates/fuel-core/src/service/sub_services.rs
@@ -1,5 +1,8 @@
 #![allow(clippy::let_unit_value)]
-use super::adapters::P2PAdapter;
+use super::adapters::{
+    P2PAdapter,
+    SyncAdapter,
+};
 
 use crate::{
     database::Database,
@@ -119,13 +122,19 @@ pub fn init_sub_services(
     let poa_config: fuel_core_poa::Config = config.try_into()?;
     let production_enabled =
         !matches!(poa_config.trigger, Trigger::Never) || config.manual_blocks_enabled;
+
     let poa = (production_enabled).then(|| {
+        let syncer = Box::new(SyncAdapter::new(
+            importer_adapter.block_importer.subscribe(),
+        ));
+
         fuel_core_poa::new_service(
             last_block.header(),
             poa_config,
             tx_pool_adapter.clone(),
             producer_adapter.clone(),
             importer_adapter.clone(),
+            syncer,
         )
     });
     let poa_adapter = PoAAdapter::new(poa.as_ref().map(|service| service.shared.clone()));
--- a/crates/services/consensus_module/poa/src/ports.rs
+++ b/crates/services/consensus_module/poa/src/ports.rs
@@ -86,3 +86,9 @@ pub trait RelayerPort {
         max_da_lag: &DaBlockHeight,
     ) -> anyhow::Result<()>;
 }
+
+#[async_trait::async_trait]
+pub trait SyncPort: Send + Sync {
+    /// await synchronization with the peers
+    async fn sync_with_peers(&mut self) -> anyhow::Result<()>;
+}


xgreenx:
image

The current code disables the `fuel-core-sync` service if the node starts in the block production mode. So you need to change this too, which may break some tests and the initial idea of this change overall. 

Also, waiting inside the `start` function is a bad idea because the start of the services should be a fast process. The introduction of the synchronization during the start will block the `FuelService` and all its consumers. So it is better to move this check into the `run` function itself. But that means that we need to listen for updates from `fuel-core-sync` during lifecycle of the `fuel-core-poa` -> We need to handle the case when  `fuel-core-block-importer` submits the block.

So this task is more tricky than we thought before.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -437,8 +438,10 @@ where
     async fn into_task(
         self,
         _: &StateWatcher,
-        _: Self::TaskParams,
+        mut syncer: Self::TaskParams,
     ) -> anyhow::Result<Self::Task> {
+        let _ = syncer.as_mut().sync_with_peers().await;
```
freesig, line 443: Maybe add a comment about why the result can be ignored?

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -107,3 +110,26 @@ impl BlockImporter for BlockImporterAdapter {
...
+const TIME_UNTIL_SYNCED: Duration = Duration::from_secs(60);
+
+#[async_trait::async_trait]
+impl SyncPort for crate::service::adapters::SyncAdapter {
+    async fn sync_with_peers(&mut self) -> anyhow::Result<()> {
+        // todo: connect to N amount of peers first!
+
+        'outer: loop {
+            tokio::select! {
+                _ = self.block_rx.recv() => {
```
freesig, line 123: What if this errors because the channel is closed?

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -107,3 +110,26 @@ impl BlockImporter for BlockImporterAdapter {
...
+                    // keep receiving them blocks
+                }
+                _ = tokio::time::sleep(TIME_UNTIL_SYNCED) => {
+                    // time expired we are synced!
+                    break 'outer;
+                }
+            }
+        }
```
freesig, line 131:
```suggestion
	while tokio::time::timeout(TIME_UNTIL_SYNCED, block_rx.recv()).await.is_ok() {}
```

crates/fuel-core/src/service/adapters.rs
```diff
@@ -87,3 +89,13 @@ impl P2PAdapter {
         Default::default()
     }
 }
+
+pub struct SyncAdapter {
+    block_rx: Receiver<Arc<ImportResult>>,
```
freesig, line 94: I think this can be a mpsc::Receiver? There's only one service that will subscribe right? Then you want get lagged errors.

leviathanbeak: I have used the one that the [Importer exposes](https://github.com/FuelLabs/fuel-core/blob/master/crates/services/importer/src/importer.rs#L109).

If it's only one service - then yeah we can switch the Importer code first.

crates/chain-config/src/config/consensus.rs
```diff
@@ -1,10 +1,33 @@
...
     Deserialize,
     Serialize,
 };
 
+use crate::default_consensus_dev_key;
+
 #[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq)]
 pub enum ConsensusConfig {
-    PoA { signing_key: Address },
+    PoA {
```
xgreenx, line 16: Because the network participants can't verify effects from other fields than `signing_key`, it can't be part of the `ChainConfig::ConsensusConfig`. 

It should be part of the `fuel_core_poa::Config`

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -137,40 +142,69 @@ pub struct Task<T, B, I> {
...
         let (request_sender, request_receiver) = mpsc::channel(100);
         let last_timestamp = last_block.time();
         let duration =
             Duration::from_secs(Tai64::now().0.saturating_sub(last_timestamp.0));
         let last_block_created = Instant::now() - duration;
+
+        let min_connected_reserved_peers = config.min_connected_reserved_peers;
+        let time_until_synced = config.time_until_synced;
+
+        let initial_sync_state =
```
xgreenx, line 172: It should be the internal logic of the `SyncTask`. The `sync_state` can be part of the `SyncTask::SharedState` and you can subscribe after creation fo the `SyncTask `=)

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -420,14 +454,14 @@ where
...
-impl<T, B, I> RunnableService for Task<T, B, I>
+impl<T, B, I> RunnableService for MainTask<T, B, I>
 where
     Self: RunnableTask,
 {
     const NAME: &'static str = "PoA";
 
     type SharedData = SharedState;
-    type Task = Task<T, B, I>;
+    type Task = MainTask<T, B, I>;
```
xgreenx, line 464: You also need to start the `SyncService` in the `into_task` function

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -457,13 +491,18 @@ where
...
+impl<D, T, B, I> RunnableTask for MainTask<T, B, I>
 where
     T: TransactionPool,
     B: BlockProducer<Database = D>,
     I: BlockImporter<Database = D>,
 {
     async fn run(&mut self, watcher: &mut StateWatcher) -> anyhow::Result<bool> {
+        // make sure we're synced first
+        if *self.sync_state.borrow() == SyncState::NotSynced {
+            self.sync_state.changed().await?;
```
xgreenx, line 503: It is better to handle `watcher` here too, maybe `sync_task` will take a lot of time to work=)

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -0,0 +1,563 @@
...
+                        self.state_sender.send(self.inner_state.sync_state())?;
+                    }
+                }
+
+            }
+            block = self.block_stream.next() => {
+                if let Some(new_block_height) = block {
+                    if self.inner_state.change_state_on_block(new_block_height) {
+                        self.set_timeout_on_state_change().await;
+                        self.state_sender.send(self.inner_state.sync_state())?;
```
xgreenx, line 93: It is better to use `send_if_modified` everywhere in our case=)

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -0,0 +1,563 @@
...
+            match self {
+                InnerSyncState::InsufficientPeers(_) => {
+                    *self = InnerSyncState::InsufficientPeers(new_block_height);
+                }
+                InnerSyncState::SufficientPeers(_) => {
+                    *self = InnerSyncState::SufficientPeers(new_block_height);
+                }
+                InnerSyncState::Synced(_) => {
+                    // we considered to be synced but we're obiously not!
+                    *self = InnerSyncState::SufficientPeers(new_block_height);
```
xgreenx, line 199: If you receive the block produced by yourself, then you don't need to change the state.

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -0,0 +1,563 @@
...
+        let mut state_changed = false;
+        let current_block_height = self.block_height();
+
+        if &new_block_height > current_block_height {
+            match self {
+                InnerSyncState::InsufficientPeers(_) => {
+                    *self = InnerSyncState::InsufficientPeers(new_block_height);
+                }
+                InnerSyncState::SufficientPeers(_) => {
+                    *self = InnerSyncState::SufficientPeers(new_block_height);
```
xgreenx, line 195: If you receive a new block, it should reset the timer always

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -0,0 +1,563 @@
...
+    ) -> bool {
+        let sufficient_peers = currently_connected_peers >= min_connected_reserved_peers;
+
+        match self {
+            InnerSyncState::InsufficientPeers(block_height) if sufficient_peers => {
+                *self = InnerSyncState::SufficientPeers(*block_height);
+                true
+            }
+            InnerSyncState::SufficientPeers(block_height)
+            | InnerSyncState::Synced(block_height)
```
xgreenx, line 221: Hmm, I think if we started producing blocks, then we should continue to do that even with the insufficient number of peers.

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -0,0 +1,563 @@
...
+            min_connected_reserved_peers,
+            time_until_synced,
+            block_stream,
+            state_sender,
+            inner_state,
+            timer,
+        }
+    }
+
+    pub async fn run(&mut self, watcher: &mut StateWatcher) -> anyhow::Result<bool> {
```
xgreenx, line 74: I see that you want to test each state transition separately in unit tests, but it seems it would be much simpler to have all transition code inside of the `SyncTask`. Maybe even all logic inside the `run` function(or you can split it into internal functions).

In the unit test, you can create a `SyncTask` instance and test each iteration of the `run` funciton.

crates/services/p2p/src/p2p_service.rs
```diff
@@ -522,14 +528,24 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
...
+                                    .connected_reserved_peers(&peer_id),
+                            })
                         }
                     }
                     PeerReportEvent::PeerDisconnected { peer_id } => {
                         if self.peer_manager.handle_peer_disconnect(peer_id) {
                             let _ = self.swarm.dial(peer_id);
                         }
-                        return Some(FuelP2PEvent::PeerDisconnected(peer_id))
+                        return Some(FuelP2PEvent::PeerDisconnected {
```
xgreenx, line 543: I think it would be better if the `PeerManager` itself provides a subscription. In this case, you don't need `.connected_reserved_peers(&peer_id)` function and the event will be simple as possible without propagation of the new information.

crates/services/consensus_module/poa/src/sync.rs
leviathanbeak: I have moved logic from `run()` method to `RunnableTask::run()` method, since having both didn't make much sense. Apart from that the code is already split into smaller encapsulated methods.

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -0,0 +1,594 @@
...
+                        } else {
+                            false
+                        }
+                    }) {
+                        self.update_timeout().await;
+                    }
+                }
+
+            }
+            block = self.block_stream.next() => {
```
xgreenx, line 143: The same here:

image

You can see that we can handle restarting of the timer in the case of `SufficientPeers` state.

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -0,0 +1,594 @@
...
+                        }
+                    });
+
+                    // update timeout on each received block
+                    self.update_timeout().await;
+
+                }
+            }
+            _ = self.timer.wait() => {
+                self.state_sender.send_if_modified(|sync_state: &mut SyncState| {
```
xgreenx, line 160: The idea of the `send_if_modified` to compare `sync_state: &mut SyncState` with a new value that you want to set=) Maybe someone already updated it before you in a separate thread.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -137,40 +142,65 @@ pub struct Task<T, B, I> {
...
+        p2p_port: P,
     ) -> Self {
         let tx_status_update_stream = txpool.transaction_status_events();
         let (request_sender, request_receiver) = mpsc::channel(100);
         let last_timestamp = last_block.time();
         let duration =
             Duration::from_secs(Tai64::now().0.saturating_sub(last_timestamp.0));
         let last_block_created = Instant::now() - duration;
+
+        let min_connected_reserved_peers = config.min_connected_reserved_peers;
```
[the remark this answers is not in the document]
leviathanbeak: Then SyncTask would need to be aware of whole Config which is unnecessary.
But I updated this method with deconstructing the Config so it looks cleaner.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -137,40 +142,65 @@ pub struct Task<T, B, I> {
...
+        let sync_task_handle = ServiceRunner::new(sync_task);
+        let sync_state = sync_task_handle.shared.clone();
+
         Self {
             block_gas_limit: config.block_gas_limit,
             signing_key: config.signing_key,
             txpool,
             block_producer,
             block_importer,
+            sync_state,
```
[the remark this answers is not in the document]
leviathanbeak: nice catch

crates/services/p2p/src/peer_manager.rs
```diff
@@ -73,16 +74,23 @@ impl PeerManager {
...
             non_reserved_connected_peers: HashMap::with_capacity(max_non_reserved_peers),
             reserved_connected_peers: HashMap::with_capacity(reserved_peers.len()),
             reserved_peers,
             connection_state,
             max_non_reserved_peers,
+            reserved_peers_updates,
         }
     }
 
+    pub fn reserved_peers_updates(&self) -> tokio::sync::broadcast::Sender<usize> {
```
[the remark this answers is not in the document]
leviathanbeak: While that makes sense, I used `Sender` here since that was the (only?) way to exposes `Receiver`s from `SharedState` since `Receiver` is not cloneable.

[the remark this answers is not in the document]
xgreenx: `Receiver` has `resubscribe` method=) 
image

crates/services/p2p/src/p2p_service.rs
```diff
@@ -955,23 +955,23 @@ mod tests {
         loop {
             tokio::select! {
                 event_from_node_a = node_a.next_event() => {
-                    if let Some(FuelP2PEvent::PeerConnected(_)) = event_from_node_a {
+                    if let Some(FuelP2PEvent::PeerConnected { .. }) = event_from_node_a {
```
xgreenx, line 958: We still can rollback all changes here=)

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -0,0 +1,370 @@
...
+            }
+            Some(block_info) = self.block_stream.next() => {
+                let new_block_height = block_info.height;
+
+                match self.inner_state {
+                    InnerSyncState::InsufficientPeers(block_height) if new_block_height > block_height => {
+                        self.inner_state = InnerSyncState::InsufficientPeers(new_block_height);
+                    }
+                    InnerSyncState::SufficientPeers(block_height) if new_block_height > block_height => {
+                        self.inner_state = InnerSyncState::SufficientPeers(new_block_height);
```
xgreenx, line 157: We also need to restart the timer=)

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -0,0 +1,377 @@
...
+                Poll::Pending
+            } else {
+                let next_item = this.items.pop_front();
+                Poll::Ready(next_item)
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn test_sync_task() {
```
xgreenx, line 300: It would be nice to add more tests for different scenarios, but we can do that in the follow-up PR.

Also, we need integration tests to check the behavior of two block producers.

xgreenx (approved):

PR #1170, opened 2023-05-02
v0.17.11 pyroscope

Voxelot:
This PR adds pyroscope to fuel-core-bin. Pyroscope is a platform that allows for continuous profiling, enabling real-time flamegraphs in deployed environments.

It also builds a separate docker image with debug symbols and profiling enabled, leaving the original docker image the same as before.

This is still in draft, and requires additional testing in the dev environment.

image

reopening of #1140, with a newer version of pyroscope which solves the openssl dependency issue

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -294,7 +294,9 @@ jobs:
         with:
           context: .
           file: deployment/Dockerfile
-          build-args: "FEATURES=production,profiling DEBUG_SYMBOLS=true"
+          build-args: |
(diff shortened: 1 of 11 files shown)


Voxelot:
Verified this is working in dev, however we need to update our grafana version to support the Phlare data connection to ingest pyroscope data.

Also needs a load tests to measure the overhead added by prof, and determine the appropriate sampling rate.

Voxelot:
closing in favor of #1324

Issue #1174, opened 2023-05-04
Help `fuel-indexer` fetch transactions efficiently 


ra0x3:
- Currently on the fuel-indexer, when fetching transactions for a block, we're grabbing the transaction fragments, and then making requests to fuel-core via the GQL client, for each transaction fragment
  - Meaning, we're making a GraphQL request for each transaction (not very efficient)
- It would be great if, for a given block, we could either fetch the transactions in bulk OR, if we could have the transactions included in the block itself
- We tried some previous work with this in https://github.com/FuelLabs/fuel-core/issues/764
  - This issue also references the original issue in fuel-indexer
- @xgreenx, @Voxelot mentions I should reach out to you personally about this :) 
- Happy to provide any other details

xgreenx:
You should already be able to fetch blocks with all transactions and any additional data via GraphQL. 
<img width="1673" alt="image" src="https://user-images.githubusercontent.com/18346821/236270955-ead1ab05-9ad8-4007-82e9-fceccfdb9ae7.png">

On the rust side, you can reuse types from `fuel-core-client` and compose them into the type that should fetch all data in one run. For example, you can define `FullBlock` type that fetches `Transaction` instead of `TransactionIdFragment`, which should return the full block for you. if you need some additional data from transactions, you can define your own `Mytransaction` that returns limited/extended fields.

<img width="591" alt="image" src="https://user-images.githubusercontent.com/18346821/236271163-d47e607b-5d00-4c68-804e-5933657bd58b.png">

xgreenx:
Hmm, the `FuelClient::query` function is private but with https://github.com/FuelLabs/fuel-core/pull/1175 you can write your own queries=)

xgreenx:
Solved on the indexer side https://github.com/FuelLabs/fuel-indexer/issues/288

PR #1175, opened 2023-05-04
Added example with custom query around the `fuel-core-client`

xgreenx:
Makes the `FuelClient::query` public to allow custom queries. Added an example of how to write custom queries for https://github.com/FuelLabs/fuel-core/issues/1174 but it requires access to the `schema.sdl`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2930,6 +2930,8 @@ dependencies = [
 name = "fuel-core-tests"
 version = "0.0.0"
 dependencies = [
+ "async-trait",
+ "cynic",
  "ethers",
  "fuel-core",
  "fuel-core-client",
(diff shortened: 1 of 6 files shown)


Voxelot:
Exposing the graphql schema directly isn't a long term viable solution given that we plan to transition off of graphql in the client eventually. Fetching a whole block should likely be a Rust-level API using primitives types to align with the work @bvrooman is doing.

xgreenx:
Of course, it is not a long-term solution=) But as a temporary solution, it will make the indexer team independent. And later, based on the API they require, we will put the same functionality in our Rust traits and implement it for RPC/GraphQL.

freesig (approved):

PR #1179, opened 2023-05-10
sync port with network info

leviathanbeak:
In this approach I have added NetworkInfo trait to SyncAdapter implementation,

it returns number of reserved peers that are currently connected, so when we are syncing with the network we first check the min amount of reserved peers are already connected.

### DIFF
--- a/crates/fuel-core/src/service/adapters.rs
+++ b/crates/fuel-core/src/service/adapters.rs
@@ -5,7 +5,10 @@ use crate::{
 use fuel_core_consensus_module::block_verifier::Verifier;
 use fuel_core_txpool::service::SharedState as TxPoolSharedState;
 use fuel_core_types::services::block_importer::ImportResult;
-use std::sync::Arc;
+use std::{
+    sync::Arc,
+    time::Duration,
+};
 use tokio::sync::broadcast::Receiver;
 
 pub mod block_importer;
@@ -118,6 +121,7 @@ pub trait NetworkInfo {
 pub struct SyncAdapter<T: NetworkInfo> {
     block_rx: Receiver<Arc<ImportResult>>,
     min_connected_reserved_peers: usize,
+    time_until_synced: Duration,
     network_info: T,
 }
 
@@ -125,11 +129,13 @@ impl<T: NetworkInfo> SyncAdapter<T> {
     pub fn new(
         block_rx: Receiver<Arc<ImportResult>>,
         min_connected_reserved_peers: usize,
+        time_until_synced: Duration,
         network_info: T,
     ) -> Self {
         Self {
             block_rx,
             min_connected_reserved_peers,
+            time_until_synced,
             network_info,
         }
     }
--- a/crates/fuel-core/src/service/adapters.rs
+++ b/crates/fuel-core/src/service/adapters.rs
@@ -2,6 +2,7 @@ use crate::{
     database::Database,
     service::Config,
 };
+use anyhow::Result;
 use fuel_core_consensus_module::block_verifier::Verifier;
 use fuel_core_txpool::service::SharedState as TxPoolSharedState;
 use fuel_core_types::services::block_importer::ImportResult;
@@ -89,11 +90,11 @@ impl P2PAdapter {
 #[cfg(feature = "p2p")]
 #[async_trait::async_trait]
 impl NetworkInfo for P2PAdapter {
-    async fn connected_reserved_peers(&self) -> usize {
+    async fn connected_reserved_peers(&self) -> anyhow::Result<usize> {
         if let Some(service) = self.service.as_ref() {
-            service.connected_reserved_peers_count().await.unwrap_or(0)
+            service.connected_reserved_peers_count().await
         } else {
-            0
+            anyhow::bail!("P2P is not enabled")
         }
     }
 }
@@ -108,20 +109,20 @@ impl P2PAdapter {
 #[cfg(not(feature = "p2p"))]
 #[async_trait::async_trait]
 impl NetworkInfo for P2PAdapter {
-    async fn connected_reserved_peers(&self) -> usize {
-        // returns fake number of connected peers - since p2p is not enabled at all
-        usize::MAX
+    async fn connected_reserved_peers(&self) -> Result<usize> {
+        Ok(0)
     }
 }
 
 #[async_trait::async_trait]
 pub trait NetworkInfo {
-    async fn connected_reserved_peers(&self) -> usize;
+    async fn connected_reserved_peers(&self) -> Result<usize>;
 }
 pub struct SyncAdapter<T: NetworkInfo> {
     block_rx: Receiver<Arc<ImportResult>>,
     min_connected_reserved_peers: usize,
     time_until_synced: Duration,
+    timeout_between_checking_peers: Duration,
     network_info: T,
 }
 
@@ -130,12 +131,14 @@ impl<T: NetworkInfo> SyncAdapter<T> {
         block_rx: Receiver<Arc<ImportResult>>,
         min_connected_reserved_peers: usize,
         time_until_synced: Duration,
+        timeout_between_checking_peers: Duration,
         network_info: T,
     ) -> Self {
         Self {
             block_rx,
             min_connected_reserved_peers,
             time_until_synced,
+            timeout_between_checking_peers,
             network_info,
         }
     }
--- a/crates/fuel-core/src/service/adapters.rs
+++ b/crates/fuel-core/src/service/adapters.rs
@@ -106,7 +106,8 @@ impl P2PAdapter {
 #[async_trait::async_trait]
 impl NetworkInfo for P2PAdapter {
     async fn connected_reserved_peers(&self) -> usize {
-        0
+        /// returns fake number of connected peers - since p2p is not enabled at all
+        usize::MAX
     }
 }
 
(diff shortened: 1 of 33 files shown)


crates/fuel-core/src/service/adapters.rs
```diff
@@ -83,19 +83,53 @@ impl P2PAdapter {
     }
 }
 
+#[cfg(feature = "p2p")]
+#[async_trait::async_trait]
+impl NetworkInfo for P2PAdapter {
+    async fn connected_reserved_peers(&self) -> usize {
+        if let Some(service) = self.service.as_ref() {
+            service.connected_reserved_peers_count().await.unwrap_or(0)
```
freesig, line 91: Why would this error and if it does is zero appropriate?

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -114,20 +116,27 @@ impl BlockImporter for BlockImporterAdapter {
...
-                    // time expired we are synced!
-                    break 'outer;
-                }
+        // todo: check if the next block to be produced equals last previous block + 1
+        // if yes, then we are synced already
+
+        // 1. check count of connected reserved peers
+        // todo: add 'n' tries or a timeout
+        loop {
+            let count = self.network_info.connected_reserved_peers().await;
```
freesig, line 127: If there is a real error in the nested function this will loop forever

freesig (dismissed):

crates/fuel-core/src/service/adapters.rs
leviathanbeak: Only reason would be because of the `oneshot` channels failing - nothing else.
So while it shouldn't fail, if it does returning 0 should cause the check to be repeated again (after some delay)

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
leviathanbeak: As mentioned in the above comment - the only "error" is oneshot channel erroring, that should not be the case, and if it fails once, it should fail the next time

crates/fuel-core/src/service/adapters.rs
freesig: Well what is holding the other end of the oneshot because if that panics then the oneshot will continuously fail.

xgreenx: I think Tommi's idea is that we want to know if it fails instead silently doing nothing. So maybe it is better to return `Result<usize, Err>` here and process the error on the `PoA` side by simple `?`. In this case the service will log this error, if something is wrong=)

crates/fuel-core/src/service/adapters.rs
```diff
@@ -83,19 +83,54 @@ impl P2PAdapter {
...
     }
 }
 
-pub struct SyncAdapter {
+#[cfg(not(feature = "p2p"))]
+#[async_trait::async_trait]
+impl NetworkInfo for P2PAdapter {
+    async fn connected_reserved_peers(&self) -> usize {
+        // returns fake number of connected peers - since p2p is not enabled at all
+        usize::MAX
```
xgreenx, line 110: I would prefer to add a new additional function, `is_enable`, that indicates whether p2p is enabled or not. In this case, we can return here 0, but on the PoA side, we can handle it.

Also, if p2p is disabled then sync service is disabled and we can use old logic=)

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -114,20 +116,27 @@ impl BlockImporter for BlockImporterAdapter {
...
+            if count >= self.min_connected_reserved_peers {
+                break
             }
+            tokio::time::sleep(Duration::from_millis(500)).await;
```
xgreenx, line 131: Hmm, it should be configurable

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -114,20 +116,27 @@ impl BlockImporter for BlockImporterAdapter {
...
+        }
+
+        // 2. receive all the blocks
+        while tokio::time::timeout(TIME_UNTIL_SYNCED, self.block_rx.recv())
```
xgreenx, line 135: And `TIME_UNTIL_SYNCED` also should be configurable.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -476,6 +479,7 @@ where
                 if let Some(request) = request {
                     match request {
                         Request::ManualBlocks((block, response)) => {
+                            let _ = self.sync_port.sync_with_peers().await;
```
xgreenx, line 482: We need to do that on each iteration of the `run` function, not only during manual block production

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
xgreenx: If `self.network_info.connected_reserved_peers` silently fails inside and returns `0`, it will be a hot loop that will consume a lot of CPU. I agree with Tommi that it is better to return an error.

You work with an interface here, and implementation always can change in the future

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -111,25 +111,34 @@ impl BlockImporter for BlockImporterAdapter {
...
-                    // keep receiving them blocks
-                }
-                _ = tokio::time::sleep(TIME_UNTIL_SYNCED) => {
-                    // time expired we are synced!
-                    break 'outer;
-                }
-            }
+#[cfg(feature = "p2p")]
+impl P2pPort for P2PAdapter {
+    fn reserved_peers_count(&self) -> Receiver<usize> {
```
xgreenx, line 116: We try not to use `tokio` types in the public API. It should return a stream as we do it in other cases.

Also, you don't need to return the closed stream if p2p is disabled. It should be an infinite stream like `pending`.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -140,29 +148,65 @@ pub struct Task<T, B, I> {
...
+        let initial_sync_state = {
+            if min_conneected_reserved_peers == 0 && time_until_sync == Duration::ZERO {
+                SyncState::Synced
+            } else {
+                SyncState::NotSynced
+            }
+        };
+
+        let (state_sender, sync_state) = tokio::sync::watch::channel(initial_sync_state);
+        let sync_notify = Arc::new(tokio::sync::Notify::new());
```
xgreenx, line 184: You can use only `watcher::changed` without `Notify`=)

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -140,29 +148,65 @@ pub struct Task<T, B, I> {
...
         let last_timestamp = last_block.time();
         let duration =
             Duration::from_secs(Tai64::now().0.saturating_sub(last_timestamp.0));
         let last_block_created = Instant::now() - duration;
+
+        // todo get from config
+        let min_conneected_reserved_peers = 0;
+        let time_until_sync = Duration::ZERO;
+
+        let initial_sync_state = {
```
xgreenx, line 175: You can set it `NotSynced` by default, and later it will be updated by the `SyncTask`

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -0,0 +1,245 @@
...
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum SyncState {
+    NotSynced,
+    Synced,
+}
+
+pub struct SyncTask<T> {
+    min_connected_reserved_peers: usize,
+    time_until_synced: Duration,
+    time_left_until_synced: Duration,
```
xgreenx, line 29: You can use a timer instead and `reset` it when you need it, like we already do for the `PoA`.

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -0,0 +1,245 @@
...
+            latest_count = reserved_peers_count.recv() => {
+                if let Ok(latest_count) = latest_count {
+                    if self.inner_state.change_state_on_peers_update(latest_count >= self.min_connected_reserved_peers) {
+                        self.state_sender.send(self.inner_state.sync_state())?;
+                    }
+                }
+            }
+            block = self.block_rx.recv() => {
+                if let Ok(block) = block {
+                    if self.inner_state.change_state_on_block(block) {
```
xgreenx, line 95: You can't change the state, because you are not connected to a sufficient amount of peers.

crates/services/consensus_module/poa/src/service.rs
leviathanbeak: nice, removed Notify

crates/services/consensus_module/poa/src/service.rs
leviathanbeak: While I changed a bit this code, wasn't the idea that if we start with 0 desired connected peers and time until sync is 0 - that we already consider to be Synced.

crates/services/consensus_module/poa/src/sync.rs
leviathanbeak: Yeah that makes it much cleaner

crates/services/consensus_module/poa/src/sync.rs
leviathanbeak: It does not change state per se, it tries to, but in case if the current state is `InnerSyncState::InsufficientPeers(_)` then it will just update the block height if it's bigger, but the state won't change.

xgreenx (approved):
Let's merge it into the original PR and continue review there=D

Issue #1155, opened 2023-04-24
ContractConfig should allow the contract id to be explicitly set


Voxelot:
Currently in `ContractConfig`, it computes the contract ID automatically by using all provided state as the initial state root. This undesirable because:

1. This may cause the contract ID to change on regenesis, breaking existing applications configured for a specific contract id.
2. It's very slow to compute the initial state root tree with large amounts of state
3. The amount of initial contract state is typically limited to what can fit inside of a transaction, making it impossible to regenesis a contract with a large amount of state.

Instead, the contract id should be an optional parameter that's used in place of autocalculated contract id if it is set. This way we don't need to preserve the initial state root of the contract if it's already included in the prior contract id.

PR #1184, opened 2023-05-22
Added `contract_id` to the `ContractConfig`

xgreenx:
Fixes https://github.com/FuelLabs/fuel-core/issues/1155


crates/chain-config/src/config/contract.rs
```diff
@@ -28,6 +28,9 @@ use serde_with::{
 #[serde_as]
 #[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq)]
 pub struct ContractConfig {
+    #[serde_as(as = "HexType")]
+    #[serde(default)]
+    pub contract_id: ContractId,
```
Voxelot, line 33: Should we make `contract_id` optional and automatically calculate it as we did before if it's set to None?

xgreenx: The primary purpose of the `ContractConfig` is the re-genesis of the blockchain, so I would prefer to keep it as a required field because each contract has a contract id used during initial creation.

Voxelot (approved):

PR #1185, opened 2023-05-23
Improve executor config

Voxelot:
1. Make executor use it's own config struct instead of reusing the fuel-core service config.
2. Make utxo_validation overridable on each execution, avoiding the need to create a fresh executor instance with a cloned config that's been slightly modified

This gives benefits such as:
  1. reduces the cost of dry-runs if the config object is large
  2. makes it easier to extract executor to a separate crate later on
  3. is a more sensible pattern to split parameters that change based on the caller vs ones that always stay the same

### DIFF
--- a/crates/fuel-core/src/executor.rs
+++ b/crates/fuel-core/src/executor.rs
@@ -480,6 +480,7 @@ where
         Ok(data)
     }
 
+    #[allow(clippy::too_many_arguments)]
     fn execute_transaction(
         &self,
         idx: u16,
@@ -609,6 +610,7 @@ where
         Ok(mint)
     }
 
+    #[allow(clippy::too_many_arguments)]
     fn execute_create_or_script<Tx>(
         &self,
         idx: u16,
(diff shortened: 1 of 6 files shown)


crates/fuel-core/src/executor.rs
```diff
@@ -458,6 +480,7 @@ where
         Ok(data)
     }
 
+    #[allow(clippy::too_many_arguments)]
```
xgreenx, line 483: Hmm, maybe it is better to create `ExecutorContext` that contains all data(like header, execution data, tx, etc) and implement all methods directly to this context. Because it seems we propagate a lot of the same object.

But it's okay to do that in follow-up PR, if you don't want to do that now=)

xgreenx (approved):

crates/fuel-core/src/executor.rs
Voxelot: Yeah I'll need to think on this, since each of these fn's has slightly different params. Although maybe there's a way to combine a few of them together.

Issue #1049, opened 2023-02-27
Transactions can be lost in the network


Voxelot:
Due to various reasons(disconnection at the moment of transaction broadcasting; non-actual state of the transaction broadcaster/receiver) transactions can be lost in the network.

For the cases like this, we need a transaction reconciliation protocol. Where missed transactions can be synced again.

freesig:
Goal: Make mempool more resilient.

MitchTurner:
Closed with #2131

PR #1187, opened 2023-05-24
Add test that show tx pool consistency can't be reached

freesig:
Currently this test fails at `Bob` because nodes score becomes too low to actually gossip with each other. They seem to be penalized for "mesh message delivery". They should not be being penalized so this is a bug @leviathanbeak 

If I manually set the penalty thresholds to super low then this test fails at the producer coming back online. This is the original issue the test was meant to show. This happens because "gossipsub" only tries to deliver messages for 3 heartbeats. So we can't rely on "gossipsub" to get the tx pools to consistency.

Note the solution to the tx pool consistency should be done at the app level not the network. So each tx pool should diff each others ids and then exchange messages each other don't have. This could be on a pretty slow loop.

Related issue #1049

### DIFF
--- a/crates/fuel-core/src/p2p_test_helpers.rs
+++ b/crates/fuel-core/src/p2p_test_helpers.rs
@@ -50,7 +50,10 @@ use rand::{
     SeedableRng,
 };
 use std::{
-    collections::HashMap,
+    collections::{
+        HashMap,
+        HashSet,
+    },
     ops::{
         Index,
         IndexMut,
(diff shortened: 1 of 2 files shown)


Voxelot:
One part of the issue might be that the bootstrap node isn't a real node and doesn't perform any application level message validation (ie. wait for the txpool to confirm the tx is valid). This test should probably wait until the validators are directly connected to each other.

xgreenx:
It was fixed during https://github.com/FuelLabs/fuel-core/pull/1434

PR #1191, opened 2023-05-25
feat: Use ID types in client api

bvrooman:


crates/client/src/client.rs
```diff
@@ -675,17 +691,17 @@ impl FuelClient {
     /// Retrieve coins to spend in a transaction
     pub async fn coins_to_spend(
         &self,
-        owner: &str,
-        spend_query: Vec<(&str, u64, Option<u64>)>,
+        owner: &Address,
+        spend_query: Vec<(AssetId, u64, Option<u64>)>,
         // (Utxos, Messages Nonce)
         excluded_ids: Option<(Vec<&str>, Vec<&str>)>,
```
Voxelot, line 697:
```suggestion
        excluded_ids: Option<(Vec<&UtxoId, Vec<&Nonce>)>,
```

crates/client/src/client.rs
```diff
@@ -1,12 +1,22 @@
...
+        coins::{
+            ExcludeInput,
+            SpendQueryElementInput,
+        },
+        contract::ContractBalanceQueryArgs,
+        tx::DryRunArg,
+        Tai64Timestamp,
+        TransactionId,
+    },
+    types::scalars::{
```
Voxelot, line 13: We should probably rename this module, as the notion of "scalars" is from GraphQL. However none of these are GraphQL types.

xgreenx: Something like `primitives`=)

crates/client/src/client.rs
```diff
@@ -750,10 +770,10 @@ impl FuelClient {
     // Retrieve a page of balances by their owner
     pub async fn balances(
         &self,
-        owner: &str,
+        owner: &Address,
         request: PaginationRequest<String>,
```
xgreenx, line 774: Hmm, maybe it also would be nice to use a real cursor instead of the `String`. If that possible, of course

xgreenx (commented):
Now the usage looks much, much better, thank you!=)

crates/client/src/client.rs
bvrooman: Thanks for the catch, will update

crates/client/src/client.rs
bvrooman: Updated

bvrooman: As discussed over a call, we can investigate this further at a separate time. I am creating this ticket to capture this tech debt: https://github.com/FuelLabs/fuel-core/issues/1197

crates/client/src/client.rs
```diff
@@ -693,24 +710,32 @@ impl FuelClient {
     /// Retrieve coins to spend in a transaction
     pub async fn coins_to_spend(
         &self,
-        owner: &str,
-        spend_query: Vec<(&str, u64, Option<u64>)>,
+        owner: &Address,
+        spend_query: Vec<(AssetId, u64, Option<u64>)>,
         // (Utxos, Messages Nonce)
-        excluded_ids: Option<(Vec<&str>, Vec<&str>)>,
+        excluded_ids: Option<(Vec<&UtxoId>, Vec<&Nonce>)>,
```
xgreenx, line 716:
```suggestion
        excluded_ids: Option<(Vec<UtxoId>, Vec<Nonce>)>,
```

xgreenx (approved):

PR #1192, opened 2023-05-26
Added `submit_and_await` endpoint to not miss the notifications

xgreenx:


crates/client/assets/schema.sdl
```diff
@@ -450,10 +450,16 @@ type Mutation {
...
-	Submits transaction to the txpool
+	Submits transaction to the `TxPool`.
+	
+	Returns submitted transaction if the transaction is included in the `TxPool` without problems.
 	"""
 	submit(tx: HexString!): Transaction!
 	"""
+	Submits transaction to the `TxPool` and await either confirmation or failure.
+	"""
+	submitAndAwait(tx: HexString!): TransactionStatus!
```
Voxelot, line 461: This should probably be a stream/subscription, as the connection is more likely to just timeout if the transaction is waiting in the transaction pool too long.

freesig: Yeh I agree

xgreenx: I wrote a test and checked that I didn't get a timeout with a 10-minute production interval. Maybe it works for me because of OS keeps alive default behavior.

But I changed to use a subscription model. Maybe we can get a timeout in the production in some other place like ingress nginx.

Voxelot (approved):

PR #1193, opened 2023-05-26
Update deploy-test-env.yml add error handling

d3287t328:
This is a rather simple boilerplate to allow users to adjust how they handle errors with this build. If this idea is useful to the other env's it can be added to the other workflow builds let me know and I can add it to those. I picked this file to make this suggested improvement on due to its low risk of security implications.


xgreenx:
It doesn't seem useful for us right now=)

PR #1196, opened 2023-05-31
Collect transaction bench output

freesig:


xgreenx:
We close it because related PR also is close for now https://github.com/FuelLabs/fuel-core/pull/1156#event-11412493150

PR #1198, opened 2023-06-01
Make `FuelClient::query` public and up version nto `0.17.12`

xgreenx:
The indexer team needs a public version of the `FuelClient::query` method to write custom queries. We don't need to deploy it to testnet, only a new version as a lirbary

### DIFF
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -164,7 +164,8 @@ impl FuelClient {
         Self::from_str(url.as_ref())
     }
 
-    async fn query<ResponseData, Vars>(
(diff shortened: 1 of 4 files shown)


crates/client/src/client.rs
```diff
@@ -164,7 +164,8 @@ impl FuelClient {
         Self::from_str(url.as_ref())
     }
 
-    async fn query<ResponseData, Vars>(
+    /// Send the GraphQL query ot the client.
```
Voxelot, line 167:
```suggestion
    /// Send the GraphQL query to the client.
```

xgreenx:
Released

PR #1202, opened 2023-06-08
setting peer reputation params

leviathanbeak:
Apart from the configuration of the reputation params, I have also removed `topics` from Cli configuration since for us they are fixed and  other parts of the code already assumed it - so it didn't make much sense having that as configurable


crates/services/p2p/src/gossipsub/config.rs
```diff
@@ -54,150 +102,104 @@ pub fn default_gossipsub_builder() -> GossipsubConfigBuilder {
...
-        /// The score threshold below which px will be ignored; this should be positive
-        /// and limited to scores attainable by bootstrappers and other trusted nodes.
-        accept_px_threshold: 10.0,
-
-        /// The median mesh score threshold before triggering opportunistic
-        /// grafting; this should have a small positive value.
-        opportunistic_graft_threshold: 20.0,
+        gossip_threshold: -4000.0,
+        publish_threshold: -8000.0,
+        graylist_threshold: GREYLIST_THRESHOLD,
```
bvrooman, line 200: Here `graylist_threshold` is spelt with an `a`, but `GREYLIST_THRESHOLD` is spelt with an `e`. Americanos will probably tell you `GRAYLIST` with an `a` is the correct spelling.

crates/services/p2p/src/gossipsub/config.rs
```diff
@@ -54,150 +102,104 @@ pub fn default_gossipsub_builder() -> GossipsubConfigBuilder {
...
-/// This function takes in `max_gossipsub_score` and sets it to `topic_score_cap`
-/// The reasoning is because app-specific is set to 0 so the max peer score
-/// can be the score of the topic score cap
-fn initialize_peer_score_params() -> PeerScoreParams {
-    PeerScoreParams {
-        // topics are added later
-        topics: Default::default(),
+fn score_parameter_decay(decay_time: Duration) -> f64 {
+    let ticks = decay_time.as_secs_f64() / DECAY_INTERVAL.as_secs_f64();
+    DECAY_TO_ZERO.powf(1.0 / ticks)
```
bvrooman, line 164: Are there any scenarios where `ticks == 0`?

crates/services/p2p/src/gossipsub/config.rs
```diff
@@ -54,150 +102,104 @@ pub fn default_gossipsub_builder() -> GossipsubConfigBuilder {
...
+}
 
-        /// Aggregate topic score cap; this limits the total contribution of topics towards a positive
-        /// score. It must be positive (or 0 for no cap).
-        topic_score_cap: MAX_GOSSIP_SCORE,
+fn decay_convergence(decay: f64, rate: f64) -> f64 {
+    rate / (1.0 - decay)
```
bvrooman, line 168: Any scenarios where `decay == 1`?

crates/services/p2p/src/gossipsub/config.rs
leviathanbeak: ah yes I should follow whatever the config is doing, which is surprising since -> https://docs.rs/libp2p/latest/libp2p/swarm/trait.NetworkBehaviour.html 😁

crates/services/p2p/src/gossipsub/config.rs
leviathanbeak: there shouldn't be, this is only a helper function used during setup, it will not called ever again.

crates/services/p2p/src/gossipsub/config.rs
```diff
@@ -25,9 +25,57 @@ use sha2::{
...
+    GossipTopic,
+    CON_VOTE_GOSSIP_TOPIC,
+    NEW_BLOCK_GOSSIP_TOPIC,
+    NEW_TX_GOSSIP_TOPIC,
+};
+
+// The number of slots in each epoch.
+const SLOTS_PER_EPOCH: u64 = 32;
+
+// The duration of each slot in seconds. This is the amount of time allotted for each opportunity to create a block.
```
xgreenx, line 38: What is the "block" here?

crates/services/p2p/src/gossipsub/config.rs
```diff
@@ -25,9 +25,57 @@ use sha2::{
...
+const SLOT: Duration = Duration::from_secs(6);
```
xgreenx, line 39: What is this "mesh" and why previously used 1 millisecond and now we want to use 6 seconds?

crates/services/p2p/src/gossipsub/config.rs
```diff
@@ -239,23 +241,43 @@ pub(crate) fn build_gossipsub_behaviour(p2p_config: &Config) -> Gossipsub {
     }
 }
 
-fn initialize_gossipsub(gossipsub: &mut Gossipsub, p2p_config: &Config) {
-    let peer_score_params = initialize_peer_score_params();
+const EXPECTED_TXS_PER_SLOT: f64 = 20.0;
```
xgreenx, line 244: It is almost impossible to predict how many transactions we expect. It fully depends on `max_tx_limit` and `max_gas_limit`.

crates/services/p2p/src/gossipsub/config.rs
leviathanbeak: SLOT represents a unit of time, in this case it represents time between two produced fuel blocks

crates/services/p2p/src/gossipsub/config.rs
leviathanbeak: mesh is the gossipsub mesh, time in mesh is the time since the peer has been part of the gossipsub mesh, previously we used the default value, but now we want a more realistic value for our usecase, which would be at least the length between two blocks

crates/services/p2p/src/gossipsub/config.rs
leviathanbeak: we can have some healthy estimate of amount of tx messages we expect to receive from a single peer within a single SLOT, it does not have to be perfect

xgreenx: But how does it affect gossiping? How will it behave if we have more than `20` transactions? And what if less? Sorry, I don't know how it works inside, and I'm trying to get it=)

leviathanbeak: For a given N amount of messages in SLOT period, gossipsub will analyze and consequently score peers and their performances. For example if we say we expect 20 messages, and the peer keeps delivering way less than that then it's reputation will decrease. Having said that, if we struggle to have an estimate for txs (unlike for some other messages) then we can also disable peer scoring for this `topic`

xgreenx (dismissed):

xgreenx (approved):

PR #1207, opened 2023-06-11
SMT storage key hashing

xgreenx:
This PR uses an upcoming release of the `fuel-vm 0.34`.

It brings a new [important feature](https://github.com/FuelLabs/fuel-vm/pull/485) - hashing the leaf key for the SMT to prevent tree structure manipulation.

As an example: Hashed storage key decreases the number of nodes in the SMT from 1.3M to 70K for 30K leaves in the `run_contract_large_state` e2e test. It improves the test time from 200 seconds to 13 seconds in Debug mode and to 1 second in the release(instead of 20 seconds).

So it improves security and performance.

The change is breaking because it affects the `state_root` field -> generated `ContractId`.

Fixes https://github.com/FuelLabs/fuel-core/issues/1143


crates/fuel-core/src/database/state.rs
```diff
@@ -144,6 +153,52 @@ impl MerkleRootStorage<ContractId, ContractsState> for Database {
     }
 }
 
+impl Database {
+    /// Initialize the state of the contract from the all leafs.
```
bvrooman, line 157:
```suggestion
    /// Initialize the state of the contract from the all leaves.
```

crates/fuel-core/src/database/state.rs
```diff
@@ -144,6 +153,52 @@ impl MerkleRootStorage<ContractId, ContractsState> for Database {
...
+    /// This method is more perform than insert state one by one.
```
bvrooman, line 158:
```suggestion
    /// This method is more performant than inserting state one by one.
```

crates/fuel-core/src/database/balances.rs
```diff
@@ -146,6 +156,53 @@ impl MerkleRootStorage<ContractId, ContractsAssets> for Database {
...
+        balances: S,
+    ) -> Result<(), StorageError>
+    where
+        S: Iterator<Item = (AssetId, Word)>,
+    {
+        if self
+            .storage::<ContractsAssetsMerkleMetadata>()
+            .contains_key(contract_id)?
+        {
+            Err(anyhow::anyhow!(
```
leviathanbeak, line 174: nit: I'd prefer explicit `return` over `?` but up to you

crates/fuel-core/src/database/state.rs
```diff
@@ -144,6 +153,52 @@ impl MerkleRootStorage<ContractId, ContractsState> for Database {
...
+        slots: S,
+    ) -> Result<(), StorageError>
+    where
+        S: Iterator<Item = (Bytes32, Bytes32)>,
+    {
+        if self
+            .storage::<ContractsStateMerkleMetadata>()
+            .contains_key(contract_id)?
+        {
+            Err(anyhow::anyhow!("The contract state is already initialized"))?;
```
leviathanbeak, line 171: same here

crates/fuel-core/src/service.rs
```diff
@@ -196,9 +196,10 @@ pub struct Task {
 
 impl Task {
     /// Private inner method for initializing the fuel service task
-    pub fn new(database: Database, config: Config) -> anyhow::Result<Task> {
+    pub fn new(database: Database, mut config: Config) -> anyhow::Result<Task> {
         // initialize state
         genesis::maybe_initialize_state(&config, &database)?;
+        config.chain_conf.initial_state = None;
```
leviathanbeak, line 202: What's the `initial_state` before this line? Isn't there any other way for this to get initialized to desired state or it has to go from Some<T> to None in here?

leviathanbeak (approved):
small nits - looks good

leviathanbeak (approved):

Issue #1217, opened 2023-06-18
Take into account the actually used gas by the transactions and fetch more transaction


xgreenx:
Beta 3 and upcoming beta 4 suffer the problem with transactions with a 10M gas limit, but the actual used gas is 5K. TX pool only returns 50 transactions because the max gas limit of the block is 500M. It decreases the tps of the blockchain, and users can't commit a lot of transactions.

We need to update the transaction selection logic to be a stream like in https://github.com/FuelLabs/fuel-core/issues/1206. This issue tracks only the streaming implementation of the transaction selection, and the soft commit part should be done during the parent issue.

Also, we need to patch the latest `0.17` and `0.18` releases with this fix.

During development, we need to consider that this stream will be later used during [parallel execution](https://github.com/FuelLabs/fuel-core/issues/1088).

PR #1220, opened 2023-06-20
Take into account used gas during block production and include more transactions if we can

xgreenx:
The PR implements https://github.com/FuelLabs/fuel-core/issues/1217 for the beta 3 testnet

### DIFF
--- a/crates/fuel-core/src/executor.rs
+++ b/crates/fuel-core/src/executor.rs
@@ -83,6 +83,7 @@ use fuel_core_types::{
         PredicateStorage,
     },
     services::{
+        block_producer::Components,
         executor::{
             Error as ExecutorError,
             ExecutionBlock,
@@ -99,8 +100,7 @@ use fuel_core_types::{
         txpool::TransactionStatus,
     },
 };
-use itertools::Itertools;
-pub use ports::RelayerPort;
+use parking_lot::Mutex as ParkingMutex;
 use std::{
     ops::{
         Deref,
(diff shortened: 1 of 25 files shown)


crates/fuel-core/src/service/adapters.rs
```diff
@@ -33,6 +37,24 @@ impl TxPoolAdapter {
...
+}
+
+impl TransactionsSource {
+    pub fn new(
+        txpool: TxPoolSharedState<P2PAdapter, Database>,
+        block_height: BlockHeight,
+    ) -> Self {
+        Self {
+            txpool,
+            _block_height: block_height,
```
Voxelot, line 53: Looks like block_height isnt used, is the idea for adding it here to enable selecting txs based on maturity requirements?

xgreenx: Yes:) The code before did the same
IMG_1069

crates/services/producer/src/ports.rs
```diff
@@ -39,13 +39,15 @@ pub trait BlockProducerDatabase: Send + Sync {
 
 #[async_trait]
 pub trait TxPool: Send + Sync {
-    fn get_includable_txs(
+    /// The source of the transactions used by the executor.
+    type TxSource;
+
+    /// Returns teh source of includable transactions.
```
leviathanbeak, line 45:
```suggestion
    /// Returns the source of includable transactions.
```

crates/fuel-core/src/executor.rs
```diff
@@ -157,7 +184,21 @@ where
...
+                gas_limit: u64::MAX,
+            }),
+            ExecutionBlock::Production(block) => ExecutionTypes::Production(Components {
+                header_to_produce: block.header,
+                transactions_source: OnceTransactionsSource::new(block.transactions),
+                gas_limit: u64::MAX,
+            }),
+            ExecutionBlock::Validation(block) => ExecutionTypes::Validation(block),
+        };
+
```
leviathanbeak, line 200: nit: maybe this From<ExecutionBlock> could be extracted

xgreenx: It is only used once here for unit tests, and it is better to do that explicitly because it does not apply to other places=)

leviathanbeak (approved):

Voxelot (approved):

PR #1228, opened 2023-06-23
WIP: Pooled transactions exchange protocol

digorithm:
**Work in progress**

## Scope and design

Partially closes https://github.com/FuelLabs/fuel-core/issues/1049. I say partially because this work will be split into two:

1. An initial pooled transactions "handshake" between two newly connected peers, where these two peers will share their pooled transactions. **This is the scope of this PR**. Then, in a follow-up PR:
2. Another sub-service will periodically request pooled transactions from a random, small subset of known peers and compare and share their pooled transactions like the first part of the protocol.
3. More details in the RFC.

These two combined will lead to the mempool being more resilient. 

## Tasks and TODOs

- [x] Create and organize the new `TxPoolSync` service under the current architecture;
- [x] Setup the communication between the new service and the main transaction pool service, and the p2p service (the majority of this work falls under this task);
- [x] Implement the pooled transaction sharing
  - [x] Current blocker: node gets penalized and disconnected when sending transactions
- [x] Write new tests
- [ ] Fix old tests and CI-related stuff broken due to these changes;
- [ ] Clean-up code (`TODOs`, improve quality, etc.)
- [ ] Write more documentation;


bvrooman:
To make things a bit easier to review, is it possible/feasible to separate out the nonfunctional formatting changes into a separate PR?

digorithm:
> To make things a bit easier to review, is it possible/feasible to separate out the nonfunctional formatting changes into a separate PR?

Yup, I'll revert those; those were automatically done by my editor + my current version of the formatter.  I'll do it asap

digorithm:
Quick update on our strategy: we'll _request_ the pooled transactions upon connection instead of sending them immediately. The reason is because returning variable-sized responses works better for our current configuration, which is @Voxelot's finding: 

> it would probably work better for peers to ask for the pooled txs and return them as a variable sized response instead of pushing the tx_ids to peers right away

xgreenx:
Closing in favor of https://github.com/FuelLabs/fuel-core/pull/2131. We still need to implement part of the functionality from here, so we can use this PR as a source later.

PR #1231, opened 2023-06-28
Speedup and stabilize unit and integration tests

xgreenx:
Speedup and stabilize unit and integration tests:
- Removed usage of `mdns = true` from all test except one(which tests it)
- `P2PService::start`is async function now and awaits getting of any address
- Updated some p2p tests to have more strict conditions to exit.

### DIFF
--- a/crates/services/p2p/src/p2p_service.rs
+++ b/crates/services/p2p/src/p2p_service.rs
@@ -262,6 +262,7 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
         }
     }
 
+    #[cfg(feature = "test-helpers")]
     pub fn multiaddrs(&self) -> Vec<Multiaddr> {
         let local_peer = self.local_peer_id;
         self.swarm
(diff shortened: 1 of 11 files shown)


crates/services/p2p/src/p2p_service.rs
```diff
@@ -707,77 +731,36 @@ mod tests {
...
+    async fn setup_bootstrap_nodes(
+        p2p_config: &Config,
+        bootstrap_nodes_count: usize,
+    ) -> (Vec<P2PService>, Vec<Multiaddr>) {
+        let nodes = join_all(
+            (0..bootstrap_nodes_count)
+                .map(|_| build_service_from_config(p2p_config.clone())),
+        )
+        .await;
+        let bootstarp_multiaddrs = nodes
```
bvrooman, line 756:
```suggestion
        let bootstrap_multiaddrs = nodes
```

leviathanbeak (dismissed):

leviathanbeak (approved):

Issue #1221, opened 2023-06-21
Expose `GasCost` from the `chain_info` endpoint


xgreenx:
The SDK can't get `GasCosts` from the `fuel-core` because we don't expose this information. The `chain_info` endpoint should also include `GasCosts` along with `ConsensusParametrs`

danielbate:
Is there any update on this? I believe this is blocking us as we are not retrieving a `predicateGasUsed` property from the `estimate-predicates` endpoint.

xgreenx:
The `estimate_predicates` should return transaction with estimated predicates to you=)
<img width="735" alt="image" src="https://github.com/FuelLabs/fuel-core/assets/18346821/49c1e215-8e6d-49ce-8587-da16fa86aee8">

danielbate:
Thankyou, this has now been resolved for us, please disregard

lucasgstz:
Good one Daniel, i was looking all those emails just now... No sense at all...
N1
TY

________________________________
De: Daniel Bate ***@***.***>
Enviado: terça-feira, 11 de julho de 2023 10:07
Para: FuelLabs/fuel-core ***@***.***>
Cc: Subscribed ***@***.***>
Assunto: Re: [FuelLabs/fuel-core] Expose `GasCost` from the `chain_info` endpoint (Issue #1221)


Thankyou, this has now been resolved for us, please disregard my comment

xgreenx:
Fixed with https://github.com/FuelLabs/fuel-core/pull/1244

PR #1244, opened 2023-07-11
Expose gas cost in chain info

MitchTurner:
https://github.com/FuelLabs/fuel-core/issues/1221

This PR adds:
- new GraphQL schema for `GasCosts` inside the `ChainInfo` schema
- public `GasCosts` types for client
- appropriate conversions
- check in tests to confirm gas costs are included in chain info


crates/client/src/client/types/gas_costs.rs
```diff
@@ -0,0 +1,345 @@
+use crate::client::schema;
+
+pub struct GasCosts {
```
xgreenx, line 3: You can create a local macro and automate `From` implementation=) We did the same [here](https://github.com/FuelLabs/fuel-vm/pull/499/files#diff-b5eeedfa720b67c4498e4b6d75e8bf7019ee7d60494d7511f86854118566b888)

xgreenx (commented):
Looks good=) Only one comment regarding `From` trait implementation

crates/client/src/client/types/gas_costs.rs
MitchTurner: Would you prefer I do a macro? I just let copilot do the work for me :).

Voxelot: A macro might be easier to maintain as we add new fields in the future

MitchTurner: This case is a bit different from the example you gave since it's a struct instead of an enum. I believe it is still possible to create a proc macro that would iterate over all the fields (see: https://github.com/dtolnay/syn/issues/516). Idk if I'm convinced it's worth my time though--I don't mind the current verbosity.

MitchTurner:
This PR was broken so created new one: https://github.com/FuelLabs/fuel-core/pull/1245

MitchTurner:
New PR didn't fix it. Reopening to keep comments.

crates/client/src/client/types/gas_costs.rs
xgreenx: You still can do it with `macro_rules` without proc macro=)

MitchTurner: Done.

xgreenx (approved):
Thank you, looks good!=)

Issue #1112, opened 2023-04-09
Reuse calculated tx id in executor


Voxelot:
Maybe we can calculate it one time at the begin of the `execute_create_or_script` method and pass it everywhere?

_Originally posted by @xgreenx in https://github.com/FuelLabs/fuel-core/pull/1107#discussion_r1160938395_

xgreenx:
Fixed by https://github.com/FuelLabs/fuel-core/pull/1248

PR #1248, opened 2023-07-12
Reuse calculated tx id in executor

MitchTurner:
https://github.com/FuelLabs/fuel-core/issues/1112

### DIFF
--- a/crates/fuel-core/src/executor.rs
+++ b/crates/fuel-core/src/executor.rs
@@ -65,6 +65,7 @@ use fuel_core_types::{
         Address,
         AssetId,
         Bytes32,
+        Cacheable,
         Input,
         Mint,
         Output,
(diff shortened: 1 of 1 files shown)


xgreenx (commented):
Could you also update places like this?
image

MitchTurner:
Yes. I had it originally and then removed it because I thought there was some ordering issues, but it was a simple fix in the end.

MitchTurner:
Actually. I was right. The `compute_inputs` needs to happen before the `into_checked_basic` because the tx is being mutated.

So we can't get the tx id from the `checked_tx`, and if we were to add a tx_id calculation before `compute_inputs` it would get called twice every time. I recommend keeping it as is since it only get's called once, either when it errors or when it does the tx_check.

xgreenx:
The `into_checked_basic` shouldn't modify `tx_id`, otherwise something is wrong in our logic=D

MitchTurner:
Sorry. That's not what I'm saying.

`compute_inputs` modifies the tx. And that's the method that you are suggesting passing `tx_id` into. As you can see, the tests are failing if I do `into_check_basic` before `compute_inputs`.

If `tx_id` is calculated during `into_check_basic` and `compute_inputs` needs to happen before `into_check_basic` I don't recommend calculating `tx_id` before `compute_inputs` and only calling `tx.id()` if an error occurs.

That is how the current code works. So, I'm suggesting we don't make this change.

xgreenx:
Modification from `compute_inputs` doesn't affect transaction id. You don't need to do `into_check_basic` before

xgreenx:
After refactoring the `TxPool`, we don't need to call `into_check_basic`, so in the future, we will use a cached value each time=)

The idea is to minimize the number of `.id(&self.config.transaction_parameters.chain_id)` in the code. Because without this change, we already use cached value everywhere. We receive transactions from `TxPool`, and their id is calculated there.

MitchTurner:
Okay. Cool, if `original_tx` has a cached id, then I take it back. I'm comfortable with that!

xgreenx (approved):

PR #1249, opened 2023-07-12
Reused `CheckedTransaction` from transaction pool in the executor

xgreenx:
The change adds a `MaybeCheckedTransaction` type that is either `CheckedTransaction` or `Transaction`.

We need it because the validation of the block works with the `Transaction` type, while block production works with `CheckedTransaction`(TxPool has already validated transactions before during insertion).

The executor calls `check_signatures` and `check_predicates` that inside only will be executed if the flag is not set.
image
image

It should speedup block production


crates/fuel-core/src/executor.rs
```diff
@@ -631,54 +649,85 @@ where
...
             .storage::<Transactions>()
-            .contains_key(&tx_id)?
+            .contains_key(tx_id)?
         {
-            return Err(ExecutorError::TransactionIdCollision(tx_id))
+            return Err(ExecutorError::TransactionIdCollision(*tx_id))
         }
 
+        let block_height = *header.height();
         match tx {
```
Voxelot, line 670: Maybe we should try to reduce some of the duplication in these match arms?

For example, we could first turn a MaybeCheckedTransaction into a CheckedTransaction, and then separately execute the normalized CheckedTransaction.

Voxelot (dismissed):

Voxelot (approved):

Issue #1186, opened 2023-05-23
Add logging of the long GraphQL queries for future debug


xgreenx:
We need to add detailed logging into the resolver for long GraphQL queries for future debugging.

Agilulfe:
I can happily give it a go.

xgreenx:
You can try to finish it, but we plan to start work on it the next week, so you need to hurry=)

Agilulfe:
You can go for it then. I thought it was stale and wanted to casually get my hands dirty.

xgreenx:
Done with https://github.com/FuelLabs/fuel-core/pull/1250

PR #1250, opened 2023-07-13
Add logging of the long GraphQL queries for future debug

MitchTurner:
https://github.com/FuelLabs/fuel-core/issues/1186

This PR:
- Creates new extension for logging
- Adds new extension to GraphQL API
- Allows caller of `new_service` to customize the `log_threshold_ms` for how long a query must take for it to get logged


MitchTurner:
I left a TODO on the actual logging because I wasn't sure if there was more details we wanted to include.

leviathanbeak (commented):
looks good, will wait for removal of `todo` to approve

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -38,6 +38,8 @@ pub type BlockProducerService = fuel_core_producer::block_producer::Producer<
 >;
 pub type GraphQL = crate::fuel_core_graphql_api::service::Service;
 
+const LOG_THRESHOLD_MS: u64 = 2000;
```
xgreenx, line 41: It would be nice if you made it configurable via CLI arguments=) Also you can use `Duration`

crates/fuel-core/src/graphql_api/logging.rs
```diff
@@ -0,0 +1,74 @@
...
+        ctx: &ExtensionContext<'_>,
+        info: ResolveInfo<'_>,
+        next: NextResolve<'_>,
+    ) -> ServerResult<Option<Value>> {
+        let start_time = Instant::now();
+        let result = next.run(ctx, info).await;
+        let elapsed = start_time.elapsed();
+
+        if elapsed > self.log_threshold_ms {
+            // TODO: Add more details to log?
```
xgreenx, line 64: We need to log all information required for reproduction: the body of the request, path, and maybe transaction(if it is `dry_run`).

crates/fuel-core/src/graphql_api/logging.rs
```diff
@@ -0,0 +1,74 @@
...
+    ServerResult,
+    Value,
+};
+use std::{
+    sync::Arc,
+    time::Duration,
+};
+use tokio::time::Instant;
+
+pub(crate) struct LoggingExtension {
```
xgreenx, line 20: I like the idea of combining `LoggingExtension` and `PrometheusExtension` into `LogsAndMetricsExtension` or `MetricsExtension`. In this case, you will spawn only one `Instance`=)

xgreenx (commented):
It would be nice if you tried to test it locally and put a screenshot with the output=)

You can start a node in debug mode and put a breakpoint somewhere in the endpoint to simulate long query=)

crates/fuel-core/src/service/sub_services.rs
MitchTurner: No problem. I'll push it up to the CLI level! Should I include a default for the CLI/make the field optional?

crates/fuel-core/src/graphql_api/logging.rs
MitchTurner: I went with separate because combining them feels like a violation of the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).

If performance with `Instance` is a big concern, I'm happy to reconsider. But I just default to architecturally cleanliness over performance.

crates/fuel-core/src/service/sub_services.rs
xgreenx: Yep, let's use 2 seconds as a default value for now=)

crates/fuel-core/src/graphql_api/logging.rs
xgreenx: The logging of long queries is more about metrics than logs. Plus, we have a separate `Tracing` extension for logs=)

Yeah, it is not integrated with Prometheus, but if we rename the extension into `MetricsExtension`, then it would be ergonomic=)

image

crates/fuel-core/src/graphql_api/logging.rs
MitchTurner: I'm not sure I know what the "path" and the "transaction" are in this case. I think I have the request working though :+1:

MitchTurner:
image

It's a little noisy, I assume because the resolvers are recursive. But that's probably useful so we can identify which request in the recursion.

Voxelot:
It might be good to log this all in one line. Otherwise if there are a lot of concurrent requests the logs linking the query args could be out of order.

MitchTurner:
Need to rename `prometheus.rs`

MitchTurner:
What do we prefer? 

Readable:
image

or compact:
image

xgreenx (dismissed):

crates/fuel-core/src/graphql_api/service.rs
```diff
@@ -161,6 +162,7 @@ pub fn new_service(
     txpool: TxPool,
     producer: BlockProducer,
     consensus_module: ConsensusModule,
+    log_threshold_ms: Duration,
```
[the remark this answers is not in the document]
MitchTurner: Isn't this syntax for vars that are unused? I'm missing why I'd change it here.

xgreenx (approved):

PR #1252, opened 2023-07-14
Migrate to buildjet cache

Voxelot:
github cache has been suffering from performance issues and the cache team at github has been very inactive. Testing out migrating to alternate CI caching schemes to improve performance.

see https://buildjet.com/for-github-actions/blog/launch-buildjet-cache


Voxelot:
Overall the results are looking much faster, several jobs that would take 4-8m on a warm cache finish in under 1m now.

Voxelot:
fixed the docker build cache - this is faster for now but may need some adjustments in the future to use branch specific cache restoration keys to avoid globbing all branches together into the same cache. But for now the bandwidth hit of a big cache isn't a bottleneck due to buildjet's cache performance.

xgreenx (approved):

Dentosal (approved):

PR #1253, opened 2023-07-14
parallel predicate verification

### LINKED ISSUE #1087: Parallel Predicate Execution
Predicates verification can be done in parallel, as they are stateless and don't depend on the results of each other. Devise a means to perform predicate verification in parallel for both the txpool and the primary block executor.

leviathanbeak:
closes #1087 on the fuel-core side


crates/services/txpool/src/service.rs
```diff
@@ -310,14 +325,28 @@ where
 impl<P2P, DB> SharedState<P2P, DB>
 where
     P2P: PeerToPeer<GossipedTransaction = TransactionGossipData>,
-    DB: TxPoolDb,
+    DB: TxPoolDb + Clone,
```
xgreenx, line 328:
```suggestion
    DB: TxPoolDb,
```

crates/services/txpool/src/service.rs
```diff
@@ -254,7 +269,7 @@ where
 //  `StorageInspect` trait.
 impl<P2P, DB> SharedState<P2P, DB>
 where
-    DB: TxPoolDb,
+    DB: TxPoolDb + Clone,
```
xgreenx, line 272:
```suggestion
    DB: TxPoolDb,
```

crates/services/txpool/src/txpool.rs
```diff
@@ -391,6 +364,96 @@ where
...
+    };
+
+    Ok(tx)
+}
+
+fn verify_tx_min_gas_price(tx: &Transaction, config: &Config) -> Result<(), Error> {
+    let price = match tx {
+        Transaction::Script(script) => script.price(),
+        Transaction::Create(create) => create.price(),
+        Transaction::Mint(_) => unreachable!(),
```
xgreenx, line 421:
```suggestion
        Transaction::Mint(_) => return Err(anyhow::anyhow!("Mint transactions is not supported")),
```

crates/services/txpool/src/txpool.rs
```diff
@@ -391,6 +364,96 @@ where
...
+        let tx = tx
+            .into_checked_basic(current_height, consensus_params)?
+            .check_signatures(&consensus_params.chain_id)?;
+
+        tx.check_predicates_async::<TokioWithRayon>(
+            consensus_params,
+            &config.chain_config.gas_costs,
+        )
+        .await?
+        .into()
```
xgreenx, line 405: It would be nice to add `debug_assert` here to verify that bitflag `Checks::All` is fulfilled. In the future, it will simplify adding new stages to the validation.

crates/services/txpool/src/txpool.rs
```diff
@@ -71,41 +79,17 @@ where
         &self.by_dependency
     }
 
-    #[tracing::instrument(level = "info", skip_all, fields(tx_id = %tx.id(&self.config.chain_config.transaction_parameters.chain_id)), ret, err)]
+    pub fn current_height(&self) -> anyhow::Result<BlockHeight> {
```
xgreenx, line 82: You don't need it anymore

crates/services/txpool/src/txpool.rs
```diff
@@ -63,6 +67,10 @@ where
         }
     }
 
+    pub fn config(&self) -> &Config {
```
xgreenx, line 70: You don't need it anymore

crates/services/txpool/src/txpool.rs
```diff
@@ -71,41 +79,17 @@ where
         &self.by_dependency
     }
 
-    #[tracing::instrument(level = "info", skip_all, fields(tx_id = %tx.id(&self.config.chain_config.transaction_parameters.chain_id)), ret, err)]
+    pub fn current_height(&self) -> anyhow::Result<BlockHeight> {
+        Ok(self.database.current_block_height()?)
+    }
+
+    // #[tracing::instrument(level = "info", skip_all, fields(tx_id = %tx.id(&self.config.chain_config.transaction_parameters.chain_id)), ret, err)]
```
xgreenx, line 86: Can we return it back?=)

crates/services/txpool/src/txpool.rs
```diff
@@ -233,37 +217,26 @@ where
...
-        Ok(())
-    }
-
     #[tracing::instrument(level = "info", skip_all)]
     /// Import a set of transactions from network gossip or GraphQL endpoints.
     pub fn insert(
         &mut self,
         tx_status_sender: &TxStatusChange,
-        txs: &[Arc<Transaction>],
+        txs: Vec<anyhow::Result<CheckedTransaction>>,
```
xgreenx, line 225: I know that you try not to break the old behavior. But I think it is better to pass only transactions which passed the validation. Otherwise, we call `lock` on the `TxPool`, do nothing, and unlock it. If we have many `submit` requests with invalid transactions, each of them will block the `TxPool` while we can avoid it.

crates/services/txpool/Cargo.toml
```diff
@@ -18,8 +18,11 @@ fuel-core-metrics = { workspace = true }
 fuel-core-services = { workspace = true }
 fuel-core-storage = { workspace = true }
 fuel-core-types = { workspace = true }
+fuel-vm-private = { workspace = true }
+futures = "0.3.28"
 parking_lot = { workspace = true }
 tokio = { workspace = true, default-features = false, features = ["sync"] }
+tokio-rayon = "2.1.0"
```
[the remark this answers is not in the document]
leviathanbeak: tokio-rayon was not included into workspace deps but I included it now

crates/services/txpool/src/service.rs
```diff
@@ -138,7 +147,7 @@ pub struct Task<P2P, DB> {
 impl<P2P, DB> RunnableService for Task<P2P, DB>
 where
     P2P: PeerToPeer<GossipedTransaction = TransactionGossipData> + Send + Sync,
-    DB: TxPoolDb,
+    DB: TxPoolDb + Clone,
```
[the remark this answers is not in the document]
leviathanbeak: not possible to remove due to `clone()` usage in `self.shared.clone()`

crates/services/txpool/src/txpool.rs
leviathanbeak: I returned the appropriate error type instead `Error::NotSupportedTransactionType`

crates/services/txpool/src/txpool.rs
leviathanbeak: what to do with `id` field in the macro then since `CheckedTransaction` does not have it (directly)

crates/services/txpool/src/txpool.rs
leviathanbeak: I turned it into test-only method since it was used in the tests in other files

crates/services/txpool/Cargo.toml
```diff
@@ -18,8 +18,11 @@ fuel-core-metrics = { workspace = true }
 fuel-core-services = { workspace = true }
 fuel-core-storage = { workspace = true }
 fuel-core-types = { workspace = true }
+fuel-vm-private = { workspace = true }
```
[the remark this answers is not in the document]
leviathanbeak: I think the reason for this is because `PredicateVerificationFailed` isn't exported from fuel_core_types::fuel_vm from what I saw

crates/services/txpool/src/txpool.rs
leviathanbeak: added it

crates/services/txpool/src/txpool.rs
leviathanbeak: good point!

crates/services/txpool/src/txpool.rs
xgreenx: You can try to use `Checked<Transaction>` instead of `CheckedTransaction`. Then you can get transaction and get the id=)

leviathanbeak: switched to `Checked<Transaction>`

crates/services/txpool/Cargo.toml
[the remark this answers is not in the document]
xgreenx: But you can export it=D

crates/services/txpool/src/service.rs
```diff
@@ -205,29 +215,38 @@ where
...
-                            Some(GossipsubMessageAcceptance::Accept)
-                        },
-                        Some(Err(_)) => {
-                            Some(GossipsubMessageAcceptance::Reject)
+                    let current_height = self.shared.db.current_block_height()?;
+
+                    // verify tx
+                    let checked_tx = check_single_tx(tx, current_height, &self.shared.config).await;
+
+                    if let Ok(tx) = checked_tx {
```
xgreenx, line 223: I guess in the case that it is not valid, we need to report `GossipsubMessageAcceptance::Reject`

crates/services/txpool/src/service.rs
```diff
@@ -313,13 +332,34 @@ where
...
+        let valid_txs = checked_txs
+            .iter()
+            .filter_map(|tx_chech| match tx_chech {
+                Ok(tx) => Some(tx.clone()),
+                Err(e) => {
+                    tracing::error!("Transaction is invalid: {}", e);
+                    None
+                }
+            })
+            .collect();
```
xgreenx, line 357: It would be nice if you could do that without cloning of the transaction. For example, you can wrap all values into `Some(result)`. After, you can extract all valid transactions and replace them with `None`, insert them, and return back. It is one available option, but maybe you can find another one=D

crates/services/txpool/src/txpool.rs
```diff
@@ -71,40 +86,15 @@ where
         &self.by_dependency
     }
 
-    #[tracing::instrument(level = "info", skip_all, fields(tx_id = %tx.id(&self.config.chain_config.transaction_parameters.chain_id)), ret, err)]
+    #[tracing::instrument(level = "info", skip_all, fields(tx_id = %tx.id()), ret, err)]
     // this is atomic operation. Return removed(pushed out/replaced) transactions
     fn insert_inner(
         &mut self,
         // TODO: Pass `&Transaction`
```
xgreenx, line 93: I guess you can remove this todo now

crates/services/txpool/src/txpool.rs
```diff
@@ -391,6 +365,96 @@ where
...
+        )
+        .await?
+    } else {
+        tx.into_checked_basic(
+            current_height,
+            &config.chain_config.transaction_parameters,
+        )?
+    };
+
+    debug_assert!(tx.checks().contains(Checks::All));
```
xgreenx, line 413: It should be in the `config.utxo_validation` section

xgreenx (approved):

Issue #1243, opened 2023-07-10
Decompose `ConsensusParameters` struct into more relevant structs


MitchTurner:
Let's break down `ConsensusParameters` into structs based on how they are used, rather than stuffing every value into one struct.

For example the fields that are used for calculating Tx cost could be included in a `TxParams` struct and the fields that include `max` values, could be in a `DataLimit` struct.

It is also possible that fields that are not included in `ConsensusParameters` but are used with it would be better off in the `TxParams`. Look at fields on `ChainInfo`, for example.

Related to: https://github.com/FuelLabs/fuel-core/issues/1221

xgreenx:
I want to clarify that the idea is to keep the `ConsensusParameters` structure and put there all data required for consensus(Like `GasCosts`, storage costs, `TxParametrs`, etc). But to avoid the mess with million fields, we need to extract fields into their own structures based on usage context.

xgreenx:
Finished with https://github.com/FuelLabs/fuel-core/pull/1262

PR #1262, opened 2023-07-26
Decompose consensus params struct

MitchTurner:
https://github.com/FuelLabs/fuel-core/issues/1243

Adding the changes from https://github.com/FuelLabs/fuel-vm/pull/514 to `fuel-core`


MitchTurner:
There is a failing benchmark test that I haven't figured out yet. It seems like it can't find some structs exposed by `fuel_core_types`, but I can't figure out what's different about the benchmark crate from the other crates that access those types :(

xgreenx:
It is because of my change in the `fuel-asm`...=( I duplicated the alias in the `fuel-asm` to remove dependency on the `fuel-types`. And now you face a problem that in the bench, you import `*` - everything, and they overlap in the naming of the `Word` alias.

image

MitchTurner:
Yeah. This is actually another test that I had fail locally. I need to solve both. I'll get them in the morning.

bin/e2e-test-client/src/test_context.rs
```diff
@@ -176,7 +176,11 @@ impl Wallet {
             amount: 0,
             asset_id,
         });
-        tx.with_params(self.consensus_params);
+        tx.with_tx_params(self.consensus_params.tx_params);
```
xgreenx, line 179: We need to support continue supporting `tx.with_params(self.consensus_params);`=) Otherwise it is a lot of boilerplate code

crates/chain-config/src/snapshots/fuel_core_chain_config__config__tests__snapshot_local_testnet_config.snap
```diff
@@ -35,21 +36,171 @@ expression: json
     ]
   },
   "transaction_parameters": {
```
xgreenx, line 38: Could you also rename this field into `consensus_parameters`, please?=)

crates/chain-config/src/snapshots/fuel_core_chain_config__config__tests__snapshot_simple_message_state.snap
```diff
@@ -18,21 +19,171 @@ expression: json
...
+        "base": 207,
+        "dep_per_unit": 18
+      },
+      "srwq": {
+        "base": 44,
+        "dep_per_unit": 5
+      }
+    }
   },
   "gas_costs": {
```
xgreenx, line 188: Now you can remove `gas_costs` field here=)

crates/client/src/client/schema/chain.rs
```diff
@@ -4,28 +4,127 @@ use crate::client::schema::{
...
+impl From<FeeParameters> for fuel_core_types::fuel_tx::FeeParameters {
+    fn from(params: FeeParameters) -> Self {
+        Self {
+            gas_price_factor: params.gas_price_factor.0,
+            gas_per_byte: params.gas_per_byte.0,
+        }
+    }
+}
+
+// macro_rules! include_from_impls {
```
xgreenx, line 108: It should be simpler for you to use macro than copy paste it=)

crates/client/src/client/schema/snapshots/fuel_core_client__client__schema__chain__tests__chain_gql_query_output.snap
```diff
@@ -38,21 +39,171 @@ query {
...
+          base
+          depPerUnit
+        }
+        srwq {
+          base
+          depPerUnit
+        }
+      }
     }
     gasCosts {
```
xgreenx, line 208: You can remove it too now

crates/client/src/client/schema/chain.rs
MitchTurner:
```
error[E0425]: cannot find value `builder` in this scope
   --> crates/client/src/client/schema/chain.rs:114:39
    |
114 |               $($field_vis $field_name: $field_type,)*
    |                                         ^^^^^^^^^^^ not found in this scope
...
128 | / include_from_impls! {
129 | | #[derive(cynic::QueryFragment, Debug)]
130 | | #[cynic(schema_path = "./assets/schema.sdl")]
131 | | pub struct GasCosts {
```

I'm getting this error when I try to use the macro. My best guess is that it doesn't like the attributes:
```
#[derive(cynic::QueryFragment, Debug)]
```

MitchTurner:
I'm currently overriding crates.io to point at the latest version of `fuel-vm` on GitHub. 
```
# Fuel dependencies
fuel-vm-private = { version = "0.35.1", package = "fuel-vm", git = "https://github.com/FuelLabs/fuel-vm.git" }
```
I don't know if that's kosher. 

Would it be better to publish a new version of `fuel-vm` before merging this PR?

Dentosal:
> I'm currently overriding crates.io to point at the latest version of fuel-vm on GitHub.
> Would it be better to publish a new version of fuel-vm before merging this PR?

> [crates.io](https://crates.io/) does not allow packages to be published with git dependencies [[docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories)]

MitchTurner:
Yes. Sorry I'm not suggesting that. I'm just asking if we need to cut a new release of fuel-vm before merging this PR.

xgreenx:
We don't merge into `master` code with `patch`=) Usually, after approving the PR, we will release all required dependencies, and you can use them to FINISH HIM(pull request)=D

crates/client/src/client/schema/chain.rs
MitchTurner: I modified the macro to include the cynic stuff so this isn't an issue.

Not the cleanest IMO, but it was the way to get the macro to work.

crates/client/src/client/schema/chain.rs
```diff
@@ -4,134 +4,234 @@ use crate::client::schema::{
...
-    pub max_storage_slots: U64,
+}
+
+impl From<TxParameters> for fuel_core_types::fuel_tx::TxParameters {
+    fn from(params: TxParameters) -> Self {
+        Self {
+            max_inputs: params.max_inputs.0,
+            max_outputs: params.max_outputs.0,
+            max_witnesses: params.max_witnesses.0,
+            max_gas_per_tx: params.max_gas_per_tx.0,
```
xgreenx, line 35: Could you use `into()` please, instead of `0` in all places?=)
```suggestion
            max_inputs: params.max_inputs.into(),
            max_outputs: params.max_outputs.into(),
            max_witnesses: params.max_witnesses.into(),
            max_gas_per_tx: params.max_gas_per_tx.into(),
```

xgreenx (commented):
Thank you for a huge work, looks much much better!=)
Let's wait for a new `fuel-vm` release

xgreenx (dismissed):

xgreenx (approved):
Thank you for finishing this issue, looks very good!:)

PR #1267, opened 2023-07-28
add metrics to services

### LINKED ISSUE #1168: Performance metrics for each service
Measure time of run function for every service

leviathanbeak:
closes #1168 

I added metrics to each `run()` method of `RunnableTask` implementor.

Improvement: this could be hidden behind `metrics` feature?


crates/services/sync/src/sync.rs
```diff
@@ -53,6 +60,8 @@ impl SyncHeights {
     /// Sync the state from the height stream.
     /// This stream never blocks or errors.
     pub(crate) async fn sync(&mut self) -> Option<()> {
+        let start = ServiceMetrics::instant();
```
xgreenx, line 63: It measures busy and idle time together, we want to know about both independently.

crates/services/sync/src/import.rs
```diff
@@ -118,8 +123,11 @@ where
         &self,
         shutdown: &mut StateWatcher,
     ) -> anyhow::Result<bool> {
+        let start = ServiceMetrics::instant();
+
         self.import_inner(shutdown).await?;
 
+        IMPORT_METRICS.observe(start);
```
xgreenx, line 130: Each service may have the instance inside of `ServiceRunner` and you don't need to use lazy=)

xgreenx:
Clsoe in favor of https://github.com/FuelLabs/fuel-core/pull/1284

Issue #1236, opened 2023-07-03
Add instruction of how to debug the `fuel-core`


xgreenx:
- Add instructions with screenshots of how to debug the project. 
   - In the Clion
   - In the VSCode
- How to build with/without p2p and relayer
- How to connect the node to the existing network

PR #1268, opened 2023-07-29
feat: Add `enable-p2p` flag to `fuel-core` CLI

bvrooman:
Related issues:
- https://github.com/FuelLabs/fuel-core/issues/1236

This PR disables P2P by default, and adds a new flag to the `fuel-core` CLI to allow users to expressly enable P2P:
```
--enable-p2p
```
Currently, the P2P service is only enabled when the user provides CLI arguments for `keypair` and `network`. If either of these values are missing, P2P is implicitly disabled without warning to the user. There is no way to tell if the user accidentally omitted these values, or deliberately omitted them in order to disable P2P. 

A better user experience is to enable P2P explicitly using `--enable-p2p` and to make `keypair` and `network` mandatory. This will raise an error if either of these values is missing. P2P should be disabled by default. This flag replaces having to add or remove the `network` argument to enable or disable P2P implicitly. 

If a user wants to run `fuel-core` with P2P, and supplies `--enable-p2p`, but forgets to include `keypair` and/or `network`, they will encounter an error:
```
error: the following required arguments were not provided:
  --keypair <KEYPAIR>
  --network <NETWORK>
  ```

### DIFF
--- a/bin/fuel-core/src/cli/run/p2p.rs
+++ b/bin/fuel-core/src/cli/run/p2p.rs
@@ -32,19 +32,19 @@ const MAX_RESPONSE_SIZE_STR: &str = const_format::formatcp!("{MAX_RESPONSE_SIZE}
 
 #[derive(Debug, Clone, Args)]
 pub struct P2PArgs {
-    /// Disable P2P. By default, P2P is enabled when the binary is compiled with the "p2p" feature
-    /// flag. Providing `--disable-p2p` will disable the P2P service.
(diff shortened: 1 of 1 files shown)


bvrooman:
As discussed in the sync today, we will change the approach to have P2P disabled by default, and instead use a flag to enable it. I will rework this PR with that feedback. In the meantime, I am putting it back in draft.

xgreenx (approved):

Issue #1133, opened 2023-04-17
Download range of headers during syncing


Voxelot:
Instead of fetching headers 1 by 1, add a new p2p api that allows entire ranges of blocks to be downloaded from a peer.

xgreenx:
Solved by https://github.com/FuelLabs/fuel-core/commit/078446b2712ea326d99846800a3e1527672a0a60

PR #1270, opened 2023-08-01
Download range of blocks during syncing

MitchTurner:
https://github.com/FuelLabs/fuel-core/issues/1133


MitchTurner:
I've added the blocks test next to the block test, but I'm concerned with the scope of the test. I feel like we'd be better off with tests that:
- injected the _correct_ event
- included a dummy DB, rather than some dummy response
- checked that expected values are returned based on the request

The current tests just show that the messages can be transmitted to peers. Which is important, but not complete. Maybe I'm missing tests that show coverage of lookup, etc. I'll keep looking,

crates/services/sync/src/ports.rs
```diff
@@ -31,6 +31,14 @@ pub trait PeerToPeerPort {
         height: BlockHeight,
     ) -> anyhow::Result<Option<SourcePeer<SealedBlockHeader>>>;
 
+    /// Request range of sealed block headers from the network
+    /// starting from block height `start` up to, and including `end`.
+    async fn get_sealed_block_headers_inclusive(
+        &self,
+        start: BlockHeight,
+        end: BlockHeight,
```
xgreenx, line 39: Maybe it is better to work with the `Range` directly

crates/services/sync/src/import.rs
```diff
@@ -250,6 +230,35 @@ where
         .in_current_span()
         .await
     }
+
+    async fn get_block_for_header(
```
xgreenx, line 234: You have some intersection with @bvrooman here, so maybe it is better not to update `fuel-core-sync` for now and only add new functionality on the p2p layer to avoid conflicts.

crates/services/p2p/src/request_response/messages.rs
```diff
@@ -49,6 +50,7 @@ pub enum RequestMessage {
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub enum ResponseMessage {
     SealedBlock(Option<SealedBlock>),
+    SealedBlocksInclusive(Vec<SealedBlock>),
```
xgreenx, line 53:
```suggestion
    SealedBlocksInclusive(Vec<Option<SealedBlock>>),
```

Maybe it is better to follow the pattern=)

crates/services/p2p/src/request_response/messages.rs
```diff
@@ -66,6 +69,7 @@ pub enum ResponseChannelItem {
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub enum NetworkResponse {
     Block(Option<Vec<u8>>),
+    BlocksInclusive(Vec<u8>),
```
xgreenx, line 72:
```suggestion
    BlocksInclusive(Option<Vec<u8>>),
```

crates/services/p2p/src/request_response/messages.rs
```diff
@@ -75,6 +79,7 @@ pub enum NetworkResponse {
 #[derive(Debug, Clone)]
 pub enum OutboundResponse {
     Block(Option<Arc<SealedBlock>>),
+    BlocksInclusive(Vec<SealedBlock>),
```
xgreenx, line 82:
```suggestion
    BlocksInclusive(Option<Arc<Vec<SealedBlock>>),
```

crates/services/p2p/src/request_response/messages.rs
```diff
@@ -75,6 +79,7 @@ pub enum NetworkResponse {
...
     SealedHeader(Option<Arc<SealedBlockHeader>>),
     Transactions(Option<Arc<Vec<Transaction>>>),
```
xgreenx, line 84: Sorry for the wrong description of the task. The initial idea was to support range requests for headers and transactions.

crates/services/sync/src/import.rs
MitchTurner: Hmm. Okay. Most of the work I've been doing today has been in understanding the sync stuff. There actually is very little p2p stuff to do, turns out.
I've also been rewriting some of the tests and over in sync. But sounds like I should cut that out.
I'll look into @bvrooman 's work before moving forward.

bvrooman: It looks like there is overlap in the `get_header_range` area.

crates/services/sync/src/ports.rs
MitchTurner: Range of `u32`?

MitchTurner: I prefer using domain types. I think `Range<BlockHeight>` might be a thing, but it doesn't implement `Iterator` for example, so it's not great to work with.

crates/services/p2p/src/p2p_service.rs
```diff
@@ -1514,6 +1527,26 @@ mod tests {
         }
     }
 
+    fn arbitrary_headers() -> Vec<SealedBlockHeader> {
+        let mut blocks = Vec::new();
+        for _ in 2..=5 {
+            let header = Default::default();
+
+            let sealed_block = SealedBlockHeader {
```
xgreenx, line 1535: All produced blocks are the same, so maybe you want to use `height` to make some difference=)

crates/services/p2p/src/ports.rs
```diff
@@ -21,6 +21,12 @@ pub trait P2pDb: Send + Sync {
         height: &BlockHeight,
     ) -> StorageResult<Option<SealedBlockHeader>>;
 
+    fn get_sealed_headers_range_inclusive(
+        &self,
+        start_height: &BlockHeight,
+        end_height: &BlockHeight,
+    ) -> StorageResult<Vec<SealedBlockHeader>>;
+
```
xgreenx, line 29: You don't need to have a `_range_inclusive` suffix. Instead, you can work directly with the `Range<BlockHeight>` or `Range<u32>` type

crates/services/p2p/src/service.rs
```diff
@@ -568,6 +606,14 @@ pub mod tests {
             }))
         }
 
+        fn get_sealed_headers_range_inclusive(
```
xgreenx, line 609: What is the difference with `get_sealed_block_headers_inclusive`?

crates/services/sync/src/import.rs
```diff
@@ -163,41 +161,22 @@ where
...
         } = &self;
-        // Request up to `max_get_header_requests` headers from the network.
-        get_header_range_buffered(range.clone(), params, p2p.clone())
+
+        let headers_res = get_header_range_single_req(range.clone(), p2p.clone()).await;
+        let unchecked_headers = match headers_res {
+            Ok(headers) => headers,
+            Err(e) => return (0, Err(e)),
+        };
+        let headers = check_headers_match_range(&unchecked_headers, range.clone());
```
xgreenx, line 170: I think it is better to move it inside of the `map` section below

crates/services/sync/src/import.rs
```diff
@@ -250,6 +229,54 @@ where
...
+            .await_da_height(&header.entity.da_height)
+            .await?;
+
+        get_transactions_on_block(p2p.as_ref(), block_id, header).await
+    }
+}
+
+// The response from the P2P network isn't guaranteed to include all the
+fn check_headers_match_range(
+    headers: &[SourcePeer<SealedBlockHeader>],
```
xgreenx, line 265: It is better to accept `Vec<_>` to avoid `clone`.

crates/services/sync/src/import.rs
```diff
@@ -250,6 +229,54 @@ where
...
+    range: RangeInclusive<u32>,
+) -> Vec<SourcePeer<SealedBlockHeader>> {
+    let mut new_headers = Vec::new();
+    let mut header_iter = headers.iter();
+    for height in range {
+        if let Some(header) =
```
xgreenx, line 271: It is a bad idea to use `find` because it is `O(n^2)` in the worst case.

You know that each block should have `i == header.height`, so you can iterate over the range and compare the height. You finish the iteration if you don't have more elements or one has an unexpected height. All valid block headers you can use for future processing. For all remaining blocks, you need to send a request again, and report about invalid behavior of the peer(we don't do that right now, but you can leave a TODO).


The right ordering(from smallest to biggest) of the block header should be a rule that each node should follow.

crates/services/sync/src/import.rs
```diff
@@ -450,3 +432,13 @@ impl<S> ScanErr<S> {
         })
     }
 }
+
+// #[cfg(test)]
```
xgreenx, line 436: Do we need it?=)

crates/services/sync/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 #![deny(unused_crate_dependencies)]
 #![deny(missing_docs)]
-#![deny(warnings)]
+// #![deny(warnings)]
```
xgreenx, line 3: Uncomment

crates/services/sync/src/ports.rs
xgreenx: You can work with `Range<u32>` for now. Leave todo to replace it with `BlockHeight` and implement `Iterator` in the `fuel-type` crate in parallel or follow-up=)

crates/services/sync/src/import.rs
MitchTurner: I agree on ordering. There is no danger to them returning the wrong order, it just will error on the check.

crates/services/sync/src/import.rs
MitchTurner: Is there a reason you'd prefer this? I just gave this a try and it got kinda messy trying to get the future types to match up.

crates/services/p2p/src/request_response/messages.rs
MitchTurner: Can we close these? Do you still prefer `Option` here? I like `Vec` because we have to check all the values anyway.

MitchTurner:
So, is the idea to implement `Step` for `BlockHeight` so we can use `Range<BlockHeight>` as `Iterator`?

crates/services/p2p/src/request_response/messages.rs
MitchTurner: This PR is still just doing headers. It might make sense to do transactions in a separate PR at this point?

xgreenx: Of course, if it makes development easier=)

crates/services/sync/src/import.rs
```diff
@@ -250,6 +230,66 @@ where
...
+    height: u32,
+) -> Option<SourcePeer<SealedBlockHeader>> {
+    if *(header.data.entity.height()) == height.into() {
+        Some(header)
+    } else {
+        None
+    }
+}
+
+// The response from the P2P network isn't guaranteed to include all the
```
Voxelot, line 275: all the ?

crates/services/sync/src/import.rs
```diff
@@ -163,41 +161,23 @@ where
             consensus,
             ..
         } = &self;
-        // Request up to `max_get_header_requests` headers from the network.
-        get_header_range_buffered(range.clone(), params, p2p.clone())
+
+        let headers_res = get_header_range_single_req(range.clone(), p2p.clone()).await;
```
Voxelot, line 165: We should test what happens when a node tries to fetch a very large range. My guess is that this will exceed the max payload size limit and be rejected, or it could consume a ton of memory if we store 1 million block headers in a vector here. We'll likely need to chunk this range up into separate requests.

MitchTurner: I'm working on the chunk code rn actually.

Voxelot: Maybe we should do a similar strategy to before where we started a stream, but instead of each item being a single block height it would be a sub-range of headers?

MitchTurner: Hmm. We could do that. What' I've done already is just turn it into a stream of headers like before--the caller can't tell that it's being batched as multiple calls under the hood

It seems to be working. I just had one test that was broken, so I hadn't pushed it up yet.

crates/services/sync/src/import.rs
MitchTurner: Got rid of this function anyway.

crates/services/sync/src/import.rs
MitchTurner: Need to add the batch sizes to config is all now.

crates/services/p2p/src/ports.rs
```diff
@@ -21,6 +22,11 @@ pub trait P2pDb: Send + Sync {
         height: &BlockHeight,
     ) -> StorageResult<Option<SealedBlockHeader>>;
 
+    fn get_sealed_headers_range(
```
bvrooman, line 25: Can we consider simply calling this `get_sealed_headers`? I think `_range` is not idiomatic in the context of the rest of our interfaces, where we may already accept a range without putting that in the function name.

crates/services/p2p/src/request_response/messages.rs
```diff
@@ -76,6 +83,7 @@ pub enum NetworkResponse {
 pub enum OutboundResponse {
     Block(Option<Arc<SealedBlock>>),
     SealedHeader(Option<Arc<SealedBlockHeader>>),
+    SealedHeadersRangeInclusive(Vec<SealedBlockHeader>),
```
bvrooman, line 86: Just curious:
- Why doesn't this one follow the same pattern of `Option<Arc<..>>` like the others?
- Is there a particular reason you went with the name `SealedHeadersRangeInclusive` instead of just `SealedHeaders`, similar to the other variants?

crates/services/sync/src/import.rs
```diff
@@ -241,15 +214,110 @@ where
...
+        })
+        .buffered(requests_in_flight)
+        .into_scan_err()
+        .scan_err()
+        .map(|val| {
+            let sorted_headers = match val {
+                Ok((mut range, headers)) => headers
+                    .into_iter()
+                    .map(move |header| {
+                        let height = range.next().unwrap();
```
bvrooman, line 292: If the `.unwrap()` is guaranteed to be safe, can we use `expect(..)` with a message explaining this instead?

crates/services/p2p/src/request_response/messages.rs
MitchTurner: > Why doesn't this one follow the same pattern of Option<Arc<..>> like the others?
I consider an empty `Vec` to be equivalent to a `None`. IF we were considering doing `Vec<Option<...>>` I'd still have to do a check on the ordering of the response, so it also didn't seem to give any meaning at this level. I discussed this a bit with @xgreenx before and I'm don't think he objected to my reasoning. Still open to changing if I'm wrong.

WRT the `Arc` I left it out expecting to get some `Sync`/`Send` error from the compiler if I needed it. Never got an error. Probably superfluous in the other cases.

> Is there a particular reason you went with the name SealedHeadersRangeInclusive instead of just SealedHeaders, similar to the other variants?
That's just left behind from when I wasn't using `range` and I've changed the name everywhere else. I'll fix that. Thanks!

crates/services/sync/src/import.rs
```diff
@@ -241,15 +214,110 @@ where
...
+}
+
+fn range_chunks(range: RangeInclusive<u32>, chunk_size: u32) -> Vec<RangeInclusive<u32>> {
+    let mut chunks = Vec::new();
+    let mut start = *range.start();
+    let end = *range.end();
+    while start <= end {
+        let chunk_start = start;
+        let chunk_end = start + chunk_size;
+        let chunk_end = if chunk_end > end { end } else { chunk_end };
```
bvrooman, line 316: I think this could simply be:
```
let chunk_end = std::min::cmp(start + chunk_size, end);
```

crates/services/p2p/src/ports.rs
MitchTurner: Yeah. Totally. I'll fix it.

crates/fuel-core/src/database/block.rs
```diff
@@ -271,9 +270,8 @@ impl Database {
             .get(commit_block_height)?
             .ok_or(not_found!(FuelBlockMerkleMetadata))?;
 
-        let storage = self.borrow();
         let tree: MerkleTree<FuelBlockMerkleData, _> =
-            MerkleTree::load(storage, commit_merkle_metadata.version)
+            MerkleTree::load(self, commit_merkle_metadata.version)
```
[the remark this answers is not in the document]
MitchTurner: I don't remember the context I was in when I changed this. I think I was just reading and felt like this was some misdirection in the code and removed the `borrow` to see what would happen.

I can't argue with your point though. I do like documentation with code. So I can change it back.

crates/services/sync/src/import.rs
MitchTurner: Oh. Yikes. Good catch. I think I'll just `.and_then` the following logic. That's good anyway because if we have a response that is longer than our `range` all the extra headers should be ignored`.

crates/services/sync/src/import.rs
```diff
@@ -241,15 +214,110 @@ where
...
+        chunks.push(chunk_start..=chunk_end);
+        start = chunk_end + 1;
+    }
+    chunks
 }
```
xgreenx, line 321:
```suggestion
fn range_chunks(
    range: RangeInclusive<u32>,
    chunk_size: u32,
) -> impl Iterator<Item = RangeInclusive<u32>> {
    let end = *range.end();
    range.step_by(chunk_size as usize).map(move |block_start| {
        let block_end = (block_start + chunk_size).min(end);
        block_start..=block_end
    })
}
```

crates/services/sync/src/import.rs
```diff
@@ -270,59 +338,13 @@ async fn wait_for_notify_or_shutdown(
...
-/// Returns a stream of network requests for headers.
-fn get_header_range(
-    range: RangeInclusive<u32>,
+async fn get_header_range_single_req(
+    range: &RangeInclusive<u32>,
     p2p: Arc<impl PeerToPeerPort + 'static>,
-) -> impl Stream<
-    Item = impl Future<Output = anyhow::Result<Option<SourcePeer<SealedBlockHeader>>>>,
-> {
-    stream::iter(range).map(move |height| {
```
xgreenx, line 295: You can use the old naming instead of creating new functions with slightly different names =) It will make fewer conflicts for @bvrooman and still works for you. Instead of `get_header_range`, you can use `get_header_range_buffered`, and instead of `get_header_range_single_req`, you can use `get_header_range`.


Also, it seems you don't need to touch other code except `get_header_range`. In this case, your change will be super tiny

crates/services/sync/src/import.rs
```diff
@@ -270,59 +338,13 @@ async fn wait_for_notify_or_shutdown(
...
-fn get_header_range_buffered(
-    range: RangeInclusive<u32>,
-    params: &Config,
-    p2p: Arc<impl PeerToPeerPort + Send + Sync + 'static>,
-) -> impl Stream<Item = anyhow::Result<SourcePeer<SealedBlockHeader>>> {
-    get_header_range(range, p2p)
-        .buffered(params.max_get_header_requests)
-        // Continue the stream unless an error or none occurs.
-        .into_scan_none_or_err()
-        .scan_none_or_err()
```
xgreenx, line 284: It would be nice to keep all these nice things from the `Stream`

crates/services/p2p/src/service.rs
```diff
@@ -568,6 +613,13 @@ pub mod tests {
             }))
         }
 
+        fn get_sealed_headers_range(
+            &self,
+            _block_height_range: Range<u32>,
+        ) -> StorageResult<Vec<SealedBlockHeader>> {
+            todo!()
```
xgreenx, line 620: Do we want to fix that during this PR?=)

MitchTurner: This is parity with what we had before. I'd be happy to do the work to get tests working for all the requests but should probably be done in a separate Issue if we want to do it correctly.

I actually wrote down some notes about this. I think this is the correct interface to test, instead of those other  tests where we manually assemble a response. I call this out at the beginning of this PR ^^^.. I'm pretty sure it would take some modifications to `Task` to make the `p2p_service` mockable (at least that's what I have in my notes, we'd have to double-check).

bin/fuel-core/src/cli/run/p2p.rs
```diff
@@ -170,12 +170,15 @@ pub struct P2PArgs {
 
 #[derive(Debug, Clone, Args)]
 pub struct SyncArgs {
```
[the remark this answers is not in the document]
MitchTurner: I'm not sure I understand. Which piece of code is looking for these values?

crates/services/p2p/src/service.rs
xgreenx: Could you use `unreachable` instead, please?

Hmm, it seems we don't test any logic except start of the service. Could you update all methods in the implementation of the `P2pDb` trait to do `unreachable!()`?

bin/fuel-core/src/cli/run/p2p.rs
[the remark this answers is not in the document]
xgreenx: I meant that while receiving the `SealedHeaders` on the receiver side of the P2P, we need to limit the size of the batch. Because right now, anyone can request 1B headers without punishment.

crates/services/sync/src/import/tests.rs
```diff
@@ -8,454 +10,483 @@ use crate::ports::{
...
-            consensus_port: DefaultMocks::times([1]),
-            executor: DefaultMocks::times([0])
+#[tokio::test]
+async fn test_import(state: State, mocks: Mocks) -> (State, bool) {
+    let state = SharedMutex::new(state);
+    test_import_inner(state, mocks, None).await
+}
+
+#[tokio::test]
+async fn import__signature_fails_on_header_5_only() {
```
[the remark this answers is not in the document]
MitchTurner: It's a naming scheme I like to use. We used it at my last job.

The idea is that every test should only be testing a single function (in this case, `import`) So for all the tests for a function, you write `<function_name>__<what you're testing in snake case>`.

It makes reading easier to have a standardized naming scheme.

crates/services/p2p/src/service.rs
MitchTurner: `unreachable!()` wouldn't be correct. I can just call the methods directly. How about `unimplemented!()`?

crates/services/sync/src/import/back_pressure_tests.rs
```diff
@@ -139,41 +145,59 @@ struct Counts {
...
+
+    async fn get_sealed_block_headers(
+        &self,
+        block_height_range: Range<u32>,
+    ) -> anyhow::Result<Vec<SourcePeer<SealedBlockHeader>>> {
+        // TODO: Does this need more count tracking?
+        for _ in block_height_range.clone() {
+            // self.counts.apply(|c| c.inc_headers());
+            self.counts.apply(|c| c.inc_blocks());
+        }
```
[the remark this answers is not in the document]
MitchTurner: I don't understand what this is for. It seems strange to me that there is so much novel logic inside a test impl.

crates/services/sync/src/import/back_pressure_tests.rs
```diff
@@ -139,41 +145,59 @@ struct Counts {
...
+struct PressureConsensus(MockConsensusPort, Duration, SharedCounts);
 
 #[async_trait::async_trait]
-impl PeerToPeerPort for PressurePeerToPeerPort {
+impl PeerToPeerPort for PressurePeerToPeer {
     fn height_stream(&self) -> BoxStream<BlockHeight> {
-        self.0.height_stream()
+        self.p2p.height_stream()
     }
     async fn get_sealed_block_header(
```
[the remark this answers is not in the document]
MitchTurner: There's also the `P2pDb` method `get_sealed_header` should we get rid of all of that and the handling? Along with `RequestMessage::SealedHeader` and all that?

crates/services/sync/src/import/back_pressure_tests.rs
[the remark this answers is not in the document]
MitchTurner: I made those changes, but it was breaking one of the tests. It was `State::new(None, 1)` which should be len 2 (for header 0 and header 1), but it was expecting block count to be 1. So, I changed it to `State::new(None, 0)`.

Not positive that's correct, but the test is passing now.

bin/fuel-core/src/cli/run/p2p.rs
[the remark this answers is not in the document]
MitchTurner: I think I understand. What is the correct behavior of the sender? Should they truncate the request or should they error?

I think I'll just truncate for now. Seems fine.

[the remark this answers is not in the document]
xgreenx: Based on the current `fuel-core-sync` implementation, it is better to reject the request=D Because if we request 100 batches and the first batch fails, we will waste a lot of resources for nothing(because all batches, except the first, will be dropped).

crates/services/sync/src/import/back_pressure_tests.rs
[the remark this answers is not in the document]
xgreenx: Yep

crates/services/sync/src/import/back_pressure_tests.rs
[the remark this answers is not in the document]
xgreenx: > I don't understand what this is for. It seems strange to me that there is so much novel logic inside a test impl.

The idea is to create some delay for each request.

bin/fuel-core/src/cli/run/p2p.rs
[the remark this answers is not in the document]
MitchTurner: Ah. Kk. Fair enough.

Voxelot:
@MitchTurner looks like there's a merge conflict to resolve

crates/services/p2p/src/p2p_service.rs
```diff
@@ -1650,6 +1709,12 @@ mod tests {
         request_response_works_with(RequestMessage::Block(0.into())).await
     }
 
+    #[tokio::test]
+    #[instrument]
+    async fn request_response_works_with_sealed_headers_range_inclusive() {
+        request_response_works_with(RequestMessage::SealedHeaders(0..0)).await
```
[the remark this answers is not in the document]
xgreenx: The comment is not resolved=)

crates/services/sync/src/import/back_pressure_tests.rs
```diff
@@ -84,26 +92,38 @@ struct Input {
...
+},
+State::new(None, 50),
+Config{
+max_get_txns_requests: 10,
+header_batch_size: 10,
+max_header_batch_requests: 10,
+}
+=> is less_or_equal_than Count{ headers: 10, consensus: 10, transactions: 10, executes: 1, blocks: 21 }
+; "50 headers with max 10 size and max 10 requests"
+)]
```
xgreenx, line 115: Could you fix the indentation, please?

crates/services/sync/src/import/tests.rs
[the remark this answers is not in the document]
xgreenx: Why do we want to extract each test case into a separate function instead of the previous approach with `test_case`?

crates/services/sync/src/import.rs
Voxelot: we can make a todo to retry batches at some point, but if this becomes an issue we can just reduce the max number of buffered header batches to avoid too much throw-away work.

crates/services/sync/src/import/tests.rs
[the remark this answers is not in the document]
Voxelot: I don't have a strong preference either way, but I can see how overly parameterized test_case's can be opaque to others reading and maintaining the test suite. However in this case, it seems like we're still using the `test_import_inner` everywhere so this doesn't really change the readability of the tests to me.

deployment/charts/values.yaml
```diff
@@ -22,6 +22,8 @@ app:
   max_transmit_size: "${fuel_core_max_buffer_size}"
   sync_max_get_header: "${fuel_core_sync_max_get_header}"
   sync_max_get_txns: "${fuel_core_sync_max_get_txns}"
+  sync_header_batch_size: "${fuel_core_sync_header_batch_size}"
```
Voxelot, line 25: we can also set defaults for these vars in the `.env` which is used in automated deployments from master: https://github.com/FuelLabs/fuel-core/blob/master/deployment/scripts/.env

crates/services/p2p/src/service.rs
```diff
@@ -288,11 +298,14 @@ where
...
                             }
-                            RequestMessage::SealedHeader(block_height) => {
-                                let response = self.db.get_sealed_header(&block_height)?
-                                    .map(Arc::new);
-
-                                let _ = self.p2p_service.send_response_msg(request_id, OutboundResponse::SealedHeader(response));
+                            RequestMessage::SealedHeaders(range) => {
+                                let max_len = self.max_headers_per_request.try_into().expect("u32 should always fit into usize");
+                                if range.len() > max_len {
+                                    return Err(anyhow!("Requested too many headers: {} > {}", range.len(), max_len));
```
xgreenx, line 304: It logs the error on the receiver side. It would be much better to return this error to the sender of the request. Otherwise, the sender should wait for the timeout.

crates/services/sync/src/import.rs
```diff
@@ -58,17 +53,20 @@ mod back_pressure_tests;
 #[derive(Clone, Copy, Debug)]
 /// Parameters for the import task.
 pub struct Config {
-    /// The maximum number of get header requests to make in a single batch.
-    pub max_get_header_requests: usize,
     /// The maximum number of get transaction requests to make in a single batch.
     pub max_get_txns_requests: usize,
+    /// The maximum number of headers to request in a single batch.
+    pub header_batch_size: u32,
```
Dentosal, line 59: Why is this `u32` while the other amounts are `usize`? To me it seems like `get_headers_buffered` is doing unnecessary type conversions because of this.

xgreenx (approved):
Thank you, looks good!=) Could you fix Hanne's comment, please? After, we are ready to go=)

Dentosal (approved):

PR #1273, opened 2023-08-02
Block gas limit calibration

xgreenx:
- Added a new benchmark, `block_target_gas`, that runs different simple scripts with opcode and measures the block execution time. It allows us to track how accurate is our gas cost. If each benchmark takes ~ the same amount of time, then our gas cost is accurate. The current costs is not accurate because the `noop` takes ~1s for 100M gas, while `meq` - ~6s, `logd` - ~23s... Later we can have more opcodes tested and the end goal is to optimize opcodes/calibrate benchmarks to have the same execution time.
- Modified the formula to calculate the dependent cost. Instead of the predicted value, we use the value from the worst scenario. But we also need to consider that the cost of most opcodes rises logarithmically. So, there is a point after which each new element is relatively cheap. And if the cost includes cases with a small number of elements as a `base` cost, we can make the cost per element much cheaper. So, the initial benchmarks with `10`, `1000`, and `10000` elements can be included in the `base` cost to decrease the cost for each new element. The predicted value from linear regression is used to find *this* point.

### DIFF
--- a/benches/src/bin/collect.rs
+++ b/benches/src/bin/collect.rs
@@ -596,6 +596,25 @@ fn linear_regression(x_y: Vec<(u64, u64)>) -> f64 {
         .map(|(x, y)| (*x as f64 - avg_x) * (*y as f64 - avg_y))
         .sum();
     let sq_x: f64 = x_y.iter().map(|(x, _)| (*x as f64 - avg_x).powi(2)).sum();
+
+    // The original formula says:
+    //
+    // y = B * x,
+    // where B = sum((x_i - x_avg) * (y_i - y_avg)) / sum((x_i - x_avg) ^ 2) = sum_x_y / sq_x
+    //
+    // We want to know how many elements fit into one `noop` opcode timing,
+    // so we need value of `x / y`:
+    //
+    // y = B * x
+    //     |
+    //    \|/
+    // 1 / B = x / y
+    //     |
+    //    \|/
+    // 1 / sum_x_y / sq_x = x / y
+    //     |
+    //    \|/
+    // sq_x / sum_x_y = x / y
     sq_x / sum_x_y
 }
 
--- a/benches/src/bin/collect.rs
+++ b/benches/src/bin/collect.rs
@@ -808,6 +808,9 @@ mod tests {
 
     #[test]
     fn test_linear_regression() {
-        assert_eq!(linear_regression(vec![(4, 2), (8, 3), (12, 4), (16, 5)]), 4);
+        assert_eq!(
+            linear_regression(vec![(4, 2), (8, 3), (12, 4), (16, 5)]),
+            4.0
+        );
     }
 }
--- a/benches/src/bin/collect.rs
+++ b/benches/src/bin/collect.rs
@@ -530,7 +530,7 @@ impl State {
                         time: *y,
                     })
                     .collect();
-                let dep_per_unit = (1.0 / slope(x_y)) as u64;
+                let (_, dep_per_unit) = dependent_cost(x_y);
                 (
                     name,
                     Cost::DependentAll {
@@ -543,8 +543,8 @@ impl State {
         } else {
             let iter = dependant_groups.into_iter().map(|(name, x_y)| {
                 groups.remove(&name);
-                let base = x_y.first().unwrap().1;
-                let dep_per_unit = (1.0 / slope(x_y)) as u64;
+
+                let (base, dep_per_unit) = dependent_cost(x_y);
                 (name, Cost::Dependent { base, dep_per_unit })
             });
             costs.0.extend(iter);
@@ -586,7 +586,7 @@ impl Costs {
     }
 }
 
-fn slope(x_y: Vec<(u64, u64)>) -> f64 {
+fn linear_regression(x_y: Vec<(u64, u64)>) -> f64 {
     let avg_x =
         x_y.iter().map(|(x, _)| x).copied().sum::<u64>() as f64 / x_y.len() as f64;
     let avg_y =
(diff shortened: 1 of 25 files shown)


benches/src/bin/collect.rs
```diff
@@ -596,7 +596,108 @@ fn slope(x_y: Vec<(u64, u64)>) -> f64 {
         .map(|(x, y)| (*x as f64 - avg_x) * (*y as f64 - avg_y))
         .sum();
     let sq_x: f64 = x_y.iter().map(|(x, _)| (*x as f64 - avg_x).powi(2)).sum();
-    sum_x_y / sq_x
+    sq_x / sum_x_y
```
Voxelot, line 599: I don't think this is correct for the least squares method of linear regression, it should be `sum_x_y / sq_x`:
image
https://en.wikipedia.org/wiki/Simple_linear_regression#Fitting_the_regression_line

xgreenx: Yes, but we need value for `y`, not `x`. And it is the same behavior as before(after `slope` we divide `1/slope(x_y)`

Voxelot: Or is it inverted because our dep_per_unit is inverted? Might be worth putting a comment explaining why this deviates from the standard formula.

xgreenx: Added a comment=)

benches/src/bin/collect.rs
```diff
@@ -596,7 +596,127 @@ fn slope(x_y: Vec<(u64, u64)>) -> f64 {
...
+    // y = B * x,
+    // where B = sum((x_i - x_avg) * (y_i - y_avg)) / sum((x_i - x_avg) ^ 2) = sum_x_y / sq_x
+    //
+    // We want to know how many elements fit into one `noop` opcode timing,
+    // so we need value of `x / y`:
+    //
+    // y = B * x
+    //     |
+    //    \|/
+    // 1 / B = x / y
```
Voxelot, line 611: [nit] could add a transformation to both sides of the equation like (rhs)^-1 = (lhs)^-1

benches/src/bin/collect.rs
```diff
@@ -596,7 +596,127 @@ fn slope(x_y: Vec<(u64, u64)>) -> f64 {
...
+    let x_y = x_y
+        .into_iter()
+        .map(|(x, y)| Point { x, y })
+        .collect::<Vec<_>>();
+
+    let first = x_y.first().unwrap();
+    let last = x_y.last().unwrap();
+
+    let near_linear = linear_regression * NEAR_LINEAR;
+    let expected_type = if (linear_regression - first.amount()).abs() < near_linear
```
Voxelot, line 671: A more sound approach for testing the goodness-of-fit using linear regression would be the $R^2$ metric, or [coefficient of determination](https://en.wikipedia.org/wiki/Coefficient_of_determination#Definitions). Where 1 represents a perfect fit, and lower values indicate a poor fit of the linear model.

image

benches/src/bin/collect.rs
```diff
@@ -596,7 +596,127 @@ fn slope(x_y: Vec<(u64, u64)>) -> f64 {
...
+                .skip_while(|p| {
+                    if p.amount() < linear_regression - base.amount() {
+                        base = *p;
+                        true
+                    } else {
+                        false
+                    }
+                })
+                .map(|p| p.amount())
+                .min_by(|a, b| a.partial_cmp(b).unwrap())
```
Voxelot, line 712: Wouldn't this potentially leave a range of inputs somewhere between the base and the min(p.amount()) where we undercharge for execution time? would it be safer to just take the slope off the base, even if it overcharges?

benches/src/bin/collect.rs
```diff
@@ -596,7 +596,127 @@ fn slope(x_y: Vec<(u64, u64)>) -> f64 {
...
+
+            let mut base = x_y.next().unwrap();
+
+            let amount = x_y
+                .skip_while(|p| {
+                    if p.amount() < linear_regression - base.amount() {
+                        base = *p;
+                        true
+                    } else {
+                        false
```
Voxelot, line 708: Should we return some kind of error if we never find a base using the slope comparison? Otherwise it will just default to the first item which could be a very small base fee and be very inadequate for charging users. If we don't find a base it probably means we don't have enough points to accurately predict where the log fn seems "linear" enough.

For example if no base is found and the first point is x=10, and then we use the lowest amount from the high end of the points like x = 1,000,000 this could severely undercharge.

benches/src/bin/collect.rs
Voxelot: also nit - isn't this a more complicated expression of simply using `.min()` here?

benches/src/bin/collect.rs
xgreenx: It shouldn't be possible. Our function is linear if the initial value is around the predicted value. So the initial price is already fair enough in this case.

benches/src/bin/collect.rs
xgreenx: Floats don't support `min` and you need to use `partial_cmp` =(

Remember, the lower value of `amount` means we charge the user more because of `args / dep_per_unit`. So we will only overcharge user, not undercharge.

benches/src/bin/collect.rs
xgreenx: We could use it here, but we don't have cases of linear behavior, so the super-complex formula is overkill. Because each opcode includes the base execution time, it looks like a logarithmic almost in all cases.

Plus, we still need to play with the threshold for `R`. The code for the logarithm behavior should cover almost all cases, and the `min` strategy means we can only overcharge user. If later, with benchmarking target block time, we will see that some opcodes are really overcharged, then we can try to improve the algorithm. But right now, it looks like our benchmarks produce undercharge costs.

Voxelot (approved):

PR #1276, opened 2023-08-04
Release v0.20.3

xgreenx:
Finalizing the chain specification for the beta 4 network.
Bumping version to `0.20.3`.

### DIFF
--- a/deployment/scripts/chainspec/beta_chainspec.json
+++ b/deployment/scripts/chainspec/beta_chainspec.json
@@ -1,6 +1,6 @@
 {
   "chain_name": "Testnet Beta 4",
-  "block_gas_limit": 5000000000,
+  "block_gas_limit": 30000000,
   "initial_state": {
     "coins": [
       {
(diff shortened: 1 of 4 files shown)


deployment/scripts/chainspec/beta_chainspec.json
```diff
@@ -52,7 +52,7 @@
     "max_predicate_length": 1048576,
     "max_predicate_data_length": 1048576,
     "max_gas_per_predicate": 10000000,
-    "gas_price_factor": 1000000000,
+    "gas_price_factor": 92,
```
xgreenx, line 55: Because we use eth as a gas asset, we want to avoid having super-expensive transactions. Because of that, we need to divide the gas price and make it cheaper than Gwei. The most expensive transaction shouldn't be more than 0.205 dollars. So the formula to calculate `gas_price_factor` is:

`ETH_PRICE * MAX_GAS_PER_TX * 10^9 / 10^18 / gas_price_factor  = tx_price`

1900 * 10000000 * 10^9/10^18 / gas_price_factor = 0.205
1900 * 10000000 * 10^9/10^18 / 0.205 = gas_price_factor 
gas_price_factor = 92

Voxelot (approved):

Issue #1280, opened 2023-08-09
Relayer spamming ethereum rpc


Voxelot:
There are some conditions in the relayer that appear to be causing an apparent infinite loop of requests to the ethereum rpc.

As seen here in our test environment, the relayer went on the fritz and made a lot of requests to eth rpc without any delay between them:
![image](https://github.com/FuelLabs/fuel-core/assets/794823/e94bf43d-c9e4-4f10-b166-78805bf82ea4)

The culprit is likely these two sections:
![image](https://github.com/FuelLabs/fuel-core/assets/794823/fa293048-3080-49c3-b169-a4d479c967d0)
![image](https://github.com/FuelLabs/fuel-core/assets/794823/aa0f70d4-15e5-493d-81b7-9c95462d4eae)

While the second image indicates that the relayer should wait before running again, this delay only takes effect if synced is set to some value. If the connection to eth drops or is rate limited, and the relayer is unable to obtain any sync state from eth, then it appears to start busily polling the connection to eth.

PR #1281, opened 2023-08-09
Fixes the relayer hot loop in the case of error

xgreenx:
Fixes https://github.com/FuelLabs/fuel-core/issues/1280

### DIFF
--- a/crates/services/relayer/src/config.rs
+++ b/crates/services/relayer/src/config.rs
@@ -45,7 +45,7 @@ pub struct Config {
 
 #[allow(missing_docs)]
 impl Config {
-    pub const DEFAULT_LOG_PAGE_SIZE: u64 = 5;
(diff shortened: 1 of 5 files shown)


Voxelot (approved):

crates/services/relayer/src/config.rs
```diff
@@ -45,7 +45,7 @@ pub struct Config {
 
 #[allow(missing_docs)]
 impl Config {
-    pub const DEFAULT_LOG_PAGE_SIZE: u64 = 5;
+    pub const DEFAULT_LOG_PAGE_SIZE: u64 = 10_000;
```
Voxelot, line 48: This is a pretty large increase, couldn't this worsen our rate limiting issues?

xgreenx: It should decrease the number of requests=) I used `100_000` locally during synchronization with the network and it was only several requests
image

Issue #940, opened 2023-01-26
test_case's should be documented with human readable descriptions, ie.


Voxelot:
test_case's should be documented with human readable descriptions, ie.

`#[test_case(...; "successfully imports genesis block when database is empty")`

_Originally posted by @Voxelot in https://github.com/FuelLabs/fuel-core/pull/935#discussion_r1087348155_

xgreenx:
Fixed with https://github.com/FuelLabs/fuel-core/pull/1286

PR #1286, opened 2023-08-10
Add readable test names to test cases

MitchTurner:
https://github.com/FuelLabs/fuel-core/issues/940

This issue seemed specific to the block import tests, but while I was here I figured I could do it for all the tests.

LMK if I missed any, since my IDE search seemed to be ignoring some.


crates/services/consensus_module/poa/src/service_test/manually_produce_tests.rs
```diff
@@ -6,30 +6,39 @@ use test_case::test_case;
...
+#[test_case(Tai64::now() + 100, 10, vec![Tai64::now() + 100; 10], Trigger::Instant, 10;
+"can manually produce blocks with txs starting in the future when trigger is Instant")]
 #[test_case(
     Tai64::now() + 100, 3, vec![Tai64::now() + 100, Tai64::now() + 110, Tai64::now() + 120],
-    Trigger::Interval { block_time: Duration::from_secs(10) }, 10
-)]
+    Trigger::Interval { block_time: Duration::from_secs(10) }, 10;
+"can manually produce blocks with txs starting in the future with different times")]
 #[tokio::test]
 async fn can_manually_produce_block(
```
MitchTurner, line 43: I'm interested to learn why we have tests for manually producing blocks. Who is the customer for manually producing blocks?

xgreenx: It is helpful for developers to test cases where you need time travel. Like you deployed a contract/called some method at height X, and at height Y, you need to test some behavior =)

MitchTurner: Cool. That's what I suspected.
```
let production_enabled =
        !matches!(poa_config.trigger, Trigger::Never) || config.manual_blocks_enabled;
```
It seems like the manual production logic is exposed in production code. I don't think that should be the case. I like the idea of having it for tests :100: , but it should be hidden behind Block Production interface in some `TestBlockProducer` adapter.

crates/services/relayer/src/ports/tests.rs
```diff
@@ -69,13 +69,13 @@ fn insert_always_raises_da_height_monotonically() {
...
-#[test_case(11, None, 11)]
-#[test_case(11, None, 10)]
+#[test_case(None, 0, 0; "can set DA height to 0 when there is none available")]
+#[test_case(None, 10, 10; "can set DA height to 10 when there is none available")]
+#[test_case(0, 10, 10; "can set DA height to 10 when it is 0")]
+#[test_case(0, None, 0; "inserts are bypassed when height goes from 0 to 0")]
+#[test_case(10, 11, 11; "can set DA height to 11 when it is 10")]
+#[test_case(11, None, 11; "inserts are bypassed when height goes from 11 to 11")]
+#[test_case(11, None, 10; "inserts are bypassed when height reverted from 11 to 10")]
 fn set_raises_da_height_monotonically(
```
MitchTurner, line 79: These tests are confusing. Especially since there are no checks on output.

I could rewrite `set_finalized_da_height_to_at_least` to do nothing and it would still pass all of these tests:
```
fn set_finalized_da_height_to_at_least(
        &mut self,
        height: &DaBlockHeight,
    ) -> StorageResult<()> {
        let mut db_tx = self.transaction();
        let db = db_tx.as_mut();
        let _ = db.get(&METADATA_KEY);
        db.insert(&METADATA_KEY, 0)?;
        Ok(())
    }
```
:(

It would be nice if these tests checked storage after to see if the values are correct. It would be nice to have some failing tests! 

`mockall` is convenient, but it often puts the tests' focus on internal details. Tests should check how the interface behaves as a black box. 
`test_case` helps keep tests DRY, but it pushes tests toward reuse rather that coherence IMO. 

"Given, When, Then" would be useful here.

crates/services/relayer/src/ports/tests.rs
```diff
@@ -69,13 +69,13 @@ fn insert_always_raises_da_height_monotonically() {
...
-#[test_case(10, 11, 11)]
-#[test_case(11, None, 11)]
-#[test_case(11, None, 10)]
+#[test_case(None, 0, 0; "can set DA height to 0 when there is none available")]
+#[test_case(None, 10, 10; "can set DA height to 10 when there is none available")]
+#[test_case(0, 10, 10; "can set DA height to 10 when it is 0")]
+#[test_case(0, None, 0; "inserts are bypassed when height goes from 0 to 0")]
+#[test_case(10, 11, 11; "can set DA height to 11 when it is 10")]
+#[test_case(11, None, 11; "inserts are bypassed when height goes from 11 to 11")]
+#[test_case(11, None, 10; "inserts are bypassed when height reverted from 11 to 10")]
```
xgreenx, line 78: Hmm, that test case looks weird to me=) Maybe we need to return an error from `grow_monotonically` if the `new_height` is lower than the `current_height`.

My comment in the [initial PR](https://github.com/FuelLabs/fuel-core/pull/971#discussion_r1096189654) was left unrsponned=) Maybe you want to update it

MitchTurner: I agree. I didn't want to change the behavior of the code at all with this work. We should create another issues for fixing this IMO.

MitchTurner: Haha. I just saw the name of the tests lol: `set_raises_da_height_monotonically`. 

As I mentioned above, I think these tests need to all be rewritten anyway. There needs to be failure modes and checked output. These tests are kinda meaningless now. I'll put together an issue.

MitchTurner: https://github.com/FuelLabs/fuel-core/issues/1292

PR #1294, opened 2023-08-13
[script] fix ci_check.sh for everyone can use

YusongWang:


xgreenx:
Thank you for the PR, but the main reason for having the `ci_checks.sh` file is to simulate all checks running by CI locally. Our CI uses [`cargo-make`](https://github.com/sagiegurari/cargo-make) to run commands per each crate instead of the whole workspace because each crate should be valid individually.

It doesn't make sense to have a unique behavior in the `ci_checks.sh` without updating our CI. But we don't have plans to remove `cargo-make`, because it works well=)

Issue #1136, opened 2023-04-17
Load CLI arguments from configuration file


Voxelot:
There's a lot of complexity involved with adding new cli args to our deployment configurations. By allowing fuel-core to load these from a file, we wouldn't need to update our helm charts as often.

Voxelot:
closed by #1308

PR #1308, opened 2023-08-16
feat: Enable .env file loading using `env` feature

bvrooman:
Related issues:
- https://github.com/FuelLabs/fuel-core/issues/1236

This PR adds conditional support for loading a local `.env` for environment variables. This support is only enabled when compiling the client binary with the `env` feature or when passing `--all-features` to Cargo. This is not enabled by default for production environments.

Users can add a `.env` to the root directory of their local copy of their Fuel node (or wherever their working directory is). 

This helps users provide CLI arguments in a way that is more convenient and more secure than entering these values in an IDE directly:
- Users can reuse environment variables across IDEs (e.g., VS Code and CLion) and IDE configurations seamlessly, rather than specifying them in each IDE specific configuration
- Users can avoid putting secrets in shared commands or configurations
- CLion users can greatly shorten their `run` commands, which is helpful since the configuration window does not word wrap long commands

The `.env` is also added to `.gitignore` to prevent users from uploading individual environment configurations that may include secrets.

This approach acts as a sort of stop gap for handling secrets before we have SOPS.

This PR also renames the internal `eth_client` component of the relayer configuration to `relayer`. This allow users to use the environment variable name `RELAYER` to supply this value. Prior to this, the CLI expected an environment variable called `ETH_CLIENT`, even though the CLI argument is named `relayer`.   

Note: Due to the order in which the initialization functions are called, we cannot supply `RUST_LOG` in the .env file; this has to be provided elsewhere or directly to Cargo.


Voxelot:
neat! I didn't consider this option when I wrote up: #1136

Voxelot (dismissed):
approving - but also might want to wait for @xgreenx to look it over

bin/fuel-core/src/cli.rs
```diff
@@ -41,6 +47,16 @@ pub enum Fuel {
 pub const LOG_FILTER: &str = "RUST_LOG";
 pub const HUMAN_LOGGING: &str = "HUMAN_LOGGING";
 
+#[cfg(feature = "env")]
+fn init_environment() -> Option<PathBuf> {
+    dotenv().ok()
```
xgreenx, line 52: It would be nice to log the error here.

bin/fuel-core/src/cli.rs
```diff
@@ -87,6 +103,11 @@ pub async fn init_logging() -> anyhow::Result<()> {
 }
 
 pub async fn run_cli() -> anyhow::Result<()> {
+    init_logging().await?;
+    if let Some(path) = init_environment() {
+        let path = path.display();
```
xgreenx, line 108: Is it possible to add a test to verify that it works properly with the `.env` file? Maybe even we can have a snapshot of `.env` file to track all future modificaitons?

xgreenx (approved):

bin/fuel-core/Cargo.toml
```diff
@@ -40,6 +41,7 @@ test-case = { workspace = true }
 [features]
 debug = ["fuel-core/debug"]
 default = ["debug", "metrics", "relayer", "rocksdb"]
+env = ["dep:dotenvy"]
```
xgreenx, line 44: Maybe it also should be part of the `production` and `debug` feature

bvrooman: I can understand adding it to the `debug` feature. I can definitely add it there.

I was looking at some of the deployment architecture to see if there would be any conflicts here, because I don't know how this will interact with existing environment variables like the ones set in K8s any configs or helm. It may override them. Is there any risk here?

xgreenx: If we don't use `.env`, it shouldn't affect the current operator or chart deployment

bvrooman:
I'm pushing changes to log an error statement when the `.env` file is missing when compiled with the `env` feature. I've also added `env` to both `debug` and `production` features.

I've been working on adding the above mentioned tests, but this is still a work in progress. See the [WIP PR here](https://github.com/FuelLabs/fuel-core/pull/1313). I've been trying an approach where I run the compiled binary and supply different arguments to the CLI to test the behaviour. However, this is proving to be complex, as it requires the compiled binary to exist prior. This makes it complicated to test different feature sets.

I think the tests can be pushed separately as to unblock this PR and the docs PR. @xgreenx @Voxelot If you can reapprove the PR in its current state, I will push it and finish these current PRs.

Voxelot (approved):

PR #1309, opened 2023-08-17
docs: Debugging instructions using IDEs

bvrooman:
Related issues:
- Closes https://github.com/FuelLabs/fuel-core/issues/1236

This PR adds documentation on debugging.


docs/developers/debugging.md
```diff
@@ -0,0 +1,131 @@
...
+
+When running the client with the P2P service enabled, i.e., building the binary with the `p2p` feature and supplying the runtime argument `--enable-p2p`, the client will connect to a Fuel network. This requires additional CLI arguments, including the `--keypair` and `--network` arguments.
+
+A key pair can be generated by running the utility binary `fuel-core-keygen`:
+
+```bash
+cargo run --bin fuel-core-keygen new
+```
+
+The resulting key pair printed to the console contains an address and secret. The `address` can be provided to the `--keypair` argument when running the node.
```
Voxelot, line 59:
```suggestion
The resulting key pair printed to the console contains an address and secret. The `secret` can be provided to the `--keypair` argument when running the node.
```

bvrooman: Thanks for the clarification!

Voxelot (dismissed):

docs/developers/debugging.md
```diff
@@ -0,0 +1,131 @@
...
+
+A key pair can be generated by running the utility binary `fuel-core-keygen`:
+
+```bash
+cargo run --bin fuel-core-keygen new
+```
+
+The resulting key pair printed to the console contains an address and secret. The `secret` can be provided to the `--keypair` argument when running the node.
+
+The `--network` argument identifies the name of the network to join. The network name is used during peer discovery. For example, users can specify `--network beta3-prod` to join the Beta 3 Production network. Similarly, setting the environment variable `NETWORK="beta3-prod"` will produce the same result.
```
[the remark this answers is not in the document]
bvrooman: Question just for my own understanding: If the names don't match, will the peers even discover each other? Is it not considered a separate network?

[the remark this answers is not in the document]
xgreenx: Nope, they will not discover each other. The connection will be reset during authentication process before we started exchange data.

Voxelot:
minor conflict to resolve in the changelog, otherwise the docs look good

xgreenx (approved):
It would be nice to get some feedback from @luizstacio =)

xgreenx:
The issue also mentioned instructions on how to connect to the network. It's already done in https://github.com/FuelLabs/docs-hub/pull/36, so maybe you can add a link to our documentation instead.

bvrooman:
> It would be nice to get some feedback from @luizstacio =)

Certainly, it is always good to get more feedback. I will wait for any comments from Luiz

> The issue also mentioned instructions on how to connect to the network. It's already done in [FuelLabs/docs-hub#36](https://github.com/FuelLabs/docs-hub/pull/36), so maybe you can add a link to our documentation instead.

Thanks for pointing that out. I will have a look, and link the appropriate networking documentation.

docs/developers/debugging.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Prerequisites
+
+Running Fuel Core with CLion requires the official JetBrains Rust plugin. Find the Rust plugin by navigating to CLion's Plugin menu, entering "Rust" in the search. Install it by clicking "Install".
+
+### Create A Run Configuration
+
+1. In the top-right corner of the CLion window, click on the configurations dropdown menu and select "Edit Configurations..."
+2. In the "Run/Debug Configurations" window, click on the "+" button and select "Cargo"
+3. In the "Name" field, give your configuration a descriptive name, such as "Run Beta 3"
```
luizstacio, line 83:
```suggestion
3. In the "Name" field, give your configuration a descriptive name, such as "Run Fuel Client"
```

Just to remove the need to update the docs. Maybe is better to set a generic name

bvrooman:
I updated the networking section with a link to the Fuel guide on Vercel. I removed some of the text explaining client networking, since the official documentation can do that better. I left the parts of the section that talk about configuration specifically, like setting the CLI argument or providing an environment variable.

docs/developers/debugging.md
bvrooman: Yep, that makes sense 👍

docs/developers/debugging.md
```diff
@@ -0,0 +1,133 @@
...
+4. In the "Command" field, enter `run --all-features --bin fuel-core -- run <ARGUMENTS>`, where `<ARGUMENTS>` represents the desired CLI arguments. Refer to the CLI documentation for guidance on which options and arguments to use.
+5. If you use environment variables for configuration (as indicated in the CLI documentation), you can set them in the "Environment variables" section. Enter the variables and their values as needed. It is recommended to set the `RUST_LOG` environment variable to either `info` or `debug`.
+6. Click "Apply" or "OK" to save your run configuration.
+
+### Example Configuration
+
+![clion_configuration.png](assets/clion_configuration.png)
```
luizstacio, line 90: Following the steps, I had to do;
5. Select the workspace folder
6. Remove on the before start section the build action

luizstacio: Once I try to run it throws the error: 
```
ERROR fuel_core::cli: 57: Unable to load .env environment variables: path not found. Please check that you have created a .env file in your working directory.
```

docs/developers/debugging.md
```diff
@@ -0,0 +1,133 @@
...
+
+Running Fuel Core with Visual Studio Code requires the CodeLLDB extension. Find the CodeLLDB extension by navigating to the "Extensions" tab and searching "CodeLLDB". Install it by clicking "Install".
+
+### Create A Run Configuration
+
+1. In VS Code, click on the "Run and Debug" icon in the left sidebar.
+2. VS Code will offer to create a `launch.json` file to configure how your client node is run. Clicking this will prompt you to select a debugger. Select the CodeLLDB debugger to proceed. VS Code will generate a `launch.json` file for you.
+3. In the "Run and Debug" menu, click the dropdown menu and select "Add Configuration..."
+4. In the `"type"` field, enter `"lldb"`. In the `"request"` field, enter `"launch"`.
+5. In the `"name"` field, give your configuration a descriptive name, such as `"Run Beta 3"`
```
luizstacio, line 110:
```suggestion
5. In the `"name"` field, give your configuration a descriptive name, such as `"Run FuelCore"`
```

docs/developers/debugging.md
```diff
@@ -0,0 +1,133 @@
...
+6. In the `"args"` array nested under `"cargo"`, enter the following arguments: `"run"`, `"--all-features"`, `"--bin"`, `"fuel-core"`, `"--"`, `"run"`, `<ARGUMENTS>`, where `<ARGUMENTS>` is a list of CLI values and arguments encapsulated by quotes. Refer to the CLI documentation for guidance on which options and arguments to use.
+7. Save these changes made to your `launch.json` file.
+
+### Example Configuration
+
+![vs_code_configuration.png](assets/vs_code_configuration.png)
```
luizstacio, line 116: We could just past here the plain json, to make it easier to copy and paste

docs/developers/debugging.md
bvrooman: Thanks for going through the steps!

I think it would be a good idea for me to add a quick line adding the workspace folder. Usually, CLion should populate this line for you to your repository root (don't change it to what's exactly in the example!) But if that is missing for whatever reason, it would be helpful to know that you should set this to the root yourself. I see that I omitted that instruction.

The "Before start" settings are optional. CLion adds "Build" by default, but it's okay because `cargo run` does a build by default. I will add a note on these settings too.

bvrooman: > Once I try to run it throws the error:
> 
> ```
> ERROR fuel_core::cli: 57: Unable to load .env environment variables: path not found. Please check that you have created a .env file in your working directory.
> ```

This is actually a new feature and it's deliberate. It simply means the client is expecting you to provide a `.env` file (probably because you compiled with `--all-features`), but it didn't find it. Simply add a `.env` file in the root of your repo with your specific configuration values, and you're good to go.

docs/developers/debugging.md
luizstacio: I was not able to make the debug work, it looks like the command never finishes. It starts but not finishes.

My config;
```
[
    "run",
    "--all-features",
    "--bin",
    "fuel-core",
    "--",
    "run",
    "--ip", "0.0.0.0",
    "--port", "4004",
    "--db-type", "in-memory",
    "--utxo-validation",
    "--vm-backtrace",
    "--poa-interval-period",
    "30s"
]
```

docs/developers/debugging.md
```diff
@@ -0,0 +1,133 @@
...
+### Prerequisites
+
+Running Fuel Core with CLion requires the official JetBrains Rust plugin. Find the Rust plugin by navigating to CLion's Plugin menu, entering "Rust" in the search. Install it by clicking "Install".
+
+### Create A Run Configuration
+
+1. In the top-right corner of the CLion window, click on the configurations dropdown menu and select "Edit Configurations..."
+2. In the "Run/Debug Configurations" window, click on the "+" button and select "Cargo"
+3. In the "Name" field, give your configuration a descriptive name, such as "Run Beta 3"
+4. In the "Command" field, enter `run --all-features --bin fuel-core -- run <ARGUMENTS>`, where `<ARGUMENTS>` represents the desired CLI arguments. Refer to the CLI documentation for guidance on which options and arguments to use.
```
luizstacio, line 84: I think we could add the example `--db-type in-memory` just to avoid errors.

docs/developers/debugging.md
luizstacio: Yeah, it worked even with the error. The real error was on the database folder: 

```
thread 'main' panicked at 'source slice length (64) does not match destination slice length (32)', <>/.cargo/registry/src/index.crates.io-6f17d22bba15001f/fuel-types-0.36.0/src/array_types.rs:389:16

docs/developers/debugging.md
bvrooman: Luiz and I discussed this specific issue over a call: It appears that the client is starting correctly, and producing the expected output in the terminal, but the VS Code debugger is not correctly attaching to the running process. Luiz and I will spend some time independently looking into this.

docs/developers/debugging.md
bvrooman: I added a Common Issues section to mention using `--db-type in-memory` in case of some errors.

docs/developers/debugging.md
bvrooman: I updated the example with a more reliable config that users can copy + paste to get started.

xgreenx (approved):
Looks good to me, thank you!=)

PR #1318, opened 2023-08-23
chore: Single buffer block sync

bvrooman:
Related issues:
- Closes https://github.com/FuelLabs/fuel-core/issues/1167

PR #1320, opened 2023-08-25
fix: fix import bench values

bvrooman:


bvrooman:
No longer needed now that https://github.com/FuelLabs/fuel-core/pull/1318 is merged.

PR #1321, opened 2023-08-25
Run unit tests without default features only for specific crates

xgreenx:


Voxelot (dismissed):

Voxelot:
fuel-core-chain-config could also be a good one to check since it's used by the SDK as well

bvrooman (approved):

Issue #1297, opened 2023-08-14
Allow `manual_blocks_enabled` only with `debug` feature enabled during compilation


xgreenx:
From the comment: https://github.com/FuelLabs/fuel-core/pull/1286/files#r1290825419

Blocked by https://github.com/FuelLabs/fuel-core/issues/1296

Voxelot:
As discussed, this should be a cli arg feature flag rather than a compile flag. Otherwise it will impact the ability of the released binaries to be used by sdk harnesses etc.

Issue #1296, opened 2023-08-14
Publish a `debug` version of the docker image per release


xgreenx:
Related https://github.com/FuelLabs/fuel-core/pull/1286/files#r1290825419

MitchTurner:
WRT manual block production (and other runtime flags), if we want to do it at runtime rather than at compile time, I think it would be best to have the consensus method fully encapsulated as an `enum` rather than having a bunch of flags.

I'd feel a lot better about exposing those flags in production in that case. Much easier understand the state of the system.

xgreenx:
Clsoed by https://github.com/FuelLabs/fuel-core/pull/1322

PR #1322, opened 2023-08-25
Added support for `debug` CLI flag.

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1297 https://github.com/FuelLabs/fuel-core/issues/1296

Removed `manual_blocks_enabled` flag.
Added a new `debug` flag:
```rust
    /// Should be used for local development only. Enabling debug mode:
    /// - Allows GraphQL Endpoints to arbitrarily advanced blocks.
    /// - Enabled debugger GraphQL Endpoints.
    /// - Allows setting `utxo_validation` to `false`.
    #[arg(long = "debug", env)]
    pub debug: bool,
 ```

Removed `debug` and `dap` features, now the code contains it by default and disables it if `debug` is `false` in the config.

### CODE REVIEW

File: crates/services/consensus_module/poa/src/verifier.rs [hunk 1]
```diff
@@ -63,14 +56,6 @@ pub fn verify_block_fields<D: Database>(
...
     );
 
-    // Skip the verification of the time if it is possible to produce blocks manually.
-    if !config.enabled_manual_blocks {
-        ensure!(
-            header.time() >= prev_header.time(),
-            "The `time` of the next block can't be lower"
-        );
-    }
-
```
xgreenx, line 73: The manual block production increases the time monotonically, so we can remove this check=)

crates/services/consensus_module/poa/src/verifier.rs
```diff
@@ -63,13 +56,10 @@ pub fn verify_block_fields<D: Database>(
...
-    if !config.enabled_manual_blocks {
-        ensure!(
-            header.time() >= prev_header.time(),
-            "The `time` of the next block can't be lower"
-        );
-    }
+    ensure!(
+        header.time() >= prev_header.time(),
+        "The `time` of the next block can't be lower"
+    );
```
xgreenx, line 62: The manual block production increases the time monotonically, so we can remove this check=)

CHANGELOG.md
```diff
@@ -10,12 +10,17 @@ Description of the upcoming release here.
 
 ### Added
 
-- [#1274](https://github.com/FuelLabs/fuel-core/pull/1274): Added tests to benchmark block synchronization.
+- [#1322](https://github.com/FuelLabs/fuel-core/pull/1322): 
+    The `debug` flag is added to the CLI. The flag should be used for local development only. Enabling debug mode:
+    - Allows GraphQL Endpoints to arbitrarily advanced blocks.
+    - Enabled debugger GraphQL Endpoints.
+    - Allows setting `utxo_validation` to `false`.
```
Voxelot, line 17: This may be a breaking change for the sway quickstart guides / lsp tooling that automatically spins up fuel-core. I think it makes sense to protect this setting, but just calling it out that this is a breaking change that needs to be flagged across all the tooling and docs that use fuel-core.

Voxelot: This will also be somewhat difficult to trace down all occurrences, since searching across the repos for `--utxo-validation` will only show the places that already do use proper validation. But there are likely many places that describe how to startup nodes that omit this flag.

crates/fuel-core/src/schema/dap.rs
```diff
@@ -295,25 +322,13 @@ impl DapMutation {
...
-    }
-
-    #[cfg(feature = "debug")]
     async fn set_single_stepping(
         &self,
         ctx: &Context<'_>,
         id: ID,
         enable: bool,
     ) -> async_graphql::Result<bool> {
+        require_debug(ctx)?;
```
MitchTurner, line 331: Where before  `#[cfg(feature = "debug")]` prevented these from showing up in the api at all, will this `require_debug` approach mean that `set_single_stepping` (and the other endpoints) will still be available to non-debug apis, but will always error?

crates/fuel-core/src/schema/dap.rs
```diff
@@ -52,9 +52,13 @@ use tracing::{
 };
 use uuid::Uuid;
 
-#[cfg(feature = "debug")]
 use fuel_core_types::fuel_vm::state::DebugEval;
 
+pub struct Config {
+    /// The `true` means that debugger functionality is enabled.
```
bvrooman, line 58:
```suggestion
    /// `true` means that debugger functionality is enabled.
```

crates/fuel-core/src/service/config.rs
```diff
@@ -39,9 +39,13 @@ pub struct Config {
     pub database_path: PathBuf,
     pub database_type: DbType,
     pub chain_conf: ChainConfig,
+    /// The `true` value:
+    /// - Enables manual block production.
+    /// - Enables debugger endpoint.
+    /// - Allows setting `utxo_validation` to `false`.
+    pub debug: bool,
```
bvrooman, line 46:
```suggestion
    /// When `true`:
    /// - Enables manual block production.
    /// - Enables debugger endpoint.
    /// - Allows setting `utxo_validation` to `false`.
    pub debug: bool,
```

crates/services/consensus_module/poa/src/verifier.rs
Voxelot: even if no time interval is specified?

xgreenx: Yes, the time always is `new >= prev`

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -134,9 +134,9 @@ pub fn init_sub_services(
     };
     let producer_adapter = BlockProducerAdapter::new(block_producer);
 
-    let poa_config: fuel_core_poa::Config = config.try_into()?;
+    let poa_config: fuel_core_poa::Config = config.into();
     let production_enabled =
-        !matches!(poa_config.trigger, Trigger::Never) || config.manual_blocks_enabled;
+        !matches!(poa_config.trigger, Trigger::Never) || config.debug;
```
bvrooman, line 139: 1. It may be useful to have an info message saying that production is enabled or disabled
2. If `debug` turns off block production, what can we do if we need extra debug features to debug block production? This also seems to imply that we shouldn't add debug features related to block production, which isn't obvious.

crates/types/Cargo.toml
```diff
@@ -19,7 +19,7 @@ version = { workspace = true }
 [dependencies]
 anyhow = { workspace = true }
 derive_more = { version = "0.99" }
-fuel-vm-private = { workspace = true }
```
Voxelot, line 22: Are we confident in enabling the debug feature for fuel-vm by default? It seems like there's a lot of extra code that gets enabled to check for breakpoints on every step etc.

MitchTurner (commented):
Is there anything preventing a "debug" node from participating in consensus with the network? I'm not sure if I saw anything like that.

Also. What does `dap` stand for?

tests/tests/blocks.rs
```diff
@@ -155,53 +153,27 @@ async fn produce_block_manually() {
...
 
     let new_height = client.produce_blocks(5, None).await;
 
     assert_eq!(
-        "Response errors; Manual Blocks must be enabled to use this endpoint",
+        "Response errors; `debug` must be enabled to use this endpoint",
         new_height.err().unwrap().to_string()
     );
-
-    let tx = Transaction::default_test_tx();
```
Voxelot, line 171: Was this removed because it wasn't relevant to testing the functionality of the feature flag?

crates/fuel-core/src/service/config.rs
```diff
@@ -79,8 +84,8 @@ impl Config {
             database_type: DbType::RocksDb,
             #[cfg(not(feature = "rocksdb"))]
             database_type: DbType::InMemory,
+            debug: true,
```
Voxelot, line 87: Are we sure that we want to always have debug enabled by default? Could this lead to unexpected results in some tests?

crates/fuel-core/src/schema/dap.rs
xgreenx: Yes, it is what we discussed=) So it is totally expected behavior

crates/types/Cargo.toml
xgreenx: Yeah, it seems this part is heavy enough 
image

We definitely need to optimize it on `fuel-vm` side.

I will create a separate PR in the `fuel-vm` repository. Meanwhile, I think we can merge this PR without waiting for the VM part.

tests/tests/blocks.rs
xgreenx: It was not relevant to testing, plus it used `utxo_validation = false` which is not possible anymore with `debug = false`.

crates/fuel-core/src/service/config.rs
xgreenx: It only broke the test that expected `manual_blocks_enabled = false`. It is `true` by default to not change the behavior for `utxo_validation`

crates/fuel-core/src/service/sub_services.rs
xgreenx: 1. Done
2. `debug` only enables production in the case when it should be off. What do you mean by "debug block production"? I don't see how we can do that=)

CHANGELOG.md
xgreenx: If this flag is omitted, then it is `true` by default, and we don't need to worry=)

xgreenx:
> Is there anything preventing a "debug" node from participating in consensus with the network? I'm not sure if I saw anything like that.

You still can participate, why not until you will produce block manually=)

> Also. What does dap stand for?

It is a debugger that allows you to put breakpoints and iterate over the execution step by step.

crates/fuel-core/src/service/sub_services.rs
bvrooman: That  makes sense. I misread `production_enabled` as `production_disabled`, and I was confused why debug would turn off block production

crates/types/Cargo.toml
xgreenx: https://github.com/FuelLabs/fuel-vm/pull/554

bvrooman (dismissed):
I have no more questions/comments about the code itself. 

Excuse me if we've already discussed this and I am just forgetting - what is the motivation behind removing the `debug` compilation making this all runtime?

bvrooman (approved):

crates/fuel-core/src/schema/dap.rs
MitchTurner: I'm kinda uncomfortable with exposing api endpoints to the standard user that they can never hit. But it's not a huge concern, it just seems messy.

Issue #1138, opened 2023-04-17
Pyroscope for fuel-core


Voxelot:
use pprof-rs and pyroscope in deployments to obtain realtime profiling info about fuel-core

PR #1324, opened 2023-08-25
Pyroscope profiling

Voxelot:
Reopening work from the previous pyroscope PR now that grafana actually supports pyroscope to work with our current master branch: https://github.com/FuelLabs/fuel-core/pull/1170

This will enable real-time flame graphs from fuel-core in deployed environments, making it easier to diagnose performance bottlenecks on the fly based on real world usage.

Closes https://github.com/FuelLabs/fuel-core/issues/1138


.github/workflows/ci.yml
```diff
@@ -284,6 +284,57 @@ jobs:
           cache-from: type=registry,ref=ghcr.io/fuellabs/fuel-core-build-cache:latest
           cache-to: type=registry,ref=ghcr.io/fuellabs/fuel-core-build-cache:latest,mode=max
 
+  # duplicate of publish-docker-image, but with profiling features enabled
+  # this is split into a separate action since it takes longer to build
+  publish-docker-image-profiling:
```
xgreenx, line 289: I guess we only want to do that during release?

.github/workflows/ci.yml
```diff
@@ -284,6 +284,57 @@ jobs:
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push the image to ghcr.io
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: deployment/Dockerfile
+          build-args: |
+            "FEATURES=production,profiling"
+            "DEBUG_SYMBOLS=true"
```
xgreenx, line 331: I didn't know that `ARG FEATURES="production"` and `ARG DEBUG_SYMBOLS=false` will not override env variables if you set them before=) Nice to know

bin/fuel-core/src/cli/run.rs
```diff
@@ -332,20 +344,46 @@ impl Command {
...
+    let config = command.get_config()?;
+
+    // start profiling agent if url is configured
+    #[cfg(feature = "profiling")]
+    let _profiling_agent = profiling
+        .pyroscope_url
+        .as_ref()
+        .map(|url| -> anyhow::Result<_> {
+            // Configure profiling backend
+            let agent = PyroscopeAgent::builder(url, &"fuel-core".to_string())
```
xgreenx, line 372: Do we want to run it here, or maybe put it along with other services? I'm okay to keep it here, but maybe it is better to follow the same pattern

.github/workflows/ci.yml
Voxelot: Following from continuous delivery best practices, every commit should be deployable

bin/fuel-core/src/cli/run.rs
Voxelot: It may make more sense to keep it at the bin level. This agent is more similar to setting up a tracing subscriber, since it's process-wide and there may be unexpected effects caused if multiple nodes have the profiling agent enabled within the same process (i.e. tests).

bin/fuel-core/src/cli/run.rs
```diff
@@ -193,6 +201,9 @@ pub struct Command {
     /// Time to wait after submitting a query before debug info will be logged about query.
     #[clap(long = "query-log-threshold-time", default_value = "2s", env)]
     pub query_log_threshold_time: humantime::Duration,
+
+    #[clap(flatten)]
+    pub profiling: profiling::ProfilingArgs,
```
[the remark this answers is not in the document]
bvrooman: For the sake of consistency, I could understand removing the cfg, like we've done elsewhere. This way, you don't have conditional CLI args. This would be better than situation where you have added the args, but the associated behaviour is removed due to the compilation configuration.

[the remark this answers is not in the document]
Voxelot: I removed the feature flag - the only downside is that this may increase compile times for those installing fuel-core for local dev testing.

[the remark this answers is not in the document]
bvrooman: I think that's a reasonable tradeoff for more consistent and predictable behaviour.

bvrooman (approved):

PR #1331, opened 2023-08-28
Report Peers that give bad Block Info

MitchTurner:
First implementation of https://github.com/FuelLabs/fuel-core/issues/943

During block import, if the peer gives incorrect values, dock their peer score. If they successfully report a block, improve their peer score.


crates/fuel-core/src/service/adapters/sync.rs
```diff
@@ -81,7 +78,45 @@ impl PeerToPeerPort for P2PAdapter {
...
+
+struct NewPeerReportReason(PeerReportReason);
+
+impl From<PeerReportReason> for NewPeerReportReason {
+    fn from(reason: PeerReportReason) -> Self {
+        Self(reason)
+    }
+}
+
+impl PeerReport for NewPeerReportReason {
```
MitchTurner, line 109: These are arbitrary values. We should do a spike or something to determine good values. Should we move these to a config?

Dentosal (dismissed):
LGTM,. The configuration should be tested somehow, but I think that can be done later; this is already an improvement

MitchTurner:
> LGTM,. The configuration should be tested somehow, but I think that can be done later; this is already an improvement

I agree. There also could be more testing at the 'P2PAdapter' level but the way we've designed it, it's hard to test. I've run into this issue in a couple stories, so might be worth looking at a refactor there :/.

CHANGELOG.md
```diff
@@ -16,14 +16,7 @@ Description of the upcoming release here.
 - [#1286](https://github.com/FuelLabs/fuel-core/pull/1286): Include readable names for test cases where missing.
 - [#1274](https://github.com/FuelLabs/fuel-core/pull/1274): Added tests to benchmark block synchronization.
 - [#1263](https://github.com/FuelLabs/fuel-core/pull/1263): Add gas benchmarks for `ED19` and `ECR1` instructions.
-
-#### Breaking
-
-- [#1322](https://github.com/FuelLabs/fuel-core/pull/1322):
```
Voxelot, line 22: Why was this removed from the breaking section?

crates/services/sync/Cargo.toml
```diff
@@ -24,6 +24,7 @@ fuel-core-trace = { path = "../../trace" }
 fuel-core-types = { path = "../../types", features = ["test-helpers"] }
 mockall = { workspace = true }
 test-case = { workspace = true }
+tracing-subscriber = "0.3.17"
```
Voxelot, line 27: Could we source this from a workspace dep? Ie workspace = true?

crates/services/sync/src/ports.rs
```diff
@@ -12,10 +12,33 @@ use fuel_core_types::{
...
+    /// Did not receive advertised block headers
+    MissingBlockHeaders,
+    /// Report a peer for sending a bad block header
+    BadBlockHeader,
+    /// Did not receive advertised transactions
+    MissingTransactions,
+    /// Received invalid transactions
+    InvalidTransactions,
+    /// Received invalid block
+    InvalidBlock,
```
Voxelot, line 39: If a block that passes all prior checks such as the PoA signature fails to import, we may want to consider this as a more serious case that indicates a fork or out of date client. Maybe the node should just stop syncing altogether if it hits this point rather than downvoting the main PoA node?

This could be a follow up as it likely requires more research.

CHANGELOG.md
MitchTurner: There were two `Breaking` sections.

crates/services/sync/src/ports.rs
MitchTurner: Interesting. I thought the previous checks were to show that the merkle root matches the txs, not that the txs are valid for the current state of the chain. Sounds like I'm missing something.

MitchTurner: Oh. But we've already checked consensus, so that's more important than checking individual txs at this level. Okay. I can just remove this check, I guess?

Voxelot: Maybe just leave it the way it is for now. But my thinking is that if the tx root is correct and matches what the PoA node signed in the header, then it's likely a logical error in the client such as being out of date etc

MitchTurner: Oh. Well. I already removed it.

crates/fuel-core/src/service/adapters/sync.rs
Voxelot: Having a config struct or constants file containing all the reputation values seems useful. They could still be hardcoded for now, but it would let us manage all the reputation behaviors from a single place that's easy to find.

crates/services/sync/src/import.rs
```diff
@@ -214,17 +218,27 @@ where
...
+
+                    let res = execute_and_commit(executor.as_ref(), &state, block).await;
+                    match &res {
+                        Ok(_) => {
+                            let _ = p2p.report_peer(peer_id.clone(), PeerReportReason::SuccessfulBlockImport)
+                                .await
+                                .map_err(|e| tracing::error!("Failed to report successful block import for peer {:?}: {:?}", peer_id, e));
+                        },
+                        Err(_) => {
+                            // If this fails, then it means that consensus has approved a block that is invalid.
```
Voxelot, line 237: We should at least log that this is a serious error instead of silently failing. We could even introduce some kind of error backoff or shutdown the sync service altogether so we don't keep importing the same bad block over and over every 5s.

MitchTurner: I'll log it for now. Any concept of sending a shutdown signal or something would fall out-of-scope for this story. Happy to create a new issue.

crates/fuel-core/src/service/adapters/sync.rs
MitchTurner: Okay. I added a config to `P2PAdapter` it's just hardcoded in the place where we init the adapter, but it gives us options in the future.

crates/services/sync/src/import.rs
```diff
@@ -381,21 +409,62 @@ async fn get_headers_batch(
...
+                            PeerReportReason::MissingBlockHeaders,
+                        )
+                        .await
+                        .map_err(|e| {
+                            tracing::error!(
+                                "Failed to report bad block header from peer {:?}: {:?}",
+                                peer_id,
+                                e
+                            )
+                        });
```
bvrooman, line 462: If the report fails, it looks like the event is permanently lost. Under what circumstances could this fail?

crates/services/sync/src/import.rs
bvrooman: I think this should be captured in a new issue and completed separately.

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -109,8 +109,23 @@ pub fn init_sub_services(
...
+    let p2p_adapter = {
+        use crate::service::adapters::PeerReportConfig;
+
+        // Hardcoded for now, but left here to be configurable in the future.
+        let peer_report_config = PeerReportConfig {
+            successful_block_import: 5.,
+            missing_block_headers: -100.,
+            bad_block_header: -100.,
+            missing_transactions: -100.,
+            invalid_transactions: -100.,
```
bvrooman, line 121: This configuration might belong in the consensus parameters. Can we capture this work in a new issue?

crates/services/sync/src/import.rs
MitchTurner: It depends on the implementation of `PeerToPeerPort` we are using for `p2p`. Potentially it could perform 

WRT our specific impl, it looks like there is some fallible code in converting the `peer_id` as well in sending the peer report request.

crates/fuel-core/src/service/sub_services.rs
MitchTurner: I don't believe this should go in consensus parameters, as it's specific to a single P2P implementation. But I agree that they could be passed in at the same place as the consensus parameters. 

I'll create an issue.

MitchTurner: Actually. I think it might still fall into this current issue:
https://github.com/FuelLabs/fuel-core/issues/943

MitchTurner: https://github.com/FuelLabs/fuel-core/issues/1340

bvrooman (approved):
LGTM

crates/services/sync/src/import.rs
MitchTurner: https://github.com/FuelLabs/fuel-core/issues/1341

crates/fuel-core/src/service/adapters/sync.rs
MitchTurner: https://github.com/FuelLabs/fuel-core/issues/1340

Voxelot (approved):

Issue #1337, opened 2023-08-29
`cargo xtask build` fails to build on master due to missing conversion


bvrooman:
Currently, `cargo xtask build` fails on master (**edit: unreleased**)  due to compilation errors in `crates/fuel-core/src/schema/chain.rs`. 

This blocks building the GQL schema.

An example of this error is:
```
error[E0277]: the trait bound `scalars::U64: std::convert::From<fuel_core_types::fuel_tx::DependentCost>` is not satisfied
   --> crates/fuel-core/src/schema/chain.rs:430:21
    |
430 |         self.0.s256.into()
    |                     ^^^^ the trait `std::convert::From<fuel_core_types::fuel_tx::DependentCost>` is not implemented for `scalars::U64`
```

The root cause of this failure appears to be some of the getter functions inside the implementation of `GasCosts`. Here, we are trying to convert `DependentCost`, which is a 2 dimensional value, to a single dimensional scalar. 

These getter functions don't appear to be used anywhere, so it may be safe to simply update the signature to return a tuple of values when using `DependentCost`.

bvrooman:
Turns out, this failure is the simply the result of using `fuel-core` master with an unreleased version of `fuel-vm` master. Since the version isn't released, failure is expected. The attached PR could be helpful for adopting the next `fuel-vm` version into `fuel-core`.

wuliuqii:
I also get an error when I tried to run `cargo xtask build` with fuel.nix, are they the same problem? I create a issue in https://github.com/FuelLabs/fuel.nix/issues/100.

![image](https://github.com/FuelLabs/fuel.nix/assets/34090258/9999de63-673a-4718-ad20-66770b6b6397)

bvrooman:
That appears to be a separate issue that you're experiencing. If you like, feel free to submit a new ticket for the issue you're experiencing.

Issue #1325, opened 2023-08-25
Use `DependentCost` for `k256`, `s256`, `mcpi` opcodes


xgreenx:
We need to update benchmarks for `k256`, `s256`, `mcpi` opcodes to fill `DependentCost` appropriately, considering arbitrary data. Corresponding change on the `fuel-vm` side: https://github.com/FuelLabs/fuel-vm/pull/537

PR #1338, opened 2023-08-29
chore: Upgrade `fuel-vm` to 0.37.0

bvrooman:
Related issues:
- Closes https://github.com/FuelLabs/fuel-core/issues/1337
- Ref https://github.com/FuelLabs/fuel-core/issues/1325

This PR upgrades `fuel-vm` to version 0.37.0. 

Because the 0.37.0 release of `fuel-vm` uses Rust 1.72, I am also upgrading `fuel-core` CI to 1.72 as well.


xgreenx (dismissed):

bvrooman:
This PR is blocked until a new VM version containing https://github.com/FuelLabs/fuel-vm/pull/537 is released.

xgreenx:
> This PR is blocked until a new VM version containing https://github.com/FuelLabs/fuel-vm/pull/537 is released.

Then how `master` fails to compile?

BTW, if you want, you can create a new release of VM=) @MitchTurner knows how to do that

bvrooman:
> > This PR is blocked until a new VM version containing [FuelLabs/fuel-vm#537](https://github.com/FuelLabs/fuel-vm/pull/537) is released.
> 
> Then how `master` fails to compile?
> 
> BTW, if you want, you can create a new release of VM=) @MitchTurner knows how to do that

It looks like I was using `fuel-core` on master, but I modified the dependencies to point to my local `fuel-vm` on master, rather than `v0.36.1`. So the failure here is expected. This fix will be necessary when upgrading. It looks like this was the only change that broke compilation.

crates/fuel-core/src/query/subscriptions/test.rs
```diff
@@ -1,3 +1,6 @@
+// Allow `arc_with_non_send_sync` to enable use of `Arbitrary`
+#![allow(clippy::arc_with_non_send_sync)]
```
bvrooman, line 2: `arc_with_non_send_sync` is a new Clippy lint that was introduced in Rust 1.72.0 (see https://rust-lang.github.io/rust-clippy/master/index.html#/arc_with_non_send_sync). Currently, structs that derive `Arbitrary` fail this lint due to its dependency on `Proptest`, which in turn uses `Arc`.

bvrooman: Looks like this was actually fixed, and we just need the update to propagate upstream: https://github.com/proptest-rs/proptest/issues/364

bvrooman:
Before merging this, we should merge the base branch https://github.com/FuelLabs/fuel-core/pull/1358 that upgrades CI to Rust 1.72.0. This is the rust version used by fuel-vm CI.

crates/client/src/client/schema/chain.rs
```diff
@@ -264,6 +265,7 @@ impl From<ConsensusParameters> for fuel_core_types::fuel_tx::ConsensusParameters
             fee_params: params.fee_params.into(),
             chain_id: params.chain_id.0.into(),
             gas_costs: params.gas_costs.into(),
+            base_asset_id: AssetId::default().into(),
```
MitchTurner, line 268: Why does one version of `ConsensusParameters` have the `base_asset_id` field and another doesn't?

Also, Idk how I feel about using `default` here. It kinda feels arbitrary.

crates/services/txpool/src/test_helpers.rs
```diff
@@ -49,7 +49,7 @@ pub(crate) fn add_coin_to_state(input: Input, mock_db: Option<&MockDb>) -> (Coin
     let coin = CompressedCoin {
         owner: *input.input_owner().unwrap(),
         amount: TEST_COIN_AMOUNT,
-        asset_id: *input.asset_id().unwrap(),
+        asset_id: *input.asset_id(&AssetId::BASE).unwrap(),
```
MitchTurner, line 52: This is about an upstream dep, so not something you can fix, but I don't love this syntax. There is another place where we pass the `AssetId::BASE` into a getter. I've looked at the source code for these methods and it seems like we return the `base_asset_id` for certain variants... so a getter that is a passthrough.

I think it would be better if those variants could know what their base asset is.

Again, not something you need to fix.

crates/client/src/client/schema/chain.rs
bvrooman: `ConsensusParameters` should always have `base_asset_id` now. I might be missing something - can you point me to where you're seeing a version of `ConsensusParameters` that doesn't?

Also, using `default` here is temporary, since the base asset ID configuration isn't done for `fuel-core` - it's currently in progress here: https://github.com/FuelLabs/fuel-core/pull/1339. I had to update the `fuel-vm` side, including the `fuel-tx::ConsensusParameters` struct, to include the base asset ID, so this is the non-parameterized version in the meantime. This will be configured by the `params` once that is updated.

crates/services/txpool/src/test_helpers.rs
bvrooman: I tend to agree that passing a value to a getter is a bit unorthodox, and it might be more desirable for the underlying struct to know what the base asset is. The challenge here specifically is that `Input`s, specifically `Message`s and `MessageCoin`s  are defined by a specification, and this includes their serialization and deserialization. The data is fed in via messages, and the message specification _does not_ include the base asset id, so we would not be able to deserialize a message if we put the base asset on the struct. Similarly, we would have to include workaround logic to handle serialization.

crates/client/src/client/schema/chain.rs
MitchTurner: Oh. Sorry. I meant to ask why you aren't using `params.base_asset_id` and you're manually initializing one. If this is an intermediate state then I won't worry about it.

crates/services/txpool/src/test_helpers.rs
MitchTurner: Yeah. I figured it was something along those lines. It's not the worst thing, just a smell. I'm not gonna worry about it too much.

MitchTurner (approved):

PR #1339, opened 2023-08-29
feat: Configurable base asset

bvrooman:
Related issues:
- Closes https://github.com/FuelLabs/fuel-vm/issues/164


bvrooman:
Currently blocked on a `fuel-vm` release with https://github.com/FuelLabs/fuel-vm/pull/573.

crates/types/src/entities/coins.rs
```diff
@@ -73,10 +73,10 @@ impl CoinType {
     }
 
     /// Returns the asset held by the coin.
-    pub fn asset_id(&self) -> &AssetId {
+    pub fn asset_id<'a>(&'a self, base_asset_id: &'a AssetId) -> &'a AssetId {
         match self {
             CoinType::Coin(coin) => &coin.asset_id,
-            CoinType::MessageCoin(_) => &AssetId::BASE,
+            CoinType::MessageCoin(_) => base_asset_id,
         }
```
Dentosal, line 80: This feels a bit weird to me, shpuld we return `Option<AssetID>` instead, returning `None` for coins? Then the caller could do `.asset_id().unwrap_or(base_asset_id)`.

crates/fuel-core/src/query/balance/asset_query.rs
```diff
@@ -120,6 +123,11 @@ impl<'a> AssetsQuery<'a> {
                 }
             });
 
+        let predicate = self
+            .assets
+            .as_ref()
+            .map(|assets| assets.contains(self.base_asset_id))
+            .unwrap_or(true);
```
Dentosal, line 130: nit: If this evaluates to `false`, we should do a short-circuiting return somehow. Sadly Rust makes this vary hard, but maybe moving the following iterator chain to a helper function would do it.

In any case I don't think this should be named `predicate`, when it's in fact a boolean. Consider something like `has_base_asset`.

Dentosal (dismissed):
Some minor suggestions, but good enough for me already.

crates/types/src/entities/coins.rs
bvrooman: I understand the motivation behind your suggestion, but I would argue that returning `Option` changes the semantics of the signature in way that is incorrect: a coin must always have an asset, in the same way that a coin must always have an amount. It is the unit in which the scalar quantity is measured. 

If we use `Option`, it would require introducing `.unwrap_or(base_asset_id)` as a ubiquitous idiom, but we can't really enforce that. 

I agree that passing the `base_asset_id` to the getter is a weird idiom itself. Unfortunately, this is the most straightforward approach, since we cannot store the asset ID on messages themselves - the serialization/ deserialization specification of messages excludes asset ID, so we have no guaranteed way of querying it from the message data. We need to feed it in externally.

crates/fuel-core/src/schema/coins.rs
```diff
@@ -96,8 +96,10 @@ impl MessageCoin {
         self.0.amount.into()
     }
 
-    async fn asset_id(&self) -> AssetId {
-        fuel_core_types::fuel_types::AssetId::BASE.into()
+    async fn asset_id(&self, ctx: &Context<'_>) -> AssetId {
```
MitchTurner, line 99: Curious why this is called `asset_id` vs `base_asset_id`

MitchTurner:
Looking good. I just have that one question.

crates/fuel-core/src/query/balance/asset_query.rs
Dentosal: Already much more readable. I dislike the `Box` here, but going halfway with

```rust
    pub fn coins(&self) -> impl Iterator<Item = StorageResult<CoinType>> + '_ {
        let has_base_asset = self.has_asset(self.base_asset_id);
        self.coins_iter().chain(self.messages_iter().take_while(move |_| has_base_asset))
    }
```

is way better already. Defining the boolean on line before it's used in `take_while` makes it less painful to understand.

crates/fuel-core/src/schema/coins.rs
bvrooman: If I had to argue in favour of the name `asset_id` (`assetId` in GraphQL parlance), I would say that this naming gives a uniform interface on top of any coin type for retrieving the Asset ID. Specifically, we can query the `assetId` of a `Coin` or `MessgeCoin` without modifying our GQL queries. It also means we don't leak internal details: Calling it `base_asset_id` would be a leaky abstraction because it ties `MessageCoin` to the base layer, even though this is an irrelevant detail at the GQL layer. A user just wants the Asset ID of a given coin - they may not care if it's a regular coin or message coin.

On the other hand, calling it `base_asset_id` explicitly communicates its relation to the base layer. Maybe some users might like that. 

Personally, I'm in favour of keeping it `asset_id` for reasons above, but I am not attached one way or another. If we want to change it, I'm happy to do so.

crates/fuel-core/src/query/balance/asset_query.rs
Dentosal: Right, that's true. We can avoid both the database read and extra `Box` by doing this:

```rust
    pub fn coins(&self) -> impl Iterator<Item = StorageResult<CoinType>> + '_ {
        let messages = self
            .has_asset(self.base_asset_id)
            .then(|| self.messages_iter());
        self.coins_iter().chain(messages.into_iter().flatten())
    }
```

bvrooman: Nice! Updated.

crates/fuel-core/src/schema/coins.rs
Voxelot: Makes sense given the context that this is on a MessageCoin, as it should be implicit that the amount of the asset is related to the base id.

However, from a user perspective it could also be confusing if they are trying to understand how to parse messages. Since if an erc20 is bridged, they may confuse the amount and asset with the erc20 coin they wanted to bridge, which would actually be encoded into the data section.

Since it's implicit anyways, do we even need the asset field here at all?

bvrooman: Yeah, you could argue we don't even need it here. If it has a greater potential for added confusion, rather than for added benefit, we could axe it. If user's goal is to get the base asset ID, they can always get it from the consensus parameters:
Screenshot 2023-09-13 at 5 42 27 PM

Removing it is a bit tangential to this PR, so I can do it here if we prefer, or in a follow up PR.

Dentosal (approved):
I think this can be merged in the current form, and the mentioned asset field removal could be a follow up it it's desired.

bvrooman:
I'm creating a new issue to remove `asset_id` from `MessageCoin`: https://github.com/FuelLabs/fuel-core/issues/1365. This can be done in a follow up.

Issue #1278, opened 2023-08-06
Process the storage error during response


xgreenx:
<img width="1597" alt="image" src="https://github.com/FuelLabs/fuel-core/assets/18346821/885aef98-57b5-40df-a9e0-e35dbfe077fb">

If the node has problems retrieving data from the storage, we need to log it and send the `None` back to the caller instead of silent do nothing.

MitchTurner:
I'm gonna wait until https://github.com/FuelLabs/fuel-core/issues/1133 is done before working on this.

xgreenx:
Closed with https://github.com/FuelLabs/fuel-core/pull/1342

PR #1342, opened 2023-09-01
Add error handling for P2P requests to return `None` to requester and log error

MitchTurner:
https://github.com/FuelLabs/fuel-core/issues/1278


MitchTurner:
I just realized why there were 2 `Breaking` sections in the last PR and replaced it--I realized that it was a subsection of `Removed`. IDK if that's standard semver... but I feel bad for removing it without discussing/checking.

Voxelot (approved):

Voxelot:
What happens on the caller side of the responder returns none? Does the channel to the p2p messaging close or stay open indefinitely?

MitchTurner:
> What happens on the caller side if the responder returns none? Does the channel to receive the p2p response close or stay open indefinitely?

They're already expecting an Option.

Issue #1163, opened 2023-04-26
Add metrics for blocks and transaction stats


Voxelot:
We should have additional Prometheus metrics for the following:
- Current block height of the node: this will allow us to observe any delays, stalls or forks during block sync
- transaction counter: this will enable us to provide a total count of transactions hitting the network, as well as enabling the ability to track TPS. We may also want to persist this total tx count to the database so that its updated between restarts.
- block execution time: we should measure how long it takes to process blocks over time. This will provide useful insight into how accurate our gas metering is and other parameters like block production interval time.

PR #1355, opened 2023-09-07
Importer Metrics

Voxelot:
closes: #1163 

Adds the following metrics to prometheus:

- `importer_tx_count`
- `importer_block_height`
- `importer_latest_block_commit_timestamp_s`
- `importer_execute_and_commit_duration_s`

Additional related changes:
- New field in the metadata table to keep track of the total number of transactions so that the metric is accurate across restarts
- Removed the compiler feature flag for "metrics" since it was sneaked into all builds anyways with the addition of futures tracking.

This allows us to _finally_ start measuring the following:
- Transactions per second
- Delays in synchronization (ie. comparing the block heights between nodes, and comparing the committed timestamps)
- Various measures of time related to block importing/execution on synchronizing nodes (i.e. validation mode) such as avg, max, min etc

### DIFF
--- a/crates/fuel-core/src/database/metadata.rs
+++ b/crates/fuel-core/src/database/metadata.rs
@@ -59,7 +59,7 @@ impl Database {
         Ok(new_tx_count)
     }
 
-    pub fn get_tx_count(&self) -> DatabaseResult<usize> {
+    pub fn get_tx_count(&self) -> DatabaseResult<u64> {
         self.get(TX_COUNT, Column::Metadata)
             .map(|v| v.unwrap_or_default())
     }
--- a/crates/fuel-core/src/database/metadata.rs
+++ b/crates/fuel-core/src/database/metadata.rs
@@ -8,6 +8,9 @@ use fuel_core_chain_config::ChainConfig;
 
 pub(crate) const DB_VERSION_KEY: &[u8] = b"version";
 pub(crate) const CHAIN_NAME_KEY: &[u8] = b"chain_name";
+/// Tracks the total number of transactions written to the chain
+/// It's useful for analyzing TPS or other metrics.
+pub(crate) const TX_COUNT: &[u8] = b"total_tx_count";
 
 /// Can be used to perform migrations in the future.
 pub(crate) const DB_VERSION: u32 = 0x00;
@@ -44,5 +47,20 @@ impl Database {
 
     pub fn get_chain_name(&self) -> DatabaseResult<Option<String>> {
         self.get(CHAIN_NAME_KEY, Column::Metadata)
+    }
+
+    pub fn increase_tx_count(&self, new_txs: u64) -> DatabaseResult<u64> {
+        // TODO: how should tx count be initialized after regenesis?
+        let current_tx_count: u64 =
+            self.get(TX_COUNT, Column::Metadata)?.unwrap_or_default();
+        // Using saturating_add because this value doesn't significantly impact the correctness of execution.
+        let new_tx_count = current_tx_count.saturating_add(new_txs);
+        self.insert::<_, _, u64>(TX_COUNT, Column::Metadata, &new_tx_count)?;
+        Ok(new_tx_count)
+    }
+
+    pub fn get_tx_count(&self) -> DatabaseResult<usize> {
+        self.get(TX_COUNT, Column::Metadata)
+            .map(|v| v.unwrap_or_default())
     }
 }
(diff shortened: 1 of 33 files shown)


crates/fuel-core/src/database/metadata.rs
```diff
@@ -45,4 +48,19 @@ impl Database {
...
+        // TODO: how should tx count be initialized after regenesis?
+        let current_tx_count: u64 =
+            self.get(TX_COUNT, Column::Metadata)?.unwrap_or_default();
+        // Using saturating_add because this value doesn't significantly impact the correctness of execution.
+        let new_tx_count = current_tx_count.saturating_add(new_txs);
+        self.insert::<_, _, u64>(TX_COUNT, Column::Metadata, &new_tx_count)?;
+        Ok(new_tx_count)
+    }
+
+    pub fn get_tx_count(&self) -> DatabaseResult<usize> {
```
bvrooman, line 62: Should this be a `u64`, since it's being inserted into the DB as a `u64`?

crates/fuel-core/src/service/adapters/block_importer.rs
```diff
@@ -113,6 +115,10 @@ impl ImporterDatabase for Database {
     fn latest_block_height(&self) -> StorageResult<BlockHeight> {
         self.latest_height()
     }
+
+    fn update_tx_count(&self, new_txs_count: u64) -> StorageResult<u64> {
```
bvrooman, line 119: Maybe `update_tx_count` should also be named `increase_tx_count`, and the inner call should be namespaced. The name `update_tx_count` suggests that we are setting the TX count to the argument we pass in, but the inner behaviour is that we are setting the TX count to the current TX count + the argument we pass in.

crates/metrics/src/importer.rs
```diff
@@ -0,0 +1,66 @@
...
+            registry,
+            total_txs_count: tx_count_gauge,
+            block_height: block_height_gauge,
+            latest_block_import_timestamp: latest_block_import_ms,
+            execute_and_commit_duration,
+        }
+    }
+}
+
+lazy_static! {
```
bvrooman, line 64: It could be worth migrating to `OnceLock` going forward, now that it is part of the standard library.

crates/services/importer/src/importer.rs
```diff
@@ -223,12 +232,55 @@ where
...
+        // Update the total tx count in chain metadata
+        let total_txs = db_after_execution
+            // Safety: casting len to u64 since it's impossible to execute a block with more than 2^64 txs
+            .update_tx_count(result.sealed_block.entity.transactions().len() as u64)?;
+
         db_tx.commit()?;
 
-        tracing::info!("Committed block");
+        // update the importer metrics after the block is successfully committed
+        IMPORTER_METRICS.total_txs_count.set(total_txs as i64);
```
bvrooman, line 243: Is a signed type necessary here? Looking at the [documentation](https://docs.rs/prometheus-client/latest/prometheus_client/metrics/gauge/struct.Gauge.html#using-atomicu64-as-storage-and-f64-on-the-interface), `i64` is the default, but it appears that this can be configured to a signed type if we so choose.

crates/fuel-core/src/database/metadata.rs
MitchTurner: Is the size bounded by `u64` or `usize` or neither? If neither, I'd say we use `u64` everywhere.

Voxelot: ah yep, missed this one

crates/services/importer/src/importer.rs
Voxelot: It only uses the AtomicU64 if the gauge is an f64

crates/services/importer/src/importer.rs
```diff
@@ -223,12 +232,55 @@ where
             .seal_block(&block_id, &result.sealed_block.consensus)?
             .should_be_unique(&expected_next_height)?;
 
+        // Update the total tx count in chain metadata
+        let total_txs = db_after_execution
+            // Safety: casting len to u64 since it's impossible to execute a block with more than 2^64 txs
+            .update_tx_count(result.sealed_block.entity.transactions().len() as u64)?;
```
MitchTurner, line 238: Related to about question about `usize` vs `u64`. Why is this being converted to `u64`?

Voxelot: I'll normalize this to u64, mainly because usize is unpredictable.

Voxelot: In this case, I think the conversion is appropriate since there's no way to get the length of an array as anything but a usize.

crates/metrics/src/importer.rs
Voxelot: This expands the scope of the PR a bit, but I went ahead and removed lazy static from the metrics crate entirely.

Voxelot:
I also noticed that we have a cli arg for toggling whether metrics are enabled which isn't being used consistently. I'll take care of that in a followup PR.

bvrooman (approved):

Issue #1348, opened 2023-09-05
Report Peers to punish unusually slow heartbeats


MitchTurner:
We don't want to ban a peer just for being slower than the other peers. There is a risk of banning good peers if we do that too much. But we want to keep an eye on super slow peers or those that have stopped sending heartbeats.

xgreenx:
Fixed with https://github.com/FuelLabs/fuel-core/pull/1356

Issue #1354, opened 2023-09-06
Refactor `fuel-core` services to be abstracted and allow testing


MitchTurner:
There are a few things preventing this code from being tested properly.

Let's look at the P2P service as an example. 

`FuelP2PService` has quite a bit of testing already, but it's incomplete. The tests manually poll the `next_event()` method to get the events and then mock the internal event handling before returning some arbitrary result. The actual event handling (as well as request handling) takes place inside the `Task`, but these tests ignore the `Task` completely.

This _might_ be okay if we could also test the the `Task` directly, but `Task` takes a direct dependency on a concrete `FuelP2PService` and `SharedState`:
```rs
pub struct Task<D> {
    p2p_service: FuelP2PService<PostcardCodec>,
    db: Arc<D>,
    next_block_height: BoxStream<BlockHeight>,
    /// Receive internal Task Requests
    request_receiver: mpsc::Receiver<TaskRequest>,
    shared: SharedState,
    max_headers_per_request: u32,
}
```
So we would need to abstract `p2p_service` so we can feed the `Task` faked events and requests and abstract `shared` to check that we are modifying state correctly, etc.

BUT even if we were to add the abstractions to `Task`, we wouldn't have coverage of how those `Task`s are spun up and operated in the wild. That logic is exempted from the `FuelP2PService` and short circuited by calling `next_event()` directly, where normally the `Task` calls `next_event()`.

What would be ideal is if one place just had full coverage. All we know is that some external party makes a request, the service does a bunch of stuff, and then gives a response and state changes. That's all we should be testing: request in, response/state change out. 

One question is whether or not libp2p should be included in the test scope. We seem to have some logic in the Libp2p code and in the Service code. Ideally, everything tested should be agnostic to _how_ or _where_ the request came from, so if we've coupled our core logic to `libp2p` we might need to mock the peering somehow, if not, then we can abstract the `libp2p` code as an adapter to some kind of request/response port.

MitchTurner:
I don't think this is worth working on just for better test coverage, but imo testable code is maintainable code and it will make our jobs easier in the future.

Voxelot:
I agree that further separation of business logic from external dependencies a-la onion model would be ideal. The relayer crate handles this a bit better already for example, by keeping the core state machine logic disentangled from the ethereum SDK. The p2p logic is highly intertwined with the inner workings of libp2p, so I understand the complexity of testing this. However, it should still be possible with more integration-level tests in the meantime right?

[...]

PR #1356, opened 2023-09-07
Add tracking for peer heartbeats

MitchTurner:
https://github.com/FuelLabs/fuel-core/issues/1348

In order to track peer heartbeats, this PR adds two new things:

- A concept of a rolling/moving average for the time between heartbeats for each peer
    - This is updated every time a heartbeat is sent from that peer
- A regular timeout in the P2P `Task` that will check peers for good heartbeat behavior

_Originally I was looking at just reporting on heartbeat, but if a peer never reported then they would be exempt from checks_

There are ~3~ 2 cases in which we will want to punish peers for their heartbeat behavior:

- ~They have never sent us a heartbeat~ (If we treat the setup as the first heartbeat, then we can merge this with the next 👇)
- They haven't sent us a heartbeat in a long time
- They are sending heartbeats very sporadically

We can also use this data to determine which peers are performing the best


MitchTurner:
Still need to:
- Parameterize the rates at which heartbeats are checked, and possibly the window size for the heartbeat rolling average?
- Actually implement the criteria for punishing the peers in `peer_heartbeat_reputation_checks`

I don't really know what any of these values _should_ be and probably need to discuss with the team.

MitchTurner:
This is a perfect example of why we need this refactor:
https://github.com/FuelLabs/fuel-core/issues/1354

It's difficult to test the surfaces of the `Task` to know that punishments are being divvied out correctly.

bvrooman:
I think the overall idea makes sense. You mentioned two criteria for which we punish peer scores:

1. They haven't sent us a heartbeat in a long time
2. They are sending heartbeats very sporadically

I imagine that for 1., you calculate the time since the last heartbeat and check if it surpasses a threshold. 

For 2., is this where the moving average comes in to play? Do we punish the peer if the time since last heartbeat falls significantly below/above the moving average?

It would be nice to consolidate, if we haven't already, on the architecture for introducing these kinds of "rules" for peer score, i.e. a way to plug inputs in to an extensible rule engine.

MitchTurner:
> I imagine that for 1., you calculate the time since the last heartbeat and check if it surpasses a threshold.

Yes. This already exists in the `peer_heartbeat_reputation_checks` method, which gets called on a regular timeout by the `select!`. And maybe that's enough.

> For 2., is this where the moving average comes in to play?

That's the idea right now. The idea originally was to check on every received heartbeat how that peer was doing. They would be scored based on how often they send heartbeats and the moving average was to introduce some leniency. 

This sounds great to me, except for the case were we find a new peer and then never get any heartbeats from them again. So I realized we need some independent peer watching, which is where the `peer_heartbeat_reputation_checks` came in.

This all seems like it could be simplified some. Maybe we actually get rid of #1 checks? Since if they are taking a long time then the average will catch it.

> It would be nice to consolidate, if we haven't already, on the architecture for introducing these kinds of "rules" for peer score, i.e. a way to plug inputs in to an extensible rule engine.

I wish, but peer scoring is kinda scattered around the code a bit. There isn't one place where we can catch all the behavior. For example, we can't capture any of the sync misbehavior checks in the P2P code since we don't have access to it. For now we just expose the `report_peer` method that other services can have a shared handle to.

Voxelot:
> This all seems like it could be simplified some. Maybe we actually get rid of https://github.com/FuelLabs/fuel-core/issues/1 checks? Since if they are taking a long time then the average will catch it.

Simple is better

MitchTurner:
Oh. I remember why I included the "time since last heartbeat" check. It's because the peer could have had a great average, but then dropped. This is the way to catch if the peer is down.

crates/services/p2p/src/peer_manager.rs
```diff
@@ -291,22 +295,11 @@ impl PeerManager {
             .send(self.reserved_connected_peers.len());
     }
 
-    fn insert_peer_info(&mut self, peer_id: &PeerId, data: PeerInfoInsert) {
-        let peers = if self.reserved_peers.contains(peer_id) {
+    fn get_relevant_peers(&mut self, peer_id: &PeerId) -> &mut HashMap<PeerId, PeerInfo> {
```
Dentosal, line 298: Nit: maybe this could be renamed to `relevant_peers_mut`, since it returns mutable access to underlying data?

crates/services/p2p/src/p2p_service.rs
```diff
@@ -275,7 +275,7 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
             .collect()
     }
 
-    pub fn get_peers_ids(&self) -> impl Iterator<Item = &PeerId> {
+    pub fn get_peers_ids_iter(&self) -> impl Iterator<Item = &PeerId> {
```
Voxelot, line 278: shouldn't it be clear enough from the method signature that it's an iter?

crates/services/p2p/src/peer_manager.rs
```diff
@@ -291,22 +295,14 @@ impl PeerManager {
             .send(self.reserved_connected_peers.len());
     }
 
-    fn insert_peer_info(&mut self, peer_id: &PeerId, data: PeerInfoInsert) {
-        let peers = if self.reserved_peers.contains(peer_id) {
+    fn get_relevant_peers_mut(
```
Voxelot, line 298: It's not clear what this is "relevant" for. Maybe something like this instead?

```suggestion
    fn get_peer_table_mut(
```

Voxelot (dismissed):
lgtm - tested the moving avg code and it seems to work 👍🏻

crates/services/p2p/src/service.rs
```diff
@@ -107,65 +118,271 @@ impl Debug for TaskRequest {
...
 /// and the top level `NetworkService`.
-pub struct Task<D> {
-    p2p_service: FuelP2PService<PostcardCodec>,
+pub struct Task<P, D, B> {
+    p2p_service: P,
     db: Arc<D>,
     next_block_height: BoxStream<BlockHeight>,
     /// Receive internal Task Requests
     request_receiver: mpsc::Receiver<TaskRequest>,
-    shared: SharedState,
```
Voxelot, line 118: minor nit: `shared` is conventional for most of our tasks in fuel core

MitchTurner:
> lgtm - tested the moving avg code and it seems to work 👍🏻 

I was considering adding some tests for this but my natural inclination is to not test internal apis. Maybe this is a good exception because we don't have a clear boundary. Might add that before merging.

crates/services/p2p/src/service.rs
MitchTurner: It is still represented in the Task trait assoc type. I'm not a fan of having a field called `shared` because it doesn't communicate at all what it's for and why it's grouped. `shared` implies to me that it's for misc data and anyone extending it might just throw it in there. Whereas the trait assoc type communicates that it would probably be shared with other Tasks. 

I was hoping to move away from that name :)

Voxelot: In this case I think it's fine, since this task is just mimicking the form, but not rigidly adhering to the standardized task framework from `fuel-core-services`. However, changing it in the broader scope will require a more comprehensive plan to improve.

MitchTurner:
> > lgtm - tested the moving avg code and it seems to work 👍🏻
> 
> I was considering adding some tests for this but my natural inclination is to not test internal apis. Maybe this is a good exception because we don't have a clear boundary. Might add that before merging.

I think I caught some bugs, so I good thing I checked.

Voxelot (approved):

PR #1358, opened 2023-09-08
chore: Upgrade to Rust 1.72.0 in CI

bvrooman:
Upgrades the Rust version used in CI to 1.72.0. This matches the Rust version used in `fuel-vm` which was previously upgraded.


crates/fuel-core/src/query/subscriptions/test.rs
```diff
@@ -1,3 +1,6 @@
+// Allow `arc_with_non_send_sync` to enable use of `Arbitrary`
+#![allow(clippy::arc_with_non_send_sync)]
```
bvrooman, line 2: `arc_with_non_send_sync` is a new Clippy lint that was introduced in Rust 1.72.0 (see https://rust-lang.github.io/rust-clippy/master/index.html#/arc_with_non_send_sync). Currently, structs that `derive` Arbitrary fail this lint due to its dependency on Proptest, which in turn uses `Arc` on `Box`.

MitchTurner (dismissed):
LGTM

MitchTurner (approved):

Issue #1234, opened 2023-06-30
Replace `MessageId` with `Nonce` in `message_proof` endpoint


xgreenx:

Voxelot:
@xgreenx plz add description

PR #1363, opened 2023-09-12
Replace message_id with nonce in `MessageProof` query

MitchTurner:
Closes https://github.com/FuelLabs/fuel-core/issues/1234

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,12 +33,12 @@ Description of the upcoming release here.
 
 ### Breaking
 - [#1363](https://github.com/FuelLabs/fuel-core/pull/1363): Change message_proof api to take `nonce` instead of `message_id`
+- [#1355](https://github.com/FuelLabs/fuel-core/pull/1355): Removed the `metrics` feature flag from the fuel-core crate, and metrics are now included by default.
 - [#1322](https://github.com/FuelLabs/fuel-core/pull/1322):
   The `debug` flag is added to the CLI. The flag should be used for local development only. Enabling debug mode:
     - Allows GraphQL Endpoints to arbitrarily advance blocks.
     - Enables debugger GraphQL Endpoints.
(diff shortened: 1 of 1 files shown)


crates/fuel-core/src/query/message.rs
```diff
@@ -185,6 +186,8 @@ pub fn message_proof<T: MessageProofData + ?Sized>(
         None => return Ok(None),
     };
 
+    let message_id = compute_message_id(&sender, &recipient, &nonce, amount, &data);
```
bvrooman, line 189: If I understand correctly, this is kinda the meat of the change: Rather than using the message id directly, we pass in the nonce. Together, these pieces of information allow us to reconstruct the message id.

bvrooman:
Does this PR close the ticket, or just progress it?

crates/fuel-core/src/query/message.rs
MitchTurner: Yep.  Maybe I could have reworded the contents of the issue.

We were just asking for more information than needed.

MitchTurner:
> Does this PR close the ticket, or just progress it?

Should close it.

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ Description of the upcoming release here.
     - Allows GraphQL Endpoints to arbitrarily advance blocks.
     - Enables debugger GraphQL Endpoints.
     - Allows setting `utxo_validation` to `false`.
+- [#1363](https://github.com/FuelLabs/fuel-core/pull/1363): Change message_proof api to take `nonce` instead of `message_id`
```
bvrooman, line 45: This could just be my interpretation, but I believe we are putting changelog updates at the start of the section - something like a reverse chronological order, so that the newest things are first.

bvrooman (approved):

Issue #1360, opened 2023-09-11
Error is logged when no env is provided, even though it's optional


Voxelot:
According to the coding guidelines, we should only use ERROR level if there is a very serious error. https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/logging.md

However, right now we log an error if we can't load the fuel-core settings from an env file even though a user may prefer to use cli args instead. This is a jarring experience for users and we should just remove the logging here.

PR #1372, opened 2023-09-20
Use right level of error for `.env` file and duplicated service

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1360

### DIFF
--- a/bin/fuel-core/src/cli.rs
+++ b/bin/fuel-core/src/cli.rs
@@ -4,10 +4,6 @@ use std::{
     path::PathBuf,
     str::FromStr,
 };
-use tracing::{
-    info,
-    warn,
-};
 use tracing_subscriber::{
     filter::EnvFilter,
     layer::SubscriberExt,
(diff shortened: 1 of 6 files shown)


bin/fuel-core/src/cli.rs
```diff
@@ -54,7 +48,7 @@ fn init_environment() -> Option<PathBuf> {
     match dotenv() {
         Ok(path) => Some(path),
         Err(e) => {
-            error!("Unable to load .env environment variables: {e}. Please check that you have created a .env file in your working directory.");
```
Voxelot, line 57: I think we should maybe not log anything, since the majority of devs won't use env files and this isn't a helpful message as the first thing they see when starting up fuel-core.

Voxelot: We should probably log something to indicate when a env file is being used instead of CLI args, instead of always notifying people whenever they don't have an env file even if they dont want to use it.

Voxelot (approved):

PR #1375, opened 2023-09-21
Release 0.20.5 with a new endpoint

xgreenx:
Add query and handling for `MessageStatus` (#1371)
The bridge UI requires an client endpoint for querying the status of a specific `Message`, whether it has been spent or not. An additional variant is added for unknown messages, i.e. messages that can't be found in the DB.

---------

Co-authored-by: Green Baneling <XgreenX9999@gmail.com>

(cherry picked from commit c6b0df7b6d6f4c5574a442639ed86865d63cb8f4)


bvrooman (approved):
It sounds like we're cherry picking this PR because the bridging UI is blocked - is this correct?

Issue #1065, opened 2023-03-17
Refactor DiscoveryBehaviour events in p2p


leviathanbeak:
At this moment there is no need for other Behaviours / Swarm to listen to internal Discovery events so they should be removed.
Upon removing the Discovery test should be updated too.

MitchTurner:
> At this moment there is no need for other Behaviours / Swarm to listen to internal Discovery events so they should be removed.

We want to keep the `DiscoveryBehavior`, right? I'm not sure how we remove the `DiscoveryEvent`s without removing the behavior. I don't believe any other behaviors are currently using the `DiscoveryEvent`s...

PR #1377, opened 2023-09-21
Remove `DiscoveryEvent` and use `KademliaEvent` instead

MitchTurner:
Closes https://github.com/FuelLabs/fuel-core/issues/1065

The idea being that we had the `DiscoveryEvent` that didn't serve any special purpose and wasn't being consumed anywhere. This simplifies `DiscoveryBehavior`.

The issue also suggested:
> Upon removing the Discovery test should be updated too.

I'm not really sure what this test is for. The actual code uses `mdns` to identify new peer addresses, whereas the test bypasses this and does it manually on `UnroutablePeer` events. I don't understand the point of this.


MitchTurner:
I might give rewriting the test a try using `libp2p-test-swarm`. It allows you to spin up ephemeral swarms with `MemoryTransport` to interact locally in tests. I might be able to test the `mdns` stuff as well with that :crossed_fingers:

MitchTurner:
> I might give rewriting the test a try using libp2p-test-swarm. It allows you to spin up ephemeral swarms with MemoryTransport to interact locally in tests. I might be able to test the mdns stuff as well with that 🤞

I don't think our current version of `libp2p` supports the `libp2p-test-swarm` extension traits. So this isn't possible without bumping our version :(

MitchTurner:
There are also `mdns` tests, so I assume we could get the discovery test to work completely, without the hacky manual insert into `kademlia`. 
https://github.com/MitchTurner/rust-libp2p/blob/master/protocols/mdns/tests/use-tokio.rs

But it doesn't look like `mdns` is identifying peers the way it's currently working--we aren't getting any `mdns` events. 

I'd again recommend we move to `libp2p-test-swarm` for our tests after we complete:
https://github.com/FuelLabs/fuel-core/issues/1298

I've created an issue for this here:
https://github.com/FuelLabs/fuel-core/issues/1378

crates/services/p2p/src/discovery.rs
```diff
@@ -238,27 +177,26 @@ impl NetworkBehaviour for DiscoveryBehaviour {
                             self.kademlia.add_address(&peer_id, multiaddr);
```
xgreenx, line 177: Hmm, if `list.len() + self.connected_peers.len() > self.max_peers_connected` we will not handle this case=) After looking at how `add_address` works, I think it is okay to call `add_address` without any limitation(I think we can remoev `if self.connected_peers.len() < self.max_peers_connected` check) because it only updates the routing table.

crates/services/p2p/src/discovery.rs
```diff
@@ -184,51 +154,20 @@ impl NetworkBehaviour for DiscoveryBehaviour {
...
+                        ))
+                    }
+                    _ => Poll::Pending,
+                },
+                _ => Poll::Ready(kad_action),
             }
-        }
-
-        while let Poll::Ready(mdns_event) = self.mdns.poll(cx, params) {
+        } else if let Poll::Ready(mdns_event) = self.mdns.poll(cx, params) {
```
xgreenx, line 170: Hmm, the previous iteration would iterate over all mdns events. Let's say, if we had many `NetworkBehaviourAction::GenerateEvent(MdnsEvent::Discovered(list))`, we would process them for one loop.

While right not it is not true. I'm okay to use `if` statement for Kademlia, but for mdns it is better to use `while` as we did before=)

crates/services/p2p/src/discovery.rs
MitchTurner: Yeah. 

````
                    if self.connected_peers.len() < self.max_peers_connected {
                        for (peer_id, multiaddr) in list {
                            self.kademlia.add_address(&peer_id, multiaddr);
                        }
                    }
```
doesn't even work as intended, I think, because it only checks before you add the first peer.

crates/services/p2p/src/discovery.rs
```diff
@@ -184,51 +154,20 @@ impl NetworkBehaviour for DiscoveryBehaviour {
...
-                _ => {}
+                NetworkBehaviourAction::GenerateEvent(event) => match event {
+                    KademliaEvent::UnroutablePeer { peer } => {
+                        Poll::Ready(NetworkBehaviourAction::GenerateEvent(
+                            KademliaEvent::UnroutablePeer { peer },
+                        ))
+                    }
+                    _ => Poll::Pending,
+                },
+                _ => Poll::Ready(kad_action),
```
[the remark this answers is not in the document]
MitchTurner: I had that before, but technically that's not what the code was doing before because it was only matching on the one event variant.

I can switch it if you think it's safe (I think it is too).

crates/services/p2p/src/discovery.rs
```diff
@@ -184,81 +154,43 @@ impl NetworkBehaviour for DiscoveryBehaviour {
...
-                    })
-                }
-                _ => {}
-            }
-        }
-
+        // poll sub-behaviors
+        let mut poll_state = None;
+        if let Poll::Ready(kad_action) = self.kademlia.poll(cx, params) {
+            poll_state = Poll::Ready(kad_action).into();
```
xgreenx, line 160: mdns can override the `poll_state` later. You can use `return` as it was before=) Events from mdns are optional, and the priority of events from Kademlia is higher=)

MitchTurner: Ugh. You're right. I just think `return` is terrible. K. I'll just do that.

xgreenx (approved):

PR #1379, opened 2023-09-22
Update `libp2p` from `0.50.0` to `0.53.1`

MitchTurner:
https://github.com/FuelLabs/fuel-core/issues/1298

### Documentation:
Removed `upgrade::read_length_prefixed` and `upgrade::write_length_prefixed`: https://github.com/libp2p/rust-libp2p/pull/4787/files
Remove `FastMessageId`: https://github.com/libp2p/rust-libp2p/issues/4138
Remove `TokioDnsConfig`: https://github.com/libp2p/rust-libp2p/commit/95890b550bc4659d7b4797daad91bf969440f03a
Implement `InboundConnectionUpgrade`/`OutboundConnectionUpgrade`: https://github.com/libp2p/rust-libp2p/issues/4307


crates/services/p2p/src/discovery.rs
```diff
@@ -2,31 +2,37 @@ use self::mdns::MdnsWrapper;
...
         PollParameters,
+        THandler,
     },
     Multiaddr,
     PeerId,
 };
-use libp2p_kad::KademliaEvent;
+
+// TODO: Figure out the right Handler
+use libp2p_kad::handler::KademliaHandler;
```
MitchTurner, line 31: This doesn't exist, but I haven't figured out what the correct value would be.

xgreenx:
Some functionality regarding connections has been moved to separate behaviors. We need to start to use them inside of the 4 new methods: https://github.com/libp2p/rust-libp2p/blob/master/swarm/CHANGELOG.md#0420

MitchTurner:
Started a convo in the `libp2p` repo. Feel free to through any on-going question in there if you get stuck:
https://github.com/libp2p/rust-libp2p/discussions/4674

crates/services/p2p/src/peer_report.rs
```diff
@@ -305,77 +273,108 @@ impl NetworkBehaviour for PeerReportBehaviour {
...
     }
+
+    fn handle_established_inbound_connection(
+        &mut self,
+        _connection_id: ConnectionId,
+        peer: PeerId,
+        local_addr: &Multiaddr,
+        remote_addr: &Multiaddr,
+    ) -> Result<THandler<Self>, ConnectionDenied> {
+        self.identify
```
MitchTurner, line 299: This is a guess.

Another approach would be to do some kind of check with `identify` to see if it's already handled... and then pass it to `heartbeat` in that case. I'm not sure how important this is though.

MitchTurner:
Tests are compiling, but 16/27 of them run indefinitely 🤪. That's not surprising, since there are some `todo`s left and some new method implementations that might be incomplete, as mentioned above.

crates/services/p2p/src/discovery.rs
```diff
@@ -301,14 +273,20 @@ mod tests {
...
+        // let swarm_builder = SwarmBuilder::without_executor(
+        //     transport,
+        //     behaviour,
+        //     keypair.public().to_peer_id(),
+        // )
+        // .dial_concurrency_factor(NonZeroU8::new(1).expect("1 > 0"));
+
+        // let mut swarm = swarm_builder.build();
+
+        let config = libp2p_swarm::Config::with_tokio_executor()
```
MitchTurner, line 285: This is a guess because there is no `without_executor` concept anymore.

MitchTurner:
My best guess for why the tests are failing, at least the `nodes_cannot_connect_due_to_different_checksum` test, is because  Kademlia dialing is failing. Specifically, the dialing opts aren't including any addresses, so it is rejected repeatedly. This is keeping the nodes from ever making a connection and it's just spinning forever trying to make a connection. 

I'm going to compare the behavior to `master` and see if indeed the addresses exist there. Then I can identify what changes I made that removed the assignment of addresses to peer_ids.

crates/services/p2p/src/discovery.rs
thomaseizinger: For testing, I'd recommend using `libp2p_swarm_test` and the `Swarm::new_ephemeral` function.

crates/services/p2p/src/discovery/discovery_config.rs
```diff
@@ -171,7 +173,7 @@ impl DiscoveryConfig {
 
         // mdns setup
         let mdns = if self.with_mdns {
-            MdnsWrapper::default()
+            MdnsWrapper::new(local_peer_id)
```
thomaseizinger, line 176: Note: We have a `Toggle` behaviour that probably can do what you need here.

MitchTurner:
@thomaseizinger Wow thanks for the review! You didn't need to do that but much appreciated.

crates/services/p2p/src/behavior.rs
```diff
@@ -99,20 +104,24 @@ impl<Codec: NetworkCodec> FuelBehaviour<Codec> {
...
 
         let mut req_res_config = RequestResponseConfig::default();
-        req_res_config.set_request_timeout(p2p_config.set_request_timeout);
-        req_res_config.set_connection_keep_alive(p2p_config.set_connection_keep_alive);
+        req_res_config
+            .clone()
+            .with_request_timeout(p2p_config.set_request_timeout);
+
+        // TODO: Do we need to set this via the `SwarmBuilder` somewhere? https://github.com/libp2p/rust-libp2p/pull/4679
+        // req_res_config.set_connection_keep_alive(p2p_config.set_connection_keep_alive);
```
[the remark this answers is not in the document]
MitchTurner: `with_idle_connection_timeout` is only on the Swarm `Config`, so it would probably have the same effect as using the method on the `SwarmBuilder`. There are the other comments here about using the `SwarmBuilder` so I assume I'd go with whichever method, the builder method or the config method depending on what we land on there.

crates/services/p2p/Cargo.toml
```diff
@@ -22,45 +22,48 @@ fuel-core-types = { workspace = true, features = [
...
+libp2p-identify = "=0.44.0"
+libp2p-kad = "=0.45.1"
+libp2p-mdns = "=0.45.1"
+libp2p-mplex = "=0.41.0"
+libp2p-noise = "=0.44.0"
+libp2p-request-response = "=0.26.0"
+libp2p-swarm = "=0.44.0"
+libp2p-tcp = "=0.41.0"
+libp2p-websocket = "=0.43.0"
+libp2p-yamux = "=0.45.0"
```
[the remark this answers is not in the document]
thomaseizinger: My recommendation would be to only depend on `libp2p` and import everything via that. We tend to issue a lot of patch-releases of the underlying crates to add new features and fix bugs. Pinning them here means you are not going to receive these updates (not sure if dependabot updates them actually).

crates/services/p2p/src/behavior.rs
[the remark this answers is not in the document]
thomaseizinger: Yes, `SwarmBuilder` has a `with_swarm_config` function that allows you to modify the config. If you construct `Swarm` directly then you can just construct the config manually yes.

crates/services/p2p/src/p2p_service.rs
```diff
@@ -196,21 +196,28 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
...
+        //     SwarmBuilder::with_tokio_executor(transport, behaviour, local_peer_id)
+        //         .connection_limits(connection_limits)
+        //         .build();
+
+        let mut swarm = Swarm::new(
+            transport,
+            behaviour,
+            local_peer_id,
+            libp2p_swarm::Config::with_tokio_executor(),
+        );
```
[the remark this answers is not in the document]
MitchTurner: It's mostly because the `with_tokio_executor` constructor wasn't there before but there was one for the config. I guess it's possible the builder is doing some things that I'm missing this way though.

How would I get parity with what I had before?
```
        let mut swarm =
            SwarmBuilder::with_tokio_executor(transport, behaviour, local_peer_id)
                .connection_limits(connection_limits)
                .build();
```
Something like
```
        let mut swarm = SwarmBuilder::with_new_identity()
            .with_tokio()
... I'm not sure
            .build();
```
For example, I don't think there is a `connection_limits` method anymore. I found what I needed going the config route (at least superficially).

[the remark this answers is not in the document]
MitchTurner: I'm realizing that the builder can handle a lot of the other things we are doing above, e.g. setting up transport, etc.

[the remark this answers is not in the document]
thomaseizinger: The builder handles all sorts of nasty details for you that are important in e.g. how you compose transports. For example, DNS must wrap everything else, except relay.

Connection limits is now a `NetworkBehaviour` in its own module! :)

crates/services/p2p/src/discovery.rs
MitchTurner: Okay. This `discovery` test was broken so I switched over to `libp2p_swarm_test` and that seemed to make it run better. It doesn't fail immediately, but it's still broken.

The test creates 25 nodes and tries to connect them all together starting with just one shared peer. It never seems to connect to anyone but the original peer--I can reduce the node count to 2 and the test passes, but at 3 nodes it fails.

It seems to be failing because of protocol negotiation failure:
```
2023-12-02T01:02:29.023409Z DEBUG new_established_connection{remote_addr=/memory/15897544716629232405/p2p/12D3KooWNUZev6j3zKSe8iEKuGW8jHxhGKe9QdeLxcyyWdaqgZBv id=3 peer=12D3KooWNUZev6j3zKSe8iEKuGW8jHxhGKe9QdeLxcyyWdaqgZBv}:Connection::poll: multistream_select::dialer_select: Dialer: Received rejection of protocol: /fuel/kad/test_network/kad/1.0.0    
```
and
```
2023-12-02T01:02:29.023766Z DEBUG Swarm::poll: libp2p_kad::behaviour: Request to peer in query failed with Upgrade(NegotiationFailed) peer=12D3KooWNUZev6j3zKSe8iEKuGW8jHxhGKe9QdeLxcyyWdaqgZBv query=QueryId(4)
```
I've been able to track the problem back as far as a `StreamUpgradeError` in the Connection's `negotiating_out`. I can keep looking up that way but I'm also wondering if this has something to do with the `mdns` stuff that got removed. I'm not sure--this is still a little magic to me. But given this is a pretty core function of kad, I thought you might have some ideas.

MitchTurner: It looks like part of the issue is that Kademlia now has "client" and "server" modes and it defaults to "client". If we change to "server", it passes for 3.

25 is still failing. It seems to break around 10 nodes (10 is flaky). Still investigating that.

thomaseizinger: > It looks like part of the issue is that Kademlia now has "client" and "server" modes and it defaults to "client". If we change to "server", it passes for 3.

Yes. Kademlia will automatically switch to server-mode if the node learns that it has an external address. You can do that as part of listening when you set up the test-swarm:

```rust
let mut swarm1 = Swarm::new_ephemeral(|identity| { ... });
swarm1.listen().with_memory_addr_external().await;
```

thomaseizinger: Any node that should be part of the routing table needs to be in server-mode.

crates/services/p2p/src/config/fuel_authenticated.rs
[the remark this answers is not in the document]
MitchTurner: The idea is that we want to protect certain nodes. We do this by only allowing it to talk to authorized "sentry" nodes. My guess is this minimizes the attack surface of bad actors trying to communicate with the protected node.

crates/services/p2p/src/behavior.rs
```diff
@@ -64,6 +66,9 @@ pub struct FuelBehaviour<Codec: NetworkCodec> {
 
     /// RequestResponse protocol
     request_response: RequestResponse<Codec>,
+
+    /// The Behaviour to manage connections to blocked peers.
+    blocked_peer: allow_block_list::Behaviour<allow_block_list::BlockedPeers>,
```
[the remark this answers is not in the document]
MitchTurner: I had to move the Gossipsub behavior to the top (second) to get some of our tests to pass. Idk if that's a bug or if it's just finicky.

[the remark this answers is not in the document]
thomaseizinger: That is odd, never heard of that before!

crates/services/p2p/src/config/fuel_authenticated.rs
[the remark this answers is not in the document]
thomaseizinger: > The idea is that we want to protect certain nodes. We do this by only allowing it to talk to authorized "sentry" nodes.
> 
> My guess is they implemented this way because it minimizes the attack surface of bad actors trying to communicate with the protected node. I'll ask about more details.

This should be easily doable using the `allow_block_list` module: https://docs.rs/libp2p/latest/libp2p/allow_block_list/struct.Behaviour.html#method.allow_peer

Only the peers added do this function are allowed to establish connections (inbound & outbound).

crates/services/p2p/src/peer_manager.rs
```diff
@@ -132,27 +132,18 @@ impl PeerManager {
     pub fn handle_peer_connected(
         &mut self,
         peer_id: &PeerId,
-        addresses: Vec<Multiaddr>,
         initial_connection: bool,
     ) -> bool {
         if initial_connection {
-            self.handle_initial_connection(peer_id, addresses)
+            self.handle_initial_connection(peer_id)
         } else {
```
MitchTurner, line 139: Should look into the `else` case. IDK if we want to remove this.

crates/services/p2p/src/peer_manager.rs
```diff
@@ -132,27 +132,18 @@ impl PeerManager {
...
-            let peers = self.get_assigned_peer_table_mut(peer_id);
-            insert_peer_addresses(peers, peer_id, addresses);
```
MitchTurner, line 142: Idk if we want to remove this. I'm not sure where we're gonna get the addresses though.

MitchTurner: I remembered that this is directly related to [this discussion](https://github.com/libp2p/rust-libp2p/discussions/4674) I had a couple of months ago.

MitchTurner: I  think I fixed it with some refactoring of `PeerReport` to move `Identify` and `Heartbeat` to `FuelBehavior` so we'd have access to the events, and with them the addresses, at the level we do this stuff ^^^.

It also cleans up the code a lot.

The UTs are passing again, so it should be good, but we can roll it back if needed.

crates/services/p2p/src/peer_report.rs
```diff
@@ -118,340 +96,107 @@ impl PeerReportBehaviour {
...
-                        self, action,
-                    );
-                if let Some(action) = action {
-                    return Poll::Ready(action)
-                }
-            }
-        }
-
-        loop {
-            // poll until we've either exhausted the events or found one of interest
```
MitchTurner, line 274: This is something that doesn't have parity with the changes. I just let the generated `poll` (from the `NetworkBehavior` macro) do its thing. That might be a problem.

We could manually implement `poll` for `FuelBehavior` to do the same thing as this, but that might mean we have to implement _all_ the methods, and some of them are hairy with the nested `Selects` etc.

crates/services/p2p/src/behavior.rs
```diff
@@ -21,37 +21,39 @@ use crate::{
...
     PeerReport(PeerReportEvent),
     Gossipsub(GossipsubEvent),
     RequestResponse(RequestResponseEvent<RequestMessage, NetworkResponse>),
+    BlockedPeers(void::Void),
 }
 
 /// Handles all p2p protocols needed for Fuel.
 #[derive(NetworkBehaviour)]
-#[behaviour(out_event = "FuelBehaviourEvent")]
+#[behaviour(to_swarm = "FuelBehaviourEvent")]
```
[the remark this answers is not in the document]
MitchTurner: https://github.com/FuelLabs/fuel-core/issues/1555

crates/services/p2p/src/discovery.rs
```diff
@@ -58,7 +59,7 @@ pub struct DiscoveryBehaviour {
     mdns: MdnsWrapper,
 
     /// Kademlia with MemoryStore
-    kademlia: Kademlia<MemoryStore>,
+    kademlia: KademliaBehavior<MemoryStore>,
```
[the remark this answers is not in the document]
MitchTurner: https://github.com/FuelLabs/fuel-core/issues/1556

crates/services/p2p/src/heartbeat/handler.rs
```diff
@@ -192,12 +148,13 @@ impl ConnectionHandler for HeartbeatHandler {
...
-                    HeartbeatFailure::Timeout,
-                ))
-            }
+            // TODO: Close connection properly: https://github.com/libp2p/rust-libp2p/pull/4755
+            // if self.failure_count >= self.config.max_failures.into() {
+            //     // Request from `Swarm` to close the faulty connection
+            //     return Poll::Ready(ConnectionHandlerEvent::Close(
+            //         HeartbeatFailure::Timeout,
+            //     ))
+            // }
```
MitchTurner, line 157: Do we need this?

crates/services/p2p/src/behavior.rs
```diff
@@ -98,21 +115,42 @@ impl<Codec: NetworkCodec> FuelBehaviour<Codec> {
 
         let peer_report = PeerReportBehaviour::new(p2p_config);
 
+        let identify = {
+            let identify_config = identify::Config::new(
+                "/fuel/1.0".to_string(),
```
Voxelot, line 120: Would we want to make a todo to make this configurable?
i.e something like
```suggestion
                format!("/fuel/{chain_id}/{p2p_protocol_version}", chain_id, P2P_VERSION),
```

crates/services/p2p/src/codecs/postcard.rs
```diff
@@ -85,28 +82,36 @@ impl RequestResponseCodec for PostcardCodec {
...
     {
-        let encoded_data = read_length_prefixed(socket, MAX_REQUEST_SIZE).await?;
+        let mut response = Vec::new();
+        socket
+            .take(self.max_response_size as u64)
+            .read_to_end(&mut response)
+            .await?;
 
-        self.deserialize(&encoded_data)
+        self.deserialize(&response)
```
Voxelot, line 97: todo: we may want to look into streaming deserialization at some point instead of buffering the whole message into a vec first.

crates/services/p2p/src/codecs/postcard.rs
```diff
@@ -14,25 +14,17 @@ use crate::{
         OutboundResponse,
         RequestMessage,
         ResponseMessage,
-        MAX_REQUEST_SIZE,
+        // MAX_REQUEST_SIZE,
```
Voxelot, line 17:
```suggestion
```

crates/services/p2p/src/config/fuel_authenticated.rs
```diff
@@ -84,24 +86,19 @@ where
...
     type Output = (PeerId, NoiseOutput<T>);
     type Error = NoiseError;
     type Future = Pin<Box<dyn Future<Output = Result<Self::Output, Self::Error>> + Send>>;
 
-    fn upgrade_outbound(self, socket: T, info: Self::Info) -> Self::Future {
+    fn upgrade_outbound(self, socket: T, _: Self::Info) -> Self::Future {
         Box::pin(
             self.noise_authenticated
-                .upgrade_outbound(socket, info)
+                .upgrade_outbound(socket, "")
```
Voxelot, line 101: What's the reason for using a blank string here?

crates/services/p2p/src/discovery.rs
```diff
@@ -187,159 +233,102 @@ impl NetworkBehaviour for DiscoveryBehaviour {
...
+        bootstrap_nodes: Vec<Multiaddr>,
+    ) -> impl FnOnce(Keypair) -> DiscoveryBehaviour {
+        |keypair| {
             let mut config = DiscoveryConfig::new(
                 keypair.public().to_peer_id(),
                 "test_network".into(),
             );
             config
-                .discovery_limit(50)
+                .max_peers_connected(50)
```
Voxelot, line 279: Maybe we should move this to a const?

```suggestion
                .max_peers_connected(MAX_PEERS)
```

crates/services/p2p/src/discovery.rs
```diff
@@ -187,159 +233,102 @@ impl NetworkBehaviour for DiscoveryBehaviour {
...
+        (swarm, listen_addr, peer_id)
     }
 
     // builds 25 discovery swarms,
     // initially, only connects first_swarm to the rest of the swarms
     // after that each swarm uses kademlia to discover other swarms
     // test completes after all swarms have connected to each other
+    // TODO: This used to fail with any connection closures, but that was causing a lot of failed tests.
+    //   Now it allows for many connection closures before failing. We don't know what caused the
+    //   connections to start failing, but had something to do with upgrading `libp2p`.
```
Voxelot, line 312: Do we have a ticket to keep track of this issue?

MitchTurner: Not yet. I can create one for investigation. I mostly left this as a note to myself and reviewers to have a discussion about the change in behavior.

crates/services/p2p/src/codecs/postcard.rs
```diff
@@ -85,28 +82,36 @@ impl RequestResponseCodec for PostcardCodec {
...
+        _: &Self::Protocol,
         socket: &mut T,
     ) -> io::Result<Self::Request>
     where
         T: AsyncRead + Unpin + Send,
     {
-        let encoded_data = read_length_prefixed(socket, MAX_REQUEST_SIZE).await?;
+        let mut response = Vec::new();
+        socket
+            .take(self.max_response_size as u64)
```
Voxelot, line 93: Is there a way to improve the error message reporting if `max_response_size` is exceeded? Otherwise it will just show up as an opaque deserialization error.

MitchTurner: It should be caught by this trace:
```
tracing::error!("RequestResponse outbound error for peer: {:?} with id: {:?} and error: {:?}", peer, request_id, error);
```

crates/services/p2p/src/discovery.rs
MitchTurner: No. I've tried re-running since we updated the codec and everything. It is still flaky.

crates/services/p2p/src/p2p_service.rs
```diff
@@ -155,62 +159,41 @@ pub enum FuelP2PEvent {
...
-                // If this is a guarded node,
-                // it should not receive any incoming connection requests.
-                // Rather, it will send outgoing connection requests to its reserved nodes
-                0
-            } else {
-                total_connections / 2
-            }
-        };
-
-        let connection_limits = ConnectionLimits::default()
```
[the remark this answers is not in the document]
MitchTurner: Good catch. We should probably use the connection_limit behavior:
https://github.com/libp2p/rust-libp2p/blob/master/misc/connection-limits/src/lib.rs

Since this isn't covered tests, I'll create a new issue for this that includes tests for proving desired behavior.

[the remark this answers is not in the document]
MitchTurner: https://github.com/FuelLabs/fuel-core/issues/1560

crates/services/p2p/src/peer_report.rs
```diff
@@ -83,32 +58,15 @@ pub enum PeerReportEvent {
 
 // `Behaviour` that reports events about peers
 pub struct PeerReportBehaviour {
```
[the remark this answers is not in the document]
MitchTurner: https://github.com/FuelLabs/fuel-core/issues/1561

crates/services/p2p/src/discovery.rs
MitchTurner: https://github.com/FuelLabs/fuel-core/issues/1562

crates/services/p2p/src/heartbeat.rs
```diff
@@ -79,19 +84,54 @@ impl Heartbeat {
...
+        _connection_id: ConnectionId,
+        _peer: PeerId,
+        _addr: &Multiaddr,
+        _role_override: Endpoint,
+    ) -> Result<THandler<Self>, ConnectionDenied> {
+        Ok(HeartbeatHandler::new(self.config.clone()))
+    }
+
+    fn on_swarm_event(&mut self, event: FromSwarm) {
+        match event {
```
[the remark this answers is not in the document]
MitchTurner: I just simplified it to `()` for simplicity.

crates/services/p2p/src/codecs/postcard.rs
MitchTurner: https://github.com/FuelLabs/fuel-core/issues/1563

crates/services/p2p/src/behavior.rs
MitchTurner: https://github.com/FuelLabs/fuel-core/issues/1564

crates/services/p2p/src/heartbeat/handler.rs
MitchTurner: Hmm. As it currently works, an unresponsive peer's reputation is going to decay naturally over time. If it runs out it will be banned, not just disconnected.

Would we prefer an unresponsive peer not be banned, like you're saying?

xgreenx: Decay doesn't increase the negative score of the peer. It only decreases the value. So it can't cause a disconnection from the peer.

crates/fuel-core/src/p2p_test_helpers.rs
```diff
@@ -378,11 +378,11 @@ pub fn make_config(name: String, mut node_config: Config) -> Config {
 pub async fn make_node(node_config: Config, test_txs: Vec<Transaction>) -> Node {
     let db = Database::in_memory();
     let node = tokio::time::timeout(
-        Duration::from_secs(1),
+        Duration::from_secs(2),
         FuelService::from_database(db.clone(), node_config),
     )
     .await
-    .expect("All services should start in less than 1 second")
+    .expect("All services should start in less than 2 second")
```
Dentosal, line 385:
```suggestion
    .expect("All services should start in less than 2 seconds")
```

crates/services/p2p/src/heartbeat/handler.rs
MitchTurner: I guess you're right since `MIN_APP_SCORE` is negative.

MitchTurner: > I think it is still better to close/restart the connection if the peer returns nothing to us for a while. But not block him=)

You mean you like this behavior?

The problem, and sorry I didn't clarify this in the original comment, is that there is no longer a `Close` variant of the `ConnectionHandlerEvent`. So I commented it out. I'll create an issue as this behavior isn't captured in any tests.

https://github.com/FuelLabs/fuel-core/issues/1567

xgreenx (approved):

thomaseizinger:
Exciting to see this merged 🎉

MitchTurner:
Thanks for all your help!

PR #1391, opened 2023-10-01
Remove `Result` from streams and use `Batch::is_err`

xgreenx:
Simplification after the review process

### DIFF
--- a/crates/services/sync/src/import.rs
+++ b/crates/services/sync/src/import.rs
@@ -2,7 +2,6 @@
 //! This module contains the import task which is responsible for
 //! importing blocks from the network into the local blockchain.
 
-use anyhow::anyhow;
 use fuel_core_services::{
     SharedMutex,
     StateWatcher,
@@ -25,7 +24,6 @@ use futures::{
     stream::StreamExt,
     FutureExt,
     Stream,
-    TryStreamExt,
 };
 use std::{
     future::Future,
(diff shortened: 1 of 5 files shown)


crates/services/sync/src/import.rs
```diff
@@ -423,34 +383,27 @@ fn check_sealed_header<
...
-) -> Result<(), ImportError> {
-    consensus
+    consensus: &Arc<C>,
+) {
+    let _ = consensus
         .await_da_height(&header.entity.da_height)
         .await
-        .map_err(ImportError::ConsensusError)?;
-    Ok(())
+        .trace_err("Failed to wait for DA layer to sync");
```
bvrooman, line 406: It looks like you're changing the function signature to be infallible by ignoring any internal error. Is this safe? What does it mean if `consensus.await_da_height(..)` fails for some reason, but the import continues?

crates/services/sync/src/import.rs
```diff
@@ -623,16 +557,13 @@ async fn execute_and_commit<E>(
     executor: &E,
     state: &SharedMutex<State>,
     block: SealedBlock,
-) -> Result<(), ImportError>
+) -> anyhow::Result<()>
```
bvrooman, line 560: I think it would be nice to move away from using `anyhow` inside internal implementations, but we don't have to solve that right now.

bvrooman (approved):
Thanks for the simplifications! LGTM as long as the logic remains correct.

crates/services/sync/src/import.rs
xgreenx: Yes, we continue to import blocks. If `consensus.await_da_height` fails, it may fail the execution of some blocks with new messages.

But it is not a huge problem. The block producer's public key is defined in the Genesis block, so awaiting the right DA block height is not critical for us if the block producer signed this block.

Issue #1334, opened 2023-08-29
TOB-FUEL-41: Message proof generation does not validate that block is ahead of genesis


xgreenx:
## Description
The message_proof implementation does not validate that the committed block is greater than the genesis block (zero) prior to subtracting one, resulting in an underflow. While the proof cannot be generated since the committed block is not in the database, this input should be handled with an error as it is elsewhere.

#### Figure 41.1: Potential overflow of block height (audit-fuel/fuel-core/crates/fuel-core/src/query/message.rs#201)

```rust
let verifiable_commit_block_height = *commit_block_header.height() - 1u32.into();
```

The following test case can be modified to observe the overflow. Note, this test uses a
mocked database but in production that

#### Figure 41.2: Modifications to existing test to demonstrate overflow

```bash
diff --git a/fuel-core/crates/fuel-core/src/query/message/test.rs
b/fuel-core/crates/fuel-core/src/query/message/test.rs
index c351b7c00..662aba5ff 100644
--- a/fuel-core/crates/fuel-core/src/query/message/test.rs
+++ b/fuel-core/crates/fuel-core/src/query/message/test.rs
@@ -83,8 +83,8 @@ mockall::mock! {
 #[tokio::test]
 async fn can_build_message_proof() {
     use mockall::predicate::*;
-    let commit_block_height = BlockHeight::from(2u32);
-    let message_block_height = BlockHeight::from(1u32);
+    let commit_block_height = BlockHeight::from(0u32);
+    let message_block_height = BlockHeight::from(0u32);
     let expected_receipt = receipt(Some(11));
     let message_id = expected_receipt.message_id().unwrap();
     let receipts: [Receipt; 4] = [
```

#### Figure 41.3: Implementation of the Sub trait with wrapping semantics (fuel-vm/fuel-types/src/numeric_types.rs#233–240)

```rust
impl Sub for $i {
    type Output = $i;
    #[inline(always)]
    fn sub(self, rhs: $i) -> $i {
        $i(self.0.wrapping_sub(rhs.0))
    } 
}
```

Furthermore, the use of the wrapping_sub and wrabbing_add in the implementation of a Sub and Add trait, respectively, is error-prone as it prevents rustc debug builds from panicking on overflow. This behavior is surprising and masked by being implemented in a macro. It would be ideal to explicitly use wrapping_add as opposed to overriding the semantics of the binary arithmetic operators in debug builds.

## Exploit Scenario
[...]

Issue #816, opened 2022-12-05
Design and implement the flow for `Message`s in the genesis block


xgreenx:
The genesis block can be filled with `Message`s, but because there is no signer and no known consensus algorithm, it is unclear how to prove and bridge messages back to the settlement layer.

This issue is partially related to the fork upgrade strategy of the Fuel chain. We need to design the flow of upgradability and data transfer from the old network to a new one or remove the initial state from the `ChainConfig`.

Voxelot:
Yeah the main issue at hand is that we currently need to post the entire block to Ethereum to prove the message really exists and was signed by consensus. We might need to update the bridging protocol to have a separate withdrawal function for genesis messages, where it authenticates the message root against the genesis message root configured on the contract during initialization, and doesn't require posting a full block.

xgreenx:
During the work on https://github.com/FuelLabs/fuel-core/pull/1625 `Message`s became part of the on-chain database, and the block importer has exclusive write ownership in the `Messages` table. 

This means that messages became part of the blockchain state when we increased the DA block height and unrevertalbe unless the block was created incorrectly(it is provable via fraud proofs mechanism).

Because messages are part of the blockchain state they also should migrate to the forked network. The https://github.com/FuelLabs/fuel-core/pull/1693 performs the migration of coins, messages, and contracts. So the issue for L1 -> L2 messages migration is solved.

Regarding the L2 -> L1 case: The L2 block already contains all the required information to withdraw the money on L1. For the different versions of the block header, the L1 contract will support different API methods. Transactions and receipts are versionable by default because they are enums, so if something changes for the `SMO` opcode, it will be reflected in a new enum variant. The proof generation to withdraw tokens from L2 uses off-chain data (receipts and transactions), so after [migration of the off-chain database](https://github.com/FuelLabs/fuel-core/issues/16190, we will also support L2 -> L1 messages migration.

I'm closing this issue since there are sub-issues that resolve L2 -> L1 messages migration.

PR #1392, opened 2023-10-02
Fix an interger overflow in message_proof when looking beyond genesis

Dentosal:
Fixes https://github.com/FuelLabs/fuel-core/issues/1334

See https://github.com/FuelLabs/fuel-vm/pull/596 for follow-up work.

### DIFF
--- a/crates/fuel-core/src/query/message.rs
+++ b/crates/fuel-core/src/query/message.rs
@@ -205,8 +205,8 @@ pub fn message_proof<T: MessageProofData + ?Sized>(
         None => return Ok(None),
     };
 
-    let block_heigth = *commit_block_header.height();
-    if block_heigth == 0u32.into() {
+    let block_height = *commit_block_header.height();
+    if block_height == 0u32.into() {
         // Cannot look beyond the genesis block
         return Ok(None)
     }
--- a/crates/fuel-core/src/query/message.rs
+++ b/crates/fuel-core/src/query/message.rs
@@ -205,7 +205,11 @@ pub fn message_proof<T: MessageProofData + ?Sized>(
         None => return Ok(None),
     };
 
-    let verifiable_commit_block_height = *commit_block_header.height() - 1u32.into();
+    let block_heigth = *commit_block_header.height();
+    if block_heigth == 0u32.into() { // Cannot look beyond the genesis block
+        return Ok(None)
+    }
+    let verifiable_commit_block_height = block_heigth - 1u32.into();
     let block_proof = database.block_history_proof(
         message_block_header.height(),
         &verifiable_commit_block_height,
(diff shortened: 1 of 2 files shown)


crates/fuel-core/src/query/message.rs
```diff
@@ -205,7 +205,12 @@ pub fn message_proof<T: MessageProofData + ?Sized>(
         None => return Ok(None),
     };
 
-    let verifiable_commit_block_height = *commit_block_header.height() - 1u32.into();
+    let block_heigth = *commit_block_header.height();
+    if block_heigth == 0u32.into() {
```
xgreenx, line 209: Hmm, actually, you should be able to use Genesis Block to generate proofs, but because it is a specific use case, we need to define how we want to do that and what the proof should look like. 

It is not related to this PR and should be solved during https://github.com/FuelLabs/fuel-core/issues/816

crates/fuel-core/src/query/message.rs
```diff
@@ -205,7 +205,12 @@ pub fn message_proof<T: MessageProofData + ?Sized>(
...
+        // Cannot look beyond the genesis block
+        return Ok(None)
```
xgreenx, line 211: Maybe it is better to return an error?=)

Dentosal: I understood that `None` means "cannot generate proof" here from the usage above. Is this not the case? It doesn't help that the function docs say this without explaining the meaning:

```rust
```

xgreenx: Yeah, I left this question=D I don't know why we want to return `None` instead of returning the reason WHY we can't generate the proof.

xgreenx (approved):

crates/fuel-core/src/query/message.rs
bvrooman, line 209:
```suggestion
    let block_height = *commit_block_header.height();
    if block_height == 0u32.into() {
```

xgreenx (approved):

crates/fuel-core/src/query/message.rs
```diff
@@ -205,7 +205,12 @@ pub fn message_proof<T: MessageProofData + ?Sized>(
         None => return Ok(None),
     };
 
-    let verifiable_commit_block_height = *commit_block_header.height() - 1u32.into();
+    let block_height = *commit_block_header.height();
+    if block_height == 0u32.into() {
+        // Cannot look beyond the genesis block
+        return Ok(None)
+    }
+    let verifiable_commit_block_height = block_heigth - 1u32.into();
```
xgreenx, line 213:
```suggestion
    let verifiable_commit_block_height = block_height - 1u32.into();
```

xgreenx (approved):

Issue #1330, opened 2023-08-28
TOB-FUEL-28: Low timeout for heartbeat protocol upgrade


xgreenx:
## Description
The substream protocol for the heartbeat protocol has a rather low timeout of 2 seconds by default. If a peer hits the timeout, then it gets disconnected. A timeout of 2 seconds could be reached during normal operation, or it could be forced by an attacker leading to a peer disconnecting from the P2P network. The following figure shows the code which configures the timeout.

#### Figure 28.1: Timeout configuration for the heartbeat subprotocol. (fuel-core/crates/services/p2p/src/heartbeat/handler.rs#261–268)

```rust
// Request new stream
self.outbound = Some(OutboundState::NegotiatingStream);
let protocol =
    SubstreamProtocol::new(ReadyUpgrade::new(HEARTBEAT_PROTOCOL), ())
        .with_timeout(self.config.send_timeout);
return Poll::Ready(ConnectionHandlerEvent::OutboundSubstreamRequest {
protocol, })
```

The next figure shows the default configuration for the timeout.

#### Figure 28.2: Default configuration for heartbeat timeouts. (fuel-core/crates/services/p2p/src/heartbeat/handler.rs#73–81)

```rust
impl Default for HeartbeatConfig {
    fn default() -> Self {
        Self::new(
            Duration::from_secs(2),
            Duration::from_secs(1),
            NonZeroU32::new(5).expect("5 != 0"),
) }
}
```

## Exploit Scenario
An attacker causes high CPU usage on a server that runs a reserved and honest fuel-node. Because of the high load, the responsiveness of the server is reduced. Ultimately, the node hits the default 2 seconds timeout and is disconnected.

## Recommendations
Short term, increase the timeout to a value that is unlikely to be hit during normal operation like 10 seconds or 1 minute.
Long term, consider introducing a separate timeout configuration for timeouts. This would simplify debugging fuel-core using a debugger utilizing breakpoints.

xgreenx:
With https://github.com/FuelLabs/fuel-core/issues/1381 we can decrease the frequency of the heartbeat protocol and significantly increase timeout.

PR #1393, opened 2023-10-02
Increase heartbeat timeout to one minute

Dentosal:
Closes https://github.com/FuelLabs/fuel-core/issues/1330

### DIFF
--- a/crates/services/p2p/src/heartbeat/handler.rs
+++ b/crates/services/p2p/src/heartbeat/handler.rs
@@ -73,7 +73,7 @@ impl HeartbeatConfig {
 impl Default for HeartbeatConfig {
     fn default() -> Self {
         Self::new(
-            Duration::from_secs(2),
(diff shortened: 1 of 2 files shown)


crates/services/p2p/src/heartbeat/handler.rs
```diff
@@ -73,7 +73,7 @@ impl HeartbeatConfig {
 impl Default for HeartbeatConfig {
     fn default() -> Self {
         Self::new(
-            Duration::from_secs(2),
+            Duration::from_secs(60),
             Duration::from_secs(1),
             NonZeroU32::new(5).expect("5 != 0"),
```
xgreenx, line 78: It would be nice to make it configurable, but I'm okay with not doing that right now because we already have a lot of configuration for p2p

xgreenx (approved):

PR #1397, opened 2023-10-04
Improve keygen

cr-fuel:
Make the keygen a crate that can be implemented and extended from other places (like a forc-plugin). Upgraded its dependencies to avoid depending on protoc for building

### DIFF
--- a/bin/keygen/src/keygen.rs
+++ b/bin/keygen/src/keygen.rs
@@ -40,7 +40,6 @@ pub struct NewKey {
 }
 
 #[derive(Clone, Debug, Default, ValueEnum)]
-#[clap(rename_all = "snake_case")]
 pub enum KeyType {
     #[default]
     BlockProduction,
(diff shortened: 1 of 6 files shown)


xgreenx (approved):
Could you update changeling please?=)

bin/keygen/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+//! Keygen crate
+
+pub const BLOCK_PRODUCTION: &str = "block_production";
```
xgreenx, line 3:
```suggestion
pub const BLOCK_PRODUCTION: &str = "block-production";
```

We use dashes in our CLI=)

bin/keygen/src/keygen.rs
```diff
@@ -25,46 +24,31 @@ use std::{
...
         value_enum,
-        default_value = "block-production"
+        default_value = BLOCK_PRODUCTION,
     )]
     key_type: KeyType,
 }
 
 #[derive(Clone, Debug, Default, ValueEnum)]
-pub(crate) enum KeyType {
+#[clap(rename_all = "snake_case")]
```
xgreenx, line 43: Why do you want to use snake case here?

xgreenx (approved):

Issue #1336, opened 2023-08-29
TOB-FUEL-43: Relayer should check for finalized blocks instead of block confirmations


xgreenx:
## Description

The Fuel relayer synchronizes the Fuel L2 with Ethereum mainnet by processing deposit transactions on Ethereum once the block containing a transaction has received sufficient confirmations, currently at least 100. This enables users to receive their deposited asset on the Fuel L2 and potentially withdraw it.

This mechanism and its configuration are not safe as Ethereum mainnet can experience periods of non-finality and block confirmations do not have any bearing on whether a transaction is final. In May 2023 the beacon chain failed to finalize for 9 epochs which corresponds to 149 blocks, indicating that Fuel’s configuration may process transasctions that are subsequently reorged and allow double spending deposits.

Instead, the Fuel relayer should only make logs available to the block producer which are finalized; that is, the block is justified (it has received attestations from two-thirds validators) and is 1 epoch behind a block is also justified. Double spends would require burning of staked Ether, which is prohibitively expensive.

#### Figure 43.1: the finality criteria defined in the [Fuel Relayer Specification](https://github.com/FuelLabs/fuel-core/tree/v0.19.0/crates/services/relayer#validity-finality-and-synchronization)
> Ethereum blocks are considered final after two epochs. Each epoch contains 32 slots. A slot takes 12s so epoch is around 6.4min and two epochs will take 12.8min. After 13min we are sure that our deposits will not be reverted by some big reorganization. So we are okay to say that everything older [sic] then ~100blocks is finalized.

#### Figure 43.2: the Relayer considers non-finalized blocks as final given sufficient confirmations (fuel-core/crates/services/relayer/src/service/state.rs#71–84)

```rust
impl EthHeights {
    /// Create a new Ethereum block height from the current
    /// block height and the desired finalization period.
    fn new(current: u64, finalization_period: u64) -> Self {
            Self(Heights(
                current.saturating_sub(finalization_period)..=current,
            ))
    }

    /// Get the finalized eth block height.
    fn finalized(&self) -> u64 {
        *self.0 .0.start()
    } 
}
```

## Exploit Scenario
The Beacon Chain fails to finalize as it did in May 2023 due to a client bug or network latency. Fuel processes deposit transactions of blocks that are produced but not finalized. Some deposit transactions are reorged from Ethereum mainnet, allowing attackers to deposit again and receive twice what they deposited.

## Recommendations
[...]

PR #1399, opened 2023-10-04
feat: Retrieve finalized blocks from Ethereum

bvrooman:
Related issues:
- Closes https://github.com/FuelLabs/fuel-core/issues/1336

This PR changes the approach a Fuel node uses to determine the finalized blocks on the Ethereum blockchain.

**The existing approach:**
- Assume a "finalization period" - a period of time after which we assume a block is finalized. Generally, we assume a block is finalized after 100 new blocks have been produced afterwards
- Observe what the current block height is as reported by the Ethereum client
- Subtract the finalization period from the current height to get an assumed finalized block, i.e. current block height - 100

The problem with this approach, as outlined by Trail of Bits, is that finality cannot be guaranteed simply by the amount of time passed alone. 

**This PR proposes a more robust approach:**
- Use the Ethereum client to query for the most recently finalized block directly. This is done using the `Finalized` tag in the query. See https://github.com/gakonst/ethers-rs/pull/1792

This approach removes any assumption on our side as to what the finalized block is, and gets the canonically accepted finalized block from Ethereum.

### DIFF
--- a/crates/services/relayer/src/service/state.rs
+++ b/crates/services/relayer/src/service/state.rs
@@ -3,7 +3,6 @@
 
 use core::ops::RangeInclusive;
 pub use state_builder::*;
-// use std::ops::Deref;
 
 mod state_builder;
 
--- a/crates/services/relayer/src/service/state.rs
+++ b/crates/services/relayer/src/service/state.rs
@@ -71,10 +71,8 @@ impl EthState {
 impl EthHeights {
     /// Create a new Ethereum block height from the current
     /// block height and the desired finalization period.
-    fn new(current: u64, finalization_period: u64) -> Self {
-        Self(Heights(
-            current.saturating_sub(finalization_period)..=current,
-        ))
+    fn new(current: u64, finalized: u64) -> Self {
+        Self(Heights(current.saturating_sub(finalized)..=current))
     }
 
     /// Get the finalized eth block height.
--- a/crates/services/relayer/src/service/state.rs
+++ b/crates/services/relayer/src/service/state.rs
@@ -3,7 +3,7 @@
 
 use core::ops::RangeInclusive;
 pub use state_builder::*;
-use std::ops::Deref;
+// use std::ops::Deref;
 
 mod state_builder;
 
@@ -14,7 +14,7 @@ mod test;
 /// The state of the Ethereum node.
 pub struct EthState {
     /// The state that the relayer thinks the remote Ethereum node is in.
-    remote: EthHeights,
+    remote: EthHeight,
     /// State related to the Ethereum node that is tracked by the relayer.
     local: Option<EthHeight>,
 }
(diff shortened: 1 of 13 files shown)


crates/services/relayer/src/service/state.rs
```diff
@@ -3,7 +3,7 @@
 
 use core::ops::RangeInclusive;
 pub use state_builder::*;
-use std::ops::Deref;
+// use std::ops::Deref;
```
Voxelot, line 6:
```suggestion
```

crates/services/relayer/src/service/state.rs
```diff
@@ -53,50 +49,34 @@ impl EthState {
...
 
     /// Get the gap between the relayer and the Ethereum node if
     /// a sync is required.
     pub fn needs_to_sync_eth(&self) -> Option<EthSyncGap> {
         (!self.is_synced()).then(|| {
-            EthSyncGap::new(
-                self.local.map(|l| l.saturating_add(1)).unwrap_or(0),
-                self.remote.finalized(),
-            )
+            let local = self.local.map(|l| l.saturating_add(1)).unwrap_or(0);
```
Voxelot, line 59: Does removing the use of an inclusive range affect the correctness of doing a saturating_add here? It seems like there's a potential that local could be higher than remote in this case.

bvrooman: I do not believe that this changes the correctness.  Prior to these changes, `EthSyncGap` used a single inclusive range to store its start and end values (`local` and `remote` respectively); with these changes, `EthSyncGap` stores its start and end values separately, and implements a conversion into an inclusive range where any range behaviour needs to be maintained. In effect, the correctness will be maintained.

At the point of instantiation, where we calculate the inputs to the `EthSyncGap` constructor, `local` remains unchanged, and `remote` is now the value returned by the Ethereum client. Before and after these changes, there is no guarantee that `local` was less than or equal to `remote`. This means that the `EthSyncGap` can be "empty", i.e., `local` is greater than or equal to `remote`. In fact, some tests check the behaviour when `local` is greater than `remote` to ensure that no sync is necessary: https://github.com/FuelLabs/fuel-core/blob/ec6a61277118d2b9e3deb9e29ae01f8129242ea8/crates/services/relayer/src/service/state/test.rs#L30

tests/tests/relayer.rs
```diff
@@ -309,6 +309,10 @@ async fn handle(
             let r = mock.get_block_number().await.unwrap();
             json!({ "id": id, "jsonrpc": "2.0", "result": r })
         }
+        "eth_getBlockByNumber" => {
```
xgreenx, line 312: I guess we can remove `eth_blockNumber` now

xgreenx (approved):

tests/tests/relayer.rs
bvrooman: Yep, that's true. If we need it again in the future for whatever reason, we can add it back.

xgreenx (approved):

Issue #1269, opened 2023-07-31
`Mint` transaction rework


xgreenx:
Rework the `Mint` transaction to not produce a new UTXO per each block. Instead, we can use contract's account-based model.

xgreenx:
This change is better to do on top of https://github.com/FuelLabs/fuel-vm/pull/533

PR #1407, opened 2023-10-09
Account-base fee collection

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1269

This PR also adapts `fuel-core` to use `fuel-vm 0.39`.
- The recipient is a `ContractId` instead of `Address`. The block producer should deploy its contract to receive the transaction fee. The collected fee is zero until the recipient contract is set.
- The `Mint` transaction is reworked with new fields to support the account-base model. It affects serialization and deserialization of the transaction and also affects GraphQL schema.
- I limited the number of selected transactions from the TxPool to `u16::MAX` to avoid overflow.
- The `Mint` transaction is the last transaction in the block instead of the first.
- Applied changes for `BlockHeight` to use `succ` and `pred` functions.


crates/types/src/services/executor.rs
```diff
@@ -247,18 +247,22 @@ impl ExecutionKind {
...
-    #[error("Not supported transaction: {0:?}")]
-    NotSupportedTransaction(TxId),
-    #[error("The first transaction in the block is not `Mint` - coinbase.")]
-    CoinbaseIsNotFirstTransaction,
-    #[error("Coinbase should have one output.")]
-    CoinbaseSeveralOutputs,
-    #[error("Coinbase outputs is invalid.")]
-    CoinbaseOutputIsInvalid,
+    #[error("The blocks is missing `Mint` transaction.")]
+    CoinbaseIsMissed,
```
bvrooman, line 257: Maybe 
```suggestion
    CoinbaseMissing,
```

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -207,7 +207,9 @@ where
     }
 
     fn next_height(&self) -> BlockHeight {
-        self.last_height + 1u32.into()
+        self.last_height
+            .succ()
+            .expect("It should be impossible to produce more blocks than u32::MAX")
```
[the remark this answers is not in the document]
xgreenx: After 132 years, we can overflow, so it is better to panic=D

crates/fuel-core/src/schema/tx/types.rs
```diff
@@ -309,6 +317,13 @@ impl Transaction {
         }
     }
 
+    async fn input_contract(&self) -> Option<input::InputContract> {
+        match &self.0 {
+            fuel_tx::Transaction::Script(_) | fuel_tx::Transaction::Create(_) => None,
+            fuel_tx::Transaction::Mint(mint) => Some(mint.input_contract().into()),
+        }
+    }
```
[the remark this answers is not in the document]
xgreenx: No, because it is getter for one field while `input_contracts` is a helper function

crates/fuel-core/src/executor.rs
```diff
@@ -691,101 +687,174 @@ where
...
+                execution_data,
+                tx_db_transaction,
+                execution_kind,
+                options,
+            ),
         }
     }
 
-    fn apply_coinbase(
+    fn execute_mint(
```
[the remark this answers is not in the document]
xgreenx: No, we still allow only one `Mint` transaction. Is what `CoinbaseSecondTransaction` error is about=) I renamed it into `MintFoundSecondEntry`

crates/fuel-core/src/executor.rs
```diff
@@ -691,101 +687,174 @@ where
...
-            .storage::<Transactions>()
-            .insert(&coinbase_id, &coinbase_tx.into())?
-            .is_some()
-        {
-            return Err(ExecutorError::TransactionIdCollision(coinbase_id))
+        execution_kind: ExecutionKind,
+        options: ExecutionOptions,
+    ) -> ExecutorResult<Transaction> {
+        if execution_data.found_mint {
+            return Err(ExecutorError::CoinbaseSecondTransaction)
```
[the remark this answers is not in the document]
bvrooman: Ah, I see now :)

crates/services/consensus_module/poa/src/service.rs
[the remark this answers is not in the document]
bvrooman: Yes, we can panic, but the panic could be inside `succ`. Right now, we always use this pattern:
```rust
Some(Self(self.0.checked_add(1)?)).expect("expect valid")
```
where we always add `expect` (or `unwrap`) - we never check the `Option`. 

We could just put the expect in  `succ` and achieve the same thing, but simpler.

But it's not a big deal - just an observation.

crates/fuel-core/src/executor.rs
```diff
@@ -779,7 +779,7 @@ where
                 *mint.mint_amount(),
             )
             .map_err(|e| anyhow::anyhow!(e))
-            .map_err(ExecutorError::CoinbaseCantIncreaseBalance)?;
+            .map_err(ExecutorError::CoinbaseCannotIncreaseBalance)?;
```
bvrooman, line 782: It looks like you're replacing `Coinbase` with `Mint` - should this be `MintCannotIncreaseBalance`?

xgreenx: Nope, because here the functionality of the coinable works incorrectly, instead of the problem with `Mint` transaction validity=)

crates/client/src/client/schema/tx/transparent_tx.rs
```diff
@@ -81,12 +83,16 @@ pub struct Transaction {
     pub tx_pointer: Option<TxPointer>,
     pub input_asset_ids: Option<Vec<AssetId>>,
     pub input_contracts: Option<Vec<ContractIdFragment>>,
+    pub input_contract: Option<InputContract>,
```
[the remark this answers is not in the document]
bvrooman: Any update here?

bvrooman (approved):
Just one remaining question about adding comments about the `input_contracts` versus `input_contract` and `output_contracts` versus `input_contract`.  It's up to you if you would like to add comments or change anything there.

crates/client/src/client/schema/tx/transparent_tx.rs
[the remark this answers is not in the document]
xgreenx: I've added the comments into the code itself=)

crates/fuel-core/src/executor.rs
```diff
@@ -528,93 +547,60 @@ where
...
-        };
-
-        let mut tx_index = if execution_kind != ExecutionKind::DryRun {
-            // Skip the coinbase transaction.
-            block.transactions.push(coinbase_tx.clone().into());
-            1
-        } else {
-            0
-        };
+        let mut execute_transaction = |execution_data: &mut ExecutionData,
```
Voxelot, line 550: It seems like it'd be cleaner to move the logic in this closure to the execute_transaction fn. Is there any reason why we need a closure for execute_transaction that wraps calls to the actual `fn execute_transction()`?

crates/fuel-core/src/executor.rs
```diff
@@ -625,36 +611,47 @@ where
...
+                },
+                amount_to_mint,
+                self.config.consensus_parameters.base_asset_id,
+            );
+
+            execute_transaction(
                 execution_data,
-                block_db_transaction,
+                MaybeCheckedTransaction::Transaction(coinbase_tx.into()),
             )?;
```
Voxelot, line 642: Should the tx_count include the mint tx? It may be confusing later on down the road if the count gets used for something else and isn't accurate.

crates/fuel-core/src/executor.rs
```diff
@@ -2198,48 +2369,36 @@ mod tests {
...
-            ));
-        }
-
-        #[test]
-        fn invalidate_block_height() {
-            let mint =
-                Transaction::mint(TxPointer::new(1.into(), Default::default()), vec![]);
+        fn invalidate_is_not_last() {
+            let mint = Transaction::mint(
+                TxPointer::new(Default::default(), 1),
```
Voxelot, line 2392: Hmm I don't think the way we check that the mint is the last tx is very robust. If I just change the tx_index to 0 here, I don't get the `MintIsNotLastTransaction` error.

crates/fuel-core/src/executor.rs
```diff
@@ -1992,24 +2069,118 @@ mod tests {
 
             assert_eq!(skipped_transactions.len(), 1);
             assert_eq!(block.transactions().len(), 2);
```
Voxelot, line 2071: if the second tx was skipped, shouldn't block.transactions().len() == 1?

crates/fuel-core/src/executor.rs
xgreenx: There are two main reason: 

1) We want to get a result from transaction execution and based on the result, decide what we need to do with it(skip or error, commit or not):
image

2) It pushes the transaction to the block and increases the tx counter:
image

crates/fuel-core/src/executor.rs
xgreenx: Yeah, because it should be `MintHasUnexpectedIndex` error=) I added proper support for `MintIsNotLastTransaction`

Voxelot (approved):

Cargo.toml
```diff
@@ -113,3 +113,6 @@ itertools = "0.10"
 insta = "1.8"
 tempfile = "3.4"
 tikv-jemallocator = "0.5"
+
+[patch.crates-io]
+fuel-vm = { git = "https://github.com/FuelLabs/fuel-vm/", branch = "master" }
```
MitchTurner, line 118: I assume this is blocked on the upgrade?

crates/client/src/client/schema/tx/transparent_tx.rs
```diff
@@ -80,13 +82,21 @@ pub struct Transaction {
     pub id: TransactionId,
     pub tx_pointer: Option<TxPointer>,
     pub input_asset_ids: Option<Vec<AssetId>>,
+    /// The list of all contracts from the inputs of the transaction.
+    ///
+    /// It is a helper function and doesn't have a corresponding field in the transaction.
```
MitchTurner, line 87: What is the "helper function"? This is a field on a struct.

crates/client/src/client/schema/tx/transparent_tx.rs
```diff
@@ -80,13 +82,21 @@ pub struct Transaction {
...
     pub input_contracts: Option<Vec<ContractIdFragment>>,
+    /// The field of the `Mint` transaction.
```
MitchTurner, line 89: the `input_contract` is a field of the `Mint` transaction?

crates/fuel-core/src/executor.rs
```diff
@@ -625,36 +611,47 @@ where
...
-                coinbase_tx,
-                Some(execution_data.coinbase),
-            )?;
-            self.apply_coinbase(
-                coinbase_tx,
-                block,
+            let coinbase_tx = Transaction::mint(
+                TxPointer::new(block_height, execution_data.tx_count),
+                input::contract::Contract {
+                    utxo_id: Default::default(),
```
MitchTurner, line 624: It seems weird to be using `Default` for these values. If the values are static, maybe have the constructor include them under the hood? But I have a feeling we should be explicit about what these values are so if the defaults change from under us we aren't surprised, etc.

crates/fuel-core/src/executor.rs
```diff
@@ -691,101 +691,171 @@ where
...
+        let (mut mint, _) = checked_mint.into();
 
-        if let Some(Output::Coin {
-            asset_id, amount, ..
-        }) = checked_mint.transaction().outputs().first()
-        {
-            if asset_id != self.config.consensus_parameters.base_asset_id() {
-                return Err(ExecutorError::CoinbaseOutputIsInvalid)
+        if mint.input_contract().contract_id != ContractId::zeroed() {
+            if *mint.mint_amount() != execution_data.coinbase {
```
MitchTurner, line 731: I think it would improve readability a lot to break down these functions into sub-functions. These functions are long.

crates/client/src/client/schema/tx/transparent_tx.rs
xgreenx: Yes

MitchTurner: Sorry. I didn't phrase this well enough. I don't understand what it means for `input_contracts` to be a field on `Mint`. Same with `output_contract`. It's a field on `Transaction`.

xgreenx: image

MitchTurner: Okay. They both have that field. They both have the `tx_pointer` field as well. I guess my point is when I'm reading through the `Transaction` code and I see the comment `The field of the `Mint` transaction.`, I'm not sure why this is important. If it is _only_ used by mints, then perhaps name it `mint_input_contract`, etc. If it serves multiple purposes, but it's important for the reader to understand that it is _also_ used on the `Mint` transaction, then maybe we could explain in the comment why it's important we take note of that.

The way it is written confused me, so I assume it will confuse someone else as well.

bvrooman: I agree that a name like `mint_input_contract` would be more clear. You could also add a comment that describes the relationship between transparent txs and mint and how the input/output contracts represent that relationship.

crates/client/src/client/schema/tx/transparent_tx.rs
MitchTurner: Gotcha. The function isn't public though, but this doc comment is public, so maybe mentioning the function isn't necessary as long as you say it's generated rather than part of the original `Transaction`.

crates/fuel-core/src/executor.rs
MitchTurner: This is an improvement. I don't know if I'm a fan of the function being in shared scope, it is about as cluttered as before. Also, having the "else" case be encapsulated as a function wouldn't hurt either, but I don't want to create too much extra work for you :). Clean Code suggests avoiding functions that are more than 20 lines long, that is an arbitrary number, but I tend to agree in spirit--breaking down functions makes a world of difference on cognitive load for a reader.

crates/client/src/client/schema/tx/transparent_tx.rs
MitchTurner: Maybe adding docs for the overall struct explaining why it's constructed of a collage of different transaction type parts would be good. Other than that, the docs you've added help a ton!

crates/fuel-core/src/executor.rs
xgreenx: I don't want to add one more function that accepts 8 arguments=)

MitchTurner: Okay. I'll leave it as a nit, however I think we shouldn't let something like that keep us from making our code more friendly and sympathetic to the next engineer.

MitchTurner (approved):

Issue #1370, opened 2023-09-18
Make most of lightweight crates `no-std` compatible


xgreenx:
Requires a new release of the `fuel-vm` with https://github.com/FuelLabs/fuel-vm/issues/536.

Make lightweight crates WASM compatible:
- [x] `fuel-core-types`
- [x] `fuel-core-storage`
- [x] `fuel-core-chain-config`
- [x] `fuel-core-client` without `reqwest` functionality

These crates(except `fuel-core-storage`) are used by the Rust SDK and are required for WASM compatibility with `fuel-indexer`.

PR #1411, opened 2023-10-10
feat: Support no_std and WASM compilation for `fuel-core` crates

bvrooman:
Related issues:
- Closes https://github.com/FuelLabs/fuel-core/issues/1370

Adds no_std compilation support for crates:
- `fuel-core-types`
- `fuel-core-storage`
- `fuel-core-client`
- `fuel-core-chain-config`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,10 +75,11 @@ fuel-core-tests = { version = "0.0.0", path = "./tests" }
 fuel-core-xtask = { version = "0.0.0", path = "./xtask" }
 
 # Fuel dependencies
-fuel-vm-private = { version = "0.38.0", package = "fuel-vm" }
+fuel-vm-private = { version = "0.38.0", package = "fuel-vm", default-features = false, features = ["alloc"] }
+#fuel-vm-private = { path = "../fuel-vm/fuel-vm", package = "fuel-vm", default-features = false, features = ["alloc"] }
 
 # Common dependencies
-anyhow = "1.0"
+anyhow = { version = "1.0", default-features = false }
 async-trait = "0.1"
 cynic = { version = "2.2.1", features = ["http-reqwest"] }
 clap = "4.1"
@@ -113,3 +114,6 @@ itertools = "0.10"
 insta = "1.8"
 tempfile = "3.4"
 tikv-jemallocator = "0.5"
+
+[patch.crates-io]
+fuel-vm-private = { git = "https://github.com/FuelLabs/fuel-vm/", package = "fuel-vm", branch = "bvrooman/feat/prepare-no-std-compatability" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -113,6 +113,3 @@ itertools = "0.10"
 insta = "1.8"
 tempfile = "3.4"
 tikv-jemallocator = "0.5"
-
-#[patch.crates-io]
-#fuel-vm = { git = "https://github.com/FuelLabs/fuel-vm/", branch = "bvrooman/feat/prepare-no-std-compatability" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,8 +75,8 @@ fuel-core-tests = { version = "0.0.0", path = "./tests" }
 fuel-core-xtask = { version = "0.0.0", path = "./xtask" }
 
 # Fuel dependencies
-#fuel-vm-private = { version = "0.38.0", package = "fuel-vm", default-features = false }
-fuel-vm-private = { path = "../fuel-vm-master/fuel-vm", package = "fuel-vm", default-features = false }
+fuel-vm-private = { version = "0.38.0", package = "fuel-vm", default-features = false }
+#fuel-vm-private = { path = "../fuel-vm-master/fuel-vm", package = "fuel-vm", default-features = false }
 
 # Common dependencies
 anyhow = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,7 +75,7 @@ fuel-core-tests = { version = "0.0.0", path = "./tests" }
 fuel-core-xtask = { version = "0.0.0", path = "./xtask" }
 
 # Fuel dependencies
-fuel-vm-private = { version = "0.38.0", package = "fuel-vm", default-features = false, features = ["alloc"] }
+fuel-vm-private = { version = "0.38.0", package = "fuel-vm", default-features = false }
 
 # Common dependencies
 anyhow = { version = "1.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -114,5 +114,5 @@ insta = "1.8"
 tempfile = "3.4"
 tikv-jemallocator = "0.5"
 
-[patch.crates-io]
-fuel-vm = { git = "https://github.com/FuelLabs/fuel-vm/", branch = "bvrooman/feat/prepare-no-std-compatability" }
+#[patch.crates-io]
+#fuel-vm = { git = "https://github.com/FuelLabs/fuel-vm/", branch = "bvrooman/feat/prepare-no-std-compatability" }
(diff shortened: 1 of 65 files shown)


crates/types/Cargo.toml
```diff
@@ -20,6 +20,7 @@ version = { workspace = true }
 anyhow = { workspace = true }
 derive_more = { version = "0.99" }
 fuel-vm-private = { workspace = true }
+getrandom = { version = "0.2", features = ["js"] }
```
bvrooman, line 23: See: https://docs.rs/getrandom/latest/getrandom/#webassembly-support

`getrandom` is a transitive dependency of `fuel-crypto`.

Cargo.toml
```diff
@@ -113,3 +113,6 @@ itertools = "0.10"
 insta = "1.8"
 tempfile = "3.4"
 tikv-jemallocator = "0.5"
+
+#[patch.crates-io]
+#fuel-vm = { git = "https://github.com/FuelLabs/fuel-vm/", branch = "bvrooman/feat/prepare-no-std-compatability" }
```
xgreenx, line 118:
```suggestion
```

benches/src/lib.rs
```diff
@@ -165,7 +162,8 @@ impl VmBench {
 
         let mut db = new_db();
 
-        db.deploy_contract_with_id(&salt, &[], &contract, &state_root, &id)?;
+        db.deploy_contract_with_id(&salt, &[], &contract, &state_root, &id)
```
xgreenx, line 165: It will be easier if you implement `impl From<fuel_core_storage::Error> for anyhow::Error`=) In this case, you don't need to update all functions related to `InterpreterStorage`

crates/chain-config/Cargo.toml
```diff
@@ -14,18 +14,20 @@ description = "Fuel Chain config types"
 anyhow = { workspace = true }
 bech32 = "0.9.0"
 fuel-core-storage = { workspace = true }
-fuel-core-types = { workspace = true, features = [
-    "serde",
-    "random",
-] }
+fuel-core-types = { workspace = true, default-features = false, features = ["random", "serde"] }
```
xgreenx, line 17: Why do we need `random` here?

crates/chain-config/src/lib.rs
```diff
@@ -11,8 +11,51 @@ pub use genesis::GenesisCommitment;
...
+//         const DEV_KEY_PHRASE: &str =
+//             "winner alley monkey elephant sun off boil hope toward boss bronze dish";
+//         let expected_key = SecretKey::new_from_mnemonic_phrase_with_path(
+//             DEV_KEY_PHRASE,
+//             "m/44'/60'/0'/0/0",
+//         )
+//         .unwrap();
+//         assert_eq!(key, expected_key);
+//     }
+// }
```
xgreenx, line 61:
```suggestion
```

crates/database/src/lib.rs
```diff
@@ -10,42 +10,41 @@
...
-impl From<Error> for std::io::Error {
-    fn from(e: Error) -> Self {
-        std::io::Error::new(ErrorKind::Other, e)
-    }
-}
+// impl From<Error> for std::io::Error {
+//     fn from(e: Error) -> Self {
+//         std::io::Error::new(ErrorKind::Other, e)
+//     }
+// }
```
xgreenx, line 47:
```suggestion
```

crates/fuel-core/src/database.rs
```diff
@@ -194,7 +194,7 @@ impl Database {
     #[cfg(feature = "rocksdb")]
     pub fn open(path: &Path, capacity: impl Into<Option<usize>>) -> DatabaseResult<Self> {
         use anyhow::Context;
-        let db = RocksDb::default_open(path, capacity.into()).context("Failed to open rocksdb, you may need to wipe a pre-existing incompatible db `rm -rf ~/.fuel/db`")?;
+        let db = RocksDb::default_open(path, capacity.into()).map_err(anyhow::Error::msg).context("Failed to open rocksdb, you may need to wipe a pre-existing incompatible db `rm -rf ~/.fuel/db`")?;
```
xgreenx, line 197: You can implement `From<DatabaseError>`

crates/chain-config/Cargo.toml
bvrooman: I will check to see what dependency/dependencies used it. If I can remove it, I will do so here.

bvrooman: It looks like `ChainConfig::local_testnet()` uses `SecretKey::random(&mut rng);`, and `SecretKey::random(..)` is behind the `random` feature. 
Instead of requiring `random`, I can add the `random` feature to `chain-config`. Then:
- we put `ChainConfig::local_testnet()` behind the `random` feature
- we only enable `fuel-core-types:random` when we use `chain-config:random`

crates/fuel-core/src/database.rs
bvrooman: Yes. I have now implemented `From` for the error classes. In cases where we use `context`, we still need to map to `anyhow::Error` because it is not implemented automatically for non-std errors.

crates/chain-config/Cargo.toml
bvrooman: After some digging, it looks like it is necessary to include `fuel-core-types:random` because `random` allows us to generate valid `SecretKey` values. Since we need to support `fuel-core-chain-config --no-default-features`, we cannot put `fuel-core-types:random` behind another feature flag: Creating keys without `random`, and using something like `SecretKey::default` instead, creates invalid keys in `local_testnet` and tests (maybe `default` should be removed, since default keys are invalid).  
The simplest approach is to keep the `random` feature in `fuel-core-types`.

benches/src/lib.rs
```diff
@@ -95,7 +92,7 @@ pub struct VmBench {
     pub prepare_call: Option<PrepareCall>,
     pub dummy_contract: Option<ContractId>,
     pub contract_code: Option<ContractCode>,
-    pub prepare_db: Option<Box<dyn FnMut(VmDatabase) -> io::Result<VmDatabase>>>,
+    pub prepare_db: Option<Box<dyn FnMut(VmDatabase) -> anyhow::Result<VmDatabase>>>,
```
MitchTurner, line 95: Did you end up deciding to use `anyhow` after all? I wasn't part of the conversation, but just for thoroughness's sake, I disagree with using `anyhow` in a lib.

crates/chain-config/Cargo.toml
bvrooman: As we discussed during our huddle, we hardcoded the `SecretKeys` in `local_testnet` (based on printing secrets to console using the existing seed) allowing us to remove the `random` feature flag. This removed our transitive dependency on `getrandom`, allowing us to successfully compile `fuel-core-chain-config` on WASM32. Lastly, I also added a new `random_testnet` function that depends on the `random` flag (thus not available in WASM32 land) if users want to have true random secret keys.

benches/src/lib.rs
MitchTurner: For example, did you consider something like this:
https://crates.io/crates/thiserror-no-std

bvrooman: At least for this PR, I'm not aiming to take a stand on the use of `anyhow` (pro or con) - I think that issue deserves its own conversation and follow up. For now, I'm simply removing the use of `std`, which includes `io` errors/results. For now, I am replacing these instances with `anyhow`, and we can later explore how best to do error handling as a practice. `thiserror-no-std` warrants a look though!

MitchTurner: Okay. Understood. I'm worried that it won't get prioritized, but I don't want to hold up this work because of that.

MitchTurner (approved):

xgreenx (approved):

Issue #1396, opened 2023-10-04
Determine file encoding for initial state


segfault-magnet:
Current development on the regenesis feature has split off the initial state from the chain config. This buys us some flexibility to select an appropriate encoding for the initial state file. 

Some points to consider:
* Storage -- how much overhead does the encoding have?
* Cursor -- how quickly can we continue decoding the file when resuming the regenesis?
* Performance -- both for decoding and encoding
* Compression -- encoding overhead can be reduced by compression. On the other hand, having a cursor to the last loaded element might get complicated. If compression is used ideally it shouldn't have system dependencies and be a well-known algorithm.
* Human readability -- keeping it human readable helps with manipulating the file, and developing tools to analyze it, testing.
* Library maturity -- general support for encoding/decoding and also rust-specific support.
* Streaming - can you decode/encode on the fly? What is the API like? Do you handle reading/saving the data?

segfault-magnet:
An update:

Wrote a small [utility](https://github.com/segfault-magnet/regenesis_storage_bench) to benchmark our use case.

Our expected data volume is enormous  -- a 1TB rocksdb database at the moment of regenesis. Benchmarking for this much data became problematic so I resorted to a mixed approach of benchmarks and linear regression.

Testing was done in memory to remove the storage speed variable.

Some expectations were given for the contract state in that it might reach numbers of 100M entries, which meant that one `ContractConfig` might take up 6.4GB of memory (or more depending on the decoding overhead) when loaded (64B per entry).

This made us uneasy with regards to keeping the `state` and `balances` collections inside `ContractConfig`.

We agreed that `ContractConfig` would need some flattening. We considered encoding the state and balances into separate files. We ultimately decided against it because of the overhead when matching the contract state and balances to the contract itself. This is manifested in the need for a 'foreign key' to be saved with each storage or balance entry. If we don't enforce an ordering while encoding then a columnar file layout would be needed so that we may locate all of the contract state and balances when parsing a single contract. We briefly considered using the rocksdb snapshot (or `sled` or `mysql`) as the target format itself but didn't pursue the idea further.

An alternative would be introducing a somewhat custom layout to fit our needs exactly. We're currently placing the following requirement on whatever data format is chosen:
* First all `CoinConfig`s are encoded, 
* Then all `MessageConfig`s are encoded,
* Finally all `ContractConfig`s are encoded. After each `ContractConfig` all of its state and balances should follow in any order before another `ContractConfig` is decoded.

[...]

PR #1413, opened 2023-10-11
Regenesis support

MujkicA:
Closes https://github.com/FuelLabs/fuel-core/issues/1450

Remaining tasks:

- [x] use new merkle root calculator
- [ ] make snapshot adhere to the new format -- #1396
- [x] add re-genesis cli command
- [ ] write back re-genesis steps to db in chunks, not per item
- [ ] adapt tests


Voxelot:
The Cargo.lock file should not be deleted.

crates/chain-config/src/config/state_importer.rs
```diff
@@ -0,0 +1,311 @@
...
+        self.dest.flush()
+    }
+
+    pub fn into_inner(self) -> Result<W, IntoInnerError<BufWriter<W>>> {
+        self.dest.into_inner()
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct StateImporter<T> {
```
MitchTurner, line 217: Would it be possible to specify what kind of state, rather than just general `StateImporter`? Like `ChainState` or whatever is most appropriate.

xgreenx:
We've split it into several smaller PRs

Issue #1329, opened 2023-08-28
TOB-FUEL-27: Dependencies are yanked or unsound


xgreenx:
## Description
The fuel-core project uses yanked dependencies like pest@2.60 and hermit-abi@0.3.1. Furthermore, the fuel-vm and fuel-core projects both depend on the unmaintained library atty with soundness issues on Windows.

## Recommendations
Short term, run cargo update which will update the yanked dependencies. Additionally, upgrade criterion to 0.5 across the projects. Unfortunately, datatest-stable has a transitive dependency on atty which can not be avoided until a new version of datatest-stable is released.
Long term, run cargo-audit in the CI and report its output. Note, that the cargo-audit tool scans the Cargo.lock file which includes false-positives due to optional dependencies.

xgreenx:
We run `cargo update` before the publishing of a new release of `fuel-core`. But it is not required action and is done manually. Maybe we can automate it somehow

PR #1415, opened 2023-10-12
Updated CI to run `cargo update` weekly

xgreenx:
I copied the flow from [rust-lang](https://github.com/rust-lang/rust/blob/master/.github/workflows/dependencies.yml).

It closes https://github.com/FuelLabs/fuel-core/issues/1329.


Dentosal (approved):
Looks neat. I wonder if this has any practical issues like causing annoyance, but if `rust-lang/rust` repo has this enabled, then it cannot be too bad.

Issue #1386, opened 2023-09-28
Sanity check opcode gas prices by executing the block with pre-defined gas limit


xgreenx:
Write a benchmark for each opcode in `benches/benches/block_target_gas.rs` and compare results with each other. The final goal is to have approximately the same execution time for each opcode.

The benchmark should create the worst possible scenario. It requires knowledge of how opcode works internally. You can look into the implementation of each opcode on the `fuel-vm` side to get a better understanding of what can affect the execution time.

- [x] `alu` (https://github.com/FuelLabs/fuel-core/pull/1419)
- [x] `contract` (https://github.com/FuelLabs/fuel-core/pull/1453)
- [x] `crypto` (https://github.com/FuelLabs/fuel-core/pull/1430)
- [x] `flow` (https://github.com/FuelLabs/fuel-core/pull/1433)
    - [ ] `ret` variants
- [x] `mem` (https://github.com/FuelLabs/fuel-core/pull/1444)
- [x] `other` (https://github.com/FuelLabs/fuel-core/pull/1476)
    - [x] `gtf` variant benchmarks

xgreenx:
Done across several pull requests

PR #1433, opened 2023-10-19
`flow` sanity benchmarks

MitchTurner:
https://github.com/FuelLabs/fuel-core/issues/1386

```
block target estimation/flow/jmp
                        time:   [1.7932 ms 1.8015 ms 1.8106 ms]
```


benches/benches/block_target_gas_set/contract.rs
```diff
@@ -0,0 +1,20 @@
+use crate::*;
```
xgreenx, line 1: Do we need this file?

benches/benches/block_target_gas_set/flow.rs
```diff
@@ -0,0 +1,72 @@
...
+    //     vec![op::ret(RegId::ONE), op::jmpb(RegId::ZERO, 0)].to_vec(),
+    //     vec![],
+    // );
+    //
+    // run(
+    //     "flow/ret_contract opcode",
+    //     group,
+    //     vec![op::ret(RegId::ONE), op::jmpb(RegId::ZERO, 0)].to_vec(),
+    //     vec![],
+    // );
```
xgreenx, line 71: Yeah, `ret` and `retd` look tricky=D But its price is essential only in the scenario if someone calls another contract and returns from it(and repeats this move many times). Because otherwise, these opcodes return to finish the current execution.

You can skip it for now, but create an issue to write a benchmark later that: the scenario either calls another contract and returns with a lot of data or runs a lot of transactions with this opcode.

benches/benches/block_target_gas_set/crypto.rs
```diff
@@ -0,0 +1,171 @@
...
+            .public
+            .as_ref()
+            .iter()
+            .chain(ed19_signature.as_ref())
+            .chain(message.as_ref())
+            .copied()
+            .collect(),
+    );
+
+    for i in generate_linear_costs() {
```
bvrooman, line 138: I understand you didn't create `generate_linear_costs()` but I'm curious to know why it's being used here. It looks like it generates `n`s that increase exponentially and then decrease exponentially, and for each `n`, we run `sha256` on `n` bytes. Is there a reason that we use this specific generator function over something like a monotonic or linear generator?

MitchTurner: I don't really understand the name--I've been trying to think of a new name. But it's used in a number of or benchmarks for dependent costs. The actual values used aren't particularly important as long as they show a good range between 0 -> a lot.

benches/benches/block_target_gas_set/contract.rs
MitchTurner: I'll probably just keep it since we're gonna add it soon anyway.

benches/benches/block_target_gas_set/flow.rs
MitchTurner: I added it as a subtask to the parent issue, like we did with `gtf`.

benches/benches/block_target_gas_set/crypto.rs
MitchTurner: Renaming `arb_dependent_cost_values` to communicate that they aren't particularly important values and what they are for.

xgreenx (approved):

PR #1435, opened 2023-10-20
Release 0.20.8

xgreenx:
Beta 4 authority node punished some sentries because of some invalid transactions they sent because of block race.

The fix is to not punish the reserved peer for sending an invalid transaction.

### DIFF
--- a/crates/services/p2p/src/p2p_service.rs
+++ b/crates/services/p2p/src/p2p_service.rs
@@ -377,8 +377,16 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
         &mut self,
         msg_id: &MessageId,
         propagation_source: PeerId,
-        acceptance: MessageAcceptance,
+        mut acceptance: MessageAcceptance,
     ) {
+        // Even invalid transactions shouldn't affect reserved peer reputation.
+        if let MessageAcceptance::Reject = acceptance {
+            if self.peer_manager.is_reserved(&propagation_source) {
+                println!("report_message_validation_result");
+                acceptance = MessageAcceptance::Ignore;
+            }
+        }
+
         if let Some(gossip_score) = self
             .swarm
             .behaviour_mut()
--- a/crates/services/p2p/src/p2p_service.rs
+++ b/crates/services/p2p/src/p2p_service.rs
@@ -382,7 +382,6 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
         // Even invalid transactions shouldn't affect reserved peer reputation.
         if let MessageAcceptance::Reject = acceptance {
             if self.peer_manager.is_reserved(&propagation_source) {
-                println!("report_message_validation_result");
                 acceptance = MessageAcceptance::Ignore;
             }
         }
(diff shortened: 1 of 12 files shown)


crates/services/p2p/src/p2p_service.rs
```diff
@@ -377,8 +377,15 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
...
-        acceptance: MessageAcceptance,
+        mut acceptance: MessageAcceptance,
     ) {
+        // Even invalid transactions shouldn't affect reserved peer reputation.
+        if let MessageAcceptance::Reject = acceptance {
+            if self.peer_manager.is_reserved(&propagation_source) {
+                acceptance = MessageAcceptance::Ignore;
+            }
+        }
+
```
xgreenx, line 388: The fix of the problem. If we remove it, a new unit test will fail.

crates/services/p2p/src/service.rs
```diff
@@ -108,7 +108,35 @@ enum TaskRequest {
...
+                write!(f, "TaskRequest::BroadcastVote")
+            }
+            TaskRequest::GetPeerIds(_) => {
+                write!(f, "TaskRequest::GetPeerIds")
+            }
+            TaskRequest::GetBlock { .. } => {
+                write!(f, "TaskRequest::GetBlock")
+            }
+            TaskRequest::GetSealedHeader { .. } => {
+                write!(f, "TaskRequest::v")
```
Voxelot, line 128:
```suggestion
                write!(f, "TaskRequest::GetSealedHeader")
```

crates/services/p2p/src/service.rs
```diff
@@ -131,9 +159,9 @@ impl<D> Task<D> {
         db: Arc<D>,
         block_importer: Arc<B>,
     ) -> Self {
-        let (request_sender, request_receiver) = mpsc::channel(100);
-        let (tx_broadcast, _) = broadcast::channel(100);
-        let (block_height_broadcast, _) = broadcast::channel(100);
```
Voxelot, line 136: interesting, we should probably compute these channel limits based on the sync parameters for batch size and number of concurrent batches etc.

tests/tests/tx_gossip.rs
```diff
@@ -90,3 +96,129 @@ async fn test_tx_gossiping() {
...
+async fn test_tx_gossiping_non_reserved_nodes_invalid_txs() {
+    // This test is opposite to the `test_tx_gossiping_reserved_nodes_invalid_txs`.
+    // It verifies that gossiping about invalid transactions from
+    // non-reserved peers disconnects them.
+    //
+    // The test sends `NUMBER_OF_INVALID_TXS` invalid transactions,
+    // and verifies that sending a valid one will be ignored.
+    let status = test_tx_gossiping_invalid_txs(BootstrapType::BootstrapNodes).await;
+
+    assert!(status.is_err());
```
bvrooman, line 223: Is it possible to assert a match for the specific error here, so we know that this error occurs because the sender was disconnected/rejected?

bvrooman (approved):

PR #1452, opened 2023-10-25
test: Add contract code root benchmark

bvrooman:
Related issues:
- https://github.com/FuelLabs/fuel-vm/issues/599

This PR adds a benchmark to determine the gas price of calculating contract roots.


benches/src/bin/collect.rs
```diff
@@ -570,9 +581,10 @@ impl State {
         let relative = groups.into_iter().filter_map(|(name, samples)| {
             let relative = samples
                 .into_iter()
-                .find(|sample| sample.ends_with(&name))
```
bvrooman, line 573: This was causing the problem with new benchmarks not appearing in the final output. 

Benchmarks have a group name and a benchmark name, and together they form a sample. E.g., a `foo_group` group with benchmarks `test_n_1`, `test_n_10`, and `test_n_100`:

```rust
pub fn foo_tests(c: &mut Criterion) {
    let mut group = c.benchmark_group("foo_group");

    group.bench_function("test_n_1", |b| {
        b.iter(|| {...})
    });

    group.bench_function("test_n_10", |b| {
        b.iter(|| {...})
    });

    group.bench_function("test_n_100", |b| {
        b.iter(|| {...})
    });

    group.finish();
}

```

 would produce the following samples: 
- `foo_group/test_n_1`
- `foo_group/test_n_10` 
- `foo_group/test_n_100`

The above check was testing that the sample _ends_ with the group name. The above example benchmarks would fail this check and be excluded by the `find` because they instead _start_ with the group name.

benches/src/bin/collect.rs
```diff
@@ -570,9 +581,10 @@ impl State {
...
+                .find(|sample| sample.starts_with(&name) && ids.contains_key(sample))
```
bvrooman, line 584: This fixes the name check issue.

`ids.contains_key(sample)` is also needed because the list of samples also includes any benches found previously in the `reports` folder. If the `"group_complete" line lists a bench for this group not found in a `"benchmark-complete"` line above it, the mapping from the sample to the mean would silently fail, causing the benchmark not to appear in the final output.

bvrooman:
Currently blocked on a new release of fuel-vm, but ready for review.

crates/client/src/client/schema/chain.rs
```diff
@@ -236,6 +237,8 @@ include_from_impls_and_cynic! {
         pub smo: DependentCost,
         pub srwq: DependentCost,
         pub swwq: DependentCost,
+
+        pub contract_root: U64,
```
xgreenx, line 241: Have you tried to write a benchmark as a `DependentCost`? Maybe it is better if we charge users based on the size of the contract. But if there is not a huge difference in result, then it's okay to charge for possible max value.

xgreenx (commented):
It would be nice if you posted an approximate gas price for root calculation based on your benchmarks.

bvrooman:
I updated the benchmark and the gas costs to make `contract_root` a `DependentCost`. Collecting gas costs locally gives me the following for `contract_root`:
```
        contract_root: DependentCost {
            base: 75,
            dep_per_unit: 1,
        },
```

I have updated this value in [my PR](https://github.com/FuelLabs/fuel-vm/pull/613) for `fuel-vm`.

benches/benches/contract_root.rs
```diff
@@ -0,0 +1,48 @@
...
+
+fn random_bytes<R: Rng + ?Sized>(n: usize, rng: &mut R) -> Vec<u8> {
+    let mut bytes = vec![0; n];
+    for chunk in bytes.chunks_mut(32) {
+        rng.fill(chunk);
+    }
+    bytes
+}
+
+pub fn contract_root(c: &mut Criterion) {
```
xgreenx, line 21: Maybe you want to move this file into `vm_set` module=)

xgreenx (approved):

benches/benches/contract_root.rs
bvrooman: I don't think I would like to do put this under `vm_set`, since this doesn't really touch the VM. I think it would be better to consolidate these files by adding a parent folder to contain both VM and non-VM items used for gas pricing. I will merge this PR and look at that for the next one.

Issue #1428, opened 2023-10-18
Investigate slow start up of the GraphQL server in beta 3


xgreenx:
The authority node in beta 3 experienced a long launch after the restart. Because the launch was longer than 30 seconds, the GraphQL server, required for liveness check, was unavailable. After 30 seconds the liveness check failed and restarted the node again. It had been happening until we removed the liveness check.

There was only one log - the version of the image:

```shell
INFO fuel_core::cli::run: 348: Fuel Core version v0.17.13
```

It seems something before starting the services or one of the services took a lot of time to start. Maybe opening the rocksdb after the invalid closing of it takes a lot of time.

PR #1456, opened 2023-10-26
Call `flush` during end of the `FuelService`

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1428

Call the `flush` function during the end of the `FuelService` to move the processing of WAL and SST fiels from the node's launching to the node's shutdown.

Added additional logs to track the performance of the starting/shutdown process.

### DIFF
--- a/crates/fuel-core/src/service.rs
+++ b/crates/fuel-core/src/service.rs
@@ -68,7 +68,6 @@ impl FuelService {
     #[tracing::instrument(skip_all, fields(name = %config.name))]
     pub fn new(database: Database, config: Config) -> anyhow::Result<Self> {
         let config = config.make_config_consistent();
-        database.init(&config.chain_conf)?;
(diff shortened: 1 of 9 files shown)


crates/fuel-core/src/service.rs
```diff
@@ -251,6 +258,7 @@ impl RunnableTask for Task {
                 );
             }
         }
+        self.shared.database.flush()?;
```
MitchTurner, line 261: Nit: There are a few places that we're calling `self.something.somethig_else...`.

It might be better to create a helper functions for each of these:
`self.flush_database()`, `self.flush_view_layer()`, etc

see: https://en.wikipedia.org/wiki/Law_of_Demeter

MitchTurner (approved):
I left one nit, but other than that LGTM.

Issue #1431, opened 2023-10-19
Reduce the impact of the OS clocks on the benchmarks results


xgreenx:
Related https://github.com/FuelLabs/fuel-core/issues/948

PR #1457, opened 2023-10-26
Fixing incorrect measurement for fast(µs) opcodes

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1431

In the case of super-fast opcodes(< than 1 µs), the `elapse` method is not accurate enough, which leads to unstable results from benchmarks.

The change aims to stabilize the results from measurements by subtracting the time required to reset the VM from the total time of the benchmark.

### DIFF
--- a/benches/benches/vm.rs
+++ b/benches/benches/vm.rs
@@ -9,7 +9,6 @@ use criterion::{
     BenchmarkGroup,
     Criterion,
 };
-use std::time::Duration;
 
 use fuel_core_benches::*;
 use fuel_core_storage::transactional::Transaction;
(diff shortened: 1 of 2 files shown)


benches/benches/vm.rs
```diff
@@ -41,25 +40,48 @@ where
...
                 let start = std::time::Instant::now();
-                match instruction {
-                    Instruction::CALL(call) => {
-                        let (ra, rb, rc, rd) = call.unpack();
-                        vm.prepare_call(ra, rb, rc, rd).unwrap();
-                    }
-                    _ => {
-                        black_box(vm.instruction(*instruction).unwrap());
+                for _ in 0..iters {
+                    vm.reset_vm_state(diff);
```
Voxelot, line 50: Is this actually measuring anything? It seems like this would essentially be a noop if we haven't executed any instructions yet 🤔

xgreenx: During setup of the bench we run instruction one time to take a `diff` 
image

Voxelot: Have we run the benchmarks to see how much of an improvement this is?

xgreenx: Yes, now it produces more stable results 
image
image

Voxelot: that's great!

Voxelot (approved):

PR #1458, opened 2023-10-27
Add a spellcheck into the CI

xgreenx:


.cargo/common.dic
```diff
@@ -0,0 +1,123 @@
...
+TODO
+tokio
+transactional
+TTL
+tx
+TxPool/S
+txs
+uncompress
+uri
+UTXO/S
```
MitchTurner, line 112: I prefer `UTxO` tbo, so maybe add that too or instead?

MitchTurner: Hmm. Maybe that's not standard:
https://en.wikipedia.org/wiki/Unspent_transaction_output

Althought I prefer it, maybe just stick with the standard.

bin/fuel-core/src/cli/run/relayer.rs
```diff
@@ -15,32 +15,32 @@ use std::str::FromStr;
 #[derive(Debug, Clone, Args)]
 pub struct RelayerArgs {
     /// Enable the Relayer. By default, the Relayer is disabled, even when the binary is compiled
-    /// with the "relayer" feature flag. Providing `--enable-relayer` will enable the relayer
+    /// with the "Relayer" feature flag. Providing `--enable-relayer` will enable the Relayer
```
MitchTurner, line 18: I'm not sure that "relayer" should be capitalized. Maybe add it to the list in lowercase?

crates/fuel-core/src/executor.rs
```diff
@@ -967,22 +967,22 @@ where
...
+                // ensure tx matches VM output exactly
                 if &tx != checked_tx.transaction() {
                     return Err(ExecutorError::InvalidTransactionOutcome {
                         transaction_id: tx_id,
                     })
                 }
             }
             ExecutionKind::DryRun | ExecutionKind::Production => {
-                // malleate the block with the resultant tx from the vm
+                // malleable the block with the resultant tx from the VM
```
MitchTurner, line 978: "malleable" isn't a verb. I'm not sure what "malleate" means either. Maybe we choose a new word.

crates/fuel-core/src/executor.rs
```diff
@@ -1485,7 +1485,7 @@ where
         }
     }
 
-    /// Log a VM backtrace if configured to do so
+    /// Log a VM backtraces if configured to do so
```
MitchTurner, line 1482: I think "backtrace" was more correct.

crates/fuel-core/src/executor/ports.rs
```diff
@@ -51,8 +51,8 @@ pub trait RelayerPort {
 
 #[cfg(test)]
 /// For some tests we don't care about the actual
-/// implementation of the RelayerPort
-/// and using a passthrough is fine.
+/// implementation of the `RelayerPort`
+/// and using a pass through is fine.
```
MitchTurner, line 55: "pass-through"

crates/fuel-core/src/schema/block.rs
```diff
@@ -69,7 +69,7 @@ type CoreConsensus = fuel_core_types::blockchain::consensus::Consensus;
 
 #[derive(SimpleObject)]
 pub struct Genesis {
-    /// The chain configs define what consensus type to use, what settlement layer to use,
+    /// The chain config define what consensus type to use, what settlement layer to use,
```
MitchTurner, line 72: config "defines"

crates/services/producer/src/block_producer.rs
```diff
@@ -88,14 +88,14 @@ where
...
         //  - get previous block info (hash, root, etc)
-        //  - select best da_height from relayer
-        //  - get available txs from txpool
+        //  - select best `da_height` from Relayer
+        //  - get available txs from TxPool
         //  - select best txs based on factors like:
         //      1. fees
         //      2. parallel throughput
-        //  - Execute block with production mode to correctly malleate txs outputs and block headers
+        //  - Execute block with production mode to correctly malleable txs outputs and block headers
```
MitchTurner, line 96: Here's another "malleable" being used as a verb.

crates/services/producer/src/block_producer.rs
```diff
@@ -144,8 +144,8 @@ where
         // The dry run execution should use the state of the blockchain based on the
         // last available block, not on the upcoming one. It means that we need to
         // use the same configuration as the last block -> the same DA height.
-        // It is deterministic from the result perspective, plus it is more performant
-        // because we don't need to wait for the relayer to sync.
+        // It is deterministic from the result perspective, plus it is more performer
```
MitchTurner, line 147: What was wrong with "performant"? "performer" is wrong.

crates/services/producer/src/block_producer.rs
```diff
@@ -201,7 +201,7 @@ where
         let best_height = self.relayer.wait_for_at_least(&previous_da_height).await?;
         if best_height < previous_da_height {
             // If this happens, it could mean a block was erroneously imported
-            // without waiting for our relayer's da_height to catch up to imported da_height.
+            // without waiting for our `Relayer`'s `da_height` to catch up to imported `da_height`.
```
MitchTurner, line 204: We use backticks here (`) but not in other places.

crates/client/assets/schema.sdl
```diff
@@ -258,7 +258,7 @@ type DependentCost {
 
 input ExcludeInput {
 	"""
-	Utxos to exclude from the selection.
+	UTXO to exclude from the selection.
```
MitchTurner, line 261: "UTXOs"

crates/services/relayer/src/service.rs
```diff
@@ -252,13 +252,13 @@ impl<D> SharedState<D> {
     /// Wait for the [`Task`] to be in sync with
     /// the data availability layer.
     ///
-    /// Yields until the relayer reaches a point where it
+    /// Yields until the Relayer reaches a point where it
```
bvrooman, line 255:
```suggestion
    /// Yields until the Relayer reaches a point where it is
```

crates/services/relayer/src/test_helpers/middleware.rs
```diff
@@ -165,7 +165,7 @@ impl Default for MockMiddleware {
 #[derive(Error, Debug)]
 /// Thrown when an error happens at the Nonce Manager
 pub enum MockMiddlewareError {
-    /// Thrown when the internal middleware errors
+    /// Thrown when the internal middle ware errors
```
MitchTurner, line 168: I think "middleware" is correct.

crates/services/src/service.rs
```diff
@@ -66,13 +66,13 @@ pub trait Service {
 /// Trait used by `ServiceRunner` to encapsulate the business logic tasks for a service.
 #[async_trait::async_trait]
 pub trait RunnableService: Send {
-    /// The name of the runnable service, used for namespacing error messages.
+    /// The name of the runnable service, used for name spacing error messages.
```
MitchTurner, line 69: "namespacing" is correct

crates/services/relayer/src/test_helpers/middleware.rs
bvrooman, line 168: This should still be `"middleware"`

crates/services/src/service.rs
bvrooman, line 69: This should still be `"namespacing"`

crates/services/sync/src/ports.rs
```diff
@@ -64,7 +64,7 @@ pub trait PeerToPeerPort {
 pub trait ConsensusPort {
     /// Check if the given sealed block header is valid.
     fn check_sealed_header(&self, header: &SealedBlockHeader) -> anyhow::Result<bool>;
-    /// await for this DA height to be sync'd.
+    /// await for this DA height to be synced.
```
bvrooman, line 67:
```suggestion
    /// await for this DA height to be synchronized.
```

crates/services/txpool/src/txpool.rs
```diff
@@ -227,8 +227,8 @@ where
         tx_status_sender: &TxStatusChange,
         txs: Vec<Checked<Transaction>>,
     ) -> Vec<anyhow::Result<InsertionResult>> {
-        // Check if that data is okay (witness match input/output, and if recovered signatures ara valid).
-        // should be done before transaction comes to txpool, or before it enters RwLocked region.
+        // Check if that data is okay (witness match input/output, and if recovered signatures  valid).
```
bvrooman, line 230:
```suggestion
        // Check if that data is okay (witness matches input/output, and recovered signatures are valid).
```

crates/types/src/blockchain/block.rs
```diff
@@ -99,7 +99,7 @@ impl Block<Transaction> {
 }
 
 impl<T> Block<T> {
-    /// Destructure into the inner types.
+    /// Destructs into the inner types.
```
bvrooman, line 102: This should still be "Destructure"

crates/types/src/blockchain/block.rs
```diff
@@ -41,7 +41,7 @@ pub type CompressedBlock = Block<TxId>;
 
 /// Fuel block with all transaction data included
 /// but without any data generated.
-/// This type can be created with unexecuted
+/// This type can be created with non-executed
```
bvrooman, line 44: This should still be "unexecuted"

.cargo/common.dic
```diff
@@ -0,0 +1,125 @@
...
+(
+)
+*
+/
+@dev
+^
++
+<
+=
+>
```
bvrooman, line 13: Why are these included in the dictionary?

.cargo/common.dic
```diff
@@ -0,0 +1,125 @@
...
+Jemalloc
+JSON
+Kademlia
+keygen
+keypair
+libp2p
+lifecycle
+MDNS
+malleate
+merkle
```
bvrooman, line 66:
```suggestion
Merkle
```

bvrooman (commented):
A couple of concerns:
- There are currently a number of things added to the dictionary that are either real words that probably don't need to be there, or fake words/shorthands that probably shouldn't be there
- This check can only do so much, and we should be more attentive to how we write our comments in general. And as part of the PR review process, we should be catching spelling and grammar mistakes at that time
- Some other acronyms like "db" should arguably be flagged, and we should instead use something like "DB". It does not seem like this check is catching that.

.cargo/common.dic
xgreenx: Otherwise, we can't use formulas in the comments=)

crates/fuel-core/src/executor.rs
Dentosal: I would suggest "mutate" here

crates/services/producer/src/block_producer.rs
Dentosal: Maybe "modify" instead?

Dentosal:
I don't think this is worth doing. Too many false positives, and the impact of typos is really low.

crates/services/producer/src/block_producer.rs
Voxelot: We could go with modify, but malleate carries some context from the bitcoin world which is useful conceptually.

xgreenx:
> I don't think this is worth doing. Too many false positives, and the impact of typos is really low.

There are indeed a lot of false positives because we need to set the dictionary that works with our style of comments=) And it is almost done because we have already received reviews from everyone.

It is nice to have a spellchecker for the open-source project. Because it is already done(we spent some time adding it and reviewing it), we can integrate it(plus it is an optional part of the CI). But I'm also okay to close this pull request if no one approves it by the end of this week=)

xgreenx:
No one approved, so I'm closing this PR=)

Issue #1450, opened 2023-10-25
Extract the `initial_state` from the ChainConfig


xgreenx:
## Overview

Extract the initial_state from the `ChainConfig`. We still need it during the start of the network, so this information should be provided along with `--chain` argument. We want to store an initial state in a separate folder. 

We have 3 flows:
- Contracts
- Messages
- UTXOs

Each flow should have its own file with the state. It allows us parallelize initialing from the snapshot in the future.

## Implementation details

The `init_height` or `base_height` should be static part of `ChainConfig`.
A new `--chain-initia-state` arguments accepts path to the folder, where initial state is stored(Not to the file). it should be part of the `fuel-core run` command.
The `import_genesis_block` function should split importing of each flow into a separate task and do in parallel. Each task should return unordered vector of merkle roots. Before calculating the final root for each flow we need to sort all roots and create a final BMT root after.

xgreenx:
Implemented during https://github.com/FuelLabs/fuel-core/pull/1459. The part related to parallelization will be done in a separate PR.

PR #1459, opened 2023-10-27
Split initial state from chain config

MujkicA:
Closes https://github.com/FuelLabs/fuel-core/issues/1450

Separates the loading of the initial chain state from loading `ChainConfig`.

The parameters to the run command have been renamed to `chain_params` and `chain_state`. Should we  rename the corresponding structs `ChainConfig` and `StateConfig` accordingly? 
The block height is moved to the `ChainConfig` to be available instantly because the `StateConfig` will be streamed in a later PR.


bin/fuel-core/src/cli/snapshot.rs
```diff
@@ -69,19 +73,25 @@ pub async fn exec(command: Command) -> anyhow::Result<()> {
...
+        } => {
+            let chain_params = match chain_config.as_str() {
+                LOCAL_TESTNET => ChainConfig::local_testnet(),
+                _ => ChainConfig::load_from_directory(&chain_config)?,
             };
+            let chain_state = StateConfig::generate_state_config(db)?;
 
-            let stdout = std::io::stdout().lock();
+            std::fs::create_dir_all(output_dir)?;
+            let parameters_writer = File::create("chain_parameters.json")?;
```
Voxelot, line 87: These strings should be constants for easier identification and later refactoring.

```suggestion
            let parameters_writer = File::create("CHAIN_PARAMS_FILENAME")?;
```

crates/chain-config/src/config/state.rs
```diff
@@ -44,9 +65,108 @@ impl StateConfig {
             coins: db.get_coin_config()?,
             contracts: db.get_contract_config()?,
             messages: db.get_message_config()?,
-            height: Some(db.get_block_height()?),
         })
     }
+
+    #[cfg(feature = "std")]
+    pub fn load_from_directory(path: &str) -> Result<Self, anyhow::Error> {
+        let path = PathBuf::from_str(path)?.join("chain_state.json");
```
Voxelot, line 73: This should be a constant that's reused for exporting state as well.

crates/chain-config/src/config/chain.rs
```diff
@@ -68,128 +56,38 @@ impl Default for ChainConfig {
...
-            chain_name: LOCAL_TESTNET.to_string(),
-            initial_state: Some(StateConfig {
-                coins: Some(initial_coins),
-                ..StateConfig::default()
-            }),
-            ..Default::default()
-        }
+    #[cfg(feature = "std")]
+    pub fn load_from_directory(path: &str) -> Result<Self, anyhow::Error> {
+        let path = PathBuf::from_str(path)?.join("chain_parameters.json");
```
Voxelot, line 70: This string should also be a constant

crates/chain-config/src/config/state.rs
```diff
@@ -44,9 +65,108 @@ impl StateConfig {
...
+            ))
+        })
+    }
+
+    #[cfg(feature = "std")]
+    pub fn local_testnet() -> Self {
+        // endow some preset accounts with an initial balance
+        tracing::info!("Initial Accounts");
+        let secrets = [
+            "0xde97d8624a438121b86a1956544bd72ed68cd69f2c99555b08b1e8c51ffd511c",
```
Voxelot, line 93: Less concerned about having these as constants, although it could be useful for testing if the secrets are easy to reference.

Voxelot:
On a quick initial pass - these changes look straightforward and also nicely cleans up some ugly parts of the code around the configurable block height. Will wait for more of the client team to review before approving.

crates/chain-config/src/config.rs
```diff
@@ -14,52 +14,31 @@ pub use state::*;
...
+        RngCore,
+        SeedableRng,
     };
 
+    use super::chain::ChainConfig;
+
     #[cfg(feature = "std")]
     #[test]
-    fn from_str_loads_from_file() {
+    fn from_file_loads_from_file() {
```
MitchTurner, line 34: I wonder if we should be testing the writing code too. That functionality is just written into the cli endpoint rather than having its own helper like `load_from_file`. 

It would be good to have a `create_config_file` rather than testing against the `serde` and `write`.

crates/chain-config/src/config/state.rs
```diff
@@ -59,3 +179,265 @@ pub trait ChainConfigDb {
...
+    }
+
+    #[cfg(feature = "std")]
+    #[test]
+    fn from_file_loads_from_file() {
+        // setup chain config in a temp file
+        let tmp_file = tmp_path();
+        let disk_config = StateConfig::local_testnet();
+        let json = serde_json::to_string_pretty(&disk_config).unwrap();
+        write(tmp_file.clone(), json).unwrap();
```
MitchTurner, line 234: Similar thing here.

crates/services/txpool/src/config.rs
```diff
@@ -63,7 +63,7 @@ impl Config {
             max_depth,
             min_gas_price,
             utxo_validation,
-            chain_config,
+            chain_parameters: chain_config,
```
xgreenx, line 66: Maybe it is better to use one name, either `chain_config` or `chain_parameters` or `chain_params`. I prefer `chain_config` more=)

crates/chain-config/src/config/state.rs
```diff
@@ -59,3 +189,254 @@ pub trait ChainConfigDb {
...
+    #[test]
+    fn can_roundtrip_simple_message_state() {
+        let config = test_message_config();
+        let json = serde_json::to_string(&config).unwrap();
+        let deserialized_config: StateConfig =
+            serde_json::from_str(json.as_str()).unwrap();
+        assert_eq!(config, deserialized_config);
+    }
+
+    fn test_config_contract(
```
MitchTurner, line 351: It doesn't look like any of these fields are used at the same time. It might make more sense to move the shared code out to a `base_state_config()` and create a small helper function for each case. e.g. 
```
fn state_config_with_state() -> StateConfig {
    let test_key: Bytes32 = rng.gen();
    let test_value: Bytes32 = rng.gen();
    let state = Some(vec![(test_key, test_value)]);
    StateConfig {
        state,
        ..base_state_config()
    }
}
```
What you have would also be more readable as a builder pattern, but I think that's more complexity than is needed here.

crates/fuel-core/src/service/genesis.rs
```diff
@@ -84,11 +85,11 @@ fn import_genesis_block(
...
-    let chain_config_hash = config.chain_conf.root()?.into();
-    let coins_root = init_coin_state(database, &config.chain_conf.initial_state)?.into();
-    let contracts_root =
-        init_contracts(database, &config.chain_conf.initial_state)?.into();
-    let messages_root = init_da_messages(database, &config.chain_conf.initial_state)?;
+    let height = config.chain_config.height;
+    let chain_config_hash = config.chain_config.root()?.into();
+    let coins_root = init_coin_state(database, &config.chain_state, height)?.into();
+    let contracts_root = init_contracts(database, &config.chain_state, height)?.into();
+    let messages_root = init_da_messages(database, &config.chain_state)?;
```
MitchTurner, line 92: Can we just have the `into` here instead of having a redundant variable right after?

crates/fuel-core/src/service/genesis.rs
```diff
@@ -151,150 +147,146 @@ fn import_genesis_block(
 
 fn init_coin_state(
     db: &mut Database,
-    state: &Option<StateConfig>,
+    state: &StateConfig,
+    height: Option<BlockHeight>,
```
MitchTurner, line 151: Just curious, why did the height get pulled out of the `StateConfig`?

MujkicA: It doesn't make a big difference in the current PR but the final goal is for the state config to be consumed as a stream because of memory limitations. 
During the streaming process, we make use of the height to validate tx pointers. Having it as a par of the chain config is more convenient since it's loaded into memory before we start streaming.

MitchTurner (approved):
LGTM. I feel like I'm lacking a little context in the goals of the overall feature and the followup work, but I'm happy with everything here.

deployment/scripts/chainspec/beta_chainstate_spec.json
```diff
@@ -0,0 +1,41 @@
+{
```
xgreenx, line 1: Because the name of state and params is hardcoded, you need to create `beta` and `dev` folders that contain both files with the right name. We use this config directly for our dev and prod environments, so it should be valid.

tests/tests/deployment.rs
```diff
@@ -1,17 +1,29 @@
...
-    include_str!("../../deployment/scripts/chainspec/dev_chainspec.json"); 
+    include_str!("../../deployment/scripts/chainspec/dev_chainparams_spec.json"),
+    include_str!("../../deployment/scripts/chainspec/dev_chainstate_spec.json");
     "Dev chainconfig"
 )]
-fn test_deployment_chainconfig(bytes: &str) {
-    let chain_config: ChainConfig =
-        serde_json::from_str(bytes).expect("Should be able to decode chain config");
-    let actual_bytes = serde_json::to_string_pretty(&chain_config)
+fn test_deployment_chainconfig(params_bytes: &str, state_bytes: &str) {
```
xgreenx, line 16: The same here. We need to work with directories and use the building methods to `ChainConfig` and `StateConfig`.

crates/fuel-core/src/service/config.rs
```diff
@@ -72,7 +74,8 @@ pub struct Config {
 
 impl Config {
     pub fn local_node() -> Self {
-        let chain_conf = ChainConfig::local_testnet();
+        let chain_params = ChainConfig::local_testnet();
```
xgreenx, line 77: Maybe it is better to keep naming according to the type `chain_config`

xgreenx (commented):
Thank you, looks definitely much better than everything in one structure!=)

Small comments that should be easy to resolve and we can merge

bin/fuel-core/src/cli/run.rs
```diff
@@ -250,7 +253,14 @@ impl Command {
...
 
-        let chain_conf: ChainConfig = chain_config.as_str().parse()?;
+        let (chain_params, chain_state) = match chain_config.as_str() {
+            LOCAL_TESTNET => (ChainConfig::local_testnet(), StateConfig::local_testnet()),
+            _ => {
+                let chain_conf = ChainConfig::load_from_directory(&chain_config)?;
+                let chain_state = StateConfig::load_from_directory(&chain_config)?;
+                (chain_conf, chain_state)
+            }
+        };
```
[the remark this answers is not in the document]
MujkicA: We still want to support someone entering `local_testnet` as well right?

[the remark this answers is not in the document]
xgreenx: Nope, we can skip it. If the user hasn't specified the path, we will use `local_testnet` by default.

tests/tests/deployment.rs
```diff
@@ -2,23 +2,27 @@ use fuel_core::chain_config::ChainConfig;
...
+
     // Deployment configuration should use gas costs from benchmarks.
-    chain_config.consensus_parameters.gas_costs = benchmark_gas_costs;
-    let actual_bytes = serde_json::to_string_pretty(&chain_config)
-        .expect("Should be able to encode the chain config");
-    assert_eq!(actual_bytes, bytes);
+    assert_eq!(
+        chain_config.consensus_parameters.gas_costs,
+        benchmark_gas_costs
+    );
```
xgreenx, line 25: Setting the `gas_costs = benchmark_gas_costs;` enforced the prices according to latest benchmarks. Could we return it back?=)

tests/tests/deployment.rs
```diff
@@ -2,23 +2,27 @@ use fuel_core::chain_config::ChainConfig;
...
+    "Dev chainconfig" 
 )]
-fn test_deployment_chainconfig(bytes: &str) {
-    let mut chain_config: ChainConfig =
-        serde_json::from_str(bytes).expect("Should be able to decode chain config");
+fn test_deployment_chainconfig(path: &str) {
+    let chain_config = ChainConfig::load_from_directory(path)
+        .expect("Should be able to load chain config");
+
+    // TODO: add state config check in subsequent PR
```
xgreenx, line 16: Could you create an issue to track it?=)

xgreenx (approved):

crates/fuel-core/src/service/genesis.rs
```diff
@@ -151,153 +146,146 @@ fn import_genesis_block(
...
+                            .chain(
+                                (generated_output_index / 255).to_be_bytes().into_iter(),
+                            )
+                            .collect_vec()
+                            .as_slice(),
+                    )
+                    .expect("Incorrect genesis transaction id byte length")
+                }),
+                coin.output_index.unwrap_or_else(|| {
+                    generated_output_index += 1;
```
bvrooman, line 173: It looks like you removed the `checked_add(1)` - is there any specific reason?

MujkicA: Must have been an accident. Thanks!

tests/tests/deployment.rs
MujkicA: This TODO will be merged into the subsequent regenesis PRs and will be addressed there with an updated test before it hits the master branch. Maybe we can skip opening an issue in this case?

bvrooman (approved):
Thanks for putting this together and addressing all the feedback!

tests/tests/deployment.rs
segfault-magnet: I'm not sure I follow what the issue is:

Previously the test had:
1. loaded the configuration from the file
2. changed the gas costs of the *loaded* configuration
3. checked that the loaded configuration, when serialized, is equal to the one from the file.

Had the gas costs in the file been anything other than what was checked in step #3 then the test would have failed.

So we interpreted the test as to mean:

Validate that:
1. you can decode the config (i.e. given file is a valid configuration)
2. check that the gas costs encoded in the file are equal to the given value (`benchmark_gas_costs`)

Since we needed to touch these tests for our changes we changed the assertion to be more direct. 
1. Check that the configuration is valid (i.e. can be decoded)
2. Check that the gas costs loaded from the file are as we expect.

We didn't check the encoding again since that is a general test that would belong over where the rest of the encoding/decoding is being tested -- i.e. the goal here is not to test the encoding/decoding but rather test that the committed configuration is valid and has expected values (i.e. prevent anyone from changing the gas costs accidentally)

Did we understand this right?

xgreenx: Initially, this test was added to verify the decoding and encoding of the chain configuration. At the same time, enforce the same layout inside the chain configuration as during encoding.

It is also our way to update chain configs by copying/pasting the result from the test into chain configuration files directly.

Later, I also added a check for gas costs later as one of the requirements for these configurations.

Without testing encoding, we can't be sure that it is exactly the same config. Plus, removing it makes the process of updating `beta_chainspec.json` and `dev_chainspec.json` much harder=)

benches/benches/vm_initialization.rs
```diff
@@ -88,7 +88,7 @@ pub fn vm_initialization(c: &mut Criterion) {
     loop {
         let size = 8 * (1 << i);
         if size as u64 > consensus_params.script_params.max_script_data_length {
-            break
+            break;
```
xgreenx, line 91: Is it possible to remove changes like that from the PR to minimise the number of conflicts in the future?

xgreenx (approved):

benches/benches/vm_initialization.rs
MujkicA: Seems like my formatter did something there. I got it removed, thanks!

xgreenx (approved):

PR #1462, opened 2023-10-30
test: state root and contract id benchmarks

bvrooman:


xgreenx (approved):

bvrooman:
I was trying to refactor some of the benches into a separate bench target so that we could organize the files a little more cleanly.

`gas_costs.rs`:
```rust
mod contract;
mod vm;

use contract::contract;
use vm::vm;

use criterion::{
    criterion_group,
    criterion_main,
    Criterion,
};

/// Run the benchmarks responsible for generating the data used to
/// calculate gas costs
pub fn gas_costs(c: &mut Criterion) {
    contract(c);
    vm(c);

criterion_group!(benches, gas_costs);
criterion_main!(benches);
```

However, the way Rust resolves files outside of libraries/binaries is a bit mysterious here, and this fails to compile the `vm` module, citing `unresolved import ...` for many things.

To minimize the change, I won't try to move files into a `src` folder (this would constitute a large change) just to make this possible. Instead, I think I am fine with keeping `contract.rs` where it is. I also don't want to move `contract.rs` into `vm_set` since it is not related to the VM, strictly speaking. I think it is fine to leave it as is, at least for now.

xgreenx (approved):

PR #1463, opened 2023-10-30
`DB::repair` breaks the table for checkpoints

xgreenx:


crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -91,7 +88,18 @@ impl RocksDb {
                         }
                         Ok(db)
                     }
-                    err => err,
+                    _ => {
+                        DB::repair(&opts, &path)
```
Voxelot, line 92: Maybe we should add some logging here so we know if the database is going into repair mode?

Voxelot (approved):

Issue #1261, opened 2023-07-25
TOB-FUEL-23: Consensus parameters are potentially incompatible on 32bit platforms


Dentosal:
## Description
The consensus parameters define the maximum length of vectors such as max_inputs, max_outputs, max_witnesses to be 64-bit values.These values are cast to usize when validating that a transaction’s field does not contain a vector which exceeds these maximums (see figure 23.2). On 32-bit architectures, usize will truncate consensus parameter values larger than 32 bits and reject different transactions than 64-bit platforms.

#### Figure 23.1: The ConsensusParameters struct definition (fuel-core/crates/client/src/client/schema/chain.rs#11–27)

```rust
pub struct ConsensusParameters {
    pub contract_max_size: U64,
    pub max_inputs: U64,
    pub max_outputs: U64,
    pub max_witnesses: U64,
    pub max_gas_per_tx: U64,
    pub max_script_length: U64,
    pub max_script_data_length: U64,
    pub max_storage_slots: U64,
    pub max_predicate_length: U64,
    pub max_predicate_data_length: U64,
    pub max_gas_per_predicate: U64,
    pub gas_price_factor: U64,
    pub gas_per_byte: U64,
    pub max_message_data_length: U64,
    pub chain_id: U64,
}
```

#### Figure 23.2: Validation of transactions against consensus parameters (fuel-vm/fuel-tx/src/transaction/validity.rs#325–335)

```rust
if tx.inputs().len() > parameters.max_inputs as usize {
    Err(CheckError::TransactionInputsMax)?
}
if tx.outputs().len() > parameters.max_outputs as usize {
    Err(CheckError::TransactionOutputsMax)?
}
if tx.witnesses().len() > parameters.max_witnesses as usize {
    Err(CheckError::TransactionWitnessesMax)?
}
```
## Recommendations
Short term, define the consensus parameters as 32-bit unsigned integers.
Long term, review platform-dependent behavior such as casts to usize and ensure that code is compatible across architectures.

## Our thoughts
The actual use occurs here:
https://github.com/FuelLabs/fuel-vm/blob/20f1e9b888e359b9d1ac45214673e97277370543/fuel-tx/src/transaction/validity.rs#L325-L335

Either change the data types to be `u32`, or remove `usize` casts from the fuel-vm side. Or maybe specify that fuel-vm/core only work on 64bit platforms and add test to that effect.

PR #1464, opened 2023-10-30
Avoid possible truncation of higher bits

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1261

Awaits reelasing of the https://github.com/FuelLabs/fuel-vm/pull/619.


crates/services/p2p/src/p2p_service.rs
```diff
@@ -167,8 +167,10 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
 
         let total_connections = {
             // Reserved nodes do not count against the configured peer input/output limits.
-            let total_peers =
-                config.max_peers_connected + config.reserved_nodes.len() as u32;
+            let total_peers = config.max_peers_connected
+                + u32::try_from(config.reserved_nodes.len()).expect(
+                    "The numbere of reserved nodes should be less than `u32::max`",
```
Dentosal, line 172:
```suggestion
                    "The number of reserved nodes should be less than `u32::MAX`",
```

Dentosal (approved):
No complants. As long as the tests still pass, this seems to be fine.

crates/client/src/client.rs
```diff
@@ -464,7 +464,7 @@ impl FuelClient {
         self.query(query).await.map(|r| r.execute)
     }
 
-    pub async fn register(&self, id: &str, register: RegisterId) -> io::Result<Word> {
+    pub async fn register(&self, id: &str, register: u32) -> io::Result<Word> {
```
bvrooman, line 467: Going from `RegisterId` to `u32` appears to be losing domain specificity. Can the `RegisterId` type alias `u32` instead?

crates/client/src/client/pagination.rs
```diff
@@ -11,7 +11,7 @@ pub struct PaginationRequest<T> {
     /// The cursor returned from a previous query to indicate an offset
     pub cursor: Option<T>,
     /// The number of results to take
-    pub results: usize,
+    pub results: i32,
```
MitchTurner, line 14: Why `i32`? Isn't this a cardinal number, will it ever be negative?

xgreenx: It can't be negative, but using `u32` will cause casting from `u32` to `i32` that potentially can truncate the last bit=) So I decided to not support the last bit at all by using `i32`

crates/fuel-core/src/executor.rs
```diff
@@ -1500,7 +1500,7 @@ where
                     backtrace.contract(),
                     backtrace.registers(),
                     backtrace.call_stack(),
-                    hex::encode(&backtrace.memory()[..backtrace.registers()[RegId::SP] as usize]), // print stack
+                    hex::encode(&backtrace.memory()[..usize::try_from(backtrace.registers()[RegId::SP]).expect("`$sp` can't be more than memory - `usize`")]), // print stack
```
MitchTurner, line 1503: This `expect` statement is confusing me. 

> memory - `usize`

Are we subtracting the `usize`? Do you mean to say

> memory (`usize`)

or

> memory, which is a `usize`

or something like that?

crates/client/src/client/pagination.rs
xgreenx: image

We are also trading the readability of the codebase. If we use `u32`, it means that all places from the screenshot will have something like `i32::try_from(results).expect("If the result is more than `i32::MAX` it is impossible to iterate with `PaginationRequest` type")`

MitchTurner: Right. Ideally you'd want to create an `Unsigned_i32` type that only can be constructed from a valid value. Or better a type that represents the domain case, rather than the implementation, e.g. `ResultCount`, because the caller doesn't care what it's using under the hood, they just want to construct a valid instance.

It all feels like a smell that it's represented as a `i32` at _any_ level though. Do you know why we're using `i32` under the hood and not `u32` or `u16` or something?

xgreenx: We do it because of the `async_graphql` API. But it seems under the hood still doesn't accept negative values... 

image

And casts it to `usize` later...=)

MitchTurner (approved):
Approved assuming we create an issue for the `i32` stuff.

Issue #1335, opened 2023-08-29
TOB-FUEL-42: Config values lack overflow checks


xgreenx:
## Description
The Fuel node has configurations for its chain and peer to peer that lack overflow checks on some parameters provided by node operators. Rather than allow a misconfigured node to continue to operate in the event of an overflow, invalid configurations should be validated upfront and rejected by using checked arithmetic.

#### Figure 42.1: Potential overflow in ChainConfig (audit-fuel/fuel-core/crates/chain-config/src/config/chain.rs#74–78)

```rust
impl Default for ChainConfig {
    fn default() -> Self {
        Self {
            chain_name: "local".into(),
            block_gas_limit: ConsensusParameters::DEFAULT.max_gas_per_tx * 10,
```

#### Figure 42.2: Potential overflow in P2P configuration (audit-fuel/fuel-core/crates/services/p2p/src/p2p_service.rs#154–171)

```rust
pub fn new(config: Config, codec: Codec) -> Self {
    let local_peer_id = PeerId::from(config.keypair.public());
    let gossipsub_data =
        GossipsubData::with_topics(GossipsubTopics::new(&config.network_name));
    let network_metadata = NetworkMetadata { gossipsub_data };
    // configure and build P2P Service
    let (transport, connection_state) = build_transport(&config);
    let behaviour = FuelBehaviour::new(&config, codec.clone());
    let total_connections = {
        // Reserved nodes do not count against the configured peer input/output
limits.
};
    let total_peers =
        config.max_peers_connected + config.reserved_nodes.len() as u32;
    total_peers * config.max_connections_per_peer
```

## Exploit Scenario
A node operator uses invalid configurations that cause an overflow. Rather than receiving an error that these configurations are invalid, the node continues to run with unanticipated behavior.

## Recommendations
Short term, use checked arithmetic and validate that configurations are within reasonable bounds.
Long term, consider setting Clippy’s integer arithmetic lint to deny and use checked, saturating, or wrapping arimhetic explicitly (same as https://github.com/FuelLabs/fuel-core/issues/1334).

PR #1466, opened 2023-10-31
Handling overflows during arithmetic operations

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1335

Handling overflows during arithmetic operations by denying `clippy::arithmetic_side_effects`.

Also, it is the last issue that we plan to fix right now in the scope of the ToB audit report. So this PR closes https://github.com/FuelLabs/fuel-vm/issues/513.


crates/fuel-core/src/service/genesis.rs
```diff
@@ -176,7 +176,9 @@ fn init_coin_state(
                         .expect("Incorrect genesis transaction id byte length")
                     }),
                     coin.output_index.unwrap_or_else(|| {
-                        generated_output_index += 1;
+                        generated_output_index = generated_output_index
+                            .checked_add(1)
+                            .expect("We don't support so many UTXOs in the genesis");
```
Voxelot, line 181:
```suggestion
                            .expect("The maximum number of UTXOs in the genesis configuration has been exceeded.");
```

crates/services/p2p/src/service.rs
```diff
@@ -519,7 +522,7 @@ where
 
                         // Note: this range has already been check for
```
Voxelot, line 523:
```suggestion
                        // Note: this range has already been checked for
```

crates/services/txpool/src/containers/dependency.rs
```diff
@@ -424,7 +425,7 @@ impl Dependency {
                         *utxo_id,
                         CoinState {
                             is_spend_by: Some(tx.id() as TxId),
-                            depth: max_depth - 1,
+                            depth: max_depth.saturating_sub(1),
```
Voxelot, line 428: Should we just throw an error here since this function returns a result anyways?

Voxelot (approved):
minor nits but I won't hold up the review over them

crates/services/txpool/src/containers/dependency.rs
xgreenx: We could, but `max_depth` above is always more than zero, so it should be impossible to overflow. I thought about using `checked_sub + expect` here to highlight it, but decided to go with saturating because depth less than zero also doesn't make any sense.

But if you think that `checked_sub + expect` is better,  I can move to it=)

Voxelot (approved):

PR #1468, opened 2023-10-31
Bumping `fuel-vm` to `v0.40.0`

xgreenx:
The new version of `fuel-vm`:
- Adds a new `max_size` into `TxParameters` and breaks the old API.
- `max_inputs` and `max_outputs` are `u8` now.
- `max_witnesses` is `u32` now.
- Add two new fields into `GasCosts.
- Charges additional `min_fee` for each Signed/Predicate input and for `Create` transaction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ Description of the upcoming release here.
 
 ### Changed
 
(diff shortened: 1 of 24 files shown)


bvrooman (dismissed):
LGTM. 

We should also do a run of the benchmarks to get more accurate gas costs for `contract_root` and `state_root` in the near future.

xgreenx:
Yeah, we will when all changes related to the benchmark are merged.

Dentosal (approved):

PR #1469, opened 2023-11-02
Avoid storage caching in benchmarks

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1441

- Disabled cache for blocks in the table
- Removed usage of in memory transaction in benchmarks and replaced it with checkpoint
- Added bloom filter
- Added customization of the block cache

### DIFF
--- a/benches/benches/vm.rs
+++ b/benches/benches/vm.rs
@@ -13,7 +13,6 @@ use criterion::{
 
 use contract::*;
 use fuel_core_benches::*;
-use fuel_core_storage::transactional::Transaction;
 use fuel_core_types::fuel_asm::Instruction;
 use vm_set::*;
 
@@ -33,14 +32,12 @@ where
                 instruction,
                 diff,
             } = &mut i;
-            let original_db = vm.as_mut().database_mut().clone();
-            let mut db_txn = {
-                let db = vm.as_mut().database_mut();
-                let db_txn = db.transaction();
-                // update vm database in-place to use transaction
-                *db = db_txn.as_ref().clone();
-                db_txn
-            };
+            let checkpoint = vm
+                .as_mut()
+                .database_mut()
+                .checkpoint()
+                .expect("Should be able to create a checkpoint");
+            let original_db = core::mem::replace(vm.as_mut().database_mut(), checkpoint);
 
             let final_time;
             loop {
@@ -80,7 +77,6 @@ where
                 }
             }
 
-            db_txn.commit().unwrap();
             // restore original db
             *vm.as_mut().database_mut() = original_db;
             final_time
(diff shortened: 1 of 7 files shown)


benches/benches/vm.rs
```diff
@@ -33,14 +32,12 @@ where
...
-                // update vm database in-place to use transaction
-                *db = db_txn.as_ref().clone();
-                db_txn
-            };
+            let checkpoint = vm
+                .as_mut()
+                .database_mut()
+                .checkpoint()
+                .expect("Should be able to create a checkpoint");
+            let original_db = core::mem::replace(vm.as_mut().database_mut(), checkpoint);
```
Voxelot, line 40: Is this really representative of real world conditions though? Since we'll pretty much always be executing this opcode from within a db transaction when going through the block executor.

xgreenx: image

The idea is to avoid optimizations from `MemoryTransactionView` in the case of non-first interaction with the storage key. Using checkpoint instead of  `MemoryTransactionView`, we go to `self.data_source.get(key, column)` part directly.

But yeah, it is not real-world execution. Because inside of the block, we have `RocksDB -> MemoryTransactionView -> MemoryTransactionView`. We can create a special `BenchmarkMemoryTransactionView` for benchmarks that does all the same thing, except it always goes to `data_source`. But I think the overhead from `MemoryTransactionView` is much lesser than direct access to the checkpoint database. The proof is a significant increase in the execution time of the opcode for case with `1`.

image

benches/src/bin/collect.rs
```diff
@@ -633,6 +633,7 @@ fn linear_regression(x_y: Vec<(u64, u64)>) -> f64 {
 fn dependent_cost(name: &String, x_y: Vec<(u64, u64)>) -> (u64, u64) {
```
MitchTurner, line 633: `x_y` isn't a very helpful name. Would `points` work here? I know we have have a `Point` type below, but I think it would still be an improvement.

If the types are getting confusing you could even set the type as `impl Into<Point>` and create a `From<(u64, u64)>` :P. That's less important than the name.

crates/fuel-core/src/state.rs
```diff
@@ -87,6 +89,12 @@ pub enum WriteOperation {
 }
 
 pub trait TransactableStorage: BatchOperations + Debug + Send + Sync {
+    fn checkpoint(&self) -> DatabaseResult<Database> {
+        Err(DatabaseError::Other(anyhow::anyhow!(
```
MitchTurner, line 93: I think it would be better to add an error variant for `Unsupported` or whatever. I think we should get out of the habit of using `anyhow`.

Voxelot: Depends on the situation, in the case of checkpoints it's only used in benchmarking / test contexts and we don't do any special logic based on the error variant other than bailing from the benchmark.

https://www.lpalmieri.com/posts/error-handling-rust/#anyhow-or-thiserror

Voxelot (approved):

PR #1474, opened 2023-11-08
Chain state encoding

MujkicA:


crates/chain-config/src/config/codec.rs
```diff
@@ -0,0 +1,262 @@
...
+        elements: Vec<ContractState>,
+    ) -> anyhow::Result<()>;
+    fn write_contract_balance(
+        &mut self,
+        elements: Vec<ContractBalance>,
+    ) -> anyhow::Result<()>;
+    fn close(self: Box<Self>) -> anyhow::Result<()>;
+}
+
+pub type DynGroupDecoder<T> = Box<dyn GroupDecoder<T>>;
```
MitchTurner, line 54: Does this need to be dynamic?

MujkicA:
During the sync we agreed to focus reviews on the codec related changes found in 
`chain_config/src/config/codec` and `chain_config/src/config/codec.rs`

crates/chain-config/src/config/codec.rs
MujkicA: What decoder actually gets used (json/parquet) is decided at runtime based on the file extension of the configs. So we resort to trait object polymorphism here to make higher level methods agnostic of the format.

MitchTurner: > So a dyn dispatch has to happen somewhere since we don't know the type until the app is ran.

Couldn't you use an enum:
```rs
#[non_exhaustive]
pub enum ConcreteDecoder<T> {
    JSON(JSONDecoder<T>),
    Parquet(ParquetDecoder<T>),
    InMemory(InMemoryDecoder<T>),
}

impl<T> GroupDecoder<T> for ConcreteDecoder<T> {
    ...
}
```

> There might be way to make it non-dynamic via an enum and lot's of if/match statements but it would affect proper separation and make it messier overall.

This feels like the idiomatic way to me. I have a general aversion to `dyn Trait` though :P Almost never use it.

segfault-magnet: Decided your suggestion a try. Turned out great in the end IMO. Lemme know your thoughts when you get the time.

MitchTurner: Nice. On a quick glance I'm not sure I follow the use of the `Iterator`, it seems like it could be valuable to capture the behavior in domain terms, rather than implementation terms. But I'm gonna give it a read in the full context and I'm sure it'll make more sense--I don't think I have a full picture yet.

bin/fuel-core/src/cli/snapshot.rs
```diff
@@ -23,17 +38,27 @@ pub struct Command {
...
+    Parquet,
+}
+
 #[derive(Debug, Clone, Subcommand)]
 pub enum SubCommands {
     /// Creates a snapshot of the entire database and produces a chain config.
     #[command(arg_required_else_help = true)]
     Everything {
-        /// Specify either an alias to a built-in configuration or filepath to a JSON file.
+        /// Specify a a path to the directory containing the chain config. Defaults used if no path
```
MitchTurner, line 52:
```suggestion
        /// Specify a path to the directory containing the chain config. Defaults used if no path
```

crates/chain-config/src/config/codec.rs
```diff
@@ -0,0 +1,151 @@
...
+    use itertools::Itertools;
+
+    use super::*;
+
+    #[test]
+    fn writes_then_reads_written() {
+        let group_size = 100;
+        let num_groups = 10;
+        let starting_group_index = 3;
+        {
```
MitchTurner, line 43: These could be separate tests for `json` and `parquet`. Easier to read.

Nit: maybe add `roundtrip` to the name too. I feel like that's the _expected_ name for tests like this--maybe not.

crates/chain-config/src/config/codec.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    fn test_write_read(
+        mut encoder: Encoder,
+        init_decoder: impl FnOnce() -> Decoder,
+        group_size: usize,
+        group_range: Range<usize>,
+    ) {
+        let num_groups = group_range.end;
+        let mut rng = rand::thread_rng();
```
MitchTurner, line 80: It's better to use seeded rng in tests:
https://docs.rs/rand/latest/rand/trait.SeedableRng.html

That way failures are reproducible. Or prop tests if you need to have coverage.

crates/chain-config/src/config/codec.rs
```diff
@@ -0,0 +1,151 @@
...
+        macro_rules! write_batches {
```
MitchTurner, line 81: I think this would be more readable as a generic function. Something like.

```rs
fn batches_for_type<T>(rng: &mut impl Rng, count: usize, group_size: usize) -> Vec<T>
   where
        Standard: Distribution<T>,
   {
   ::std::iter::repeat_with(|| rng.gen::<T>())
                    .chunks(group_size)
                    .into_iter()
                    .map(|chunk| chunk.collect_vec())
                    .enumerate()
                    .map(|(index, data)| Group { index, data })
                    .take(count)
                    .collect_vec()
}
```
(IDK if that quite compiles)

and then the test could just be:

```rs
#[test]
fn json_roundtrip_read_write_coins() {
    let mut rng = StdRng::seed_from_u64(42);
    let temp_dir = tempfile::tempdir().unwrap();
    
    // given
    let group_size = 100;
    let num_groups = 10;
    let starting_group_index = 3;
    
    let encoder = Encoder::json(temp_dir.path());
    let decoder = Decoder::json(temp_dir.path(), group_size).unwrap();
    
    let coin_batches = batches_for_type<CoinConfig>(rng, num_groups, group_size);
    
    // when
    encoder.write_coins(coin_batches);
    let read = decoder.coins().unwrap();
    
    // then
    let actual = read.skip(starting_group_index).collect::<Result<Vec<_>, _>>().unwrap();
    let expected = original[starting_group_index..];

    pretty_assertions::assert_eq!(actual, expected);
}
```
Then the test just reads smoothly top-to-bottom. You don't need to look at how everything is working for different tests to understand this one test. Sure, it's less DRY, but much easier to grok quickly.

crates/metrics/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 #![deny(clippy::arithmetic_side_effects)]
 #![deny(clippy::cast_possible_truncation)]
 #![deny(unused_crate_dependencies)]
-#![deny(warnings)]
```
MitchTurner, line 4: Don't forget this

crates/services/consensus_module/poa/src/lib.rs
```diff
@@ -2,7 +2,7 @@
 #![deny(clippy::cast_possible_truncation)]
 #![deny(unused_crate_dependencies)]
 #![deny(unused_must_use)]
-#![deny(warnings)]
```
MitchTurner, line 5: And this

crates/metrics/src/lib.rs
MitchTurner: Looks like there are a lot of them. So I won't note them all :P

crates/chain-config/src/config/codec.rs
segfault-magnet: Added it. The only implementations inside the `rand` crate had this warning:
> The algorithm is deterministic but should not be considered reproducible due to dependence on configuration and possible replacement in future library versions. For a secure reproducible generator, we recommend use of the [rand_chacha](https://crates.io/crates/rand_chacha) crate directly.

I'm fine with this level of reproducibility. If you deem the considerations above important enough to drag in another dependency please reopen this thread :)

crates/chain-config/src/config/codec.rs
segfault-magnet: Implemented something very similar to the suggestion. Let me know if you like it :)

crates/chain-config/src/config/codec.rs
MitchTurner: This is fine. I've used seedable many times and never had any trouble. It's just a precaution anyway.

crates/chain-config/src/config/codec.rs
MitchTurner: We've really been pushing for the "given/when/then" format for tests.

 It doesn't have to be exactly my suggestion either. I am curious if you had any specific issues with what I wrote.

MitchTurner:
I haven't finished my review of this, btw. Just got part way down from the top. I'll keep returning to it when I have chances.

crates/chain-config/src/config/codec.rs
segfault-magnet: > I am curious if you had any specific issues with what I wrote.

Nope. When I looked at your suggestion I made the following mental notes:
1. Ok, dislikes macros for readability. DRY what you can without them.
2. Dislikes testing both formats at once. Separating them increases readability.

I didn't even notice the given/when/then suggestion :) 

My only gripe is I dislike two "events" for the "when" phase. So I'll split these tests up into one for each entity and have the encoding be part of the "given" phase -- as in "given something was encoded like so when you decode then... "

> I'm also a fan of tests only testing a single thing
Me as well. The reason why I didn't do it like that is parquet is not human-readable when encoded. Rather than make a test whose assertions/setup are a binary blob, I went with the roundtrip option.

MitchTurner: >  DRY what you can without them.
DRY is one of my lowest priorities when I code. Especially with tests. It's a nice-to-have, but never at the expense of clarity.

>My only gripe is I dislike two "events" for the "when" phase. So I'll split these tests up into one for each entity and have the encoding be part of the "given" phase -- as in "given something was encoded like so when you decode then... "

Yep. That's what I meant when I said "I'm also a fan of tests only testing a single thing--roundtrip is pushing it already IMO." Roundtrip is kinda an exception to the rule, but I'm never going to complain about splitting it up :).

MitchTurner: Oh. I see. You kept the roundtrip but only one of the actions is in the `when`.

Is writing not also the subject of these tests?

segfault-magnet: Oh, never mind, I forgot what the tests look like. 

I'll group the actions under `when`. Makes sense since the test is 'roundtrip` after all.

crates/chain-config/Cargo.toml
```diff
@@ -14,9 +14,12 @@ description = "Fuel Chain config types"
 anyhow = { workspace = true }
 bech32 = { version = "0.9.0", optional = true }
 fuel-core-storage = { workspace = true }
-fuel-core-types = { workspace = true, default-features = false, features = ["serde"] }
+fuel-core-types = { workspace = true, default-features = false, features = [
+  "serde",
+] }
 hex = { version = "0.4", features = ["serde"] }
 itertools = { workspace = true, optional = true }
+parquet = "49.0"
```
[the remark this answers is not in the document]
xgreenx: Hmm, `cargo check -p fuel-core-chain-config --target wasm32-unknown-unknown --no-default-features` fails for me

crates/chain-config/src/config/codec/decoder.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+impl Decoder {
+    pub fn json(
+        snapshot_dir: impl AsRef<Path>,
+        group_size: usize,
+    ) -> anyhow::Result<Self> {
+        let path = snapshot_dir.as_ref().join("state.json");
+
+        // This is a workaround until the Deserialize implementation is fixed to not require a
```
xgreenx, line 74: It is fixed as part of the [v0.43.1](https://github.com/FuelLabs/fuel-vm/releases/tag/v0.43.1)

crates/chain-config/src/config/codec/decoder.rs
```diff
@@ -0,0 +1,170 @@
...
+pub struct Decoder {
+    data_source: DataSource,
+}
+
+impl Decoder {
+    pub fn json(
+        snapshot_dir: impl AsRef<Path>,
+        group_size: usize,
+    ) -> anyhow::Result<Self> {
+        let path = snapshot_dir.as_ref().join("state.json");
```
xgreenx, line 72: A magic constant, maybe instead of directory we can expect a path to he file from the user

crates/chain-config/src/config/codec/decoder.rs
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+
+    pub fn detect_encoding(
+        snapshot_dir: impl AsRef<Path>,
+        default_group_size: usize,
+    ) -> anyhow::Result<Self> {
+        let snapshot_dir = snapshot_dir.as_ref();
+
+        if snapshot_dir.join("state.json").exists() {
```
xgreenx, line 105: Hmm, it is not an ideal way of doing that=) Do we need this method?

crates/chain-config/src/config/codec/encoder.rs
```diff
@@ -0,0 +1,304 @@
...
+            encoder: EncoderType::Json {
+                state_file_path: snapshot_dir.as_ref().join("state.json"),
+                buffer: StateConfig::default(),
+            },
+        }
+    }
+
+    pub fn parquet(
+        snapshot_dir: impl AsRef<Path>,
+        compression_level: u8,
```
xgreenx, line 58: Maybe it is better to use enum?

crates/chain-config/src/config/codec/encoder.rs
```diff
@@ -0,0 +1,304 @@
...
+        where
+            T: Schema,
+        {
+            let path = path.join(format!("{name}.parquet"));
+            let file = std::fs::File::create(path)?;
+            parquet::Encoder::new(file, compression)
+        }
+
+        let path = snapshot_dir.as_ref();
+        let compression =
```
xgreenx, line 74: Do we need to know `compression` during decoding?

crates/chain-config/src/config/codec/encoder.rs
```diff
@@ -0,0 +1,304 @@
...
+                messages,
+                contracts,
+                contract_state,
+                contract_balance,
+            } => {
+                coins.close()?;
+                messages.close()?;
+                contracts.close()?;
+                contract_state.close()?;
+                contract_balance.close()?;
```
xgreenx, line 169: What kind of error we can get here? If closing `coins` fails, how does it affect other encoders?

crates/storage/src/lib.rs
```diff
@@ -10,6 +10,8 @@
 #![deny(missing_docs)]
 #![deny(warnings)]
 
+use std::array::TryFromSliceError;
```
xgreenx, line 13:
```suggestion
use core::array::TryFromSliceError;
```

crates/fuel-core/src/database/contracts.rs
```diff
@@ -149,6 +153,42 @@ impl StorageWrite<ContractsRawCode> for Database {
...
+                let value = res.1;
+
+                Ok(ContractStateConfig {
+                    contract_id,
+                    key,
+                    value,
+                })
+            })
+    }
+    pub fn iter_contract_balance_configs(
```
xgreenx, line 173:
```suggestion
    }

    pub fn iter_contract_balance_configs(
```

bin/fuel-core/src/cli/run.rs
```diff
@@ -256,6 +258,8 @@ impl Command {
             }
         };
 
+        let state_decoder = Decoder::in_memory(state_config.clone(), 1);
```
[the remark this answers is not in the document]
segfault-magnet: > What is 1 here?

In-memory decoders don't have premade groups. This sets the group size to `1` . 

But I guess the question is "why is this hardcoded" :D I'll give it a name somewhere.

> Also, maybe we don't want to load the whole state here. Instead, maybe we want to load it during genesis block generation?

Giving just a path here will cause friction with tests. There are tests where a `Config` is given some handmade state. If we only accept a path here then it cannot be in-memory.

Then we'd have to do something like `Either<PathBuf, StateConfig>`. Or make the `Config` generic and accept a: 
```rust
state_reader: impl Fn() -> anyhow::Result<StateReader>
```
The state config (in `master`) is already loaded into `Config` before genesis and stays there afterward. Don't know about it causing issues like cloning overhead?

bin/fuel-core/src/cli/run.rs
```diff
@@ -304,13 +308,18 @@ impl Command {
...
+
         let config = Config {
             addr,
             api_request_timeout: api_request_timeout.into(),
-            max_database_cache_size,
-            database_path,
-            database_type,
+            db_config,
             chain_config: chain_conf.clone(),
+            snapshot_decoder: state_decoder,
```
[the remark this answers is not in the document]
segfault-magnet: Answered [here](https://github.com/FuelLabs/fuel-core/pull/1474#discussion_r1434262689).

bin/fuel-core/src/cli/snapshot.rs
```diff
@@ -53,46 +78,110 @@ pub async fn exec(command: Command) -> anyhow::Result<()> {
...
+        data: impl Iterator<Item = StorageResult<T>>,
+        group_size: usize,
+        mut write: impl FnMut(Vec<T>) -> anyhow::Result<()>,
+    ) -> anyhow::Result<()> {
+        data.chunks(group_size)
+            .into_iter()
+            .try_for_each(|chunk| write(chunk.try_collect()?))
+    }
+
+    let group_size = 1000;
```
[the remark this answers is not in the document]
segfault-magnet: For `parquet` the group size gets baked into each group so that during decoding we don't need to know the group size.

For `json` we decided against encoding group sizes so we'd have a simpler schema for users (also fuel-core tests and the SDKs). When decoding `json` you need to specify the group sizes and they will be created on the fly.

I'll add the group size as a CLI parameter. Ideally, make it available only if the encoding is `parquet`.

xgreenx:
@MujkicA Could you solve conflicts, please?=)

crates/chain-config/src/config/codec/encoder.rs
segfault-magnet: Enumerated all twelve valid values.

bin/fuel-core/src/cli/snapshot.rs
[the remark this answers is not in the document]
segfault-magnet: Added the group size as well as the encoding to the CLI.

crates/chain-config/src/config/codec/encoder.rs
segfault-magnet: It is read automatically from the metadata inside the parquet files.

crates/chain-config/src/config/codec/encoder.rs
segfault-magnet: If `coins` fails nothing else will get closed. 

Closing a parquet file is to flush any buffers and to write the metadata at the end of the file.

Seeing as a snapshot is only complete if all of the parts are properly written, didn't see a need to even try to close other files if one fails.

I'll probably go ahead and parallelize this file writing rn. If I succeed then I will have to handle each failure separately and not short-circuit them.

crates/chain-config/src/config/codec/decoder.rs
segfault-magnet: We can either ask the user to specify the encoding while reading the snapshot or we detect it for them. This is the latter approach.

I opted for detecting since a snapshot is now a directory and we're enforcing naming rules for the files within. If we have those expectations then detection can be robust.

Otherwise the CLI will get messy:

```
core run --regenesis --json dir/state.json
```
```
core run --regenesis --parquet --coins dir/coins.parquet --messages dir/messages.parquet ....
```

crates/chain-config/src/config/codec/decoder.rs
segfault-magnet: Shared my thoughts [here](https://github.com/FuelLabs/fuel-core/pull/1474/files#r1435846137).

bin/fuel-core/Cargo.toml
```diff
@@ -34,9 +36,9 @@ tikv-jemallocator = { workspace = true }
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true, features = [
-    "ansi",
-    "env-filter",
-    "json",
+  "ansi",
```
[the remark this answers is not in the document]
segfault-magnet: It's the autoformatter for toml/cargo on my setup. I'll leave this comment unresolved. When this becomes a blocker for merging I'll manually ident it and merge.

crates/chain-config/src/config/codec/parquet/encode.rs
```diff
@@ -0,0 +1,466 @@
...
+        &self,
+        index: usize,
+        column: &mut SerializedColumnWriter<'_>,
+    ) -> anyhow::Result<()>;
+}
+
+impl ColumnEncoder for Vec<ContractConfig> {
+    type ElementT = ContractConfig;
+
+    fn encode_column(
```
xgreenx, line 106: Do we really need to encode it field by field? Maybe instead, we can simply encode the whole type into a one-byte array and store it? In this case, the code will be more permanent (because we don't need to clone additionally) and flexible in terms of support.

Do we use columns somehow?

crates/chain-config/src/config/state.rs
```diff
@@ -179,282 +200,64 @@ impl StateConfig {
...
-    }
-
-    #[test]
-    fn can_roundtrip_simple_contract() {
-        let config = config_contract();
-        let json = serde_json::to_string(&config).unwrap();
-        let deserialized_config: StateConfig =
-            serde_json::from_str(json.as_str()).unwrap();
-        assert_eq!(config, deserialized_config);
+impl<T> ChainStateDb for &T
```
xgreenx, line 227: You can use `#[impl_tools::autoimpl(for<T: trait> &T)]` to automate it

crates/fuel-core/src/service.rs
```diff
@@ -297,7 +276,7 @@ mod tests {
                 task.sub_services()[i].stop_and_await().await.unwrap();
                 assert!(!task.run(&mut watcher).await.unwrap());
             } else {
-                break
+                break;
```
xgreenx, line 279:
```suggestion
                break
```

crates/chain-config/src/config/codec/decoder.rs
segfault-magnet: I'll go with the metadata file route.

crates/chain-config/src/config/codec/decoder.rs
segfault-magnet: Sry, must have lost that commit somewhere. We have a constant for it and used in the state reading PR. 

I'll change the magic values in this PR as well and deal with the conflicts later.

crates/chain-config/src/config/codec/parquet/encode.rs
segfault-magnet: Implemented parquet encoding with Postcard and Canonical support. Lemme know how it looks.

crates/chain-config/src/config/codec.rs
```diff
@@ -0,0 +1,366 @@
...
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct Group<T> {
+    pub index: usize,
+    pub data: Vec<T>,
+}
+type GroupResult<T> = anyhow::Result<Group<T>>;
+
+#[cfg(all(feature = "std", feature = "random"))]
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
segfault-magnet: We'll make a note of it. Need to see what else needs changing to support recreating any table. I'll create a task to revisit this later when we've finished the initial regenesis scope.

bin/fuel-core/Cargo.toml
```diff
@@ -47,8 +49,19 @@ test-case = { workspace = true }
 default = ["env", "relayer", "rocksdb"]
 env = ["dep:dotenvy"]
 p2p = ["fuel-core/p2p", "const_format"]
-relayer = ["fuel-core/relayer", "dep:url", "dep:serde_json"]
-rocksdb = ["fuel-core/rocksdb"]
-rocksdb-production = ["fuel-core/rocksdb-production"]
+relayer = ["fuel-core/relayer", "dep:url"]
+parquet = ["fuel-core-chain-config/parquet"]
+rocksdb = [
```
xgreenx, line 54: BTW, if you want, you can create a separate PR on the `master` to use your formatting for the toml files. But you also need to add a CI action that will verify that formating is the same for everyone=)

crates/fuel-core/src/service/genesis.rs
```diff
@@ -141,160 +198,180 @@ fn import_genesis_block(
...
+        .is_some()
+    {
+        return Err(anyhow!("Contract utxo should not exist"));
     }
-    Ok(contracts_tree.root())
+    Ok(())
 }
 
-fn init_contract_state(
+fn _init_contract_state(
```
xgreenx, line 374: What does `_` mean here? Do you plan to rework it in the future?

crates/fuel-core/src/service/genesis.rs
```diff
@@ -141,160 +198,180 @@ fn import_genesis_block(
...
+                utxo_id,
+                tx_pointer,
+            },
+        )?
+        .is_some()
+    {
+        return Err(anyhow!("Contract utxo should not exist"));
     }
-    Ok(contracts_tree.root())
+    Ok(())
```
xgreenx, line 371: Do you plan to return hash of the whole contract?

crates/fuel-core/src/service/genesis.rs
```diff
@@ -4,11 +4,13 @@ use crate::{
 };
 use anyhow::anyhow;
 use fuel_core_chain_config::{
```
xgreenx, line 6: This file seems unfinished=) Do we need to review it, or do you plan to finish it in another PR?

bin/fuel-core/src/cli.rs
```diff
@@ -55,7 +57,7 @@ fn init_environment() -> Option<PathBuf> {
     None
 }
 
-pub async fn init_logging() -> anyhow::Result<()> {
+pub fn init_logging() {
```
MitchTurner, line 60: Why not return the errors rather than `expect`ing them?

I realize that it's not the end of the world if there is a panic right on startup, but curious since it's there.

bin/fuel-core/src/cli.rs
```diff
@@ -127,3 +129,417 @@ pub async fn run_cli() -> anyhow::Result<()> {
         }
     }
 }
+
+#[cfg(any(feature = "rocksdb", feature = "rocksdb-production"))]
+#[cfg(test)]
+mod tests {
```
MitchTurner, line 135: These are some stellar tests. I've never really tested against the cli commands directly before, but I like how this makes very living docs of how it works :muscle:

bin/fuel-core/src/cli/run.rs
```diff
@@ -250,11 +256,13 @@ impl Command {
         let (chain_conf, state_config) = match genesis_config.as_deref() {
             None => (ChainConfig::local_testnet(), StateConfig::local_testnet()),
             Some(path) => {
-                let chain_conf = ChainConfig::load_from_directory(path)?;
-                let state_config = StateConfig::load_from_directory(path)?;
+                let metadata = SnapshotMetadata::read(path)?;
+                let chain_conf = ChainConfig::from_snapshot(&metadata)?;
```
MitchTurner, line 260: I think the name of the function would make more sense as `from_snapshot_metadata`.

bin/fuel-core/src/cli/snapshot.rs
```diff
@@ -44,55 +101,137 @@ pub enum SubCommands {
...
+
     Ok(())
 }
+
+fn load_chain_config(
+    chain_config: Option<PathBuf>,
+) -> Result<ChainConfig, anyhow::Error> {
+    let chain_config = match chain_config {
+        Some(file) => ChainConfig::load(file)?,
+        None => ChainConfig::local_testnet(),
```
MitchTurner, line 226: Idk if I like this implicit behavior. This method should only do the loading. The caller should handle mapping a `None`. 

I also think there should be a message emitted or some way to notify the caller that they are getting a `local_testnet` config.

crates/chain-config/src/config/chain.rs
```diff
@@ -64,23 +64,27 @@ impl ChainConfig {
...
+        serde_json::from_reader(&file).map_err(|e| {
             anyhow::Error::new(e).context(format!(
                 "an error occurred while loading the chain state file: {:?}",
                 path.to_str()
             ))
         })
     }
 
+    #[cfg(feature = "std")]
+    pub fn from_snapshot(snapshot_metadata: &SnapshotMetadata) -> anyhow::Result<Self> {
```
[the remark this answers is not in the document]
MujkicA: Thanks, I've opened #1615

crates/fuel-core/src/service/genesis.rs
MujkicA: The changes to `genesis.rs` are placeholders to appease the CI. They undergo heavy changes in https://github.com/FuelLabs/fuel-core/pull/1519 and can be ignored here.

crates/fuel-core/src/service/genesis.rs
MujkicA: See https://github.com/FuelLabs/fuel-core/pull/1474#discussion_r1463488334

crates/fuel-core/src/service/genesis.rs
MujkicA: See https://github.com/FuelLabs/fuel-core/pull/1474#discussion_r1463488334

bin/fuel-core/src/cli/run.rs
MujkicA: I'm a b it hesitant about `from_snapshot_metadata` since it makes it sound as if the config is loaded from the metadata file instead of the snapshot.
Unless that gives you second thoughts as well, I'm ready to go with `from_snapshot_metadata`

MitchTurner: Maybe I'm just misreading what's happening. It looks like you're loading the `SnapshotMetadata` and passing it to the `from_snapshot` method. i.e. you're passing the metadata, not the snapshot to load from. 

Just seems like a mismatch of names. Are you really loading from the snapshot, not the metadata?

bin/e2e-test-client/tests/integration_tests.rs
```diff
@@ -97,15 +98,12 @@ async fn execute_suite(config_path: String) {
...
-        ChainConfig::load_from_directory("../../deployment/scripts/chainspec/dev")
-            .expect("Should be able to load chain config");
-    // TODO: don't use serde here
-    let state_config =
-        fs::read_to_string("../../deployment/scripts/chainspec/dev/state_config.json")
-            .expect("Should be able to read state config");
-    let state_config: StateConfig = serde_json::from_str(&state_config)
-        .expect("Should be able to decode state config");
+    let snapshot =
+        SnapshotMetadata::open("../../deployment/scripts/chainspec/dev").unwrap();
```
MitchTurner, line 102: The unwrap should prolly be an expect here too.

crates/chain-config/src/config/codec/decoder.rs
```diff
@@ -0,0 +1,191 @@
...
+            |state| &state.contract_balance,
+            #[cfg(feature = "parquet")]
+            |files| &files.contract_balance,
+        )
+    }
+
+    fn create_iterator<T: Clone>(
+        &self,
+        extractor: impl FnOnce(&StateConfig) -> &Vec<T>,
+        #[cfg(feature = "parquet")] file_picker: impl FnOnce(
```
MitchTurner, line 153: I think this should just be a different function for `feature = "parquet"`. Having a different function signature depending on the features sounds like a pitfall :hole:

MitchTurner: I guess it's not public so it's not the worst. It just feels a little clever. I'll downgrade this to a "Nit"

crates/chain-config/src/config/codec/encoder.rs
```diff
@@ -0,0 +1,537 @@
...
+    Level17,
+    Level18,
+    Level19,
+    Level20,
+    Level21,
+    Max,
+}
+
+#[cfg(feature = "parquet")]
+impl TryFrom<u8> for ZstdCompressionLevel {
```
MitchTurner, line 78: Do we want round-trip tests for these conversions?

crates/chain-config/src/config/codec/parquet.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    use crate::{
+        CoinConfig,
+        Group,
+    };
+
+    use super::*;
+
+    #[test]
+    fn can_skip_groups_without_reading_whole_file() {
```
MitchTurner, line 48: Nit: The tests should prolly be at the bottom of the `mod`

crates/chain-config/src/config/state.rs
```diff
@@ -52,61 +49,85 @@ pub const TESTNET_WALLET_SECRETS: [&str; 5] = [
...
-// TODO: do streaming deserialization to handle large state configs
-#[serde_as]
-#[skip_serializing_none]
 #[derive(Clone, Debug, Default, Deserialize, Serialize, Eq, PartialEq)]
 pub struct StateConfig {
     /// Spendable coins
-    pub coins: Option<Vec<CoinConfig>>,
-    /// Contract state
-    pub contracts: Option<Vec<ContractConfig>>,
+    pub coins: Vec<CoinConfig>,
```
MitchTurner, line 55: I'm a little confused with the nominclature. It looks like its a holdover from before, but I want to make sure I understand. Some of these fields are `Vec<SomeConfig>`, but then the docstring says "state".

To me config and state are different things--apps with the same configs could hold different state. If `Config` is the right term here, I'm confused what it would mean to have a vec of configs in general.

I think the names /docstrings/both need some clarity here.

crates/fuel-core/src/database/contracts.rs
```diff
@@ -149,6 +153,43 @@ impl StorageWrite<ContractsRawCode> for Database {
 }
 
 impl Database {
+    pub fn iter_contract_state_configs(
+        &self,
+    ) -> impl Iterator<Item = StorageResult<ContractStateConfig>> + '_ {
```
MitchTurner, line 158: Nit: shouldn't we know this type before compile time? It might be ugly and this might be cleaner, it just feels interesting hiding the type from caller this way.

crates/fuel-core/src/service/genesis.rs
```diff
@@ -69,33 +71,89 @@ pub fn maybe_initialize_state(
...
-    let mut database_transaction = Transactional::transaction(original_database);
+    let block_height = config.chain_config.height.unwrap_or_default();
+
+    let coins_reader = config.state_decoder.coins()?;
+    let mut _coin_roots =
+        import_coin_configs(&original_database, coins_reader, block_height)?;
+
+    _coin_roots.sort();
+
+    // TODO: other threads should be killed if one encounters a failure
```
MitchTurner, line 93: Is there a follow-up item for this `TODO`?

crates/fuel-core/src/service/genesis.rs
```diff
@@ -69,33 +71,89 @@ pub fn maybe_initialize_state(
...
+fn commit_genesis_block(
+    config: &Config,
+    original_database: &Database,
+) -> anyhow::Result<()> {
+    let mut database_transaction = Transactional::transaction(original_database);
     let database = database_transaction.as_mut();
-    // Initialize the chain id and height.
 
-    let height = config.chain_config.height;
+    // TODO: load roots from storage
```
MitchTurner, line 149: And this `TODO`?

MitchTurner (commented):
Looking really good. The tests warm my heart. Just left a few comments and nits.

crates/fuel-core/src/service/genesis.rs
MujkicA: These TODOs in `genesis.rs` are addressed in the follow up PR

bin/fuel-core/src/cli.rs
MujkicA: Props to @segfault-magnet

crates/chain-config/src/config/codec/decoder.rs
```diff
@@ -0,0 +1,191 @@
...
+    pub fn for_snapshot(
+        snapshot_metadata: crate::config::SnapshotMetadata,
+        default_group_size: usize,
+    ) -> anyhow::Result<Self> {
+        use crate::StateEncoding;
+
+        let decoder = match snapshot_metadata.take_state_encoding() {
+            StateEncoding::Json { filepath } => Self::json(filepath, default_group_size)?,
+            #[cfg(feature = "parquet")]
+            StateEncoding::Parquet { filepaths, .. } => Self::parquet(filepaths),
```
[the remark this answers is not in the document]
MujkicA: The decoder is able to deduce `compression` and `group_size` from metadata stored in the parquet file so we are safe

bin/fuel-core/src/cli.rs
MujkicA: Can't speak for the author but the expect messages seem more straightforward than e.g. printing `ParseBoolError` at some point.

crates/chain-config/src/config/state.rs
MujkicA: My understanding was these are parts of the initial state config, so if you were to use the same configs you would end up with the same initial state. 
I'm not sure what naming convention to go with to best clarify this, for now I just removed "state" from the docstring. If you have some ideas how to improve clarity I'm open for it.

crates/chain-config/src/config/codec/parquet/encode.rs
```diff
@@ -0,0 +1,122 @@
...
+    W: Write + Send,
+    E: Encode<T>,
+{
+    pub fn write(&mut self, elements: Vec<T>) -> anyhow::Result<()> {
+        let mut group = self.writer.next_row_group()?;
+        let mut column = group
+            .next_column()?
+            .ok_or_else(|| anyhow::anyhow!("Missing column. Check the schema!"))?;
+
+        let values: Vec<_> = elements
```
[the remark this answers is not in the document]
MujkicA: This might not play well with parquet. AFAIK it's not straightforward to go from a row which is supposed to be a single element to treating it like it's a collection. 
I'll set this to resolved to move forward with the feature unless you think it's worth it to investigate further.

crates/chain-config/src/config/state.rs
MitchTurner: Sounds good enough for now. I don't want to slow you down too much.

MitchTurner (approved):

PR #1475, opened 2023-11-08
Upgrade ed25519_dalek 1.0.1 -> 2.0.0

Dentosal:
Resolves https://github.com/FuelLabs/fuel-core/security/dependabot/29.


MitchTurner:
I talked to @Dentosal  about this off-line, but for the record:

This doesn't remove our dependency on the vulnerable version of `ed25519-dalek` completely since `libp2p` also depends on the bad version, as I mention here:
https://github.com/FuelLabs/fuel-core/issues/1298#issuecomment-1679739832

Here is the WIP changes for upgrading `libp2p`:
https://github.com/FuelLabs/fuel-core/pull/1379

That work has been put on the back burner in favor of beta-5 work.

Dentosal:
Exactly as @MitchTurner said. This only removes our direct dependency, so after this only the libp2p upgrade is needed.

MitchTurner (approved):

PR #1479, opened 2023-11-09
Update README.md - Fix broken link

MartinKong1990:
Old: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/struct.EnvFilter.html#examples

New: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ ulimit -n 10240
 
 #### Log level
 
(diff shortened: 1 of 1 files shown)


xgreenx:
Thank you, but we prefer to close this PR because the change is too small. If you wish, you can fix other links in readme file=)

PR #1487, opened 2023-11-15
#1486: remove the (seemingly) unnecessary borrow()

### LINKED ISSUE #1486: A warning (treated as error) when building using 1.73
```
error: call to `.borrow()` on a reference in this situation does nothing
   --> crates/fuel-core/src/database/block.rs:274:27
    |
274 |         let storage = self.borrow();
    |                           ^^^^^^^^^ help: remove this redundant call
    |
    = note: the type `database::Database` does not implement `Borrow`, so calling `borrow` on `&database::Database` copies the reference, which does not do anything and can be removed
note: the lint level is defined here
   --> crates/fuel-core/src/lib.rs:2:9
    |
2   | #![deny(warnings)]
    |         ^^^^^^^^
    = note: `#[deny(noop_method_call)]` implied by `#[deny(warnings)]`

error: could not compile `fuel-core` (lib) due to previous error
```

This `borrow` indeed seems to be redundant, [the archaeology](https://github.com/FuelLabs/fuel-core/pull/1071/files#diff-9c5a14fda963ea35dbb447b92c5cd59b5e8cc63c8a81cdde7697edbff5df98ecR274) gave no explanation either.

RGafiyatullin:
close #1486

### DIFF
--- a/crates/fuel-core/src/database/block.rs
+++ b/crates/fuel-core/src/database/block.rs
@@ -42,7 +42,6 @@ use fuel_core_types::{
 use itertools::Itertools;
 use std::{
     borrow::{
-        Borrow,
         BorrowMut,
         Cow,
     },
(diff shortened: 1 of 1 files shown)


xgreenx:
Thank you for you PR, but it was fixed before
image

PR #1489, opened 2023-11-15
Fix macos binary publishing

xgreenx:


.github/workflows/ci.yml
```diff
@@ -448,6 +448,7 @@ jobs:
         uses: docker/setup-buildx-action@v1
 
       - name: Log in to the ghcr.io registry
+        if: matrix.job.os == 'buildjet-4vcpu-ubuntu-2204'
```
Voxelot, line 451: This should also be added to the next login right? line 459-460

Voxelot (approved):

PR #1492, opened 2023-11-16
Support backward iteration in the RocksDB

xgreenx:
Reverting the change https://github.com/FuelLabs/fuel-core/pull/1004 and adds support for reverse prefix iteration in the RocksDB.


crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -503,6 +558,17 @@ impl TransactableStorage for RocksDb {
     }
 }
 
+/// The `None` means overflow, so there is not following prefix.
+fn next_prefix(mut prefix: Vec<u8>) -> Option<Vec<u8>> {
+    for byte in prefix.iter_mut().rev() {
+        if byte != &u8::MAX {
+            *byte = byte.checked_add(1).expect("We've just checked it above");
+            return Some(prefix)
+        }
```
MitchTurner, line 567:
```suggestion
        if let Some(new_byte) = byte.checked_add(1) {
            *byte = new_byte
            return Some(new_byte)
        }
```

crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -321,13 +321,68 @@ impl KeyValueStore for RocksDb {
...
-                let mut opts = ReadOptions::default();
-                opts.set_prefix_same_as_start(true);
+                // RocksDB prefix iteration doesn't support reverse order,
+                // but seeking the start key and iterating in reverse order works.
+                // So we can create a workaround. We need to find the next available
+                // element and use it as an anchor for reverse iteration,
+                // but skip the first element to jump on the previous prefix.
+                // If we can't find the next element, we are at the end of the list,
+                // so we can use `IteratorMode::End` to start reverse iteration.
+                if direction == IterDirection::Reverse {
```
MitchTurner, line 331: I feel like this section could be a little easier to read.

Something like:
```rs
match direction {
    IterDirection::Reverse => self.reverse_prefix_iter(prefix, ...),
    _ => self.prefix_iter(prefix, ...),
}
```

Then inside the new `reverse_prefix_iter`, instead of returning early, just do something like:
```rs
let maybe_next_item = next_prefix(prefix.to_vec())
    .and_then(|next_prefix| self
                            .iter_all(
                                column,
                                Some(next_prefix.as_slice()),
                                None,
                                IterDirection::Forward,
                            )
                            .next())
    .and_then(|res| res.ok());
if let Some(next_item) = maybe_next_item {
    ...
} else {
    ...
}
```
Of course, breaking those down into into smaller functions wouldn't hurt either, but those are the changes that would improve readability the most.

tests/tests/tx.rs
```diff
@@ -475,6 +464,99 @@ async fn get_transactions_by_owner_forward_and_backward_iterations() {
...
+        .unwrap();
+    let transactions_backward = response
+        .results
+        .into_iter()
+        .map(|tx| {
+            assert!(matches!(tx.status, TransactionStatus::Success { .. }));
+            tx.transaction
+        })
+        .collect_vec();
+    assert_eq!(transactions_backward.len(), 5);
```
bvrooman, line 498: This looks like you're testing that the request returns paginated data with the correct length. After this, you test that the request returns paginated data with the expected results. Lastly, you test that the request returns paginated data with the expected results when using a cursor. Therefore, I think this test could be broken up into two or three separate tests, since these are three separate concerns.

crates/fuel-core/src/state/rocks_db.rs
bvrooman: Alternatively,

```rust
/// The `None` means overflow, so there is not following prefix.
fn next_prefix(mut prefix: Vec<u8>) -> Option<Vec<u8>> {
    prefix.iter_mut().rev().find_map(|p| {
        *p = p.checked_add(1)?;
        Some(())
    })?;
    Some(prefix)
}
```

But I think @MitchTurner's suggestion is easiest to read.

MitchTurner: Yeah. I'm not a fan of early `return`, but I also don't love doing mutation with a `map`. So I think I prefer the succinct approach.

tests/tests/tx.rs
xgreenx: Splitting this text into three doesn't make much sense because the third test will be almost the same except for two asserts. It increases the test file size and makes it harder to navigate by adding a very slight improvement to the one test readability.

I agree that we need to try to test different things with different tests, but when one test is the copy of another + one new line, it is not very nice.

bvrooman: This is my opinion on testing, with a bias from working in TDD:

Each unit test should illustrate the conformance of a single, functional behaviour (a unit). The behaviour being tested should be described succinctly by the test name. The test suite creates a specification that outlines the complete behaviour of the subject under test, and this specification acts as a promise to the developers using the code: The test names tell us what it will do and the test body proves that it does it. A well-written test suite can serve as living documentation for the code. It clearly demonstrates the intended use cases and the expected behaviour of the software. E.g., `"backward_pagination_returns_correct_number_of_results"`, "`backward_pagination_returns_results_in_reverse_order`", `"backward_pagination_supports_cursor`", etc. 

Therefore, the goal isn't simply readability, and the fact that test code is duplicated is not really important. What's more important is that the test suite communicates the intended behaviour succinctly, and test names are the easiest way to achieve that. When I read the test mod, or run all the tests, I can simply look at the list of test names to have an overall sense of what the behaviours are.

If that resonates with you, feel free to update the test suite. If not, no worries - again, this is just my opinion.

xgreenx: https://github.com/FuelLabs/fuel-core/pull/1492/commits/ef6cfc6964bd486b267b72f9ac9b52041753149a

bvrooman (approved):

PR #1494, opened 2023-11-17
Adjusting of the prices benchmarks after sanity check benchmarks

xgreenx:
After running sanity check benchmarks, we've got inconsistent results for some opcodes. Part of that is because of the wrong configuration of the price benchmarks, some of them because of the wrong configuration of the sanity check benchmarks.

This PR addresses them and produces more consistent results.


benches/benches/block_target_gas_set/memory.rs
```diff
@@ -28,12 +28,7 @@ pub fn run_memory(group: &mut BenchmarkGroup<WallTime>) {
     run(
         "memory/aloc opcode",
         group,
-        [
-            op::movi(0x10, 1000),
-            op::aloc(0x10),
-            op::jmpb(RegId::ZERO, 0),
-        ]
-        .to_vec(),
+        [op::movi(0x10, 0), op::aloc(0x10), op::jmpb(RegId::ZERO, 0)].to_vec(),
```
MitchTurner, line 31: What's the difference between `0` and `1000` here?

benches/benches/vm.rs
```diff
@@ -38,8 +38,11 @@ where
             let mut total = core::time::Duration::ZERO;
             for _ in 0..iters {
                 let original_db = vm.as_mut().database_mut().clone();
-                let database_tx = original_db.transaction().as_ref().clone();
-                *vm.as_mut().database_mut() = database_tx;
+                // Simulates the block production/validation with three levels of database transaction.
```
MitchTurner, line 41: I'm not following this what this is for. These each seem to be converted to the same `DatabaseTransaction` type under the hood.

benches/benches/vm_set/alu.rs
```diff
@@ -37,7 +37,8 @@ pub fn run(c: &mut Criterion) {
     run_group_ref(
         &mut c.benchmark_group("aloc"),
         "aloc",
-        VmBench::new(op::aloc(0x10)),
+        VmBench::new(op::aloc(0x10))
+            .with_prepare_script(vec![op::movi(0x10, (1 << 18) - 1)]),
```
MitchTurner, line 41: The sanity check we are doing `0` and here we do `(1 << 18) - 1`. Is that okay that they are different?

benches/benches/vm_set/blockchain.rs
```diff
@@ -571,10 +598,19 @@ pub fn run(c: &mut Criterion) {
                 op::addi(0x15, 0x15, 1),
                 op::aloc(0x15),
                 op::move_(0x14, RegId::HP),
+                op::gtf_args(0x27, 0x00, GTFArgs::ScriptData),
+                op::addi(0x27, 0x27, ContractId::LEN.try_into().unwrap()),
+                op::addi(0x27, 0x27, WORD_SIZE.try_into().unwrap()),
+                op::addi(0x27, 0x27, WORD_SIZE.try_into().unwrap()),
+                op::addi(0x27, 0x27, AssetId::LEN.try_into().unwrap()),
```
bvrooman, line 605: Can you provide some context on why we want these extra op codes on post call?

It looks like you're setting register `0x27`, which is the `$rc` in the upcoming `srwq` opcode, to the script data pointer, then bumping it by a contact ID, a word, a word, and lastly, an asset ID. Is this some sort of additional setup for the `srwq` at the next iteration?

benches/benches/vm_set/mem.rs
```diff
@@ -164,13 +164,15 @@ pub fn run(c: &mut Criterion) {
     run_group_ref(
         &mut c.benchmark_group("pshh"),
         "pshh",
-        VmBench::new(op::pshh(full_mask)),
+        VmBench::new(op::pshh(full_mask))
+            .with_prepare_script(vec![op::pshh(full_mask); 10000]),
```
MitchTurner, line 168: Doesn't this affect the execution time?

benches/benches/block_target_gas_set/memory.rs
xgreenx: If you allocate `1000` you can run out of memory faster than run out of gas=) Because it doesn't make any difference to allocate `0` or `1000` for stability of the check, it is better to allocate zero.

benches/benches/vm.rs
xgreenx: It is converted to the same type, but inclusion depth increases with each `DatabaseTransaction`. 

During the execution of the block we have several layers of `DatabaseTransaction`.

When you call `get` on `DatabaseTransaction`, it goes to the internal database, which is `DatabaseTransaction`, that goes to another `DatabaseTransaction` that goes to `RocksDB`. 3 layers of `DatabaseTransaction` simulates the real execution environment complexity.

benches/benches/vm_set/alu.rs
xgreenx: With a sanity check, we may run out of memory before running out of gas. While in cost benchmarks we revert the state of VM after each iteration.

There is no difference in using `0` or `u18::MAX` with the current implementation, but we need this difference to potentially catch mismatching in sanity check benchmarks and cost benchmarks.

benches/benches/vm_set/blockchain.rs
xgreenx: The `VmBench::contract_using_db` already sets some script data inside, so we need to skip data set by the `contract_using_db` to get access to the data that we set inside of the benchmark in line 614.

image

image

benches/benches/vm_set/mem.rs
xgreenx: The same idea as with `aloc`. There is no difference in the current implementation, but maybe later, it can help us catch some mismatching in cost and results from sanity check.

benches/benches/vm.rs
MitchTurner: Hmm. Okay. This might be naive of me, but I feel like we should be benching against the surface _above_ where this happens, that way we don't have to simulate it. That might not be possible, and I'm not able to suggest a way to do it, so I'm not gonna push too hard on that.

benches/benches/block_target_gas.rs
```diff
@@ -251,7 +250,9 @@ fn run(
                     else {
                         panic!("The execution should fails with out of gas")
                     };
-                assert!(reason.contains("OutOfGas"));
+                if !reason.contains("OutOfGas") {
+                    panic!("The test failed because of {}", reason);
+                }
```
Dentosal, line 255: nit:
```suggestion
                assert!(reason.contains("OutOfGas", "The test failed because of {reason}");
```

Dentosal (approved):

Issue #1499, opened 2023-11-20
Sanity-check `gtf` opcodes


MitchTurner:
Following https://github.com/FuelLabs/fuel-core/issues/1386

Might not need to create a benchmark for every `gtf` opcode, but rather choose a good candidate for worst-case scenario.

Specifically, the opcodes that get metadata from inputs/outputs can take longer if there are more inputs/outputs. We might be able to choose a worst-case opcode and overcharge for the others since `gtf` isn't called many times in a single script.

xgreenx:
Closed by https://github.com/FuelLabs/fuel-core/pull/1503

PR #1503, opened 2023-11-21
Sanity-check `gtf` opcode

MitchTurner:
https://github.com/FuelLabs/fuel-core/issues/1499

vm benches:
```
gtf/gtf                
time:   [208.43 ns 208.99 ns 209.60 ns]
```

sanity benches:
```
block target estimation/other/gtf
time:   [65.877 ms 66.094 ms 66.323 ms]
```


benches/benches/block_target_gas.rs
```diff
@@ -345,6 +381,97 @@ fn service_with_contract_id(
...
+        .consensus_parameters
+        .fee_params
+        .gas_per_byte = 0;
+    config.utxo_validation = false;
+    config.block_production = Trigger::Instant;
+
+    let mut storage_key = primitive_types::U256::zero();
+    let mut key_bytes = Bytes32::zeroed();
+
+    for contract_id in contract_ids.iter() {
```
xgreenx, line 436: Initializing state and balances for all contracts will kill us=D It's okay to init only for `ContractId::zeroed()`.

benches/benches/block_target_gas.rs
```diff
@@ -345,6 +381,97 @@ fn service_with_contract_id(
     (service, rt)
 }
 
+fn service_with_many_contracts(
```
xgreenx, line 384: We have three almost similar functions to run benchmarks. Maybe we can somehow join them into one customizable?

benches/benches/block_target_gas_set/other.rs
```diff
@@ -37,5 +37,23 @@ pub fn run_other(group: &mut BenchmarkGroup<WallTime>) {
             .run(id, group, instructions, script_data.clone());
     }
 
-    // gtf: TODO: As part of parent issue (https://github.com/FuelLabs/fuel-core/issues/1386)
+    // gtf
+    {
+        let count = 254;
```
xgreenx, line 42: It would be nice if we updated the corresponding benchmark to get the right price for GTF=)

benches/benches/block_target_gas.rs
MitchTurner: How will it kill us? It works. I'm not doing full state like for the other tests--I can still reduce it some as well. But I need separate contract ids for each input/output. Each contract needs to exist in the DB as well.

xgreenx: 10M state for 254 contracts will take a lot of time to initialize=) We only need to init state for the `ContractId::zeroed()`. You can simply deploy other contracts

MitchTurner: Like I said,  I'm not doing the full 10M state for all of them:
```rs
    pub fn new_with_many_contracts(
        contract_ids: Vec<ContractId>,
    ) -> MultiContractBenchmarkRunnerBuilder {
        let state_size = 1000; // Arbitrary small state size
        let (service, rt) = service_with_many_contracts(state_size, contract_ids.clone());
        let rng = rand::rngs::StdRng::seed_from_u64(2322u64);
        MultiContractBenchmarkRunnerBuilder {
            service,
            rt,
            contract_ids,
            rng,
        }
    }
```

as opposed to what we're doing for the shared contract:
```rs
    pub fn new_shared(contract_id: ContractId) -> SharedSanityBenchmarkRunnerBuilder {
        let state_size = crate::utils::get_state_size();
        let (service, rt) = service_with_contract_id(state_size, contract_id);
        let rng = rand::rngs::StdRng::seed_from_u64(2322u64);
        SharedSanityBenchmarkRunnerBuilder {
            service,
            rt,
            contract_id,
            rng,
        }
    }
```

We could restructure to specify the state size for each contract though if we wanted. I don't think it's hurting any of our benchmarks significantly, currently though.

benches/benches/block_target_gas.rs
MitchTurner: Did a bunch of refactoring around the services and the `run` functions.

benches/benches/block_target_gas_set/other.rs
MitchTurner: Do you mean using `GTFArgs::InputContractOutputIndex` instead of `GTFArgs::ScriptData` in the `vm_set` benchmark? So they are testing the same argument for `gtf`?

xgreenx: Yes, because we need to use the price for the GTF worst case. The getter for script data is fast

MitchTurner:
For some reason the vm benches error with:
```
thread 'main' panicked at benches/benches/utils.rs:96:10:
Context should not be full: Panic(TooManyReceipts)
```
But it's not related to my new `gtf` opcode--I can comment out the other `run` calls and it goes away--so I don't think I should be worried about it.

benches/benches/block_target_gas_set/min_fee.rs
```diff
@@ -0,0 +1,5 @@
+use super::*;
```
xgreenx, line 1: I think we can exclude this file for now=)

xgreenx (approved):

PR #1506, opened 2023-11-21
Added validation of the coin's fields during block production and validation

xgreenx:
- Added validation of the coin's fields during block production and validation.
- Moved field comparison to `fuel-core-types` crates from `fuel-core-txpool`. Reused it in the executor.
- Added verification that contract exists. It helps to catch the error before hitting the VM.
- Included `Nonce`, `UtxoId`, `ContractId` into error message.


Dentosal (approved):
LGTM. Huge readability improvement as well.

PR #1508, opened 2023-11-23
Deny warnings only in CI and xtask

### LINKED ISSUE #1493: Remove `!#[deny(warnings)]` to improve DEX
`!#[deny(warnings)]` impedes quick prototyping, forcing all code to be without warnings for it to compile.

Also, it prevents Clippy from giving suggestions when dependencies fail to compile due to warnings.

Find a way of removing this, turning it into a CI step, **but** make sure that `cargo make` in the CI doesn't pass if any warnings are left.

Dentosal:
Closes #1493

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -22,6 +22,7 @@ skip_crate_env_info = true
 
 [env]
 CARGO_MAKE_EXTEND_WORKSPACE_MAKEFILE = true
+RUSTFLAGS = "-D warnings"
 
 [tasks.check]
 description = "Compile code without linking. Faster than build to detect problems."
--- a/bin/fuel-core-client/src/main.rs
+++ b/bin/fuel-core-client/src/main.rs
@@ -1,7 +1,6 @@
 #![deny(clippy::arithmetic_side_effects)]
 #![deny(clippy::cast_possible_truncation)]
 #![deny(unused_crate_dependencies)]
-#![deny(warnings)]
 
 use clap::Parser;
 use fuel_core_client::client::FuelClient;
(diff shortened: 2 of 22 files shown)


Makefile.toml
```diff
@@ -22,6 +22,7 @@ skip_crate_env_info = true
 
 [env]
 CARGO_MAKE_EXTEND_WORKSPACE_MAKEFILE = true
+RUSTFLAGS = "-D warnings"
```
xgreenx, line 25: Using `-D warnings` for `cargo make` while not using the same `RUSTFLAGS` for regular build invalidates the building cache.

You can run `./ci_checks.sh` twice a row, and you will see that it starts building from scratch.

xgreenx (changes requested):

xgreenx:
I still think it is better to use `!#[deny(warnings)]` on the crate level because it doesn't cause the rebuilding of all dependencies, plus it can be easily disabled without rebuilding all dependencies. It works in all cases and covers only the crate that we are interested in.

We can introduce a new feature "ignore-warning" for each crate and use `#![cfg_attr(not(feature = "ignore-warning"), deny(warnings))] `

MitchTurner:
I agree that the warnings are annoying sometimes and I also don't like modifying the attributes during development--I'd end up forgetting to replace them.

I feel like overriding the lints via your `cargo build` command is a good solution. I haven't figured out how to get it to work yet though. This doesn't seem to do the trick:
```
RUSTFLAGS="-W warnings" cargo build
```
So I'm not sure if it's possible...

Issue #1510, opened 2023-11-23
Remove usage of `iter_all_filtered` from `merkle_contract_state_range`


xgreenx:
Replace the usage of `iter_all_filtered` with sequential call of the `get` method.

<img width="1054" alt="image" src="https://github.com/FuelLabs/fuel-core/assets/18346821/990fb21f-9817-4f74-be54-7932d920d925">

Note: We need to update the price of `srwq` opcode then.

cc @Salka1988

PR #1512, opened 2023-11-23
Remove usage of iter_all_filtered from merkle_contract_state_range

Dentosal:
Closes https://github.com/FuelLabs/fuel-core/issues/1510

### DIFF
--- a/crates/fuel-core/src/database/vm_database.rs
+++ b/crates/fuel-core/src/database/vm_database.rs
@@ -227,47 +227,16 @@ impl InterpreterStorage for VmDatabase {
         start_key: &Bytes32,
         range: usize,
     ) -> Result<Vec<Option<Cow<Bytes32>>>, Self::DataError> {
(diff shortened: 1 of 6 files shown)


crates/fuel-core/src/database/vm_database.rs
```diff
@@ -227,47 +227,16 @@ impl InterpreterStorage for VmDatabase {
...
-        }
-
-        // Fill not initialized slots with `None`.
-        while results.len() < range {
-            results.push(None);
-            expected_key.increase()?;
+        let mut results = Vec::new();
+        for _ in 0..range {
+            key.to_big_endian(&mut multikey[32..]);
+            results.push(self.database.get(&multikey, Column::ContractsState)?);
```
xgreenx, line 237: Please use `storage::<ContractsState>().get` instead. In this case we can extract `VmDatabase` from `fuel-core`(we don't need to know about `Column`).

xgreenx (approved):

Issue #1209, opened 2023-06-13
Rework the handling of the `ChainConfig.initial_state` during re-genesis


xgreenx:
We must process a huge state if we need to re-genesis in the mainnet. It can be so huge that it can not even fit into the RAM. So, we need to re-work the initial state handling. 

The obvious solution is to stream the state from the file and process it with chunks. Because this process can be long, we also need to support interruption and continuing from the place where we've stopped. Also, the `initial_state` field in the `ChainConfig` should be replaced with the final result like genesis hash or genesis metadata instead of storing the whole state in the RAM.

xgreenx:
Based on the internal conversation, we agreed that the work can be paralyzed in several ways:

1. Extract the `initial_state` from the `ChainConfig`. We still need it during the start of the network, so this information should be provided along with `ChainConfig`. But it can be done as a separate file or as an optional field. It would be nice to integrate it with the loading snapshot functionality.
1. The current serde JSON format is hard to stream. It can be replaced with a custom or any already implemented format. The person who works on this part should compare existing solutions and decide which to select based on the integration complexity and performance.
1. The streaming itself should be pausable/resumable. It means we need to store a temporary, not finalized, database somewhere. When the snapshot loading is finished, we must make a temporary database our main database. Also, the temporary database may store some information about where we stopped. This information also should be cleaned up at the end.
1. The code requires adaptation to work with streamed data. It may require the introduction of some abstraction(mostly by traits) to support streaming. Also, we may support different formats in the future, so it would be nice if this abstraction is format agnostic.
1. During the initial benchmark, it becomes clear that BMT consumes a lot of memory for huge snapshots. The solution is to create an optimized version of the BMT that doesn't store all leaves and nodes. We can store only the nodes required to build the next nodes. It should drastically reduce memory consumption. The alternative solution is to use storage to store all nodes, but it is better to do this after optimized BMT.

xgreenx:
Also, maybe we need to think about migration. Do we want to make it part of the regenesis, or do we want to have a separate flow to migrate it?

PR #1519, opened 2023-11-27
Stream reader for chain state

MujkicA:
Relevant issue: #1209

Relevant previous PRs: #1474

This PR builds upon the previous work by introducing a resumable, parallelized (re)genesis process.

Starting with the run command, users provide a snapshot metadata file containing paths to the chain config file and files containing chain state items (such as coins, messages, contracts, contract states, and balances), which are loaded via streaming.

Each item group in the genesis process is handled by a separate worker, allowing for parallel loading. Workers stream file contents in batches, with the batch size adaptable for future performance optimization.

A database transaction is committed every time an item group is succesfully loaded. Resumability is achieved by recording the last loaded group index within the same db tx. If loading is aborted, the remaining workers are shutdown. Upon restart, workers resume from the last processed group.

Database tables used for the genesis process are cleared once it's finalized.

**Contract States and Balances**
The use of uniform-sized batches may result in batches containing items from multiple contracts. We expect that best performance is achieved by selecting a batch size that typically encompasses an entire contract's state or balance, allowing for immediate initialization of relevant Merkle trees through `nodes_from_set`.

**Note:** 
- The significant diff count primarily arises from configuring and updating the test contract with a large state.
- The most significant changes start from the `genesis.rs`, so it might be a good entry point for the review. Analysing how workers process the batches will guide you through the majority of the changes. Coin and message configs are simply loaded into the db, contract configs are loaded but the contracts tree root cannot be calculated until the contract states and balances workers are finished.
When inserting contract states and balances, we also have to init/update the corresponding trees.


MujkicA:
Note: make sure deployment tests are properly updated

crates/chain-config/src/config/contract_balance.rs
```diff
@@ -24,3 +27,5 @@ impl crate::Randomize for ContractBalance {
         }
     }
 }
+
+impl NonSkippingSerialize for ContractBalanceConfig {}
```
xgreenx, line 31: We need to remove this trait=) It super unclear what is its purpose

bin/e2e-test-client/src/tests/script.rs
```diff
@@ -70,28 +80,64 @@ pub async fn dry_run(ctx: &TestContext) -> Result<(), Failed> {
...
     let dry_run = include_bytes!("test_data/large_state/tx.json");
     let dry_run: Transaction = serde_json::from_slice(dry_run.as_ref())
         .expect("Should be able do decode the Transaction");
 
     // If the contract changed, you need to update the
     // `f4292fe50d21668e140636ab69c7d4b3d069f66eb9ef3da4b0a324409cc36b8c` in the
-    // `test_data/large_state/contract.json` together with:
+    // `test_data/large_state/state_config.json` together with:
     // 244, 41, 47, 229, 13, 33, 102, 142, 20, 6, 54, 171, 105, 199, 212, 179, 208, 105, 246, 110, 185, 239, 61, 164, 176, 163, 36, 64, 156, 195, 107, 140,
-    contract_config.calculate_contract_id();
```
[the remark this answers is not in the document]
MujkicA: `load_contract` already recomputes the id via `update_contract_id`

crates/chain-config/src/config/coin.rs
```diff
@@ -33,6 +42,40 @@ pub struct CoinConfig {
     pub asset_id: AssetId,
 }
 
+impl NonSkippingSerialize for CoinConfig {
```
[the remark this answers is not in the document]
segfault-magnet: Formats that don't include a schema cannot tolerate fields being skipped. Json can because from the encoded data it is clear that the field is missing. Postcard must read something for the optional field. Bincode as well. 

The problem is that the derived `Serialize` implementation skips over `Option<?>` fields if they are `None` so that we may have more friendly json encoding. But that causes the above described issue for `Postcard`. 

So you can either:
1. Don't skip the fields in the derived `Serialize`. This means that you need to manually encode the json if you want to retain the nice property of not having `null`s scattered about. 
2. Have the derived `Serialize` skip the fields but have another trait that, when implemented, gives a guarantee that the type either didn't skip a field or will provide a `Serialize` implementation that will not skip a field. 

I've tried option 1 and it turned out to be more verbose with all the .is_none checks.

Option 2 was easier to code. If a type's derived `Serialize` doesn't skip fields then we can just implement the thait `NonSkippingSerialize` that has a default implementation of the single method it declares.

If it does skip then it becomes a simple 'forward calls to .serialize for all fields of the struct'. No checks for null like option 1 warranted.

I've looked into the possibility of writing our own derive macro (Like `Bincode` did for this exact reason) but didn't know if this was enough of a painpoint to warrant a whole proc macro crate.

CHANGELOG.md
```diff
@@ -193,6 +193,7 @@ FuelVM received a lot of safety and stability improvements:
 - Refactored errors, returning more detailed errors to the user, simplifying debugging.
 
 ### Added
+- [#1519](https://github.com/FuelLabs/fuel-core/pull/1519): Stream chain state from file
```
MujkicA, line 196: The breaking changes can be added in the final feature branch to better summarise them

crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -484,6 +491,21 @@ where
             .write(batch)
             .map_err(|e| DatabaseError::Other(e.into()).into())
     }
+
+    // use delete_range to delete all keys in a column
+    fn delete_all(&self, column: Self::Column) -> StorageResult<()> {
```
xgreenx, line 496: Maybe we want to add a unit test to verify the behaviour of removing leys with different length.

crates/fuel-core/src/service/genesis/workers.rs
```diff
@@ -0,0 +1,289 @@
...
+        &mut self,
+        group: Vec<ContractBalanceConfig>,
+        tx: &mut Database,
+    ) -> anyhow::Result<()> {
+        tx.update_contract_balances(group)?;
+        Ok(())
+    }
+}
+
+impl ProcessState<ContractId> for Handler {
```
xgreenx, line 281: Maybe instead of implementing the `ProcessState` trait for `ContractId`, we could create a new type(a wrapper around `ContractId`) that will have the same meaning as `ContractBalanceConfig` or `ContractStateConfig>`. In this case it would be expected how `genesis_contract_ids_iter` function is used and it will contain it with this place

crates/fuel-core/src/service/genesis/runner.rs
```diff
@@ -0,0 +1,666 @@
...
+impl<T, K> ProcessStateGroup<K> for T
+where
+    T: ProcessState<K>,
+{
+    fn process_group(&mut self, item: Vec<K>, tx: &mut Database) -> anyhow::Result<()> {
+        item.into_iter().try_for_each(|item| self.process(item, tx))
+    }
+}
+
+pub trait HandlesGenesisResource {
```
xgreenx, line 56: Optional: I think the type that implements `ProcessState` and `HandlesGenesisResource` can be the same. It means we will create `Handler<T>` for the type `T`, and the handler will know how to process the type instead of the type itself.

xgreenx (commented):
Super puper cool change, thank you a lot!=) Left some optional comments but I think its already in a good shape to be merged into the feature branch=)

I think we need to merge it to the `master` as sooner as possible because it touches some ongoing areas

crates/fuel-core/src/database.rs
```diff
@@ -267,6 +283,18 @@ where
     ) -> StorageResult<()> {
         self.as_ref().batch_write(entries)
     }
+
+    fn delete_all(&self, column: Self::Column) -> StorageResult<()> {
```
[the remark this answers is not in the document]
MujkicA: This call should have been delegated to the inner types batch operation just as the write from above

crates/fuel-core/src/service/genesis/runner.rs
MujkicA: I've made the abstractions more straightforward at the cost of slight code duplication and a some more verbosity in the tests.

.github/workflows/ci.yml
```diff
@@ -57,6 +57,19 @@ jobs:
         run: cargo binstall --no-confirm cargo-sort
       - name: Run Cargo.toml sort check
         run: cargo sort -w --check
+      - name: Install Prettier and TOML Plugin
```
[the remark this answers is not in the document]
MujkicA: Done!

@segfault-magnet do you want to open up the new PR?

crates/fuel-core/src/service/genesis/workers.rs
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    fn genesis_resource() -> GenesisResource {
+        GenesisResource::Contracts
+    }
+}
+
+// wrapper for ContractId to be used in the ProcessState trait
+
+impl ProcessState for Handler<ContractId> {
```
xgreenx, line 257: Hmm, it seems part of the comments were missed after I pushed changes=(

We discussed on the call that it would be nice to use something else than `ContractId`. Maybe `ContractRootCalculatorConfig` or something with a similar idea to be along with another implementation of this trait for the config related only to the regenesis.

MujkicA: I've changed the implementation to use the already existing `ContractsInfo` table since it already contains the loaded contract ids. 
IIUC the problem with `ContractId` is that the the handler implementation for it comes as a surprise but it differs from the other *Config items in that it isn't read from a file.
I've added the `GenesisImportedContractId` alias to make it's purpose more obvious. What do you think?

xgreenx (approved):

Issue #649, opened 2022-09-27
Expose P2P network information from the P2P orchestrator 


bvrooman:
Getting currently connected peers:
- [ ] `get_peers()` method on network service returns the list of currently connected peers
- [ ] `Peers` schema in `fuel-core`
- [ ] GraphQL endpoint for retrieving `Peers` where the list of peers is populated by calling `get_peers()`  on network service
- [ ] Schema wrapper in `fuel-client`
- [ ] `get_peers()` method on client returns the list of peers populated by querying GraphQL for peers

bvrooman:
Blocked on ongoing P2P upgrades

Voxelot:
this should be unblocked now

xgreenx:
@leviathanbeak @bvrooman What is the status of this issue right now?

PR #1524, opened 2023-12-01
Show info about connected peers in the gql api

Voxelot:
closes: #649 

Show information about connected peers that may be useful for debugging purposes via the gql api.

Note: in the future it may make sense to protect this API with a special authentication token.


Voxelot:
todo:

- [ ] switch time_since_last_heartbeat to time_of_last_heartbeat
- [ ] throw runtime errors on the peers api if p2p is disabled

crates/fuel-core/src/service/adapters/graphql_api.rs
```diff
@@ -264,4 +273,58 @@ impl DryRunExecution for BlockProducerAdapter {
 
 impl BlockProducerPort for BlockProducerAdapter {}
 
-use super::BlockProducerAdapter;
+#[async_trait::async_trait]
+impl P2pPort for P2PAdapter {
+    async fn connected_peers(&self) -> anyhow::Result<Vec<PeerId>> {
```
xgreenx, line 278: Hmm, it seems we don't use it. Do we need it?

xgreenx (approved):

xgreenx (approved):

Issue #1535, opened 2023-12-11
Execution Error When Calling `bal` After Low-Level Call and Calling The Other Contract First


bitzoic:
The following describes a bug resulting in an execution error when checking the balance of a contract after performing a low-level call transferring some tokens and calling the other contract first.

In our test case, we have two contracts:

1. A dummy contract with a simple function:
```sway
contract;

abi DummyContract {
    fn dummy_function();
}

impl DummyContract for Contract {
    fn dummy_function() {
        let val = 0;
    }
}

```

2. A contract that performs a low level call and transfers some value and has a balance function:
```sway
contract;

use std::{
    bytes::Bytes,
    context::this_balance,
    low_level_call::{
        call_with_function_selector,
        CallParams,
    },
};

abi CallerContract {
    // Just performs a low level call
    #[storage(read, write)]
    fn make_low_level_call(
        target: ContractId,
        function_selector: Bytes,
        calldata: Bytes,
        single_value_type_arg: bool,
        asset_id: AssetId,
        amount: u64,
        gas: u64,
    );
    // Just returns the balance
    fn balance(asset_id: AssetId) -> u64;
}

impl CallerContract for Contract {
    #[storage(read, write)]
    fn make_low_level_call(
        target: ContractId,
        function_selector: Bytes,
        calldata: Bytes,
        single_value_type_arg: bool,
        asset_id: AssetId,
        amount: u64,
        gas: u64,
    ) {
        // Create the CallParams
        let call_params = CallParams {
            coins: amount,
            asset_id: asset_id,
            gas: gas,
        };
        // Perform the low level call
        call_with_function_selector(
            target,
            function_selector,
            calldata,
            single_value_type_arg,
            call_params,
        );
    }

    fn balance(asset_id: AssetId) -> u64 {
        // Returns the balance
        this_balance(asset_id)
    }
}
```

If we call `dummy_function()` first, then we call `make_low_level_call()`, and finally calling `balance()` will result in the following error:

```
failures:

---- test_low_level_transfer stdout ----
thread 'test_low_level_transfer' panicked at tests/harness.rs:100:78:
[...]

PR #1536, opened 2023-12-11
Fixed contracts tables to touch only own SMT nodes

xgreenx:
It was possible to remove the SMT nodes in the foreign contract because it was not prefixed with the `ContractId`. The caller could use the same storage key/asset ID with exactly the same values to delete the SMT nodes in another contract.

Fixes https://github.com/FuelLabs/fuel-core/issues/1535


Dentosal (dismissed):
Seems pretty straightforward, but maybe should wait for @bvrooman to review as well?

crates/fuel-core/src/database/balances.rs
```diff
@@ -438,6 +436,42 @@ mod tests {
         assert_eq!(root_0, root_2);
     }
 
+    #[test]
+    fn updating_foreign_contract_doesnt_affect_the_given_contract_insertion() {
```
bvrooman, line 440: Nit:
```suggestion
    fn updating_foreign_contract_does_not_affect_the_given_contract_insertion() {
```

crates/fuel-core/src/database/state.rs
```diff
@@ -433,6 +431,42 @@ mod tests {
         assert_eq!(root_0, root_2);
     }
 
+    #[test]
+    fn updating_foreign_contract_doesnt_affect_the_given_contract_insertion() {
```
bvrooman, line 435: Nit:
```suggestion
    fn updating_foreign_contract_does_not_affect_the_given_contract_insertion() {
```

bvrooman (approved):

bvrooman (approved):

PR #1542, opened 2023-12-12
Peer API fix for library users of fuel-core

Voxelot:
This PR restructures the changes from #1524 to fix an error that would block users from calling `client.chain_info()` without having the p2p feature enabled.

While throwing errors on apis that use disabled features is our current convention, it would mean that the sdk testing harness would fail when trying to get the consensus parameters unless it also enabled the p2p feature flag.

What changed in this PR:

- Information about peers now part of NodeInfo instead of ChainInfo. This seemed like a more suitable location, as chain info relates more towards global network settings and state. NodeInfo on the other hand shows details about the current node instance you've connected to.
- Peers are not returned in the default query for NodeInfo, and instead use a special query fragment that's used to only return info about peers. This is beneficial since it will allow users without p2p enabled to still fetch NodeInfo, and also reduce api bandwidth by only returning potentially bulky information about peers through a dedicated api request instead of by default.

*Note: This is a release blocker for 0.22 to unblock the sdk*


bvrooman (approved):
This appears to be a fairly straightforward migration of the peer info query from chain info to node info.

PR #1527, opened 2023-12-03
fix: typos in the docs

tudorpintea999:
Hello
I fixed some typos in the docs,
Hope it helps.
Br,
Tudor.

PR #1554, opened 2023-12-15
Implement codespell as a workflow

Frierened:
#1527 follow up

I have no admit it's my first time doing that, so please check

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -70,9 +70,9 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.6"
+version = "0.8.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91429305e9f0a25f6205c5b8e0d2db09e0708a7a6df0f42212bb56c32c8ac97a"
+checksum = "77c3a9648d43b9cd48db467b3f87fdd6e146bcc88ab0180006cef2179fe11d01"
 dependencies = [
  "cfg-if",
  "once_cell",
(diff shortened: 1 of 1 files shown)


.codespellrc
```diff
@@ -0,0 +1,4 @@
+[codespell]
+skip = *.rs,*.json,*.js,*.html
```
xgreenx, line 2: Why do we want to skip `rs` files?

xgreenx (commented):
Thank you for the PR! It's funny that PR is created by another person(not original one) with a private profile=)

Hmm, why it doesn't include [files](https://github.com/FuelLabs/fuel-core/pull/1527/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055) from the original PR?

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+
+    - name: Install codespell
+      run: pip install codespell
+
+    - name: Run codespell
+      run: codespell
```
xgreenx, line 28: It will not fail the CI if we have some spelling mistakes. Could you also apply suggestions from `codespell` to the whole codebase, please?

xgreenx (changes requested):

.github/workflows/codespell.yml
xgreenx: I meant to `fuel-core`. I tried to run `codespell` locally, and it seems to produce a lot of warnings.

xgreenx (approved):

.github/workflows/ci.yml
```diff
@@ -89,12 +89,32 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           slack_webhook: ${{ secrets.SLACK_WEBHOOK_NOTIFY_BUILD }}
 
+  spellcheck:
+    name: Run codespell
+    runs-on: ubuntu-latest
```
xgreenx, line 94:
```suggestion
    runs-on: buildjet-4vcpu-ubuntu-2204
```

xgreenx (dismissed):

xgreenx (dismissed):

xgreenx (commented):
Another thing to consider is that the original PR had more suggestions, so maybe we can use something more advanced to fix spelling.

xgreenx:
Since PR is not active, closing it for now. You are welcome to create a new one on top of the `master`=)

Issue #1548, opened 2023-12-14
Abstract database to work with raw bytes instead of types


xgreenx:
### Overview

With each Fuel node, the data from the blockchain is stored in a database. The tables on the database expect specific types,  so as the blockchain evolves and those types change, we need the database to accept updated and new types.

The most general way of doing this is to always store the data as raw bytes, rather than strong types, and as those types change shape they can just be serialized accordingly.

### Design Thoughts/Questions

- What happens if a network update requires new tables added to the db, not just changes in the shape stored on the tables?
- What is stored on db is rawbytes, but the interface for the DB should just be a `non_exhaustive` `enum` right? We don't want to be stringly typed at our interface (port).

MitchTurner:
Could the database types not just be an `non_exhaustive` `enum` instead of bytes? Maybe there's not a meaningful difference between that and what is proposed above since it all gets serialized anyway?

PR #1559, opened 2023-12-18
Remove duplicating logic in the `KeyValueStore` trait

xgreenx:
Preparation before start work on https://github.com/FuelLabs/fuel-core/issues/1548.

The `KeyValueStore` trait has some duplicated logic. This PR removes it, minimizing the number of methods that we need to implement. Also I applied the original ordering of the method as in the trait.

### DIFF
--- a/bin/fuel-core/src/cli/run.rs
+++ b/bin/fuel-core/src/cli/run.rs
@@ -425,16 +425,12 @@ async fn shutdown_signal() -> anyhow::Result<()> {
 
         let mut sigint =
             tokio::signal::unix::signal(tokio::signal::unix::SignalKind::interrupt())?;
-        loop {
-            tokio::select! {
-                _ = sigterm.recv() => {
-                    tracing::info!("sigterm received");
-                    break;
-                }
-                _ = sigint.recv() => {
-                    tracing::info!("sigint received");
-                    break;
-                }
+        tokio::select! {
+            _ = sigterm.recv() => {
+                tracing::info!("sigterm received");
+            }
+            _ = sigint.recv() => {
+                tracing::info!("sigint received");
             }
         }
     }
--- a/crates/fuel-core/src/database.rs
+++ b/crates/fuel-core/src/database.rs
@@ -56,6 +56,7 @@ type DatabaseResult<T> = Result<T>;
 // TODO: Extract `Database` and all belongs into `fuel-core-database`.
 #[cfg(feature = "rocksdb")]
 use crate::state::rocks_db::RocksDb;
+use crate::state::Value;
 #[cfg(feature = "rocksdb")]
 use std::path::Path;
 #[cfg(feature = "rocksdb")]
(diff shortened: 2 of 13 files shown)


bin/fuel-core/src/cli/run.rs
```diff
@@ -425,16 +425,12 @@ async fn shutdown_signal() -> anyhow::Result<()> {
...
-                _ = sigint.recv() => {
-                    tracing::info!("sigint received");
-                    break;
-                }
+        tokio::select! {
+            _ = sigterm.recv() => {
+                tracing::info!("sigterm received");
+            }
+            _ = sigint.recv() => {
+                tracing::info!("sigint received");
```
Dentosal, line 433: I don't think this change has anything to do with the `KeyValueStore` trait?

crates/fuel-core/src/state.rs
```diff
@@ -27,16 +27,9 @@ pub trait KeyValueStore {
...
-        &self,
-        key: &[u8],
-        column: Column,
-        buf: &[u8],
-    ) -> DatabaseResult<(usize, Option<Value>)>;
-
     fn take(&self, key: &[u8], column: Column) -> DatabaseResult<Option<Value>>;
 
-    fn delete(&self, key: &[u8], column: Column) -> DatabaseResult<Option<Value>>;
+    fn delete(&self, key: &[u8], column: Column) -> DatabaseResult<()>;
```
Dentosal, line 32: These should have doc comments explaining the behavior

Dentosal (changes requested):

bin/fuel-core/src/cli/run.rs
xgreenx: It is clippy requirement from a new rust=)

crates/fuel-core/src/database.rs
```diff
@@ -299,36 +327,19 @@ impl Database {
         self.data.batch_write(&mut set.into_iter())
     }
 
-    fn remove<V: DeserializeOwned>(
+    fn take<V: DeserializeOwned>(
```
Voxelot, line 330: did we rename this to take in case we'd want to have a separate remove fn that doesn't need to load the previous value?

xgreenx: Yes, for example in `batch_write`. Plus in the case of clear sequential opcode. Plus later we can optimise VM execution to use it too in some cases.
image

bin/fuel-core/src/cli/run.rs
Dentosal: Oh ok, then it's fine of course.

crates/fuel-core/src/state.rs
```diff
@@ -13,49 +13,98 @@ use std::{
...
         &self,
         key: &[u8],
-        column: Column,
+        column: Self::Column,
         value: Value,
-    ) -> DatabaseResult<Option<Value>>;
-
-    fn write(&self, key: &[u8], column: Column, buf: &[u8]) -> DatabaseResult<usize>;
+    ) -> DatabaseResult<Option<Value>> {
+        // FIXME: This is a race condition. We should use a transaction.
```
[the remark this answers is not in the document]
xgreenx: I moved it here from RocksDB implementation to remember that it really may hurt because it is not an atomic operation, and if someone modifies the database at the same time, the result can be wrong.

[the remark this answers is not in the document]
Dentosal: We should have an issue for that, so we'll remember to fix it.

Dentosal (approved):

PR #1565, opened 2023-12-19
Add `graph-node-fuel` support

Salka1988:


Voxelot:
this PR is huuuge lol - maybe we could split some of this out into a separate repo?

Salka1988:
> this PR is huuuge lol - maybe we could split some of this out into a separate repo?

Let us get this working and then we can see how to proceed. I think we will make a PR on the Graph node side, and then the this PR will be much smaller.

Dentosal:
Migrated to https://github.com/FuelLabs/fuel-subgraph

Issue #1350, opened 2023-09-06
Report Peers that give incorrect (not deserializable) data


MitchTurner:
This will require some error handling and possibly moving away from `anyhow` in our binary.

xgreenx:
I guess any error that we receive from the request/response protocol should decrease the reputation of the peer regardless of the reason.

xgreenx:
The current implementation ignores the error from the `convert_to_response` method.

<img width="1657" alt="image" src="https://github.com/FuelLabs/fuel-core/assets/18346821/4a431fbb-c815-4f84-a6a4-7e4e332a17af">

We need to handle it and, [in the same way](https://github.com/FuelLabs/fuel-core/issues/1345), send an error or `None` into the `channel`.

Issue #1345, opened 2023-09-05
Support timeouts on Request/Response protocol level


xgreenx:
## Problem description

Any error(it includes timeouts) that occurs during the request-response protocol is ignored. If we receive an error, the parts of the blockchain that await the response will be paralyzed. In the case of the `fuel-core-sync`, it may stuck the synchronization. In other cases, it can consume resources forever.

<img width="1658" alt="image" src="https://github.com/FuelLabs/fuel-core/assets/18346821/2ce750ef-43e1-482a-b294-608fe6a7c4f2">

## Solution

Instead of silently removing the channel related to `request_id` we need to send either an error or `None`. It allows other parts of the project to handle it and request it again(maybe from another peer).

## Implementation details

The approach with error requires modification of the response type. But it triggers cascade changes in the places where we started a request. It will allow us to check all related places and to be sure that we handled an error response correctly and at least created retry logic(or any other logic that suits us to be sure that the node is in the "fine" state after this request).

The main goal of this issue is to have timeouts for requests. We need to add an integration test that verifies that. Also, maybe we need to make a timeout threshold configurable(the default value could be 20 seconds).

PR #1573, opened 2023-12-22
Simplify p2p request/response message serialization

Dentosal:
For some reason we had a two layers of serialization for request/response messages. This doesn't seem useful at all, and complicates e.g. error handling. This PR removes the extra layer, substantially simplifying that logic. One major upside of this is that  #1345 and #1350 can now be solved in a single follow-up PR.

~~Hopefully this doesn't conflict too much with the ongoing libp2p update PR #1379.~~


Voxelot (approved):
nice cleanup 👍🏻 

One thing to think about is how we could version the encoding we use in case we want backwards compatibility even if we improve the encoding in the future. But that can be solved later when we start working on all the stories around protocol versioning.

crates/services/p2p/src/p2p_service.rs
```diff
@@ -567,54 +557,36 @@ impl<Codec: NetworkCodec> FuelP2PService<Codec> {
...
-                            if channel.send(*block).is_err() {
-                                tracing::error!(
-                                    "Failed to send through the channel for {:?}",
-                                    request_id
-                                );
-                            }
-                        }
+                            ResponseChannelItem::Block(channel),
+                            ResponseMessage::Block(block),
+                        ) => channel.send(block.map(|b| Arc::into_inner(b).expect("There are not other references, we just received this from the network"))).is_ok(),
```
xgreenx, line 570: Do we even need to use `Arc` for block headers and transactions? Maybe we can  work directly with types

xgreenx (approved):

xgreenx (approved):

PR #1574, opened 2023-12-23
Decrease peer reputation on request timeouts and decode errors 

### LINKED ISSUE #1345: Support timeouts on Request/Response protocol level

### LINKED ISSUE #1346: Subtract Peer Reputation for Request Timeout
Part of:
https://github.com/FuelLabs/fuel-core/issues/943

Peers should receive a penalty for not responding to requests.

### LINKED ISSUE #1350: Report Peers that give incorrect (not deserializable) data

Dentosal:
Closes #1345. Closes #1346 Closes #1350.

This PR stops discarding request errors from libp2p, and instead returns them to the sender of the request. Also penalizes peers for sending invalid responses or for not replying at all.

Making penalty configurable should be a follow-up PR, as there are other penalties that should be configurable as well

TODO:
- [x] Make timeout configutable: Already seems to be case on master branch
- [x] Add tests
- [x] Fix current tests that for some reason don't terminate


Dentosal:
Tests fail in CI due to a timeout. Cannot reproduce locally. Investigating.

crates/services/p2p/src/p2p_service.rs
```diff
@@ -96,15 +99,19 @@ pub struct FuelP2PService {
...
-    /// Once the ResponseMessage is received from the p2p Network
-    /// It will send it to the NetworkOrchestrator via its unique Sender    
-    outbound_requests_table: HashMap<OutboundRequestId, ResponseChannelItem>,
+    /// Holds active outbound requests and associated oneshot channels.
+    /// When we send a request to the p2p network, we add it here. The sender
+    /// must provide a channel to receive the response.
+    /// Whenever a response (or an error) is received from the p2p network,
+    /// the request is removed from this table, and the channel is used to
+    /// send the result to the caller.
+    outbound_requests_pending: HashMap<OutboundRequestId, TypedResponseChannel>,
```
MitchTurner, line 108: I like the descriptive "table". Curious why you changed the name of these fields.

crates/services/p2p/src/request_response/messages.rs
```diff
@@ -28,27 +34,43 @@ pub enum RequestMessage {
...
-    Block(Option<SealedBlock>),
+    Block(Option<Arc<SealedBlock>>),
     SealedHeaders(Option<Vec<SealedBlockHeader>>),
     Transactions(Option<Vec<Transactions>>),
 }
 
+pub type OnResponse<T> = oneshot::Sender<Result<T, ResponseError>>;
+
+#[derive(Debug)]
+pub enum TypedResponseChannel {
```
MitchTurner, line 47: These aren't channels. The old name `ResponseChannelItem` isn't the best. "Item" isn't a helpful word, but it's the Item that comes over the "response channel", so it's accurate.

`ChannelResponse`? I'd prefer something that describes what the channel is used for, rather than just saying it is used by a channel.

crates/services/p2p/src/p2p_service.rs
Dentosal: Reverted. I personally think "table" is redundant as the data type is already HashMap which is basically a "table". But then again the name is ok for networking contexts.

crates/services/p2p/src/request_response/messages.rs
Dentosal: Renaemd to `ResponseSender`

Dentosal:
One of the test cases seems flakey, looking into that one

crates/services/p2p/src/request_response/messages.rs
```diff
@@ -28,27 +34,38 @@ pub enum RequestMessage {
...
-    Block(oneshot::Sender<Option<SealedBlock>>),
-    SealedHeaders(oneshot::Sender<(PeerId, Option<Vec<SealedBlockHeader>>)>),
-    Transactions(oneshot::Sender<Option<Vec<Transactions>>>),
-}
-
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum ResponseMessage {
-    Block(Option<SealedBlock>),
+    Block(Option<Arc<SealedBlock>>),
```
MitchTurner, line 39: Do we really need these `Arc`s? There are a few here and I'm not sure what they're for. Possible it's something I haven't seen yet.

Dentosal: Not anymore, at least. Removed in c97533d3

crates/services/p2p/src/service.rs
```diff
@@ -717,10 +716,17 @@ impl SharedState {
...
             .await?;
 
-        receiver
-            .await
-            .map(|(peer_id, headers)| (peer_id.to_bytes(), headers))
-            .map_err(|e| anyhow!("{}", e))
+        let (peer_id, response) = receiver.await.map_err(|e| anyhow!("{e}"))?;
+
+        let Ok(data) = response else {
+            let _ = self.report_peer(
```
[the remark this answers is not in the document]
Dentosal: Removed in 8c5ab861

xgreenx (approved):

PR #1577, opened 2023-12-29
Moved insertion of the blocks into the `BlockImporter` instead of the executor

xgreenx:
Related to https://github.com/FuelLabs/fuel-core/issues/1549

We don't need to insert the block in the executor because it is the block importer's responsibility. It verifies the block's validity and decides whether we want to insert a new block or not. Plus, storing blocks and transactions is not part of the state transition.

This change also adds the ability to produce the block with a defined order of the transactions. It may be useful in the tests.

### DIFF
--- a/crates/fuel-core/src/service/adapters/block_importer.rs
+++ b/crates/fuel-core/src/service/adapters/block_importer.rs
@@ -130,11 +130,11 @@ impl ImporterDatabase for Database {
 }
 
 impl ExecutorDatabase for Database {
-    fn store_block(
+    fn store_new_block(
         &mut self,
         chain_id: &ChainId,
         block: &SealedBlock,
-    ) -> StorageResult<Option<()>> {
+    ) -> StorageResult<bool> {
         let block_id = block.entity.id();
         let mut found = self
             .storage::<FuelBlocks>()
@@ -152,7 +152,7 @@ impl ExecutorDatabase for Database {
                 .insert(&tx.id(chain_id), tx)?
                 .is_some();
         }
-        Ok(found.then_some(()))
+        Ok(!found)
     }
 }
 
--- a/crates/fuel-core/src/service/adapters/block_importer.rs
+++ b/crates/fuel-core/src/service/adapters/block_importer.rs
@@ -22,6 +22,7 @@ use fuel_core_storage::{
     tables::{
         FuelBlocks,
         SealedBlockConsensus,
+        Transactions,
     },
     transactional::StorageTransaction,
     Result as StorageResult,
@@ -29,18 +30,16 @@ use fuel_core_storage::{
 };
 use fuel_core_types::{
     blockchain::{
-        block::{
-            Block,
-            CompressedBlock,
-        },
+        block::Block,
         consensus::Consensus,
-        primitives::{
-            BlockId,
-            DaBlockHeight,
-        },
+        primitives::DaBlockHeight,
         SealedBlock,
     },
-    fuel_types::BlockHeight,
+    fuel_tx::UniqueIdentifier,
+    fuel_types::{
+        BlockHeight,
+        ChainId,
+    },
     services::executor::{
         ExecutionTypes,
         Result as ExecutorResult,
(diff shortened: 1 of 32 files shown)


crates/services/importer/src/importer.rs
```diff
@@ -228,17 +242,18 @@ where
...
             return Err(Error::InvalidDatabaseStateAfterExecution(
-                expected_next_height,
+                expected_height,
                 actual_height,
             ))
         }
 
         db_after_execution
-            .seal_block(&block_id, &result.sealed_block.consensus)?
+            .block(&self.chain_id, &result.sealed_block)?
```
Voxelot, line 256: `block` seems too generic - it's not clear from the name what this function actually does with the block.

```suggestion
            .store_block(&self.chain_id, &result.sealed_block)?
```

Voxelot:
Good idea! This should also help with parallelization in case we ran multiple executor instances across several threads.

Voxelot:
If we don't include blocks in any provable state transition logic, wouldn't this mean we couldn't prove that transactions or messages are included when they are supposed to be? This seems like it could be an issue for things like forced-transaction inclusion.

xgreenx:
> If we don't include blocks in any provable state transition logic, wouldn't this mean we couldn't prove that transactions or messages are included when they are supposed to be? This seems like it could be an issue for things like forced-transaction inclusion.

Regarding messages and transactions that come to L1: 
The state transition function produces a block, and we will prove that the block was generated according to the state transition rules. The header itself contains `transactions_root` and `messages_root`. These roots are our proof that transactions and messages were included in the block. Inserting them into storage proves nothing because the host function that inserts something into the storage can simply do nothing=)

Regarding messages that come from L1: 
If the transaction inside the blocks tries to use some message bridge from L1, the relayer should provide a message. If it does not exist, the execution will fail. If a message exists, L1 can verify it by looking into the message portal. If the message exists but is spent, anyone can provide proof of where it was spent.

Regarding transactions that come from L1: 
The state transition function can create receipts for each transaction that comes from L1 during block verification. The L1 contract can go over these receipts and check that all transactions were included.

crates/fuel-core/src/service/adapters/block_importer.rs
```diff
@@ -42,16 +48,20 @@ use fuel_core_types::{
...
-use super::MaybeRelayerAdapter;
+use super::{
+    MaybeRelayerAdapter,
+    TransactionsSource,
+};
 
 impl BlockImporterAdapter {
     pub fn new(
         config: Config,
+        chain_config: &ChainConfig,
```
MitchTurner, line 59: Maybe we can combine the configs into a `BlockImporterConfig`?

crates/services/importer/src/ports.rs
```diff
@@ -32,21 +35,22 @@ pub trait Executor: Send + Sync {
 /// The database port used by the block importer.
 pub trait ImporterDatabase {
     /// Returns the latest block height.
-    fn latest_block_height(&self) -> StorageResult<BlockHeight>;
+    fn latest_block_height(&self) -> StorageResult<Option<BlockHeight>>;
```
MitchTurner, line 38: I'm confused by the cardinality of this. In which scenario would there be `None` block height? Won't we always have at least a genesis block?

I'm guessing some change upstream prompted you to change this, but from the context of an interface, the types are confusing to the consumer for ^^^ reason.

xgreenx: The block importer is already responsible for committing the genesis block. Before, we expected that the underlying database would not be initialized while the database transaction would be. Now, we expect that both don't have a genesis block. The `Option` is better than the old way of checking the error.

image

MitchTurner: Okay. I see now. Then I agree.

crates/services/importer/src/ports.rs
```diff
@@ -32,21 +35,22 @@ pub trait Executor: Send + Sync {
...
+    // TODO: Remove `chain_id` from the signature, but for that transactions inside
+    //  the block should have `cached_id`. We need to guarantee that from the Rust-type system.
+    fn store_block(
         &mut self,
-        block_id: &BlockId,
-        consensus: &Consensus,
-    ) -> StorageResult<Option<Consensus>>;
+        chain_id: &ChainId,
+        block: &SealedBlock,
+    ) -> StorageResult<Option<()>>;
```
MitchTurner, line 53: Why not return `StorageResult<Bool>`?

MitchTurner: Also, could we have a self-documenting name? Why it's returning a value rather than `Result<()>` isn't self-evident. Maybe `store_block_if_new` or something like that.

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -91,11 +106,18 @@ impl fuel_core_poa::ports::BlockProducer for BlockProducerAdapter {
         &self,
         height: BlockHeight,
         block_time: Tai64,
+        txs: Option<Vec<Transaction>>,
```
MitchTurner, line 109: I think I might prefer a separate method for this. It's not explicit what is happening with those txs vs the txs included from txpool (for example it might be reasonable to believe that it will include the specified txs and fill the rest with some from the txpool, for example).

The `BlockProducer` could have a `produce_and_execute_block_for_specific_txs` and `produce_and_execute_block_with_txpool_txs` or something like that.

This change would also helps push the logic into our domain rather the adapter.

MitchTurner: An alternative would be specifying the tx source in the function call. IDK if I like that as much though.

MitchTurner (commented):
This is great. Left a couple more comments, but should be ready to approve after those get resolved!

MitchTurner (approved):

PR #1579, opened 2024-01-04
Extract off chain logic from the executor

xgreenx:


crates/fuel-core/src/service/adapters/block_importer.rs
```diff
@@ -72,11 +72,7 @@ impl BlockImporterAdapter {
         &self,
         sealed_block: SealedBlock,
     ) -> anyhow::Result<()> {
-        tokio::task::spawn_blocking({
-            let importer = self.block_importer.clone();
-            move || importer.execute_and_commit(sealed_block)
-        })
-        .await??;
+        self.block_importer.execute_and_commit(sealed_block).await?;
```
xgreenx, line 75: Now the importer uses `tokio_rayon` inside.

crates/services/importer/src/config.rs
```diff
@@ -22,7 +22,7 @@ impl Config {
 impl Default for Config {
     fn default() -> Self {
         Self {
-            max_block_notify_buffer: 1 << 10,
+            max_block_notify_buffer: 1,
```
Dentosal, line 25: Does this field still need to exist if we always notify on every block?

Dentosal (dismissed):
Lots of changes, but it seems right. Probably someone else should go through this as well.

crates/fuel-core/src/service/adapters/block_importer.rs
web3reality5: perfect

Voxelot (approved):

Dentosal (approved):

Issue #1589, opened 2024-01-08
Move the `DatabaseTransaction` logic into `StorageTransaction`


xgreenx:
## Problem overview

The current logic of the executor works with the `Transaction` trait that hides the implementation and how database transaction works. 

```rust
/// The type is storage transaction and holds uncommitted state.
pub trait Transaction<Storage: ?Sized>:
    AsRef<Storage> + AsMut<Storage> + Send + Sync
{
    /// Commits the pending state changes into the storage.
    fn commit(&mut self) -> StorageResult<()>;
}

/// The storage transaction for the `Storage` type.
pub struct StorageTransaction<Storage: ?Sized> {
    transaction: Box<dyn Transaction<Storage>>,
}

impl<Storage: ?Sized> Transaction<Storage> for StorageTransaction<Storage> {
    fn commit(&mut self) -> StorageResult<()> {
        self.transaction.commit()
    }
}
```

Because implementation is hidden, the executor can't be sure that `StorageMutate` methods don't update the underlying storage. It is crucial for the fraud proofs, and it will be useful to have control over that for forkless upgrades. 

## Solution

The solution is to move the logic of the `DatabaseTransaction` to the `fuel-core-storage` crate. But because it is still up to the database implementation of how the final commitment is done, we will add several requirements:

- The database may be only modified through the `commit_changes` method.
  ```rust
  #[impl_tools::autoimpl(for<T: trait> &mut T, Box<T>)]
  pub trait Modifiable {
      fn commit_changes(&mut self, changes: Changes) -> StorageResult<()>;
  }
  
  pub type Changes = HashMap<(u32, Vec<u8>), WriteOperation>;
  
  pub struct StorageTransaction<Storage> {
      changes: Changes,
      storage: Storage,
  }
  
  impl<Storage> StorageTransaction<Storage>
  where
      Storage: Modifiable,
  {
      pub fn commit(&mut self) -> StorageResult<()> {
          self.storage
              .commit_changes(core::mem::take(&mut self.changes))
      }
  }
  ```
- The `StorageMutate` is only implemented for the `StorageTransaction`, and the `Storage` should only implement read methods(and only one mutate `commit_changes` method).
[...]

PR #1590, opened 2024-01-08
Use `AtomicView` in the `TxPool`

xgreenx:
The change is related to https://github.com/FuelLabs/fuel-core/issues/1589.

The idea of the change is to start using the `AtomicView` inside of the `TxPool` to generate consistent database representation during the insertion of the transactions.


Dentosal (approved):

crates/fuel-core/src/service/adapters/graphql_api.rs
```diff
@@ -145,3 +148,18 @@ impl worker::BlockImporter for BlockImporterAdapter {
...
+    type View = Database;
+
+    fn view_at(&self, _: BlockHeight) -> StorageResult<Self::View> {
+        unimplemented!(
+            "Unimplemented until of the https://github.com/FuelLabs/fuel-core/issues/451"
+        )
+    }
+
+    fn latest_view(&self) -> Self::View {
+        // TODO: https://github.com/FuelLabs/fuel-core/issues/1581
```
Dentosal, line 162: I feel like we should fix this TODO as well, now that the linked issue is no longer blocked. I'm fine with a follow-up PR as well.

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -57,25 +57,77 @@ pub type OnChainView = Arc<dyn OnChainDatabase>;
 /// The off-chain view of the database used by the [`ReadView`] to fetch off-chain data.
 pub type OffChainView = Arc<dyn OffChainDatabase>;
 
+/// The GraphQL can't work with the generics in [`async_graphql::Context::data_unchecked`] and requires a known type.
+/// It is an `Arc` wrapper around the generic for on-chain and off-chain databases.
+struct ArcWrapper<Provider, ArcView> {
```
Voxelot, line 62: maybe it would make sense to split this wrapper into a separate module so it doesn't clutter the main database implementation?

Voxelot (approved):

Issue #1544, opened 2023-12-14
Versioning of public types


xgreenx:
We need to make all public complex types like `Block`s, `Transaction`s, `Input`s, `Output`s, and `Consensus` to support versioning. It is not related to the primitive types.

<details>
<summary>Using Rust enums for versioning types in the context of blockchains</summary>
<br>
Enums can be used to version types in the context of blockchains. This is particularly useful for representing changes to the blockchain protocol over time without breaking compatibility with existing nodes.

**Example:**

```rust
enum ConsensusAlgorithmVersion {
  V1(ConsensusAlgorithmV1),
  V2(ConsensusAlgorithmV2),
}

struct ConsensusAlgorithmV1 {
    voting_mechanism: VotingMechanism,
}

struct ConsensusAlgorithmV2 {
    voting_mechanism: VotingMechanism,
}

impl From<ConsensusAlgorithmV1> for ConsensusAlgorithmV2 {
    fn from(algorithm: ConsensusAlgorithmV1) -> ConsensusAlgorithmV2 {
        ConsensusAlgorithmV2 {
            voting_mechanism: algorithm.voting_mechanism,
        }
    }
}
```

Existing code that interacts with the blockchain can use the `ConsensusAlgorithmVersion` enum to determine the specific version of the consensus algorithm that is used for the current block. This allows you to introduce new versions of the consensus algorithm without breaking existing code that handles consensus messages.

#### Benefits of using enums for versioning types

* **Flexibility:** You can easily add new versions of the type without modifying existing code.
* **Backward compatibility:** Existing code that handles the older version of the type will continue to work seamlessly with the new version.
* **Maintainability:** It is easier to track and manage changes to the type over time.

</details>

Most of them already are enums. But some of them(like `Block`) - not. In the scope of this task, we need to revise all existing types and decide which ones should be enums and which should not. It may also require `fuel-specs` modifications.

### Notes:

Add `#[non_exhaustive]` flag.

Dentosal:
Should we use `#[non_exhaustive]` for the version enum? This way users of the type must handle the cases where they're not aware of the new values yet.

Voxelot:
This story should also include the ability to configure the block height at which these versions take effect

MitchTurner:
We need to make sure we are covering all types that get included in the DB. There are also subtypes, like `BlockHeader` that  definitely needs to be versioned.

How can we feel confident that we have caught all relevant types?

PR #1596, opened 2024-01-10
Make `Consensus` version-able

MitchTurner:
Part of: https://github.com/FuelLabs/fuel-core/issues/1544


Dentosal (dismissed):

crates/fuel-core/src/schema/block.rs
```diff
@@ -83,6 +86,16 @@ pub struct PoAConsensus {
     signature: Signature,
 }
 
+#[derive(ThisError, Debug)]
+pub enum BlockError {
+    #[error("Error from `async_graphql`")]
+    AsyncGraphQl(async_graphql::Error),
+    #[error("Error from `fuel_core_storage`")]
+    FuelCoreStorage(fuel_core_storage::Error),
```
xgreenx, line 94: You shallow underlying error here=) We need to return it via `{0}`

crates/fuel-core/src/schema/block.rs
```diff
@@ -83,6 +86,16 @@ pub struct PoAConsensus {
     signature: Signature,
 }
 
+#[derive(ThisError, Debug)]
+pub enum BlockError {
```
xgreenx, line 90: Hmm, I'm not sure if we really want to have a custom error here. Since on the client side, it still will be a string. Maybe `anyhow::Error` will be enough to not complicate things.

MitchTurner: My issue is primarily that we don't own the `async_graphql::Error` and we are exposing it. On second look though, these are all non-`pub` functions, so I don't know that it actually matters.

In fact, I'm not sure who uses these functions at all. My IDE doesn't seem to know.

MitchTurner: Oh.... It's an `#[Object]`, that's right. I'll just put things back then. I get what's going on.

xgreenx (approved):

PR #1600, opened 2024-01-15
Upgrade to the fuel-vm 0.45.0

Dentosal:
Closes https://github.com/FuelLabs/fuel-core/issues/1544

Addresses breaking changes from https://github.com/FuelLabs/fuel-vm/pull/654


crates/fuel-core/src/schema/tx/input.rs
```diff
@@ -298,6 +298,7 @@ impl From<&fuel_tx::Input> for Input {
                 predicate: HexString(predicate.clone()),
                 predicate_data: HexString(predicate_data.clone()),
             }),
+            input => todo!("No mapping for input {input:?}"),
```
MitchTurner, line 301: We might need to use a `TryFrom` instead of a `From` so we don't have panicking code.

Dentosal:
Some parts of the code just cannot work with variants they don't know about. For instance this part here

https://github.com/FuelLabs/fuel-core/blob/b442485050b7668f079b7ed6f66b4c3d681ffb64/crates/services/txpool/src/containers/dependency.rs#L159

@MitchTurner any ideas on how to handle this gracefully?

MitchTurner:
I think I would be comfortable just doing nothing in that case.

I'm sure this could introduce a bug if we create new variants, but that's the cost we get to pay for using `non_exhaustive`.

xgreenx (approved):
It is very sad that with wildcard `_ =>,` we lose the ability to track new variants in the code. It means we need to manually find all places in the code where we need to update the logic during the addition of a new variant.

It looks like it would be better to not use `non_exhaustive` for `Transaction`, `Input`, and `Output`=)

But I'm okay with any of the two approaches

Issue #1605, opened 2024-01-18
Optimal type encoding for DA layer storage


Dentosal:

xgreenx:
We can consider it solved for now=) https://github.com/FuelLabs/fuel-vm/pull/670

PR #1609, opened 2024-01-18
DA compression

Dentosal:
Related https://github.com/FuelLabs/fuel-core/issues/1605. VM PR https://github.com/FuelLabs/fuel-vm/pull/670.

This PR  adds DA compression crate for Fuel blocks, performed upon block creation. The compressed blocks are stored into the offchain database and can be fetched using the GraphQL API.

## Note for reviewers

To keep this reasonably compact, decompression support is not included in this PR, and will be done as a follow-up. As a result, the full data roundtrip testing is not part of this PR. There's no proof here that compression of full blocks is reversible.

#### Features
- [x] Temporal registry db support
- [x] Optimize temporal registry eviction implementation
- [x] Implement TxId ↔ TxPointer lookups
- [x] Integrate with the block committer (GraphQL interface, probably)
#### Tests
- [x] compressed blocks are available from non-block-producer nodes
- [ ] e2e test for the full decompression cycle (moved to a follow-up)

## Follow-up issues
- Sync the node from L1: https://github.com/FuelLabs/fuel-core/issues/2208
- Decompression roudntrip tests https://github.com/FuelLabs/fuel-core/issues/2238
- Figure out which cache eviction algorithm/behavior is wanted: https://github.com/FuelLabs/fuel-core/issues/2231
- Figure out if we need to remove the compressed blocks from the db after a while
- Merkle roots for fraud proofs https://github.com/FuelLabs/fuel-core/issues/2232


crates/compression/src/services/compress.rs
```diff
@@ -0,0 +1,87 @@
...
+        <Vec<Transaction> as CompressibleBy<_>>::compress_with(&target, &mut prepare_ctx)
+            .await?;
+
+    let mut ctx = CompressCtx {
+        db: prepare_ctx.db,
+        cache_evictor: CacheEvictor {
+            keep_keys: prepare_ctx.accessed_keys,
+        },
+        changes: Default::default(),
+    };
```
rymnc, line 57:
```suggestion
    let mut ctx: CompressCtx = prepare_ctx.into();
```

where 
```rust
impl<D: CompressDb> Into<CompressCtx<D>> for PrepareCtx<D> {
  fn into(self) -> CompressCtx<D> {
    CompressCtx {
      db: self.db,
      cache_evictor: CacheEvictor {
        keep_keys: self.accessed_keys,
      },
      changes: Default::default()
    }
  }
}
```

crates/fuel-core/src/graphql_api/da_compression.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        // Remove the overwritten value from index, if any
+        self.0
+            .storage_as_mut::<DaCompressionTemporalRegistryIndex>()
+            .remove(&(keyspace, value.clone()))?;
+
+        // Add the new value to the index
+        self.0
+            .storage_as_mut::<DaCompressionTemporalRegistryIndex>()
+            .insert(&(keyspace, value), &key)?;
```
rymnc, line 89: is there a way we can make this group of operations atomic?

crates/fuel-core/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 #![deny(clippy::arithmetic_side_effects)]
 #![deny(clippy::cast_possible_truncation)]
 #![deny(unused_crate_dependencies)]
-#![deny(warnings)]
+// #![deny(warnings)]
```
rymnc, line 4:
```suggestion
```

rymnc (commented):
first pass, few questions!
looking good :)

crates/fuel-core/src/graphql_api/da_compression.rs
Dentosal: They are atomic, as they operate on an atomic transaction

crates/compression/src/services/compress.rs
```diff
@@ -0,0 +1,87 @@
...
+    Other(#[from] anyhow::Error),
+}
+
+pub trait CompressDb: TemporalRegistry + UtxoIdToPointer {}
+impl<T> CompressDb for T where T: TemporalRegistry + UtxoIdToPointer {}
+
+pub async fn compress<D: CompressDb>(db: D, block: &Block) -> Result<Vec<u8>, Error> {
+    // if *block.header().height() != db.next_block_height()? {
+    //     return Err(Error::NotLatest);
+    // }
```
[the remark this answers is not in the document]
Dentosal: I'm not yet sure if that code must exist or not.

crates/compression/src/eviction_policy.rs
```diff
@@ -0,0 +1,33 @@
...
+        // TODO: this can be optimized by keeping a counter of the last key used
+        // TODO: use a proper algo, maybe LRU?
+        let mut key = RegistryKey::ZERO;
+        while self.keep_keys[keyspace].contains(&key) {
+            key = key.next();
+            assert_ne!(key, RegistryKey::ZERO, "Ran out of keys");
+        }
+
+        self.keep_keys[keyspace].insert(key);
+        Ok(key)
```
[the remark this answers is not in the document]
Dentosal: Issue listed on PR description, but I'll create a real one as well. I don't think it makes sense to collect metrics here, as they can be indepdendently derived from the blocks themselves.

crates/compression/src/context/compress.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    let ser_value = postcard::to_stdvec(value)?;
+    if let Some(found) = ctx.db.registry_index_lookup(keyspace, ser_value)? {
+        return Ok(found);
+    }
+
+    let key = ctx.cache_evictor.next_key(keyspace)?;
+    let old = ctx.changes[keyspace].insert(key, PostcardSerialized::new(value)?);
+    assert!(old.is_none(), "Key collision in registry substitution");
+    Ok(key)
```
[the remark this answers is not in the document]
Dentosal: `cache_evictor` isn't used by decompression. DB operations are not shared between compression and decompression either. I'm not completely sure what you're looking to abstract away here?

crates/compression/src/services/compress.rs
Dentosal: I don't feel like hiding the behavior like this makes much sense, especially when it's only used in a single location.

crates/compression/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+            header: Header {
+                da_height: DaBlockHeight::default(),
+                prev_root: Default::default(),
+                height: 3u32.into(),
+                consensus_parameters_version: 1,
+                state_transition_bytecode_version: 2,
+                time: Tai64::UNIX_EPOCH,
+            },
+            transactions: vec![],
+        };
```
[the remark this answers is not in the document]
Dentosal: Changed this to use proptest for the inputs in 4748041a3b

crates/compression/src/eviction_policy.rs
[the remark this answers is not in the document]
Dentosal: https://github.com/FuelLabs/fuel-core/issues/2231

crates/fuel-core/src/schema/da_compressed.rs
```diff
@@ -0,0 +1,51 @@
...
+        HexString(self.bytes.clone())
+    }
+}
+
+#[derive(Default)]
+pub struct DaCompressedBlockQuery;
+
+#[Object]
+impl DaCompressedBlockQuery {
+    #[graphql(complexity = "2 * QUERY_COSTS.storage_read + child_complexity")]
```
Dentosal, line 40: This I still have to figure out. Reviewers are free to point out solutions.

crates/compression/src/compress.rs
```diff
@@ -0,0 +1,95 @@
...
+            height: *block.header().height(),
+            time: block.header().time(),
+        },
+        transactions,
+    };
+
+    let version = 0u8;
+
+    let compressed =
+        postcard::to_allocvec(&(version, compact)).expect("Serialization cannot fail");
```
xgreenx, line 92: I think it is better to use an enum instead of a tuple with a version. The variants of the enum will be our "versions"

crates/compression/src/context/compress.rs
```diff
@@ -0,0 +1,117 @@
...
+        PostcardSerialized,
+        RegistryKeyspace,
+    },
+};
+
+pub struct CompressCtx<D> {
+    pub db: D,
+    pub cache_evictor: CacheEvictor,
+    /// Changes to the temporary registry, to be included in the compressed block header
+    pub changes: PerRegistryKeyspace<HashMap<RegistryKey, PostcardSerialized>>,
```
xgreenx, line 35: Everywhere in the codebase, we heavily rely on the `PostcardSerialized` type and on the fact that we use poscard. Even in the database where it is up to the database to decide how to serialize and deserialize values. 

I think we don't need to rely on postcard, and we can stick with actual types.

As an example:

We already use `tables!` macro rule to define the `PerRegistryKeyspace` type with a separate field for each table. Instead of using `HashMap<RegistryKey, PostcardSerialized>` for each table, we could use `HashMap<RegistryKey, $type>` inside of the macro.

```rust
tables!(
    address: Address,
    asset_id: AssetId,
    contract_id: ContractId,
    script_code: Vec<u8>,
    predicate_code: Vec<u8>,
);
```
In this case, we move the decision about what serialization type to use to later stage.

crates/compression/src/eviction_policy.rs
```diff
@@ -0,0 +1,41 @@
...
+        &mut self,
+        db: &mut D,
+        keyspace: RegistryKeyspace,
+    ) -> anyhow::Result<RegistryKey>
+    where
+        D: EvictorDb,
+    {
+        // Pick first key not in the set
+        // TODO: use a proper algo, maybe LRU?
+        let mut key = db.read_latest(keyspace)?;
```
xgreenx, line 30: It seems you can cache the latest key during the first `next_key` and use local value instead doing it each time

crates/compression/src/eviction_policy.rs
```diff
@@ -0,0 +1,41 @@
...
+        while self.keep_keys[keyspace].contains(&key) {
+            key = key.next();
+            assert_ne!(key, RegistryKey::ZERO, "Ran out of keys");
```
xgreenx, line 33: It seems we can hit this assert just during iteration. 

image

If the key is `Self::DEFAULT_VALUE - 1`, then the next key return a zero key.

crates/compression/src/ports.rs
```diff
@@ -0,0 +1,85 @@
...
+        keyspace: RegistryKeyspace,
+        key: RegistryKey,
+    ) -> anyhow::Result<Vec<u8>>;
+
+    /// Reads a value from the registry at its current height.
+    fn write_registry(
+        &mut self,
+        keyspace: RegistryKeyspace,
+        key: RegistryKey,
+        value: Vec<u8>,
```
xgreenx, line 34: I think we rely too much on postcard serialization. It is up to the storage to decide what serialization format to use for data.

I think the `TemporalRegistry` should be generic over a `T` type. Something like: 

```rust
pub trait TemporalRegistry<T> {
     /// Reads a value from the registry at its current height.
     fn read_registry(
         &self,
         key: RegistryKey,
     ) -> anyhow::Result<T>;

     /// Reads a value from the registry at its current height.
     fn write_registry(
         &mut self,
         key: RegistryKey,
         value: T,
     ) -> anyhow::Result<()>;

     /// Lookup registry key by the value.
     fn registry_index_lookup(
         &self,
         value: T,
     ) -> anyhow::Result<Option<RegistryKey>>;
 }
```

In this case, the implementation of the `registry_index_lookup` for script and predicate can handle the logic of hashing the value

image

crates/compression/src/ports.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/// Information about a message.
+#[derive(Debug, Clone)]
+pub struct MessageInfo {
+    pub sender: Address,
+    pub recipient: Address,
+    pub amount: Word,
+    pub data: Vec<u8>,
+}
```
xgreenx, line 72: [nit] I think we can use `CompressedCoin` and `MessageCoin` types from `fuel-core-types` here. But it is up to use=)

crates/fuel-core/src/graphql_api/da_compression.rs
```diff
@@ -0,0 +1,158 @@
...
+
+impl<'a, Tx> UtxoIdToPointer for CompressTx<'a, Tx>
+where
+    Tx: OffChainDatabaseTransaction,
+{
+    fn lookup(
+        &self,
+        utxo_id: fuel_core_types::fuel_tx::UtxoId,
+    ) -> anyhow::Result<fuel_core_types::fuel_tx::CompressedUtxoId> {
+        for event in self.1 {
```
xgreenx, line 114: I like that we don't need to look into the database and we can get this data from the `ImportResult`=D

crates/fuel-core/src/graphql_api/da_compression.rs
```diff
@@ -0,0 +1,158 @@
...
+        .expect("The current implementation resolved all futures instantly")?;
+
+    transaction
+        .storage_as_mut::<DaCompressedBlocks>()
+        .insert(&block.header().consensus().height, &compressed)?;
+
+    Ok(())
+}
+
+struct CompressTx<'a, Tx>(&'a mut Tx, &'a [Event]);
```
xgreenx, line 50: [nit] I would prefer to have fields like `.block_db` and `.block_events` instead of `.0` and `.1`

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -114,10 +119,10 @@ pub struct Task<TxPool, D> {
     continue_on_error: bool,
 }
 
-impl<TxPool, D> Task<TxPool, D>
+impl<'a, TxPool, D> Task<TxPool, D>
 where
     TxPool: ports::worker::TxPool,
-    D: ports::worker::OffChainDatabase,
+    D: ports::worker::OffChainDatabase + 'a,
```
xgreenx, line 125: Hmm, I don't see why we need a lifetime here=)

crates/fuel-core/src/schema/da_compressed.rs
xgreenx: Maybe we want to have a new cost just for the compressed block since it doesn't have any children. Something like `1000` for compressed block

crates/fuel-core/src/service/adapters/da_compression.rs
```diff
@@ -0,0 +1 @@
+
```
xgreenx, line 1: Do we need this file?

xgreenx (commented):
There is one more question that we need to resolve:

When do we want to enable the DA compression? Do we want to enable it starting a specific block, or do we want to start it from the genesis block?

If the first one, when we need to coordinate it with the ecosystem. If the second one is, then we need to come up with a migration solution.

crates/fuel-core/src/service/adapters/da_compression.rs
Dentosal: 50d53d1424

crates/fuel-core/src/graphql_api/worker_service.rs
Dentosal: 50d53d1424

crates/fuel-core/src/graphql_api/da_compression.rs
Dentosal: 50d53d1424

crates/compression/src/eviction_policy.rs
Dentosal: Yep it was clearly incorrect. Fixed in e0ddd85622.

crates/compression/src/eviction_policy.rs
Dentosal: I'm not sure if optimizations like this make sense. The DB is supposed to be quite fast, and trading off code readibility for speed is not a thing I like to do without a good reason.

crates/compression/src/ports.rs
Dentosal: The proposed solution is not viable, since there's no single `T` that covers the input types. It's different for different keyspaces. Currently I'm planning to just have enum over all keyspaces, and that can be used to pass in the values.

xgreenx: It is the main idea that there are not single `T`. The register should implement `TemporalRegistry<T>` for each type, that we want to support. As an example: Database can implement `TemporalRegistry<Address>`, `TemporalRegistry<ContractId`, etc.

Dentosal: This has some storage overhead, but that's likely not a huge issue.

crates/compression/src/context/compress.rs
Dentosal: Got rid of the `PostcardSerialized` type in e0e505de032427bb941a48a9526558f5ba6f97e2

crates/fuel-core/src/schema/da_compressed.rs
Dentosal: e165e9e533

crates/compression/src/compress.rs
```diff
@@ -0,0 +1,95 @@
...
+    let registrations = RegistrationsPerTable::try_from(registrations)?;
+
+    // Apply changes to the db
+    registrations.write_to_registry(&mut ctx.db)?;
+
+    // Construct the actual compacted block
+    let compact = CompressedBlockPayload {
+        registrations,
+        registrations_root: Bytes32::default(), /* TODO: https://github.com/FuelLabs/fuel-core/issues/2232 */
+        header: Header {
```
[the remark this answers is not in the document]
Dentosal: Just to clarify: do you mean `PartialBlockHeader` or `BlockHeader`? I.e. should we also include generated fields?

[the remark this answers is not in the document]
xgreenx: Hmm, I thought about `BlockHeader`, but maybe you are right and it will be enough to use `PartialBlockHeader`. The problem with `PartialBlockHeader` is that it is not versioned, but I think maybe it is okay since we have version in the compressed payload=)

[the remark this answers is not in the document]
Dentosal: Changed to use `PartialBlockHeader` in e036b824b9

crates/compression/src/compress.rs
Dentosal: Changed to use an enum in 9a75b4b58d

crates/compression/src/ports.rs
Dentosal: They contain extra fields, so leaving this as-is for now.

crates/compression/src/compress.rs
```diff
@@ -0,0 +1,95 @@
...
+    db: D,
+    block: &Block,
+) -> Result<Vec<u8>, Error> {
+    // if *block.header().height() != db.next_block_height()? {
+    //     return Err(Error::NotLatest);
+    // }
+
+    let target = block.transactions().to_vec();
+
+    let mut prepare_ctx = PrepareCtx {
```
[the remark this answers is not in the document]
Dentosal: `CompressCtx` doesn't currently handle that. The logic will be more complex if we want to do it without `PrepareCtx`, and in addition we now re-register values more than we strictly have to. Quick esitmate says we lose maybe one value per five hours of blocks, which is not much.

I made a separate PR for this code so you can see if you prefer it: https://github.com/FuelLabs/fuel-core/pull/2251/files

I slightly prefer the version with PrepareCtx as it seems more obviously correct as there's less state to manage.

Dentosal:
> There is one more question that we need to resolve:
> 
> When do we want to enable the DA compression? Do we want to enable it starting a specific block, or do we want to start it from the genesis block?
> 
> If the first one, when we need to coordinate it with the ecosystem. If the second one is, then we need to come up with a migration solution.

Don't we already post the blocks to L1? Maybe we could call those "V0" comprssed blocks, and allow the decompressor to work on those as well? I'm not sure if they are easily recognizable, though.

Is anyone in the ecosystem depending on the posted blocks? What kind of coordination is needed?

xgreenx:
> Don't we already post the blocks to L1? Maybe we could call those "V0" comprssed blocks, and allow the decompressor to work on those as well? I'm not sure if they are easily recognizable, though.

We do. Those blocks don't have versioning.

> Is anyone in the ecosystem depending on the posted blocks? What kind of coordination is needed?

Our sentries depend on it=)  And anyone in the network who wants to follow the compression. The network should agree starting which block height we are doing compression.

It can be the first blob with a new compression type, but then we have a problem with our cluster(how sentries know what block to start). Plus, we need add support for blobs into our Relayer to track this information=)


I think just setting block height in the `ChainConfig` will be the simplest option. But we need to discuss it cc @Voxelot

crates/compression/src/compress.rs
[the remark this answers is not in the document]
xgreenx: I don't know=) I like both variants, so it is up to you

crates/compression/src/tables.rs
```diff
@@ -0,0 +1,210 @@
...
+        pub enum RegistryKeyspace {
+            $(
+                $name,
+            )*
+        }
+
+        #[doc = "RegistryKey namespace with an associated typed value"]
+        #[derive(Debug, Clone, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
+        #[allow(non_camel_case_types)] // Match names in structs exactly
+        pub enum RegistryKeyspaceValue {
```
xgreenx, line 48: I prefer if we define this type outside the macros since it is not related to compression and belongs to the storage part

crates/compression/src/tables.rs
```diff
@@ -0,0 +1,210 @@
...
+
+        #[doc = "Key-value mapping for each keyspace"]
+        #[derive(Debug, Clone, Default)]
+        pub struct PerRegistryKeyspaceMap {
+            $(pub $name: HashMap<RegistryKey, $type>,)*
+        }
+
+        impl PerRegistryKeyspaceMap {
+            #[cfg(test)]
+            pub fn insert(&mut self, key: RegistryKey, value: RegistryKeyspaceValue) {
```
xgreenx, line 88: Where do we call this function?

crates/compression/src/tables.rs
```diff
@@ -0,0 +1,210 @@
...
+            }
+        }
+
+        #[doc = "The set of registrations for each table, as used in the compressed block header"]
+        #[derive(Debug, Clone, PartialEq, Default, serde::Serialize, serde::Deserialize)]
+        pub struct RegistrationsPerTable {
+            $(pub $name: Vec<(RegistryKey, $type)>,)*
+        }
+
+        impl TryFrom<PerRegistryKeyspaceMap> for RegistrationsPerTable {
```
xgreenx, line 105: Why do we implement `TryFrom` instead of `From`?

crates/compression/src/ports.rs
```diff
@@ -0,0 +1,72 @@
...
+/// Rolling cache for compression.
+/// Holds the latest state which can be event sourced from the compressed blocks.
+/// The changes done using this trait in a single call to `compress` or `decompress`
+/// must be committed atomically, after which block height must be incremented.
+pub trait TemporalRegistry<T> {
+    /// Reads a value from the registry at its current height.
+    fn read_registry(&self, key: RegistryKey) -> anyhow::Result<T>;
+
+    /// Reads a value from the registry at its current height.
+    fn write_registry(&mut self, key: RegistryKey, value: T) -> anyhow::Result<()>;
```
xgreenx, line 26:
```suggestion
    fn write_registry(&mut self, key: RegistryKey, value: &T) -> anyhow::Result<()>;
```

crates/fuel-core/src/graphql_api/da_compression.rs
```diff
@@ -0,0 +1,193 @@
...
+                    }
+                }
+
+                fn write_registry(
+                    &mut self,
+                    key: fuel_core_types::fuel_compression::RegistryKey,
+                    value: $type,
+                ) -> anyhow::Result<()> {
+                    // Write the actual value
+                    self.db_tx
```
xgreenx, line 98: I think we do too many clones=) While it is okay for `ContractId` and similar types, it can be an issue for script and predicate. If we used separate tables for each type we could avoid it

crates/fuel-core/src/graphql_api/da_compression.rs
```diff
@@ -0,0 +1,193 @@
...
+                    let output_index = coin.utxo_id.output_index();
+                    return Ok(fuel_core_types::fuel_tx::CompressedUtxoId {
+                        tx_pointer: coin.tx_pointer,
+                        output_index,
+                    });
+                }
+                _ => {}
+            }
+        }
+        panic!("UtxoId not found in the block events");
```
xgreenx, line 163: We should return an error instead of panic, and maybe `debug_assert!(false)` if you want to catch it during local developing

crates/fuel-core/src/graphql_api/da_compression.rs
```diff
@@ -0,0 +1,193 @@
...
+                        .get(&RegistryKeyspaceValue::$name(value.clone()))?
+                        .map(|v| v.into_owned()))
+                }
+            }
+
+        )*
+    };
+}
+
+// Arguments here should match the tables! macro from crates/compression/src/tables.rs
```
xgreenx, line 132: It would be better if the implementation of the storage was independent of the `tables!` macro. It seems the removing of `RegistryKeyspace` and `RegistryKeyspaceValue` from implementation will solve that.

crates/fuel-core/src/graphql_api/storage/da_compression.rs
```diff
@@ -0,0 +1,169 @@
...
+        Self::Column::DaCompressionTemporalRegistry
+    }
+}
+
+pub struct DaCompressionTemporalRegistryIndex;
+
+impl Mappable for DaCompressionTemporalRegistryIndex {
+    type Key = Self::OwnedKey;
+    // TODO: should we hash the key?
+    type OwnedKey = RegistryKeyspaceValue;
```
xgreenx, line 61: We should use hash in the case of scripts and predicates. I think it is better to define a separate enum just for this table

crates/fuel-core/src/graphql_api/da_compression.rs
Dentosal: Maybe `unreachable!` instead of `panic!`? But I don't see why we should ever get here. What good does it do to return an error, this in unrecoverable anyway? If we cannot produce a compressed block, halting the chain is a reasonable action.

crates/compression/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+anyhow = { workspace = true }
+fuel-core-types = { workspace = true, features = [
+  "alloc",
+  "serde",
+  "da-compression",
+] }
+postcard = { version = "1.0", features = ["use-std"] }
+rand = { workspace = true, optional = true }
+serde = { version = "1.0", features = ["derive"] }
+thiserror = { workspace = true }
```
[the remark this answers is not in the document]
Dentosal: `serde` is not `no_std` compatible either and it's a required dependency here

crates/compression/src/tables.rs
Dentosal: https://github.com/FuelLabs/fuel-core/blob/0957584e90b3ed344cd5e2e31502ffc13958b127/crates/compression/src/lib.rs#L133

crates/compression/src/tables.rs
Dentosal: It used to be fallible when we used serde. Now no reason, so I'll change it.

crates/fuel-core/src/graphql_api/storage/da_compression.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+pub struct DaCompressionTemporalRegistry;
+
+impl Mappable for DaCompressionTemporalRegistry {
+    type Key = Self::OwnedKey;
+    type OwnedKey = (RegistryKeyspace, RegistryKey);
+    type Value = Self::OwnedValue;
+    type OwnedValue = RegistryKeyspaceValue;
+}
```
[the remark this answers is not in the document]
Dentosal: This creates so much copy paste I cannot imagine it being worth it. For instance, now the following new bounds are required for `OffChainDatabaseTransaction`:

```rust
        + StorageMutate<DaCompressedBlocks, Error = StorageError>
        + StorageMutate<DaCompressionTemporalRegistryAddress, Error = StorageError>
        + StorageMutate<DaCompressionTemporalRegistryIndexAddress, Error = StorageError>
        + StorageMutate<DaCompressionTemporalRegistryEvictorAddress, Error = StorageError>
        + StorageMutate<DaCompressionTemporalRegistryAssetId, Error = StorageError>
        + StorageMutate<DaCompressionTemporalRegistryIndexAssetId, Error = StorageError>
        + StorageMutate<DaCompressionTemporalRegistryEvictorAssetId, Error = StorageError>
        + StorageMutate<DaCompressionTemporalRegistryContractId, Error = StorageError>
        + StorageMutate<DaCompressionTemporalRegistryIndexContractId, Error = StorageError>
        + StorageMutate<
            DaCompressionTemporalRegistryEvictorContractId,
            Error = StorageError,
        > + StorageMutate<DaCompressionTemporalRegistryScriptCode, Error = StorageError>
        + StorageMutate<DaCompressionTemporalRegistryIndexScriptCode, Error = StorageError>
        + StorageMutate<
            DaCompressionTemporalRegistryEvictorScriptCode,
            Error = StorageError,
        > + StorageMutate<DaCompressionTemporalRegistryPredicateCode, Error = StorageError>
        + StorageMutate<
            DaCompressionTemporalRegistryIndexPredicateCode,
            Error = StorageError,
        > + StorageMutate<
            DaCompressionTemporalRegistryEvictorPredicateCode,
            Error = StorageError,
        >
```

crates/fuel-core/src/graphql_api/storage/da_compression.rs
```diff
@@ -0,0 +1,169 @@
...
+    type Column = super::Column;
+
+    fn column() -> Self::Column {
+        Self::Column::DaCompressionTemporalRegistryIndex
+    }
+}
+
+/// This table is used to hold "next key to evict" for each keyspace.
+/// In the future we'll likely switch to use LRU or something, in which
+/// case this table can be repurposed, iff migrations have been figured out.
```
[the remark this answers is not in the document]
Dentosal: Removed in 946a2e8f23fef2109fb9ba286ebd2d451c2af7cb

crates/fuel-core/src/graphql_api/storage/da_compression.rs
```diff
@@ -0,0 +1,169 @@
...
+pub struct DaCompressionTemporalRegistryEvictor;
+
+impl Mappable for DaCompressionTemporalRegistryEvictor {
+    type Key = Self::OwnedKey;
+    type OwnedKey = RegistryKeyspace;
```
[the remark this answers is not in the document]
Dentosal: Using a separate evictor column for each keyspace in 946a2e8f23fef2109fb9ba286ebd2d451c2af7cb

crates/fuel-core/src/graphql_api/storage/da_compression.rs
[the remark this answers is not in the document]
Dentosal: Implemented in 946a2e8f23fef2109fb9ba286ebd2d451c2af7cb

crates/fuel-core/src/graphql_api/da_compression.rs
Dentosal: It is independent. The comment here just tells the future person modifying the code where to look for values that are used here. But now it's removed in 946a2e8f23fef2109fb9ba286ebd2d451c2af7cb anyway.

crates/fuel-core/src/graphql_api/storage/da_compression.rs
Dentosal: Moved to use a hash in 946a2e8f23fef2109fb9ba286ebd2d451c2af7cb. The enum is gone.

crates/fuel-core/src/graphql_api/da_compression.rs
Dentosal: Clones are removed in 946a2e8f23fef2109fb9ba286ebd2d451c2af7cb

crates/compression/Cargo.toml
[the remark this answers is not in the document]
Dentosal: Rermoved in c5ac5db3cf

crates/compression/src/tables.rs
Dentosal: I just removed it in 7fe125fdfe3a48e1f9de1dd400027ee2d06d66e4

crates/compression/src/tables.rs
Dentosal: Fixed in 8d5fdf20c5d55347dee3724f823d3b1479de5d07

crates/compression/src/eviction_policy.rs
Dentosal: Ok now done in eda9e77797

crates/compression/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+            Just(RegistryKeyspace::AssetId),
+            Just(RegistryKeyspace::ContractId),
+            Just(RegistryKeyspace::ScriptCode),
+            Just(RegistryKeyspace::PredicateCode),
+        ]
+    }
+
+    proptest! {
+        /// Serialization for compressed transactions is already tested in fuel-vm,
+        /// but the rest of the block de/serialization is be tested here.
```
xgreenx, line 75: While it is true, it doesn't change the fact that we need to test compression and decompression here. In the `fuel-tx,`, we test those traits for compression, and their deriving works correctly. Here, we need to test that the logic of the compress and decompress contexts is correct. We need to be sure that we collect the right addresses, IDs, and codes and store them in the database. That we assign correct register keys, and can handle cases when we use the same register key during one compression.

crates/compression/src/ports.rs
```diff
@@ -0,0 +1,103 @@
...
+pub struct MessageInfo {
+    pub sender: Address,
+    pub recipient: Address,
+    pub amount: Word,
+    pub data: Vec<u8>,
+}
+
+/// Evictor registry to keep track of the latest used key for the type `T`.
+pub trait EvictorDb<T> {
+    fn read_latest(&self) -> anyhow::Result<RegistryKey>;
```
xgreenx, line 87: Maybe we want to rename it `get_latest_assigned_key ` for more clarity

crates/compression/src/ports.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    fn write_latest(&mut self, key: RegistryKey) -> anyhow::Result<()>;
```
xgreenx, line 89: Maybe `store_latest_assigned_key` is more clear

crates/fuel-core/Cargo.toml
```diff
@@ -44,6 +45,7 @@ hyper = { workspace = true }
 indicatif = { workspace = true, default-features = true }
 itertools = { workspace = true }
 num_cpus = { version = "1.16.0", optional = true }
+paste = "1"
```
xgreenx, line 48: We use it in several places, I think it will be better to use `{ workspace = true }` isntead

crates/fuel-core/src/graphql_api/da_compression.rs
```diff
@@ -0,0 +1,190 @@
...
+
+                // Remove the overwritten value from index, if any
+                self.db_tx
+                    .storage_as_mut::<[< DaCompressionTemporalRegistryIndex $type >]>()
+                    .remove(&value_in_index)?;
+
+                // Add the new value to the index
+                self.db_tx
+                    .storage_as_mut::<[< DaCompressionTemporalRegistryIndex $type >]>()
+                    .insert(&value_in_index, key)?;
```
xgreenx, line 104: Why do we need to remove it first and insert after? Insert will replace the value automaticly

Dentosal: This was broken in some refactor. Fixed in ee3e844626.

crates/compression/src/ports.rs
Dentosal: b5edb0d51b renames this to `set_latest_assigned_key`

crates/compression/src/ports.rs
Dentosal: b5edb0d51b renames this to `get_latest_assigned_key`

crates/compression/src/eviction_policy.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub fn new_from_db<D>(
+        db: &mut D,
+        keep_keys: HashSet<RegistryKey>,
+    ) -> anyhow::Result<Self>
+    where
+        D: EvictorDb<T>,
+    {
+        Ok(Self {
+            keep_keys,
+            next_key: db.read_latest()?,
```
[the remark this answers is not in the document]
Dentosal: Merged.

crates/compression/src/compress.rs
```diff
@@ -0,0 +1,223 @@
...
+                    if self == &Default::default() {
+                        return Ok(RegistryKey::DEFAULT_VALUE);
+                    }
+                    if let Some(found) = ctx.db.registry_index_lookup(self)? {
+                        return Ok(found);
+                    }
+
+                    let key = ctx.$ident.cache_evictor.next_key();
+                    let old = ctx.$ident.changes.insert(key, self.clone());
+                    assert!(old.is_none(), "Key collision in registry substitution");
```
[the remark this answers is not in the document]
Dentosal: We don't handle it here, and allocate two keys in that case. That's not correct and it's fixed in 40a1a3dc2e.

We would not hit the assert anyway, since that requires that the cache evictor would return the same key twice, which will not happen.

crates/compression/src/compress.rs
```diff
@@ -0,0 +1,122 @@
...
+    // Construct the actual compacted block
+    let compact = CompressedBlockPayloadV0 {
+        registrations,
+        registrations_root: Bytes32::default(), /* TODO: https://github.com/FuelLabs/fuel-core/issues/2232 */
+        header: block.header().into(),
+        transactions,
+    };
+
+    let compressed = postcard::to_allocvec(&CompressedBlock::V0(compact))
+        .expect("Serialization cannot fail");
```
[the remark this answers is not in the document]
Dentosal: Merged.

crates/compression/src/lib.rs
Dentosal: Yep. The testing is clearly not sufficient. I've taken this PR back to draft stage until I manage to write those.

crates/fuel-core/Cargo.toml
Dentosal: Sure. 1e70c771aa

crates/compression/src/lib.rs
xgreenx: We can merge this PR if we do not enable block compression in the `fuel-core` to simplify the review process; what do you think?

crates/fuel-core/Cargo.toml
xgreenx: We have two more places=)

image

Dentosal:
Design notes from a call with @xgreenx:

* only one dedicated node handling compression
    * if it crashes, it syncs from L1. must not be down over L1 retention time.
    * compressor is part of fuel-core, have a flag to enable
    * enable for only one sentry
* only store data for past two weeks in the temporal registry
    * block header timestamp is used for timing
    * add a CLI flag for controlling retention
* block committercan binary seach to figure which compressed blocks are available
* remove smt root field from v0 compression

crates/compression/src/lib.rs
Dentosal: We can. However, I don't trust the code to work correctly, and I'd like the code on `master` to be in a good state. That said, easing up the review process could be worth it.

Cargo.lock
```diff
@@ -169,9 +169,9 @@ dependencies = [
 
 [[package]]
```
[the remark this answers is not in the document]
Dentosal: 2264fbd12a

rafal-ch (commented):
Partial review.

crates/compression/README.md
```diff
@@ -0,0 +1,27 @@
+# Compression and decompression of transactions for the DA layer
+
+## Compressed block header
+
+Each compressed block begins with a version field, so that it's possible to change the format later.
```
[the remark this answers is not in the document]
Dentosal: We no longer guarantee the single-byteness. Enum tag is serialized as varint instead of u8 by postcard.

crates/compression/src/compress.rs
```diff
@@ -0,0 +1,247 @@
...
+};
+use std::collections::{
+    HashMap,
+    HashSet,
+};
+
+pub trait CompressDb: TemporalRegistryAll + UtxoIdToPointer {}
+impl<T> CompressDb for T where T: TemporalRegistryAll + UtxoIdToPointer {}
+
+/// This must be called for all new blocks in sequence, otherwise the result will be garbage.
```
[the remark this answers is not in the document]
Dentosal: The registry is valid for only the current block height. On any other height you could be referring to keys that have already been overwritten, or have not been written to yet.

crates/compression/src/compress.rs
```diff
@@ -0,0 +1,247 @@
...
+    };
+
+    Ok(VersionedCompressedBlock::V0(compact))
+}
+
+/// Preparation pass through the block to collect all keys accessed during compression.
+/// Returns dummy values. The resulting "compressed block" should be discarded.
+pub struct PrepareCtx<D> {
+    pub config: Config,
+    /// Current timestamp
```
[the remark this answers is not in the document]
Dentosal: Yep. It says "current" referring to "current block", to separate it from timestamps of the old blocks from the registry.

crates/compression/src/compress.rs
```diff
@@ -0,0 +1,247 @@
...
+    let _ = target.compress_with(&mut prepare_ctx).await?;
+
+    let mut ctx = prepare_ctx.into_compression_context()?;
+    let transactions = target.compress_with(&mut ctx).await?;
+    let registrations: RegistrationsPerTable = ctx.finalize()?;
+
+    // Apply changes to the db
+    registrations.write_to_registry(&mut db, block.header().consensus().time)?;
+
+    // Construct the actual compacted block
```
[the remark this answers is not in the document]
Dentosal: Removed uses of `compact` in 9c888f87f011e0d0b5a76abbf8838680c0de6e07

crates/compression/src/compress.rs
```diff
@@ -0,0 +1,223 @@
...
+            /// Commits the cache evictor states to the database.
+            fn finalize(mut self) -> anyhow::Result<RegistrationsPerTable> {
+                let mut registrations = RegistrationsPerTable::default();
+                $(
+                    self.$ident.cache_evictor.commit(&mut self.db)?;
+                    for (key, value) in self.$ident.changes.into_iter() {
+                        registrations.$ident.push((key, value));
+                    }
+                )*
+                Ok(registrations)
```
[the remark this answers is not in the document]
Dentosal: Moved there in 9c888f87f011e0d0b5a76abbf8838680c0de6e07

crates/compression/src/decompress.rs
```diff
@@ -0,0 +1,336 @@
...
+                    ctx: &DecompressCtx<D>,
+                ) -> anyhow::Result<Self> {
+                    if key == RegistryKey::DEFAULT_VALUE {
+                        return Ok(<$type>::default());
+                    }
+                    let key_timestamp = ctx.db.read_timestamp(&key)?;
+                    if !ctx.config.is_timestamp_accessible(ctx.timestamp, key_timestamp)? {
+                        anyhow::bail!("Timestamp not accessible");
+                    }
+                    ctx.db.read_registry(&key)
```
xgreenx, line 125: We can do that in a separate PR, but it would be nice to have internal cache in the context to avoid reading data from the database. Even it in the RocksDB cache, we still spend time on deserialization.

crates/compression/src/registry.rs
```diff
@@ -0,0 +1,109 @@
...
+                [<$type>],
+            )*
+        }
+
+        #[doc = "A value for each keyspace"]
+        #[derive(Debug, Clone, PartialEq, Eq, Default)]
+        pub struct PerRegistryKeyspace<T> {
+            $(pub $ident: T,)*
+        }
+        impl<T> core::ops::Index<RegistryKeyspace> for PerRegistryKeyspace<T> {
```
xgreenx, line 32: I like that we can use `[RegisterKeyspace]` syntax to access the field. I was not aware that we can do that with any type=)

crates/fuel-core/src/graphql_api/da_compression.rs
```diff
@@ -0,0 +1,212 @@
...
+                timestamp: Tai64,
+            ) -> anyhow::Result<()> {
+                // Write the actual value
+                let old_value = self.db_tx
+                    .storage_as_mut::<[< DaCompressionTemporalRegistry $type >]>()
+                    .replace(key, value)?;
+
+                // Remove the overwritten value from index, if any
+                if let Some(old_value) = old_value {
+                    let old_reverse_key = (&old_value).into();
```
xgreenx, line 119: [nit] We also can implement `From<ReverseKey>` to avoid `(&old_value)`=)

crates/fuel-core/src/graphql_api/da_compression.rs
```diff
@@ -0,0 +1,212 @@
...
+
+impl<'a, Tx> UtxoIdToPointer for CompressTx<'a, Tx>
+where
+    Tx: OffChainDatabaseTransaction,
+{
+    fn lookup(
+        &self,
+        utxo_id: fuel_core_types::fuel_tx::UtxoId,
+    ) -> anyhow::Result<fuel_core_types::fuel_tx::CompressedUtxoId> {
+        for event in self.block_events {
```
xgreenx, line 196: If we have many events in the transaction is can be really long loop. It would be nice to index all events on construction and store in the hash map.

crates/fuel-core/src/graphql_api/worker_service/tests.rs
```diff
@@ -81,6 +81,7 @@ fn worker_task_with_block_importer_and_db<D: ports::worker::OffChainDatabase>(
         block_importer,
         database,
         chain_id,
+        da_compression_config: DaCompressionConfig::Disabled,
```
xgreenx, line 84: It would be nice to have a unit test for it later to be sure that we process the block when it is enabled. Can be a good first(for new team members) follow-up PR=)

crates/fuel-core/src/schema/da_compressed.rs
```diff
@@ -0,0 +1,51 @@
...
+    fn from(bytes: Vec<u8>) -> Self {
+        Self { bytes }
+    }
+}
+
+#[Object]
+impl DaCompressedBlock {
+    async fn bytes(&self) -> HexString {
+        HexString(self.bytes.clone())
+    }
```
xgreenx, line 32: Not related to this PR, but I think `HexString` should take reference instead. Since cloning in some cases can be expensive. Plus I think we clone for transaction as well. What is not good. Would be nice to create an issue for it

crates/storage/src/lib.rs
```diff
@@ -209,12 +209,12 @@ mod test {
...
             format!("{}", not_found!("BlockId")),
-            format!("resource of type `BlockId` was not found at the: {}:{}", file!(), line!() - 1)
+            format!("resource was not found in table `BlockId` at the: {}:{}", file!(), line!() - 1)
         );
         #[rustfmt::skip]
         assert_eq!(
             format!("{}", not_found!(Coins)),
-            format!("resource of type `fuel_core_types::entities::coins::coin::CompressedCoin` was not found at the: {}:{}", file!(), line!() - 1)
+            format!("resource was not found in table `fuel_core_storage::tables::Coins` at the: {}:{}", file!(), line!() - 1)
         );
```
xgreenx, line 218: I did these changes, they are not related to this PR, but helped to debug issue that I had=D Blame me for modifying it=)

xgreenx (commented):
LGTM!=)

tests/tests/da_compression.rs
```diff
@@ -0,0 +1,105 @@
...
+    },
+    secrecy::Secret,
+};
+use rand::{
+    rngs::StdRng,
+    SeedableRng,
+};
+
+#[tokio::test]
+async fn can_fetch_da_compressed_block_from_graphql() {
```
xgreenx, line 30: Since we have an integration test that compresses block with one transactions, maybe we could  also try to decompress transaction, since all required traits are implemented?

tests/tests/da_compression.rs
```diff
@@ -0,0 +1,105 @@
...
+    let v_client = FuelClient::from(validator.node.shared.graph_ql.bound_address);
+
+    // Insert some txs
+    let expected = producer.insert_txs().await;
+    producer.consistency_10s(&expected).await;
+    validator.consistency_20s(&expected).await;
+
+    let block_height = 1u32.into();
+
+    let _ = v_client
```
xgreenx, line 100: What do you think about comparing blocks between produce and validator?

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -134,7 +147,7 @@ where
         let height = block.header().height();
         let block_id = block.id();
         transaction
-            .storage::<FuelBlockIdsToHeights>()
+            .storage_as_mut::<FuelBlockIdsToHeights>()
```
rafal-ch, line 150: This change is not needed (?).

Not sure though, because I'm a little bit confused by both `fn storage()` and `fn storage_as_mut()` being equivalent.

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -93,9 +98,16 @@ use std::{
 #[cfg(test)]
 mod tests;
 
+#[derive(Debug, Clone)]
```
rafal-ch, line 101: Could also be `Copy`, because `fuel_core_compression::config::Config` is.

However, when it comes to configs I prefer explicit cloning.

tests/tests/da_compression.rs
```diff
@@ -0,0 +1,105 @@
...
+#[tokio::test]
+async fn can_fetch_da_compressed_block_from_graphql() {
+    let mut rng = StdRng::seed_from_u64(10);
+    let poa_secret = SecretKey::random(&mut rng);
+
+    let db = CombinedDatabase::default();
+    let mut config = Config::local_node();
+    config.consensus_signer = SignMode::Key(Secret::new(poa_secret.into()));
+    config.da_compression = DaCompressionConfig::Enabled(fuel_core_compression::Config {
+        temporal_registry_retention: Duration::from_secs(3600),
```
rafal-ch, line 38: Did you consider testing this mechanism with duration set to `0`?

rafal-ch (commented):
Looking good, just added a few more minor comments and questions.

crates/compression/src/config.rs
```diff
@@ -0,0 +1,33 @@
...
+}
+
+impl Config {
+    /// Given timestamp of the current block and a key in an older block,
+    /// is the key is still accessible?
+    /// Returns error if the arguments are not valid block timestamps,
+    /// or if the block is older than the key.
+    pub fn is_timestamp_accessible(
+        &self,
+        block_timestamp: Tai64,
```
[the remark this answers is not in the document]
Dentosal: I don't think the context where the function is used should affect the signature of a function The arguments have names, and that should be enough.

crates/compression/src/decompress.rs
Dentosal: My first intuition says that the cache should live on the thing that implements the trait, and not here. I also don't like the idea of caching without first measuring that this is indeed a performance issue.

crates/compression/src/registry.rs
Dentosal: Yep it is nice.

crates/fuel-core/src/graphql_api/da_compression.rs
Dentosal: Agreed. Follow-up: https://github.com/FuelLabs/fuel-core/issues/2279

crates/fuel-core/src/graphql_api/worker_service/tests.rs
Dentosal: We do have an integration test here https://github.com/FuelLabs/fuel-core/blob/dc34257c3263639f1a023ea337b464f192487fde/tests/tests/da_compression.rs#L30

but maybe we could have an unit test as well, yes.

crates/fuel-core/src/schema/da_compressed.rs
Dentosal: I'm not sure how that would work. The lifetime of hex string is tied to the lifetime of the original bytes?

crates/storage/src/lib.rs
Dentosal: Do you want them reverted?

tests/tests/da_compression.rs
Dentosal: The test was like that before, but now only the validator crates the compressed block, so it's not possible anymore.

tests/tests/da_compression.rs
Dentosal: Not yet, but will be done in https://github.com/FuelLabs/fuel-core/issues/2238

crates/compression/src/compress.rs
```diff
@@ -0,0 +1,242 @@
...
+                    if let Some(found) = ctx.db.registry_index_lookup(self)? {
+                        let key_timestamp = ctx.db.read_timestamp(&found)
+                            .context("Database invariant violated: no timestamp stored but key found")?;
+                        if ctx.config.is_timestamp_accessible(ctx.timestamp, key_timestamp)? {
+                            return Ok(found);
+                        }
+                    }
+                    if let Some(found) = ctx.$ident.changes_lookup.get(self) {
+                        return Ok(*found);
+                    }
```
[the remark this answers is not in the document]
Dentosal: The first part is fixed in 0967f631978d39ea2f83a21dd69b2ed3177d13e8. Not sure what you mean about reusing `accessed_keys` here.

crates/fuel-core/src/graphql_api/da_compression.rs
```diff
@@ -0,0 +1,212 @@
...
+        ContractId,
+        ScriptCode,
+    },
+    services::executor::Event,
+    tai64::Tai64,
+};
+use futures::FutureExt;
+
+/// Performs DA compression for a block and stores it in the database.
+pub fn da_compress_block<T>(
```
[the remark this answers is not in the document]
Dentosal: I don't think the `_and_store` part is needed. This takes a mutable database transaction as an argument, so it's quite clear what's going on.

crates/fuel-core/src/schema/da_compressed.rs
xgreenx: Yes, or `HexString` could create hexed string on construction instead of cloning and creating it after.

crates/compression/src/compress.rs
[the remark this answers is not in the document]
xgreenx: I meant in the `PrepareCtx`. We also can avoid calling database by checking `accessed_keys` first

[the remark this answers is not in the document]
Dentosal: Like this? 8787971aa53bd341a013080aa4079499d8d5641a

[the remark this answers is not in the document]
xgreenx: Yep. I also thought that we could maybe have a cache for `self`, but since it is the type that we need to store and requires hashing of the predicates and script, it is better to leave it as it is now=)

crates/fuel-core/src/graphql_api/da_compression.rs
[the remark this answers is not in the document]
rafal-ch: Certainly not a blocker.

tests/tests/da_compression.rs
Dentosal: But nothing implements `HistoryLookup`?

xgreenx: Hmm, looks like it should be easy to implement.

image

But okay, could you create a follow-up PR with its implementation and this test, just to check that we can decompress stuff?=)

rafal-ch (dismissed):
No blockers.

xgreenx (approved):

xgreenx (approved):

rafal-ch (approved):

Issue #1583, opened 2024-01-05
Move off-chain database updates from storage trait to off-chain worker


xgreenx:
<img width="1380" alt="image" src="https://github.com/FuelLabs/fuel-core/assets/18346821/113311cd-2043-4c73-95bb-0deffda053b1">

It is better to do after https://github.com/FuelLabs/fuel-core/pull/1576 is merged.

PR #1612, opened 2024-01-21
Use `AtomicView` in all services

xgreenx:
Related to https://github.com/FuelLabs/fuel-core/issues/1589 and preparation for https://github.com/FuelLabs/fuel-core/issues/1583.

All services that use the database for read-only purposes use `AtomicView` now instead of direct access to the database.

- Removed `Relayer` from the `Verifier` because if it is not used for now, plus it may not be needed because of the shared sequencer and its consensus rules.
- Added verification of the transactions root hash into `Verifier`.
- Removed requesting of the one block from p2p, it is possible to use range request for that purposes.
- Removed not used `get_sealed_header` and `get_sealed_block` method.
- Added the `latest_height` method to `AtomicView` because the database always knows its latest height.
- Added customisation of the `Height` used by the `AtomicView`. In the case of the relayer, we want to use `DaBlockHeight` as a primary key to create a snapshot, while in the case of the Fuel's databases, we want to use `BockHeight` as a primary key.
- Renamed `Executor` into `ExecutionInstance` and changed it to be a one-time usable instance. It consumes the `self` to perform the execution of the block. The instance has a view of the database and execution options.


crates/fuel-core/src/database.rs
```diff
@@ -411,6 +415,65 @@ impl fuel_core_storage::vm_storage::VmStorageRequirements for Database {
...
+
+    fn latest_height(&self) -> BlockHeight {
+        // TODO: The database should track the latest height inside of the database object
+        //  instead of fetching it from the `FuelBlocks` table. As a temporary solution,
+        //  fetch it from the table for now.
+        self.latest_height().unwrap_or_default()
+    }
+
+    fn view_at(&self, _: &BlockHeight) -> StorageResult<Self::View> {
+        // TODO: Unimplemented until of the https://github.com/FuelLabs/fuel-core/issues/451
```
MitchTurner, line 431: It looks like this is merged now.

crates/fuel-core/src/database.rs
```diff
@@ -411,6 +415,65 @@ impl fuel_core_storage::vm_storage::VmStorageRequirements for Database {
...
+
+impl AtomicView for RelayerReadDatabase {
+    type View = Database;
+    type Height = DaBlockHeight;
+
+    fn latest_height(&self) -> Self::Height {
+        #[cfg(feature = "relayer")]
+        {
+            use fuel_core_relayer::ports::RelayerDb;
+            // TODO: The database should track the latest da height inside of the database object
```
MitchTurner, line 457: Is there a follow-up issue fore this?

crates/fuel-core/src/database.rs
xgreenx: Without state rewind we can't implement it properly=) It uses the latest for now to simulate the current behavior=)

crates/fuel-core/src/database.rs
xgreenx: Yep, https://github.com/FuelLabs/fuel-core/issues/1589

crates/fuel-core/src/service/adapters.rs
```diff
@@ -85,6 +133,16 @@ pub struct BlockImporterAdapter {
         Arc<fuel_core_importer::Importer<Database, ExecutorAdapter, VerifierAdapter>>,
 }
 
+impl BlockImporterAdapter {
+    pub fn events(&self) -> BoxStream<SharedImportResult> {
+        use futures::StreamExt;
```
MitchTurner, line 138: Do you have a reason why you'd put the import here rather than at the top. I see this sometimes and I'm curious what the justification for it is. I don't have anything against it.

crates/services/executor/src/executor.rs
```diff
@@ -159,18 +163,97 @@ impl TransactionsSource for OnceTransactionsSource {
     }
 }
 
-/// ! The executor is used for block production and validation. Given a block, it will execute all
-/// the transactions contained in the block and persist changes to the underlying database as needed.
-/// In production mode, block fields like transaction commitments are set based on the executed txs.
-/// In validation mode, the processed block commitments are compared with the proposed block.
+/// The executor is used for block production and validation.
+/// It acts like a factory creating an `ExecutionInstance` valid only for one block.
```
MitchTurner, line 167: I think this is an implementation detail. I don't know if it caller needs to know this.

crates/fuel-core/src/service/adapters.rs
xgreenx: It is used only in one place in that file. If you decide to remove that code later, you don't need to clean up imports too=) Plus it creates fewer conflicts with other people.

crates/services/executor/src/executor.rs
```diff
@@ -196,17 +279,28 @@ impl From<&Config> for ExecutionOptions {
     }
 }
 
-impl<R, D> Executor<R, D>
+/// The executor instance performs block production and validation. Given a block, it will execute all
+/// the transactions contained in the block and persist changes to the underlying database as needed.
+/// In production mode, block fields like transaction commitments are set based on the executed txs.
+/// In validation mode, the processed block commitments are compared with the proposed block.
+#[derive(Clone, Debug)]
+pub struct ExecutionInstance<R, D> {
```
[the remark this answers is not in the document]
xgreenx: You are right, it can be private=)

MitchTurner:
> Added customisation of the Height used by the AtomicView. In the case of the relayer, we want to use DaBlockHeight as a primary key to create a snapshot, while in the case of the Fuel's databases, we want to use BockHeight as a primary key.

I don't think I want to block the PR for this, but this is an example of what I was talking about with business domain. the relayer and Fuel's db might be isomorphic, but they serve very different roles in the domain of our code. I feel like the abstraction should be for the function, rather than the shape.

xgreenx:
> the relayer and Fuel's db might be isomorphic, but they serve different roles in the domain of our code.

It is true. The `AtomicView` provides customization for the `Height` not only because of how DB works. Mostly, it is dictated by the `Executor` and `BlockHeader`. 

Because the `BlockHeader` has `DaBlockHeight` and `BlockHeight`,it creates two data sources for the `Executor`. The first one is on-chain storage, and the other one is relayer storage. It is why `Executor` wants to get the view of the relayer at `DaBlockHeight` and the on-chain storage at `BlockHeight`.

Relayer doesn't need to be a database to implement the `AtomicView` trait. You can see that example in the [tests](https://github.com/FuelLabs/fuel-core/pull/1625/files#diff-379b2602e8ed0e8048291c20eff0d6ab59d4cfc1e6c5c9c6d5513b33008ba1c1). The main ideas is that with this trait relayer should provide an immutable view of the relayer at `DaBlockHeight`.

MitchTurner (approved):
Okay. This looks good to me, although I do have a small reservation about reusing the abstractions.

PR #1614, opened 2024-01-23
Remove dev consensus key

MujkicA:
This PR adds the following changes:

- removes `consensus_dev_key since the `produce_block` functionality can now be controlled through the `debug` flag
- uses the dev consensus key in the debug mode regardless of the trigger mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Description of the upcoming release here.
 ### Changed
 
 - [#1591](https://github.com/FuelLabs/fuel-core/pull/1591): Simplify libp2p dependencies and not depend on all sub modules directly.
+- [#1590](https://github.com/FuelLabs/fuel-core/pull/1590): Use `AtomicView` in the `TxPool` to read the state of the database during insertion of the transactions.
 - [#1587](https://github.com/FuelLabs/fuel-core/pull/1587): Use `BlockHeight` as a primary key for the `FuelsBlock` table.
 - [#1585](https://github.com/FuelLabs/fuel-core/pull/1585): Let `NetworkBehaviour` macro generate `FuelBehaviorEvent` in p2p
 - [#1579](https://github.com/FuelLabs/fuel-core/pull/1579): The change extracts the off-chain-related logic from the executor and moves it to the GraphQL off-chain worker. It creates two new concepts - Off-chain and On-chain databases where the GraphQL worker has exclusive ownership of the database and may modify it without intersecting with the On-chain database.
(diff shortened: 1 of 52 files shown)


bin/fuel-core/src/cli/run.rs
```diff
@@ -268,15 +262,14 @@ impl Command {
 
         // if consensus key is not configured, fallback to dev consensus key
         let consensus_key = load_consensus_key(consensus_key)?.or_else(|| {
```
xgreenx, line 264: Could you also move the code below here?

```rust
        if consensus_key.is_some() && trigger == Trigger::Never {
            warn!("Consensus key configured but block production is disabled!")
        }
```

We need to print it only in the case if the `load_consensus_key` returns something

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Description of the upcoming release here.
 
 ### Changed
 
+- [#1614](https://github.com/FuelLabs/fuel-core/pull/#1614): Use default consensus key regardless of trigger mode.
```
xgreenx, line 13: It is a breaking change, because affects CLI, could you elaborate more how the new usage looks like, please?=)

bin/fuel-core/src/cli/run.rs
```diff
@@ -268,23 +262,22 @@ impl Command {
...
-            if consensus_dev_key && trigger != Trigger::Never {
+            if debug {
                 let key = default_consensus_dev_key();
                 warn!(
                     "Fuel Core is using an insecure test key for consensus. Public key: {}",
                     key.public_key()
                 );
+                if trigger == Trigger::Never {
```
xgreenx, line 271: We need to log it in the case of suer provided key

```rust
if let Some(key) = load_consensus_key(consensus_key) {
    if trigger == Trigger::Never {
        warn!("Consensus key configured but block production is disabled!")
    }
}
```

bin/fuel-core/src/cli/run.rs
```diff
@@ -266,25 +260,25 @@ impl Command {
...
 
+        let consensus_key = load_consensus_key(consensus_key)?;
+        if consensus_key.is_some() && trigger == Trigger::Never {
+            warn!("Consensus key configured but block production is disabled!");
+        }
+
         // if consensus key is not configured, fallback to dev consensus key
-        let consensus_key = load_consensus_key(consensus_key)?.or_else(|| {
-            if consensus_dev_key && trigger != Trigger::Never {
+        let consensus_key = consensus_key.or_else(|| {
```
xgreenx, line 269: I would prefer to merge that into one "if/else" statement, but it's okay to keep it like this

xgreenx (approved):

PR #1616, opened 2024-01-24
Make `BlockHeader` a versionable enum

MitchTurner:
Part of: https://github.com/FuelLabs/fuel-core/issues/1544

### DIFF
--- a/crates/fuel-core/src/service/adapters/p2p.rs
+++ b/crates/fuel-core/src/service/adapters/p2p.rs
@@ -55,7 +55,7 @@ impl BlockHeightImporter for BlockImporterAdapter {
         Box::pin(
             BroadcastStream::new(self.block_importer.subscribe())
                 .filter_map(|result| result.ok())
-                .map(|result| result.sealed_block.entity.header().consensus().height),
+                .map(|result| *result.sealed_block.entity.header().height()),
         )
     }
 }
--- a/crates/fuel-core/src/service/adapters/p2p.rs
+++ b/crates/fuel-core/src/service/adapters/p2p.rs
@@ -55,7 +55,7 @@ impl BlockHeightImporter for BlockImporterAdapter {
         Box::pin(
             BroadcastStream::new(self.block_importer.subscribe())
                 .filter_map(|result| result.ok())
-                .map(|result| result.sealed_block.entity.header().consensus.height),
+                .map(|result| result.sealed_block.entity.header().consensus().height),
         )
     }
 }
--- a/crates/types/src/blockchain/header.rs
+++ b/crates/types/src/blockchain/header.rs
@@ -404,7 +404,7 @@ impl core::ops::Deref for BlockHeader {
     type Target = ApplicationHeader<GeneratedApplicationFields>;
 
     fn deref(&self) -> &Self::Target {
-        &self.application()
+        self.application()
     }
 }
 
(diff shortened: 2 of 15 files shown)


crates/types/src/blockchain/header.rs
```diff
@@ -37,6 +47,58 @@ pub struct BlockHeader {
...
+
+    /// Getter for application portion of header
+    pub fn application(&self) -> &ApplicationHeader<GeneratedApplicationFields> {
+        match self {
+            BlockHeader::V1(v1) => &v1.application,
+        }
+    }
+
+    /// Getter for metadata portion of header
+    pub fn metadata(&self) -> &Option<BlockHeaderMetadata> {
```
xgreenx, line 72: Metadata should be a private field. We don't want to expose=)

crates/types/src/blockchain/header.rs
```diff
@@ -37,6 +47,58 @@ pub struct BlockHeader {
...
+    pub fn application_mut(
+        &mut self,
+    ) -> &mut ApplicationHeader<GeneratedApplicationFields> {
+        match self {
+            BlockHeader::V1(v1) => &mut v1.application,
+        }
+    }
+
+    /// Mutable getter for metadata portion of header
+    pub fn metadata_mut(&mut self) -> &mut Option<BlockHeaderMetadata> {
```
xgreenx, line 95: And we definitely don't want to allow modification of the metadata. You use that only in one place and you can work directly with enum there.

crates/types/src/blockchain/header.rs
```diff
@@ -37,6 +47,58 @@ pub struct BlockHeader {
...
 
+impl From<BlockHeaderV1> for BlockHeader {
+    fn from(v1: BlockHeaderV1) -> Self {
+        BlockHeader::V1(v1)
+    }
+}
+
+impl BlockHeader {
+    /// Getter for consensus portion of header
+    pub fn consensus(&self) -> &ConsensusHeader<GeneratedConsensusFields> {
```
xgreenx, line 58: Hmm, the `BlockHeader` is mostly a composition of the consensus and application headers, and, likely, updating one of them will break these getters=)

Maybe we want to consider add getter for some specific fields like `applciation_hash`, `transactions_root`, etc.

In that case, maybe instead of providing a mutable reference getter like `applciation_hash_mut`, we want to provide an explicit setter(`set_applciation_hash`) instead.

MitchTurner: Yep. I agree. I'll do that.

MitchTurner: Hmm. Some of these might be better off under a test feature. On the other hand, I'm not exactly sure what the risk of having these mutable though.

crates/fuel-core/src/service/adapters/p2p.rs
```diff
@@ -55,7 +55,7 @@ impl BlockHeightImporter for BlockImporterAdapter {
         Box::pin(
             BroadcastStream::new(self.block_importer.subscribe())
                 .filter_map(|result| result.ok())
-                .map(|result| result.sealed_block.entity.header().consensus.height),
+                .map(|result| result.sealed_block.entity.header().consensus().height),
```
xgreenx, line 58:
```suggestion
                .map(|result| *result.sealed_block.entity.header().height()),
```

xgreenx (dismissed):

bvrooman (approved):

PR #1625, opened 2024-01-24
Making relayer independent from the executor

xgreenx:
It is preparation for the https://github.com/FuelLabs/fuel-core/issues/1568.

The changes make relayer storage independent from the executor storage. Before, the relayer and executor shared write ownership to the `Messages` table. The relayer was inserting new messages, and the executor was removing them.

With this change, only the executor modifies the `Messages` table(inserts and removes messages). The relayer has its own new `History` table, that stores all events from the DA layer per each height.

This change also makes the insertion of upcoming events from DA as part of the state transition, allowing in the future handle [force transaction inclusion](https://github.com/FuelLabs/fuel-core/issues/1626). 

The change:
- Adds blanket implementation for the `VmStorageRequirements` since the executor requires access to the `FuelBlocks` table, and we can inherit this implementation.
- Adds new tests for the executor that verifies fetching data from the relayer.
- Introduces a new general `Event` type for messages and forced transactions (in the future).


crates/fuel-core/src/executor.rs
```diff
@@ -2764,4 +2800,242 @@ mod tests {
...
+            db_transaction
+                .as_mut()
+                .storage::<FuelBlocks>()
+                .insert(&0.into(), &block)
+                .expect("Should insert genesis block without any problems");
+            db_transaction.commit().expect("Should commit");
+            db
+        }
+
+        fn add_message_to_relayer(db: &mut Database, message: Message) {
```
MitchTurner, line 2834: It seems like this is adding a message to the database, but the name is `add_message_to_relayer`. Is this name right?

xgreenx: With the https://github.com/FuelLabs/fuel-core/pull/1629, the name of the function will have more sense=) Since it will be `Database<Relayer>`

crates/services/executor/src/executor.rs
```diff
@@ -650,6 +655,48 @@ where
...
+        let block_height = *header.height();
+        let prev_block_height = block_height
+            .pred()
+            .ok_or(ExecutorError::ExecutingGenesisBlock)?;
+
+        let prev_block_header = block_st_transaction
+            .storage::<FuelBlocks>()
+            .get(&prev_block_height)?
+            .ok_or(ExecutorError::PreviousBlockIsNotFound)?;
+        let previous_da_height = prev_block_header.header().da_height;
```
MitchTurner, line 672: This made me smile. Get the current height -> get pred height -> get the da_block for pred height -> get da_block height. The pred height and the da_block height should be the same, but I guess you still want to do the `get` to make sure it exists.

I guess if there was an `exists(&prev_block_height)` it could be more explicit, but I don't care that much.

crates/services/relayer/README.md
```diff
@@ -2,8 +2,8 @@
 # Relayer
 
 The Relayer connects Fuel to the DA (data availability) layer contract.
-The primary functionality is to track the finality of blocks in the DA layer and download all log messages for blocks that are considered final.
-Messages are then made available to fuel via the database.
+The primary functionality is to track the finality of blocks in the DA layer and download all log events for blocks that are considered final.
```
MitchTurner, line 5: substrate

crates/services/relayer/src/ports/tests.rs
```diff
@@ -24,49 +29,93 @@ fn test_insert_messages() {
...
+
+    let db = MockStorage::default();
+
+    // When
     let mut db = db.into_transactional();
-    db.insert_messages(&5u64.into(), &messages[..5]).unwrap();
+    let result = db.insert_events(&last_height.into(), &events);
+
+    // Then
+    assert!(result.is_err());
```
MitchTurner, line 94: Hmm. I would say I'd prefer this checks that it's the correct Error variant, but `StorageError` is pretty type-erased :(

MitchTurner: Actually. I've been thinking about this a bit. I think it kinda makes sense that `StorageError` is somewhat erased, since it's up the implementer to create their own implementation-specific errors. I think it is still possible to define the port errors with some domain specificity. That probably falls out of the scope of this PR though.

crates/services/relayer/src/storage.rs
```diff
@@ -0,0 +1,166 @@
...
+impl TableWithBlueprint for RelayerMetadata {
+    type Blueprint = Plain<Postcard, Primitive<8>>;
+
+    fn column() -> Column {
+        Column::RelayerMetadata
+    }
+}
+
+/// The table contains history of the relayer.
+pub struct History;
```
MitchTurner, line 48: Is this just a placeholder for now?

MitchTurner, line 48: I think maybe `EventHistory` might be a better name.

MitchTurner:
I'm about ready to approve. Just had a couple minor things I pointed out.

crates/services/executor/src/executor.rs
xgreenx: The code here:
1. Get the current block height
2. Get the block height of the previous block(current - 1)
3. We get the da block height of the previous block.
4. We get the next unprocessed da block height starting from the previous block(prev_da_block_height + 1).
5. We process all DA blocks between `next_unprocessed_da_block_height`..=`current_da_block_height`.

We need to do something like this to handle cases:
1. The DA block height didn't change between fuel blocks.
2. The DA block height changed increased by X, where X > 0(it can be any value like 1, 10, or 1000000).

MitchTurner: OOooh. I misread one of the lines. This looks right.

MitchTurner (approved):

PR #1628, opened 2024-01-25
feat: Versionable CompressedCoin

bvrooman:


crates/types/src/entities/coins/coin.rs
```diff
@@ -40,21 +40,36 @@ pub struct Coin {
...
 /// the proper work of the blockchain.
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum CompressedCoin {
+    /// CompressedCoin Version 1
+    V1(CompressedCoinV1),
+}
+
+impl Default for CompressedCoin {
```
xgreenx, line 63: It is better to only allow to use `Default` implementation in the tests. In the real code we want to create explicitly specific version

crates/services/executor/src/executor.rs
```diff
@@ -1508,13 +1507,12 @@ where
...
         // if there's a revert or panic and shouldn't be added to the utxo set.
         if *amount > Word::MIN {
-            let coin = CompressedCoin {
-                owner: *to,
-                amount: *amount,
-                asset_id: *asset_id,
-                maturity: 0u32.into(),
-                tx_pointer: TxPointer::new(block_height, tx_idx),
-            };
+            let mut coin = CompressedCoin::default();
```
xgreenx, line 1510: We want to use explicitly `V1` here. Plus CI fails=) 
image

crates/fuel-core/src/service/genesis.rs
```diff
@@ -181,28 +181,33 @@ fn init_coin_state(
...
-                    owner: coin.owner,
-                    amount: coin.amount,
-                    asset_id: coin.asset_id,
-                    maturity: coin.maturity.unwrap_or_default(),
-                    tx_pointer: TxPointer::new(
-                        coin.tx_pointer_block_height.unwrap_or_default(),
-                        coin.tx_pointer_tx_idx.unwrap_or_default(),
-                    ),
-                };
+                let mut compressed_coin = CompressedCoin::default();
```
xgreenx, line 184: Here we also want to use `V1` explicitly

crates/services/executor/src/executor.rs
bvrooman: Oops, missed this!

MitchTurner (approved):
LGTM

xgreenx (approved):

PR #1629, opened 2024-01-25
Use a separate database for each data domain

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1568

The change splits the `Database` into 3 databases: 
- `Database<OnChain>` - Stores only data required for normal work of the blockchain.
- `Database<OffChain>` - Stores only data used by the off-chain services like GraphQL.
- `Database<Relayer>` - Stores relayer-related data like events(messages or transactions) from L1.

The `Database<Description>` type has a generic `Description` that implements the `DatabaseDescription` trait:

```rust
/// The description of the database that makes it unique.
pub trait DatabaseDescription: 'static + Clone + Debug + Send + Sync {
    /// The type of the column used by the database.
    type Column: StorageColumn + strum::EnumCount + enum_iterator::Sequence;
    /// The type of the height of the database used to track commits.
    type Height: Copy;

    /// Returns the expected version of the database.
    fn version() -> u32;

    /// Returns the name of the database.
    fn name() -> &'static str;

    /// Returns the column used to store the metadata.
    fn metadata_column() -> Self::Column;

    /// Returns the prefix for the column.
    fn prefix(column: &Self::Column) -> Option<usize>;
}
```

Each database has its folder, defined by the `DatabaseDescription::name`, where actual data is stored.

image

Each database has its own `Column` type that describes all columns, avoiding overlaps with other tables. The change updates a little bit `StrucutredStorage` implementation and `TableWithBlueprint` to be more flexible and use the `Column` defined by the table, instead of hardcoded `fuel_core_storage::column::Column`.

Other small changes:
- Unified the logic of storing the database's metadata. It will be useful for https://github.com/FuelLabs/fuel-core/issues/1589 to implement a unified `commit_chagnes` function.
- The `latest_height` function now uses the height from the metadata table.
- Removed relayers-related tables and columns from the `fuel-core-storage` crate.
- Removed part of GraphQL tables and columns from the `fuel-core-storage`. The last part will be removed during https://github.com/FuelLabs/fuel-core/issues/1583.
- Moved `tx_count` metrics from `BlockImporter` to GraphQL off-chain worker. Any statistic that requires a persistent state in the database may be done outside of the blockchain.
- Remove `chain_name` from the database. The `ConsensusParameters` already contains this information.
- Removed the `checkpoint` function from the `RocksDB` since it is not used. Later it will be added again back but with another implementation during https://github.com/FuelLabs/fuel-core/issues/1589.
- Removed `Column::ForeignColumn`, since each database has its own `Column` type. Removed the macro rules added to handle `ForeignColumn`.


crates/fuel-core/src/database.rs
```diff
@@ -66,25 +74,26 @@ use std::path::Path;
...
 pub mod storage;
 pub mod transaction;
 pub mod transactions;
 
-pub type Column = fuel_core_storage::column::Column;
-
 #[derive(Clone, Debug)]
-pub struct Database {
-    data: StructuredStorage<DataSource>,
+pub struct Database<Description = OnChain>
```
MitchTurner, line 92: Why be opinionated about a default value for `Description`?

crates/fuel-core/src/database.rs
```diff
@@ -443,24 +487,43 @@ impl AtomicView for RelayerReadDatabase {
...
+        use enum_iterator::all;
+        use fuel_core_storage::kv_store::StorageColumn;
+        use strum::EnumCount;
+        for column in all::<Description::Column>() {
+            assert!(column.as_usize() < Description::Column::COUNT);
+        }
+    }
+
+    #[test]
+    fn column_keys_not_exceed_count_test() {
```
MitchTurner, line 524: Nit: I'd prefer these each as separate tests.

crates/fuel-core/src/database/metadata.rs
```diff
@@ -1,127 +1,109 @@
...
+impl<Description> Mappable for MetadataTable<Description>
 where
-    V: Clone,
+    Description: DatabaseDescription,
 {
-    type Key = str;
-    type OwnedKey = String;
-    type Value = V;
-    type OwnedValue = V;
+    type Key = ();
```
MitchTurner, line 35: This feels like a smell to me. Does this need to be `Mappable`?

crates/fuel-core/src/graphql_api/storage.rs
```diff
@@ -0,0 +1,54 @@
...
+    /// See [`Receipts`](receipts::Receipts)
+    Receipts = 1,
+    /// The column of the table that stores `true` if `owner` owns `Coin` with `coin_id`
+    OwnedCoins = 2,
+    /// Transaction id to current status
+    TransactionStatus = 3,
+    /// The column of the table of all `owner`'s transactions
+    TransactionsByOwnerBlockIdx = 4,
+    /// The column of the table that stores `true` if `owner` owns `Message` with `message_id`
+    OwnedMessageIds = 5,
```
MitchTurner, line 31: What is the difference between a "Message" and an "Event". I remember in the last PR you changed a lot of "message"s to "event"s.

crates/fuel-core/src/database.rs
xgreenx: I did it to minimize the number of changes -> conflicts with regenesis PR

crates/fuel-core/src/graphql_api/storage.rs
xgreenx: The `Message` is something spendable on the Fuel blockchain, while the `Event` is something bridged from the L1. `Event::Message` and `Message` are basically the same, but we can have more events from L1(like forced transactions).

crates/fuel-core/src/database/metadata.rs
xgreenx: My idea was to have a compilation know metadata types instead of `(string, string)` pairs.

Since each database only has one metadata entry, we don't need something else than `()`=) But it works worse in the case of the huge metadata entry since it is slower to serialize/deserialize.

I'm open to any suggestions=) I only tried to make it more safe(type defined).

crates/fuel-core/src/combined_database.rs
```diff
@@ -0,0 +1,92 @@
...
+};
+use fuel_core_storage::Result as StorageResult;
+use fuel_core_types::{
+    blockchain::primitives::DaBlockHeight,
+    fuel_types::BlockHeight,
+};
+
+/// A database that combines the on-chain, off-chain and relayer databases into one entity.
+#[derive(Default, Clone)]
+pub struct CombinedDatabase {
```
MitchTurner, line 18: Why do these need to be combined? From the domain perspective, if there is an on-chain and an off-chain db, then they should be handed to the domain separately imo.

MitchTurner: I do see that it gets separated when you create the graph_ql service, for example. So I figure this is just a convenience thing for moving them around.

crates/fuel-core/src/state/in_memory/transaction.rs
```diff
@@ -41,22 +45,31 @@ use std::{
...
 
 #[derive(Debug)]
-pub struct MemoryTransactionView {
-    view_layer: MemoryStore,
+pub struct MemoryTransactionView<Description = OnChain>
+where
+    Description: DatabaseDescription,
+{
+    view_layer: MemoryStore<Description>,
     // TODO: Remove `Mutex`.
```
MitchTurner, line 53: Looks like you completed this `TODO`.

crates/storage/src/column.rs
```diff
@@ -4,190 +4,92 @@
 
 use crate::kv_store::StorageColumn;
 
-/// Helper macro to generate the `Column` enum and its implementation for `as_u32` method.
-macro_rules! column_definition {
```
MitchTurner, line 8: :fireworks:

crates/fuel-core/src/database/metadata.rs
MitchTurner: It's not the worst thing. Just seems like not everything needs to be `Mappable`. I prefer this over stringly-typing things for sure :+1:

MitchTurner (commented):
Looks like it's just that one comment that needs to be removed. Otherwise LGTM.

crates/fuel-core/src/state/in_memory/transaction.rs
xgreenx: No, that one we will be able to remove with https://github.com/FuelLabs/fuel-core/issues/1589 =)

MitchTurner: Oh. Lol. Misread that. NVM. I thought the `Mutex` was gone. Clearly not.

MitchTurner (approved):

PR #1631, opened 2024-01-26
feat: dry run multiple transactions

### LINKED ISSUE #1533: Add multi-call support to dry-run
See [eth_multicallV1](https://mirror.xyz/killaridev.eth/NXR9v4r8b4SWl-hbqJqRQyIAcE4GACobzMObREvL0fo) for inspiration.

This is essentially a collection of features, most importantly:

* Allow chaining multiple dependent txs in a single dry-run, and observe state afterwards
* Allow disabling or bypassing certain validation rules (e.g. signature checks)
* And possibly in the future, adding debug overrides to cryptographic opcodes (this might need Sway support as well)

matt-user:
closes #1533


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Description of the upcoming release here.
 ### Changed
 
 - [#1633](https://github.com/FuelLabs/fuel-core/pull/1633): Notify services about importing of the genesis block.
+- [#1631](https://github.com/FuelLabs/fuel-core/pull/1631): Modify api endpoint to dry run multiple transactions.
```
Dentosal, line 14: This is a breaking change, so it should be under `### Breaking`

crates/services/producer/src/block_producer.rs
```diff
@@ -178,23 +178,22 @@ where
     Executor: ports::DryRunner + 'static,
 {
     // TODO: Support custom `block_time` for `dry_run`.
-    /// Simulate a transaction without altering any state. Does not aquire the production lock
+    /// Simulates mutliple transactions without altering any state. Does not aquire the production lock
```
Dentosal, line 181:
```suggestion
    /// Simulates multiple transactions without altering any state. Does not acquire the production lock.
```

Dentosal (dismissed):

crates/client/src/client.rs
```diff
@@ -358,26 +361,33 @@ impl FuelClient {
     }
 
     /// Default dry run, matching the exact configuration as the node
-    pub async fn dry_run(&self, tx: &Transaction) -> io::Result<Vec<Receipt>> {
-        self.dry_run_opt(tx, None).await
+    pub async fn dry_run(
+        &self,
+        txs: &[&Transaction],
```
MitchTurner, line 366: Do you expect to already have a `Vec<&Transaction>`? I feel like this could just be `&[Transaction]`.

matt-user: hmm I think I did try to use `&[Transaction]` at first, but I was running into some errors.  I'll take another look at it.

MitchTurner: Happy to look at it with you.

matt-user: lol nvm, `&[Transaction]` totally works.  Thanks for the feedback :)

crates/fuel-core/src/schema/tx/types.rs
```diff
@@ -620,6 +624,71 @@ impl Transaction {
...
+
+pub struct DryRunTransaction(pub TransactionExecutionStatus);
+
+#[Object]
+impl DryRunTransaction {
+    async fn id(&self) -> TransactionId {
+        TransactionId(self.0.id)
+    }
+
+    async fn status(&self) -> async_graphql::Result<DryRunTransactionStatus> {
```
MitchTurner, line 683: I don't think these two need to be fallible. I think they are just doing a conversion.

crates/fuel-core/src/schema/tx/types.rs
```diff
@@ -620,6 +624,71 @@ impl Transaction {
...
+    async fn program_state(&self) -> Option<ProgramState> {
+        self.result.map(Into::into)
+    }
+
+    async fn reason(&self) -> String {
+        self.reason.clone()
+    }
+}
+
+pub struct DryRunTransaction(pub TransactionExecutionStatus);
```
MitchTurner, line 675: I'm not a big fan of this name, in reality it's still the status not the tx.

Also, should this be `pub`?

matt-user: We use the inner `pub` [here](https://github.com/FuelLabs/fuel-core/pull/1631/files#diff-2507047aabf2de10d20852af1dd31abceed668d66964340ce6fd0143d93c8a46R263)

I agree with your first point.  What alternative do you like `DryRunStatus` `DryRunTransactionStatus` `DryRunTransactionExecutionStatus`

matt-user: I forgot I already used `DryRunTransactionStatus` for an enum.  It's a bit verbose but maybe `DryRunTransactionExecutionStatus` is best?

crates/services/producer/src/block_producer.rs
```diff
@@ -203,25 +202,28 @@ where
...
-        if is_script && res.is_empty() {
-            return Err(anyhow!("Expected at least one set of receipts"))
+        let tx_statuses = tokio_rayon::spawn_fifo(
+            move || -> anyhow::Result<Vec<TransactionExecutionStatus>> {
+                Ok(executor.dry_run(component, utxo_validation)?)
+            },
+        )
+        .await?;
+
+        for (transaction, tx_status) in transactions.iter().zip(tx_statuses.iter()) {
```
MitchTurner, line 219: Nit: If you want to save yourself the `clone` of `transations` above, you could just `iter` over the `transactions` earlier to get the `is_script` for each of them. I'm only guessing that the `clone` is more expensive than the second loop of `transactions`.

MitchTurner: Very Nit: Then this for-loop could be replaced with:
```
let with_empty_reciepts = tx_statuses.iter().map(|status| status.receipts.is_empty());
if transactions.iter().zip(with_empty_receipts).any(|(is_script, is_empty)| is_script && is_empty) {
    return Err(anyhow!("Expected at least one set of receipts"))
}
```

crates/fuel-core/src/schema/tx/types.rs
MitchTurner: I prefer verbose over unclear :+1:

bin/e2e-test-client/src/tests/script.rs
```diff
@@ -67,7 +69,29 @@ pub async fn dry_run(ctx: &TestContext) -> Result<(), Failed> {
...
+// Dry run multiple transactions
+pub async fn dry_run_multiple_txs(ctx: &TestContext) -> Result<(), Failed> {
+    let transaction1 = tokio::time::timeout(
+        ctx.config.sync_timeout(),
+        ctx.alice.transfer_tx(ctx.bob.address, 0, None),
+    )
+    .await??;
+    let transaction2 = tokio::time::timeout(
+        ctx.config.sync_timeout(),
+        ctx.alice.transfer_tx(ctx.alice.address, 0, None),
```
MitchTurner, line 84: Idk if it matters, but did you mean to transfer from alice to alice?

MitchTurner (commented):
Looking good. What is there left to remove the "Draft"?

bin/e2e-test-client/src/tests/script.rs
matt-user: yeah, I had to make the two transactions different otherwise they would have the same Id and this would cause an error.  I'm not sure if there's a way to add a salt.  I also didn't know if these accounts actually had funds, but if they do we could use bob again.

crates/services/producer/src/block_producer.rs
```diff
@@ -216,14 +221,23 @@ where
...
-            let is_script = transaction.is_script();
-            if is_script && tx_status.receipts.is_empty() {
-                return Err(anyhow!("Expected at least one set of receipts"))
-            }
+        if are_txs_script
+            .iter()
+            .zip(tx_statuses.iter())
+            .any(|(is_script, tx_status)| *is_script && tx_status.receipts.is_empty())
+        {
+            return Err(anyhow!("Expected at least one set of receipts"));
```
MitchTurner, line 229: You don't need to `return` here, you can just have the else case be `Ok`.

crates/services/producer/src/block_producer.rs
```diff
@@ -216,14 +221,23 @@ where
...
         }
-
         Ok(tx_statuses)
+
+        // for (transaction, tx_status) in transactions.iter().zip(tx_statuses.iter()) {
```
MitchTurner: Left this commented code here.

crates/services/producer/src/block_producer.rs
```diff
@@ -194,6 +194,11 @@ where
                 .expect("It is impossible to overflow the current block height")
         });
 
+        let are_txs_script = transactions
+            .iter()
+            .map(|tx| tx.is_script())
+            .collect::<Vec<bool>>();
```
MitchTurner, line 200: You don't need to `collect` since you just call `iter` on it the next time you use it. Should improve the performance more.

MitchTurner (changes requested):

crates/services/producer/src/block_producer.rs
MitchTurner: Hmm. I just realized that this collect is essentially the same as a `clone` as both allocates to the heap. I'm p sure leaving it as an `iter` solves that problem :crossed_fingers:

Sorry to tie you down with these nits lol.

crates/services/producer/src/block_producer.rs
matt-user: interesting in other languages I was taught it is preferable to eliminate the else since it is unnecessary.  So in rust it is preferable to eliminate return statements instead of the else case?

crates/services/producer/src/block_producer.rs
matt-user: no worries I appreciate the feedback!

matt-user: hmm I don't think we can accomplish this without the `collect` as if we leave it as an `iter` it still points to the original `transactions` vec and eventually gets borrowed later in the if statement

MitchTurner: Sorry. Yeah. I was worried about that. I think your original clone is better than the collect. It's easier to read.

crates/services/producer/src/block_producer.rs
MitchTurner: Its preference. I just don't like returns. The compiled code won't be very different.

MitchTurner (approved):

PR #1635, opened 2024-01-29
Move updating of the owned messages and coins to off-chain worker

xgreenx:
Almost closes https://github.com/FuelLabs/fuel-core/issues/1583

The change moves the `OwnedCoins` and `OwnedMessageIds` tables and their implementation to the GraphQL module. Now, the off-chain worker is responsible for tracking owners of coins and messages. For that purpose, `Executor` generates execution events to track which messages/coins are consumed and created during execution. The worker uses these events and updates corresponding tables. 

It is not required to produce these events from the executor's perspective since anyone can calculate them based on the imported block and events from the relayer(knowing the execution rules of the state transition). However, I decided to embed that logic into the executor to simplify support for it. In the future, if we decide to change how coins and messages are created/spent, we don't need to update a bunch of places. 

Events from the `ExecutionResult` are insufficient to complete the change since we also need to support coins and messages from the genesis block. I added a new function into the `genesis` module to perform an update of the `OwnedCoins` and `OwnedMessageIds` tables from the `StateConfig`(in the future, it will be done by an off-chain regenesis process). The initial idea was to emit events, but after reviewing the https://github.com/FuelLabs/fuel-core/pull/1519 I realized that we can have so many events that it will be hard to fit them into the memory. So, I decided to implement a separate function that later can work with batches and be parallelizable.


crates/fuel-core/src/service.rs
```diff
@@ -253,13 +253,20 @@ impl RunnableService for Task {
...
-                let result = execute_genesis_block(&self.shared.config, &view)?;
+                let result = execute_genesis_block(&self.shared.config, &on_view)?;
 
                 self.shared.block_importer.commit_result(result).await?;
+
+                let off_chain_db_transaction = genesis::off_chain::execute_genesis_block(
+                    &self.shared.config,
+                    &off_view,
+                )?;
+                off_chain_db_transaction.commit()?;
```
Dentosal, line 269: Shouldn't we run both both offchain and onchain parts before committing either? Not sure if that actually matters.

crates/services/executor/src/executor.rs
```diff
@@ -1201,25 +1215,54 @@ where
...
+                    maturity,
+                    ..
+                }) => {
                     // prune utxo from db
-                    db.storage::<Coins>().remove(utxo_id)?;
+                    let coin = db
+                        .storage::<Coins>()
+                        .remove(utxo_id)?
+                        .map(Ok)
+                        .unwrap_or_else(|| {
```
Dentosal, line 1243: This doesn't look correct. `.map(Ok).unwrap_or_else(...)` looks like a no-op.

crates/types/src/services/executor.rs
```diff
@@ -45,6 +49,21 @@ pub struct ExecutionResult {
...
+#[derive(Debug, Clone)]
+pub enum Event {
+    /// Imported a new spendable message from the relayer.
+    NewMessage(Message),
+    /// The message was consumed by the transaction.
+    ConsumeMessage(Message),
+    /// Created a new spendable coin, produced by the transaction.
+    NewCoin(Coin),
+    /// The coin was consumed by the transaction.
+    ConsumeCoin(Coin),
```
Dentosal, line 66: nit: Naming seems inconsistent. I'd suggest `MessageImported` and `MessageConsumed`.

crates/fuel-core/src/service.rs
xgreenx: Each of them lives inside its own database, so we can't create one database transaction for both databases.

crates/fuel-core/src/coins_query.rs
```diff
@@ -921,6 +933,7 @@ mod tests {
         coin_result
     }
 
+    /// TODO: Should use any mock database instead of the `fuel_core::CombinedDatabase`.
```
MitchTurner, line 936:
```suggestion
    // TODO: Should use any mock database instead of the `fuel_core::CombinedDatabase`.
```

crates/services/executor/src/executor.rs
```diff
@@ -1189,25 +1203,55 @@ where
...
                     continue
                 }
                 Input::MessageCoinSigned(MessageCoinSigned { nonce, .. })
                 | Input::MessageCoinPredicate(MessageCoinPredicate { nonce, .. })
-                | Input::MessageDataSigned(MessageDataSigned { nonce, .. }) // Spend only if tx is not reverted
-                | Input::MessageDataPredicate(MessageDataPredicate { nonce, .. }) // Spend only if tx is not reverted
-                => {
+                | Input::MessageDataSigned(MessageDataSigned { nonce, .. })
+                | Input::MessageDataPredicate(MessageDataPredicate { nonce, .. }) => {
+                    // `MessageDataSigned` and `MessageDataPredicate` are spend only if tx is not reverted
```
Voxelot, line 1254:
```suggestion
                    // `MessageDataSigned` and `MessageDataPredicate` are spent only if tx is not reverted
```

Voxelot (dismissed):

Voxelot (approved):

PR #1637, opened 2024-01-29
Ubuntu vulnerability container image scan

rfuelsh:
Implement ubuntu vunerability container scan into container image build-publish process

actions is based on: https://github.com/marketplace/actions/container-scan#scan-image


xgreenx (commented):
Hmm, since the `crazy-max/ghaction-container-scan@v3` action is maintained by someone else, it potentially allows them to nullify this check by updating the action to do nothing. Maybe it may make sense to fork this action into the `Fuel` organization and use it instead.

rfuelsh:
@xgreenx - maybe that is something @Voxelot can chime in- i don't have those github permissions to fork repos I believe

Voxelot:
Agree - this action seems to have a small community. We should audit the action and likely fork it to ensure it doesn't become a malicious attack vector.

rfuelsh:
@Voxelot - would it better to investigate another action or can you assist with the forking on this action code repo? (I don't have the fork repo permissions I think)

Voxelot:
I've created a fork here: https://github.com/FuelLabs/ghaction-container-scan

Looking through the code it seems like a pretty lightweight wrapper around trivy, but we could have appsec review it as well.

xgreenx:
@rfuelsh @Voxelot Do we need this PR?

Issue #1647, opened 2024-02-05
Implement new `latest_gas_price` api


MitchTurner:
Add a new graphql endpoint that will return the gas price on the most recent block.

Should probably include the price as well as the block number.

Result of the research done here: https://github.com/FuelLabs/fuel-core/issues/1606

PR #1649, opened 2024-02-06
Latest gas price endpoint

MitchTurner:
Closes: https://github.com/FuelLabs/fuel-core/issues/1647

### DIFF
--- a/crates/client/src/client/types/gas_price.rs
+++ b/crates/client/src/client/types/gas_price.rs
@@ -2,7 +2,7 @@ use crate::client::schema;
 
 pub struct LatestGasPrice {
     pub gas_price: u64,
-    // pub block_height: u64,
+    pub block_height: u64,
 }
 
 // GraphQL Translation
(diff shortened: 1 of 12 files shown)


crates/fuel-core/src/schema/gas_price.rs
```diff
@@ -0,0 +1,51 @@
...
+};
+use async_graphql::{
+    Context,
+    Object,
+};
+use fuel_core_types::blockchain::block::Block;
+
+pub struct LatestGasPrice {
+    pub gas_price: U64,
+    pub block_height: U64,
```
bvrooman, line 17: Nit: This could probably be a `U32` since block height is based on a `u32`

crates/client/src/client/types/gas_price.rs
```diff
@@ -0,0 +1,16 @@
+use crate::client::schema;
+
+pub struct LatestGasPrice {
+    pub gas_price: u64,
+    pub block_height: u64,
```
bvrooman, line 5: Nit: This could be a u32

crates/fuel-core/src/schema/gas_price.rs
MitchTurner: Thanks. Nice catch. I meant to switch these earlier and I forgot. I can't remember why I didn't do `U32` to start out.

bvrooman (approved):

Issue #1648, opened 2024-02-05
Implement new `estimate_gas_price` api


MitchTurner:
It will take a block horizon and return what the most expensive the gas price will be over that horizon.

For now, it should just return the hard-coded value that is already available via the `NodeInfo`.

Followup to this work: https://github.com/FuelLabs/fuel-core/issues/1606

Curraso:
f

xgreenx:
@MitchTurner Do we need to rename `NodeInfo.min_gas_price` to `gas_price` as well?

PR #1650, opened 2024-02-06
Estimate gas price API Endpoint

MitchTurner:
Closes: https://github.com/FuelLabs/fuel-core/issues/1648

### DIFF
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -12,6 +12,7 @@ use crate::client::{
         TransactionId,
     },
     types::{
+        gas_price::LatestGasPrice,
         message::MessageStatus,
         primitives::{
             Address,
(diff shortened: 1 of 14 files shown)


crates/client/src/client.rs
```diff
@@ -354,6 +355,14 @@ impl FuelClient {
         self.query(query).await.map(|r| r.latest_gas_price.into())
     }
 
+    pub async fn estimate_gas_price(
+        &self,
+        block_horizon: u32,
```
bvrooman, line 360: Would it make more sense to accept `block_horizon` as an instance of `BlockHorizon` here, rather than converting in the function itself? Or maybe some `T` that can `.into()` into a `BlockHorizon`?

bvrooman (dismissed):

bvrooman (approved):

PR #1658, opened 2024-02-12
Fix for the race condition with tx status and receipts

xgreenx:
Example of the failed CI https://github.com/FuelLabs/fuel-core/actions/runs/7865306004/job/21471566710?pr=1656

The change moves receipts into the `TransationStatus` to avoid requests to the database. The `TxPool` directly listens for blocks and transaction statuses to notify the user via subscription. But when it notifies the users, the off-chain database can still be outdated, leading to empty receipts.

This PR fixes the race condition for the receipts. However, the problem still exists for the end user if they want to fetch some updated information from the off-chain database. I created a separate issue to track it: https://github.com/FuelLabs/fuel-core/issues/1659.

It is the fix for the https://github.com/FuelLabs/fuel-core/pull/1656


crates/fuel-core/src/graphql_api/storage.rs
```diff
@@ -19,18 +18,16 @@ pub mod transactions;
 pub enum Column {
     /// The column id of metadata about the blockchain
     Metadata = 0,
-    /// See [`Receipts`](receipts::Receipts)
-    Receipts = 1,
     /// The column of the table that stores `true` if `owner` owns `Coin` with `coin_id`
```
Dentosal, line 21: Hmmm I'm not sure if it actually makes sense to reassign all of the numbers instead just leaving a gap here

Dentosal (approved):
This is a readibility improvement even if it wasn't for the fix.

.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
-  RUST_VERSION: 1.73.0
+  RUST_VERSION: 1.74.0
```
MitchTurner, line 19: Idk if we need https://github.com/FuelLabs/fuel-core/pull/1656 if we want to just bump these here.

MitchTurner:
You said you ran this a bunch of times locally and it seemed to be fixed. I was never able to repro this failing test locally, did you ever see it fail locally?

xgreenx:
Yeah, I tried to run e2e tests in the loop and sometime it was failing=)

PR #1663, opened 2024-02-13
Soften txpool p2p reputation requirements

Voxelot:
It is fairly common for gossipped transactions to fail due to transient reasons such as UTXO's being spent before the transaction is included. A transaction that was valid when broadcast could become invalid if its inputs were spent in a newly confirmed block. Penalizing nodes for this could unfairly punish nodes that are simply "behind" in their view of the blockchain state.

In bitcoin, only incorrectly formatted transactions (i.e. invalid signatures or corrupt data according to consensus rules) is considered immediately punishable, and all transient failures are ignored. https://github.com/bitcoin/bitcoin/blob/6ff0aa089c01ff3e610ecb47814ed739d685a14c/src/net_processing.cpp#L1849

However, bitcoin does implement rate limiting on peers to prevent spam of either valid or invalid transactions to work around this. That is not currently implemented in this PR.

Note that in the event of transactions with high validation costs due to predicates, this would still be punishable.

### DIFF
--- a/crates/services/txpool/src/service.rs
+++ b/crates/services/txpool/src/service.rs
@@ -251,9 +251,11 @@ where
                                 Some(Ok(_)) => {
                                     GossipsubMessageAcceptance::Accept
                                 },
-                                Some(Err(_)) => {
+                                // Use similar p2p punishment rules as bitcoin
+                                // https://github.com/bitcoin/bitcoin/blob/6ff0aa089c01ff3e610ecb47814ed739d685a14c/src/net_processing.cpp#L1856
+                                Some(Err(Error::ConsensusValidity(_))) => {
                                     GossipsubMessageAcceptance::Reject
-                                }
+                                },
                                 _ => GossipsubMessageAcceptance::Ignore
                             }
                         }
@@ -355,7 +357,7 @@ where
     pub async fn insert(
         &self,
         txs: Vec<Arc<Transaction>>,
-    ) -> Vec<anyhow::Result<InsertionResult>> {
+    ) -> Vec<Result<InsertionResult, Error>> {
         // verify txs
         let current_height = *self.current_height.lock();
 
--- a/crates/types/src/services/txpool.rs
+++ b/crates/types/src/services/txpool.rs
@@ -33,7 +33,10 @@ use crate::{
     },
     services::executor::TransactionExecutionResult,
 };
-use fuel_vm_private::checked_transaction::CheckedTransaction;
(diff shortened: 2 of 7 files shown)


MitchTurner (commented):
Is this still WIP? It looks like you are just punishing in the case of a `CheckError`? Is that a good synopsis?

Voxelot:
> Is this still WIP? It looks like you are just punishing in the case of a `CheckError`? Is that a good synopsis?

Yep that's a good synopsis. I'll mark this as ready for now, the only remaining change I was thinking of fixing was making the new Database error variant more friendly to use.

crates/types/src/services/txpool.rs
```diff
@@ -306,7 +307,19 @@ pub enum Error {
     TTLReason,
     #[error("Transaction squeezed out because {0}")]
     SqueezedOut(String),
+    #[error("Invalid transaction data: {0:?}")]
+    ConsensusValidity(CheckError),
+    #[error("Mint transactions are disallowed from the txpool")]
+    MintIsDisallowed,
+    #[error("Database error: {0}")]
+    Database(String),
```
xgreenx, line 315: It seems you can remove `Other` variant=)

crates/types/src/services/txpool.rs
```diff
@@ -306,7 +307,19 @@ pub enum Error {
     TTLReason,
     #[error("Transaction squeezed out because {0}")]
     SqueezedOut(String),
+    #[error("Invalid transaction data: {0:?}")]
+    ConsensusValidity(CheckError),
+    #[error("Mint transactions are disallowed from the txpool")]
+    MintIsDisallowed,
```
xgreenx, line 313: I think we want to punish for that as well=)

xgreenx (commented):
Looks good, but we need to implement rate limiting before mainnet then

crates/types/src/services/txpool.rs
Voxelot: It's used in a couple edge cases still, since we didn't have anyhow to lean on anymore.

crates/services/txpool/src/service.rs
```diff
@@ -264,14 +264,13 @@ where
                         }
                     };
 
-                    if acceptance != GossipsubMessageAcceptance::Ignore {
```
Voxelot, line 265: By not notifying the ignore status, these messages will be forced to sit in a cache until they expire taking up gossipsub message capacity.

xgreenx (approved):

PR #1664, opened 2024-02-13
Fix for the long database initialization after restart

xgreenx:
The change adds 64 MB limit for the WAL file. Previously, it was unlimited, causing [long initialization of the database](https://github.com/facebook/rocksdb/wiki/Speed-Up-DB-Open#wal-replaying).

The change also bumps the `fuel-core` version to `0.22.1`. Later, we need to duplicate the same change to the `master` branch.

### DIFF
--- a/crates/fuel-core/src/state/rocks_db.rs
+++ b/crates/fuel-core/src/state/rocks_db.rs
@@ -138,11 +138,13 @@ impl RocksDb {
         let mut opts = Options::default();
         opts.create_if_missing(true);
         opts.set_compression_type(DBCompressionType::Lz4);
-        opts.set_bytes_per_sync(1 * 1024 * 1024);
+        opts.set_bytes_per_sync(1024 * 1024);
         // TODO: Use value from the database config from https://github.com/FuelLabs/fuel-core/pull/1519
         opts.set_max_total_wal_size(64 * 1024 * 1024);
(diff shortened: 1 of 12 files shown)


crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -137,6 +138,13 @@ impl RocksDb {
         let mut opts = Options::default();
         opts.create_if_missing(true);
         opts.set_compression_type(DBCompressionType::Lz4);
+        opts.set_bytes_per_sync(1024 * 1024);
+        // TODO: Use value from the database config from https://github.com/FuelLabs/fuel-core/pull/1519
```
MitchTurner, line 142: This should probably have a follow-up item.

MitchTurner (changes requested):
LGTM, just one recommendation. 

What led you to go with 64 MB specifically? That sounds kinda small. Also, the values you chose for the other opts. I don't really have any context for what is good on those.

xgreenx:
As I mentioned on the call, I played with beta 4 synchronization and tried to use different values. With 64 MB the opening of the database doesn't have any delays, while increasing this value adds several seconds to the opening of the database. So basically it is an arbitrary value based on local observerations=D

By making it customizable we can control that behavior later.

MitchTurner (approved):

PR #1671, opened 2024-02-15
Extraction of the `FuelBlockSecondaryKeyBlockHeights` table to an off-chain worker

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1583

The change moves the `FuelBlockSecondaryKeyBlockHeights` table and related logic to the off-chain worker. Along with this, the change adds a new `Merklelized` blueprint that maintains the binary Merkle tree over the storage data. It supports only the insertion of the objects without removing them. This blueprint replaces the logic that previously was defined in the `Database`.

Some side changes caused by the main change:
- Now, each blueprint provides its own set of tests related to the logic of this blueprint.
- The `TransactionStatus` uses `block_height` inside instead of the `block_id`.
- The key for the dense merkle tree looks like:
  ```rust
  pub enum DenseMetadataKey<PrimaryKey> {
      /// The primary key of the `DenseMerkleMetadata`.
      Primary(PrimaryKey),
      #[default]
      /// The latest `DenseMerkleMetadata` of the table.
      Latest,
  }
  ```


crates/fuel-core/Cargo.toml
```diff
@@ -46,6 +46,7 @@ rand = { workspace = true }
 rocksdb = { version = "0.21", default-features = false, features = [
     "lz4",
     "multi-threaded-cf",
+    "jemalloc",
```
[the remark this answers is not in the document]
Voxelot: This is feature flagged using `rocksdb-production = ["rocksdb", "rocksdb/jemalloc"]`, to reduce build times of using rocksdb if not using production flag.

crates/client/src/client.rs
```diff
@@ -745,9 +745,12 @@ impl FuelClient {
         Ok(block)
     }
 
-    pub async fn block_by_height(&self, height: u32) -> io::Result<Option<types::Block>> {
+    pub async fn block_by_height(
```
MitchTurner, line 748: The types on `latest_gas_price` include a `u32`  to represent block height. Should we use `BlockHeight` there?

crates/storage/src/blueprint/merklelized.rs
```diff
@@ -0,0 +1,490 @@
...
+            Self::remove(storage, key_bytes.as_ref(), column)?;
+        }
+        Ok(())
+    }
+}
+
+/// The macro that generates basic storage tests for the table with the merklelized structure.
+/// It uses the [`InMemoryStorage`](crate::structured_storage::test::InMemoryStorage).
+#[macro_export]
+macro_rules! basic_merklelized_storage_tests {
```
MitchTurner, line 290: Should these macros be behind a `test-features` flag?

crates/types/src/blockchain/header.rs
```diff
@@ -112,26 +112,31 @@ impl BlockHeader {
     /// Set the block height for the header
     pub fn set_block_height(&mut self, height: BlockHeight) {
         self.consensus_mut().height = height;
+        self.recalculate_metadata();
```
MitchTurner, line 115: It looks like we have this method in a separate `impl` block for some reason. Both blocks for `BlockHeader` have the same (no) constraints. Can we combine the `impl` blocks?

crates/storage/src/tables.rs
```diff
@@ -132,6 +132,31 @@ pub mod merkle {
...
+    /// The key for the corresponding `DenseMerkleMetadata` type.
+    /// The `Latest` variant is used to have the access to the latest dense Merkle tree.
+    #[derive(Default, Debug, Clone, serde::Serialize, serde::Deserialize)]
+    pub enum DenseMetadataKey<PrimaryKey> {
+        /// The primary key of the `DenseMerkleMetadata`.
+        Primary(PrimaryKey),
+        #[default]
+        /// The latest `DenseMerkleMetadata` of the table.
+        Latest,
+    }
```
MitchTurner, line 144: I want to understand this better. If `Latest` gives you the entire latest Merkle tree, the `PrimaryKey` gives you a previous tree? Or a subsection?

Maybe this would be obvious if I understood the MTs terminology better.

crates/fuel-core/src/database/block.rs
```diff
@@ -110,40 +76,14 @@ impl StorageMutate<FuelBlocks> for Database {
...
-        // Generate new metadata for the updated tree
-        let root = tree.root();
-        let version = tree.leaves_count();
-        let metadata = DenseMerkleMetadata::new(root, version);
-        self.storage::<FuelBlockMerkleMetadata>()
-            .insert(height, &metadata)?;
-
         // TODO: Temporary solution to store the block height in the database manually here.
         //  Later it will be controlled by the `commit_changes` function on the `Database` side.
         //  https://github.com/FuelLabs/fuel-core/issues/1589
```
Voxelot, line 81: is this comment still relevant? It looks like we're storing by id vs height here now.

crates/types/src/blockchain/header.rs
```diff
@@ -112,26 +112,31 @@ impl BlockHeader {
...
     }
 
     /// Set the previous root for the header
     pub fn set_previous_root(&mut self, root: Bytes32) {
         self.consensus_mut().prev_root = root;
+        self.recalculate_metadata();
```
Voxelot, line 121: Should we use a builder in the hot paths instead of calling these setters one by one, now that they all have to recalculate the application header hash? We could save several hashing calls in the executor by calculating this after all the fields are set.

crates/types/src/services/txpool.rs
```diff
@@ -180,7 +180,7 @@ pub enum TransactionStatus {
     /// Transaction was successfully included in a block
     Success {
         /// Included in this block
-        block_id: BlockId,
+        block_height: BlockHeight,
```
Voxelot, line 183: If there's a possibility of reorgs in the future, would only using block height on the api be problematic?

crates/storage/src/blueprint/merklelized.rs
```diff
@@ -0,0 +1,490 @@
...
+    ($table:ident, $key:expr, $value_insert:expr, $value_return:expr, $random_key:expr) => {
+        $crate::paste::item! {
+        #[cfg(test)]
+        #[allow(unused_imports)]
+        mod [< $table:snake _basic_tests >] {
```
Voxelot, line 295: These look identical to the basic storage tests, should we just reuse those here?

Also, perhaps we should have some merkle tree related tests as part of this macro?

crates/types/src/services/txpool.rs
xgreenx: With a shared sequencer, I don't see how we can occur fork with several valid blocks for the same height. It is possible to have invalid blocks and valid blocks, but we will skip invalid due to invalidity and commit only a valid one.

crates/storage/src/blueprint/merklelized.rs
xgreenx: The tests are not the same since the removal of elements is not allowed=)

I moved old tests for the `FuelBlocks` table in the `blocks.rs` instead of adding new tests here. I did it for the simplicity of the PR, but if you think it is better to make it part of this macro now, then later, I will do it=)

crates/types/src/blockchain/header.rs
xgreenx: We use these methods only in the tests, so it doesn't hurt us=) I will put them under `test-helper` feature

crates/types/src/blockchain/header.rs
xgreenx: They are segregated by their business logic. I'm okay with joining them into one/two impl blocks, but since they are in different places in the file, I prefer to not shuffle them in this PR to minimize number of changes=)

crates/storage/src/tables.rs
xgreenx: Yes. The Merkle tree that we are using here is append-only. Because of that, we can only add new stuff at the end, and the `Latest` key allows us to do that. At the same time, appending a new leaf doesn't invalidate all previous Merkle trees, and `PrimaryKey` allows us to get the valid `MerkleTree` at some `Key` that we added before. For example, in the case of the `FuelBlocks`, we can get a valid Merkle tree at height X, X-1, .., X - Y and use it in proofs that some block is part of the Merkle tree at a specific height.

crates/types/src/services/txpool.rs
Voxelot: so if there are several proposed block ids for the same block height, and a user tries to see the block containing their tx, the tx might not exist in the block that actually gets accepted right?

xgreenx: Yes, the transaction has `Committed` status only when the corresponding block is finalized by the shared sequencer.

crates/client/src/client/schema/tx.rs
```diff
@@ -178,7 +178,7 @@ pub struct SubmittedStatus {
 #[cynic(schema_path = "./assets/schema.sdl")]
 pub struct SuccessStatus {
     pub transaction_id: TransactionId,
-    pub block: BlockIdFragment,
+    pub block: BlockHeightFragment,
```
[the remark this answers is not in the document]
xgreenx: It is named `block` since from the GraphQL perspective you receive the block=) The `block_height` is ID, but it could be `BlockId` as well(both works)

crates/storage/src/column.rs
```diff
@@ -62,10 +62,8 @@ pub enum Column {
     // Below are the tables used for p2p, block production, starting the node.
     /// The column id of metadata about the blockchain
     Metadata = 17,
-    /// See `FuelBlockSecondaryKeyBlockHeights`
-    FuelBlockSecondaryKeyBlockHeights = 18,
     /// See [`SealedBlockConsensus`](crate::tables::SealedBlockConsensus)
-    FuelBlockConsensus = 19,
+    FuelBlockConsensus = 18,
```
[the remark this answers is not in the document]
xgreenx: Since we don't have mainnet it's okay to break the numbering and invalidate the database. But yeah, later we will try to avoid breaking stuff like this

crates/fuel-core/src/query/block.rs
```diff
@@ -16,18 +19,15 @@ use fuel_core_types::{
...
 };
 
 pub trait SimpleBlockData: Send + Sync {
     fn block(&self, id: &BlockHeight) -> StorageResult<CompressedBlock>;
-
-    fn block_by_id(&self, id: &BlockId) -> StorageResult<CompressedBlock>;
 }
 
-impl<D: OnChainDatabase + ?Sized> SimpleBlockData for D {
+impl<D: OnChainDatabase + OffChainDatabase + ?Sized> SimpleBlockData for D {
```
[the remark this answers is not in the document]
Voxelot: It looks like this constraint isn't even needed. I removed it from here and line 53 `impl<D: OnChainDatabase + ?Sized> BlockQueryData for D {` and the tests passed.

crates/fuel-core/src/graphql_api/storage/blocks.rs
```diff
@@ -0,0 +1,42 @@
...
+use fuel_core_types::{
+    blockchain::primitives::BlockId,
+    fuel_types::BlockHeight,
+};
+
+/// The table of fuel block's secondary key - `BlockId`.
+/// It links the `BlockId` to corresponding `BlockHeight`.
+pub struct FuelBlockSecondaryKeyBlockHeights;
+
+impl Mappable for FuelBlockSecondaryKeyBlockHeights {
```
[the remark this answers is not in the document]
Voxelot: maybe FuelBlockIdsToHeights?

crates/client/src/client/schema/tx.rs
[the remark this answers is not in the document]
Voxelot: Agreed - since this could return a full block, we must use a fragment query here to only return a partial view of the object.

crates/types/src/blockchain/header.rs
MitchTurner: Okay. I just got confused where a method was just reading the PR diffs, but with an IDE it doesn't really affect anything.

Voxelot (approved):

PR #1674, opened 2024-02-16
Update verify_chart_version.sh

Kewe63:
File Existence Check: in the first code there is no function or direct command to check the existence of certain files.  But in the code I added a function called file_exists to check for the existence of certain files.


xgreenx:
Thank you for the contribution, but with new plans regarding operator and deployment, we don't need this action. Soon we will remove it=)

Issue #1652, opened 2024-02-08
Simplify GraphQL interfaces


Voxelot:
We should simplify some of the graphql interfaces to avoid issues like this from happening: https://github.com/FuelLabs/fuels-ts/issues/1706

Specifically, users shouldn't need to be able to query every field of a contract within a receipt, and this could just be simplified to returning only the ContractId.

PR #1675, opened 2024-02-17
Remove nested contract resolvers in GQL schema

Voxelot:
closes: #1652

Having the ability to fetch all data related to a contract is overkill and costly in most cases, and usually done by accident. This api change simplifies the GQL schema to simply return a contractId instead of a contract resolver for tx inputs, receipts, and more.

Note that the full contract details (such as the contract bytecode) can still be fetched directly via the contract query endpoint.


Torres-ssf (approved):
LGTM 🚀

xgreenx (approved):

xgreenx:
Do we want to create a breaking patch for beta 5?

Voxelot:
> Do we want to create a breaking patch for beta 5?

It's been addressed at the sdk level for now and we shouldn't push any breaking changes as a patch to beta-5.

Issue #623, opened 2022-09-15
Review GraphQL endpoints to avoid node overload


xgreenx:
Some GraphQL endpoints have complexity `O(n)`, `O(n^2)`, where `n` can be a number of arguments or internal data stored in the blockchain.

In the mainnet, it can affect GraphQL nodes and cause them to be stuck.

Each request should have some limitations. It can be a limitation of input or output arguments or query frequency.

The code has many simple queries (like getting block by height, transaction by id, and so on). The DevOps and infrastructure can limit the rates for those requests(for example, Apollo, but it requires investigation). So we don't need to modify the code and only care for it during cluster deployment.

We need to review all queries and decide which queries should be limited in the code(we need to implement some custom logic, maybe pagination, etc.) and which by the infra team(also, we need to decide how to do that).

ControlCplusControlV:
GraphQL endpoints should probably also implement some size check on request bodies to prevent that DOS vector

PR #1676, opened 2024-02-17
Prevent api spam with GQL complexity limits

### LINKED ISSUE #1651: Implement query & depth restrictions on the GQL interface
## Overview

Currently, GraphQL queries are unlimited in depth and complexity. It is possible to request all blocks with all transactions and receipts in one query causing OOM on the `fuel-core` node(plus it consumes a lot of execution).

## Solution

We need to implement limits on the complexity and depth of graphql queries to protect the APIs from DOS attacks. This will need to be done carefully, ensuring that the SDK isn't rendered unusable by excessive limits. 

## Implementation details

https://async-graphql.github.io/async-graphql/en/depth_and_complexity.html

Voxelot:
Closes #1651
Closes https://github.com/FuelLabs/fuel-core/issues/623

Updated the code to use "complexity" for all queries with context.
Moved all related fields/arguments into GraphQL config/arguments.
Added new CLI arguments:
- graphql-max-depth
- graphql-max-complexity
- graphql-max-recursive-depth

Changed `api-request-timeout` to be `30s`.
Breaking:
- Changed the `Block` request to only fetch transaction IDs.
- Changed the `SuccessStatus` and `FailureStatus` requests to only fetch block height.


xgreenx (approved):

crates/fuel-core/src/graphql_api.rs
```diff
@@ -13,13 +16,54 @@ pub(crate) mod view_extension;
...
+    estimate_predicates: 3000,
+    dry_run: 3000,
+    submit: 5000,
+    submit_and_await: 10000,
+    status_change: 10000,
+    raw_payload: 10,
+    storage_read: 10,
+    storage_iterator: 100,
+    bytecode_read: 2000,
+};
```
xgreenx, line 57: Default gas costs for main operations/queries.

xgreenx (approved):

xgreenx (dismissed):

crates/client/assets/schema.sdl
```diff
@@ -50,6 +50,7 @@ type Block {
 	height: U32!
 	header: Header!
 	consensus: Consensus!
+	transactionsIds: [TransactionId!]!
```
Dentosal, line 53:
```suggestion
	transactionIds: [TransactionId!]!
```

crates/client/src/client/schema/block.rs
```diff
@@ -90,7 +87,7 @@ pub struct Block {
     pub id: BlockId,
     pub header: Header,
     pub consensus: Consensus,
-    pub transactions: Vec<TransactionIdFragment>,
+    pub transactions_ids: Vec<TransactionId>,
```
Dentosal, line 90:
```suggestion
    pub transaction_ids: Vec<TransactionId>,
```

crates/fuel-core/src/schema/block.rs
```diff
@@ -115,12 +117,24 @@ impl Block {
...
     }
 
+    #[graphql(complexity = "QUERY_COSTS.storage_read + child_complexity")]
     async fn consensus(&self, ctx: &Context<'_>) -> async_graphql::Result<Consensus> {
         let query: &ReadView = ctx.data_unchecked();
         let height = self.0.header().height();
         Ok(query.consensus(height)?.try_into()?)
     }
 
+    async fn transactions_ids(&self) -> Vec<TransactionId> {
```
Dentosal, line 127:
```suggestion
    async fn transaction_ids(&self) -> Vec<TransactionId> {
```

Dentosal (approved):

Dentosal:
Approved as-is, but maybe it would be nice to have some tests that actually verify all of the endpoints respect the timeouts and such.

Voxelot:
@xgreenx these breaking changes seem pretty significant. I think there may be a way to allow users to restrict the complexity of their queries without breaking the api.

crates/client/src/client/schema/snapshots/fuel_core_client__client__schema__block__tests__block_by_height_query_gql_output.snap
```diff
@@ -35,8 +35,6 @@ query($height: U32) {
         signature
       }
     }
-    transactions {
-      id
-    }
+    transactionIds
```
Voxelot, line 38: why are we making a separate field for these ids?

crates/fuel-core/src/schema/balance.rs
```diff
@@ -65,10 +67,9 @@ impl BalanceQuery {
         Ok(balance)
     }
 
-    // TODO: We can't paginate over `AssetId` because it is not unique.
-    //  It should be replaced with `UtxoId`.
-    //  This API should be migrated to the indexer for better support and
+    // TODO: This API should be migrated to the indexer for better support and
     //  discontinued within fuel-core.
+    #[graphql(complexity = "QUERY_COSTS.balance_query")]
```
Voxelot, line 72: Shouldn't this be multiplied by the number of assets? Ie if you request your first 10 balances, it should be 10 * balance_query

xgreenx: The code inside iterates over all assets in the storage regardless of input parameters.

crates/client/src/client/schema/snapshots/fuel_core_client__client__schema__block__tests__block_by_height_query_gql_output.snap
xgreenx: In this case we don't resolve the `transactions` field. Resolving of the field fetch al transactions from the storage.

image

xgreenx:
> @xgreenx these breaking changes seem pretty significant. I think there may be a way to allow users to self restrict the complexity of their queries without breaking the api.
>
>at this stage, huge changes like this could break integrations with indexers or others. If we're going to drastically change the kinds of allowed queries, this should be held off to V2.

Indexers can just increase the allowed complexity. You can see an example in the test, where I increased it to `1_000_000`. To be able to request a lot of blocks.

xgreenx (approved):

tests/tests/dos.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    let node = FuelService::new_node(Config::local_node()).await.unwrap();
+    let url = format!("http://{}/v1/graphql", node.bound_address);
+
+    let result = send_query(&url, query).await;
+    assert!(result.contains("transactions"));
+}
+
+#[tokio::test]
+async fn complex_queries__50_block__query_to_complex() {
```
MitchTurner, line 110:
```suggestion
async fn complex_queries__50_block__query_too_complex() {
```

crates/fuel-core/src/schema/balance.rs
```diff
@@ -50,6 +51,7 @@ pub struct BalanceQuery;
 
 #[Object]
 impl BalanceQuery {
+    #[graphql(complexity = "QUERY_COSTS.balance_query")]
```
MitchTurner, line 54: I'm confused where these `complexity` values are being parsed. I guess I'm not familiar enough with the `async_graphql::Object` api, but looking through the docs I can't find where you can extend it with custom fields like `complexity`.

MitchTurner (commented):
Looking good. Just want to understand something first.

crates/fuel-core/src/schema/balance.rs
Voxelot: It's built into the proc-macro system: https://async-graphql.github.io/async-graphql/en/depth_and_complexity.html

crates/fuel-core/src/graphql_api.rs
```diff
@@ -13,13 +16,58 @@ pub(crate) mod view_extension;
...
+
+pub const QUERY_COSTS: Costs = Costs {
+    // balance_query: 4000,
+    balance_query: 10001,
+    coins_to_spend: 10001,
+    // get_peers: 2000,
+    get_peers: 10001,
+    // estimate_predicates: 3000,
+    estimate_predicates: 10001,
+    dry_run: 3000,
```
Voxelot, line 52: Isn't a dry-run potentially more expensive than estimating predicates?

crates/fuel-core/src/graphql_api.rs
```diff
@@ -13,13 +16,58 @@ pub(crate) mod view_extension;
...
+    pub status_change: usize,
+    pub raw_payload: usize,
+    pub storage_read: usize,
+    pub storage_iterator: usize,
+    pub bytecode_read: usize,
+}
+
+pub const QUERY_COSTS: Costs = Costs {
+    // balance_query: 4000,
+    balance_query: 10001,
```
Voxelot, line 46: [nit] maybe we should have a constant for the default max complexity, and then use that here (ie. MAX_COMPLEXITY_DEFAULT / 2 + 1)

crates/fuel-core/src/graphql_api.rs
```diff
@@ -13,13 +16,58 @@ pub(crate) mod view_extension;
...
+    // get_peers: 2000,
+    get_peers: 10001,
+    // estimate_predicates: 3000,
+    estimate_predicates: 10001,
+    dry_run: 3000,
+    // submit: 5000,
+    submit: 10001,
+    submit_and_await: 10001,
+    status_change: 10001,
+    raw_payload: 10,
```
Voxelot, line 57: a raw payload could also include the same complexity as a bytecode if a block is full of create txs right?

crates/fuel-core/src/graphql_api.rs
xgreenx: It is with more transactions, but we want to allow the use of the feature with several dry runs.

xgreenx (approved):

PR #1693, opened 2024-02-21
Regenesis support

MujkicA:
Closes https://github.com/FuelLabs/fuel-core/issues/1209

The change separates the initial chain state from the chain config and stores them in separate files when generating a snapshot. The state snapshot can be generated in a new format where parquet is used for compression and indexing while postcard is used for encoding. This enables importing in a stream like fashion which reduces memory requirements. Json encoding is still supported to enable easy manual setup. However, parquet is prefered for large state files.

   ### Snapshot command

   The CLI was expanded to allow customizing the used encoding. Snapshots are now generated along with a metadata file describing the encoding used. The metadata file contains encoding details as well as the location of additional files inside the snapshot directory containing the actual data. The chain config is always generated in the JSON format.

   The snapshot command now has the '--output-directory' for specifying where to save the snapshot.

   ### Run command

   The run command now includes the 'db_prune' flag which when provided will prune the existing db and start genesis from the provided snapshot metadata file or the local testnet configuration.

   The snapshot metadata file contains paths to the chain config file and files containing chain state items (coins, messages, contracts, contract states, and balances), which are loaded via streaming.

   Each item group in the genesis process is handled by a separate worker, allowing for parallel loading. Workers stream file contents in batches.

   A database transaction is committed every time an item group is succesfully loaded. Resumability is achieved by recording the last loaded group index within the same db tx. If loading is aborted, the remaining workers are shutdown. Upon restart, workers resume from the last processed group.

   ### Contract States and Balances

   Using uniform-sized batches may result in batches containing items from multiple contracts. Optimal performance can presumably be achieved by selecting a batch size that typically encompasses an entire contract's state or balance, allowing for immediate initialization of relevant Merkle trees.


crates/chain-config/src/config/snapshot_metadata.rs
```diff
@@ -0,0 +1,450 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct SnapshotMetadata {
+    metadata: Metadata,
+}
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, PartialEq)]
+struct Metadata {
+    chain_config: PathBuf,
```
xgreenx, line 95: Hmm, why do we need this field at all if we are using `CHAIN_CONFIG_FILENAME`?

tests/tests/balances.rs
```diff
@@ -38,44 +40,41 @@ async fn balance() {
...
-                    data: vec![],
-                    da_height: DaBlockHeight::from(1usize),
-                })
-                .collect(),
-        ),
-    });
+        ..Default::default()
+    };
+    let config = Config {
+        state_reader: StateReader::in_memory(state_config, MAX_GROUP_SIZE),
```
xgreenx, line 75: We propagate `MAX_GROUP_SIZE` in a lot of places. It would be nice if we somehow only defined it in one place and were using it everywhere. I mean, maybe `StateReader::in_memory` may don't require it

crates/chain-config/src/config/contract.rs
```diff
@@ -1,84 +1,84 @@
...
-    /// UtxoId: auto-generated if None
-    #[serde_as(as = "Option<HexType>")]
-    #[serde(default)]
-    pub tx_id: Option<Bytes32>,
-    /// UtxoId: auto-generated if None
-    #[serde_as(as = "Option<HexNumber>")]
-    #[serde(default)]
-    pub output_index: Option<u8>,
-    /// TxPointer: auto-generated if None
+    #[serde(default = "random_tx_id")]
```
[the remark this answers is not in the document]
MujkicA: My thinking was it can be made deterministic by providing the value inside the config if need be

tests/tests/balances.rs
MujkicA: I reckon we still want to control how large the batches are that end up in db transactions. I've made it optional for now and default to MAX on None so we avoid propagating. What do you think?

xgreenx: Yeah, it works. Also we can add a separate method for tests=)

crates/chain-config/src/config/contract_state.rs
```diff
@@ -0,0 +1,26 @@
...
+use serde::{
+    Deserialize,
+    Serialize,
+};
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, PartialOrd, Ord)]
+pub struct ContractStateConfig {
+    pub contract_id: ContractId,
+    pub key: Bytes32,
+    pub value: Bytes32,
```
Dentosal, line 14: Potential conflict with https://github.com/FuelLabs/fuel-vm/pull/671

crates/fuel-core/src/service/genesis/workers.rs
```diff
@@ -0,0 +1,308 @@
...
+use fuel_core_types::fuel_types::BlockHeight;
+use tokio::sync::Notify;
+use tokio_util::sync::CancellationToken;
+
+pub struct GenesisWorkers {
+    db: Database,
+    cancel_token: CancellationToken,
+    block_height: BlockHeight,
+    state_reader: StateReader,
+    finished_signals: [Arc<Notify>; 5],
```
Dentosal, line 41: Using a struct with name for each signal would feel less error-prone.

Cargo.lock
```diff
@@ -95,6 +97,21 @@ version = "0.2.16"
...
+checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
+
+[[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
+dependencies = [
+ "libc",
+]
```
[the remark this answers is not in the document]
MujkicA: parquet depends on chrono which for some reason also pulls in the android platform specific deps.
But the same doesn't happen when chrono is present as a sub dependancy through ethers for example. 
I'm not sure why this is happening.

tests/tests/balances.rs
MujkicA: Good point, the repeated `None` wasn't pretty either

xgreenx (approved):
Thank you for the huge work, it is really nice! Sorry for so many conflicts=) I added several TODOs that we can address in the follow-up PRs=)

I'm waiting for performance results and e2e from https://github.com/FuelLabs/fuel-core-regenesis to see how fast is regenesis=) But it is really great that you implemented interrupt/resume functionality.

Issue #1189, opened 2023-05-25
Prepare the executor for parallelisation


xgreenx:
Update API of the `TxPool` to return `Vec<Vec<CheckedTransaction>>` batched by dependencies for parallel execution(but now only one batch).
On the `executor` side use `rayon` to execute them in one thread.

Voxelot:
related #1088

Issue #1088, opened 2023-03-30
Parallel transaction execution within blocks


Voxelot:
One of the key benefits of fuel's transaction format and architecture is the capability to parallelize transactions execution. This issue is intended to cover the initial design and considerations for implementing parallel execution, but will likely be comprised of many smaller tasks. This design is subject to change based on feedback and discovery.

### Scheduling

Transactions are schedule-able for parallelization according to their input and output UTXOs, which could be considered as a resource access list. If any resource overlap exists between transactions, then those will need to be grouped and executed sequentially. Any subsequent overlapping transactions will be added to this concurrency grouping.

Resource overlap conditions:
 - Matching contract ID
 - Coin input that depends on another unprocessed output.

![Parallel Transactions](https://user-images.githubusercontent.com/794823/228685590-0eda9f3e-0252-45dd-8296-0b88b1585ed5.jpeg)

This diagram illustrates the optimal grouping of transactions. Also note in Group B, transactions which have no direct dependency on each other will still need to be ordered within the same group if there is a transitive dependency between them.

### Gas Limits & Block Capacity

In traditional blockchain architectures, each block has a total gas limit which is used to bound execution time. However, a single global limit doesn't take into account the level of concurrency. If the block gas limit is set based upon the desired execution deadline for single threaded execution, then the overall throughput would not be increased by making execution parallel. Similarly, if the gas limit of a block is taken simply as the `single threaded limit * number of cores`, this may allow blocks to exceed their desired execution deadline if all the transactions are combined into one concurrency group due to shared resource access.

In order to correctly bound execution time, our block gas limit should be considered as the limit per concurrency group. A max concurrency limit should also be provided as a consensus rule. If there are more concurrency groups than the concurrency limit, then those groups should be recombined to meet the concurrency restriction. If there's no way to fit all the transactions into the maximum number of concurrency groups without exceeding the per group gas limit when validating a block, it will be considered invalid.

### Transaction Selection & Packing

Transaction selection should seek to fully saturate each concurrency group in a way that maximizes the overall fee collected from a block. A fully optimal solution may involve a more complicated variant of the knapsack problem which has to factor in dependencies between transactions. To avoid this overhead, the selection process could greedily select transactions based on gas price. Greedy selection would attempt to pack all concurrency groups by taking transactions in order of their gas price. It would then need to skip over transactions that contain resource overlap with full concurrency groups.

#### Deterministic Packing

[...]

PR #1694, opened 2024-02-21
Moved `StorageTransaction` to the `fuel-core-storage` crate

xgreenx:


crates/fuel-core/src/database/storage.rs
```diff
@@ -51,7 +55,7 @@ where
     }
 }
 
-// TODO: After https://github.com/FuelLabs/fuel-vm/pull/679 implement it only for `feature = "test-helpers"`
+#[cfg(feature = "test-helpers")]
```
MitchTurner, line 58: Why is this just a test impl? `StorageMutate` and `Database` seem like implementation concepts, not just test concepts. Is it more to do with the constraints on the blanket impl?

MitchTurner: > The StorageMutate is only implemented for the StorageTransaction, and the Storage should only implement read methods(and only one mutate commit_changes method).
I just read this in the issue. Probably the answer to my question.

benches/benches/vm.rs
```diff
@@ -37,15 +39,17 @@ where
 
             let clock = quanta::Clock::new();
 
+            let original_db = vm.as_mut().database_mut().clone();
+            // Simulates the block production/validation with three levels of database transaction.
+            let block_database_tx = original_db.clone().into_transaction();
+            let relayer_database_tx = block_database_tx.into_transaction();
+            let thread_database_tx = relayer_database_tx.into_transaction();
+            let tx_database_tx = thread_database_tx.into_transaction();
```
MitchTurner, line 47: I'm not sure what's going on here. You seem to be wrapping each in another `StorageTransaction`, but I don't know how that relates to `block`, `relayer`, `thread`, etc.

crates/fuel-core/src/database.rs
```diff
@@ -83,164 +82,89 @@ pub mod message;
...
-    }
-}
-
-impl Drop for DropResources {
-    fn drop(&mut self) {
-        if let Some(drop) = self.drop.take() {
-            (drop)()
-        }
-    }
+    height: Arc<Mutex<Option<Description::Height>>>,
```
MitchTurner, line 93: I feel like since `height` is part of this, we could change the name of `Database` to something more specific and domain related. Will this always be the block height?

crates/fuel-core/src/database.rs
```diff
@@ -83,164 +82,89 @@ pub mod message;
...
+    pub fn new(data_source: DataSource<Description>) -> Self {
+        let mut database = Self {
+            height: Arc::new(Mutex::new(None)),
+            data: data_source,
+        };
+        let height = database
+            .latest_height()
+            .expect("Failed to get latest height during creation of the database");
+
+        database.height = Arc::new(Mutex::new(height));
```
MitchTurner, line 111: Why do we need to set the `height` if there is a `latest_height()` method that works without that field?

crates/fuel-core/src/database.rs
```diff
@@ -83,164 +82,89 @@ pub mod message;
...
 
-    pub fn with_drop(mut self, drop: DropFn) -> Self {
-        self._drop = Arc::new(drop.into());
-        self
+        database
     }
 
     #[cfg(feature = "rocksdb")]
-    pub fn open(path: &Path, capacity: impl Into<Option<usize>>) -> DatabaseResult<Self> {
+    pub fn open(path: &Path, capacity: impl Into<Option<usize>>) -> Result<Self> {
```
MitchTurner, line 117: This should be called something like `new_rocksdb` or something like that imo.

crates/fuel-core/src/database.rs
MitchTurner: I see the TODO to add this below. Is this just to reduce lookups?

crates/fuel-core/src/database.rs
```diff
@@ -421,11 +239,8 @@ impl AtomicView for Database<OnChain> {
 
     type Height = BlockHeight;
 
-    fn latest_height(&self) -> BlockHeight {
-        // TODO: The database should track the latest height inside of the database object
-        //  instead of fetching it from the `FuelBlocks` table. As a temporary solution,
-        //  fetch it from the table for now.
-        self.latest_height().unwrap_or_default()
+    fn latest_height(&self) -> Option<Self::Height> {
+        *self.height.lock().expect("poisoned")
```
MitchTurner, line 243: For these `expect`s, Instead of telling why it panicked, we should explain why this shouldn't panic. I'm skeptical.

crates/fuel-core/src/database.rs
```diff
@@ -520,18 +493,486 @@ mod tests {
...
+            assert_eq!(database.latest_height().unwrap(), None);
+
+            // When
+            database
+                .storage_as_mut::<Coins>()
+                .insert(&Default::default(), &Default::default())
+                .unwrap();
+
+            // Then
+            assert_eq!(database.latest_height().unwrap(), None);
```
MitchTurner, line 543: Maybe make another test for if the `latest_height` isn't `None` to start.

crates/fuel-core/src/database.rs
```diff
@@ -520,18 +493,486 @@ mod tests {
...
+        #[test]
+        fn database_not_advances_without_block() {
+            // Given
+            let mut database = Database::<OnChain>::default();
+            assert_eq!(database.latest_height().unwrap(), None);
+
+            // When
+            database
+                .storage_as_mut::<Coins>()
+                .insert(&Default::default(), &Default::default())
```
MitchTurner, line 539: It's not quite clear what the `without_block` is here. `Default` obscures what is actually being inserted.

crates/fuel-core/src/database.rs
```diff
@@ -520,18 +493,486 @@ mod tests {
...
+
+            // Then
+            assert_eq!(database.latest_height().unwrap(), Some(advanced_height));
+        }
+
+        #[test]
+        fn database_not_advances_without_block() {
+            // Given
+            let mut database = Database::<OffChain>::default();
+            assert_eq!(database.latest_height().unwrap(), None);
```
MitchTurner, line 694: Same here: maybe do another where it's not `None` to start.

crates/fuel-core/src/database.rs
```diff
@@ -520,18 +493,486 @@ mod tests {
...
+
+            // Then
+            assert_eq!(database.latest_height().unwrap(), Some(advanced_height));
+        }
+
+        #[test]
+        fn database_not_advances_without_block() {
+            // Given
+            let mut database = Database::<Relayer>::default();
+            assert_eq!(database.latest_height().unwrap(), None);
```
MitchTurner, line 856: Here too.

crates/fuel-core/src/database.rs
xgreenx: The `Description` makes it kind of more domain-specific like `Database<OnChain>` or `Database<OffChain>`. I'm okay with improving naming if it is not good enough, do you have any suggestions?=)

crates/fuel-core/src/database.rs
xgreenx: It reduces lookup into the database, plus avoids any concurrent modification of the database because we verify the block height during `commit_changes`, and it should advance each commit.

benches/benches/vm.rs
xgreenx: Added a comment with more detailed description=)

crates/fuel-core/src/state/in_memory.rs
```diff
@@ -1,2 +1,2 @@
+// TODO: Move the implementation from the module to here.
```
MitchTurner, line 1: Is this part of a future PR?

xgreenx: I don't want to move it in this PR not to mix the changes and provide a clear diff of what actually was changed in the `in_memory` module. In can be done by someone who will notice this todo=)

crates/fuel-core/src/database.rs
xgreenx: We have one here, or do you mean something else?

image

crates/services/executor/src/executor.rs
```diff
@@ -548,9 +554,22 @@ where
...
         }
 
+        // The block level storage transaction that also contains data from the relayer.
+        // Starting this point modifications from each thread should be independent
+        // and shouldn't touch the same data.
+        let mut block_with_relayer_data_transaction = block_st_transaction.write_transaction()
+                // Enforces independent changes from each thread.
+                .with_policy(ConflictPolicy::Fail);
+
+        // We execute transactions in a single thread right now, but later,
```
MitchTurner, line 567: Is there an issue for tracking this?

xgreenx: https://github.com/FuelLabs/fuel-core/issues/1088 and https://github.com/FuelLabs/fuel-core/issues/1189

crates/storage/src/blueprint/merklized.rs
```diff
@@ -321,7 +338,7 @@ macro_rules! basic_merklelized_storage_tests {
             #[test]
             fn get() {
                 let mut storage = InMemoryStorage::default();
-                let mut structured_storage = StructuredStorage::new(&mut storage);
```
MitchTurner, line 324: These all still have the name `structured_storage` in all these tests. Is that still accurate?

crates/storage/src/blueprint/plain.rs
```diff
@@ -181,7 +198,7 @@ macro_rules! basic_storage_tests {
             #[test]
             fn get() {
                 let mut storage = InMemoryStorage::default();
-                let mut structured_storage = StructuredStorage::new(&mut storage);
+                let mut structured_storage = storage.write_transaction();
```
MitchTurner, line 201: Same question here about `structured_storage`.

crates/storage/src/blueprint/sparse.rs
```diff
@@ -494,7 +499,7 @@ macro_rules! root_storage_tests {
             #[test]
             fn root() {
                 let mut storage = InMemoryStorage::default();
-                let mut structured_storage = StructuredStorage::new(&mut storage);
+                let mut structured_storage = storage.write_transaction();
```
MitchTurner, line 502: Same here.

crates/storage/src/transactional.rs
```diff
@@ -1,96 +1,478 @@
...
     }
 }
 
-/// Provides a view of the storage at the given height.
-/// It guarantees to be atomic, meaning the view is immutable to outside modifications.
-pub trait AtomicView: Send + Sync {
-    /// The type of the storage view.
-    type View;
+#[cfg(feature = "test-helpers")]
+mod test {
```
MitchTurner, line 385: I think all these tests could us a little clarification on what's being tested and what the expected results are. They're pretty close.

crates/storage/src/blueprint/merklized.rs
MitchTurner: Oh. Hmm:
```
pub type StorageTransaction<S> = StructuredStorage<StorageTransactionInner<S>>;
```
Either way I guess...

crates/fuel-core/src/state/in_memory/transaction.rs
```diff
@@ -1,708 +0,0 @@
...
-        for lock in self.changes.iter() {
-            lock.lock().expect("poisoned lock").clear();
-        }
-        self.view_layer.flush()?;
-        self.data_source.flush()
-    }
-}
-
-#[cfg(test)]
-mod tests {
```
MitchTurner, line 292: Do these tests all have parity with new tests? There aren't nearly as many `StorageTransactionInner` tests, but maybe there doesn't need to be.

MitchTurner (commented):
Woof. Finished my first pass. This is a beast lol.

crates/fuel-core/src/state/in_memory/transaction.rs
xgreenx: You are right, moved them into `crates/storage/src/transactional.rs`

CHANGELOG.md
```diff
@@ -38,6 +40,30 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+
+    - One of the main ideas of the change is to allow for the actual storage only to implement `KeyValueInspect` and `Modifiable` without the `KeyValueMutate`. It simplifies work with the databases and provides a safe way of interacting with them (Modification into the database can only go through the `Modifiable::commit_changes`). This feature is used to [track the height](https://github.com/FuelLabs/fuel-core/pull/1694/files#diff-c95a3d57a39feac7c8c2f3b193a24eec39e794413adc741df36450f9a4539898) of each database during commits and even limit how commits are done, providing additional safety. This part of the change was done as a [separate commit](https://github.com/FuelLabs/fuel-core/pull/1694/commits/7b1141ac838568e3590f09dd420cb24a6946bd32).
+    
+    - The `StorageTransaction` is a `StructuredStorage` that uses `StorageTransactionInner` inside to accumulate modifications. Only `StorageStorageInner` has a real implementation of the `KeyValueMutate`(Other types only implement it in tests).
```
Voxelot, line 51: what is Storage Storage Inner?

CHANGELOG.md
```diff
@@ -38,6 +40,30 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+    
+    - As was planned in the issue, now the executor has full control over how commits to the storage are done.
+    
+    - All mutation methods now require `&mut self` - exclusive ownership over the object to be able to write into it. It almost negates the chance of concurrent modification of the storage, but it is still possible since the `Database` implements the `Clone` trait. To be sure that we don't corrupt the state of the database, the `commit_changes` function implements additional safety checks to be sure that we commit updates per each height only once time.
+
+    - Side changes:
+      - The `drop` function was moved from `Database` to `RocksDB` as a preparation for the state rewind since the read view should also keep the drop function until it is destroyed.
+      - The `StatisticTable` table lives in the off-chain worker.
+      - Removed duplication of the `Database` from the `dapp::ConcreteStorage` since it is already available from the VM.
```
Voxelot, line 62:
```suggestion
      - Removed duplication of the `Database` from the `dap::ConcreteStorage` since it is already available from the VM.
```

crates/fuel-core/src/database.rs
```diff
@@ -491,6 +289,176 @@ impl AtomicView for Database<Relayer> {
...
+    for<'a> StorageTransaction<&'a &'a mut Database<Description>>:
+        StorageMutate<MetadataTable<Description>, Error = StorageError>,
+{
+    // Gets the all new heights from the `changes`
+    let iterator = ChangesIterator::<Description>::new(&changes);
+    let new_heights = heights_lookup(&iterator)?;
+
+    // Changes for each block should be committed separately.
+    // If we have more than one height, it means we are mixing commits
+    // for several heights in one butch - return error in this case.
```
Dentosal, line 378:
```suggestion
    // for several heights in one batch - return error in this case.
```

crates/types/src/services.rs
```diff
@@ -10,46 +10,39 @@ pub mod txpool;
 
 // TODO: Define a one common error for all services like
 
-/// The uncommitted `Result` of some action with database transaction.
+/// The uncommitted `Result` of some action with storage changes.
 /// The user should commit the result by itself.
 #[derive(Debug)]
-pub struct Uncommitted<Result, DatabaseTransaction> {
+pub struct Uncommitted<Result, Changes> {
```
Dentosal, line 16: This probaby wants a `#[must_use]`

crates/storage/src/transactional.rs
```diff
@@ -1,96 +1,811 @@
...
+
+    /// Returns the view of the storage for the latest block height.
+    fn latest_view(&self) -> Self::View;
+}
+
+/// Storage transaction on top of the storage.
+pub type StorageTransaction<S> = StructuredStorage<StorageTransactionInner<S>>;
+
+/// The inner representation of the storage transaction.
+#[doc(hidden)]
```
Dentosal, line 57: `#[doc(hidden)]` should be explained in a comment here. Is this type within semver guarantees? Why it shouldn't be used outside this crate? Why is it available?

crates/fuel-core/src/database.rs
```diff
@@ -491,6 +289,176 @@ impl AtomicView for Database<Relayer> {
...
+        (Some(prev_height), Some(new_height)) => {
+            // Each new commit should be linked to the previous commit to create a monotonically growing database.
+
+            let next_expected_height = prev_height
+                .advance_height()
+                .ok_or(DatabaseError::FailedToAdvanceHeight)?;
+
+            // TODO: After https://github.com/FuelLabs/fuel-core/issues/451
+            //  we can replace `next_expected_height > new_height` with `next_expected_height != new_height`.
+            if next_expected_height > new_height {
```
bvrooman, line 404: Is there a specific reason that using `!=` depends on issue 451? 
It looks like this is saying that `next_expected_height` is allowed to be equal to or less than `new_height`.

crates/fuel-core/src/database.rs
```diff
@@ -491,6 +289,176 @@ impl AtomicView for Database<Relayer> {
...
+                }
+                .into());
+            }
+        }
+        (None, Some(_)) => {
+            // The new height is finally found; starting at this point,
+            // all next commits should be linked(the height should increase each time by one).
+        }
+        (Some(prev_height), None) => {
+            // In production, we shouldn't have cases where we call `commit_chagnes` with intermediate changes.
```
bvrooman, line 417:
```suggestion
            // In production, we shouldn't have cases where we call `commit_changes` with intermediate changes.
```

bvrooman (commented):
Review in progress (10/83 files so far).
It looks like this change means that we only make one database commit per block height, and we accumulate all changes in a transaction. If the commit contains multiple changes to block height, we raise an error; if the commit contains no changes to block height, i.e. does not advance block height, it is also an error. Is that correct?

crates/fuel-core/src/database.rs
xgreenx: Right now, the relayer commits only when there are messages. If the DA block has no messages, the relayer doesn't commit.

bvrooman (commented):
47 / 86 files reviewed so far, will continue tomorrow

crates/fuel-core/src/database/balances.rs
```diff
@@ -51,8 +70,30 @@ impl Database {
...
+                                &balance_entry.contract_id,
+                                &balance_entry.asset_id,
+                            );
+                            (db_key, balance_entry.amount)
+                        })
+                        .collect_vec();
+
+                    #[allow(clippy::map_identity)]
+                    let balance_entries_iter =
+                        balance_entries.iter().map(|(key, value)| (key, value));
```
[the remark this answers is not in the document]
xgreenx: We want to convert `&(k, v)` into `(&k, &v)`=)

crates/services/relayer/src/storage.rs
```diff
@@ -9,16 +13,19 @@ use fuel_core_storage::{
...
     Error as StorageError,
     Mappable,
     Result as StorageResult,
     StorageAsMut,
-    StorageAsRef,
     StorageMutate,
 };
 use fuel_core_types::{
     blockchain::primitives::DaBlockHeight,
+    fuel_merkle::storage::StorageInspect,
```
bvrooman, line 28: This looks like a bad auto-import - you probably want `StorageInspect` from `fuel_core_storage`

crates/storage/src/structured_storage.rs
```diff
@@ -264,80 +371,50 @@ where
...
         Result as StorageResult,
     };
-    use std::{
-        cell::RefCell,
-        collections::HashMap,
-    };
+    use std::collections::BTreeMap;
 
-    type Storage = RefCell<HashMap<(u32, Vec<u8>), Vec<u8>>>;
+    type Storage = BTreeMap<(u32, Vec<u8>), Value>;
```
bvrooman, line 384: Just curious - why did you change this from a `HashMap` to a `BTreeMap`? Is order important? Or perhaps just performance?

bvrooman (commented):
Reviewed all 86 files at the time of this comment. A couple of small nits and questions. If you are still interested in doing a group review, that may be helpful for a better understanding of the PR from a high level.

crates/fuel-core/src/database/balances.rs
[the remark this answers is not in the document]
bvrooman: I guess this is a false positive from Clippy!

crates/services/importer/src/importer.rs
```diff
@@ -206,14 +207,13 @@ where
...
-    ) -> Result<(), Error>
-    where
-        ExecutorDatabase: ports::ExecutorDatabase,
-    {
-        let (result, mut db_tx) = result.into();
+    fn _commit_result(&self, result: UncommittedResult<Changes>) -> Result<(), Error> {
+        let mut guard = self
+            .database
+            .try_lock()
+            .expect("Semaphore prevents from multiple access to teh database");
```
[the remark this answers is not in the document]
xgreenx: I'm using semaphore because it is the name of the type=D. It helps to connect it with the real object that guarantees safety.
image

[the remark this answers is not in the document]
bvrooman: Yes, but (if I am reading this right) this line accesses `Importer::database` which is a `Mutex<D>` on the database, not on the `guard` (`Semaphore`) on the Importer itself.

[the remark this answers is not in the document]
xgreenx: And the statement references to the `guard`. In all places where we call `_commit_result` we call take a lock for the semaphore first=)

Dentosal (dismissed):
Ok I think this is finally good to go. Phew.

crates/services/importer/src/importer.rs
```diff
@@ -455,7 +460,13 @@ where
         }
 
         let start = Instant::now();
-        let commit_result = self._commit_result(result);
+
+        let mut guard = self
+            .database
+            .try_lock()
+            .expect("Semaphore prevents from multiple access to teh database");
```
bvrooman, line 467:
```suggestion
            .expect("Semaphore prevents concurrent access to the database");
```

crates/services/importer/src/importer.rs
```diff
@@ -192,8 +193,13 @@ where
         if let Some(channel) = previous_block_result {
             let _ = channel.await;
         }
+        let mut guard = self
+            .database
+            .try_lock()
+            .expect("Semaphore prevents from multiple access to teh database");
```
bvrooman, line 199:
```suggestion
            .expect("Semaphore prevents concurrent access to the database");
```

crates/fuel-core/src/database.rs
MitchTurner: `Database<OnChain>` still doesn't help the reader understand what the database is used for, just that it for information that is on a (block)chain.

The fact that it is constrained to things that have a `height` implies that it isn't a general-purpose database.

Based on the traits implemented for the type, maybe `ChainStateDatabase`, `IterableChainStateDatabase`, or `AtomicIterableChainStateDatabase`? Sorry, I'm realizing I'm making a jab at inheritance design patterns :P
`ChainStateDatabase` probably is the most appropriate.

xgreenx: I'm okay with renaming it to `ChainStateDatabase` or `HeightBasedDatabase` or `BlockchainDatabase`=)  But I will do that in a separate PR to not change a lot of files

Theghost256 (approved):

Theghost256 (approved):

PR #1713, opened 2024-02-28
feat: Impl StorageRead and StorageWrite

bvrooman:
Related issues:
- https://github.com/FuelLabs/fuel-vm/issues/655

This PR removes the specific implementation of `StorageRead` for `ContractsRawCode`, and replaces it with a generic implementation over `Blueprint` tables and `StructuredStorage`. This means that other implementations of `TableWithBlueprint` will automatically receive this implementation, including `ContractsStorage`, which uses the `Sparse` Blueprint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


xgreenx (changes requested):
Could you add tests for the `ContractsState` table, please, that using `write`, `replace` and `insert` produces the same output? And could you do the same for `read` and `get`, please=)

xgreenx (approved):

Issue #1641, opened 2024-02-01
Add a way to fetch base gas price from the blockchain


xgreenx:

MitchTurner:
I think this might be a dupe of these issues:
https://github.com/FuelLabs/fuel-core/issues/1647
https://github.com/FuelLabs/fuel-core/issues/1648

PR #1714, opened 2024-02-29
Bump `fuel-vm` dep to 0.47.1

MitchTurner:
The change bumps the `fuel-vm` to `0.47.1`. It breaks several breaking changes into the protocol:
- All malleable fields are zero during the execution and unavailable through the GTF getters. Accessing them via the memory directly is still possible, but they are zero.
- The `Transaction` doesn't define the gas price anymore. The gas price is defined by the block producer and recorded in the `Mint` transaction at the end of the block. A price of future blocks can be fetched through a [new API nedopoint](https://github.com/FuelLabs/fuel-core/issues/1641) and the price of the last block can be fetch or via the block or another [API endpoint](https://github.com/FuelLabs/fuel-core/issues/1647).
- The `GasPrice` policy is replaced with the `Tip` policy. The user may specify in the native tokens how much he wants to pay the block producer to include his transaction in the block. It is the prioritazation mechanism to incentivize the block producer to include users transactions earlier.
- The `MaxFee` policy is mandatory to set. Without it, the transaction pool will reject the transaction. Since the block producer defines the gas price, the only way to control how much user agreed to pay can be done only through this policy. 
- The `maturity` field is removed from the `Input::Coin`. The same affect can be achieve with the `Maturity` policy on the transaction and predicate. This changes breaks how input coin is created and removes the passing of this argument.
- The metadata of the `Checked<Tx>` doesn't contain `max_fee` and `min_fee` anymore. Only `max_gas` and `min_gas`. The `max_fee` is controlled by the user via the `MaxFee` policy.
- Added automatic `impl` of traits `StorageWrite` and `StorageRead` for `StructuredStorage`. Tables that use a `Blueprint` can be read and written using these interfaces provided by structured storage types.


xgreenx (approved):

Dentosal (approved):
Lgtm. I'm starting to wish we had a monorepo for this.

Issue #1547, opened 2023-12-14
Forkless state transition


xgreenx:
Implement the WASM version of the state transition function that can be represented by the byte code and shared with the rest of the network.

The WASM state transition function should generate the same changes as the std version.

PR #1716, opened 2024-03-01
Forkless state transition with upgradable WASM executor

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1547

## Change overview

The change adds the WASM version of the `fuel_core_executor::Executor` and the upgradable executor that works with native(std) and WASM(no-std) versions of the executor. Currently,  it uses either the WASM or the std version. But in the follow-up PRs it will decide which version to use based on the block header. For now, this behavior is controlled by the "wasm-executor" feature. The CI runs tests two times, one with the WASM executor and the second with std, to verify that behaviors are the same. 

## Details on how WASM integration is done

The `fuel-core-upgradable-executor` uses the `wasmtime` library to compile and run the WASM bytecode. The `wastime::Engine` allows customization of how the `wastime::Module` compiles the bytecode. For simplicity, we are using the default configuration that works on all platforms for now, but later, it is possible to do more specific compilation.

We are using `wastime::Linker` to add host functions that can be called from the WASM runtime via `extern "C"` API:

```rust
#[link(wasm_import_module = "host_v0")]
extern "C" {
    /// Returns the size of the value from the storage.
    pub(crate) fn storage_size_of_value(
        key_ptr: Ptr32<[u8]>,
        key_len: u32,
        column: u32,
    ) -> u64;

    /// Returns the value from the storage.
    pub(crate) fn storage_get(
        key_ptr: Ptr32<[u8]>,
        key_len: u32,
        column: u32,
        out_ptr: Ptr32Mut<[u8]>,
        out_len: u32,
    ) -> ReturnResult;
}
```

The host functions are closures that capture data required for the work, such as the storage or relayer. Host functions have access to the `ExecutionState`, which accumulates intermediate data.

Data is passed between the host and runtime through memory, where one side gives the pointer where data should be stored, and another side writes the memory using this pointer and the size. The host is not responsible for allocating and deallocating memory; the memory management is done within WASM runtime. Currently, when we return something from the runtime to the host, we "leak" it to avoid deallocation. It allows the host to read the value from the memory.

Side changes:
- The coinbase contract id is not a part of the `Config` anymore. It is part of the `Components` and seats along with the `gas_price` because it is defined by the block producer and can always be changed on the fly.
- The usage of `anyhow::Error` in the `ExecutorError` was replaced by `String` because it doesn't implement `serde` ser/des.
- The `ExecutionOptions`, `WriteOperation`, `ExecutionResult`, `TransactionExecutionResult`, `TransactionExecutionStatus`, `ExecutorError`, and `TransactionValidityError` support `serde` ser/des.
- The `fuel-core-executor` only provides the `ExecutionInstance` without `Executor`. The logic of the `Executor` was moved to the `fuel-core-upgradable-executor`.


MitchTurner:
I'm just starting looking at this PR, so I don't know if it's completely relevant, but have you looked into the component model and using WIT Contracts to define your host functions for the WASM code?

https://component-model.bytecodealliance.org/design/wit.html

I've played around with it a bit in the past and it might be a little cleaner than using `extern "C"` and leaking.

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -178,9 +181,7 @@ where
         let tx_idx = u16::try_from(tx_idx).map_err(|e| {
             anyhow::anyhow!("The block has more than `u16::MAX` transactions, {}", e)
         })?;
-        let tx_id = tx.cached_id().expect(
-            "The imported block should contains only transactions with cached id",
-        );
+        let tx_id = tx.id(chain_id);
```
MitchTurner, line 184: Why don't we use the cached value anymore?

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -0,0 +1,281 @@
...
+    /// but with the ability to specify the coinbase recipient and the gas price.
+    pub fn execute_without_commit_with_coinbase(
+        &self,
+        block: fuel_core_types::services::executor::ExecutionBlock,
+        coinbase_recipient: fuel_core_types::fuel_types::ContractId,
+        gas_price: u64,
+    ) -> fuel_core_types::services::executor::Result<UncommittedResult<Changes>> {
+        let component = match block {
+            ExecutionTypes::DryRun(_) => {
+                panic!("It is not possible to commit the dry run result");
```
MitchTurner, line 121: I know this is just a copy and paste, but should we return an error instead of a `panic` here?

crates/fuel-core/src/graphql_api/worker_service.rs
xgreenx: With a WASM executor, we deserialize the result of the execution, and because of that, all metadata is empty.

xgreenx:
> I'm just starting looking at this PR, so I don't know if it's completely relevant, but have you looked into the component model and using WIT Contracts to define your host functions for the WASM code?

The problem with WIP is that we need to generate a rust interface based on the WIP file. But it means some additional dependency that I don't want to bring=)

crates/services/upgradable-executor/src/executor.rs
xgreenx: This function is marked `#[cfg(any(test, feature = "test-helpers"))]` and used only in tests, so it is okay to panic=)

crates/services/upgradable-executor/src/instance.rs
```diff
@@ -0,0 +1,592 @@
...
+}
+
+/// This stage adds host functions to get transactions from the transaction source.
+pub struct Source {
+    linker: Linker<ExecutionState>,
+}
+
+impl Instance<Created> {
+    /// Adds host functions for the `source`.
+    pub fn add_source<TxSource>(
```
MitchTurner, line 114: These 3 `add` methods are all really long and have a fair bit of `wasmtime` DSL. Could you break them down to be a series of declarative helper functions? e.g. instead of having the entire `Func::wrap` closure, just have a method like:
```
let peek_next_txs_size_func = self.get_peek_next_txs_size_func(...)
```
`add_source` could be like 7 lines + the return and it would be easier to read.

MitchTurner: Well, it might make more sense to have the function `get` and the linker call be in single action; the functions doesn't mean much if you don't see how it's being passed to the linker. So might be fewer than 7 lines. LMK if that doesn't make sense.

crates/services/upgradable-executor/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+    Module,
+};
+
+pub mod config;
+pub mod executor;
+
+#[cfg(feature = "wasm-executor")]
+pub mod instance;
+
+/// The WASM bytecode of the native executor.
```
MitchTurner, line 19: This comment is confusing. "of the native executor"? Do you just mean the "executor"? It's not native anymore if it's Wasm.

MitchTurner (commented):
I've finished my first pass. This is all pretty straightforward and besides the refactoring suggestions I made, LGTM.

MitchTurner (approved):
Those methods are much cleaner. Thanks. The `add_method` function really helps the reader see what's going on and how each function differs.

MitchTurner:
You requested a review after I'd already approved. Maybe you didn't notice.

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -0,0 +1,281 @@
...
+            Result as ExecutorResult,
+            TransactionExecutionStatus,
+        },
+        Uncommitted,
+    },
+};
+use std::sync::Arc;
+
+/// The upgradable executor supports the WASM version of the state transition function.
+/// If the block allows it, the executor uses a native state transition function.
```
Dentosal, line 43: What does this mean exactly "if the block allows it"?

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -0,0 +1,281 @@
...
+        let instance = crate::instance::Instance::new(&self.engine)
+            .add_source(source)?
+            .add_storage(storage)?
+            .add_relayer(relayer)?
+            .add_input_data(block, options)?;
+
+        instance.run(&self.module)
+    }
+
+    #[allow(dead_code)]
```
Dentosal, line 257: Why is this `dead_code` instead of a `cfg(feature = ...)`?

crates/services/upgradable-executor/src/instance.rs
```diff
@@ -0,0 +1,574 @@
...
+    Linker,
+    Memory,
+    Module,
+    Store,
+};
+
+trait CallerHelper {
+    /// Writes the encoded data to the memory at the provided pointer.
+    fn write(&mut self, ptr: u32, encoded: &[u8]) -> anyhow::Result<()>;
+}
```
Dentosal, line 51: Why this isn't just part of `Caller`?

crates/services/upgradable-executor/src/instance.rs
```diff
@@ -0,0 +1,574 @@
...
+                .entry(encoded_size)
+                .or_default()
+                .push(encoded_txs);
+            Ok(encoded_size)
+        };
+
+        Func::wrap(&mut self.store, closure)
+    }
+
+    fn consume_next_txs(&mut self) -> Func {
```
Dentosal, line 197: nit: this consumes only a single tx
```suggestion
    fn consume_next_tx(&mut self) -> Func {
```

crates/services/upgradable-executor/src/instance.rs
xgreenx: The `Caller` is the type defined inside of the `wasmtime`.

crates/services/upgradable-executor/build.rs
```diff
@@ -0,0 +1,69 @@
...
+    },
+    process::Command,
+};
+
+fn main() {
+    let wasm_executor_enabled = env::var_os("CARGO_FEATURE_WASM_EXECUTOR").is_some();
+
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-changed=wasm-executor/src/lib.rs");
+    println!("cargo:rerun-if-changed=wasm-executor/src/main.rs");
```
Voxelot, line 15: Would we need to rerun if _any_ files in the `wasm-executor/src/*` dir change?

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -0,0 +1,280 @@
...
+    engine: wasmtime::Engine,
+    #[cfg(feature = "wasm-executor")]
+    module: wasmtime::Module,
+}
+
+impl<S, R> Executor<S, R> {
+    pub fn new(
+        storage_view_provider: S,
+        relayer_view_provider: R,
+        config: Arc<Config>,
```
bvrooman, line 59: It looks like all instances of `new` create the `Arc` inline. I might suggest taking `config` as `config: Config` and  doing:
```
Self {
   ...
   config: Arc::new(config)
}
```

to simplify things a little.

crates/services/upgradable-executor/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+description = "Fuel Block Upgradable Executor"
+build = "build.rs"
+
+[dependencies]
+anyhow = { workspace = true, optional = true }
+fuel-core-executor = { workspace = true }
+fuel-core-storage = { workspace = true }
+fuel-core-types = { workspace = true }
+fuel-core-wasm-executor = { workspace = true, optional = true }
+lazy_static = { version = "1.4", optional = true }
```
bvrooman, line 19:
```suggestion
lazy_static = { workspace = true, optional = true }
```

crates/services/upgradable-executor/src/config.rs
```diff
@@ -0,0 +1,23 @@
+use fuel_core_executor::executor::ExecutionOptions;
+use fuel_core_types::fuel_tx::ConsensusParameters;
+
+#[derive(Clone, Debug, Default)]
+pub struct Config {
+    /// Network-wide common parameters used for validating the chain.
+    /// The executor already has these parameters, and this field allows override the used value.
```
bvrooman, line 7:
```suggestion
    /// The executor already has these parameters, and this field allows us to override the existing value.
```

bvrooman (commented):
Reviewed 26 files so far, will continue

crates/services/upgradable-executor/wasm-executor/src/ext.rs
```diff
@@ -0,0 +1,281 @@
...
+        ));
+    }
+
+    if !exists || size == 0 {
+        return Ok(vec![]);
+    }
+
+    let mut encoded_events = vec![0u8; size as usize];
+
+    unsafe {
```
Voxelot, line 270: Generally it's good practice to add some comments to unsafe code to explain why it's marked unsafe and any precautions taken to ensure no memory corruption occcurs or similar justifications

xgreenx: The function is unsafe because we call a method from `extern`. I'm not sure what comment should say in that case=) "We assume that the host implements corresponding function overwise it should fail before we even started execution of the WASM."

crates/services/upgradable-executor/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+
+/// The WASM version of the underlying [`fuel_core_executor::executor::ExecutionInstance`].
+#[cfg(feature = "wasm-executor")]
+pub const WASM_BYTECODE: &[u8] = include_bytes!(concat!(
+    env!("OUT_DIR"),
+    "/wasm32-unknown-unknown/release/fuel-core-wasm-executor.wasm"
+));
+
+#[cfg(feature = "wasm-executor")]
+lazy_static::lazy_static! {
```
Voxelot, line 27: Might not need lazy static anymore.

https://docs.rs/once_cell/latest/once_cell/#safe-initialization-of-global-data

Voxelot (approved):

crates/services/upgradable-executor/src/lib.rs
xgreenx: Fixed in https://github.com/FuelLabs/fuel-core/pull/1781

PR #1717, opened 2024-03-01
Bugfix: Calling of the contract with enabled `utxo_validation` fails

bvrooman:
The fix for the [#1657](https://github.com/FuelLabs/fuel-core/pull/1657) to include the contract into `ContractsInfo` table.

Thanks @Torres-ssf for pointing and debugging it.


crates/services/executor/src/executor.rs
```diff
@@ -1073,6 +1074,20 @@ where
             })
         }
 
+        // TODO: Move to an off-chain worker: https://github.com/FuelLabs/fuel-core/issues/1721
```
xgreenx, line 1077: Since it is a temporary solution, I'm okay with doing it here. Later, we will move this code to the off-chain worker.

xgreenx (dismissed):

xgreenx (approved):

PR #1722, opened 2024-03-04
Bugfix: Zero `predicate_gas_used` field during validation of the produced block

xgreenx:


crates/services/executor/src/executor.rs
```diff
@@ -1025,6 +1025,40 @@ where
...
+                    })
+                    | Input::MessageDataPredicate(MessageDataPredicate {
+                        predicate_gas_used,
+                        ..
+                    }) => {
+                        *predicate_gas_used = gas_used;
+                    }
+                    _ => {
+                        debug_assert!(false, "This error is not possible unless VM changes the order of inputs, \
+                        or we added a new predicate inputs.");
```
bvrooman, line 1053: nit:
```suggestion
                        debug_assert!(false, "This error is not possible unless VM changed the order of inputs, \
                        or we added a new predicate input.");
```

bvrooman (approved):

Dentosal (approved):

Issue #1659, opened 2024-02-12
Move subscription notification logic from the `TxPool` to the `GraphQL` service


xgreenx:
In the current codebase, the `TxPool` service handles subscriptions for the transaction statuses. It is expected since the `TxPool` knows when a transaction is removed from it. But it leads to race conditions with GraphQL of chain workers. Sometimes, the user may receive a notification about the transaction when the GraphQL off-chain database is not updated.

During this task, we need to either move the subscription logic from the `TxPool` to the `GraphQL` service or await the processing of the transaction by the GraphQL off-chain worker.

PR #1723, opened 2024-03-04
Notify about imported blocks from the off-chain worker

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1659


Dentosal (approved):
No complaints.

Issue #1661, opened 2024-02-12
Add versioning suffix to the GraphQL endpoint


xgreenx:
Like `0.0.0.0:4000/v1/graphql`

PR #1725, opened 2024-03-04
Added versioning suffix to the GraphQL endpoint

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1661

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 #### Breaking
 
+- [#1725](https://github.com/FuelLabs/fuel-core/pull/1725): All API endpoints now are prefixed with `/v1` version. New usage looks like: `/v1/playground`, `/v1/graphql`, `/v1/graphql-sub`, `/v1/metrics`, `/v1/health`.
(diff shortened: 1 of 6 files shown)


xgreenx:
cc @rfuelsh @liviudm @tirkesi For the next testnet you need to update the link for endpoints=)

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 #### Breaking
```
Dentosal, line 43: We could make this non-breaking by implicitly treating anything without `/v*/` prefix as `/v1/`. I like not doing so, but just pointing out that we necessarily don't have to break compatibiltiy here.

xgreenx: It is more of a question to the @Voxelot . I thought the idea is to break it now to require all SDK handle different version instead of blindly using `/graphql`.

Dentosal (approved):
I was expecting something like this to break lots of tests, but the CI is green, and the URLs don't seem hardcoded, so I guess it's fine.

CHANGELOG.md
Voxelot: Im fine with this option as well if we want to limit churn.

Issue #1642, opened 2024-02-01
Prepare the codebase to use base gas price during block production


xgreenx:
It can be done by a new service or can be part of the block producer logic.

PR #1727, opened 2024-03-04
On-Hold: Refactor Execution Logic

MitchTurner:
Introduced as part of:
https://github.com/FuelLabs/fuel-core/issues/1642

Before, we had a single `execute` function for all the cases we want to "execute" a block: production, validation, and dry run. The block would just be wrapped in an `ExecutionTypes` to specify which type of execution we were expecting. This resulted in _very_ convoluted execution logic with many special cases. 

Now that some of the blocks come with a pre-determined `gas_price` (in the form of the `Mint` tx for validated blocks) and others need to be assigned a `gas_price` (production), I'm taking advantage of this change to refactor `execute` into separate functions to isolate behavior and move to . This will clean up the code a lot and unburden the callers by giving each function a single responsibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Breaking
+
+- [#1727](https://github.com/FuelLabs/fuel-core/pull/1727): Remove unused `map_v` and `map_p` functions on `ExecutionTypes`
+
 ## [Version 0.23.0]
 
 ### Added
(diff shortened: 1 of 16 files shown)


MitchTurner:
Gonna put the refactoring on hold and move that work to another issue:
https://github.com/FuelLabs/fuel-core/issues/1751

I'll cherry-pick the changes for managing base gas price to another branch and keep this one as a placeholder.

xgreenx:
Done as part of the https://github.com/FuelLabs/fuel-core/issues/1751

PR #1733, opened 2024-03-05
Release v0.22.3

Dentosal:
Also add changelog that I forgot from https://github.com/FuelLabs/fuel-core/pull/1732

Why did I even make this a separate PR?


xgreenx (approved):
You made this PR just for the history to not mix things and allow reverting it separately=)

PR #1740, opened 2024-03-06
Remove optional fields from configs

### LINKED ISSUE #1668: Remove usage of `Option` from configs related to regenesis

MujkicA:
Closes #1668

Removes optional fields from coin and contract configs. 
Added the `CoinConfigGenerator` to enable easier setup of coin configs for testing. It holds an internal counter that's used to init the `tx_id` and `output_index`.

A caveat is that we have to be careful when creating coins/contracts in unit tests. If we use default values for `tx_id` there could be overlapping utxos and the node setup will fail. Care also needs to be applied if tests setup configs from multiple sources.


Cargo.lock
```diff
@@ -1027,9 +1027,9 @@ dependencies = [
 
 [[package]]
 name = "chrono"
-version = "0.4.34"
+version = "0.4.35"
```
xgreenx, line 1030: We prefer run `cargo update` only once per week and before release=D Could you revert this file, please?

crates/fuel-core/src/service/genesis.rs
```diff
@@ -174,30 +169,13 @@ pub async fn execute_and_commit_genesis_block(
...
-    )
-}
-
 fn init_coin(
     db: &mut Database,
     coin: &CoinConfig,
-    output_index: u64,
     height: BlockHeight,
 ) -> anyhow::Result<()> {
     // TODO: Store merkle sum tree root over coins with unspecified utxo ids.
```
xgreenx, line 177: I think we can remove that TODO=)

deployment/scripts/chainspec/dev/state_config.json
```diff
@@ -124,10 +124,10 @@
       "contract_id": "7777777777777777777777777777777777777777777777777777777777777777",
       "code": "9000000994318e6e453f30e85bf6088f7161d44e57b86a6af0c955d22b353f91b2465f5e6140000a504d00205d4d30001a4860004945048076440001240400005050c0043d51345024040000",
       "salt": "1bfd51cb31b8d0bc7d93d38f97ab771267d8786ab87073e0c2b8f9ddc44b274e",
-      "tx_id": null,
-      "output_index": null,
-      "tx_pointer_block_height": null,
-      "tx_pointer_tx_idx": null
+      "tx_id": "0000000000000000000000000000000000000000000000000000000000000000",
```
xgreenx, line 127: I'm curios why zero tx id is at the end=D

MujkicA: I felt the contract was special 😅

xgreenx (dismissed):

xgreenx (approved):

PR #1743, opened 2024-03-08
Added blacklisting transaction on the `TxPool`

xgreenx:


bin/fuel-core/src/cli/run/tx_pool.rs
```diff
@@ -0,0 +1,104 @@
...
+};
+
+#[derive(Debug, Clone, clap::Args)]
+pub struct TxPoolArgs {
+    /// The max time to live of the transaction inside of the `TxPool`.
+    #[clap(long = "tx-pool-ttl", default_value = "5m", env)]
+    pub tx_pool_ttl: humantime::Duration,
+
+    /// The max number of transactions that the `TxPool` can simultaneously store.
+    #[clap(long = "tx-max-number", default_value = "4064", env)]
```
bvrooman, line 19: Where did the "4064" come from?

bvrooman (approved):
It sounds like every node will need to have the same blacklist in their startup args. Otherwise, some nodes will accept the TX, while some will not.

bin/fuel-core/src/cli/run/tx_pool.rs
Voxelot: looks like it was just copied over from before

Voxelot (approved):
lgtm! pushed a version bump commit as well so we can release this.

PR #1747, opened 2024-03-11
Regenesis should also store da block height

xgreenx:


crates/services/consensus_module/src/block_verifier/config.rs
```diff
@@ -1,24 +1,32 @@
...
 /// The config of the block verifier.
 pub struct Config {
     /// The chain configuration.
     pub chain_config: ChainConfig,
-    // TODO: Decide do we need it here or inside of the `ChainConfig` or somewhere else.
-    //  https://github.com/FuelLabs/fuel-core/issues/1667
-    /// The block height at genesis
+    /// The block height of the genesis block.
     pub block_height: BlockHeight,
+    /// The DAT block height at genesis block.
```
segfault-magnet, line 15:
```suggestion
    /// The DA block height at genesis block.
```

crates/services/txpool/src/txpool.rs
```diff
@@ -252,7 +252,7 @@ where
         self.insert_inner(tx, &view)
     }
 
-    #[tracing::instrument(level = "info", skip_all, fields(tx_id = %tx.id()), ret, err)]
+    #[tracing::instrument(level = "debug", skip_all, fields(tx_id = %tx.id()), ret, err)]
```
segfault-magnet, line 255: Should this remain as debug?

xgreenx: Yep, because it logs each inserted and removed transactions

Dentosal (approved):
This looks like the correct approach.

PR #1752, opened 2024-03-13
 Prepare the codebase to use base gas price during block production #1642 

MitchTurner:
Closes https://github.com/FuelLabs/fuel-core/issues/1642

Introduce new `ProducerGasPrice` interface for getting gas price for a given block.

Create simple `StaticGasPrice` impl for parity with current behavior.

### DIFF
--- a/crates/fuel-core/src/schema/gas_price.rs
+++ b/crates/fuel-core/src/schema/gas_price.rs
@@ -5,13 +5,22 @@ use super::scalars::{
 use crate::{
     fuel_core_graphql_api::database::ReadView,
     graphql_api::api_service::GasPriceProvider,
-    query::BlockQueryData,
+    query::{
+        BlockQueryData,
+        SimpleTransactionData,
+    },
 };
 use async_graphql::{
     Context,
     Object,
 };
-use fuel_core_types::blockchain::block::Block;
+use fuel_core_types::{
+    blockchain::block::Block,
+    fuel_tx::{
+        field::MintGasPrice,
+        Transaction,
+    },
+};
 
 pub struct LatestGasPrice {
     pub gas_price: U64,
@@ -38,20 +47,18 @@ impl LatestGasPriceQuery {
         &self,
         ctx: &Context<'_>,
     ) -> async_graphql::Result<LatestGasPrice> {
-        let gas_price_provider = ctx.data_unchecked::<GasPriceProvider>();
         let query: &ReadView = ctx.data_unchecked();
 
         let latest_block: Block<_> = query.latest_block()?;
-        let block_height = u32::from(*latest_block.header().height());
-        let gas_price = gas_price_provider
-            .known_gas_price(block_height.into())
-            .await
-            .ok_or(async_graphql::Error::new(format!(
-                "Gas price not found for latest block:{block_height:?}"
-            )))?;
-
+        let block_height: u32 = (*latest_block.header().height()).into();
+        let mut gas_price: U64 = 0.into();
+        if let Some(tx_id) = latest_block.transactions().last() {
+            if let Transaction::Mint(mint_tx) = query.transaction(tx_id)? {
+                gas_price = (*mint_tx.gas_price()).into();
+            }
+        }
         Ok(LatestGasPrice {
-            gas_price: gas_price.into(),
+            gas_price,
             block_height: block_height.into(),
         })
     }
@@ -83,8 +90,7 @@ impl EstimateGasPriceQuery {
     ) -> async_graphql::Result<EstimateGasPrice> {
         let query: &ReadView = ctx.data_unchecked();
 
-        let latest_block: Block<_> = query.latest_block()?;
-        let latest_block_height = u32::from(*latest_block.header().height());
+        let latest_block_height: u32 = query.latest_block_height()?.into();
         let target_block = block_horizon
             .and_then(|h| h.0.checked_add(latest_block_height))
             .ok_or(async_graphql::Error::new(format!(
--- a/crates/fuel-core/src/schema/gas_price.rs
+++ b/crates/fuel-core/src/schema/gas_price.rs
@@ -3,10 +3,8 @@ use super::scalars::{
     U64,
 };
 use crate::{
-    fuel_core_graphql_api::{
-        database::ReadView,
-        Config as GraphQLConfig,
-    },
+    fuel_core_graphql_api::database::ReadView,
+    graphql_api::api_service::GasPriceProvider,
     query::BlockQueryData,
 };
 use async_graphql::{
@@ -40,14 +38,20 @@ impl LatestGasPriceQuery {
         &self,
         ctx: &Context<'_>,
     ) -> async_graphql::Result<LatestGasPrice> {
-        let config = ctx.data_unchecked::<GraphQLConfig>();
-
+        let gas_price_provider = ctx.data_unchecked::<GasPriceProvider>();
         let query: &ReadView = ctx.data_unchecked();
+
         let latest_block: Block<_> = query.latest_block()?;
         let block_height = u32::from(*latest_block.header().height());
+        let gas_price = gas_price_provider
+            .known_gas_price(block_height.into())
+            .await
+            .ok_or(async_graphql::Error::new(format!(
+                "Gas price not found for latest block:{block_height:?}"
+            )))?;
 
         Ok(LatestGasPrice {
-            gas_price: config.min_gas_price.into(),
+            gas_price: gas_price.into(),
             block_height: block_height.into(),
         })
     }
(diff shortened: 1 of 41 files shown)


crates/fuel-core/src/schema/gas_price.rs
```diff
@@ -40,14 +38,20 @@ impl LatestGasPriceQuery {
...
     ) -> async_graphql::Result<LatestGasPrice> {
-        let config = ctx.data_unchecked::<GraphQLConfig>();
-
+        let gas_price_provider = ctx.data_unchecked::<GasPriceProvider>();
         let query: &ReadView = ctx.data_unchecked();
+
         let latest_block: Block<_> = query.latest_block()?;
         let block_height = u32::from(*latest_block.header().height());
+        let gas_price = gas_price_provider
+            .known_gas_price(block_height.into())
```
xgreenx, line 47: Is that gas price used by the block producer at the last block, or is that the gas price for the next block?

If it is gas price used for the last block, then maybe we can get it from the `Mint` transaction=)

crates/fuel-core/src/schema/gas_price.rs
```diff
@@ -77,13 +81,23 @@ impl EstimateGasPriceQuery {
...
         block_horizon: Option<U32>,
     ) -> async_graphql::Result<EstimateGasPrice> {
-        // TODO: implement dynamic calculation based on block horizon
-        //   https://github.com/FuelLabs/fuel-core/issues/1653
-        let _ = block_horizon;
+        let query: &ReadView = ctx.data_unchecked();
 
-        let config = ctx.data_unchecked::<GraphQLConfig>();
-        let gas_price = config.min_gas_price.into();
+        let latest_block: Block<_> = query.latest_block()?;
```
xgreenx, line 86: Getting `Block` is super expensive here, maybe we can get the latest height?=D

crates/services/producer/src/block_producer/gas_price.rs
```diff
@@ -0,0 +1,51 @@
...
+    }
+}
+
+/// Interface for retrieving the gas price for a block
+pub trait ProducerGasPrice {
+    /// The gas price for all transactions in the block.
+    fn gas_price(&self, params: GasPriceParams) -> Option<u64>;
+}
+
+pub struct MockProducerGasPrice {
```
xgreenx, line 31: Maybe we can use `mockall` instead?

crates/services/producer/src/block_producer/tests.rs
```diff
@@ -203,12 +210,44 @@ async fn production_fails_on_execution_error() {
...
+    //     .produce_and_execute_block_txpool(1u32.into(), Tai64::now(), 1_000_000_000)
+    //     .await
+    //     .unwrap();
+    //
+    // // then
+    // let mint_tx = changes.get_mint_tx();
+    // assert_eq!(mint_tx.gas_price, gas_price);
+}
+
+// TODO: Add test that checks the failure case for gas_price after `Executor` refactor
```
xgreenx, line 236: I didn't get why we can't implement this test now. Could you elaborate more on it, please?=)

crates/services/producer/src/block_producer/tests.rs
```diff
@@ -203,12 +210,44 @@ async fn production_fails_on_execution_error() {
     );
 }
 
+// TODO: Add test that checks the gas price on the mint tx after `Executor` refactor
```
xgreenx, line 213: We can test for now that we are passing the gas price from the GP to the executor.

crates/services/txpool/src/service.rs
```diff
@@ -138,32 +144,35 @@ impl<P2P, ViewProvider> Clone for SharedState<P2P, ViewProvider> {
...
         }
     }
 }
 
-pub struct Task<P2P, ViewProvider> {
+pub struct Task<P2P, ViewProvider, GasPriceProvider> {
     gossiped_tx_stream: BoxStream<TransactionGossipData>,
     committed_block_stream: BoxStream<SharedImportResult>,
-    shared: SharedState<P2P, ViewProvider>,
+    tx_pool_shared_state: SharedState<P2P, ViewProvider, GasPriceProvider>,
```
xgreenx, line 155: I'm not sure that we need to rename the field of the `TxPool` task to contain `tx_pool` prefix=) It is like defining `TxPoolService` in the `tx_pool` module.

crates/services/txpool/src/service.rs
```diff
@@ -120,16 +121,21 @@ impl TxStatusChange {
...
     current_height: Arc<ParkingMutex<BlockHeight>>,
     config: Config,
+    gas_price_provider: GasPriceProvider,
 }
 
-impl<P2P, ViewProvider> Clone for SharedState<P2P, ViewProvider> {
+impl<P2P, ViewProvider, GasPriceProvider> Clone
+    for SharedState<P2P, ViewProvider, GasPriceProvider>
+where
+    GasPriceProvider: Clone,
```
xgreenx, line 137: You don't need to require `Clone`, it is better to use `Arc` instead.

crates/services/txpool/src/txpool.rs
```diff
@@ -76,6 +76,33 @@ pub struct TxPool<ViewProvider> {
     database: ViewProvider,
 }
 
+pub trait TxPoolGasPrice {
+    fn gas_price(&self, block_height: BlockHeight) -> Option<GasPrice>;
+}
+
+#[derive(Debug, Clone)]
+pub struct MockTxPoolGasPrice {
```
xgreenx, line 84: Why not to use `mockall`?

xgreenx, line 84: Mocked types should be defined in the tests

crates/services/txpool/src/txpool/tests.rs
```diff
@@ -83,14 +95,15 @@ async fn insert_simple_tx_succeeds() {
...
+    let gas_price = 0;
 
     let (_, gas_coin) = context.setup_coin();
     let tx = TransactionBuilder::script(vec![], vec![])
         .script_gas_limit(GAS_LIMIT)
         .add_input(gas_coin.clone())
         .finalize_as_transaction();
     let mut txpool = context.build();
-    let tx = check_unwrap_tx(tx, &txpool.config).await;
+    let tx = check_unwrap_tx(tx, &txpool.config, gas_price).await;
```
xgreenx, line 106: The `check_unwrap_tx` function is a test helper function. If you are using zero gas price everywhere, it is better to use zero inside of this function without modifying the signature.

In the tests where gas price matters, you can use another function=)

crates/services/txpool/src/txpool.rs
```diff
@@ -76,6 +76,33 @@ pub struct TxPool<ViewProvider> {
     database: ViewProvider,
 }
 
+pub trait TxPoolGasPrice {
+    fn gas_price(&self, block_height: BlockHeight) -> Option<GasPrice>;
```
xgreenx, line 80: Hmm, I think the port should only request either the price from the last block or for the next block. Overwise, it is harder to understand what your expectations are.

xgreenx, line 80: BTW, comment would be helpful=D

crates/services/producer/src/block_producer/gas_price.rs
MitchTurner: I dislike `mockall`.

We have a lot of tests in our codebase that test the internals of our code rather the externals. We do that with `mockall`'s `expect_`, `with`, `times`, and similar methods. It's true that `mockall` would give us a struct that is basically identical to what I added, but it also gives us a bunch of things that are used to write bad tests, so I avoid it.

crates/services/txpool/src/service.rs
MitchTurner: I think having something called `SharedState` is bad. My code should never care that the data is shared with something else. Instead, it can know that it is communicating with some outside process, but that should be explicitly stated for each line of communication.

I don't have time to fix this in our code, so instead I'm trying to constrain it a little. Specifically, we have multiple types of `SharedState` in our code and this is of the `tx_pool` variety. That doesn't really help the reader much more but at least I don't accidentally mistake it for a POA, a relayer, or a P2P `SharedState`.

xgreenx: The `tx_pool::Task` only uses `SharedState` from the `tx_pool` module, and this rule applies to each service. The `tx_pool_service.shared` always will be a `SharedState` from `tx_pool` module.

Renaming `shared` to `tx_pool_shared_state` makes sense outside of the module, for example how it is done in the `fuel-core` service:

image

crates/services/producer/src/block_producer/gas_price.rs
xgreenx: If you don't like these functions, you can not use them=)

If you have a nice tool with some feature that you don't need, simply don't use them=) Auto-generation structures for testing simplify supporting of this code in the future.

If you worry that in the future someone will use `expect` and other methods, then usage of the custom structure will not solve that=D The person simply will use `mockall` and will use `expect` because he wants to=)

crates/services/txpool/src/txpool.rs
MitchTurner: That's what I started with. I definitely think that would be a cleaner interface, but the problem is I don't know what "current block" really means. Like every time you call it it gives a new gas price? or does it have some kind of tracking behind the scenes to know when a block was committed and now we're on to the next block?

I actually used this:
```pub struct GasPriceParams {
    block_height: BlockHeight,
}
```
instead of `BlockHeight` on the `ProducerGasPrice` and got lazy with the other ones. Honestly, I don't really know what the method should take yet, but `BlockHeight` seems like a good place to start, since it's absolute.

crates/services/producer/src/block_producer/tests.rs
MitchTurner: You're right. I've implemented it.

xgreenx (approved):
I still think it is better to use the `mockall` for simple cases=)

Theghost256 (approved):

PR #1760, opened 2024-03-18
Added tests to verify that the network operates with a custom chain id and base asset id

xgreenx:
Added tests to verify that the network operates with a custom chain id and base asset id. While we added customization feature, we haven't actually tested it yet.

cc @dmihal It looks like the custom base asset id works already=)

### DIFF
--- a/tests/tests/chain.rs
+++ b/tests/tests/chain.rs
@@ -171,6 +171,6 @@ async fn network_operates_with_non_zero_base_asset_id() {
     let balance = client
         .balance(&owner, Some(&new_base_asset_id))
         .await
-        .expect("Should fetch hte balance");
+        .expect("Should fetch the balance");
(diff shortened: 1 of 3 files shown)


tests/tests/chain.rs
```diff
@@ -24,3 +50,127 @@ async fn chain_info() {
...
+        .await
+        .expect("transaction should insert");
+
+    // Then
+    let expected_fee = 1;
+    assert!(matches!(result, TransactionStatus::Success { .. }));
+    let balance = client
+        .balance(&owner, Some(&new_base_asset_id))
+        .await
+        .expect("Should fetch hte balance");
```
bvrooman, line 174:
```suggestion
        .expect("Should fetch the balance");
```

bvrooman (approved):

Theghost256 (approved):

Issue #1753, opened 2024-03-13
Upgradable consensus parameters


xgreenx:
- Make `ConsensusParameters` and `GasCost` versionable(`fuel-vm` and `fuel-core`).
- Create a new type of the transaction that allow permissioned upgrade of consensus parameters. It gasless, it should support predicates and signatures(it should be customisable).
- Add new table to store CP. Update application block header to contain version of the CP. Implement simple increasing of the version during procession of a new transaction. Update block producer and executor to use a new CP from the table.

MitchTurner:
I feel like versions for CP could mean two things: change the values, but have the same fields, or add a new type with different/ new fields.

What are we thinking here?

xgreenx:
It means both. The main point is to indicate that there are new consensus parameters. How exactly they look like is defined by serialization and deserialization.

MitchTurner:
> It means both. The main point is to indicate that there are new consensus parameters. How exactly they look like is defined by serialization and deserialization.

Meaning that the mapping of a version to the params is for the serialized parameters and not for the current Rust type?

xgreenx:
Yes. The database will contain a table where the key is a version, and the value is serialized consensus parameters.

PR #1767, opened 2024-03-19
Added consensus parameters version and state transition version to the `ApplicationHeader`

xgreenx:
It is a preparation of the public types for the next changes:

https://github.com/FuelLabs/fuel-core/issues/1754
https://github.com/FuelLabs/fuel-core/issues/1753


crates/services/producer/src/block_producer/tests.rs
```diff
@@ -138,6 +142,66 @@ async fn can_produce_next_block() {
     assert!(result.is_ok());
 }
 
+#[tokio::test]
+async fn can_produce_next_block_that_contains_non_zero_versions() {
```
MitchTurner, line 146: We should include the SUT in the test name IMO: `produce_and_execute_block_txpool`.

Also, this test is probably testing too much. We should probably split this into a test for param version and state transition version.

MitchTurner: I also don't really understand why we should care that the version is non-zero... or why we should assume `rng.gen()` is non-zero for that matter.

MitchTurner (approved):
LGTM

Issue #1741, opened 2024-03-07
The fuel header should mark imported content from the L1


xgreenx:
Currently, each node simply imports all events received from the ethereum node. But in the case if the ethereum node returns different events than expected, it is impossible to prove that the state is incorrect(like some messages are missed). It is not critical in the case of missing events because we can index them later. But in the case of incorrect events(not real events) it can bring the network into an incorrect state.

For the safety of the network, it would be better if the block producer also posts the hash of the imported events along with the `da_block_height`. In this case, we can easily prove/identify an incorrect state.

Voxelot:
It might be better to make this a merkle root of imported events, and we'll also need to formally specify the event encoding format that is used to serialize these into bytes.

xgreenx:
I think we can hash all fields emitted by the event. In this case L1 contract can do the same on his side and it will be enough during fraud prooving.

PR #1769, opened 2024-03-20
Keep record of events from L1 in Block Header

MitchTurner:
closes https://github.com/FuelLabs/fuel-core/issues/1741

This PR introduces a new field on  the block header generated section called `event_inbox_root`. This is calculated after execution by adding all the hashes of the events from the L1 relayer to a merkle tree and generating the root.

Also renamed the message receipt root to `message_outbox_root` to keep up the theme.


Dentosal (approved):
LGTM; nice and small changeset

crates/types/src/blockchain/header.rs
```diff
@@ -365,17 +369,31 @@ impl PartialBlockHeader {
         self,
         transactions: &[Transaction],
         message_ids: &[MessageId],
+        events: &[Event],
```
Voxelot, line 372: This is the wrong event type, these events are internal to fuel and not the same as the relayer events. This should be based on the events that are used in the `process_da` function from the executor, ie. 

https://github.com/FuelLabs/fuel-core/blob/master/crates/types/src/services/relayer.rs#L11

Voxelot (changes requested):

crates/services/executor/src/executor.rs
```diff
@@ -724,13 +728,37 @@ where
...
+        // let event_inbox_root = events
+        //     .iter()
+        //     .map(|event| event.hash())
+        //     .fold(MerkleRootCalculator::new(), |mut tree, event| {
+        //         tree.push(event.as_ref());
+        //         tree
+        //     })
+        //     .root()
+        //     .into();
+
```
Voxelot, line 751: looks like this can be removed?

crates/fuel-core/src/schema/block.rs
```diff
@@ -165,7 +165,7 @@ impl Header {
 
     /// Merkle root of message receipts in this block.
     async fn message_receipt_root(&self) -> Bytes32 {
-        self.0.message_receipt_root.into()
+        self.0.message_outbox_root.into()
     }
```
Voxelot, line 170: we should also have the inbox root available here

crates/fuel-core/src/schema/block.rs
```diff
@@ -165,7 +165,7 @@ impl Header {
 
     /// Merkle root of message receipts in this block.
     async fn message_receipt_root(&self) -> Bytes32 {
```
Voxelot, line 167: We should also rename this function, as it's what's used to generate the fields of the gql api.

crates/types/src/blockchain/block.rs
```diff
@@ -218,8 +220,13 @@ impl PartialFuelBlock {
     ///
     /// Message ids are produced by executed the transactions and collecting
     /// the ids from the receipts of messages outputs.
-    pub fn generate(self, message_ids: &[MessageId]) -> Block {
-        Block::new(self.header, self.transactions, message_ids)
+    pub fn generate(self, message_ids: &[MessageId], event_inbox_root: Bytes32) -> Block {
```
Voxelot, line 223: Maybe we should clarify that these message ids are for the outbox?

```suggestion
    pub fn generate(self, outbox_message_ids: &[MessageId], event_inbox_root: Bytes32) -> Block {
```

Voxelot:
we should do something like a `sed -i -- 's/message_receipt_root/message_outbox_root/g' *` to cleanup all the old references to the previous name of the field.

crates/types/src/blockchain/header.rs
Voxelot: this is resolved now but theres some cleanup to do still.

crates/client/src/client/types/block.rs
```diff
@@ -36,7 +36,7 @@ pub struct Header {
     pub transactions_count: u64,
     pub message_receipt_count: u64,
     pub transactions_root: MerkleRoot,
-    pub message_receipt_root: MerkleRoot,
+    pub message_outbox_root: MerkleRoot,
     pub height: u32,
```
Voxelot, line 40: we should also include the event inbox root here, otherwise it won't be exposed to the rust-sdk or other consumers of the gql-client library.

crates/client/src/client/types/block.rs
```diff
@@ -77,7 +78,7 @@ impl From<schema::block::Header> for Header {
             transactions_count: value.transactions_count.into(),
             message_receipt_count: value.message_receipt_count.into(),
             transactions_root: value.transactions_root.into(),
-            message_receipt_root: value.message_receipt_root.into(),
+            message_outbox_root: value.message_outbox_root.into(),
             height: value.height.into(),
```
Voxelot, line 82: also here too

crates/client/src/client/types/block.rs
```diff
@@ -36,7 +36,8 @@ pub struct Header {
     pub transactions_count: u64,
     pub message_receipt_count: u64,
     pub transactions_root: MerkleRoot,
-    pub message_receipt_root: MerkleRoot,
+    pub message_outbox_root: MerkleRoot,
+    pub event_inbox_root: MerkleRoot,
```
Voxelot, line 40: it doesn't look like this actually gets set by anything?

MitchTurner: I'm not sure what you mean. The only constructor I see is the `From`. Is there another place you'd expect it to be set?

crates/services/executor/src/executor.rs
```diff
@@ -724,13 +728,16 @@ where
...
+        let mut root_calculator = MerkleRootCalculator::new();
+
         for da_height in next_unprocessed_da_height..=header.da_height.0 {
             let da_height = da_height.into();
             let events = self
                 .relayer
                 .get_events(&da_height)
                 .map_err(|err| ExecutorError::RelayerError(err.into()))?;
             for event in events {
+                root_calculator.push(event.hash().as_ref());
```
xgreenx, line 740: Hmm, I'm curious: does Ethereal RPC guarantee that the order of events is the same as they were emitted? @MitchTurner , Could you double-check that?=)

xgreenx (approved):

crates/client/src/client/types/block.rs
Voxelot: Ah I added this comment before I saw the latest commit

Voxelot (approved):

crates/services/executor/src/executor.rs
Voxelot: Looking at ethers, there is a field for indexing fields within each block. We could sort all the logs by this in the relayer to play it safe.

image

Voxelot: We could also consider lexicographically sorting all the event hashes before merklizing

MitchTurner: The order we receive the events shouldn't be dependent on the Ethereal RPC specifically, rather our implementation of `Relayer` as a whole. 

IF the order isn't deterministic, then I agree with @Voxelot that we could just do some deterministic sorting in our `Relayer` adapter.

Do we want to do that as part of this PR?

MitchTurner: Or we could just order them after we get them from the `Relayer`.

Voxelot: The best ordering Imo is to use log_index, as it ensures fair ordering of event processing. I'd prefer to avoid muddling our events in the executor with ethereum specific fields, so I'd be fine with making a separate PR to ensure ordering on the relayer side.

The alternative is to add some kind of ordering field to events, and making the executor sort them every time. However, this would also incur additional consensus overhead which I think we could avoid.

PR #1770, opened 2024-03-21
Feature/new fti event

### LINKED ISSUE #1742: Add a new L1 event for FTI
Add a new event - `Transaction(max_gas: u64, canoniclly_serialzied_tx: Vec<u8>)` that we can receive from the L1.

We need to know the signature for the event, so the bridge team should provide an ABI for a new event. After that, we can fully integrate it into the state transition function and int block production.

Voxelot:
closes #1742 

Adds a new relayer event type in support of forced transactions, assuming the following ABI from L1:

```
[
    event MessageSent(bytes32 indexed sender, bytes32 indexed recipient, uint256 indexed nonce, uint64 amount, bytes data)
    event Transaction(uint64 max_gas, bytes canonically_serialized_tx)
]
```

Specifically, this PR adds:

- Parsing FTI logs from ethereum
- Saving parsed FTI events to the relayer events table
- Reshuffles the message entity in fuel-core-types to be grouped under a relayer module, which also includes the new entity for RelayedTransaction's (this may cause a lot of breaking changes for the regenesis workstream, so if this seems like more trouble than it's worth I can put things back to how they were)


crates/services/relayer/src/mock_db.rs
```diff
@@ -25,6 +29,8 @@ use std::{
 #[derive(Default)]
 pub struct Data {
     pub messages: BTreeMap<DaBlockHeight, HashMap<Nonce, Message>>,
+    pub transactions:
+        BTreeMap<DaBlockHeight, HashMap<RelayedTransactionId, RelayedTransaction>>,
```
Dentosal, line 33: Why is `BTreeMap` used here? Do we need this to be sorted by `DaBlockHeight`?

Voxelot: This is just a mockdb, so I don't think it really matter a whole lot. Was just following the convention we already had for the mock messages table.

crates/types/src/entities/relayer/transaction.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+impl RelayedTransactionV1 {
+    /// The hash of the relayed transaction (max_gas (big endian) || serialized_transaction)
+    pub fn relayed_transaction_id(&self) -> RelayedTransactionId {
+        let hasher = fuel_crypto::Hasher::default()
+            .chain(self.max_gas.to_be_bytes())
+            .chain(self.serialized_transaction.as_slice());
+        // TODO: We need some kind of assurance from L1 that this ID is unique
```
xgreenx, line 98: Could you elaborate more on why it is a problem?

crates/services/relayer/src/log.rs
```diff
@@ -97,6 +122,35 @@ impl TryFrom<&Log> for EthEventLog {
                     ),
                 })
             }
+            n if n == *config::ETH_FORCED_TX => {
+                if log.topics.len() != 1 {
+                    return Err(anyhow!("Malformed topics for Message"))
```
xgreenx, line 127:
```suggestion
                    return Err(anyhow!("Malformed topics for forced Transaction"))
```

crates/types/src/entities/relayer/transaction.rs
Voxelot: as discussed, it should be caught by our existing check for duplicate transactions if it's valid. And if it's invalid, the duplicate hashes shouldn't cause any issues.

crates/fuel-core/src/executor.rs
```diff
@@ -3112,11 +3125,23 @@ mod tests {
...
+                let mut transaction = RelayedTransaction::default();
+                transaction.set_da_height(da_height.into());
+                transaction.set_max_gas(da_height);
+                transaction.set_serialized_transaction(da_height.to_be_bytes().to_vec());
+                root_calculator.push(Bytes32::from(transaction.relayed_id()).as_ref());
+                // add events to relayer
+                add_events_to_relayer(
+                    &mut relayer_db,
+                    da_height.into(),
+                    &[message.into(), transaction.into()],
```
MitchTurner, line 3143: Good. I'm glad you thought to add this.

crates/services/relayer/tests/integration.rs
```diff
@@ -124,6 +127,40 @@ async fn can_get_messages() {
...
+    // Setup the eth node with a block high enough that there
+    // will be some finalized blocks.
+    eth_node.update_data(|data| data.best_block.number = Some(100.into()));
+    let relayer = new_service_test(eth_node, mock_db.clone(), config);
+    relayer.start_and_await().await.unwrap();
+
+    relayer.shared.await_synced().await.unwrap();
+
+    for tx in expected_transactions {
+        assert_eq!(mock_db.get_transaction(&tx.relayed_id()).unwrap(), tx);
```
MitchTurner, line 160: I think we should make it clear that `start_and_await` is the SUT here:

Given an eth node with these logs,
when we run `start_and_await`,
then those logs are picked up and stored in the db as txs

crates/types/src/entities/relayer/transaction.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub fn relayed_id(&self) -> RelayedTransactionId {
+        match &self {
+            RelayedTransaction::V1(tx) => tx.relayed_transaction_id(),
+        }
+    }
+}
+
+impl RelayedTransactionV1 {
+    /// The hash of the relayed transaction (max_gas (big endian) || serialized_transaction)
+    pub fn relayed_transaction_id(&self) -> RelayedTransactionId {
```
MitchTurner, line 111: Is there a reason you went with the long name instead of just `id`?

crates/types/src/services/relayer.rs
```diff
@@ -13,20 +16,24 @@ use std::ops::Deref;
 pub enum Event {
     /// The message event which was sent to the bridge.
     Message(Message),
+    /// A transaction that was forced included from L1
```
bvrooman, line 19:
```suggestion
    /// A transaction that was force included from L1
```

crates/types/src/entities/relayer/transaction.rs
```diff
@@ -0,0 +1,123 @@
...
+impl RelayedTransaction {
+    /// Get the DA height that originated this transaction from L1
+    pub fn da_height(&self) -> DaBlockHeight {
+        match self {
+            RelayedTransaction::V1(transaction) => transaction.da_height,
+        }
+    }
+
+    /// Set the da height
+    pub fn set_da_height(&mut self, height: DaBlockHeight) {
```
bvrooman, line 61: For this setter, and all other setters in this interface, we should mark these with `#[cfg(test)]` or `#[cfg(any(test, feature = "test-helpers"))]` since they should only be used in test.

crates/services/relayer/src/log.rs
```diff
@@ -97,6 +122,35 @@ impl TryFrom<&Log> for EthEventLog {
                     ),
                 })
             }
+            n if n == *config::ETH_FORCED_TX => {
+                if log.topics.len() != 1 {
```
bvrooman, line 126: The `1` is a bit of a magic number. I looked up solidity topics and found:

> `topic[0]` always refers to the hash of the hash of the event itself, and can have up to 3 indexed arguments, which will each be reflected in the topics.

 Maybe it would be worth using a constant with a comment about why we use this value?

MitchTurner (dismissed):
Nice. LGTM. I don't have an opinion on the namespacing changes. Might want to check with someone who does before merging.

crates/types/src/entities/relayer/transaction.rs
Voxelot: done

crates/services/relayer/src/log.rs
Voxelot: added some comments for clarity

xgreenx (dismissed):

xgreenx (approved):

Issue #1619, opened 2024-01-24
Off-chain database migration


xgreenx:

MitchTurner:
I'm not sure if this is an issue or a feature. I think we should make a clearer distinction between those. Typically Scrum has Issues, Features, and Epochs, where issues are the smallest and features capture a group of issues and epochs capture a long-term business objective.

xgreenx:
It is a master issue to track progress on the migration of the off-chain table. Since we decided to split migration into two parts, on-chain and off-chain, it has its own deliverables. The whole database migration belongs to the regenesis feature, but because the regenesis feature itself took 4 months, we need smaller tickets to track work=)

xgreenx:
Is done as part of the https://github.com/FuelLabs/fuel-core/issues/1545

PR #1771, opened 2024-03-21
refactor: Prepare (re)genesis for off chain tables

segfault-magnet:
Mostly refactoring to allow for #1619 . Not final but close. Some cleanup remains here or there. That will be bundled with the actual feature itself.

The `SnapshotReader` and `SnapshotWriter` (previously the `StateReader` and `StateWriter`) now:
* have a single `read` (or `write`) method generic over the table type
* no longer use `*Config` structs in the interface -- all reading and writing is to be done over tables 
* `ChainConfig` is now written and read by the `SnapshotWriter` and `SnapshotReader` (since it is actually part of the snapshot). 

The genesis progress is now a generic `String` -> `u64` mapping with no enumerations for the key (previously we had an enum with a variant for each table). 

Every table gets its own parquet file. Json is still in a single file with the `StateConfig` schema. Depending on whether we want the offchain tables in the `StateConfig` we can either drop them or include them in the next PR.

I avoided enumerating the tables as much as possible to lessen coupling. A new table should ideally require:
1. a call to write::<NewTable> when generating the snapshot
2. an implementation of `ProcessState<NewTable>` to describe how to import it on regenesis
3. an implementation of `AsTable` to describe how (if at all) this table can be extracted from a in-memory `StateConfig`
4. a call to `workers.spawn::<NewTable>` to run a worker to import it from the snapshot.

After this PR we'll add in the off-chain tables and identify and regenerate dependent tables (ideally with batching + resumability).


Cargo.toml
```diff
@@ -76,8 +76,8 @@ fuel-core-types = { version = "0.23.0", path = "./crates/types", default-feature
 fuel-core-tests = { version = "0.0.0", path = "./tests" }
 fuel-core-xtask = { version = "0.0.0", path = "./xtask" }
 
-# Fuel dependencies
-fuel-vm-private = { version = "0.47.1", package = "fuel-vm", default-features = false }
+# Fuel dependencies 
+fuel-vm-private = { package = "fuel-vm", git = "https://github.com/FuelLabs/fuel-vm", branch = "feature/regenesis_off_chain", default-features = false }
```
segfault-magnet, line 80: Not sure what to do about this one. Branched off of the latest release. It seems there are some commits after the release that aren't compatible with `fuel-core` `master`.

Cargo.toml
```diff
@@ -76,8 +76,8 @@ fuel-core-types = { version = "0.23.0", path = "./crates/types", default-feature
 fuel-core-tests = { version = "0.0.0", path = "./tests" }
 fuel-core-xtask = { version = "0.0.0", path = "./xtask" }
 
-# Fuel dependencies
-fuel-vm-private = { version = "0.47.1", package = "fuel-vm", default-features = false }
+# Fuel dependencies 
+fuel-vm-private = { package = "fuel-vm", version = "0.47.1", git = "https://github.com/FuelLabs/fuel-vm", branch = "feature/regenesis_off_chain", default-features = false }
```
xgreenx, line 80: Why do we need to use `master` of `fuel-vm`? What doesn't work with the `0.47` release?=)

segfault-magnet: The `fuel-vm` branch `feature/regenesis_off_chain` was forked off of the `0.47.1` tag. It includes `serde::Serialize` and `serde::Deserialize` implementations for `double_key` needed for some tables.

The issue is that I cannot merge that because that would require a new release but there seems to be some changes after `0.47.1` that aren't compatible with `fuel-core`'s `master`. 

Any ideas?

xgreenx: Could you create PR on the `fuel-vm` side with se/de please?=) I think we will release it this week

crates/chain-config/src/config/state/writer.rs
```diff
@@ -157,148 +171,103 @@ impl From<ZstdCompressionLevel> for ::parquet::basic::Compression {
...
         })
     }
 
-    pub fn write(mut self, state_config: StateConfig) -> anyhow::Result<()> {
-        self.write_coins(state_config.coins)?;
-        self.write_contracts(state_config.contracts)?;
-        self.write_messages(state_config.messages)?;
-        self.write_contract_state(state_config.contract_state)?;
-        self.write_contract_balance(state_config.contract_balance)?;
+    pub fn write_state_config(
```
xgreenx, line 217: Hmm, it seems we use this function only in tests=) Would be nice to mark it `feature = "test-helpers"`

crates/chain-config/src/config/state/reader.rs
```diff
@@ -1,47 +1,53 @@
...
 
     fn next(&mut self) -> Option<Self::Item> {
         match self {
             IntoIter::InMemory { groups } => groups.next(),
-            IntoIter::Parquet { decoder } => decoder.next(),
+            IntoIter::Parquet { decoder } => Some(decoder.next()?.map(|bytes_group| {
+                let decoded = bytes_group
+                    .data
+                    .into_iter()
+                    .map(|bytes| postcard::from_bytes(&bytes).unwrap())
```
xgreenx, line 44: We need to handle the case of unsuccessful deserialization

crates/chain-config/src/config/snapshot_metadata.rs
```diff
@@ -3,184 +3,97 @@ use std::path::{
...
     }
-}
 
-impl SnapshotMetadata {
-    pub fn chain_config(&self) -> &Path {
-        &self.metadata.chain_config
+    pub fn prepend_path(&mut self, dir: &Path) {
+        self.chain_config = dir.join(&self.chain_config);
+        self.table_encoding.prepend_path(dir);
     }
```
xgreenx, line 90: It seems these methods can be private and we can call them only on `read` and `write`.

xgreenx: Also, I would prefer to remove these methods completely and put the underlying logic directly into the constructor/flush functions. But it is up to you=)

bin/fuel-core/src/cli/snapshot.rs
```diff
@@ -240,26 +277,29 @@ fn load_chain_config(
     Ok(chain_config)
 }
 
-fn open_db(path: &Path) -> anyhow::Result<Database> {
-    Database::<OnChain>::open_rocksdb(path, None)
+fn open_db(path: &Path) -> anyhow::Result<CombinedDatabase> {
+    CombinedDatabase::open(path, 1024 * 1024 * 1024)
```
xgreenx, line 281: Do we need to make `1GB` configurable?

bin/e2e-test-client/src/test_context.rs
```diff
@@ -315,10 +319,25 @@ impl Wallet {
...
             code: bytes,
+            states,
             ..
         } = config;
-        let state_root = Contract::initial_state_root(slots.iter());
-        let mut tx = TransactionBuilder::create(bytes.into(), salt, slots);
+
+        let state = states
+            .iter()
+            .map(|ContractStateConfig { key, value }| {
```
xgreenx, line 328: It seems you can use `try_from` that you've added

Cargo.toml
segfault-magnet: [PR](https://github.com/FuelLabs/fuel-vm/pull/704)

Cargo.lock
```diff
@@ -84,9 +84,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.2"
+version = "1.1.3"
```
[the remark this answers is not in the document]
segfault-magnet: Sorry, was debugging something and ran `cargo update` out of desperation for not seeing some changes.

Will revert.

tests/tests/coins.rs
```diff
@@ -709,8 +718,9 @@ mod all_coins {
 
 async fn empty_setup() -> TestContext {
     // setup config
+    let chain_config = ChainConfig::local_testnet();
     let config = Config {
-        state_reader: StateReader::in_memory(StateConfig::default()),
+        snapshot_reader: SnapshotReader::in_memory(StateConfig::default(), chain_config),
```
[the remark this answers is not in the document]
segfault-magnet: Went through the invocations, there are about 2/32 places where we could have something like `SnapshotReader::local_testnet()` that would do `SnapshotReader::in_memory(StateConfig::local_testnet(), ChainConfig::local_node())` 

The rest either have:
* some local changes to `StateConfig` + `ChainConfig::local_node()`
* Use `StateConfig::default()` (which is empty) + `ChainConfig::local_node()`
* Use their own state and chain config

Only thing that would remove duplication is `SnapshotReader::in_memory_local_node_chain_config(state_config: StateConfig)` but I can't find a good name for it.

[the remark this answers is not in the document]
xgreenx: We can have something like `SnapshotReader::in_memory().with_state_config(state_config)`. Or `SnapshotReader::in_memory(state_config)` and `SnapshotReader::in_memory(state_config).with_chain_config(chain_config)`

bin/fuel-core/src/cli/snapshot.rs
segfault-magnet: Don't know but I made it a parameter with a default.

crates/chain-config/src/config/state/writer.rs
```diff
@@ -308,72 +277,105 @@ impl StateWriter {
...
-                buffer,
-                state_file_path,
-            } => {
-                let file = std::fs::File::create(state_file_path)?;
-                serde_json::to_writer_pretty(file, &buffer)?;
-                Ok(())
+            EncoderType::Json { builder } => {
+                let state_config = builder.build()?;
+
+                let state_file_path = self.dir.join("state_config.json");
```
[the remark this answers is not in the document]
segfault-magnet: Should we use a public constant? With the `metadata.json` no expectations are placed on the naming of the state config file. Whatever is the value of the `state_config` json-object key is going to be used as the path from which to read the `StateConfig`. 

As such maybe it would be better to keep the name we're currently using (`state_config.json`) an implementation detail housed solely in `close` where it's used? As such maybe to delete the already existing constant `STATE_CONFIG_FILENAME`?

crates/chain-config/src/config/state/writer.rs
```diff
@@ -1,39 +1,53 @@
...
-        contract_balance: parquet::encode::PostcardEncoder<ContractBalanceConfig>,
-        block_height: parquet::encode::PostcardEncoder<BlockHeight>,
-        da_block_height: parquet::encode::PostcardEncoder<DaBlockHeight>,
+        compression: ZstdCompressionLevel,
+        table_encoders: std::collections::HashMap<
+            String,
+            (PathBuf, parquet::encode::Encoder<std::fs::File>),
+        >,
+        block_height: PathBuf,
+        da_block_height: PathBuf,
```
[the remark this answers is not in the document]
segfault-magnet: Could cause issues.

The way block and da heights are encoded is as a group of 1 in parquet terms. We expect this when we read the data from the parquet files (e.g. `block_height.parquet` and `da_block_height.parquet`). 

If I keep encoders then I would have to guard against `write_block_height` not being called twice. Or would have to adopt some convention as: "the last group of the file contains the block_height`. 

It seemed cleaner to keep the file paths and when somebody wants to write the block height we will overwrite the whole parquet file and always only have one group inside.

Thoughts?

crates/fuel-core/src/service/query.rs
```diff
@@ -45,7 +45,8 @@ impl FuelService {
         let id = tx.id(&self
             .shared
             .config
-            .chain_config
+            .snapshot_reader
```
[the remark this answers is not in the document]
segfault-magnet: Agreed that this will be a separate PR.

Cargo.lock
xgreenx, line 87: Could we revert this file one more time, pelase?=D

bin/fuel-core/src/cli/snapshot.rs
```diff
@@ -118,31 +154,59 @@ pub fn exec(command: Command) -> anyhow::Result<()> {
...
+
+    let block = db.on_chain().latest_block()?;
+    let mut writer = SnapshotWriter::json(output_dir);
+
+    writer.write(vec![code])?;
+    writer.write(vec![utxo])?;
+    writer.write(state)?;
+    writer.write(balance)?;
     writer.write_block_data(*block.header().height(), block.header().da_height)?;
+    writer.write_chain_config(&ChainConfig::local_testnet())?;
```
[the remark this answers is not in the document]
segfault-magnet: A snapshot must contain the chain info. 

When reading a snapshot some things are eagerly read -- like the chain config and block heights -- so `reader.chain_config()` don't have to have to return a `Result`.

So even though we may not need the chain config in the e2e tests, we still write it because of the eager-reading of it by the `SnapshotReader`.

xgreenx (approved):

Theghost256 (approved):

Theghost256 (approved):

Theghost256 (approved):

Issue #1772, opened 2024-03-22
Ensure that L1 Events received from the relayer are in a deterministic order


MitchTurner:
We are now writing the merkle root of all the events received via the relayer to the header of each block. It is important that the order of those events is the same for all nodes verifying blocks so that the merkle root is the same.

Related to https://github.com/FuelLabs/fuel-core/pull/1769

PR #1779, opened 2024-03-25
Modify Relayer service to order Events from L1 by block index

MitchTurner:
closes https://github.com/FuelLabs/fuel-core/issues/1772

***

Had to modify `MockDB` to also maintain order.

This code only works if the Relayer DB maintains order. In production it is stored as a `Mappable` with value being `[Event]`:

```
impl Mappable for EventsHistory {
    /// The key is the height of the DA.
    type Key = Self::OwnedKey;
    type OwnedKey = DaBlockHeight;
    /// The value is an events happened at the height.
    type Value = [Event];
    type OwnedValue = Vec<Event>;
}
```
So it should be good.

This approach is flakier than having the relayer reader in the executor ensure the order, but it could add a lot of extra execution cost to the executor code if we put the sorting on that side. We are already making assumptions about what the writing code is doing though. For example, if an event is missing from L1, correct order doesn't matter. So why not add more assumptions :P

***

We require that all `Logs`include the `log_index` so we can sort them. So I added a test `relayer__if_a_log_does_not_include_index_then_event_not_included` to check the sad path. **I'm not positive this is the behavior we want.** It is erroring in the `sort_events_by_log_index` method but I'm not exactly sure where that bubbles up to: `start_and_await` still returns an `Ok(Starting)` not an `Ok(StoppedWithError)`. That might be a bug in our `Service` code, or might be meant to be used differently. Not sure. So I just check that the `Log` isn't included in the test...

~I had to modify `MockMiddleware` to only let you take the batch once. Before it would error and then just try again and call the same value over and over and never exited. Making some assumptions there too.~


MitchTurner:
Another approach could be just to set the index to `0` if `None` is specified.

crates/services/relayer/src/service/get_logs.rs
```diff
@@ -109,3 +109,18 @@ where
...
     Ok(())
 }
+
+fn sort_events_by_log_index(events: Vec<Log>) -> anyhow::Result<Vec<Log>> {
+    let mut with_indexes = events
+        .into_iter()
+        .map(|e| {
+            let log_index = e
+                .log_index
+                .ok_or(anyhow::anyhow!("Log missing `log_index`: {e:?}"))?;
```
MitchTurner, line 119: The alternative here would be set to 0 if None.

I think I prefer the error, but there are still ways for this to be non-deterministic, e.g. if there are more than one log with the same index.

So, maybe we need to have additional checks that there are no duplicates? And do we want to accept batches with missing log indices?

If we don't want to worry about all the edge cases, then we could just do .unwrap_or(0) instead of .ok_or?

MitchTurner:
The `can_paginate_logs` and `test_da_height_updates` shouldn't exist. I kept them because my changes broke them, but just because our `MockMiddleware` changed behavior, not because of changed business code.

Instead of `can_paginate_logs` there should be a `MockMiddleware` that returns values in batches and we should show that they are stitched together correctly. `download_logs` shouldn't be the SUT--it's not even public.

crates/services/relayer/src/test_helpers/middleware.rs
```diff
@@ -51,7 +51,7 @@ struct InnerState {
 pub struct MockData {
     pub is_syncing: SyncingStatus,
     pub best_block: Block<TxHash>,
-    pub logs_batch: Vec<Vec<Log>>,
+    pub logs_batch: Option<Vec<Vec<Log>>>,
```
MitchTurner, line 54: Can remove `Option` now. Not doing `take` anymore.

crates/services/relayer/src/service/get_logs.rs
MitchTurner: Talked to the team about this offline. It sounds like we are okay with throwing an error here and not including the batch. In practice, this value should always be `Some`. 

In a similar vein, we can probably assume they will always be in order and all unique.

Voxelot: Are all events coming through this sort helper `fn` guaranteed to be grouped by da_height?

MitchTurner: They get sorted into blocks after they are sorted by index. So it will maintain order once sorted by block.

xgreenx (approved):

PR #1781, opened 2024-03-26
Removed the usage of the `lazy_static` from teh codebase

xgreenx:
It is always good to minimize the number of deps=D


Dentosal (approved):
Nice, especially just using the std `OnceLock` now that it's stable

PR #1783, opened 2024-03-28
Patch to use `fuel-vm 0.48.0`

xgreenx:
The PR upgrade `fuel-vm` to `0.48.0` release. Because of some breaking changes, we also adapted our codebase to follow them:

- Implementation of `Default` for configs was moved under the `test-helpers` feature. The `fuel-core` binary uses testnet configuration instead of `Default::default`(for cases when `ChainConfig` was not provided by the user).
- All parameter types are enums now and require corresponding modifications across the codebase(we need to use getters and setters). The GraphQL API remains the same for simplicity, but each parameter now has one more field - `version`, that can be used to decide how to deserialize. 
- The `UtxoId` type now is 34 bytes instead of 33. It affects hex representation and requires adding `00`.
- The `block_gas_limit` was moved to `ConsensusParameters` from `ChainConfig`. It means the block producer doesn't specify the block gas limit anymore, and we don't need to propagate this information.
- The `bytecodeLength` field is removed from the `Create` transaction.
- Removed `ConsensusParameters` from executor config because `ConsensusParameters::default` is not available anymore. Instead, executors fetch `ConsensusParameters` from the database.


crates/storage/src/codec/manual.rs
```diff
@@ -3,48 +3,5 @@
...
-};
-use std::borrow::Cow;
-
 /// The codec allows the definition of manual implementation for specific type `T`.
 pub struct Manual<T>(core::marker::PhantomData<T>);
-
-// TODO: Use `Raw` instead of `Manual` for `ContractsAssetKey`, `ContractsStateKey`, and `OwnedMessageKey`
-//  when `double_key` macro will generate `TryFrom<&[u8]>` implementation.
-
-impl Encode<ContractsAssetKey> for Manual<ContractsAssetKey> {
```
xgreenx, line 22: Now we can use `Raw` codec

Dentosal (approved):

Issue #1750, opened 2024-03-11
The block producer should take into account the total gas used by the L1 transactions


xgreenx:
Each forced transaction from L1 will also have a corresponding gas limit. The L1 contract is responsible for queuing transactions on L1 to not exceed the block gas limit.

Right now, the block producer service blindly increases the da block height to the last available height. However, this may cause a situation when we need to include more transactions than fit into the block gas limit. To avoid that, the block producer should take into account the gas used by the transactions from the L1.

shexinyi9527:
how to account the  gas used by the L1 ?

MitchTurner:
Clarification: This is for the L2 gas to include the L1 transactions on L2.

PR #1785, opened 2024-03-28
 Modify block producer to take into account the total gas used by the L1 transactions

MitchTurner:
closes https://github.com/FuelLabs/fuel-core/issues/1750


crates/fuel-core/src/service/adapters/producer.rs
```diff
@@ -123,31 +123,54 @@ impl fuel_core_producer::ports::DryRunner for ExecutorAdapter {
...
+    let da_height = DaBlockHeight(height);
+    let cost = sync
+        .database()
+        .storage::<fuel_core_relayer::storage::EventsHistory>()
+        .get(&da_height)?
+        .map(|cow| cow.into_owned())
+        .unwrap_or_default()
+        .iter()
+        .map(|event| event.cost())
+        .sum();
```
Voxelot, line 170: While probably very unlikely, I wonder if we need to do anything here to handle an overflow in case the cumulative max_gas is somehow too large.

MitchTurner: I'll change it to a saturated add and fold.

crates/fuel-core/src/service/adapters/producer.rs
```diff
@@ -123,31 +123,53 @@ impl fuel_core_producer::ports::DryRunner for ExecutorAdapter {
...
+    ) -> anyhow::Result<Vec<(DaBlockHeight, u64)>> {
         #[cfg(feature = "relayer")]
         {
             if let Some(sync) = self.relayer_synced.as_ref() {
-                sync.await_at_least_synced(height).await?;
-                sync.get_finalized_da_height()
+                sync.await_at_least_synced(starting_from).await?;
+                let highest = sync.get_finalized_da_height()?;
+                (starting_from.0..=highest.0)
+                    .map(|height| get_gas_cost_for_height(height, sync))
```
xgreenx, line 136: I'm afraid that the current usage of the Genesis da block height will lead to iterations from 0..5M.

It may not be a huge problem in the long term, but it may slow down the start of the `fuel core` for local development.

crates/fuel-core/src/service/adapters/producer.rs
```diff
@@ -123,31 +123,53 @@ impl fuel_core_producer::ports::DryRunner for ExecutorAdapter {
 
 #[async_trait::async_trait]
 impl fuel_core_producer::ports::Relayer for MaybeRelayerAdapter {
-    async fn wait_for_at_least(
+    async fn get_latest_da_blocks_with_costs(
         &self,
-        height: &primitives::DaBlockHeight,
-    ) -> anyhow::Result<primitives::DaBlockHeight> {
+        starting_from: &DaBlockHeight,
```
xgreenx, line 128: It would be nice, maybe, to move this logic to block producer. 

In this case, the block producer will have control over the number of iterations. Because we had a long delay/finalization and we have 100 blocks with a maxed gas limit, you will do 100 iterations while you need only one.

We could use something like:

```rust
#[async_trait::async_trait]
pub trait Relayer: Send + Sync {
    /// Wait for the relayer to reach at least this height and return the
    /// latest height (which is guaranteed to be >= height).
    async fn wait_for_at_least(
        &self,
        height: &DaBlockHeight,
    ) -> anyhow::Result<DaBlockHeight>;

    /// Returns the gas cost of events at the given height.
    async fn gas_cost_at_height(
        &self,
        height: &DaBlockHeight,
    ) -> anyhow::Result<Word>;
}
```

crates/fuel-core/src/service/adapters/producer.rs
```diff
@@ -123,31 +123,53 @@ impl fuel_core_producer::ports::DryRunner for ExecutorAdapter {
...
+    sync: &fuel_core_relayer::SharedState<
+        Database<crate::database::database_description::relayer::Relayer>,
+    >,
+) -> anyhow::Result<(DaBlockHeight, u64)> {
+    let da_height = DaBlockHeight(height);
+    let cost = sync
+        .database()
+        .storage::<fuel_core_relayer::storage::EventsHistory>()
+        .get(&da_height)?
+        .map(|cow| cow.into_owned())
```
xgreenx, line 166: You don't need to call `into_owned` here. It is the place where we want to use references if possible.

crates/services/relayer/src/service/synced.rs
```diff
@@ -1,58 +0,0 @@
...
-    // until a future state change that puts the relayer in sync
-    // with the ethereum node.
-    //
-    // previous_state, new_state => keep waiting
-    #[test_case(None, None => true; "if nothing updated with nothing then keep waiting")]
-    #[test_case(None, Some(0) => false; "if nothing updated with something then stop waiting")]
-    #[test_case(Some(0), Some(0) => false; "if something updated with same thing then stop waiting")]
-    #[test_case(Some(0), None => true; "if something updated with nothing then keep waiting")]
-    #[test_case(Some(0), Some(1) => false; "if something updated with something new then stop waiting")]
-    fn can_update_sync(previous_state: Option<u64>, new_state: Option<u64>) -> bool {
```
xgreenx, line 52: Why do we want to lose these tests?

MitchTurner: Added some tests on `Task` for the same coverage.

xgreenx (approved):

PR #1786, opened 2024-03-28
feat: Add some off chain tables to regenesis

segfault-magnet:
related to: #1619 

Doesn't close the issue, still need to migrate a few tables.

The following tables are now part of regenesis:
OnChain:
* Transactions (saved in snapshot)

OffChain:
* TransactionStatuses (saved in snapshot)
* OwnedTransactions (saved in snapshot)
* OwnedMessageIds (derived from Messages in snapshot)
* OwnedCoins (derived from Coins in snapshot)
* ContractsInfo (derived from Transactions in snapshot)

We have open questions to @xgreenx: 
1. Should we regenesize `FuelBlockIdsToHeights`? We attempted it but it caused issues with the "don't commit changes related to more than one block" guard. 
2. Also what about the restoring the following tables:
* Metadata
* Statistics
* All relayer tables
* ProcessedTransactions

There are opportunities for optimization, namely we're reading some snapshot data twice (e.g. Transactions are read once to restore the `Transactions` table and once to derive the `ContractsInfo` table). That could probably be done in one go writing to both on chain and off chian tables at once.


crates/fuel-core/src/graphql_api/storage.rs
```diff
@@ -58,20 +58,22 @@ pub mod transactions;
 pub enum Column {
     /// The column id of metadata about the blockchain
     Metadata = 0,
+    /// Metadata for genesis progress
+    GenesisMetadata = 1,
```
xgreenx, line 62: Do we need a separate column? Maybe we can use `Metadata` column instead?

bin/fuel-core/src/cli/snapshot.rs
```diff
@@ -230,33 +236,39 @@ fn full_snapshot(
...
+    write::<ContractsRawCode, OnChain>(db, group_size, &mut writer)?;
+    write::<ContractsLatestUtxo, OnChain>(db, group_size, &mut writer)?;
+    write::<ContractsState, OnChain>(db, group_size, &mut writer)?;
+    write::<ContractsAssets, OnChain>(db, group_size, &mut writer)?;
+    write::<Transactions, OnChain>(db, group_size, &mut writer)?;
 
-    let block = db.on_chain().latest_block()?;
+    let db = combined_db.off_chain();
+    write::<TransactionStatuses, OffChain>(db, group_size, &mut writer)?;
+    write::<OwnedTransactions, OffChain>(db, group_size, &mut writer)?;
```
xgreenx, line 269: Maybe we need to create an issue for the future to parallelize snapshot creation=)

xgreenx (approved):

bin/fuel-core/src/cli/snapshot.rs
segfault-magnet: https://github.com/FuelLabs/fuel-core/issues/1789

crates/fuel-core/src/graphql_api/storage.rs
segfault-magnet: Would cause concurrency issues. The metadata is sa `() -> Metadata` mapping. All threads would then need to update the same key. Since we must always write the whole `OwnedValue` that could cause one genesis worker to overwrite another if we don't sync the access between them.

If you have another solution or feel that a mutex is acceptable if it means getting rid of the `GenesisMetadata` table then please write and I'll tackle it in another PR.

crates/fuel-core/src/service/genesis/workers.rs
```diff
@@ -67,14 +73,28 @@ impl GenesisWorkers {
...
+        tracing::info!("Running off-chain imports");
         tokio::try_join!(
-            self.spawn_worker::<Coins>()?,
-            self.spawn_worker::<Messages>()?,
-            self.spawn_worker::<ContractsRawCode>()?,
-            self.spawn_worker::<ContractsLatestUtxo>()?,
-            self.spawn_worker::<ContractsState>()?,
-            self.spawn_worker::<ContractsAssets>()?,
+            self.spawn_worker_off_chain::<TransactionStatuses, TransactionStatuses>()?,
+            self.spawn_worker_off_chain::<OwnedTransactions, OwnedTransactions>()?,
```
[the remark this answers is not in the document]
segfault-magnet: The key for the table is:
```rust
pub struct OwnedTransactionIndexKey {
    pub owner: Address,
    pub block_height: BlockHeight,
    pub tx_idx: TransactionIndex,
}
```

We would have to figure out the height and tx idx of the tx and that isn't stored in the `Transactions` table. Would need to be some query over `FuelBlocks` and `Transactions` so we left it for another time after we measure how big of an impact actually saving it has on the snapshot size/performance.

Theghost256 (approved):

Theghost256 (approved):

Issue #1626, opened 2024-01-24
Implement force transaction inclusion


xgreenx:
https://www.notion.so/fuellabs/Forced-Transaction-Inclusion-FTI-b7c6391efa1646fb9d96f939adf78342

PR #1787, opened 2024-03-29
Handle FTI messages in executor

Voxelot:
Closes https://github.com/FuelLabs/fuel-core/issues/1626

todo:

tests / requirements - 

- [x]  query the status of failed forced transactions
- [x]  prevent non-chargeable txs from being included
- [x]  handle duplicate relayed txs ids (what should gql api report if there are duplicate FTIs?)
    - ~NOTE: The current solution is each duplicate gets added to the `skipped_transactions` for the block. We might want to find a different solution if UX is bad--but it's _findable_ at least currently.~ Edit: We are now reporting duplicate txs just like any other execution error for relayed txs. Since they will all have unique nonces in practice, all of their ids will be unique.
- [x]  ensure that skipped fti's due to invalid state (ie. missing utxo inputs) have their status reported
- [x]  ensure that sentry nodes also include the status of failed FTIs without failing the processing of the block
- [x]  ensure users aren't double charged for gas payed on l1


crates/fuel-core/src/graphql_api/worker_service/tests.rs
```diff
@@ -0,0 +1,81 @@
...
+        _block_height: &BlockHeight,
+        _status: TransactionStatus,
+    ) {
+        // Do nothing
+        ()
+    }
+}
+
+#[tokio::test]
+async fn run__relayed_transaction_events_are_added_to_storage() {
```
MitchTurner, line 36: Clean this up. Maybe use a builder or something to keep tidy.

tests/tests/relayer.rs
```diff
@@ -252,6 +263,38 @@ async fn messages_are_spendable_after_relayer_is_synced() {
     eth_node_handle.shutdown.send(()).unwrap();
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn can_find_failed_relayed_tx() {
```
MitchTurner, line 267: Clean this up.

crates/fuel-core/src/executor.rs
```diff
@@ -2896,52 +2904,52 @@ mod tests {
         }
 
         #[test_case::test_case(
-            Input {
-                relayer_da_height: 10,
-                block_height: 1,
-                block_da_height: 10,
-                genesis_da_height: Some(0),
-            } => matches Ok(()) ; "block producer takes all 10 messages from the relayer"
+        Input {
```
MitchTurner, line 2907: Reformat this. Idk why it is doing this.

crates/client/assets/schema.sdl
```diff
@@ -850,6 +850,7 @@ type Query {
 	messages(owner: Address, first: Int, after: String, last: Int, before: String): MessageConnection!
 	messageProof(transactionId: TransactionId!, nonce: Nonce!, commitBlockId: BlockId, commitBlockHeight: U32): MessageProof
 	messageStatus(nonce: Nonce!): MessageStatus!
+	status(id: Bytes32!): RelayedTransactionStatus
```
Voxelot, line 853: Would be nice to use a GraphQL scalar that wraps the RelayedTransactionId type. Using more specific newtypes/scalars helps with typescript parsing of queries.

MitchTurner: Oh yeah. Thanks for the reminder. I tried that and it wouldn't compile for some reason, so I just went with`Bytes32` to keep moving. I can try again though.

crates/fuel-core/src/executor.rs
MitchTurner: Seems like the problem is only on my desktop. My laptop doesn't reformat this.

crates/client/src/client.rs
```diff
@@ -968,6 +971,24 @@ impl FuelClient {
 
         Ok(proof)
     }
+
+    pub async fn relayed_transaction_status(
+        &self,
+        id: &Bytes32,
```
Voxelot, line 977: Should we also use RelayedTransactionId here?

crates/client/src/client/schema/relayed_tx.rs
```diff
@@ -0,0 +1,41 @@
...
+};
+
+#[derive(cynic::QueryFragment, Clone, Debug)]
+#[cynic(
+    schema_path = "./assets/schema.sdl",
+    graphql_type = "Query",
+    variables = "RelayedTransactionStatusArgs"
+)]
+pub struct RelayedTransactionStatusQuery {
+    #[arguments(id: $ id)]
```
Voxelot, line 15: [nit]

```suggestion
    #[arguments(id: $id)]
```

crates/client/src/client/types.rs
```diff
@@ -169,3 +170,33 @@ impl TryFrom<OpaqueTransaction> for TransactionResponse {
         })
     }
 }
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum RelayedTransactionStatus {
+    Failed {
```
Voxelot, line 176: Are there any other statuses besides failure we should keep track of?

crates/client/src/client/types.rs
```diff
@@ -145,7 +146,7 @@ impl TryFrom<SchemaTxStatus> for TransactionStatus {
                 TransactionStatus::SqueezedOut { reason: s.reason }
             }
             SchemaTxStatus::Unknown => {
-                return Err(Self::Error::UnknownVariant("SchemaTxStatus"))
+                return Err(Self::Error::UnknownVariant("SchemaTxStatus"));
```
Voxelot, line 149: [nit] seems like an unnecessary change
```suggestion
                return Err(Self::Error::UnknownVariant("SchemaTxStatus"))
```

crates/client/src/client/types.rs
MitchTurner: I think probably. But they can be defined in other Issues/PRs.

crates/client/src/client.rs
MitchTurner: The type we are using for `transaction_status` is `TxId` which is owned by `fuel-vm`, but is just an alias of `Bytes32`, so I just left this... Maybe we should consider changing both? I'm not too opinionated.

I think the best maybe is to define a type in `/client` that is the alias of `Bytes32`. I like the idea of having the client own its types, but I'm not too worried either way on this one.

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -165,6 +172,25 @@ where
...
+                    block_time: *block_time,
+                    failure: failure.clone(),
+                };
+
+                block_st_transaction
+                    .storage_as_mut::<RelayedTransactionStatuses>()
+                    .insert(&Bytes32::from(id.to_owned()), &status)?;
+            }
+            _ => {
+                // unknown execution event (likely due to a runtime upgrade)
```
Dentosal, line 192: Should we log these?

tests/tests/relayer.rs
```diff
@@ -252,6 +263,41 @@ async fn messages_are_spendable_after_relayer_is_synced() {
...
+    let block_time = Tai64::UNIX_EPOCH;
+    let failure = "lolz".to_string();
+
+    // given
+    let status = FuelRelayedTransactionStatus::Failed {
+        block_height,
+        block_time,
+        failure: failure.clone(),
+    };
+    db.off_chain_mut()
```
Voxelot, line 280: Should we have an integration test involving mocked events from the relayer? That way we know the graphql worker task properly updates the offchain table.

crates/fuel-core/src/graphql_api/storage/relayed_transactions.rs
```diff
@@ -0,0 +1,73 @@
...
+    codec::{
+        postcard::Postcard,
+        raw::Raw,
+    },
+    structured_storage::TableWithBlueprint,
+    Mappable,
+};
+use fuel_core_types::{
+    entities::relayer::transaction::{
+        // RelayedTransactionId,
```
Voxelot, line 19:
```suggestion
```

tests/tests/relayer.rs
MitchTurner: We do have a UT that is kinda doing that: `run__relayed_transaction_events_are_added_to_storage`.

I'm trying to understand the full flow you want. Are you looking to have the 
1. relayer send a _bad_ transaction that is then 
2. included in block production 
3. the worker would then see the failure event in the block and record it in the DB and
4. then we would query it with the client?

I think we have coverage of all these sections from the UT perspective, but you're right that we don't see them all working together in one flow

Voxelot: Yeah after reviewing some more, it does seem like we have full e2e coverage split across various unit tests. We can defer full e2e testing to the bridge integration if we feel confident in what we already have here.

MitchTurner: That sounds good to me.

crates/services/executor/src/executor.rs
```diff
@@ -470,33 +474,36 @@ where
...
             .with_policy(ConflictPolicy::Overwrite);
 
-        // ALl transactions should be in the `TxSource`.
-        // We use `block.transactions` to store executed transactions.
         debug_assert!(block.transactions.is_empty());
-        let mut iter = source.next(remaining_gas_limit).into_iter().peekable();
+        // initiate transaction stream with relayed (forced) transactions first,
+        // and pull the rest from the TxSource (txpool) if there is remaining blockspace available.
+        // We use `block.transactions` to store executed transactions.
+        let relayed_tx_iter = forced_transactions.into_iter().peekable();
```
Voxelot, line 502: [nit] maybe we should just move this line down to the for loop for this iterator? seems odd to have it way up here still.

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -165,6 +172,28 @@ where
...
+                    block_height: *block_height,
+                    block_time: *block_time,
+                    failure: failure.clone(),
+                };
+
+                block_st_transaction
+                    .storage_as_mut::<RelayedTransactionStatuses>()
+                    .insert(&Bytes32::from(id.to_owned()), &status)?;
+            }
+            _ => {
```
Voxelot, line 191: We probably don't need to make this non-exhaustive since changing this enum won't really work in a forkless upgrade anyways

crates/client/src/client.rs
Voxelot: Yeah alias is probably the only safe way at this point to avoid breaking changes on apis that are already present. For the new API being added here we could use the RelayedTransactionId though since it wouldn't break anything.

crates/services/executor/src/executor.rs
```diff
@@ -648,16 +680,72 @@ where
...
+        consensus_params: &ConsensusParameters,
+    ) -> Result<MaybeCheckedTransaction, ForcedTransactionFailure> {
+        let parsed_tx = Transaction::from_bytes(relayed_tx.serialized_transaction())
+            .map_err(|_| ForcedTransactionFailure::CodecError)?;
+
+        let check_tx_res = parsed_tx
+            .into_checked(header.consensus.height, consensus_params)
+            .map_err(ForcedTransactionFailure::CheckError)?;
+
+        let checked_tx = check_tx_res
```
xgreenx, line 730: Before we run any predicates, we need to check that the calculated `max_gas <= relayed_max_gas`.

crates/services/executor/src/executor.rs
```diff
@@ -676,26 +764,31 @@ where
...
         if tx_st_transaction
             .storage::<ProcessedTransactions>()
             .contains_key(tx_id)?
         {
-            return Err(ExecutorError::TransactionIdCollision(*tx_id))
+            return Err(ExecutorError::TransactionIdCollision(*tx_id));
         }
 
         let block_height = *header.height();
+        let mut relayed_tx: Option<RelayedTransactionId> = None;
```
xgreenx, line 779: It looks hacky with combination of the code below=)

image

crates/services/executor/src/executor.rs
```diff
@@ -549,14 +566,26 @@ where
             Ok(())
         };
 
-        while iter.peek().is_some() {
-            for transaction in iter {
-                execute_transaction(&mut *execution_data, transaction)?;
+        let relayed_tx_iter = forced_transactions.into_iter();
+        for transaction in relayed_tx_iter {
+            const RELAYED_GAS_PRICE: Word = 0;
+            execute_transaction(&mut *execution_data, transaction, RELAYED_GAS_PRICE)?;
```
xgreenx, line 572: What do you think about handling error and emit the event here? It will remove `let mut relayed_tx: Option<RelayedTransactionId>` hack=) Plus maybe we can remove `ExecutorError::RelayedTransactionFailed` variant

crates/services/executor/src/executor.rs
```diff
@@ -549,14 +566,26 @@ where
...
+        }
+
+        let remaining_gas_limit = block_gas_limit.saturating_sub(execution_data.used_gas);
+
+        // L2 originated transactions should be in the `TxSource`. This will be triggered after
+        // all relayed transactions are processed.
+        let mut regular_tx_iter = source.next(remaining_gas_limit).into_iter().peekable();
+        while regular_tx_iter.peek().is_some() {
+            for transaction in regular_tx_iter {
+                execute_transaction(&mut *execution_data, transaction, gas_price)?;
```
xgreenx, line 582: And here we can handle skipped transactions, in this case `execute_transaction` will be cleaner=)

image

crates/services/executor/src/executor.rs
xgreenx: Hmm, and it will remove `MaybeCheckedTransaction::RelayedCheckedTransaction`!=)

crates/fuel-core/src/executor.rs
```diff
@@ -3038,6 +3046,314 @@ mod tests {
...
+            let not_base_asset = AssetId::new([5; 32]);
+            let tx = script_tx_for_amount(100, not_base_asset);
+            let tx_bytes = tx.to_bytes();
+            relayed_tx.set_serialized_transaction(tx_bytes);
+
+            relayer_db_for_events(&[relayed_tx.into()], da_height)
+        }
+
+        #[test]
+        fn execute_without_commit__duplicated_relayed_tx_not_included_in_block() {
```
xgreenx, line 3128: We also need a test case when `relayer_max_gas`  is `1` or `0` and actually gas is `block_gas_limit`(or around). We need to reject there transactions as well

crates/fuel-core/src/executor.rs
```diff
@@ -3038,6 +3046,314 @@ mod tests {
             assert_eq!(actual, expected);
         }
 
+        #[test]
+        fn execute_without_commit__relayed_tx_included_in_block() {
```
xgreenx, line 3050: It would be nice to have a one-positive test with the message event and FTI that tries to spend that message in the same block(it is a real use case that we may have on the bridge).

crates/client/assets/schema.sdl
```diff
@@ -850,6 +850,7 @@ type Query {
 	messages(owner: Address, first: Int, after: String, last: Int, before: String): MessageConnection!
 	messageProof(transactionId: TransactionId!, nonce: Nonce!, commitBlockId: BlockId, commitBlockHeight: U32): MessageProof
 	messageStatus(nonce: Nonce!): MessageStatus!
+	status(id: RelayedTransactionId!): RelayedTransactionStatus
```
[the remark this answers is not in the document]
MitchTurner: I feel like it's likely we'll want to add more statuses in the future, but that didn't fall into the requirements of this issue.

Or is it YAGNI?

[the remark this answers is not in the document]
xgreenx: I'm okay to keep it as an enum=) So `relayedTransactionStatus` name works=)

crates/types/src/entities/relayer/transaction.rs
```diff
@@ -148,3 +149,18 @@ impl From<RelayedTransactionV1> for RelayedTransaction {
...
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Clone, Debug, PartialEq, Eq)]
+/// Potential states for the relayed transaction
+pub enum RelayedTransactionStatus {
+    /// Transaction was included in a block, but the execution was reverted
+    Failed {
+        /// The height of the block that processed this transaction
+        block_height: BlockHeight,
+        /// The time of the block that processed this transaction
+        block_time: Tai64,
```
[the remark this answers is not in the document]
MitchTurner: Hmm. You might be right. I think we just modeled it off of the  `TransactionStatus::Failure`, which has it. 

I think I'll remove it.

crates/fuel-core/src/graphql_api/storage/relayed_transactions.rs
```diff
@@ -0,0 +1,70 @@
...
+    },
+    structured_storage::TableWithBlueprint,
+    Mappable,
+};
+use fuel_core_types::{
+    entities::relayer::transaction::RelayedTransactionStatus,
+    fuel_tx::Bytes32,
+};
+
+pub struct RelayedTransactionStatuses;
```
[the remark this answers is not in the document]
MitchTurner: https://github.com/FuelLabs/fuel-core/issues/1824

crates/fuel-core/src/executor.rs
```diff
@@ -3038,6 +3046,314 @@ mod tests {
...
+            let mut relayed_tx = RelayedTransaction::default();
+            let tx = script_tx_for_amount_of_base_asset(100);
+            let tx_bytes = tx.to_bytes();
+            relayed_tx.set_serialized_transaction(tx_bytes);
+
+            relayer_db_for_events(&[relayed_tx.into()], da_height)
+        }
+
+        #[test]
+        fn execute_without_commit_with_coinbase__relayed_tx_can_have_no_base_asset_and_mint_will_have_no_fees(
```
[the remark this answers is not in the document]
MitchTurner: Yeah. Brandon and I couldn't come up with a better way to test that it doesn't charge a fee directly. So instead we just said let's submit a tx that doesn't have any base assets to implicitly show that no fee could have been taken. We also show that the mint tx doesn't have any fees taken. Neither of these are the same thing as what we really, want but are both implicitly giving us the same coverage.

I didn't want to falsely advertise the test as something it wasn't: showing that the tx would have had a fee if it was a regular tx.

If you have a suggestion on how to write a test that affirmatively shows that no L2 fee was charged for the relayed tx, I'd prefer that to this test.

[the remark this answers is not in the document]
MitchTurner: I guess we could add another test that _does_ have base asset, and show that the mint tx is still zero. Maybe we could show that the inputs and the outputs are equal as well?

crates/client/assets/schema.sdl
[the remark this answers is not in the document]
xgreenx: The name is still not updated=D

crates/fuel-core/src/query/relayed_tx.rs
```diff
@@ -0,0 +1 @@
+
```
xgreenx, line 1: It seems you can remove this file

crates/services/executor/src/executor.rs
```diff
@@ -549,14 +529,50 @@ where
...
+        while regular_tx_iter.peek().is_some() {
+            for transaction in regular_tx_iter {
+                let tx_id = transaction.id(&self.consensus_params.chain_id());
+                match execute_transaction(&mut *execution_data, transaction, gas_price) {
+                    Ok(_) => {}
+                    Err(err) => {
+                        execution_data
+                            .skipped_transactions
+                            .push((tx_id, err.clone()));
+                    }
```
xgreenx, line 568: We only need to add transaction into skipped transaction in production mode. In other cases we need to fail execution of the whole block since it is invalid

image

crates/services/executor/src/executor.rs
```diff
@@ -648,18 +667,115 @@ where
...
+    }
+
+    /// Parse forced transaction payloads and perform basic checks
+    fn validate_forced_tx(
+        relayed_tx: RelayedTransaction,
+        header: &PartialBlockHeader,
+        consensus_params: &ConsensusParameters,
+    ) -> Result<CheckedTransaction, ForcedTransactionFailure> {
+        let parsed_tx = Self::parse_tx_bytes(&relayed_tx)?;
+        let checked_tx =
```
xgreenx, line 710: What is the reason to put 2-3 lines of code into a separate function? It creates code duplication like below:

image

crates/services/executor/src/executor.rs
```diff
@@ -648,18 +667,115 @@ where
...
+        if actual_max_gas > claimed_max_gas {
+            return Err(ForcedTransactionFailure::InsufficientMaxGas {
+                claimed_max_gas,
+                actual_max_gas,
+            });
+        }
         Ok(())
     }
 
+    fn predicates_on_tx_are_valid(
```
xgreenx, line 769: Hmm, `into_checked` perfromces all checks inside. so we can remove `check_predicate`

image

crates/services/executor/src/executor.rs
```diff
@@ -648,18 +667,115 @@ where
...
+    fn validate_forced_tx(
+        relayed_tx: RelayedTransaction,
+        header: &PartialBlockHeader,
+        consensus_params: &ConsensusParameters,
+    ) -> Result<CheckedTransaction, ForcedTransactionFailure> {
+        let parsed_tx = Self::parse_tx_bytes(&relayed_tx)?;
+        let checked_tx =
+            Self::get_checked_tx(parsed_tx, *header.height(), consensus_params)?;
+        Self::tx_is_valid_variant(&checked_tx)?;
+        Self::relayed_tx_claimed_enough_max_gas(&checked_tx, &relayed_tx)?;
```
xgreenx, line 713: We need to do this check even before `Self::get_checked_tx`, because `into_checked` may be heavy in the case of `Create` transactions.

crates/fuel-core/src/executor.rs
```diff
@@ -3038,6 +3040,485 @@ mod tests {
...
+            let events = result.events;
+            let fuel_core_types::services::executor::Event::ForcedTransactionFailed {
+                failure: actual,
+                ..
+            } = &events[0]
+            else {
+                panic!("Expected `ForcedTransactionFailed` event")
+            };
+            let expected = "Insufficient max gas.";
+            assert!(actual.starts_with(expected));
```
xgreenx, line 3244: Instead of using hardcoded strings you can generate them from corresponding error variant.

```rust
assert_eq!(ForcedTransactionFailure::MaxGas.to_string(), actual);
```

The same applies to all tests

crates/fuel-core/src/executor.rs
```diff
@@ -3038,6 +3040,485 @@ mod tests {
...
+            let mut relayed_tx = RelayedTransaction::default();
+            let tx = script_tx_for_amount(100);
+            let tx_bytes = tx.to_bytes();
+            relayed_tx.set_serialized_transaction(tx_bytes);
+            relayed_tx.set_max_gas(max_gas);
+            let mut bad_relayed_tx = relayed_tx.clone();
+            let new_nonce = [9; 32].into();
+            bad_relayed_tx.set_nonce(new_nonce);
+            // let mut bad_tx = tx.clone();
+            // bad_relayed_tx.set_serialized_transaction(bad_tx_bytes);
```
xgreenx, line 3311: I think we can remove it=)

crates/fuel-core/src/executor.rs
MitchTurner: I should have been clearer:
```
            assert!(actual.starts_with(expected));
 ```
 Is not an `assert_eq` it's checking that the start of the message is right.
 
 I can rename the `expected` to something better.

File: crates/services/executor/src/executor.rs (hunk 26 above)
Comment (MitchTurner) [reply crates/services/executor/src/executor.rs:718]: The purpose was to clean up the code. It goes a long way when you see a function called `validate_forced_tx` to be able to immediately read, line-by-line, each of the validations that are happening and not get distracted with the implementation. I didn't do it to reduce the numbers of lines. And "is a valid variant" is separate from "the max gas is correct" when it comes to requirements, so I wrote them as separate, readable requirements.

**Review by xgreenx — APPROVED** [review]
Thank you @MitchTurner for finishing this PR!=) Looks cleaner than the initial variant=)

And thank you one more time for many tests

File: crates/fuel-core/src/executor.rs [hunk 31]
```diff
@@ -3038,6 +3044,531 @@ mod tests {
...
+            let consensus_params = ConsensusParameters::default();
+            // let actual_max_gas = if let CheckedMetadata::Script(metadata) = tx
+            //     .into_checked(block_height.into(), &consensus_params)
+            //     .unwrap()
+            //     .metadata()
+            // {
+            //     metadata.max_gas
+            // } else {
+            //     panic!("Expected `CheckedMetadata::Script`")
+            // };
```
Voxelot, line 3263: The cleanup commit added a bunch of commented out code 😂

MitchTurner: Naturally :blush: 

These tests take so long to compile that I kinda lose track of things by the time they complete.

xgreenx (approved):

PR #1795, opened 2024-04-01
Improvements for the `fuel-core-upgradable-executor` build script

xgreenx:
- Added better handling of errors and formatting.
- Fixed the building of the `fuel-core-wasm-executor` to work outside of the `fuel-core` context. The change uses the path to the manifest file of the `fuel-core-upgradable-executor` to build the `fuel-core-wasm-executor` instead of relying on the workspace.


MitchTurner (approved):
LGTM. Just a small thought to consider.

Issue #1801, opened 2024-04-03
Unable to start the node after restart on `fuel-core 0.24.1`


xgreenx:
The off-chain worker fails inside of the `into_task` function while starting the service because of `db_tx.commit()?;`.

<img width="926" alt="image" src="https://github.com/FuelLabs/fuel-core/assets/18346821/64c61814-457f-4b5b-8fcf-1ec52f50ff1d">

We need to fix the problem to avoid committing non-block data on the startup of the service.
Add a unit test that:
1. Starts the node
1. Produces several blocks(maybe with some useful data like transactions or contract calls)
1. Shutdown the node
1. Start again and re-use data from the second step

PR #1802, opened 2024-04-03
fix: Fix commit error on GraphQL service startup 

bvrooman:
Related issues:
- Closes https://github.com/FuelLabs/fuel-core/issues/1801

This PR removes the unnecessary DB commit that occurs during startup of the GraphQL API service. Because the committed transaction has no corresponding change in block height, the state transition fails and throws an error. This error then causes startup of the Fuel node to fail.

This can currently happen in local and production builds, but we did not have coverage for this in test. This is because this requires loading the off-chain database using something like `CombinedDatabase::open(...)`, while most tests in the codebase use a default off-chain database, i.e. `from_database`:

```
    pub async fn from_database(
        database: Database,
        config: Config,
    ) -> anyhow::Result<Self> {
        let combined_database =
            CombinedDatabase::new(database, Default::default(), Default::default());
        Self::from_combined_database(combined_database, config).await
    }
```

This PR introduces a test that performs the following steps:
1. Creates a node using a RocksDB database on the filesystem for on-chain, off-chain, and relayer ("combined") data
2. Sends some transactions to the node
3. Shuts down the node
4. Creates a second node by loading the same database on the filesystem for on-chain, off-chain, and relayer data
5. Waits for the second node to become healthy

Before removing the commit, this test failed with the observed failure:

```
The initialization of the service failed.: error occurred in the underlying datastore `NewHeightIsNotSet { prev_height: 3 }`
```

With the code change, this test passes.


crates/fuel-core/src/graphql_api/storage.rs
```diff
@@ -141,6 +142,14 @@ where
         Ok(new_tx_count)
     }
 
+    fn get_tx_count(&self) -> StorageResult<u64> {
+        const TX_COUNT: &str = "total_tx_count";
```
xgreenx, line 146: It would be nice to move it on one level upper. And reuse inside of the `increase_tx_count`

bvrooman: By one level up, do you mean the `Transactional` trait?

xgreenx: I mean reuse one constant in two functions=) Define it on file level instead of function level

xgreenx (approved):

xgreenx (approved):

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -323,10 +323,11 @@ where
         _: &StateWatcher,
         _: Self::TaskParams,
     ) -> anyhow::Result<Self::Task> {
-        let mut db_tx = self.database.transaction();
-        let total_tx_count = db_tx.increase_tx_count(0).unwrap_or_default();
-        db_tx.commit()?;
-        graphql_metrics().total_txs_count.set(total_tx_count as i64);
+        {
+            let db_tx = self.database.transaction();
```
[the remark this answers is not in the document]
xgreenx: No, we don't, but it will require changing of constraints for the port=D I'm okay with using transaction for now to just fix the issue and release a new `fuel-core` with the fix=)

PR #1803, opened 2024-04-04
Fix: Produce block when da height haven't changed

xgreenx:


MitchTurner (commented):
Makes sense. Good catch. I'm confused why we were able to merge before despite this being broken.

xgreenx:
Actually, it failed=D But yeah, it is strange why we even merged it with failed tests.

image

MitchTurner (approved):

PR #1808, opened 2024-04-06
Fetch consensus parameters from the provider

xgreenx:
Related issue https://github.com/FuelLabs/fuel-core/issues/1753

Replaced the direct usage of the `ConsensusParameters` with `ConsensusParametersProvider` that gets the consensus parameters from the database.

Some services use `ChainId`, but since `ChainId` is immutable for the network, it is okay to pass during the construction of the service(we don't need to fetch it from the database).

The change prepares the codebase to use `ConsensusParametersProvider`. The fetching of the consensus from the database will be done in the follow up PR.


crates/services/txpool/src/txpool.rs
```diff
@@ -371,19 +361,6 @@ where
             return Err(Error::NoMetadata)
         }
 
-        // verify max gas is less than block limit
-        let block_gas_limit = self
-            .config
-            .chain_config
-            .consensus_parameters
-            .block_gas_limit();
-        if tx.max_gas() > block_gas_limit {
```
xgreenx, line 380: The conversion into a `Checked` transaction verifies that `max_gas` is less than `tx_max_gas`.

Dentosal (approved):

PR #1809, opened 2024-04-06
Fetch `ConsensusParameters` from the database

xgreenx:
Related issue https://github.com/FuelLabs/fuel-core/issues/1753

The change fetches the `ConsensusParameters` from the database and caches it inside of the `ConsensusParametersProvider`. The `ConsensusParametersProvider` has a service that runs in the background and listens for a new block from the block importer. If a new block uses a new consensus parameters version, the `ConsensusParametersProvider` will update the internal `latest_consensus_parameters_version` variable and cache a new `ConsensusParameters`.

### Before requesting review


crates/services/txpool/src/service.rs
```diff
@@ -254,6 +250,7 @@ where
 
                     let acceptance = match checked_tx {
                         Ok(tx) => {
+                            let id = tx.transaction().cached_id().expect("`Checked` tx should have cached id");
```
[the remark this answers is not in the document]
xgreenx: It is the `expect` because this kind of information should be available now. But because we forgot to expose it, we don't have access right now. I don't want to use `?` because it will hide the problem if, at some point, this information is not available.

bvrooman (approved):

crates/fuel-core/src/service/adapters/consensus_parameters_provider.rs
```diff
@@ -0,0 +1,177 @@
...
+use std::{
+    collections::HashMap,
+    sync::Arc,
+};
+
+#[derive(Clone, Debug)]
+pub struct SharedState {
+    latest_consensus_parameters_version: SharedMutex<ConsensusParametersVersion>,
+    consensus_parameters:
+        SharedMutex<HashMap<ConsensusParametersVersion, Arc<ConsensusParameters>>>,
```
bvrooman, line 36: I would be interested to understand better if we really need `Arc<ConsensusParameters>` here instead of just `ConsensusParameters`:
- If the `SharedState` is already shared, then the contents of `SharedState` are also shared, and the hashmap values can be accessed by cloning this `Arc` and locking `consensus_parameters`
- If `SharedState` owns the `HashMap`,  and the `HashMap` owns the value type, we know that the lifetime of the hashmap is maintained by `SharedState`, which in turn has a lifetime equal to the long running task. We have clear ownership semantics so we probably don't need RC here

Let me know if that sounds possible or if I'm simply misunderstanding anything here.

xgreenx: We can't return a reference to the `ConsensusParameters` because it is owned by the `MutexGuard`, so it has a local lifetime. Plus, if we could, it would create a situation where you can return a reference, lock the `HashMap`, and remove the value, making the reference invalid.

Dentosal (approved):

Dentosal (approved):

Issue #1790, opened 2024-03-29
Add migration for remaining off chain tables 


segfault-magnet:
@xgreenx wrote:

> We need to migrate as well:
> `FuelBlocks` - for that we need to disable guard on commit.
> `FuelBlockIdsToHeights` - Cab be generated form the `FuelBlocks`.
> `SpentMessages` - can be derived from Transactions or removed entirly
> `ProcessedTransactions` - can be derived from Transactions
> `FuelBlockConsensus`

and

> Actually, maybe transactions and blocks should be migrated to the off-chain database. Because we are only migrating them for GraphQL service and only to support queries for old data.

and

> [...] the idea is to have `OldFuelBlocks` and `OldTransactions` as part of the off-chain database. If the requested block is lower than genesis block height, we go to `off_chain::OldFuelBlocks` , otherwise to `on_chain::FuelBlocks`. In the case of transactions, yeah, first we go to on_chain::Transactions  and if no success, go to `off_chain::OldTransactions`

From the above, we should have a table for blocks and transactions in the off chain db and populate it upon regenesis.

PR #1811, opened 2024-04-08
Store old blocks and txs after regenesis

Dentosal:
Work towards #1790

This PR adds `OldFuelBlocks`, `OldFuelBlockConsensus` and `OldTransactions` tables to the off-chain database. If the requested block is lower than genesis block height, we go to `off_chain::OldFuelBlocks` , otherwise to `on_chain::FuelBlocks`. For transactions, we try `on_chain::Transactions` and fallback to `off_chain::OldTransactions`.

Also exposes `fuel_core_bin` as a library to help with e2e testing.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog (no breaking changes)

### Before requesting review

### Before merge
- [x] Retarget to master branch


bin/fuel-core/chainspec/testnet/state_config.json
```diff
@@ -13,6 +13,9 @@
       "balances": []
     }
   ],
+  "blocks": [],
+  "block_consensus": [],
+  "transactions": [],
```
xgreenx, line 18: I think we don't need to migrate blocks, transactions, and consensus information via `StateConfig`. The `StateConfig` is only used for the JSON serialization and only used to generate a genesis block in a user-friendly form. For the real re-genesis process, we use only parquet.

crates/chain-config/src/config/state.rs
```diff
@@ -364,14 +408,56 @@ impl AddTable<ContractsLatestUtxo> for StateConfigBuilder {
...
-    fn as_table(&self) -> Vec<TableEntry<Transactions>> {
-        Vec::new() // Do not include these for now
+    fn as_table(&self, chain_id: &ChainId) -> Vec<TableEntry<Transactions>> {
+        self.transactions
+            .iter()
+            .map(|tx| TableEntry {
+                key: tx.id(chain_id),
+                value: tx.clone().into(),
+            })
+            .collect()
```
xgreenx, line 418: It seems if we remove transactions from the `StateConfig` we don't need to pass `ChainId`=)

crates/fuel-core/src/database.rs
```diff
@@ -102,6 +102,13 @@ impl Database<OnChain> {
             .map(|(_, block)| block)
             .ok_or_else(|| not_found!("FuelBlocks"))
     }
+
+    pub fn first_block_height(&self) -> StorageResult<BlockHeight> {
```
xgreenx, line 106: You can re-use `genesis_block` function

image

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -117,12 +119,56 @@ impl DatabaseBlocks for ReadView {
...
+                IterDirection::Forward => self
+                    .off_chain
+                    .old_blocks(height, direction)
+                    .chain(self.on_chain.blocks(height, direction))
+                    .into_boxed(),
+                IterDirection::Reverse => self
+                    .on_chain
+                    .blocks(height, direction)
+                    .chain(self.off_chain.old_blocks(height, direction))
+                    .into_boxed(),
```
xgreenx, line 160:
```suggestion
                IterDirection::Forward => self
                    .off_chain
                    .old_blocks(None, direction)
                    .chain(self.on_chain.blocks(None, direction))
                    .into_boxed(),
                IterDirection::Reverse => self
                    .on_chain
                    .blocks(None, direction)
                    .chain(self.off_chain.old_blocks(None, direction))
                    .into_boxed(),
```

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -117,12 +119,56 @@ impl DatabaseBlocks for ReadView {
...
+            }
+        }
     }
 
     fn latest_height(&self) -> StorageResult<BlockHeight> {
         self.on_chain.latest_height()
     }
+
+    fn first_height(&self) -> StorageResult<BlockHeight> {
```
xgreenx, line 169: The `genesis_height` sounds better=) Plus, since we may have many genesis heights, maybe it would be nice to highlight it somehow `active_genesis_height`, `actual_genesis_height`, `on_chain_genesis_height`=)

crates/fuel-core/src/graphql_api/storage/old.rs
```diff
@@ -0,0 +1,99 @@
...
+
+impl Mappable for OldFuelBlocks {
+    type Key = Self::OwnedKey;
+    type OwnedKey = BlockHeight;
+    type Value = Self::OwnedValue;
+    type OwnedValue = CompressedBlock;
+}
+
+impl TableWithBlueprint for OldFuelBlocks {
+    type Blueprint = Plain<Postcard, Postcard>;
```
xgreenx, line 32:
```suggestion
    type Blueprint = Plain<Primitive<4>, Postcard>;
```

crates/fuel-core/src/graphql_api/storage/old.rs
```diff
@@ -0,0 +1,99 @@
...
+
+impl Mappable for OldFuelBlockConsensus {
+    type Key = Self::OwnedKey;
+    type OwnedKey = BlockHeight;
+    type Value = Self::OwnedValue;
+    type OwnedValue = Consensus;
+}
+
+impl TableWithBlueprint for OldFuelBlockConsensus {
+    type Blueprint = Plain<Postcard, Postcard>;
```
xgreenx, line 59:
```suggestion
    type Blueprint = Plain<Primitive<4>, Postcard>;
```

tests/tests/regenesis.rs
```diff
@@ -0,0 +1,252 @@
...
+    pub async fn spawn<I, S>(args: I) -> Self
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<OsStr>,
+    {
+        // Generate temp params
+        let db_dir = tempdir().expect("Failed to create temp dir");
+        let port = portpicker::pick_unused_port().expect("Failed to pick unused port");
+
+        let mut process = Command::new(env!("CARGO"))
```
xgreenx, line 58: It would be nice if, instead of starting a new child process that will compile the binary,  we used a `Command` directly from `fuel_core_bin`.

image

Like we did for snapshot tests in the https://github.com/FuelLabs/fuel-core/blob/master/bin/fuel-core/src/cli.rs#L181

tests/Cargo.toml
```diff
@@ -47,6 +47,7 @@ futures = "0.3"
 hyper = { workspace = true, features = ["server"] }
 insta = { workspace = true }
 itertools = { workspace = true }
+portpicker = "0.1"
```
xgreenx, line 50: If we use stuff from `fuel-core-bin`, then we don't need to pick a port=)

crates/chain-config/src/config/state.rs
Dentosal: True.

crates/fuel-core/src/graphql_api/database.rs
Dentosal: Maybe `latest_genesis_height`? I'll rename it.

crates/fuel-core/src/database.rs
Dentosal: Where's that defined? `rg 'fn genesis_block'` doesn't find it in this branch or `master`.

xgreenx: Oh, my bad, it is not merged=D https://github.com/FuelLabs/fuel-core/pull/1826

bin/fuel-core/chainspec/testnet/state_config.json
xgreenx: It doesn't hurt, but I prefer to avoid dead code since we need to support it, plus it affects our interfaces=)

I don't see any usage of it right now. Transactions are unique and can't be used by other transactions(only `UtxoId` contains `TxId`, but we can use the `coins` field to generate any input coins). Blocks and consensus info contain only transactions and are not used by transactions as well, so it is also hard to imagine why we need it for `StateConfig`.

crates/fuel-core/src/graphql_api/storage/old.rs
Dentosal: e7ef64c

crates/fuel-core/src/graphql_api/storage/old.rs
Dentosal: e7ef64c

crates/fuel-core/src/graphql_api/database.rs
Dentosal: e7ef64c

crates/fuel-core/src/graphql_api/database.rs
Dentosal: e7ef64c

bin/fuel-core/chainspec/testnet/state_config.json
Dentosal: 3a721e5bf

crates/chain-config/src/config/state.rs
Dentosal: 3a721e5bf

crates/fuel-core/src/database.rs
Dentosal: I'm going to keep it as-is now, and after that's merged we can combine the implementations.

tests/tests/regenesis.rs
Dentosal: 8d5f1437e

tests/Cargo.toml
Dentosal: Really? I think running two tests like that in parallel will break if we do not randomize ports.

bin/fuel-core/src/main.rs
```diff
@@ -1,15 +1,8 @@
-#![deny(clippy::arithmetic_side_effects)]
-#![deny(clippy::cast_possible_truncation)]
-#![deny(unused_crate_dependencies)]
-#![deny(warnings)]
```
xgreenx, line 4: I think you can keep it in both places=)

crates/fuel-core/src/graphql_api/storage/old.rs
xgreenx: It is still `Postcard`=)

tests/tests/regenesis.rs
```diff
@@ -0,0 +1,211 @@
...
+        let mut args = vec![
+            "_IGNORED_",
+            "--port",
+            &port_str,
+            "--db-path",
+            db_dir.path().to_str().unwrap(),
+        ];
+        args.extend(extra_args);
+
+        let process = task::spawn(run::exec(run::Command::parse_from(args)));
```
xgreenx, line 55: Hmm, do you really need to use `run::Command` to start a node? It seems you still can use `FuelService::new_node` to do that and work directly with the `SharedState.graph_ql.bound_address` to send transactions and get data. In this case you don't need to specify the port. 

Or you can split the `run::exec`command into two functions where one function has the logic around stop signal:

image


And another one contains all initialisation logic:
image

bin/fuel-core/src/main.rs
Dentosal: The first two don't affect the file at all. `#![deny(unused_crate_dependencies)]` will error about many crates that only the library depends on directly, and not the `main.rs`.

crates/fuel-core/src/graphql_api/storage/old.rs
Dentosal: 0f6e3516b

tests/tests/regenesis.rs
Dentosal: Yep, done in b630cd9a3

tests/tests/regenesis.rs
```diff
@@ -0,0 +1,216 @@
...
+}
+impl FuelCoreDriver {
+    pub async fn spawn(from_snapshot: Option<&Path>) -> anyhow::Result<Self> {
+        // Generate temp params
+        let db_dir = tempdir()?;
+
+        let mut config = Config::local_node();
+        config.debug = true;
+        config.block_production = Trigger::Instant;
+        config.combined_db_config.database_path = db_dir.path().to_path_buf();
```
xgreenx, line 51: I still think it would be better to re-use the part of the `run::exec` command and use CLI arguments to init all what you need=) In this case, you will not duplicate the logic of `fuel-core-bin`.

crates/fuel-core/src/service/genesis/exporter.rs
```diff
@@ -84,7 +86,9 @@ where
             ContractsLatestUtxo,
             ContractsState,
             ContractsAssets,
-            Transactions
+            Transactions,
+            FuelBlocks,
+            SealedBlockConsensus
```
xgreenx, line 91: I've just realized that while it works for the first-time regenesis, it will not work for the next regenesises because it ignores `Old*` tables.

We need to migrate(export and import) them as well as part of the off-chain snapshot.

You have two variants, you can export and import all `Old*` tables, and it should work. Or, during exporting instead of exporting them into `Transactions`/`FuelBlocks`/`SealedBlockConsensus`, you can export directly into `OldTransactions/`OldFuelBlocks`/`OldFuelBlockConsensus`.

tests/tests/regenesis.rs
Dentosal: Hmmm I though you argued against doing that above. I think this approach here is quite clean. I'm not sure what logic is duplicarted here?

xgreenx: image

We can reuse the code from here. Instead of creating `Config::local_node()` we can get it from the command like `command.get_config()?`. As an example, we don't use `Config::local_node()` inside of the `fuel-core-bin` anymore, instead it uses config for the testnet=) So to avoid different behavior it is better to use part of the `run::exec`=)

xgreenx (approved):

PR #1814, opened 2024-04-09
Bugfix: the `iter_all_by_prefix` was not working for all tables

xgreenx:
We used `RocksDB` filtering by prefix during `iter_all_by_prefix`. But if the prefix is not set for the column, the `RocksDB` ignores it. 

We don't set the prefix for all tables because of performance purposes. But it means that iteration sometimes can be wrong.

The change adds a `Rust` level filtering since not all tables really need a prefix.

The issue found by @segfault-magnet in https://github.com/FuelLabs/fuel-core/pull/1799#discussion_r1556745105

### Before requesting review


Dentosal (dismissed):
Less traits, more macros. Color me happy.

PR #1816, opened 2024-04-10
Use state transition bytecode from the database

xgreenx:
Related issue https://github.com/FuelLabs/fuel-core/issues/1754

Updated the upgradable executor to fetch the state transition bytecode from the database when the version doesn't match a native one. This change enables the WASM executor in the "production" build and requires a `wasm32-unknown-unknown` target.

The executor has a `VERSION` constant that must be updated when the behavior of the state transition function is changed. That constant is used to decide what executor to use(WASM or native).

## Checklist

### Before requesting review


crates/services/upgradable-executor/src/executor.rs
```diff
@@ -291,4 +404,407 @@ where
...
+
+                // Then
+                assert_eq!(Ok(()), result);
+            }
+        }
+
+        // The test verifies that `Executor::get_module` method caches the compiled WASM module.
+        // If it doesn't cache the modules, the test will fail with a timeout.
+        #[test]
+        #[ntest::timeout(40_000)]
```
Dentosal, line 791: nit: Potentially flaky, especially in the CI. Cannot thing of an easy way to do this better, though.

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -291,4 +404,407 @@ where
...
+            let wrong_version = Executor::<Storage, DisabledRelayer>::VERSION + 1;
+            let block = valid_block(wrong_version);
+
+            // When
+            let result = executor
+                .execute_without_commit(ExecutionTypes::Validation(block))
+                .map(|_| ());
+
+            // Then
+            assert!(result.is_err());
```
Dentosal, line 636: nit: use `.expect_err` for better error output

Dentosal (approved):
The logic itself looks good. I left a couple of nits on tests.

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -225,16 +307,48 @@ where
...
+        let block_version = block.state_transition_version();
+        if block_version == Self::VERSION {
+            match &self.execution_strategy {
+                ExecutionStrategy::Native => self.native_execute_inner(block, options),
+                ExecutionStrategy::Wasm { module } => {
+                    self.wasm_execute_inner(module, block, options)
+                }
+            }
+        } else {
+            let module = self.get_module(block_version)?;
```
Voxelot, line 319: Do we have a way to ensure that the initial version of the executor is available in wasm? For example, if there have been several upgrades to the network and a node on the latest native version attempts to sync the network, will the bytecode for the initial executor version (before any upgrades occured) be available?

xgreenx: No, we don't have a way. It should already be part of the snapshot/genesis block.

Dentosal (dismissed):
LGTM. @Voxelot still has one open comment, so you might want to wait for him to review as well. Also needs conflict resolution.

crates/services/upgradable-executor/src/executor.rs
xgreenx: Automated the process of insertion of the Wasm bytecode into the `ChainConfig` https://github.com/FuelLabs/fuel-core/pull/1826

Dentosal (approved):

Issue #1820, opened 2024-04-11
Parquet snapshot currently doesn't tolerate missing tables


segfault-magnet:
@MujkicA found during e2e tests that if the parquet snapshot is used then we will error out if any of the tables aren't present in the snapshot.

Although this isn't expected in real life, it does create an unnecessary edge case where if you don't do anything with the core (not even a single tx) then the snapshot generated from such a state will be unimportable (because, for example, you have no file for the `Transactions` table because you've had no transactions).

Issue #1819, opened 2024-04-11
Genesis progress report


segfault-magnet:
Genesis could take a long time depending on the snapshot size. Some feedback on the progress made should be given periodically.

Issue #1800, opened 2024-04-02
Genesis import graceful shutdown


segfault-magnet:
The cancellation signal has not been propagated to the genesis state import process. This is essential for resumability.

PR #1821, opened 2024-04-11
feat: (re)genesis graceful shutdown

segfault-magnet:
closes: #1800 
closes: #1819
closes: #1820 

Resumability:
(Re)genesis can now be paused/resumed. In order to do so the start of `FuelService` in `cli/run` needed to be split into `starting` and `running` phases, both interrupt-able by signals.

Progress:
Progress bars are now shown when `stderr` is a `tty`. Otherwise logs will be printed containing the current group, number of groups, table name and ETA to completion.

Bars:
image

Logs:
image

Empty tables:
@MujkicA found an edge case while e2e testing. Fixed here by allowing tables to not be present in snapshot when using parquet.


CHANGELOG.md
```diff
@@ -10,9 +10,11 @@ Description of the upcoming release here.
 ### Fixed
 
 - [#1814](https://github.com/FuelLabs/fuel-core/pull/1814): Bugfix: the `iter_all_by_prefix` was not working for all tables. The change adds a `Rust` level filtering.
+- [#1821](https://github.com/FuelLabs/fuel-core/pull/1821): Can handle missing tables in snapshot.
 =======
 ### Added
 
+- [#1821](https://github.com/FuelLabs/fuel-core/pull/1821): Shutdown signal propagated to (re)genesis. Progress indication added
```
Dentosal, line 17:
```suggestion
- [#1821](https://github.com/FuelLabs/fuel-core/pull/1821): Propagate shutdown signal to (re)genesis. Also add progress bar for (re)genesis.
```

Dentosal (approved):

crates/fuel-core/src/service/genesis/importer/import_task.rs
```diff
@@ -97,49 +101,51 @@ where
...
 
-        tracing::info!(
-            "Finishing genesis runner. Read: {} wrote into {}",
-            Logic::TableInSnapshot::column().name(),
-            Logic::TableBeingWritten::column().name()
-        );
+        result?;
 
-        result
+        Ok(!is_cancelled)
```
xgreenx, line 134: Maybe makes more sense to return an error that import process was canceled

crates/fuel-core/src/service/genesis/importer/progress.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+}
+
+pub struct MultipleProgressReporter {
+    multi_progress: MultiProgress,
+}
+
+impl MultipleProgressReporter {
+    pub fn new_sterr() -> Self {
+        Self::new_target(ProgressDrawTarget::stderr())
```
xgreenx, line 70: Why do we report progress into stderr?

crates/fuel-core/src/service/genesis/exporter.rs
```diff
@@ -57,12 +57,13 @@ where
...
         group_size: usize,
     ) -> Self {
+        let (_, receiver) = watch::channel(State::Started);
         Self {
             db,
             prev_chain_config,
             writer,
             group_size,
-            task_manager: TaskManager::new(CancellationToken::new()),
+            task_manager: TaskManager::new(receiver.into()),
```
xgreenx, line 66: Do we even use it in the case of exporter? It would be nice to avoid passing unused parameters. Or maybe it makes sense to process export cancelation as well=)

crates/fuel-core/src/service/genesis/importer/progress.rs
segfault-magnet: As opposed to `stdout`? `stderr` is traditionally the place for diagnostic output.

Or did you mean something else?

crates/fuel-core/src/service/genesis/exporter.rs
segfault-magnet: It probably would be more user friendly if the process could be stopped. 

Do something like write into a tempdir in the same fs (so that moving/renaming is cheap), if stopped remove the temp dir. If not stopped, at the end just rename the dir to the location originally indented for the snapshot.

I could implement that in a follow up PR if you agree.

xgreenx: I think we don't need to bother with temporary storage and resumability=) We can just shut the process and next time start from scratch.

xgreenx: If it is hard to integrate in this PR, you can leave a TODO comment with the link to the issue and do in follow up PR

crates/fuel-core/src/service/genesis/importer/progress.rs
segfault-magnet: Some of the motivation is mentioned [here](https://unix.stackexchange.com/questions/331611/do-progress-reports-logging-information-belong-on-stderr-or-stdout).

Some points:
- Posix defines `stderr` as unbuffered -- so a natural place for "fast" updating spinners and UI elements
- `stderr` was originally created so that you can dump there any info that isn't meant for consumption by the next program in the pipeline, e.g. `cat data.txt | ./process_data.py | sort > data.csv` `process_data.py` can print any spinners and progress bars to `stderr` without impacting `sort` and the rest of the pipeline. `fuel-core` isn't meant for such usage but the tradition still remains.

crates/fuel-core/src/service/genesis/exporter.rs
segfault-magnet: I'll do it in this PR. 

> I think we don't need to bother with temporary storage and resumability=) 

I'd just implement the 'rename-when-finished-delete-temp-if-cancelled` if that's ok. Seems better to leave the filesystem clean if the snapshot generation was cancelled. No way of generating a broken snapshot then (barring ungraceful shutdown ofc but even then it would be some temp file that would be left).

xgreenx (approved):

PR #1826, opened 2024-04-13
Support upgradability of the consensus parameters and state transition bytecode in genesis

xgreenx:
The changes make the state transition bytecode part of the `ChainConfig`. It guarantees the state transition's availability for the network's first blocks.

The change has many minor improvements in different areas related to the state transition bytecode:
- The state transition bytecode lies in its own file(`state_transition_bytecode.wasm`) along with the chain config file. The `ChainConfig` loads it automatically when `ChainConfig::load` is called and pushes it back when `ChainConfig::write` is called.
- The `fuel-core` release bundle also contains the `fuel-core-wasm-executor.wasm` file of the corresponding executor version.
- The regenesis process now considers the last block produced by the previous network. When we create a (re)genesis block of a new network, it has the `height = last_block_of_old_netowkr + 1`. It continues the old network and doesn't overlap blocks(before, we had `old_block.height == new_genesis_block.hegiht`).
- Along with the new block height, the regenesis process also increases the state transition bytecode and consensus parameters versions. It guarantees that a new network doesn't use values from the previous network and allows us not to migrate `StateTransitionBytecodeVersions` and `ConsensusParametersVersions` tables.
- Added a new CLI argument, `native-executor-version,` that allows overriding of the default version of the native executor. It can be useful for side rollups that have their own history of executor upgrades.
- Replaced:
  
  ```rust
           let file = std::fs::File::open(path)?;
           let mut snapshot: Self = serde_json::from_reader(&file)?;
  ```
  
  with a:
  
  ```rust
           let mut json = String::new();
           std::fs::File::open(&path)
               .with_context(|| format!("Could not open snapshot file: {path:?}"))?
               .read_to_string(&mut json)?;
           let mut snapshot: Self = serde_json::from_str(json.as_str())?;
  ```
  because it is faster in 100 times for big JSON files.
- Updated all tests to use `Config::local_node_*` instead of working with the `SnapshotReader` directly. It is the preparation of the tests for the futures bumps of the `Executor::VERSION`. When we increase the version, all tests continue to use `GenesisBlock.state_transition_btyecode = 0` while the version is different, which forces the usage of the WASM executor, while for tests, we still prefer to test native execution. The `Config::local_node_*` handles it and forces the executor to use the native version.
- Reworked the `build.rs` file of the upgradable executor. The script now has a cache used between builds to avoid recompilation of the WASM bytecode. Also, fixed the issue with outdated WASM bytecode. The script reacts on any modifications of the `fuel-core-wasm-executor` and forces recompilation(it is why we need cache), so WASM bytecode always is actual now.

## Checklist

### Before requesting review


crates/fuel-core/src/service/genesis.rs
```diff
@@ -154,31 +171,67 @@ pub async fn execute_and_commit_genesis_block(
...
+    let height;
+    let da_height;
+    let consensus_parameters_version;
+    let state_transition_bytecode_version;
+
+    // If the rollup continues the old rollup, the height of the new block should
+    // be higher than that of the old chain by one to make it continuous.
+    // The same applies to the state transition functions and consensus
+    // parameters since it is a new chain.
+    if let Some(latest_block) = config.snapshot_reader.last_block_config() {
```
MitchTurner, line 183: It would be good to have some tracing on which branch it takes here.

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -59,30 +59,37 @@ pub fn init_sub_services(
     config: &Config,
     database: CombinedDatabase,
 ) -> anyhow::Result<(SubServices, SharedState)> {
+    let chain_config = config.snapshot_reader.chain_config();
```
MitchTurner, line 62: I know that this isn't your change, but it feels bad that our `chain_config` is inside our `config.snapshot_reader`. Like, that's very unintuitive for a couple reasons: 

1. what's the difference between a `Config` and a `ChainConfig`?
1. why would anyone looking for the `consensus` (or any other `ChainConfig` value) go to the `snapshot_reader` to find that? Like, even now that I've found it, I'm worried that there must be another field by the same name someplace else because this feels namespaced into some snapshot sub-category.

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -59,30 +59,37 @@ pub fn init_sub_services(
...
             utxo_validation_default: config.utxo_validation,
+            native_executor_version: config.native_executor_version,
         },
     );
 
-    let verifier =
-        VerifierAdapter::new(&config.snapshot_reader, database.on_chain().clone());
+    let verifier = VerifierAdapter::new(
+        &genesis_block,
+        config.snapshot_reader.chain_config().consensus,
```
MitchTurner, line 90: Shouldn't this be `chain_config.consensus`?

crates/services/upgradable-executor/src/config.rs
```diff
@@ -1,11 +1,14 @@
 use fuel_core_executor::executor::ExecutionOptions;
+use fuel_core_types::blockchain::header::StateTransitionBytecodeVersion;
 
 #[derive(Clone, Debug, Default)]
 pub struct Config {
     /// Print execution backtraces if transaction execution reverts.
     pub backtrace: bool,
     /// Default mode for utxo_validation
     pub utxo_validation_default: bool,
+    /// The version of the native executor to determine usage of native vs WASM executor.
```
MitchTurner, line 10: > to determine usage of native vs WASM executor

Meaning that if this is `None`, it will use Wasm? How does it determine which one to use?

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -133,6 +134,10 @@ impl<S, R> Executor<S, R> {
         }
     }
 
+    pub fn native_executor_version(&self) -> StateTransitionBytecodeVersion {
+        self.config.native_executor_version.unwrap_or(Self::VERSION)
```
MitchTurner, line 138: Does something break if we update the native code but we don't increment the version?

tests/tests/blocks.rs
```diff
@@ -77,24 +77,25 @@ async fn block() {
 
 #[tokio::test]
 async fn get_genesis_block() {
```
MitchTurner, line 79: I'm not really sure what the SUT is of this test. Can you clarify what we're testing here in the test name? I don't understand the significance of `14` in `block_by_height` for example.

tests/tests/tx/tx_pointer.rs
```diff
@@ -163,7 +163,7 @@ async fn tx_pointer_set_from_previous_block() {
     let ret_tx2 = ret_tx2.as_script().unwrap();
 
     // verify coin tx_pointer is correctly set
-    let expected_tx_pointer = TxPointer::new((block_height + 1u32).into(), 0);
+    let expected_tx_pointer = TxPointer::new((previous_block_height + 2u32).into(), 0);
```
MitchTurner, line 166: Why `+ 2` now?

crates/chain-config/Cargo.toml
```diff
@@ -13,6 +13,7 @@ description = "Fuel Chain config types"
 [dependencies]
 anyhow = { workspace = true }
 bech32 = { version = "0.9.0", default-features = false, optional = true }
+derivative = { workspace = true }
```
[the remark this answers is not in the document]
xgreenx: I agree. I have the same feeling, but we are already using this dependency in the `fuel-tx` everywhere. [An example](https://github.com/FuelLabs/fuel-vm/blob/master/fuel-tx/src/contract.rs#L43).

crates/fuel-core/src/service/genesis.rs
xgreenx: Added tracing into the `execute_genesis_block` function since it is the place where the choice matter

crates/services/upgradable-executor/src/executor.rs
xgreenx: No in terms of execution, but we will use WASM executor instead of native which is slower=)

CHANGELOG.md
```diff
@@ -30,6 +30,32 @@ Description of the upcoming release here.
...
+      with a:
+      
+      ```rust
+               let mut json = String::new();
+               std::fs::File::open(&path)
+                   .with_context(|| format!("Could not open snapshot file: {path:?}"))?
+                   .read_to_string(&mut json)?;
+               let mut snapshot: Self = serde_json::from_str(json.as_str())?;
+      ```
+      because it is faster in 100 times for big JSON files.
```
Dentosal, line 56:
```suggestion
      because it is 100 times faster for big JSON files.
```

CHANGELOG.md
```diff
@@ -30,6 +30,32 @@ Description of the upcoming release here.
...
+    - Updated all tests to use `Config::local_node_*` instead of working with the `SnapshotReader` directly. It is the preparation of the tests for the futures bumps of the `Executor::VERSION`. When we increase the version, all tests continue to use `GenesisBlock.state_transition_btyecode = 0` while the version is different, which forces the usage of the WASM executor, while for tests, we still prefer to test native execution. The `Config::local_node_*` handles it and forces the executor to use the native version.
```
Dentosal, line 57:
```suggestion
    - Updated all tests to use `Config::local_node_*` instead of working with the `SnapshotReader` directly. It is the preparation of the tests for the futures bumps of the `Executor::VERSION`. When we increase the version, all tests continue to use `GenesisBlock.state_transition_bytecode = 0` while the version is different, which forces the usage of the WASM executor, while for tests, we still prefer to test native execution. The `Config::local_node_*` handles it and forces the executor to use the native version.
```

CHANGELOG.md
```diff
@@ -30,6 +30,32 @@ Description of the upcoming release here.
...
+    - Reworked the `build.rs` file of the upgradable executor. The script now has a cache used between builds to avoid recompilation of the WASM bytecode. Also, fixed the issue with outdated WASM bytecode. The script reacts on any modifications of the `fuel-core-wasm-executor` and forces recompilation(it is why we need cache), so WASM bytecode always is actual now.
```
Dentosal, line 58:
```suggestion
    - Reworked the `build.rs` file of the upgradable executor. The script now caches WASM bytecode to avoid recompilation. Also, fixed the issue with outdated WASM bytecode. The script reacts on any modifications of the `fuel-core-wasm-executor` and forces recompilation (it is why we need the cache), so WASM bytecode always is actual now.
```

bin/fuel-core/chainspec/testnet/state_transition_bytecode.wasm
Dentosal, line 1: Do we actually want to commit wasm files?

crates/chain-config/Cargo.toml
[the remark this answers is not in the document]
Dentosal: For vetting I've heard good things about https://github.com/crev-dev/cargo-crev

crates/services/upgradable-executor/src/executor.rs
MitchTurner: I'll word my question in a different way:
As a dev working on the executor, will I ever be prompted to bump the hard-coded `Self::VERSION`?

The block could be running different versions of the native code, but the version number is the same so it doesn't use Wasm since the version number matches what is on the block. 

Or is there something preventing committing changes to the executor without bumping the version?

xgreenx: https://github.com/FuelLabs/fuel-core/issues/1835

crates/chain-config/Cargo.toml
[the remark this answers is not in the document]
MitchTurner: https://github.com/FuelLabs/fuel-core/issues/1836

MitchTurner (approved):

Dentosal (approved):

PR #1829, opened 2024-04-16
Remove SpentMessages table

### LINKED ISSUE #1790: Add migration for remaining off chain tables

Dentosal:
Work towards #1790

Depends on #1811

Removes `SpentMessages` table.

## Checklist

### Before requesting review

### Before merging
- [ ] Retarget to `master`

### DIFF
--- a/crates/fuel-core/src/service/genesis/importer/on_chain.rs
+++ b/crates/fuel-core/src/service/genesis/importer/on_chain.rs
@@ -18,6 +18,7 @@ use fuel_core_storage::{
         ContractsRawCode,
         ContractsState,
         Messages,
+        ProcessedTransactions,
         Transactions,
     },
     transactional::StorageTransaction,
@@ -143,6 +144,8 @@ impl ImportTable for Handler<Transactions> {
         for transaction in &group {
             tx.storage::<Transactions>()
                 .insert(&transaction.key, &transaction.value)?;
+            tx.storage::<ProcessedTransactions>()
+                .insert(&transaction.key, &())?;
         }
         Ok(())
     }
--- a/crates/fuel-core/src/service/genesis/importer/on_chain.rs
+++ b/crates/fuel-core/src/service/genesis/importer/on_chain.rs
@@ -144,8 +144,6 @@ impl ImportTable for Handler<Transactions> {
         for transaction in &group {
             tx.storage::<Transactions>()
                 .insert(&transaction.key, &transaction.value)?;
-            tx.storage::<ProcessedTransactions>()
-                .insert(&transaction.key, &())?;
         }
         Ok(())
     }
--- a/crates/types/src/entities/relayer/message.rs
+++ b/crates/types/src/entities/relayer/message.rs
@@ -273,40 +273,9 @@ impl MessageProof {
 }
 
 /// Represents the status of a message
-pub struct MessageStatus {
-    /// The message state
-    pub state: MessageState,
-}
-
-impl MessageStatus {
-    /// Constructor for `MessageStatus` that fills with `Unspent` state
-    pub fn unspent() -> Self {
-        Self {
-            state: MessageState::Unspent,
-        }
-    }
-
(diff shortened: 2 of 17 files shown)


Voxelot:
are we removing spent messages due to the extra hassle of regenesis migration? We had that table to provide extra redundancy protection against double spending messages in case of any relayer hiccups.

xgreenx:
> are we removing spent messages due to the extra hassle of regenesis migration? We had that table to provide extra redundancy protection against double spending messages in case of any relayer hiccups. Are we just going to rely on the relayer offchain tables to track included events now?

We added the `SpentMessages` table for extra redundancy protection against double spending messages because we had two places that can be written into the `Messages` table. In the past, the relayer inserted new messages into the `Messages` table, and the executor removed messages from the `Messages` table.

However, with a new flow, the relayer writes all events into the `EventsHistory` table, and the executor writes and removes messages from the `Messages` table while increasing the da height. 

The new flow provides the same protection by default because relayer hiccups/pruning no longer affect the `Messages` table.

crates/types/src/entities/relayer/message.rs
```diff
@@ -273,40 +273,9 @@ impl MessageProof {
...
-        Self {
-            state: MessageState::NotFound,
-        }
-    }
-}
-
-/// The possible states a Message can be in
-pub enum MessageState {
-    /// Message is still unspent
+pub enum MessageStatus {
```
xgreenx, line 276: Ohh, I forgot that it affects the GraphQL API endpoint. Hmm, we need to find a way to still support this endpoint because the bridge relies on it.

Maybe we need to move this table to off-chain database instead.

crates/fuel-core/src/service/genesis/importer/on_chain.rs
```diff
@@ -143,6 +144,8 @@ impl ImportTable for Handler<Transactions> {
         for transaction in &group {
             tx.storage::<Transactions>()
                 .insert(&transaction.key, &transaction.value)?;
+            tx.storage::<ProcessedTransactions>()
+                .insert(&transaction.key, &())?;
```
xgreenx, line 148: Hmm, this migration even shouldn't exist because we have `OldTransactions` for that. I think we forgot to remove it during https://github.com/FuelLabs/fuel-core/pull/1811.

Hmm, it seems we need to migrate the `ProcessedTransactions` table instead of fetching it from `OldTransactions` because we can't rely on off-chain data during migration. Off-chain data migration is optional and only for nodes that want to support GraphQL API

Dentosal: #1811 is not merged yet, so we can still include that there. I'm going to remove this for now, we can do that in a separate PR if not in #1811. Removed in 435d1c66c.

crates/types/src/entities/relayer/message.rs
Dentosal: Hmmm in that case I'm probably just going to redo the whole PR.

PR #1830, opened 2024-04-16
Use versioning enum for WASM executor input and output

xgreenx:
Live improvement for later upgrades. The usage of the enum makes versioning flexible, and we can update only the enum instead of adding new host functions with new types of input.

### Before requesting review


crates/services/upgradable-executor/src/executor.rs
```diff
@@ -386,7 +386,11 @@ where
             .add_relayer(relayer)?
             .add_input_data(block, options)?;
 
-        instance.run(module)
+        let output = instance.run(module)?;
+
+        match output {
+            fuel_core_wasm_executor::utils::ReturnType::V1(result) => result,
+        }
```
bvrooman, line 393: Given that the method signature defines the return type as `ExecutorResult<..>`, and not a generic/`dyn`, it appears that you will aways want the version variant of `output` to wrap  `ExecutorResult` or implement `Into<ExecutorResult>`. 

Assuming the signature is correct, then you can probably just move the `match` into a function (e.g., `fn result(..)`) on the implementation of `ResultType`, and call `output.result().into()` here at the call site.

But if the return type in the method signature may change in the future, I think it would make sense to reflect that change in this PR, to communicate that the type depends on the `ReturnType` version.

bvrooman (approved):

bvrooman (approved):
Overall, LGTM. One small suggestion regarding the new enum interface.

Issue #1828, opened 2024-04-15
Support graceful shutdown during the exporting of the snapshot


xgreenx:
https://github.com/FuelLabs/fuel-core/pull/1821#discussion_r1562395711

PR #1832, opened 2024-04-17
feat: snapshot generation graceful shutdown

segfault-magnet:
Closes: #1828 

Snapshot exporting can now be cancelled. Progress updates provided as well.

Decided not to do any cleanup upon cancelling the snapshot generation. If you feel otherwise please write. We could do some of the following:
* write to `/tmp` (or whatever the temp dir is) and move only when the snapshot is finished (considerations what if the move isn't cheap due to all sorts of reasons). 
* write to a 'hidden' `.some_temp_dir` in the same parent directory as the user provided path. (considerations: better than the above in terms of potentially expensive rename/move operations on the tmp dir)
* just remove the snapshot dir if the process is cancelled (considerations: what if the user gives the home directory as the snapshot dir, do we delete his $HOME :) , would need to be coupled with a check beforehand that the dir doesn't exist, fail if it does. 

So I just decided not to do anything rn wrt.


CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ Description of the upcoming release here.
 
 ### Added
 
+- [#1832](https://github.com/FuelLabs/fuel-core/pull/1832): Snapshot generation can be cancelled. Progress is also reported.
```
Voxelot, line 33: this will need to be moved to unreleased

crates/fuel-core/src/service/genesis/progress.rs
```diff
@@ -0,0 +1,133 @@
...
+        if let Target::Cli(message) = &target {
+            bar.set_message(message.clone());
+
+            bar.set_style(Self::style(max.is_some()));
+        }
+
+        ProgressReporter { bar, target }
+    }
+
+    fn style(lenght_known: bool) -> ProgressStyle {
```
Voxelot, line 52:
```suggestion
    fn style(length_known: bool) -> ProgressStyle {
```

crates/fuel-core/src/service/genesis/progress.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+impl Default for ProgressReporter {
+    fn default() -> Self {
+        Self::new(Target::Logs(tracing::info_span!("default")), None)
+    }
+}
+
+#[derive(Clone)]
+pub enum Target {
```
Voxelot, line 31: maybe add some rustdocs explaining what this does?

crates/fuel-core/src/service/genesis/importer.rs
```diff
@@ -171,11 +164,7 @@ impl SnapshotImporter {
                 progress_reporter,
             );
             async move {
-                if num_groups >= GROUPS_NUMBER_FOR_PARALLELIZATION {
-                    tokio_rayon::spawn(move || task.run()).await?;
-                } else {
-                    task.run()?;
-                }
+                tokio_rayon::spawn(move || task.run()).await?;
```
Voxelot, line 167: is this change related to cancellation? are there risks to creating too many spawned rayon tasks here?

segfault-magnet: > is this change related to cancellation?

No. This got introduced at some point (don't know when), I just caught it now.

Callbacks given to `TaskManager::spawn` must be light. A [rule of thumb](https://ryhl.io/blog/async-what-is-blocking/) given by a tokio developer is between 10 and 100 us between awaits.

Allowing any number of batches to execute directly in the callback is problematic since batch size can be arbitrary. The import logic is not async and will not yield until all batches are done.

> are there risks to creating too many spawned rayon tasks here?

It's suboptimal. The rayon thread pool thread number is fixed by default, also low, read somewhere it is 1x core count.

Our workload is a combination of blocking IO (rocksdb) and CPU intensive work (decompression, decoding, merkle calculations etc) probably best fitted to spawning a thread per job since we have a limited amount of tables (jobs) to migrate.

So exhausting the thread pool might provide suboptimal CPU utilization when threads are stuck on blocking IO. Also will introduce a sort of semaphore where whole tables will not begin the import process due to a lack of available threads.

Rayon isn't fit for long running IO-blocking tasks. Should be used for processing batches of CPU intensive work (has support for work-stealing).

 I'll use `tokio::spawn_blocking` for now since the backing thread pool seems to expand based on demand. I'll go back to rayon in a followup PR but will actually sub-divide the work and send batches off to be processed as rayon tasks.

crates/fuel-core/src/service/genesis/task_manager.rs
```diff
@@ -1,74 +1,126 @@
...
+    }
+}
+
+#[async_trait::async_trait]
+impl NotifyCancel for MultiCancellationToken {
+    async fn wait_until_cancelled(&self) -> anyhow::Result<()> {
+        if self.sources.is_empty() {
+            futures::future::pending::<()>().await;
+            Ok(())
+        } else {
```
xgreenx, line 80: It is not clear to me why we want to return to `pending`. More logical looks to return `Ok(())`.

Plus, it seems impossible to have zero sources because you have one constructor that initializes at least one source.

crates/fuel-core/src/service/genesis/task_manager.rs
```diff
@@ -1,74 +1,126 @@
...
+            FuturesUnordered::from_iter(
+                self.sources.iter().map(|s| s.wait_until_cancelled()),
+            )
+            .next()
+            .await
+            .expect("At least one source should be available")
```
xgreenx, line 86: You already return a result, so it is okay to return an error instead of panic.

crates/fuel-core/src/service/genesis/task_manager.rs
```diff
@@ -1,74 +1,126 @@
...
+
+    /// Note: Adding a new source to the token will not affect already running futures.
+    pub fn insert(&mut self, source: impl NotifyCancel + Send + Sync + 'static) {
+        self.sources.push(Arc::new(source));
+    }
+}
+
+#[async_trait::async_trait]
+impl NotifyCancel for MultiCancellationToken {
+    async fn wait_until_cancelled(&self) -> anyhow::Result<()> {
```
xgreenx, line 76: It is not clear what is the purpose of implementing `NotifyChannel` for the `MultiCancellationToken`.

It seems you don't use `MultiCancellationToken::wait_until_cancelled` anywhere except tests=) You only use `is_cancelled` where you select between `StateWatcher::is_cancelled || CancellationToken:: is_cancelled`. Maybe it makes more sense to just create a simple type that has two fields=)

image

segfault-magnet: The thought process:
* `TaskManager` needs to be given something to cancel it. In the case of import, cancellation is signaled via the `StateWatcher` being in a `Stopping` or `Stopped` state. For exports you don't have the `StateWatcher`, it felt unnatural to create one force it to be in `Started` and later on have it be in `Stopping` upon receiving a signal. Since we don't have a service for exporting a cancellation token fit best.
* the task manager also has an internal reason to stop -- upon encountering any error. So it needs to register its own way of signaling to stop. Currently that is a cancellation token.
* The tasks need to be given something that signals when either when something on the outside said stop (a `StateWatcher` or a outside `CancellationToken`) or whether the `TaskManager` itself ordered the stop (due to an error). So that means either combine those two (the outside signal and the internal one) into one struct or pass two arguments to each task, which is unsightly and would beg the question "why two sources of cancellation?"

So whatever I replace the `MultiCancellationToken` with needs to have two items:
a) an outside signal (a `StateWatcher` or a `CancellationToken`)
b) an internal signal (a `CancellationToken`)

I couldn't find a name for such a structure and opted for generalizing the problem into a listen to n sources and report if any of them signal cancelled.

crates/fuel-core/src/service/genesis/task_manager.rs
segfault-magnet: Why return an error when I've previously checked that the collection is not empty? I.e. like doing:
```rust
if !vec.is_empty() {
if let Some(last) = vec.last() {
/// ...
}
else {
bail!("A bug in the std");
}
}
```
instead of 
```rust
if !vec.is_empty() {
   vec.last().expect("Checked not empty");
}
```

crates/fuel-core/src/service/genesis/task_manager.rs
```diff
@@ -1,74 +1,126 @@
...
+        self.is_cancelled()
+    }
+}
+
+#[async_trait::async_trait]
+impl NotifyCancel for StateWatcher {
+    async fn wait_until_cancelled(&self) -> anyhow::Result<()> {
+        while !self.is_cancelled() {
+            self.clone().changed().await?;
         }
```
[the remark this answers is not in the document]
segfault-magnet: Like this:
```rust
    async fn wait_until_cancelled(&self) -> anyhow::Result<()> {
        let mut state = self.clone();
        while !state.is_cancelled() {
            state.changed().await?;
        }

        Ok(())
    }

```
?

crates/fuel-core/src/service/genesis/exporter.rs
```diff
@@ -174,13 +181,20 @@ where
 
         let db = db_picker(self).clone();
         let prefix = prefix.map(|p| p.to_vec());
+        let progress_tracker =
+            self.multi_progress.table_reporter(None, T::column().name());
```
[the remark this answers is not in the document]
segfault-magnet: I tried getting the info but it isn't straightforward, gave up at some point. It seems rocksdb cannot give an accurate number (due to some possibly being marked for deletion but not yet pruned, or whatever). 

I read [here](https://github.com/aayushKumarJarvis/rocks-wiki/blob/master/RocksDB-FAQ.md)
that a query can be made:

> Q: How to know the number of keys stored in a RocksDB database?
> 
> A: Use GetIntProperty(cf_handle, “rocksdb.estimate-num-keys") to obtain an estimated number of keys stored in a column family, or use GetAggregatedIntProperty(“rocksdb.estimate-num-keys", &num_keys) to obtain an estimated number of keys stored in the whole RocksDB database.

But didn't feel like propagating that up through the abstractions. It [seems](https://docs.rs/rocksdb/latest/rocksdb/properties/constant.ESTIMATE_NUM_KEYS.html) we can query for properties.

There is no size hint either for the returned iterator so we cannot even approximately know how many there are.

Wdyt, should we propagate this through the abstractions in this PR?

crates/fuel-core/src/service/genesis/exporter.rs
```diff
@@ -174,13 +181,20 @@ where
...
         self.task_manager.spawn(move |cancel| {
             tokio_rayon::spawn(move || {
                 db.entries::<T>(prefix, IterDirection::Forward)
                     .chunks(group_size)
                     .into_iter()
                     .take_while(|_| !cancel.is_cancelled())
-                    .try_for_each(|chunk| writer.write(chunk.try_collect()?))?;
+                    .enumerate()
+                    .try_for_each(|(index, chunk)| {
+                        progress_tracker.set_progress(index);
```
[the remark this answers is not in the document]
segfault-magnet: But it might not advance the progress. `set_progress` sets the progress to whatever you give it, even if you give it a lower value than previously. Propagates to `set_position` on the inner bar.

Simpler than keeping a counter inside of the progress, especially since we always seem to know the progress outside -- i.e. we have to keep track of the group number due to the need to update the cursors.

I'll add some docs to make that clearer.

crates/fuel-core/src/service/genesis/importer.rs
```diff
@@ -197,19 +186,15 @@ impl SnapshotImporter {
         let groups = self.snapshot_reader.read::<TableInSnapshot>()?;
         let num_groups = groups.len();
 
-        if num_groups == 0 {
```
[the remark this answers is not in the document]
segfault-magnet: It seemed like a premature optimization. Genesis can last for hours but a useless task of a few millis cannot be tolerated. 

Now I understand that there are tests that run this a lot and you'd like to optimize for that. I'll add a comment so that this isn't removed again in the future.

crates/fuel-core/src/service/genesis/importer.rs
```diff
@@ -219,47 +204,11 @@ impl SnapshotImporter {
...
             );
-            async move {
-                if num_groups >= GROUPS_NUMBER_FOR_PARALLELIZATION {
-                    tokio_rayon::spawn(move || task.run()).await?;
-                } else {
-                    task.run()?;
-                }
-                Ok(())
-            }
+            tokio_rayon::spawn(move || task.run())
```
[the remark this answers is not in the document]
segfault-magnet: I've changed this to tokio's spawn blocking to handle the exhaustion issue discussed in another comment. I'll track the  CI and see if it now has issues.

crates/fuel-core/src/service/genesis/importer.rs
segfault-magnet: I've changed it to use the tokio thread pool for blocking tasks, i.e. `join_set.spawn_blocking`. There are a lot more threads in that pool by default. Will see if that solves the issue. 

I'd like to avoid to do any amount of batches inline since we may prevent the executor from driving other futures. But if spawn_blocking is still slow (i.e. flaky p2p tests in the CI, I'll return the job inlining).

crates/fuel-core/src/service/genesis/task_manager.rs
xgreenx: `CancelationToken` already is good enough=) It is just an internal type for the `TaskManager`

segfault-magnet: I've made it more specific. Lemme know your thoughts.

crates/fuel-core/src/service/genesis/exporter.rs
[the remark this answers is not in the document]
xgreenx: It is better to make it as a part of a separate PR. We just can leave a todo for now and create a new issue with super low priority

crates/fuel-core/src/service/genesis/importer.rs
```diff
@@ -1,15 +1,18 @@
...
-use super::task_manager::TaskManager;
+
+use super::{
+    progress::MultipleProgressReporter,
+    task_manager::TaskManager,
+};
+mod import_task;
+mod off_chain;
+mod on_chain;
+use std::marker::PhantomData;
```
xgreenx, line 13: Could you return the modules to the old place, please?=) Usually we have imports first, and models after

crates/fuel-core/src/service/genesis/task_manager.rs
xgreenx: You don't need to implement `NotifyCancel` for `CancellationToken` and you don't need to test how it works=)

segfault-magnet: I need a unified interface because `CancellationToken` is being used by the tasks spawned on `TaskManager`. Either I implement the trait or I just copy the trait implementation into an `impl CancellationToken` block. 

If `NotifyCancel` is a thing, I'd expect it to be implemented on `task_manager::CancellationToken`.

If it has a behavior such that it will signal cancel if either the internal signal or the internal one is cancelled then it needs tests to validate this behavior.

crates/fuel-core/src/service/genesis/exporter.rs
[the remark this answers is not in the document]
xgreenx: It is up to you, but for me, keeping the counter inside and advancing it looks more natural. In this case, you guarantee that the progress tracker can't emit the same progress twice. Plus the type is self-contained in this case.

crates/fuel-core/src/service/genesis/task_manager.rs
xgreenx: https://github.com/FuelLabs/fuel-core/pull/1832/commits/2f023820df86ef7ee207ed950622675425aca78e

xgreenx (approved):

PR #1833, opened 2024-04-17
Regenesis of `SpentMessages` and `ProcessedTransactions`

xgreenx:
Re-implementation of the https://github.com/FuelLabs/fuel-core/pull/1829
Work towards https://github.com/FuelLabs/fuel-core/issues/1790

The change migrates the `ProcessedTransactions` table as part of the on-chain snapshot. Since it is used to validate blocks, the root of all processed transactions must be part of the `Genesis` structure along with messages, coins, and contracts.

The changes also moved `SpentMessages` from the on-chain database to the off-chain database since now it is only used for the GraphQL API.

### Before requesting review


Dentosal (dismissed):
Looks good otherwise, but maybe we should have some test verifying the behavior?

Dentosal (dismissed):

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -459,26 +459,37 @@ where
...
-                        self.last_block_created = last_block_created;
-                    }
+                _ = state.changed() => {
+                    break;
+                }
+                _ = self.tx_status_update_stream.next() => {
+                    // ignore txpool events while syncing
+                }
+                _ = self.timer.wait() => {
+                    // ignore timer events while syncing
```
xgreenx, line 478: During debugging tests, I realized that for `Never` mode, we will always push events but not process them, causing a huge buffer. Just add ignoring of them here to remove the from the buffer.

bin/fuel-core/chainspec/testnet/state_transition_bytecode.wasm
Dentosal, line 1: Why are these removed?

Dentosal (approved):

bin/fuel-core/chainspec/testnet/state_transition_bytecode.wasm
xgreenx: Only removed 895 bytes=D  The file still exists
image

Issue #1751, opened 2024-03-12
Refactor `execute` functions to decouple `Validation`, `Production`, and `DryRun`


MitchTurner:
Currently, our code uses the same functions for executing a block for validation and for production. To specify which you want, you wrap the block in the `ExecutionTypes` enum. The problem is each has different requirements, bot for inputs as well as implementation. 

We want to split the execution functions up in to `validate` and `produce` methods to maintain separation of concerns, and single responsibility. It will clean up our code to remove all the matches and if/thens that are require to handle all the cases.

### Move block building logic from `Executor` to `Producer` 
Currently the logic for building the block isn't actually on the `Producer` but on the `execute` method for `Production`. This makes it difficult to test the behavior of the `Producer` without the `Executor` impl. We should move that logic to the `Producer` and only use the executor to check individual txs in the context provided by the producer.

Should implement missing tests for `Producer`

PR #1837, opened 2024-04-17
Refactor/separate validation from other executions

MitchTurner:
Part of: https://github.com/FuelLabs/fuel-core/issues/1751

`Executor` is use both for producing a block, executing dry-run txs, and validating blocks. This is done using the same functions and we have internal conditional logic for each use case. This makes the code more complicated than it needs to be and makes it hard to read.

This PR removes the `Validation` variant from `ExecutionTypes` and creates new methods for just validation. Since there is only a set number of places in our code  that uses this variant, it cleans the code up a lot to just have it be separate method, thus removing the need for all the internal conditional logic for each variant.

Remaining work to do in subsequent PRs:
- Remove block production code from validation branch (Just use Block everywhere instead of Partial Block and Components, etc)
- Separate `DryRun` from `Production` and remove `ExecutionType`
- Hopefully remove `ExecutionKind` as well. It goes deep so I still need to figure out how to untangle it.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/executor/src/executor.rs
```diff
@@ -624,7 +659,144 @@ where
...
+                    &mut thread_block_transaction,
+                    &mut data,
+                    transaction,
+                    gas_price,
+                    coinbase_contract_id,
+                    execution_kind,
+                )?;
+            }
+
+            let new_remaining_gas_limit = block_gas_limit.saturating_sub(data.used_gas);
```
xgreenx, line 785: Hmm, instead of `saturating_sub` it should be `checked_add`. Because maybe bloc producer included more transactions than was allowed=)

Also, maybe we need to update `execute_inner` as well

crates/types/src/services/executor.rs
```diff
@@ -336,7 +311,7 @@ impl ExecutionKind {
         match self {
             ExecutionKind::DryRun => ExecutionTypes::DryRun(t),
             ExecutionKind::Production => ExecutionTypes::Production(t),
-            ExecutionKind::Validation => ExecutionTypes::Validation(t),
+            ExecutionKind::Validation => todo!("remove this variant"),
```
Dentosal, line 314: If this `todo!` still supposed to be here? Or is this part of the follow-up PR to remove `ExecutionKind`?

crates/fuel-core/src/service/adapters/executor.rs
```diff
@@ -55,6 +58,13 @@ impl ExecutorAdapter {
     ) -> ExecutorResult<Vec<TransactionExecutionStatus>> {
         self.executor.dry_run(block, utxo_validation)
     }
+
+    pub(crate) fn _validate_block(
```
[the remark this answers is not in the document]
MitchTurner: Can we just get rid of these "private" methods? They are all just used in one/two places and think I like the idea of just calling `executor.validate`, etc in those places.

crates/services/executor/src/executor.rs
```diff
@@ -624,7 +659,144 @@ where
...
+        &self,
+        block: &mut PartialFuelBlock,
+        thread_block_transaction: &mut StorageTransaction<
+            &StorageTransaction<&StorageTransaction<D>>,
+        >,
+        execution_data: &mut ExecutionData,
+        tx: MaybeCheckedTransaction,
+        gas_price: Word,
+        coinbase_contract_id: ContractId,
+        execution_kind: ExecutionKind,
```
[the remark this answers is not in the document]
MitchTurner: Nice. That seems like it might be the right approach to get rid of `ExecutionKind`.

crates/services/executor/src/executor.rs
```diff
@@ -624,7 +659,144 @@ where
...
+                        failure: err.to_string(),
+                    };
+                    data.events.push(event);
+                }
+            }
+        }
+
+        // L2 originated transactions should be in the `TxSource`. This will be triggered after
+        // all relayed transactions are processed.
+        let mut regular_tx_iter = source.next(block_gas_limit).into_iter().peekable();
```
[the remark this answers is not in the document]
MitchTurner: Done in the next PR:
https://github.com/FuelLabs/fuel-core/pull/1847

crates/services/executor/src/executor.rs
MitchTurner: In the follow-up PR, I don't event have this variable:
https://github.com/FuelLabs/fuel-core/pull/1847

I just expect the execution to fail if it exceeds the limit.

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ Description of the upcoming release here.
 - [#1844](https://github.com/FuelLabs/fuel-core/pull/1844): Fixed the publishing of the `fuel-core 0.25.1` release.
 - [1842](https://github.com/FuelLabs/fuel-core/pull/1842): Ignore RUSTSEC-2024-0336: `rustls::ConnectionCommon::complete_io` could fall into an infinite loop based on network
 
+### Changed
+
+- [#1837](https://github.com/FuelLabs/fuel-core/pull/1837): Refactor the executor and separate validation from the other use cases
```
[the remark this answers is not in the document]
xgreenx: Still not removed=)

crates/services/executor/src/executor.rs
```diff
@@ -624,7 +659,144 @@ where
         }
 
         data.changes = self.block_st_transaction.into_changes();
+        Ok(data)
+    }
+
+    // TODO: Refactor this function to reduce the number of arguments.
+    #[allow(clippy::too_many_arguments)]
+    fn execute_transaction_and_commit<
+        W: WriteTransaction + KeyValueInspect<Column = Column> + Modifiable,
```
xgreenx, line 668:
```suggestion
        W: KeyValueInspect<Column = Column> + Modifiable,
```

crates/services/executor/src/executor.rs
```diff
@@ -624,7 +659,144 @@ where
...
+        let mut data = ExecutionData::new();
+
+        let block = component.empty_block;
+        let source = component.transactions_source;
+        let gas_price = component.gas_price;
+        let coinbase_contract_id = component.coinbase_contract_id;
+        let block_height = *block.header.height();
+        let execution_kind = ExecutionKind::Validation;
+
+        let forced_transactions = if self.relayer.enabled() {
```
xgreenx, line 722: Maybe it make sense to move relayer related logic into a separate functions and reuse it in `validate_block` and `execute_block`=) It can be a part of the follow up

xgreenx (approved):
You can fix comments in the follow-up PR=)

PR #1847, opened 2024-04-19
Refactor/Simplify validation logic

MitchTurner:
Part of: https://github.com/FuelLabs/fuel-core/issues/1751

Have `validation` functions just take a `Block` rather than components. Remove `Validation` variant of `ExecutionKind`. Clean up a bunch of other little details by removing generalizations that existed before for things like tx `source`.

A lot of these changes are breaking down huge functions and trying to make it a bit more declarative. This will hopefully mitigate the need for hand-holding comments all the way down and you can just read the name of the functions to know what is going on. For example:
```rs
        // Throw a clear error if the transaction id is a duplicate
        if tx_st_transaction
            .storage::<ProcessedTransactions>()
            .contains_key(tx_id)?
        {
            return Err(ExecutorError::TransactionIdCollision(*tx_id))
        }
 ```
 became
 ```
 Self::check_tx_is_not_duplicate(tx_id, tx_st_transaction)?;
 ```
 about 20x better.
 
 I'm still kinda unsatisfied with the clarity with some spots though. For example, I'm not sure what state is supposed to change in the case of a `revert`, I'm just trusting the order things existed before and the tests. I think it could be better to have a match case for `if reverted { ... } else { ... }` instead of threading the `reverted` bool through to subsequent functions.

## Checklist

### Before requesting review

- [ ] Someone else?

### CODE REVIEW

File: crates/services/executor/src/executor.rs [hunk 1]
```diff
@@ -408,59 +413,40 @@ where
     }
 
     fn validate_inner(self, block: Block) -> ExecutorResult<UncommittedResult<Changes>> {
-        // Compute the block id before execution if there is one.
         let pre_exec_block_id = block.id();
```
MitchTurner, line 416: Can probably remove this.

crates/services/executor/src/executor.rs
```diff
@@ -1030,148 +1080,342 @@ where
...
+
+            if self.options.utxo_validation {
+                self.verify_inputs_exist(
+                    block_st_transaction,
+                    inputs.as_mut_slice(),
+                    header.da_height,
+                )?;
+            }
+
+            self.validate_inputs_state(
```
MitchTurner, line 1279: I had this missing for a while and the tests were still passing. Should we add coverage for this or is it unnecessary?

crates/services/executor/src/executor.rs
```diff
@@ -1030,228 +1080,671 @@ where
...
-    fn execute_chargeable_transaction<Tx, T>(
+    fn validate_utxos<Tx, T>(
         &self,
         mut checked_tx: Checked<Tx>,
         header: &PartialBlockHeader,
+        tx_st_transaction: &mut StorageTransaction<T>,
+    ) -> ExecutorResult<Checked<Tx>>
+    where
+        Tx: ExecutableTransaction + PartialEq + Cacheable + Send + Sync + 'static,
+        <Tx as IntoChecked>::Metadata: CheckedMetadata,
```
xgreenx, line 1703:
```suggestion
        Tx: ExecutableTransaction + Cacheable,
```

You need to clean up constraints in a lot of places, please check contains for each new function that you've added=)

crates/services/executor/src/executor.rs
```diff
@@ -1030,228 +1080,671 @@ where
...
         });
+        Ok(())
+    }
 
-        if block_st_transaction
-            .storage::<ProcessedTransactions>()
-            .insert(&coinbase_id, &())?
-            .is_some()
+    fn update_input_used_gas<Tx>(
+        checked_tx: &mut Checked<Tx>,
```
xgreenx, line 1652:
```suggestion
        checked_tx: &Checked<Tx>,
```

crates/services/executor/src/executor.rs
```diff
@@ -1030,228 +1080,671 @@ where
...
+    fn update_execution_data(
+        execution_data: &mut ExecutionData,
+        receipts: Vec<Receipt>,
+        used_gas: Word,
+        tx_fee: Word,
+        reverted: bool,
+        state: ProgramState,
+        tx_id: TxId,
+    ) -> ExecutorResult<()> {
+        execution_data.coinbase = execution_data
```
xgreenx, line 1615: It seems you can call `let (used_gas, tx_fee) = self.total_fee_paid(&tx, &receipts, gas_price)?;` inside of this function instead of passing arguments

crates/services/executor/src/executor.rs
```diff
@@ -1030,228 +1080,671 @@ where
...
+            tx_fee,
+            reverted,
+            state,
+            tx_id,
+        )?;
+
+        Ok(tx.into())
+    }
+
+    fn validate_tx_outcome<Tx: PartialEq + IntoChecked + UniqueIdentifier>(
```
xgreenx, line 1572: You can move this check to the `validate_block` function instead

crates/services/executor/src/executor.rs
```diff
@@ -1030,228 +1080,671 @@ where
...
+        <Tx as IntoChecked>::Metadata: CheckedMetadata,
+        T: KeyValueInspect<Column = Column>,
+    {
+        let tx_id = checked_tx.id();
+
+        if self.options.utxo_validation {
+            checked_tx = self.validate_utxos(checked_tx, header, tx_st_transaction)?;
+        }
+
+        let (reverted, state, mut tx, receipts) = self.attempt_tx_execution_with_vm(
```
xgreenx, line 1463: You need to `self.validate_inputs_state` before executing the transaction. User may lie to you about transaction's fields.

crates/services/executor/src/executor.rs
```diff
@@ -1030,228 +1080,671 @@ where
...
+            .insert(&coinbase_id, &())?
+            .is_some()
+        {
+            return Err(ExecutorError::TransactionIdCollision(coinbase_id))
+        }
+        Ok(tx)
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    fn execute_chargeable_transaction<Tx, T>(
```
xgreenx, line 1443: The `execute_chargeable_transaction` and `validate_chargeable_transaction` should be the same, except the `Self::validate_tx_outcome`. But as I mentioned before, you can move this check to `validate_block` level.

In will remove code duplication

crates/services/executor/src/executor.rs
```diff
@@ -1030,228 +1080,671 @@ where
...
-                inputs.as_slice(),
+
+            *mint.input_contract_mut() = input;
+            *mint.output_contract_mut() = output;
+        }
+
+        Self::store_mint_tx(mint, execution_data, coinbase_id, block_st_transaction)
+    }
+
+    fn validate_mint<T>(
```
xgreenx, line 1243: As I said in the past, validation of the block is equal to producing this block from the validator's perspective and comparing it.

It is the main idea of what we need to follow. The `execute_mint` and `validate_mint` should be the same and do the same steps.

You can compare the produced mint transaction at the end of the `validate_block` function to be sure that it was created correctly.

crates/services/executor/src/executor.rs
```diff
@@ -1030,228 +1080,671 @@ where
...
                 execution_data,
                 tx_st_transaction,
-                execution_kind,
             ),
         }
     }
 
     #[allow(clippy::too_many_arguments)]
-    fn execute_mint<T>(
+    fn validate_transaction<T>(
```
xgreenx, line 1088: I left comments in other places. Addressing them would help us to use `execute_transaction` instead of `validate_transaction` which should remove redundancy without affecting the business logic or codebase with strange `ExecutionKind` branches=)

crates/services/executor/src/executor.rs
```diff
@@ -322,11 +326,12 @@ where
             ExecutionTypes::DryRun(component) => {
```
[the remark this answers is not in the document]
MitchTurner: image
I thought you meant here.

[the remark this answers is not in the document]
xgreenx: `validate_inner` it is the place where we want to keep this code=) The `execute_inner` is the place from which we want to remove it=)

crates/services/executor/src/executor.rs
```diff
@@ -1848,8 +2233,10 @@ where
                     db,
                 )?,
                 Output::Contract(contract) => {
-                    if let Some(Input::Contract(Contract { contract_id, .. })) =
-                        inputs.get(contract.input_index as usize)
+                    if let Some(Input::Contract(input::contract::Contract {
+                        contract_id,
+                        ..
+                    })) = inputs.get(contract.input_index as usize)
```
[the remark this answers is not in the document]
MitchTurner: There are places we are using `output::contract::Contract` right next to `input::contract::Contract`. It is better that we include the namespace, or I can alias the type if you prefer.

crates/services/executor/src/executor.rs
```diff
@@ -1331,89 +1792,10 @@ where
...
-            id: tx_id,
-            result: status,
-        });
-
-        Ok(final_tx)
+        Ok((reverted, state, tx, receipts))
     }
 
-    fn verify_input_state<T>(
+    fn verify_inputs_exist<T>(
```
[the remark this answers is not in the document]
MitchTurner: `verify_input_state` tells the reader nothing about what its doing. Just that it's verifying something.

I actually think the new name is good. It checks that the input exists, both in that there is an UTxO with the right id and that the values are the same. That's the best way to check that it already exists.

Would you prefer  `verify_inputs_exist_and_values_match`?

crates/services/executor/src/executor.rs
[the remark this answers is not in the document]
MitchTurner: Okay. That makes sense. I wasn't even thinking about this side.

For `validate`, I moved it to
```
Self::check_new_block_id(partial_block, block, &data)?;
```
in `validate_block`. Do you see a problem with that?

crates/services/executor/src/executor.rs
MitchTurner: There is also this section from the old code:
```
        if execution_kind == ExecutionKind::Validation {
            self.validate_inputs_state(
                checked_tx.transaction().inputs(),
                tx_id,
                tx_st_transaction,
            )?;
        }
```
Is that not needed? I'm just copying what we had before, but happy to remove unneeded stuff.

crates/services/executor/src/executor.rs
MitchTurner: As I mention above, before that was only for `Validation` before.

crates/services/executor/src/executor.rs
MitchTurner: I think it needs all of them except  the `PartialEq` on this one:
```rs
#[async_trait::async_trait]
impl<Tx> CheckPredicates for Checked<Tx>
where
    Tx: ExecutableTransaction + Send + Sync + 'static,
    <Tx as IntoChecked>::Metadata: crate::interpreter::CheckedMetadata + Send + Sync,
{
    fn check_predicates(
        mut self,
        params: &CheckPredicateParams,
    ) -> Result<Self, CheckError> {
        if !self.checks_bitmask.contains(Checks::Predicates) {
            Interpreter::<PredicateStorage, _>::check_predicates(&self, params)?;
            self.checks_bitmask.insert(Checks::Predicates);
        }
        Ok(self)
    }
```

crates/services/executor/src/executor.rs
```diff
@@ -1030,228 +1080,671 @@ where
...
+                }
+            }
         }
-        Ok(tx)
+        Ok(())
     }
 
-    #[allow(clippy::too_many_arguments)]
-    fn execute_chargeable_transaction<Tx, T>(
+    fn validate_utxos<Tx, T>(
```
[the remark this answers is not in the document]
MitchTurner: On closer inspection, it is doing more than just "validating inputs" or "validating utxos", but I was just going off of this flag:
```rs
        if self.options.utxo_validation {
            checked_tx = self.validate_utxos(checked_tx, header, tx_st_transaction)?;
        }
```

Maybe `utxo_validation` should have a different name or something?

crates/services/executor/src/executor.rs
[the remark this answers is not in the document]
xgreenx: You can use an alias `InputContract` and `OutputContract`, but I'm curios why we haven't this problem with namespaces before=D

crates/services/executor/src/executor.rs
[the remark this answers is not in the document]
xgreenx: It can live in a separate function, but I would expect a function like `compare_blocks(prodecer_block, validator_block)` or something like that. Not only comparison of the `BlockId`=)

crates/services/executor/src/executor.rs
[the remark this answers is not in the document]
xgreenx: Yeah, it works

crates/services/executor/src/executor.rs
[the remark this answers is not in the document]
xgreenx: Yeah, the current usage of the `utxo_validation` flag is broader than before; it also skips signature and predicate verification. I don't know which name suits you better.

crates/services/executor/src/executor.rs
xgreenx: Okay, this function was a bad example=D But in other places we don't need so many constrains=)

crates/services/executor/src/executor.rs
xgreenx: The comparison of the transactions already includes this check. We can still call this function after the tx was executed in the VM, just like a double check that all fields set by the block producer are correct. It applies to validation and execution.

crates/services/executor/src/executor.rs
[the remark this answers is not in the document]
MitchTurner: Lol.  There is a trait we import called `InputContract` already. I think I'll just leave this.

I've added some helper functions that return these types. We didn't have anything returning `output::contract::Contract` before, so we didn't need to distinguish.

crates/services/executor/src/executor.rs
```diff
@@ -738,66 +654,126 @@ where
...
+                &mut thread_block_transaction,
+                &mut data,
+                maybe_checked_tx,
+                gas_price,
+                coinbase_contract_id,
+            )?;
+            #[cfg(debug_assertions)]
+            if _new_transaction != transaction {
+                error!("Provided transaction does not match the transaction built by the executor");
+            }
```
MitchTurner, line 681: I'm not sure I really like this. It's kinda chatty in the tests. Is there another approach we could take? We can't have it throw an error because the tests are expecting the error that will be used in prod.

crates/services/executor/src/executor.rs
```diff
@@ -210,8 +211,9 @@ impl ExecutionData {
 /// Per-block execution options
 #[derive(serde::Serialize, serde::Deserialize, Clone, Default, Debug)]
 pub struct ExecutionOptions {
+    // TODO: This is a bad name and the real motivation for this field should be specified
     /// UTXO Validation flag, when disabled the executor skips signature and UTXO existence checks
-    pub utxo_validation: bool,
+    pub extra_tx_checks: bool,
```
MitchTurner, line 216: This should simply be named after the situation we want to use it. I'm not sure what it is yet. For now, I just named it `extra_tx_checks` which is vague, but `utxo_validation` was incorrect, so it's an improvement.

crates/services/executor/src/executor.rs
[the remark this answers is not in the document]
MitchTurner: Changed the name to `extra_tx_checks`, which is a bad name but at least it's not incorrect.

crates/services/executor/src/executor.rs
MitchTurner: I'm not sure if this addressed now that everything is moved into one function.

crates/services/executor/src/executor.rs
```diff
@@ -379,23 +378,6 @@ where
...
         let block = block.generate(&message_ids[..], event_inbox_root);
 
-        let finalized_block_id = block.id();
-
-        debug!(
-            "Block {:#x} fees: {} gas: {}",
-            pre_exec_block_id.unwrap_or(finalized_block_id),
-            coinbase,
-            used_gas
-        );
```
xgreenx, line 389: I think we still want to have debug log here=)

crates/services/executor/src/executor.rs
```diff
@@ -689,41 +615,32 @@ where
...
                 execution_data,
-                execution_kind,
                 &mut tx_st_transaction,
             )?;
             tx_st_transaction.commit()?;
             tx
         };
 
-        block.transactions.push(tx);
+        block.transactions.push(tx.clone());
```
xgreenx, line 624: It may be too expensive to clone transaction. You could return a reference to it. Like `block.transactions.as_ref().expect("We just added transaction")` instead.

crates/services/executor/src/executor.rs
```diff
@@ -689,41 +615,32 @@ where
...
         let mut data = ExecutionData::new();
 
-        let block = component.empty_block;
-        let source = component.transactions_source;
-        let gas_price = component.gas_price;
-        let coinbase_contract_id = component.coinbase_contract_id;
-        let block_height = *block.header.height();
-        let execution_kind = ExecutionKind::Validation;
+        let mut partial_block = PartialFuelBlock::from(block.clone());
+        let transactions = core::mem::take(&mut partial_block.transactions);
```
xgreenx, line 637: It would be nice if you avoided cloning of the block here=) You don't need to modify `transactions`, so you can work with the reference

crates/services/executor/src/executor.rs
xgreenx: Instead of comparing transactions here, we can do that inside of the `check_block_matches`. You can even iterate over the block's transactions there and compare transactions and have a behavior like before the change(return `InvalidTransactionOutcome`)=)

crates/services/executor/src/executor.rs
```diff
@@ -1044,302 +1037,223 @@ where
...
-            debug_assert_eq!(tx.id(&self.consensus_params.chain_id()), tx_id);
+    fn check_mint_has_expected_index(
+        checked_mint: &Checked<Mint>,
+        execution_data: &ExecutionData,
+    ) -> ExecutorResult<()> {
+        if checked_mint.transaction().tx_pointer().tx_index() != execution_data.tx_count {
+            return Err(ExecutorError::MintHasUnexpectedIndex)
         }
+        Ok(())
+    }
```
xgreenx, line 1170: I think adding 3 new methods, 3 lines each, is overkill since the `if` statement and return error type are already descriptive enough.

Could you return it back, please=)

crates/services/executor/src/executor.rs
```diff
@@ -1044,302 +1037,223 @@ where
...
-        {
-            return Err(ExecutorError::TransactionIdCollision(coinbase_id))
-        }
-        Ok(tx)
+        Self::store_mint_tx(mint, execution_data, coinbase_id, block_st_transaction)
     }
 
     #[allow(clippy::too_many_arguments)]
-    fn execute_chargeable_transaction<Tx, T>(
+    fn execute_mint_with_vm<T>(
```
xgreenx, line 1088: It seems this new function shifted `execute_chargeable_transaction` below and the change diff looks bad now. Could you swap `execute_mint_with_vm` with `execute_chargeable_transaction` please for the history sake

crates/services/executor/src/executor.rs
```diff
@@ -1044,302 +1037,223 @@ where
...
-                outputs.as_slice(),
-            )?;
-            self.compute_state_of_not_utxo_outputs(
-                outputs.as_mut_slice(),
-                inputs.as_slice(),
-                coinbase_id,
                 block_st_transaction,
+                &coinbase_id,
+                &mut mint,
+                &mut inputs,
```
xgreenx, line 1077: It would be nice if you pass `input` instead=)

crates/services/executor/src/executor.rs
```diff
@@ -1044,302 +1037,223 @@ where
...
+        &self,
+        mut checked_tx: Checked<Tx>,
+        header: &PartialBlockHeader,
+        coinbase_contract_id: ContractId,
+        gas_price: Word,
+        execution_data: &mut ExecutionData,
+        tx_st_transaction: &mut StorageTransaction<T>,
+    ) -> ExecutorResult<Transaction>
+    where
+        Tx: ExecutableTransaction + PartialEq + Cacheable + Send + Sync + 'static,
```
xgreenx, line 1237:
```suggestion
        Tx: ExecutableTransaction + Cacheable + Send + Sync + 'static,
```

crates/services/executor/src/executor.rs
```diff
@@ -1409,11 +1348,155 @@ where
...
+        let mut vm = Interpreter::with_storage(
+            vm_db,
+            InterpreterParams::new(gas_price, &self.consensus_params),
+        );
+
+        let gas_costs = self.consensus_params.gas_costs();
+        let fee_params = self.consensus_params.fee_params();
+
+        let ready_tx = checked_tx
+            .clone()
```
xgreenx, line 1465: Now(because we don't need to compare `checked_tx` with `tx`) you can create a vector of `predicate_gas_used` from inputs and then you don't need to `checked_tx.clone()`=)

crates/services/executor/src/executor.rs
```diff
@@ -1349,7 +1263,35 @@ where
             tx.outputs(),
         )?;
 
-        // We always need to update outputs with storage state after execution.
+        self.update_tx_outputs(tx_st_transaction, tx_id, &mut tx)?;
```
xgreenx, line 1266: I know that it was before like this, but it would be nice if you moved this function to the `attempt_tx_execution_with_vm`. Then we can remove `mut` part from the `tx`=)

xgreenx (commented):
Almost ideal!=)

crates/services/executor/src/executor.rs
MitchTurner: Turns out we don't need it because we are inspecting the txs after.

crates/services/executor/src/executor.rs
MitchTurner: I strongly prefer the new code. I think the reduction of noise and keeping the caller function as flat as possible is worth the cost of maintaining another function. These benefits add up quickly and I'd like to go farther tbh but I think it's time to wrap up this PR.

crates/services/executor/src/executor.rs
xgreenx: We want to have one log for the executor and one log for validator

xgreenx (commented):
Look perfect to me=) Would be nice if @Dentosal also reviewed it=)

crates/services/executor/src/executor.rs
```diff
@@ -1044,154 +1070,51 @@ where
...
-            if *mint.mint_amount() != execution_data.coinbase {
-                return Err(ExecutorError::CoinbaseAmountMismatch)
-            }
-            if *mint.gas_price() != gas_price {
-                return Err(ExecutorError::CoinbaseGasPriceMismatch)
-            }
-
-            let block_height = *header.height();
+            Self::check_mint_amount(&mint, execution_data.coinbase)?;
+            Self::check_gas_price(&mint, gas_price)?;
```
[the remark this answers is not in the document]
MitchTurner: https://github.com/FuelLabs/fuel-core/issues/1867

xgreenx (approved):

PR #1848, opened 2024-04-19
feat: Support block and header versions gql

bvrooman:
Related issues:
- Closes https://github.com/FuelLabs/fuel-core/issues/1796

This PR adds a `version` field to the `Block` and `BlockHeader` entities returned by the GraphQL API.
 
Screenshot 2024-04-19 at 9 31 42 PM

Additionally, the `Block` and `BlockHeader` client types are updated to include the `version` as well.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/client/src/client/schema.rs
```diff
@@ -328,3 +328,9 @@ impl From<TryFromSliceError> for ConversionError {
         ConversionError::BytesLength
     }
 }
+
+#[derive(cynic::QueryFragment, Clone, Debug, Eq, PartialEq)]
+#[cynic(schema_path = "./assets/schema.sdl")]
+pub struct Version {
```
bvrooman, line 334: I moved this out of `chain.rs` and into the parent level `schema.rs` for reuse in other submodules.

bvrooman:
This PR updates the relevant snapshot tests. However, it does not appear that we have tests for all resources against the GQL server-level API or client API (integration or mock GQL). We have some high level tests for specific actions and resources like TXs but it could be worth having more tests for GQL responses (as long as this responsibility lies with `fuel-core` and not an indexer).

MitchTurner (commented):
Looking good. Can you add a test to `tests/tests/blocks.rs` that checks the version?

bvrooman:
> Looking good. Can you add a test to `tests/tests/blocks.rs` that checks the version?
> 
> edit: I just saw your last comment. Yeah. I think `tests/tests/blocks.rs` is the right place for that coverage

This is a bit tricky: The version isn't actually a field on the client `Block` type; rather, the version is used to instruct the client on how to construct the client `Block` type. This means that we cannot simply test for the `version` number. Real unit tests would require having other versions of the `Block`, which would allow us to test things like:
- different constructions of `Block` (i.e. version 1, version 2, etc.) 
- unsupported version (i.e. server returns a version not supported by the client)

Because we don't have these, I created some contrived versions and tests here: https://github.com/FuelLabs/fuel-core/pull/1852/files#diff-b43da8aec184b70c3e3ac74b3f156eb2142bf651b724ccfb3d7dc9e6d7d70b01R120. This is just for reference, but should illustrate the kinds of tests we can have in the future, and why we can't really test for that now.

CHANGELOG.md
```diff
@@ -13,7 +13,11 @@ Description of the upcoming release here.
 ### Fixed
 
 - [#1844](https://github.com/FuelLabs/fuel-core/pull/1844): Fixed the publishing of the `fuel-core 0.25.1` release.
-- [1842](https://github.com/FuelLabs/fuel-core/pull/1842): Ignore RUSTSEC-2024-0336: `rustls::ConnectionCommon::complete_io` could fall into an infinite loop based on network
+- [#1842](https://github.com/FuelLabs/fuel-core/pull/1842): Ignore RUSTSEC-2024-0336: `rustls::ConnectionCommon::complete_io` could fall into an infinite loop based on network
+
+### Added
+
+- [#1848](https://github.com/FuelLabs/fuel-core/pull/1848): Added `version` field to the `Block` and `BlockHeader` GraphQL entities. Added corresponding `version` field to the `Block` and `BlockHeader` client types in `fuel-core-client`.
```
xgreenx, line 20: We need to move it to "Unreleased" section

crates/client/src/client/schema/block.rs
```diff
@@ -75,10 +76,11 @@ pub struct BlockEdge {
     pub node: Block,
 }
 
-/// Block with transactiuon ids
+/// Block with transaction ids
 #[derive(cynic::QueryFragment, Clone, Debug)]
 #[cynic(schema_path = "./assets/schema.sdl")]
 pub struct Block {
+    pub version: Version,
```
xgreenx, line 83: We have a separate enum with all supported versions for `ConsensusParameters`(and other parameters). This strategy forces us to cover all possible values during compilation time, and I think it is still good to continue using it.

image

bvrooman: Updated to use `enum` at the `async-graphql` server level and at the `cynic` client level. This gives us exhaustive variant checks.

crates/fuel-core/src/schema.rs
```diff
@@ -35,6 +36,15 @@ pub mod tx;
...
 pub mod relayed_tx;
 
+pub struct Version(u8);
+
+#[Object]
+impl Version {
+    async fn value(&self) -> scalars::U8 {
+        self.0.into()
+    }
+}
```
xgreenx, line 46: I think we can remove this one=)

xgreenx (commented):
Looks good to me=)

MitchTurner:
> This is a bit tricky: The version isn't actually a field on the client Block type; rather, the version is used to instruct the client on how to construct the client Block type. This means that we cannot simply test for the version number. Real unit tests would require having other versions of the Block, which would allow us to test things like:

Okay. That makes sense. I feel like someone mentioned creating something like
```
    #[cfg(feature = "test_helpers")]
    VTest(VTestBlock)
```
variant that we could use to show upgradability. 

Maybe I misheard. But I'm kinda impartial to something like that. On one hand it would be a decent tracer bullet to prove that we have upgradability, but it also doesn't really capture any desired behavior. Obviously we could get rid of it when we add a real version, but we could also just add the test once we add thet variant.

:shrug:

bvrooman:
> > This is a bit tricky: The version isn't actually a field on the client Block type; rather, the version is used to instruct the client on how to construct the client Block type. This means that we cannot simply test for the version number. Real unit tests would require having other versions of the Block, which would allow us to test things like:
> 
> Okay. That makes sense. I feel like someone mentioned creating something like
> 
> ```
>     #[cfg(feature = "test_helpers")]
>     VTest(VTestBlock)
> ```
> 
> variant that we could use to show upgradability.
> 
> Maybe I misheard. But I'm kinda impartial to something like that. On one hand it would be a decent tracer bullet to prove that we have upgradability, but it also doesn't really capture any desired behavior. Obviously we could get rid of it when we add a real version, but we could also just add the test once we add thet variant.
> 
> 🤷

I like the idea. I hacked together some "test" blocks in [another branch](https://github.com/FuelLabs/fuel-core/pull/1852/files#diff-b43da8aec184b70c3e3ac74b3f156eb2142bf651b724ccfb3d7dc9e6d7d70b01R120) to ensure the functionality, which is how I discovered that our previous approach (using `Union`) didn't actually work. So there's definitely value there.

Adding a test block variant will require adding the supporting test code. This means adding test code to all the match statements, places requiring destructuring of the block into parts (transactions, etc), and other areas touched. The cost of this is added maintenance and code clutter. And ultimately, this is fine - I think it may be a worthwhile cost! 

I'd like to try it locally with a cleaner version of my above mentioned branch, including using the proper compiler flags.

bvrooman:
After spending some time on adding the `Test` variant, I think there are practical reasons that make this infeasible.

For one, this kind of test requires supporting the Test variant at both the server level and client level. Whenever an object is added to the server API, it must be included in the SDL. We don't have the ability to have separate SDLs based on what is and isn't hidden behind the test compiler flag (i.e., `#[cfg(test)]`). We can simply choose to always include the `Test` variant in the SDL, but now we are leaking test code into production. We also cannot use `#[cfg(test)]` at the client level at this point, because the SDL doesn't distinguish it as test code, and the compiler complains that the variant is not covered. Again, this would require us to leak test code into production.

For two, supporting this variant introduces a lot of test code that requires maintenance. There are a few places that require us to retrieve details about the `Block`, such as the header and transactions. To support this, we need to fill out the `Test` block with almost the same data and transformations as a regular `V1` block. The maintenance cost is fairly high.

In the end, I recommend against codifying this test.

Issue #162, opened 2022-02-04
Make RocksDB more upgradeable


Voxelot:
Problem: Adding new column families to RocksDB breaks the database and requires a complete wipe.

We make extensive use of RocksDB column families to partition various data types. These column families are identified using integer based column id's configured in the [database columns constants module](https://github.com/FuelLabs/fuel-core/blob/master/fuel-core/src/database.rs#L35).

However, due to the way we open and initialize RocksDB, any new columns added to the schema breaks the database and it is no longer able to open. Adding new columns shouldn't break the entire database and require it to be scrapped.

The primary method involved is [RocksDb::open](https://github.com/FuelLabs/fuel-core/blob/master/fuel-core/src/state/rocks_db.rs#L24).

simrankedia:
@Voxelot  I have tried to raise a PR to fix this issue https://github.com/FuelLabs/fuel-core/pull/1851 - please let me know if the direction is correct - I will accordingly complete and release a new PR with tests etc.

PR #1853, opened 2024-04-22
Added a test for the case of adding new columns into the existing RocksDB database

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/162

Added a test case to verify the database's behavior when new columns are added to the RocksDB database.

### Before requesting review


bvrooman (approved):

crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -618,6 +631,29 @@ mod tests {
...
+    #[test]
+    fn open_new_columns() {
+        let tmp_dir = TempDir::new().unwrap();
+
+        // Given
+        let old_columns =
+            vec![Column::Coins, Column::Messages, Column::UploadedBytecodes];
+        let database_with_old_columns =
+            RocksDb::<OnChain>::open(tmp_dir.path(), old_columns.clone(), None)
+                .expect("Failed to open database with old columns");
```
MitchTurner, line 643: Nit: I'm okay with using `unwrap` in tests, if you'd prefer `expect`, it's more standard to explain the reason it should be `Ok`, rather than just telling the user it failed:
https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style

crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -227,36 +220,56 @@ where
...
-                        let cf_descriptors = columns.clone().into_iter().map(|i| {
-                            ColumnFamilyDescriptor::new(
-                                Self::col_name(i.id()),
-                                Self::cf_opts(i, &block_opts),
-                            )
-                        });
-                        DB::open_cf_descriptors(&opts, &path, cf_descriptors)
-                    }
-                }
+        let existing_column_families = DB::list_cf(&opts, &path).unwrap_or_else(|err| {
```
MitchTurner, line 223: This is a pretty big wall of imperative code. Could you break it down into a series of declarative helper functions?

PR #1856, opened 2024-04-23
fix: Use `Enum` for `ConsensusParametersVersion` and related types

bvrooman:
Related issues:
- Untracked

While testing the feature for sending `Block` versions over GraphQL, we discovered that the current approach to GraphQL type versioning doesn't work as expected. Using `async-graphql::Union` is poorly suited for versioning because:
- union types cannot define multiple variants with the same inner types (e.g., you cannot define a union with  `V1(Version), V2(Version), ...`)
- union types are designed to represent sum types/disjoint union types (e.g., `HeavyOperation` and `LightOperation` where the type must be one or the other)

This prevents us from defining a second version. I.e.,
```
#[derive(Union)]
pub enum ConsensusParametersVersion {
    V1(Version),
    V2(Version),
}
```
will fail to compile, preventing us from releasing new versions.

The solution is to use `async-graphql::Enum`. `Enum` writes the enum in the SDL as a real enum. This allows us to reuse types within variants or provide empty variants. This is better suited for versioning. This will allow us to define future versions easily: 

```
#[derive(Enum)]
pub enum ConsensusParametersVersion {
    V1,
    V2,
}
```

This change must be used to patch `0.24` and `0.25`. Once these versions are updated:
- We must update the image in the dev cluster
- SDK teams must update their versions to include the fix
- Sway team must be updated to use new `fuels-rs`

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/client/assets/schema.sdl
+++ b/crates/client/assets/schema.sdl
@@ -195,7 +195,9 @@ type ConsensusParametersPurpose {
 	checksum: Bytes32!
 }
 
-union ConsensusParametersVersion = Version
+enum ConsensusParametersVersion {
+	V1
+}
 
 type Contract {
 	id: ContractId!
@@ -264,7 +266,9 @@ type ContractParameters {
 	maxStorageSlots: U64!
 }
 
-union ContractParametersVersion = Version
+enum ContractParametersVersion {
+	V1
+}
 
 union DependentCost = LightOperation | HeavyOperation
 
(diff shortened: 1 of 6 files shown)


crates/client/assets/schema.sdl
```diff
@@ -1122,10 +1134,6 @@ type VariableOutput {
 	assetId: AssetId!
 }
 
-type Version {
-	value: U8!
-}
```
bvrooman, line 1127: It appears that no schema types use `U8` anymore, so the SDL compiler removed it from the final SDL definition. This means I also needed to remove `U8` from the client side definitions.

xgreenx (approved):

Issue #1545, opened 2023-12-14
Database state migration


xgreenx:
Versionable data types:

- [x] https://github.com/FuelLabs/fuel-core/issues/1544
- [x] https://github.com/FuelLabs/fuel-core/issues/1552

Migrations:
- [x] https://github.com/FuelLabs/fuel-core/issues/1570
- [x] https://github.com/FuelLabs/fuel-core/issues/1572
- [x] https://github.com/FuelLabs/fuel-core/issues/1619

https://www.notion.so/fuellabs/Upgradeability-Planning-181702b241ac4002882bfc9e361aa4eb

PR #1860, opened 2024-04-24
regenesis: migrate FuelBlockIdsToHeights

### LINKED ISSUE #1790: Add migration for remaining off chain tables

Dentosal:
Closes #1790 by completing the last missing item.
Closes https://github.com/FuelLabs/fuel-core/issues/1545 by completing the last missing item as well.

This PR regenerates `FuelBlockIdsToHeights` for both blocks on regenesis.

## Checklist

### Before requesting review


crates/storage/src/transactional.rs
```diff
@@ -190,13 +229,24 @@ where
...
     }
 }
 
 impl<Storage> Modifiable for InMemoryTransaction<Storage> {
-    fn commit_changes(&mut self, changes: Changes) -> StorageResult<()> {
+    /// Panics height policy mismatches the height policy fo the transaction.
+    fn commit_changes(
+        &mut self,
+        changes: Changes,
+        height_policy: ModifyHeightPolicy,
```
xgreenx, line 244: Internal details of how the database works should not affect the interfaces.

xgreenx (commented):
I think we need to introduce a `Genesis` phase into the database that allows violation of some rules. Or even create a new `GenesisDatabase` type that doesn't have guarantees/strict rules on the `commit_chagnes` method.

crates/storage/src/transactional.rs
Dentosal: I rewrote the PR to use a flag in the Database itself. The solution seems quite clean. Let me know what you think about this one.

crates/services/src/service.rs
```diff
@@ -16,27 +18,6 @@ use std::any::Any;
 use tokio::sync::watch;
 use tracing::Instrument;
 
-/// Alias for `Arc<T>`
```
Dentosal, line 19: note: moved these to a separate module for clarity

crates/fuel-core/src/database.rs
```diff
@@ -90,7 +90,10 @@ pub struct Database<Description = OnChain>
 where
     Description: DatabaseDescription,
 {
+    /// Cached value from Metadata table, used to speed up lookups.
```
Dentosal, line 93: See https://github.com/FuelLabs/fuel-core/pull/1694/commits/7b1141ac838568e3590f09dd420cb24a6946bd32#r1504911777 for the explanation behind this comment

crates/fuel-core/src/state.rs
```diff
@@ -33,11 +33,7 @@ where
 
 pub trait TransactableStorage<Height>: IterableStore + Debug + Send + Sync {
     /// Commits the changes into the storage.
-    fn commit_changes(
-        &self,
-        height: Option<Height>,
-        changes: Changes,
-    ) -> StorageResult<()>;
+    fn commit_changes(&self, changes: Changes) -> StorageResult<()>;
```
xgreenx, line 36: We need `height: Option<Height>` for the state rewind feature=) I'm working on it right now, so I need this argument=D

I think maybe it makes sense to treat block height `None` as a regenesis and, at the end of the regenesis, set the block heights with information from the genesis block.

Dentosal: It would have been nice to leave a comment indicating that the useless argument was there for a reason.

Dentosal: Restored in 451f8be5d

xgreenx (approved):

PR #1866, opened 2024-04-25
test: restart with relayer data

bvrooman:
The relayer service fails to start if the relayer database is already populated with a `DaHeightTable` entry. When the relayer starts, it attempts to set the DA height to the config's `da_deploy_height`. If the `da_deploy_height` defined by the config (where the default is 0) is not a monotonically increasing value compared to the database entry, the service panics. 

Instead of setting we DA height at relayer start up, we can simply continue from where the relayer has previously left off. We do not need to make this invalid database commit at relayer startup. This PR removes the invalid commit. 

## Checklist

### Before requesting review

- [ ] Someone else?


bin/fuel-core/src/cli/run/relayer.rs
```diff
@@ -31,10 +28,6 @@ pub struct RelayerArgs {
     #[arg(long = "relayer-v2-listening-contracts", value_parser = parse_h160, env)]
     pub eth_v2_listening_contracts: Vec<H160>,
 
-    /// Number of da block that the contract is deployed at.
-    #[clap(long = "relayer-da-deploy-height", default_value_t = Config::DEFAULT_DA_DEPLOY_HEIGHT, env)]
-    pub da_deploy_height: u64,
```
xgreenx, line 36: We still need a way to pass the block height from which we need to start requesting ethereum blocks. Otherwise we will start from zero and it could take a lot of time to catch up with the current block height.

crates/services/relayer/src/storage.rs
```diff
@@ -143,29 +147,29 @@ where
         }
 
         db_tx.storage::<EventsHistory>().insert(da_height, events)?;
-
-        grow_monotonically(&mut db_tx, da_height)?;
+        db_tx
```
xgreenx, line 150: We don't need `DaHeightTable` table anymore=)

xgreenx (approved):

PR #1868, opened 2024-04-26
Fix `BlockHeader` hash 

MitchTurner:
- Add `event_inbox_root` to hash calcuation
- Use `to_be_bytes` instead of `to_bytes` to avoid unnecessary padding of `u32` to `u64`.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/types/src/blockchain/header.rs
```diff
@@ -441,6 +441,7 @@ impl ApplicationHeader<GeneratedApplicationFields> {
         hasher.input(self.message_receipt_count.to_be_bytes());
         hasher.input(self.transactions_root.as_ref());
         hasher.input(self.message_outbox_root.as_ref());
+        hasher.input(self.event_inbox_root.as_ref());
```
MitchTurner, line 444: Is there any logic to the order these things get added to the hash?

crates/types/src/blockchain/header.rs
```diff
@@ -436,13 +436,18 @@ impl ApplicationHeader<GeneratedApplicationFields> {
     pub fn hash(&self) -> Bytes32 {
         // Order matters and is the same as the spec.
         let mut hasher = crate::fuel_crypto::Hasher::default();
-        hasher.input(self.da_height.to_bytes().as_slice());
+        hasher.input(self.da_height.to_be_bytes().as_slice());
         hasher.input(self.transactions_count.to_be_bytes());
```
MitchTurner, line 440: @xgreenx said on Slack:
>The TxPointer.index_output is u16, so it means we can’t have more transactions than u16, but we use u64. The same for the messages receipts, we can’t have more than transactions_count * 255, so it can’t be more than u32=)
If we don’t need to pad them and it is a bug in the transactions_count and message_receipt_count, then we can remove it=)

But the docs still say:
```
| `txCount`                        | `uint64`   | Number of [transaction](../tx-format/transaction.md)s in this block.                                                                                                                                                                                |
| `message_receipt_count`          | `uint64`   | Number of [output message](../abi/receipts.md#messageout-receipt)s in this block.     
```
So I kept these to match the specs.

crates/types/src/blockchain/header.rs
xgreenx: The order of fields in the header

crates/types/src/blockchain/header.rs
xgreenx: https://github.com/FuelLabs/fuel-specs/pull/578

xgreenx (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- [#1868](https://github.com/FuelLabs/fuel-core/pull/1868): Include the `event_inbox_root` in header hash.
```
MitchTurner, line 11: We also updated the types.

crates/types/src/blockchain/block.rs
```diff
@@ -88,12 +91,16 @@ impl Block<Transaction> {
         transactions: Vec<Transaction>,
         outbox_message_ids: &[MessageId],
         event_inbox_root: Bytes32,
-    ) -> Self {
```
MitchTurner, line 91: I don't think `new` should return `Result` here. Can we just pass it a valid `BlockHeader` instead of a `PartialBlockHeader`?

xgreenx: Passing the block header still requires verification that transactions match the header=)

MitchTurner: Ah. That's right. Okay, I guess it's unavoidable. 

I think `new` gives the wrong impression still. Maybe `construct`? I just think of `new` as being a dumb constructor generally, but it's not that important. I'm fine with this.

xgreenx (dismissed):

xgreenx (approved):

Issue #1807, opened 2024-04-05
Update VM initialisation gas cost benchmark


xgreenx:
After https://github.com/FuelLabs/fuel-vm/pull/697, the allocation of the stack or heap may cause the copying of many bytes. The VM initialization should charge the user for the worst possible scenario where we will resize the heap/stack 26 times to be sure that new behaviour is not exploited by the attacker.

This strategy overcharges the user but improves the network's safety. Later, we can refund gas if the user doesn't initiate re-allocations(it will be a separate feature).

PR #1870, opened 2024-04-29
Fixed benchmarks for the `0.25.3`

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1807

The PR updates benchmarks to work properly with new VM rules and fixes the issue with non-optimal `InMemoryTransaction` behavior for the `size_of_value` method.

Because of the new VM rules: the memory that we are accessing should be initialized. The `cfe` opcode in the benchmarks expands the stack to allocate the memory that will later be used by the instruction that we are benchmarking.

### Before requesting review


benches/benches/vm_set/flow.rs
```diff
@@ -70,7 +70,7 @@ pub fn run(c: &mut Criterion) {
             format!("{i}"),
             VmBench::contract(rng, op::retd(RegId::ONE, 0x10))
                 .unwrap()
-                .with_post_call(vec![op::movi(0x10, *i)])
+                .with_post_call(vec![op::movi(0x10, *i), op::cfe(0x10)])
```
bvrooman, line 73: I'm not sure I understand the significance of adding `cfe` here or why it uses these specific registers. Can you elaborate on why we need this change? Is this just to make sure our benchmarks account for the worst case scenario?

bvrooman (approved):
Just a small optional correct and question

benches/benches/vm_initialization.rs
```diff
@@ -85,6 +87,9 @@ pub fn vm_initialization(c: &mut Criterion) {
         if size as u64 > consensus_params.script_params().max_script_data_length() {
             break
         }
+        if 2 * size as u64 > consensus_params.tx_params().max_size() {
```
MitchTurner, line 90: Can we put this into a helper with the name that explains why this should `break`? Same with the above `break`.

```
if `something_size_is_foo_bar_baz(size, consensus_params)` {
    break
}
```

It's not clear to me what the significance of `size` (size of what?) being half as big as the max tx size.

bvrooman (approved):

benches/benches/vm_initialization.rs
```diff
@@ -99,8 +104,36 @@ pub fn vm_initialization(c: &mut Criterion) {
                     Interpreter::<_, Script, NotSupportedEcal>::with_memory_storage(),
                 );
                 let ready_tx = tx.clone().test_into_ready();
-                black_box(vm.init_script(ready_tx))
-                    .expect("Should be able to execute transaction");
+
+                // Initialize the VM and require the allocation of the whole memory
+                // to charge for the worst possible case.
+                black_box({
```
MitchTurner, line 110: I could say a the same for a lot of this. Can we move everthing in these curlies into some helper?

crates/storage/src/transactional.rs
```diff
@@ -249,6 +283,31 @@ where
             self.storage.get(key, column)
         }
     }
+
+    fn read(
```
MitchTurner, line 287: I'm confused. Did we not have implementations for these methods before on `InMemoryTransaction`?

xgreenx: We had a default implementation that was calling a `get` method which is less optimal in the case of RocksDB.

image

benches/benches/vm_initialization.rs
```diff
@@ -75,36 +79,71 @@ fn transaction<R: Rng>(
...
+                let vm = black_box(
                     Interpreter::<_, Script, NotSupportedEcal>::with_memory_storage(),
                 );
-                let ready_tx = tx.clone().test_into_ready();
-                black_box(vm.init_script(ready_tx))
-                    .expect("Should be able to execute transaction");
+
+                // Initialize the VM and require the allocation of the whole memory
+                // to charge for the worst possible case.
+                black_box(initialize_vm(black_box(tx.clone()), vm));
```
MitchTurner, line 110: This is even more confusing from before. I'm not sure where the extra `black_box` came from either. 

Can we just have a method that says what we want, and then hide the implementation behind it? We probably can get rid of the comment in that case too.

`unoptimized_and_fully_saturated_vm` or something like that. Then have the internal code create the `vm` and wrap the `vm` and the `tx` in `black_box`?

MitchTurner (approved):

PR #1873, opened 2024-04-29
Separate production from dry runs in executor & Cleanup all execution paths :)

MitchTurner:
Closes: https://github.com/FuelLabs/fuel-core/issues/1751

- Separate `Production` from `DryRun` and remove `ExecutionKind` and `ExecutionType`.
- Refactor code down
- Remove `thread_block_transaction` concept because YAGNI
- Remove `PartialBlockComponent` type
- Remove intermediate `inner` functions
- Separate `block_storage_tx` from `BlockExecutor`. This was for a couple of reasons, but the big one is that all the `&mut self`s of the `BlockExecutor` can only be handed out one at a time and that was making the code more complicated. We could have just deconstructed `self`, but then we wouldn't have access to any methods. This was a clean way to separate them, and it makes more sense logically IMO. `ExecutorInstance` still keeps them together, so maybe there is an argument so separate them higher, but this is a good improvement in the mean time.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/executor/src/executor.rs
```diff
@@ -564,55 +609,132 @@ where
...
-                output::contract::Contract {
-                    input_index: 0,
-                    balance_root: Bytes32::zeroed(),
-                    state_root: Bytes32::zeroed(),
-                },
-                amount_to_mint,
-                *self.consensus_params.base_asset_id(),
-                gas_price,
-            );
+        let amount_to_mint = if coinbase_contract_id != ContractId::zeroed() {
```
MitchTurner, line 612: Move to helper

crates/services/executor/src/executor.rs
```diff
@@ -486,137 +375,220 @@ where
...
+            data,
+            MaybeCheckedTransaction::Transaction(coinbase_tx.into()),
+            gas_price,
+            coinbase_contract_id,
+        )?;
+
+        Ok(())
+    }
+
+    #[tracing::instrument(skip_all)]
```
MitchTurner, line 480: Is this redundant? We also have a `skip_all` on `dry_run_without_commit`

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -339,21 +346,70 @@ where
...
         let native_executor_version = self.native_executor_version();
         if block_version == native_executor_version {
             match &self.execution_strategy {
-                ExecutionStrategy::Native => self.native_execute_inner(block, options),
+                ExecutionStrategy::Native => self.native_dry_run_inner(block, options),
                 ExecutionStrategy::Wasm { module } => {
-                    self.wasm_execute_inner(module, block, options)
+                    self.wasm_dry_run_inner(module, block, options)
                 }
             }
```
bvrooman, line 414: Although this introduces some code duplication, the benefit is that it's simple to read and understand. However, I'm curious if you explored any alternative ways to implement these functions that may be a little DRYer. It is understandably challenging given the number of permuations dry run, validation, and Native/Wasm/wasm-executor enabled.

crates/services/executor/src/executor.rs
```diff
@@ -486,137 +375,220 @@ where
...
+
+    fn process_l1_and_l2_txs<T, TxSource>(
+        &self,
+        mut block: PartialFuelBlock,
+        components: &Components<TxSource>,
+        l1_transactions: Vec<CheckedTransaction>,
+        storage_tx: &mut T,
+        data: &mut ExecutionData,
+        skip_failed_txs: bool,
+    ) -> ExecutorResult<PartialFuelBlock>
```
bvrooman, line 528: Would it be feasible to break this up into something like `process_l1` and `process_l2`, and simply call these two functions sequentially?

MitchTurner: Yeah. It's worth looking at a bit more. It does seem like a smell to me as well. I included them together because `produce` and `dry_run` used the exact same code block, so it was a nice way to group those together. I think there probably exists a cleaner solution.

crates/services/upgradable-executor/src/executor.rs
MitchTurner: Been looking at this. I think the only part I'd really want to DRY up more would be 
```rs
        let block_version = block.header_to_produce.state_transition_bytecode_version;
        let native_executor_version = self.native_executor_version();
        if block_version == native_executor_version { .. }
```
But honestly, that would save us like 1 line, especially since validate takes a different `Block` type than the other two, so to DRY all three, we'd have to get the version as a separate line.

I think it's already flat, and easy to read.

WRT to the functions doing essentially the same thing, I think they can all vary independently, so I don't think we should create an abstraction here.

MitchTurner: I was able to find a little place to DRY up the code a bit. Added `new_native_executor_instance` method.

xgreenx (commented):
I've [updated](https://github.com/FuelLabs/fuel-core/pull/1873/commits/042054f123f0736c52aa5bee440203e0f3310716) one of the tests to show the problem with block validation=)

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -339,46 +345,89 @@ where
...
+        } else {
+            Err(ExecutorError::Other(format!(
+                "Not supported version `{block_version}`. Expected version is `{}`",
+                Self::VERSION
+            )))
+        }
+    }
+
+    #[cfg(feature = "wasm-executor")]
+    fn dry_run_inner<TxSource>(
```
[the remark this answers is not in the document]
MitchTurner: Okay. I just added a `dry_run` boolean param to some of the internal functions to branch. I don't think we need a type. I kept the external functions the same, partly because we have a bunch of test functions that won't ever be used for dry_run, so I didn't want to clutter the interfaces with the extra param.

MitchTurner:
@xgreenx 
>I've [updated](https://github.com/FuelLabs/fuel-core/pull/1873/commits/042054f123f0736c52aa5bee440203e0f3310716) one of the tests to show the problem with block validation=)

I've added a solution. The route I went will check that all the L1 transactions are included in the block before executing those txs. Is that what we want though? Does the validator care whether the txs come from L1 or L2? We want to have the producer always include them immediately, at least for now, but IDK if the validator should check that.

What do you think? Instead we could just remove the L1 tx checks completely.

crates/types/src/services/executor.rs
```diff
@@ -379,6 +242,8 @@ pub enum Error {
     RelayerGivesIncorrectMessages,
     #[display(fmt = "Consensus parameters not found for version {_0}")]
     ConsensusParametersNotFound(ConsensusParametersVersion),
+    #[display(fmt = "The transaction is not in the block")]
+    L1TransactionNotInBlock,
```
Dentosal, line 246: nit: should this be `DaTransactionNotInBlock`?

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -498,39 +516,52 @@ where
...
+    //     block: Components<TxSource>,
+    //     options: ExecutionOptions,
+    // ) -> ExecutorResult<Uncommitted<ExecutionResult, Changes>>
+    // where
+    //     TxSource: TransactionsSource + Send + Sync + 'static,
+    // {
+    //     let instance = self.new_native_executor_instance(options);
+    //     instance.dry_run_without_commit(block)
+    // }
+
```
Dentosal, line 543: Commented-out code should probably be removed

crates/types/src/services/executor.rs
MitchTurner: This is being removed because we changed how the checks work.

But, in general, I tend to think of "DA" as being for fraud proving stuff, which is irrelevant from this perspective. What's relevant is that it's coming from the parent chain, so "L1" or "Relayed" are my chosen descriptors here.

Yes, our implementation will have "DA" and "L1" be the same, but the block producer/validator just needs txs from some source.

Dentosal (approved):

xgreenx (approved):

Issue #1882, opened 2024-05-07
Pass the `Block` by reference and return the `ValidationResult`


xgreenx:
We need to pass the `Block` by reference and return the `ValidationResult` without the `Block` type inside. It can be follow-up=)

_Originally posted by @xgreenx in https://github.com/FuelLabs/fuel-core/pull/1873#discussion_r1592146690_

PR #1886, opened 2024-05-07
Use ref instead of owned `Block` in validation

MitchTurner:
https://github.com/FuelLabs/fuel-core/issues/1882

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/services/upgradable-executor/src/executor.rs
+++ b/crates/services/upgradable-executor/src/executor.rs
@@ -500,24 +500,7 @@ where
 
         match output {
             fuel_core_wasm_executor::utils::ReturnType::V1(result) => {
-                let (
-                    ExecutionResult {
-                        skipped_transactions,
-                        tx_status,
-                        events,
-                        ..
-                    },
-                    changes,
-                ) = result?.into();
-
-                Ok(Uncommitted::new(
-                    ValidationResult {
-                        skipped_transactions,
-                        tx_status,
-                        events,
-                    },
-                    changes,
-                ))
+                Ok(result?.into_validation_result())
             }
         }
     }
--- a/crates/services/upgradable-executor/src/executor.rs
+++ b/crates/services/upgradable-executor/src/executor.rs
@@ -54,6 +54,10 @@ use fuel_core_storage::{
 use fuel_core_types::blockchain::block::PartialFuelBlock;
 #[cfg(any(test, feature = "test-helpers"))]
 use fuel_core_types::services::executor::UncommittedResult;
+use fuel_core_types::services::executor::{
+    UncommittedValidationResult,
+    ValidationResult,
+};
 
 #[cfg(feature = "wasm-executor")]
 enum ExecutionStrategy {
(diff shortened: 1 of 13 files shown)


crates/services/upgradable-executor/src/executor.rs
```diff
@@ -235,11 +238,13 @@ where
         self.produce_without_commit_with_coinbase(block, Default::default(), 0)
     }
 
+    // TODO: Can we remove this and just use `validate` everywhere?
```
xgreenx, line 241: Yeah, we can do it now=)

crates/services/upgradable-executor/src/instance.rs
```diff
@@ -478,13 +478,16 @@ impl Instance<Relayer> {
...
     }
 
+    //
     pub fn add_validation_input_data(
         self,
-        block: Block,
+        block: &Block,
         options: ExecutionOptions,
     ) -> ExecutorResult<Instance<InputData>> {
+        // TODO: Can we avoid cloning the block?
```
xgreenx, line 487: Yeah, you can have two types. One for serialization on our side and one for deserialization on the WASM side. On our side you can pass reference

crates/services/upgradable-executor/wasm-executor/src/main.rs
```diff
@@ -114,15 +115,16 @@ fn execute_validation(
...
             skipped_transactions,
             tx_status,
             events,
         },
         changes,
-    ) = instance.validate_without_commit(block)?.into();
+    ) = instance.validate_without_commit(&block)?.into();
 
+    // We have to return the same type here because if the input parsing fails it won't know if this
+    // is a validation or production block.
```
MitchTurner, line 127: Maybe instead of serializing `ReturnType::V1(result)` in `extern "C" fn execute`, we serialize some `Result<ReturnType, Error>`?

That way we don't need the `ReturnType` variant to return an `Err`.

i.e. We can do something like:
```
enum ReturnType {
    V1(V1ReturnType)
}

enum V1ReturnType {
    Execute(ExecutionResult),
    Validate(ValidationResult),
}
```

xgreenx: We want to return an `ExecutorResult` because we may change the error type later. We could add ` V2` that returns a validation result and name it `Validation` instead of `V2`=)

But I don't see any problems in returning `ExecutionResult` with a `block` type since we don't consume `block`, and it doesn't hurt the performance.

xgreenx: Actually, it affects performance. We need to serialize and deserialize the block on the upgradeable executor side, so maybe we should add a new return variant. But let’s do it in a separate PR, we need to add an integration test for it to be sure that we did everything correct=)

MitchTurner: https://github.com/FuelLabs/fuel-core/issues/1887

xgreenx (approved):

PR #1888, opened 2024-05-08
Reuse VM memory across executions

### LINKED ISSUE #1878: Reuse the same instance of the VM for txs in block production and validation
The change https://github.com/FuelLabs/fuel-vm/pull/697 adds a way for a nice optimization to use the one an instance of the `FuelVM` for executing all transactions.

It should remove the burden on the end user to pay for the initialization of the VM.

Dentosal:
Closes #1878. Requires https://github.com/FuelLabs/fuel-vm/pull/732 and a new fuel-vm release.

Introduces a per-thread pool for allocated VM memory instances. Whenever a new VM instance is needed for a tx, the memory is taken from this pool. After the tx has been executed, the memory is reset to an empty state without deallocating, and returned to the pool.

A per-thread pool was chosen to no pass allocations between threads unnecessarily.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog (none)

### Before requesting review


crates/services/executor/src/vm_pool.rs
```diff
@@ -0,0 +1,39 @@
...
+
+thread_local! {
+    static POOL: RefCell<Vec<Memory>> = RefCell::new(Vec::new());
+}
+
+/// Gets a new VM memory instance from the pool.
+pub(crate) fn get_vm_memory() -> Memory {
+    POOL.with(|pool| {
+        let mut pool = pool.borrow_mut();
+        pool.pop().unwrap_or_else(Memory::new)
```
Voxelot, line 13: We should probably have some kind constraint on the max number of items allowed in each pool? Otherwise it will wait if there are none available.

Dentosal: We didn't do this earlier, and I don't see why it would make sense to track the items here, instead of just limiting concurrency of the executor itself. That said, keeping a pool per thread is by itself problematic and not useful, as:

1. The threads are not CPU-pinned anyway
2. The async runtime might decide to create new threads arbitarily, which would essentially cause a memory leak here

So I' changed the code to not keep separate pools in 5caec42

crates/services/executor/src/executor.rs
```diff
@@ -1430,33 +1436,61 @@ where
...
+        recycling_try!(self.compute_inputs(tx.inputs_mut(), storage_tx));
 
         // only commit state changes if execution was a success
         if !reverted {
             self.log_backtrace(&vm, &receipts);
+            self.vm_pool.recycle(mem::take(vm.memory_mut()));
             let changes = sub_block_db_commit.into_changes();
             storage_tx.commit_changes(changes)?;
+        } else {
+            self.vm_pool.recycle(mem::take(vm.memory_mut()));
```
MitchTurner, line 1493: Since we are always recycling the memory, why do we even need the `recycling_try` macro?

Dentosal: Since we need to recycle just before returning early. The continuing branch still needs the vm.

Dentosal: I made it quite a bit prettier on the fuel-vm side, now it just auto-recycles on drop.

xgreenx (approved):

Issue #1624, opened 2024-01-24
Feature: Implement algorithm to determine the base gas price


xgreenx:
We want to move to something similar to Ethereum's [EIP-1559 ](https://eips.ethereum.org/EIPS/eip-1559) base fees. The idea being that the price of gas will fluctuate from block to block depending on the demand on the network. We also need to take into account demand on the DA layer.

This is related as well to the `estimate_gas_price` api endpoint that needs some rules for how much the price is allowed to change from one block to the next. This work is captured in [this](https://github.com/FuelLabs/fuel-core/issues/1653) issue.

This will require at least 3 inputs:
- DA Layer posting costs
- Shared sequencer payment
- Fuel Node maintainance

PR #1889, opened 2024-05-08
Add new `FuelGasPriceProvider`

MitchTurner:
Part of: https://github.com/FuelLabs/fuel-core/issues/1624

Here is the current specs for the gas price algorithm:
https://www.notion.so/fuellabs/Gas-Price-Algo-V1-Specs-2e7a06263c4f46928b6d179e90a4ba1a

Introduce a new type to replace `StaticGasPrice` that will actually calculate new gas prices based on historical data.

This work is largely exploratory since we don't know exactly what the algorithm will land on. But this PR includes work defining the architecture of our `FuelGasPriceProvider` as well as some exploratory work defining a proof-of-concept algorithm. This is the main reason this PR is still in "Draft".

The current design separates the algorithm from the `FuelBlockProvider`. This helps identify what behavior is independent of the specifc algorithm we end up choosing and helps us determine the minimum set of variables the algorithm requires. Currently the algorithm looks like this:

```rs
pub trait GasPriceAlgorithm {
    fn calculate_da_gas_price(
        &self,
        previous_da_gas_prices: u64,
        total_production_reward: u64,
        total_da_recording_cost: u64,
    ) -> u64;

    fn calculate_execution_gas_price(
        &self,
        previous_execution_gas_prices: u64,
        block_fullness: BlockFullness,
    ) -> u64;

    fn maximum_next_gas_prices(&self, previous_gas_prices: GasPrices) -> GasPrices;
}
```

The actual impllementation of the `FuelGasPriceAlgo` for now is going to take inspiration from a [PID control loop](https://en.wikipedia.org/wiki/Proportional%E2%80%93integral%E2%80%93derivative_controller) on the DA gas price side, but ideally as simple as possible and as deterministic as possible (we don't want to use floats in places like `BlockFullness`).

Architecture Diagram (WIP, I'll try to update as it changes). CLICK IF NOT SHOWING UP PROPERLY IN DARK MODE:
image

## Checklist

### Before requesting review

- [ ] Someone else?


crates/fuel-core/src/service/adapters/fuel_gas_price_provider.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    fn total_cost(&self) -> u64 {
+        *self.profitablility_totals.total_cost.borrow()
+    }
+
+    fn totaled_block_height(&self) -> BlockHeight {
+        *self.profitablility_totals.totaled_block_height.borrow()
+    }
+
+    fn update_totals(&self, latest_block_height: BlockHeight) -> Result<()> {
```
MitchTurner, line 158: We need to take into account the fact that DA recording has some lag to it. Possibly 12-24 blocks? How does that effect the `gas_price` calculation? The current design doesn't take that into account.

crates/fuel-core/src/service/adapters/fuel_gas_price_provider.rs
```diff
@@ -0,0 +1,188 @@
...
+        let new_gas_price_candidate = self.algorithm.calculate_gas_price(
+            previous_gas_price,
+            self.total_reward(),
+            self.total_cost(),
+            block_fullness,
+        );
+        let new_gas_price = core::cmp::min(
+            new_gas_price_candidate,
+            self.algorithm.maximum_next_gas_price(previous_gas_price),
+        );
```
MitchTurner, line 142: This is kinda a smell. I'm not sure if this should be in the provider or the algo. I'm getting a feeling like the provider and the algo abstractions could be merged some, especially when we start looking at the estimation trait impl.

crates/fuel-core/src/service/adapters/fuel_gas_price_provider.rs
MitchTurner: This is implemented. Might need more testing.

crates/fuel-core/src/service/adapters/fuel_gas_price_provider.rs
```diff
@@ -0,0 +1,228 @@
...
+            .ok_or(Error::GasPricesNotFoundForBlockHeight(latest_block_height))?;
+        let latest_da_recorded_block = self
+            .da_recording_cost_history
+            .latest_height()
+            .map_err(Error::UnableToGetLatestBlockHeight)?;
+
+        let new_da_gas_price = if latest_da_recorded_block
+            <= self.profitablility_totals.totaled_block_height()
+        {
+            previous_gas_prices.da_gas_price
```
xgreenx, line 155: Since the records are always behind for some period of time, we should always go into this branch. But since `self.profitablility_totals.totaled_block_height()` depends on the number of `self.update_totals` calls, it will not happen. But because of that, it creates very unintuitive code.

What do you think if we always call `self.algorithm.calculate_da_gas_price`, but the `self.total_reward()` and `self.total_cost()` are different? We can achieve almost the same effect just by advancing reward and cost.

Because we always don't have live data from DA, we can predict the cost based on the previous blocks. We can calculate the da gas price per byte based on data from the past(total_cost/total_bytes). Using this price, we can calculate the cost for the upcoming block(txs_size * calculated_gas_price). The `total_cost` is `known_cost_from_the_past` + `cumulutive_predicted_cost`. Using `total_cost`, you can calculate the next gas price, which will give you a reward.

crates/fuel-core/src/service/adapters/fuel_gas_price_provider.rs
```diff
@@ -0,0 +1,228 @@
...
+        } else {
+            self.update_totals(latest_block_height)?;
+            self.algorithm.calculate_da_gas_price(
+                previous_gas_prices.da_gas_price,
+                self.total_reward(),
+                self.total_cost(),
+            )
+        };
+
+        let block_fullness = self
```
xgreenx, line 165: I think the previous block's fullness should be part of the `GasProviderService`. On each block imported from the block importer, the service will update its internal fullness variables and dump new values into the database.

crates/fuel-core/src/service/adapters/fuel_gas_price_provider/algorithm_adapter.rs
```diff
@@ -0,0 +1,184 @@
...
+                previous_da_gas_prices,
+                total_production_reward,
+                total_da_recording_cost,
+            ),
+        }
+    }
+
+    fn calculate_execution_gas_price(
+        &self,
+        previous_execution_gas_prices: u64,
```
xgreenx, line 36:
```suggestion
        previous_execution_gas_price: u64,
```

crates/fuel-core/src/service/adapters/fuel_gas_price_provider/ports.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+pub trait GasPriceHistory {
+    fn gas_prices(&self, height: BlockHeight) -> ForeignResult<Option<GasPrices>>;
+    // TODO: Can we make this mutable? The problem is that the `GasPriceProvider` isn't mut, so
+    //   we can't hold mutable references. Might be able to make `GasPriceProvider` `&mut self` but will
+    //  take some finangling
+    fn store_gas_prices(
+        &self,
```
xgreenx, line 106: Yeah, it should be mutable. The gas price provider service should have exclusive ownership of the database(we can create a separate database for it).

crates/gas-price-algorithm/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+
+pub struct AlgorithmV1 {
+    // DA
+    da_max_change_percent: u8,
+    min_da_price: u64,
+    da_p_value_factor: i64,
+    da_d_value_factor: i64,
+    moving_average_profit: RefCell<i64>,
+    last_profit: RefCell<i64>,
+    profit_slope: RefCell<i64>,
```
xgreenx, line 17: Why do we need `RefCell`? If you need to update the internal state, then let's use `&mut`=)

crates/gas-price-algorithm/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+        };
+        max(new, self.min_exec_price)
+    }
+
+    fn calculate_new_moving_average(&self, new_profit: i64) {
+        let old = *self.moving_average_profit.borrow();
+        *self.moving_average_profit.borrow_mut() = (old
+            * (self.moving_average_window - 1))
+            .saturating_add(new_profit)
+            .saturating_div(self.moving_average_window);
```
xgreenx, line 102: Hmm, it seems that after restarting the node `self.moving_average_profit` and `self.last_profit` will be zero, which means the algorithm will work differently. We need to persist all information required for the algorithm.

Hmm, it would be nice if we used one iteration formula to avoid a persistent state. But I assume you haven't found any suitable for us=)

xgreenx: Or maybe 

```rust
        old_da_gas_price: u64,
        old_total_production_reward: u64,
        old_total_da_recording_cost: u64,
```

can be stored inside of the algorithm and passed during creation. Then we should be able to calculate profit form them.

crates/services/producer/src/block_producer/gas_price.rs
```diff
@@ -29,7 +29,10 @@ impl From<BlockHeight> for GasPriceParams {
 /// Interface for retrieving the gas price for a block
 pub trait GasPriceProvider {
     /// The gas price for all transactions in the block.
-    fn gas_price(&self, params: GasPriceParams) -> Option<u64>;
+    fn gas_price(
+        &self,
+        params: GasPriceParams,
+    ) -> Result<u64, Box<dyn std::error::Error>>;
```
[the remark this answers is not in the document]
MitchTurner: I guess so. Since our "Producer Error" is `anyhow::Error`. 

There are other places where I might invert the error dep for the adapters as well. 

I'm torn on this in Rust in general, because adapters necessarily have their own error type, so you have to use type erasure. And because of that, I figured we could just put the domain errors on the business side rather than expecting the adapter to return the right error type.

For example:

```rs
        let previous_gas_prices = self
            .gas_price_history
            .gas_prices(latest_block_height)
            .map_err(Error::UnableToGetGasPrices)?
            .ok_or(Error::GasPricesNotFoundForBlockHeight(latest_block_height))?;
```

But that's noisy, and we are already trusting the adapter code to uphold the contract of the port, so maybe we just push into the adapters instead of using this `ForiegnError` construct that I used.

crates/gas-price-algorithm/src/lib.rs
MitchTurner: I agree. Kinda related to the `TODO` above.

crates/fuel-core/src/service/adapters/fuel_gas_price_provider.rs
MitchTurner: Can we know the bytes of the block before we calculate the gas price? I guess theoretically, but the way the code is currently written the gas price is selected before we add the txs to the block, that way we can check that it includes enough to cover fees.

The gas price will never exceed the max estimated gas price, so we can check that each tx has enough to cover that value when adding, then calculate gas price and change after? But that's not how the producer code works rn AFAIK.

We can of course know the older blocks byte size, but I don't know if that helps if we can't compare our current block to it.

MitchTurner: >Because we always don't have live data from DA, we can predict the cost based on the previous blocks. We can calculate the da gas price per byte based on data from the past(total_cost/total_bytes). Using this price, we can calculate the cost for the upcoming block(txs_size * calculated_gas_price). The total_cost is known_cost_from_the_past + cumulutive_predicted_cost. Using total_cost, you can calculate the next gas price, which will give you a reward.

This is what Optimism does:
>The L1 Data Fee is automatically charged for any transaction that is included in an OP Mainnet block. This fee is deducted directly from the address that sent the transaction. The exact amount paid depends on the estimated size of the transaction in bytes after compression, the current Ethereum gas price and/or blob gas price, and several small parameters.

We don't have the luxury of deducting from the address. And we decided to only have one gas price. So currently we calculate the gas price before we add txs to the block, when we don't know how many bytes the txs will take up in that block. Of course, we could use `dry_run` to calculate the max fee, so that's still an option if we want to separate them.

MitchTurner: Also relevant from Optimism:
> It is currently not possible to limit the maximum L1 Data Fee that a transaction is willing to pay. This is the result of limitations in existing Ethereum transaction formats and tradeoffs that the OP Stack makes to retain better EVM equivalence. Work is underway to improve this situation with the introduction of a new standardized transaction type. For now, you should be aware that the L1 Data Fee can fluctuate with the Ethereum gas price.

crates/fuel-core/src/service/adapters/fuel_gas_price_provider.rs
```diff
@@ -0,0 +1,228 @@
...
+                .map_err(Error::UnableToGetGasPrices)?
+                .ok_or(Error::GasPricesNotFoundForBlockHeight(
+                    requested_block_height,
+                ))?;
+            Ok(gas_prices.total())
+        } else if Self::asking_for_next_block(latest_block, requested_block_height) {
+            let new_gas_prices = self.calculate_new_gas_price(latest_block)?;
+            self.gas_price_history
+                .store_gas_prices(latest_block, &new_gas_prices)
+                .map_err(Error::UnableToStoreGasPrices)?;
```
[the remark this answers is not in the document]
MitchTurner: So, you're suggesting that the `GasPriceProvider` trait should have more methods? Like maybe a `get_gas_price`, a `new_gas_price` and a `write_gas_price`?

We create a new gas price, build the block, and then write it once it's built?

I don't know if the producer trait `GasPriceProvider` will ever need to get historical gas prices. But some of the other traits might.

crates/fuel-core/src/service/adapters/fuel_gas_price_provider/ports.rs
MitchTurner: > gas price provider service

I don't think of this as being a service. I just thing of it as being an adapter handed to the producer, txpool, and graphql like here:
https://github.com/FuelLabs/fuel-core/blob/acd8d1931eba42e82fcfd7396c5d23882b0e82a9/crates/services/producer/src/block_producer.rs#L124-L127

crates/gas-price-algorithm/src/lib.rs
MitchTurner: This is a great point. Configuring of that might be a pain if the services aren't running when we start the algorithm, but it would be a solution. But how do we know what values it died on?

Another solution is we have an assoc type for the algo trait that the algo returns with arbitrary metadata. The type needs to match some data storage on the provider at compilation time, and we can have that type be versioned so the data storage can still be backwards compatible. Seems like a lot of trouble lol.

```
impl<FB, DA, A, GP> FuelGasPriceProvider<FB, DA, A, GP>
where
    FB: FuelBlockHistory,
    DA: DARecordingCostHistory,
    A: GasPriceAlgorithm<Metadata=<Self as GasPriceHistory>::Metadata>,
    GP: GasPriceHistory,
{
```
ooooor we can add storage to the algorithm--which is just so many traits 
:turtle: 
:turtle: 
:turtle: 
:turtle: 

I'd suggest that we just merge the algo abstraction into the provider, but I think that would make versioning harder. 

I feel like there is a cleaner solution here.

crates/fuel-core/src/service/adapters/fuel_gas_price_provider.rs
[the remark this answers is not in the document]
MitchTurner: I remember why I'm storing it here. Because we only have one `gas_price`, but under the hood we need to record the `da_gas_price` and the `exec_gas_price` separately.

But you're right, if that block fails for some reason, then we have wrong values. I think that's fine because we are getting `latest_block_height` independently from `requested_block_height`.

That means that this should never be a getter though. So it assumes that the requested block is always the head of the chain and it can remove "bad branches".

crates/fuel-core/src/service/adapters/fuel_gas_price_provider.rs
```diff
@@ -0,0 +1,228 @@
...
+        latest_block: BlockHeight,
+        block_height: BlockHeight,
+    ) -> bool {
+        *latest_block + 1 == *block_height
+    }
+
+    fn calculate_new_gas_price(
+        &self,
+        latest_block_height: BlockHeight,
+    ) -> Result<GasPrices> {
```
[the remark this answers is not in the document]
MitchTurner: Okay. If we are able to move the logic for updating DA information in the non-live scenarios into the algorithm, then we can get rid of database tracking outside of the algorithm.

crates/fuel-core/src/service/adapters/fuel_gas_price_provider.rs
```diff
@@ -0,0 +1,225 @@
...
+                .gas_prices(requested_block_height)
+                .map_err(Error::UnableToGetGasPrices)?
+                .ok_or(Error::GasPricesNotFoundForBlockHeight(
+                    requested_block_height,
+                ))?;
+            Ok(gas_prices.total())
+        } else if Self::asking_for_next_block(latest_block, requested_block_height) {
+            let new_gas_prices = self.calculate_new_gas_price(latest_block)?;
+            self.gas_price_history
+                .store_gas_prices(requested_block_height, &new_gas_prices)
```
MitchTurner, line 120: This was `store_gas_prices(latest_block)` before and the tests were still passing. Fixing it didn't break the tests either, so the coverage needs to be updated if we want to keep this approach.

crates/fuel-core/src/service/adapters/fuel_gas_price_provider.rs
MitchTurner: I'm not sure what you mean `GasProviderService`. Right now the producer, the sync, and graphql each get a copy of the `impl GasPriceProvider for StaticGasPrice`.

That's how I imagine it will still work, but just with this new `FuelGasPriceProvider` instead of `StaticGasPrice`. I'm assuming that all the `GasPriceProvider` needs is a `BlockHeight` and then it has access to all the DBs it needs to assemble the gas price. I don't know why it needs to store `block_fullness` when it can just look up the block's data in its DB.

PR #1914, opened 2024-05-29
Fixed halting of the node in rare conditions

xgreenx:
My node halted during synchronization with the dev cluster. According to the logs, the importing of the blocks stopped because someone was still processing the previous block. I've added a 20-second timeout on the block processing to debug this issue more easily in the future.

During my node's graceful shutdown, the `PoA` service was waiting for the child `PoASyncTask` to finish his work. The child process was locked and captured the `ImportResult`, which caused the importer service to be locked. I assume one of `DeadlineClock`'s await functions caused that. I haven't looked into how exactly it could happen; I just replaced the usage of the custom `DeadlineClock` with an `Interval` since it doesn't require `await`. It is just a bad practice to call `await` inside of the `tokio::select`. 

Maybe, we need to remove `DeadlineClock` from the `PoA` service as well later just to minimize `await` in the `tokio::select`. Or re-write it without async functions.

### Before requesting review


MitchTurner (approved):
> Maybe, we need to remove DeadlineClock from the PoA service as well later just to minimize await in the tokio::select. Or re-write it without async functions.

I was just going to ask if we still need `DeadlineClock` when I reread this. I'll create an issue to investigate:
https://github.com/FuelLabs/fuel-core/issues/1917

Issue #1592, opened 2024-01-08
Remove deprecated `libp2p` `set_max_buffer_size` method call


MitchTurner:
Waiting on response from `libp2p` people here:
https://github.com/libp2p/rust-libp2p/discussions/5073

PR #1915, opened 2024-05-29
Fixed reconnection issue in the dev cluster with AWS cluster

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1592

This change improves and simplifies some aspects of the P2P service. It also fixes the issue of not reconnecting to the reserved nodes when the reserved node is restarted and got new IP.

- The change moves the reconnection handling into the `PeerReport` behavior. It removes ping-ponging reserved peers between the primary behavior and the `PeerReport` behavior and encapsulates the logic inside the `PeerReport`. Also, it eliminates the timer and replaces it with the queue of reconnections, reducing noise in logs(before, we had much more trash errors).
- Added logs for cases when the dial fails. They are very helpful to debug issues with connection.
- Simplified initialization of the `ConnectionTracker` and `FuelAuthenticated`. It allows the reuse of libp2p built-in connections builder.
- Removed the usage of the Mplex since it doesn't have a backpressure mechanism. Now we use Yamux by default. It is breaking the change since nodes with the old codebase can't connect to new nodes.
- Propagated `max_concurrent_streams` for request-response protocol.

## Checklist

### Before requesting review


crates/services/p2p/src/p2p_service.rs
```diff
@@ -174,11 +175,18 @@ impl FuelP2PService {
...
+        let tcp_config = tcp::Config::new().port_reuse(true);
         let behaviour = FuelBehaviour::new(&config, codec.clone());
 
         let mut swarm = SwarmBuilder::with_existing_identity(config.keypair.clone())
             .with_tokio()
-            .with_other_transport(transport_function)
+            .with_tcp(
+                tcp_config,
+                transport_function,
+                libp2p::yamux::Config::default,
```
xgreenx, line 186: Breaking change: We don't support Mplex anymore.

crates/services/p2p/src/peer_report.rs
```diff
@@ -142,17 +193,30 @@ impl NetworkBehaviour for Behaviour {
...
+            return Poll::Ready(event)
         }
 
-        if self.decay_interval.poll_tick(cx).is_ready() {
-            return Poll::Ready(ToSwarm::GenerateEvent(PeerReportEvent::PerformDecay))
+        if let Some((instant, peer_id)) = self.reserved_nodes_to_connect.front() {
+            if instant.elapsed() > Duration::from_secs(HEALTH_CHECK_INTERVAL_IN_SECONDS) {
+                let peer_id = *peer_id;
+                self.reserved_nodes_to_connect.pop_front();
+                let multiaddrs = self
```
xgreenx, line 203: The actual fix for the main reconnection problem that I faced with the dev cluster. The initial DNS address was replaced with a real IP, but when the sentry crashed, the IP was another one. Using initial multiaddrs here allows you to reconnect and get a new IP again.

Dentosal: We should probably have a comment somewhere here explaining this.

Dentosal (approved):

PR #1916, opened 2024-05-29
Speeds up synchronisation of the blocks for the `fuel-core-sync` service

xgreenx:
The change speeds up synchronisation and removes noise logs when the P2P request crashes.

The synchronization speed improvement is done by:
- Making all requests asynchronous. Before, the first request to fetch the blocked header was synchronous and required awaiting it because of `and_then`.
- The process of block fetching is separated from block execution. Before, we fetched X blocks and waited for their execution before requesting new X blocks. The fetching is a new parallel task that uses the `batch_sender` variable for backpressure. The task fetches blocks in parallel and sends them into a channel with buffer `1`. The executor's task is to receive the batch from the channel and execute it, allowing the request for a new batch to be made in parallel.

If the execution speed is faster than fetching, we request more than one batch in parallel, up to the buffer limit. Tests shows that.

The change also handles cases better when a P2P request fails without a known `PeerId`.

Without change, I was syncing 2M blocks for 4 hours. With this change, it takes me 20 minutes.

## Checklist

### Before requesting review


crates/services/sync/src/state.rs
```diff
@@ -83,16 +83,26 @@ impl State {
...
-            Status::Processing(range) => match height.cmp(range.end()) {
-                // The commit is less than the end of the range, so the range
-                // is still being processed.
-                Ordering::Less => {
-                    Some(Status::Processing(height.saturating_add(1)..=*range.end()))
+            Status::Processing(range) => {
+                if height < *range.start() {
+                    // The committed height is less than the start of the processing range,
+                    // it is a lag between committing and processing.
+                    None
```
xgreenx, line 90: It removes strange logs where committed height goes down. Plus removes duplicate logs

crates/services/sync/src/import.rs
```diff
@@ -300,68 +361,76 @@ fn get_block_stream<
...
-    params: &Config,
+    params: Config,
     p2p: Arc<P>,
-) -> impl Stream<Item = SealedHeaderBatch> {
+) -> impl Stream<Item = impl Future<Output = SealedHeaderBatch>> {
     let Config {
         header_batch_size, ..
     } = params;
-    let ranges = range_chunks(range, *header_batch_size);
-    futures::stream::iter(ranges).then(move |range| {
```
MitchTurner, line 364: So was `then` causing the inner inner future to get awaited? Was this blocking subsequent items in the stream?

crates/services/sync/src/import/tests.rs
```diff
@@ -1026,6 +1027,65 @@ async fn import__missing_transactions_sends_peer_report() {
         .await;
 }
 
+#[tokio::test]
+async fn import__execution_error_on_header_4_when_awaits_for_1000000_blocks() {
```
MitchTurner, line 1031: I'm not sure I understand this test. What does `1000000` have to do with the error? Is it just to show that we don't import them out of order? i.e. if header 4 doesn't import correctly, none after will either?

xgreenx: It shows that if the first query fails, it stops requesting all remaining queries. You can check [this commit](https://github.com/FuelLabs/fuel-core/pull/1916/commits/241e6ecf951a14ce2dbfd361e38c5a5a0e05fc6c) to see the problematic place=)

MitchTurner: Okay. I see. Makes sense.

MitchTurner (approved):

Issue #1621, opened 2024-01-24
Shared sequencer integration


xgreenx:

MitchTurner:
Will staking and staking parameters need to be included in this feature?

MitchTurner:
We need to figure out what data we will send from the block builder to the SS. This includes signatures and other data formatting stuff.

xgreenx:
@Dentosal You should have access to the repository right now.

Bellow instruction which we received.

> Hey @xgreenx99 the documentation on the Sequencer's README should be helpful for this. https://github.com/fuel-infrastructure/fuel-sequencer
>
>I point you specifically to the make run-sequencer-no-sidecar command, which will require you to have Golang 1.21 installed and Ignite CLI (linked in the README).
>
>This command will start up the Sequencer with a set of accounts (refer to config-no-sidecar.yml file for a full breakdown of accounts), and it will also install the fuelsequencerd binary for interacting with the network.

Maybe instead of building it manually we could try to use docker from  https://github.com/FuelLabs/fuel-rollup/tree/main/docker.

>For posting blobs, you will need to set up a client to the Sequencer. This is equivalent to setting up a client to a typical Cosmos SDK chain.
>
>The Sequencer-specific thing will be the message itself - MsgPostBlob. You can find the definition of it here https://github.com/fuel-infrastructure/fuel-sequencer/blob/e60fd0bf9d1f47b87a19cf81668f9ba5ab5f6c84/proto/fuelsequencer/sequencing/v1/tx.proto#L44-L62
>
>(Also described in the Sequencer Spec - https://github.com/fuel-infrastructure/notes/blob/main/spec/Spec.md#user-transactions)
>I wish to point out is a possible issue you might run into 🚨 We've used the bytes type for the topic ID and the blob data. By default, Cosmos SDK will expect base64-encoded data for bytes. If we require support for hex topic IDs and hex-encoded blob data, then this will need to be tweaked from our side 

Our goal is to use [PostBlob](https://github.com/fuel-infrastructure/notes/blob/main/spec/Spec.md#user-transactions) transaction to post Fuel blocks into the shared sequencer. The `From` field is an address of the PoA node. Based on the protobuf transaction definition, we can generate a Rust version of it. We could try to use [cosmos Rust SDK](https://github.com/cosmos/cosmos-rust/tree/main/cosmrs) to sign transactions, but if we can do that without it, it would be awesome.

[...]

PR #1922, opened 2024-05-30
Shared sequencer integration

Dentosal:
Related https://github.com/FuelLabs/fuel-core/issues/1621

As part of the feature, we created a new crate https://github.com/FuelLabs/sequencer-rs-sdk with protobuf types from the SS.

The PR doesn't have any testing currently. The `fuel-core` was tested with enabled SS functionality on the local node created from https://github.com/FuelLabs/fuel-rollup/blob/main/docker/docker-compose.yml, and on the SS testnet using data from https://github.com/fuel-infrastructure/networks/tree/main/seq-testnet-1.

We need to add integration tests and unit tests in the follow-up PR.

## Checklist

### Before requesting review


xgreenx:
@Dentosal Could you apply latest changes from the https://github.com/FuelLabs/fuel-core/pull/2451 and https://github.com/FuelLabs/fuel-core/pull/2452 please?=)

xgreenx:
@Dentosal Could you also include https://github.com/FuelLabs/fuel-core/pull/2495 into the PR, please?=)

xgreenx:
The `CosmosError` was removed because it was not used after https://github.com/FuelLabs/fuel-core/pull/2451/commits/8bf2a3e0ea5ce72d7e9e1c1eda92477c60bb7857 commit

xgreenx (dismissed):

xgreenx (approved):

ci_checks.sh
```diff
@@ -31,7 +31,7 @@ cargo make check --all-features --locked &&
 cargo make check --locked &&
 OVERRIDE_CHAIN_CONFIGS=true cargo test --test integration_tests local_node &&
 cargo nextest run --workspace &&
-FUEL_ALWAYS_USE_WASM=true cargo nextest run --all-features --workspace &&
+cargo nextest run --all-features --workspace &&
```
AurelienFT, line 34: What's the rationale behind this change on this PR ?

AurelienFT (approved):
I already reviewed when it was a PR for 0.40.x version. Seems similar LGTM with a little question

crates/services/shared-sequencer/src/lib.rs
```diff
@@ -0,0 +1,265 @@
...
+pub mod service;
+
+/// Shared sequencer client
+pub struct Client {
+    endpoints: Endpoints,
+    topic: [u8; 32],
+    chain_id: Id,
+    gas_price: u128,
+    coin_denom: Denom,
+    account_prefix: String,
```
acerone85, line 52: nit: missing documentation for the fields. 
But in general I'd like to understand what's the account prefix?

crates/services/shared-sequencer/src/http_api.rs
```diff
@@ -0,0 +1,189 @@
...
+
+    #[derive(Debug, Deserialize)]
+    pub struct TopicResponse {
+        pub topic: TopicInfo,
+    }
+
+    #[derive(Debug, Deserialize)]
+    pub struct TopicInfo {
+        pub owner: String,
+        pub order: String,
```
acerone85, line 52: In general I'm a fan of the newtype pattern to give context about what a value represents. 
But in abserve of that, who is the owner (I guess either a Fuel or Ethereum address) and what is the order of a topic?

acerone85: nvm, I see that this struct is redefined later and you use these fields only for deserialization purposes. Closing

crates/services/shared-sequencer/src/lib.rs
```diff
@@ -0,0 +1,265 @@
...
+impl Client {
+    /// Create a new shared sequencer client from config.
+    pub async fn new(endpoints: Endpoints, topic: [u8; 32]) -> anyhow::Result<Self> {
+        let coin_denom = http_api::coin_denom(&endpoints.blockchain_rest_api)
+            .await?
+            .parse()
+            .map_err(|e| anyhow::anyhow!("{e:?}"))?;
+        let account_prefix =
+            http_api::get_account_prefix(&endpoints.blockchain_rest_api).await?;
+        let chain_id = http_api::chain_id(&endpoints.blockchain_rest_api)
```
acerone85, line 64: this gets fetched from the shared sequencer node. Is this supposed to be the same as the fuel chain id? If not, maybe best to rename the field as `ss_chain_id` or similar to avoid ambiguity

crates/services/shared-sequencer/src/service.rs
```diff
@@ -0,0 +1,269 @@
...
+        })
+    }
+}
+
+impl<S> Task<S>
+where
+    S: Signer,
+{
+    // This function is not cancel-safe because it calls `sleep` inside.
+    async fn blobs(&mut self) -> anyhow::Result<Option<SSBlobs>> {
```
acerone85, line 144: This function seems to be doing three things at once: setting the account metadata, setting the previous topic order (not sure what that is), and then taking ownership of the blobs. However, it is not clear to me how these three are related

acerone85: would it be possible/advisable to split this function in two?

crates/services/shared-sequencer/src/service.rs
```diff
@@ -0,0 +1,269 @@
...
+            blobs = self.blobs() => {
+                let blobs = match blobs {
+                    Ok(blobs) => blobs,
+                    Err(err) => return TaskNextAction::ErrorContinue(err),
+                };
+
+                // The `blobs` function is not cancel safe, as it calls sleep inside.
+                // If someone adds new logic into the `tokio::select`, please
+                // rework the `blobs` function to be cancel safe and use interval inside.
+                if let Some(blobs) = blobs {
```
acerone85, line 214: nit: 
```rust
              let Some(blobs) = blobs  else { TaskNextAction::Continue } 
```
seems to be slightly more readable

crates/services/shared-sequencer/src/lib.rs
```diff
@@ -0,0 +1,265 @@
...
+    pub async fn send_raw<S: Signer>(
+        &self,
+        timeout_height: u32,
+        signer: &S,
+        account: AccountMetadata,
+        order: u64,
+        topic: [u8; 32],
+        data: Bytes,
+    ) -> anyhow::Result<()> {
+        // We want to estimate the transaction to know hat amount and fee to use.
```
acerone85, line 168:
```suggestion
        // We want to estimate the transaction to know what amount and fee to use.
```

crates/fuel-core/src/service.rs
```diff
@@ -519,6 +519,10 @@ mod tests {
             // p2p & sync
             expected_services += 2;
         }
+        #[cfg(feature = "shared-sequencer")]
+        {
+            expected_services += 1;
+        }
```
acerone85, line 525: not for this PR, but maybe it's better here to have an enum that lists all the expected services, with feature-specific variants, and then use `strum_enum::count` to retrieve the expected services.

acerone85:
I have reviewed the whole code. 
My understanding is that the shared sequencer service keeps monitoring for new ImportedBlocks from the BlockImporter and posts the corresponding blobs to the sequencer node. 

From the architectural point of view I would like to understand the following (note that my questions probably spill into the shared sequencer domain): 
1. What's the high level architecture of the shared sequencer? I understand that nodes with shared sequencer integration enabled will connect  and post imported blocks to a ss node, and then these will be responsible to agree that they imported the same blocks using tendermint. But I'm trying connect dots and making assumptions here. Is my understanding correct? 
2. What is validated by the consensus layer? Is it just the block content? 
3. How is the result of the consensus used? In particular, I'd expect each tendermint iterations to produce a quorum certificate: how is it used? Is it posted to L1?

acerone85 (approved):

ci_checks.sh
Dentosal: A mistake. Removed in b3ea6058e1.

crates/services/shared-sequencer/src/lib.rs
Dentosal: Agreed. Renamed in b3ea6058e1.

crates/services/shared-sequencer/src/service.rs
Dentosal: Refactored in b3ea6058e1

crates/services/shared-sequencer/src/service.rs
Dentosal: b3ea6058e1

crates/services/shared-sequencer/src/lib.rs
Dentosal: b3ea6058e1

acerone85 (dismissed):

crates/services/shared-sequencer/src/service.rs
```diff
@@ -0,0 +1,250 @@
...
+                    let mut lock = self.blobs.lock().await;
+                    core::mem::take(&mut *lock)
+                };
+                if blobs.is_empty() {
+                    tokio::time::sleep(Duration::from_secs(1)).await;
+                    return TaskNextAction::Continue;
+                };
+
+                let mut account = self.account_metadata.take().expect("Account metadata is not set");
+                let next_order = self.prev_order.map(|prev| prev.wrapping_add(1)).unwrap_or(0);
```
rafal-ch, line 209: It seems that `ensure_prev_order` doesn't really ensure that we have the "prev_order", because the `http_api::get_topic` can still return `None`.

It's probably fine to start at `0` in this case, but I'm just leaving the comment for others to double-check :)

crates/services/shared-sequencer/src/service.rs
```diff
@@ -0,0 +1,250 @@
...
+                let ss =  self.shared_sequencer_client
+                    .as_ref().expect("Shared sequencer client is not set");
+                let blobs_bytes = postcard::to_allocvec(&blobs).expect("Failed to serialize SSBlob");
+
+                if let Err(err) = ss.send(self.signer.as_ref(), account, next_order, blobs_bytes).await {
+                    return TaskNextAction::ErrorContinue(err);
+                }
+
+                tracing::info!("Posted block to shared sequencer {blobs:?}");
```
rafal-ch, line 218: How many blobs we can potentially have in the collection? Asking because we may risk clogging the log with super long messages on `info` level. Maybe logging the `.len()` would suffice? Or something like range: "blobs.first().height, blobs.last().height"?

crates/services/shared-sequencer/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+        order: u64,
+        blob: Vec<u8>,
+    ) -> anyhow::Result<()> {
+        let latest_height = self.latest_block_height().await?;
+
+        self.send_raw(
+            // We don't want our transactions to be stay in the mempool for a long time,
+            // so we set the timeout height to be 64 blocks ahead of the latest block height.
+            // The 64 is an arbitrary number.
+            latest_height.saturating_add(64),
```
rafal-ch, line 147: Do we know at what pace the sequencer chain produces blocks?

crates/services/shared-sequencer/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+
+        let sign_doc_bytes = sign_doc
+            .clone()
+            .into_bytes()
+            .map_err(|err| anyhow!("{err:?}"))?;
+        let signature = signer.sign(&sign_doc_bytes).await?;
+
+        // Convert the signature to non-normalized form
+        let mut signature_bytes = *signature;
+        signature_bytes[32] &= 0x7f;
```
rafal-ch, line 260: A nit: It could be a little bit more clear if we have the `fn non_normalized()` added to `impl Signature`.

crates/services/shared-sequencer/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+        let ss_config = http_api::config(&endpoints.blockchain_rest_api).await?;
+
+        let mut minimum_gas_price = ss_config.minimum_gas_price;
+
+        if let Some(index) = minimum_gas_price.find('.') {
+            minimum_gas_price.truncate(index);
+        }
+        let gas_price: u128 = minimum_gas_price.parse()?;
+        // Ceil the gas price to the next integer.
+        let gas_price = gas_price.saturating_add(1);
```
rafal-ch, line 77: Do we have a rationale for this piece of code. I was wondering:
1. why we truncate? (and how can we be sure that _dot_ will be used as decimal separator)?
2. why we ceil to the next integer?

crates/services/shared-sequencer/src/service.rs
```diff
@@ -0,0 +1,250 @@
...
+    fn shared_data(&self) -> Self::SharedData {
+        EmptyShared
+    }
+
+    async fn into_task(
+        mut self,
+        _: &StateWatcher,
+        _: Self::TaskParams,
+    ) -> anyhow::Result<Self::Task> {
+        let shared_sequencer_client = if let Some(endpoints) = &self.config.endpoints {
```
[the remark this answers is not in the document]
xgreenx: We want to connect to SS network, to get the initial information about it(and verify that we can connect to it), but maybe posting right now is not available because of empty balance, or maybe we want to check that URL are valid but do posting from another node. Many reasons from the deployment perspective.

crates/services/shared-sequencer/src/service.rs
xgreenx: If we use 0 when it shouldn't be zero, SS will return an error and we will re-fetch the `pred_order`

crates/services/shared-sequencer/src/lib.rs
Dentosal: Once per second, IIRC

crates/services/shared-sequencer/src/lib.rs
Dentosal: Yep. This has to be done in  `fuel-vm`, though. I'll open an issue: https://github.com/FuelLabs/fuel-vm/issues/886

crates/services/shared-sequencer/src/lib.rs
Dentosal: We need an integer for this. Yep, dot is the separator here. Ceil makes sure that the number is always at least large enough.

crates/services/shared-sequencer/src/service.rs
Dentosal: My understanding is that the blobs posted here aren't necessarily adjacent, so logging only len or first and last isn't sufficient. In any case `SSBlob` is quite small.

```rust
pub struct SSBlob {
    block_height: BlockHeight,
    block_id: BlockId,
}
```

AurelienFT (approved):

ci_checks.sh
AurelienFT: Huh I think you changed an other line instead of removing this change.

AurelienFT (changes requested):

ci_checks.sh
Dentosal: Not sure what happeded here. Well, now I forced the file to match latest `master` in a0f34f68f6

AurelienFT (approved):

crates/services/shared-sequencer/src/lib.rs
xgreenx: SS network produces block each 6 seconds

xgreenx (approved):

AurelienFT (approved):

PR #1934, opened 2024-06-04
Use `DependentCost` for `aloc` opcode

xgreenx:
PR upgrades the `fuel-vm` to `0.52.0`. A new VM changes `aloc` opcode to be `DependentCost`. It requires updated benchmarks.

The PR also updates default gas costs according to a new Vultr instance we use to do benchmarks. It works faster than the previous one, so all opcodes became cheaper.

The `vm_initialziation` became much cheaper and now has reasonable values. The `bloc` opcode became expensive and depended on a number of allocated bytes.

## Checklist

### Before requesting review

- [x] Both SDKs


crates/client/src/client/schema/chain.rs
```diff
@@ -324,10 +324,9 @@ impl TryFrom<GasCosts> for fuel_core_types::fuel_tx::GasCosts {
     fn try_from(value: GasCosts) -> Result<Self, Self::Error> {
         match value.version {
             GasCostsVersion::V1 => {
```
Voxelot, line 326: This doesn't seem to match, it still says V1 but it's using V2 internally.

crates/client/src/client/schema/snapshots/fuel_core_client__client__schema__chain__tests__chain_gql_query_output.snap
```diff
@@ -82,7 +82,6 @@ query {
         version
         add
         addi
-        aloc
```
Voxelot, line 85: could this break older SDK versions loading the gas costs for predicate estimation?

benches/benches/vm_initialization.rs
```diff
@@ -110,56 +100,14 @@ pub fn vm_initialization(c: &mut Criterion) {
         group.throughput(Throughput::Bytes(tx_size as u64));
         group.bench_function(name, |b| {
             b.iter(|| {
-                unoptimized_vm_initialization_with_allocating_full_range_of_memory(
-                    &mut vm, &tx,
+                #[allow(clippy::unit_arg)]
+                black_box(
```
Dentosal, line 104: Does this `black_box` actually do anything?

crates/client/src/client/schema/chain.rs
xgreenx: GraphQL version doesn't need to match the Rust version of the type. It was done to maintain backward compatibility. Because SDKs will throw an error if the version is unknown.

crates/fuel-core/src/schema/chain.rs
```diff
@@ -338,7 +338,7 @@ impl GasCosts {
     }
 
     async fn aloc(&self) -> U64 {
```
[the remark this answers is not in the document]
xgreenx: Yep, the only problem is that if Rust SDK is old and connected to `fuel-core 0.27.0`, the predicate estimation can be incorrect if it uses an `alco` opcode. But it should be addressed on Rust SDK side: https://github.com/FuelLabs/fuels-rs/issues/1409

benches/benches/vm_initialization.rs
xgreenx: Removing `black_box` improves speed 
image

Cargo.toml
```diff
@@ -81,7 +81,7 @@ fuel-core-wasm-executor = { version = "0.27.0", path = "./crates/services/upgrad
 fuel-core-xtask = { version = "0.0.0", path = "./xtask" }
 
 # Fuel dependencies
-fuel-vm-private = { version = "0.51.0", package = "fuel-vm", default-features = false }
+fuel-vm-private = { version = "0.52.0", package = "fuel-vm", default-features = false }
```
MitchTurner, line 84: The changelog says your updating to `0.51.0`, not `0.52.0`.

crates/client/src/client/schema/chain.rs
MitchTurner: Will it get confusing if the versions start diverging a lot over time?

Cargo.toml
MitchTurner:
```
[#1888](https://github.com/FuelLabs/fuel-core/pull/1888): Upgraded `fuel-vm` to `0.51.0`. See [release](https://github.com/FuelLabs/fuel-vm/releases/tag/v0.51.0) for more information.
```

xgreenx: The changelog related for `0.51.0` is from another PR, I just grouped everything together=)

crates/client/src/client/schema/chain.rs
xgreenx: When it will diverging a lot, we just can jump straight to the highest=) I was done for backward compatibility until SDKs will handle it gracefully.

Cargo.toml
MitchTurner: seems like we are being inconsistent about when we note these things, but I don't know that it matters

MitchTurner (approved):

PR #1938, opened 2024-06-05
Add Gas Price Updater Service

MitchTurner:
Closes: https://github.com/FuelLabs/fuel-core/issues/1956

This is a subtask of https://github.com/FuelLabs/fuel-core/issues/1624

In this PR we add the generic service that will post an algorithm for the providers to use. Not bothering with a _real_ algorithm, that will be implemented later. For now, just show that the provider can get the value generated by the service.

## Checklist

### Before requesting review

- [ ] [Swhttps://github.com/FuelLabs/fuel-core/issues/1956ay compiler](https://github.com/FuelLabs/sway/)
- [ ] Someone else?


Cargo.toml
```diff
@@ -79,6 +80,7 @@ fuel-core-tests = { version = "0.0.0", path = "./tests" }
 fuel-core-upgradable-executor = { version = "0.26.0", path = "./crates/services/upgradable-executor" }
 fuel-core-wasm-executor = { version = "0.26.0", path = "./crates/services/upgradable-executor/wasm-executor", default-features = false }
 fuel-core-xtask = { version = "0.0.0", path = "./xtask" }
+gas_price_service = { path = "crates/services/gas_price_service" }
```
xgreenx, line 83: Could you also specify a `version` please?=)

crates/services/gas_price_service/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "gas_price_service"
```
xgreenx, line 2:
```suggestion
name = "fuel-core-gas-price-provider"
```
```suggestion
name = "gas_price_service"
```

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -160,14 +178,21 @@ pub fn init_sub_services(
     #[cfg(not(feature = "p2p"))]
     let p2p_adapter = P2PAdapter::new();
 
-    let gas_price_provider = StaticGasPrice::new(config.static_gas_price);
+    let old_gas_price_provider = StaticGasPrice::new(config.static_gas_price);
+
+    let update_algo = StaticAlgorithmUpdater::new(config.static_gas_price);
+    let gas_price_task = GasPriceService::new(last_height, update_algo);
+    let next_algo = gas_price_task.next_block_algorithm();
+    let gas_price_service = ServiceRunner::new(gas_price_task);
```
xgreenx, line 186: It would be nice if you followed pattern from other services. Usually each service defined `new_service` function that is doing everything inside. If you need access to shared state, you can call `let also = service.shared_state()`.

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -160,14 +178,21 @@ pub fn init_sub_services(
     #[cfg(not(feature = "p2p"))]
     let p2p_adapter = P2PAdapter::new();
 
-    let gas_price_provider = StaticGasPrice::new(config.static_gas_price);
+    let old_gas_price_provider = StaticGasPrice::new(config.static_gas_price);
```
xgreenx, line 181: It seems we should be able to remove this=)

crates/services/gas_price_service/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+}
+
+#[async_trait]
+impl<A, U> RunnableService for GasPriceService<A, U>
+where
+    U: UpdateAlgorithm<Algorithm = A> + Send,
+    A: Send + Sync,
+{
+    const NAME: &'static str = "GasPriceUpdater";
+    type SharedData = Arc<RwLock<(BlockHeight, A)>>;
```
xgreenx, line 88: It would be nice if we had a separate struct maybe `struct SharedState`( or any other name=)) which contains all shared fields.

crates/services/gas_price_service/src/static_updater.rs
```diff
@@ -0,0 +1,40 @@
...
+
+#[async_trait]
+impl UpdateAlgorithm for StaticAlgorithmUpdater {
+    type Algorithm = StaticAlgorithm;
+
+    fn start(&self, _for_block: BlockHeight) -> Self::Algorithm {
+        StaticAlgorithm::new(self.static_price)
+    }
+
+    async fn next(&mut self, _for_block: BlockHeight) -> Self::Algorithm {
```
xgreenx, line 37: Maybe because we don't use `_for_block` it is better to remove for now and add then we really need it

crates/fuel-core/src/service/adapters/fuel_gas_price_provider.rs
```diff
@@ -0,0 +1,78 @@
...
+            Ordering::Greater => Err(Error::AlgorithmNotUpToDate {
+                requested_height: block_height,
+                latest_height: height,
+            }),
+            Ordering::Less => Err(Error::RequestedOldBlockHeight {
+                requested_height: block_height,
+                latest_height: height,
+            }),
+        }
+    }
```
xgreenx, line 67: It sounds like the responsibility of the gas price provider itself.

crates/services/gas_price_service/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+}
+
+#[async_trait]
+pub trait UpdateAlgorithm {
+    type Algorithm;
+
+    fn start(&self, for_block: BlockHeight) -> Self::Algorithm;
+    async fn next(&mut self, for_block: BlockHeight) -> Self::Algorithm;
+}
```
xgreenx, line 51: I think the algorithm itself should not be dependent on the block height. It has an internal state and `next` and `update` functions to iterate over this state.

The handling of the `for_block` should be done on the service level, since `service` knows what is the current block height and when the next block is imported.

crates/fuel-core/src/service/sub_services.rs
MitchTurner: Yes. The new provider isn't implemented for the tx pool and graphql ports yet.

crates/services/gas_price_service/src/lib.rs
MitchTurner: Yeah. It would be nice, if only to have getters etc. The provider should be able to depend on the service types without breaking dependency inversion. So I can do that.

crates/services/gas_price_service/src/lib.rs
MitchTurner: How does the producer know that the gas price service is on the same block as itself though? I feel like there needs to be a check somewhere that the producer is getting the algorithm for the block it's building.

MitchTurner:
There is a bug in the `StaticAlgorithmUpdater` because there is nothing stopping it from increasing the block height.

I guess I'll remove the block height from the `BlockGasPriceAlgorithm` for now because we aren't using it anywhere and any fix to the bug would be unneeded work.

crates/services/gas_price_service/src/lib.rs
MitchTurner: I've removed it for a separate reason. It works right now, but we might need to reconsider later.

crates/services/gas_price_service/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+    U: UpdateAlgorithm<Algorithm = A>,
+    A: Send + Sync,
+{
+    async fn update(&mut self, new_algorithm: A) {
+        self.next_block_algorithm.update(new_algorithm).await;
+    }
+}
+
+#[derive(Debug)]
+pub struct BlockGasPriceAlgo<A>(Arc<RwLock<A>>);
```
Dentosal, line 73: This wrapper type seems unnecessary and a bit misleading. It's just `Arc<RwLock<_>>`, and contains no gas price -related functionality by itself.

crates/services/gas_price_service/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "fuel-core-gas-price-service"
+description = "Service for updating the gas price algorithm"
+version = { workspace = true }
+authors = { workspace = true }
+edition = { workspace = true }
+homepage = { workspace = true }
+license = { workspace = true }
+repository = { workspace = true }
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
Dentosal, line 10:
```suggestion
```

crates/services/gas_price_service/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+        }
+    }
+
+    pub fn next_block_algorithm(&self) -> BlockGasPriceAlgo<A> {
+        self.next_block_algorithm.clone()
+    }
+}
+
+#[async_trait]
+pub trait UpdateAlgorithm {
```
Dentosal, line 55: Missing docs. What does this trait exactly do?

crates/services/gas_price_service/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+    A: Send + Sync,
+{
+    let service = GasPriceService::new(current_fuel_block_height, update_algo);
+    Ok(ServiceRunner::new(service))
+}
+
+pub struct GasPriceService<A, U> {
+    next_block_algorithm: BlockGasPriceAlgo<A>,
+    update_algorithm: U,
+}
```
Dentosal, line 33: Missing struct and field docs. What they mean isn't immediately clear to me.

crates/types/src/services/txpool.rs
```diff
@@ -284,7 +284,7 @@ pub fn from_executor_to_status(
 #[non_exhaustive]
 pub enum Error {
     #[error("Gas price not found for block height {0}")]
-    GasPriceNotFound(BlockHeight),
+    GasPriceNotFound(String),
```
Dentosal, line 287: The error message looks wrong when given non-`BlockHeight` values as you do in `crates/services/txpool/src/txpool.rs`.

Dentosal: Consider introducing new error types for different errors as well, maybe `"Failed to calculated max block bytes"` should be a separate variant, for instance.

crates/services/txpool/src/ports.rs
```diff
@@ -61,10 +61,14 @@ pub trait TxPoolDb: Send + Sync {
     fn message(&self, message_id: &Nonce) -> StorageResult<Option<Message>>;
 }
 
+#[async_trait::async_trait]
 /// Trait for getting gas price for the Tx Pool code to look up the gas price for a given block height
 pub trait GasPriceProvider {
     /// Get gas price for specific block height if it is known
```
Dentosal, line 67: Doc seems outdated now

crates/services/gas_price_service/src/lib.rs
Dentosal: If I read it correctly, it looks like we should probably use tokio channels for the updating instead of a custom trait.

crates/services/gas_price_service/src/lib.rs
MitchTurner: The name `Block` is outdated. I'll change that.

Maybe `SharedGasPriceAlgo`.

Ideally, we'd have two types:
`GasPriceAlgoWriter` and `GasPriceAlgoReader` and then protect them so that mutability is only controlled by the service, but our services are over-constrained with the `SharedData` assoc type, so it kinda makes us use the same type here and on the provider. At least `update` is `&mut self` 🤷

crates/services/gas_price_service/src/lib.rs
MitchTurner: This allows us to customize the behavior of `next`. In our case, we are going to be watching both l2 block commits as well as da recordings. If either of those things happen, we want to update to the `next` gas price algorithm.

Dentosal: Ok that makes sense.

crates/services/gas_price_service/src/lib.rs
Dentosal: I think I understand the tradeoff here. I don't think the layer of mental indirection is worth it. Since this disagreement seems to be a just matter of taste, I'll just approve it.

Dentosal (dismissed):

crates/fuel-core/src/service/adapters/fuel_gas_price_provider/ports.rs
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    pub fn used(&self) -> u64 {
+        self.used
+    }
+
+    pub fn capacity(&self) -> u64 {
+        self.capacity
+    }
+}
```
xgreenx, line 21: It seems is not used anywhere

crates/fuel-core/src/service/adapters/txpool.rs
```diff
@@ -141,9 +141,10 @@ impl fuel_core_txpool::ports::TxPoolDb for Database {
     }
 }
 
+#[async_trait::async_trait]
 impl GasPriceProvider for StaticGasPrice {
-    fn gas_price(&self, _block_height: BlockHeight) -> Option<u64> {
-        Some(self.gas_price)
+    async fn gas_price(&self, _block_bytes: u64) -> TxPoolResult<u64> {
```
xgreenx, line 146: Hmm, for TxPool I think it is okay to use the last gas price(we don't need to predict it).

crates/services/gas_price_service/src/lib.rs
```diff
@@ -0,0 +1,234 @@
+use async_trait::async_trait;
```
xgreenx, line 1:
```suggestion
#![deny(clippy::arithmetic_side_effects)]
#![deny(clippy::cast_possible_truncation)]
#![deny(unused_crate_dependencies)]
#![deny(warnings)]

use async_trait::async_trait;
```

crates/services/gas_price_service/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+    let service = GasPriceService::new(current_fuel_block_height, update_algo);
+    Ok(ServiceRunner::new(service))
+}
+
+/// The service that updates the gas price algorithm.
+pub struct GasPriceService<A, U> {
+    /// The algorithm that can be used in the next block
+    next_block_algorithm: SharedGasPriceAlgo<A>,
+    /// The code that is run to update your specific algorithm
+    update_algorithm: U,
```
xgreenx, line 35: For me, it looks like the updater and algorithm should be one entity since only the algorithm knows how it works inside and what data it requires.

But we can leave it for now as it is and see how it will look like with a nonstatic algorithm.

crates/services/gas_price_service/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+
+        // given
+        let (price_sender, price_receiver) = mpsc::channel(1);
+        let updater = TestAlgorithmUpdater {
+            start: TestAlgorithm { price: 0 },
+            price_source: price_receiver,
+        };
+        let service = GasPriceService::new(0.into(), updater);
+        let (watch_sender, watch_receiver) = tokio::sync::watch::channel(State::Started);
+        let watcher = StateWatcher::from(watch_receiver);
```
xgreenx, line 217: You can re-use helper function 
image

crates/services/producer/src/block_producer.rs
```diff
@@ -219,9 +230,23 @@ where
...
+        let gas_per_bytes = consensus_params.fee_params().gas_per_byte();
+        let max_gas_per_block = consensus_params.block_gas_limit();
+        let max_block_bytes = max_gas_per_block
+            .checked_div(gas_per_bytes)
+            .ok_or(anyhow!("Failed to calculated max block bytes"))?;
+
+        let gas_price = if let Some(inner) = gas_price {
+            inner
+        } else {
+            self.gas_price_provider.gas_price(max_block_bytes).await?
```
xgreenx, line 248: It is better to move all calculations here=) Could you move the duplicated code into a small function that we could reuse in both places, please?=)

Plus, maybe for `dry_run`, we could just use the previous gas price without.

crates/services/txpool/src/ports.rs
```diff
@@ -61,10 +61,14 @@ pub trait TxPoolDb: Send + Sync {
...
+#[async_trait::async_trait]
 /// Trait for getting gas price for the Tx Pool code to look up the gas price for a given block height
 pub trait GasPriceProvider {
-    /// Get gas price for specific block height if it is known
-    fn gas_price(&self, block_height: BlockHeight) -> Option<GasPrice>;
+    /// Calculate gas price for the next block with a given size `block_bytes`.
+    async fn gas_price(
+        &self,
+        block_bytes: u64,
+    ) -> fuel_core_types::services::txpool::Result<GasPrice>;
```
xgreenx, line 71:
```suggestion
    ) -> TxPoolResult<GasPrice>;
```

crates/services/gas_price_service/src/lib.rs
MitchTurner: I'm leaning to this approach so that the consumer of the algo doesn't need to have all the generics for the updater to read the values.

crates/fuel-core/src/service/adapters/producer.rs
```diff
@@ -224,9 +223,10 @@ impl fuel_core_producer::ports::BlockProducerDatabase for Database {
     }
 }
 
+#[async_trait::async_trait]
 impl GasPriceProvider for StaticGasPrice {
-    fn gas_price(&self, _block_height: GasPriceParams) -> Option<u64> {
-        Some(self.gas_price)
+    async fn gas_price(&self, _block_bytes: u64) -> anyhow::Result<u64> {
```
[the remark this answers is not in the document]
MitchTurner: It's not really predicting. It's the value we are going to use. We don't adjust it after getting it here.

crates/services/gas_price_service/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+
+impl<A> SharedGasPriceAlgo<A>
+where
+    A: Send + Sync,
+{
+    pub fn new(algo: A) -> Self {
+        Self(Arc::new(RwLock::new(algo)))
+    }
+
+    pub async fn read(&self) -> RwLockReadGuard<A> {
```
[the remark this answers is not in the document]
MitchTurner: Yeah. You're right. I don't like it just for code cleanliness either.

To add an `execute` and `estimate` I'll have to constrain `A` somehow with a new trait.

The other approach would be just make `A` `Clone`, and clone it out of the lock.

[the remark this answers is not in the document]
xgreenx: I think it is okay to add constrains for the algorithm=) Gas provider expects that algorithm will provide next gas price=D

crates/services/txpool/src/ports.rs
```diff
@@ -61,10 +61,12 @@ pub trait TxPoolDb: Send + Sync {
...
 }
 
+#[async_trait::async_trait]
 /// Trait for getting gas price for the Tx Pool code to look up the gas price for a given block height
 pub trait GasPriceProvider {
-    /// Get gas price for specific block height if it is known
-    fn gas_price(&self, block_height: BlockHeight) -> Option<GasPrice>;
+    /// Calculate gas price for the next block with a given size `block_bytes`.
+    async fn last_gas_price(&self)
+        -> fuel_core_types::services::txpool::Result<GasPrice>;
```
xgreenx, line 69:
```suggestion
    async fn last_gas_price(&self) -> TxPoolResult<GasPrice>;
```
```suggestion
    async fn last_gas_price(&self)
        -> fuel_core_types::services::txpool::Result<GasPrice>;
```

crates/services/producer/src/block_producer.rs
```diff
@@ -145,6 +142,25 @@ where
...
     }
+
+    async fn calculate_gas_price(
+        &self,
+        header: &PartialBlockHeader,
+    ) -> anyhow::Result<u64> {
+        let consensus_params = self
+            .consensus_parameters_provider
+            .consensus_params_at_version(&header.consensus_parameters_version)
+            .unwrap();
```
xgreenx, line 153:
```suggestion
        let consensus_params = self
            .consensus_parameters_provider
            .consensus_params_at_version(&header.consensus_parameters_version)?;
```

xgreenx (commented):
LGTM!=)

xgreenx (approved):

PR #1942, opened 2024-06-06
Sequential relayer's commits

xgreenx:
Updated relayer to do sequential commits. This means that if we don't have any DA events for height `X`, we will still insert empty arrays. In most cases, almost every block on DA will contain at least one event, so overhead will be small and only affect the start of the network.

### Before requesting review


tests/test-helpers/src/builder.rs
```diff
@@ -210,7 +211,7 @@ impl TestSetupBuilder {
 
         let latest_block = self.starting_block.map(|starting_block| LastBlockConfig {
             block_height: starting_block,
-            state_transition_version: 0,
+            state_transition_version: LATEST_STATE_TRANSITION_VERSION - 1,
```
MitchTurner, line 214: Is this related to the other changes? Why are we pointing to the "old" version here?

MitchTurner (commented):
A small question, but other than that it LGTM.

Why do we care if there are some missing blocks? Is this solving for a bug? This isn't connected to an issue.

Voxelot (dismissed):

tests/test-helpers/src/builder.rs
xgreenx: Yeah, it is not related to the change, just saw in the codebase=D We want to use previous version because after regenesis we want to use a native executor just to speed up tests.

xgreenx:
> Why do we care if there are some missing blocks? Is this solving for a bug? This isn't connected to an issue.

We just want to call the `commit` function for each separate block to grow blocks one by one and make them connected to each other. It is used by the state rewind feature to create a view of the database at each block height.

PR #1943, opened 2024-06-06
Fix code coverage compilation and tests

Dentosal:
The compilation was failing on the latest nightly. Most imporantly, it caught one test that was incorrectly disabled due to incorrect `cfg`-gating. Found in preparation for code coverage setup. After this PR, the following command should work for getting codecov:

```bash
cargo +nightly-2024-06-05 llvm-cov --html --branch --all-features
```

## Checklist

### Before requesting review


crates/fuel-core/src/database/state.rs
```diff
@@ -231,29 +232,23 @@ mod tests {
...
-            let original_state = state_groups
+            let mut original_state = state_groups
                 .into_iter()
                 .flatten()
-                .sorted()
+                .map(|v| v.value)
                 .collect::<Vec<_>>();
 
+            states_in_db.sort_by(|a, b| a.as_ref().cmp(b.as_ref()));
+            original_state.sort_by(|a, b| a.as_ref().cmp(b.as_ref()));
```
xgreenx, line 250: Before we were comparing `ContractStateConfig`, now we only compare values. Why do we want to change it?

Dentosal: 751c407b0 Now back to comparing both keys and values

crates/fuel-core/src/database/state.rs
```diff
@@ -231,29 +232,26 @@ mod tests {
...
-                .sorted()
+                .map(|v| (v.key, v.value))
                 .collect::<Vec<_>>();
 
+            states_in_db.sort_by(|(ak, av), (bk, bv)| {
+                (ak.as_ref(), av.as_ref()).cmp(&(bk.as_ref(), bv.as_ref()))
+            });
+            original_state.sort_by(|(ak, av), (bk, bv)| {
+                (ak.as_ref(), av.as_ref()).cmp(&(bk.as_ref(), bv.as_ref()))
+            });
```
xgreenx, line 253: The previous code was much simpler, why we don't want to compare `ContractStateConfig`?

tests/tests/poa.rs
```diff
@@ -110,7 +110,7 @@ mod p2p {
     // after the first_producer stops, second_producer should start producing blocks
     #[tokio::test(flavor = "multi_thread")]
     async fn test_poa_multiple_producers() {
-        const SYNC_TIMEOUT: u64 = 30;
+        const SYNC_TIMEOUT: u64 = 120;
```
xgreenx, line 113: Why do we need to do this? The test should pass in 10 seconds, we set 30 just because of CI. Why 120?

crates/services/upgradable-executor/src/lib.rs
```diff
@@ -3,6 +3,13 @@
 #![deny(unused_crate_dependencies)]
 #![deny(warnings)]
 
+#[cfg(test)]
+mod _ignore {
+    #![allow(unexpected_cfgs)] // for cfg(coverage)
+    #[cfg(coverage)]
+    use ntest as _; // Only used outside cdg(coverage)
```
xgreenx, line 10: You can move it to `executor.rs`
image

Cargo.lock
```diff
@@ -150,9 +150,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle-query"
-version = "1.0.3"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a64c907d4e79225ac72e2a354c9ce84d50ebb4586dee56c82b3ee73004f537f5"
```
xgreenx, line 155: Do we need to bump version or we can leave it for a separate PR?

tests/tests/poa.rs
Dentosal: It's too slow in branch coverage mode. I'll just disable it.

Cargo.lock
Dentosal: Apparently not. I had issues with an old `ahash` version, but cargo clean and reinstall was enough.

.cargo/config.toml
Dentosal, line 1: The filename without `.toml` is deprecated

Cargo.lock
Dentosal: Removed the changes in 1c6f607e79dd0c8dbbee8c8610bb949c0bd45b6b

crates/services/upgradable-executor/src/lib.rs
Dentosal: Sure, done in 1c6f607e79dd0c8dbbee8c8610bb949c0bd45b6b

tests/tests/poa.rs
Dentosal: 1c6f607e79dd0c8dbbee8c8610bb949c0bd45b6b

crates/fuel-core/src/database/state.rs
Dentosal: The old version didn't compile, so it was quite easy for it to be simple. I've tried to improve this a bit in 1c6f607e79dd0c8.

xgreenx (approved):

Issue #1825, opened 2024-04-12
transactions() on beta 5 has some issues with cursors.


andy-thomason:
Hi Everyone. I'm enjoying exploring the codebase and doing some experiments with the public beta 5 endpoint.

I don't have a particular template to work from, so I'm making some assumptions here:

Given

```rust
#[tokio::main]
async fn main() -> anyhow::Result<()> {
    let client = FuelClient::new("https://beta-5.fuel.network/graphql")?;

    get_transactions(&client, 0).await?;

    Ok(())
}

async fn get_transactions(client: &FuelClient, start_block: u32) -> anyhow::Result<()> {
    use fuel_core_client::client::types::primitives::TransactionId;
    let tx = TransactionId::zeroed();
    let mut cursor = (start_block != 0).then_some(format!("{start_block}#0x{tx}"));
    let results = 10;
    for _ in 0.. {
        let req = PaginationRequest {cursor, results, direction: PageDirection::Forward };
        let txs = client.transactions(req).await?;

        cursor = txs.cursor;

        for tx in txs.results {
            println!("{tx:#?}");
        }
    }
    Ok(())
}
```

Two things go wrong.

1) If we initialise with a cursor with a form of *block*#0x0000...0000 then the endpoint hangs.
2) If we initialise with a cursor of None, the next cursor has the block number formatted in hex
     such as `Some("0000000a#0x49c6fee039066353084bdf8585930bba4cd9e89433c2abce3f8c7c35c9b8ac4e")`
    which fails because the endpoint expects a decimal number.
    
 Am I being an idiot, or should these two expectations be met?

andy-thomason:
I'm using the `release/v0.22.1` fuel-core.

Voxelot:
Nice finds! These are real bugs, the SortedTxCursor decodes block height using `u32::from_str(block_height)` but the display impl for BlockHeight does indeed use `<Self as fmt::LowerHex>::fmt(&self, f)`.

`BlockHeight` used to be a type-alias for u32, and this bug must've slipped in when we migrated to using a dedicated newtype for BlockHeight.

hummusonrails:
Hi @andy-thomason 👋,

Thank you for taking the initiative to open this bug report in `fuel-core`! You taking the time to report back on the bugs you found shows a lot about your commitment to open source and your contribution is highly valued! 🌟

[...]

PR #1950, opened 2024-06-09
Fix cursor block height decoding in SortedTXCursor

AurelienFT:
https://github.com/FuelLabs/fuel-core/issues/1825

The `BlockHeight` was encoded in hexadecimal in response of `transactions` endpoint and decoded with `u32::from_str` in parameter of type `SortedTXCursor` which was preventing from making chained API calls.
This PR, fixes this by decoding `BlockHeight` from hexadecimal form in `SortedTXCursor`.

This bug wasn't catch by the test `tx::get_transactions()` because the test was using only txs included in block 3 that has a string representation that is the same in an hexadecimal and decimal encoding. I added the creation of 10 blocks at start of the test so that now the txs are included in blocks that have a hexadecimal and decimal encoding form that differs.

I didn't spotted any hangs on the endpoint when the cursor is initialized manually, however as we can't ask to start at a precise block height without specifying a previous `tx_id` it's not very usable to make manual search.

PS: This is my first contribution and so to what I saw my PR doesn't brake anything else and I tried to follow the guidelines of the repository, however maybe their things that I missed and so I'm very open to criticise !

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- [#1950](https://github.com/FuelLabs/fuel-core/pull/1950): Fix cursor `BlockHeight` encoding in `SortedTXCursor`
+
 ## [Version 0.28.0]
 
 ### Changed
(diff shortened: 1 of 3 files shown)


Voxelot (approved):
LGTM - Thank you!

PR #1974, opened 2024-06-15
Optimized the work of `InMemoryTransaction` for lookups and empty insertion

xgreenx:
Just was walking around this code and noticed that we can remove not needed `to_vec` and improve the case when we commit changes into empty `InMemoryTransaction`.

### Before requesting review


crates/storage/src/transactional.rs
```diff
@@ -233,10 +307,9 @@ where
     S: KeyValueInspect<Column = Column>,
 {
     fn get_from_changes(&self, key: &[u8], column: Column) -> Option<&WriteOperation> {
-        let k = key.to_vec();
         self.changes
             .get(&column.id())
-            .and_then(|btree| btree.get(&k))
+            .and_then(|btree| btree.get(key))
```
xgreenx, line 312: We use property that collections work with `Q` generic. The `ReferenceBytesKey` implements `Borrow<[u8]>` that allows usage of `&[u8]` for lookups. It is very important for `InMemoryTransaction` since `get` is mostly used method=)


image

crates/storage/src/transactional.rs
```diff
@@ -198,7 +265,14 @@ where
 impl<Storage> Modifiable for InMemoryTransaction<Storage> {
     fn commit_changes(&mut self, changes: Changes) -> StorageResult<()> {
         for (column, value) in changes.into_iter() {
-            let btree = self.changes.entry(column).or_default();
+            let btree = match self.changes.entry(column) {
+                hash_map::Entry::Vacant(vacant) => {
+                    vacant.insert(value);
+                    continue
+                }
```
xgreenx, line 272: If `BTreeMap` is not exist, we just set it to be `value` because it can't have any conflicts=)

Voxelot (approved):

Issue #451, opened 2022-06-29
State Rewind


Voxelot:
The database in fuel-core needs the ability to rewind state to a previous point in time. This will allow nodes to contest invalid state transitions on L1 via IVG fraud-proofs.

Each atomic batch-write to the database is composed of either insert `(column id: u32, key: Vec<u8>, value: Vec<u8>)` or delete operations `(column id: u32, key: Vec<u8>)`. By keeping a history of these batches, we'll be able to create an in-memory overlay over the database where batches are used to rewind from the current state tip. Both insert and remove operations will now also need to include the previous value associated with the key in order to be rewound: `(column id: u32, key: Vec<u8>, value: Vec<u8>, prev_value: Vec<u8>)` & `(column id: u32, key: Vec<u8>, prev_value: Vec<u8>)`

Add a new column to the database that contains a serialized blob of these operations, bundled together by block height:
 - key: block_height
 - value: `Vec<WriteOperation>`

In order to prevent ongoing state changes (e.g. new blocks committed during an IVG session) from altering the rewind overlay, we'll also need the ability to snapshot the current database state at the start of the rewind operation. This should be fairly efficient as RocksDB can make the snapshot from symlinks as long as the snapshot is on the same filesystem as the database: http://rocksdb.org/blog/2015/11/10/use-checkpoints-for-efficient-snapshots.html

The in-memory database will also need to support state rewind, but it could just clone all the data for snapshotting since it doesn't need to be very efficient and we don't expect the in-memory db to achieve significant usage outside of short-lived integ tests.

rakita:
Blockchain database growth is the one of major problems in space that yet needs to be solved.
There is two parts here, one is state growth (how much of current UTXO we have and contract slots), and second is history growth (UTXO that were prunned, contract slots changes per block or transactions)

This is extreme example where history and every change set is saved in eth `archive` node and size goes very high: https://etherscan.io/chartsync/chainarchive
you can see that increase in database history goes in few terabytes per year.

This is data footprint of eth prunned node where historic blocks are all saved but history change is saved for (i think OE is like last ~100 block while geth has like ~1000 or something, it is configurable not sure what is set in etherscan): https://etherscan.io/chartsync/chaindefault

Current state of art database model in ETH is erigons where they managed to shrink 10tb of history model to 1-2tb: https://github.com/ledgerwatch/erigon/blob/devel/docs/programmers_guide/db_walkthrough.MD
[...]

PR #1989, opened 2024-06-26
Extract `HistoricalView` trait from the `AtomicView`

xgreenx:
It is preparation for https://github.com/FuelLabs/fuel-core/issues/451.

Most of the services don't need historical functionality; only the executor uses the view of the storage at a specific height.

The change extracts this functionality into its own `HistoricalView` trait.

During the implementation of the state rewind, I realized that the historical view has some limitations(like we are not able to iterate). So because of that, the historical view uses another type that only implements the `KeyValueInspect` trait.

## Checklist

### Before requesting review


bin/fuel-core/chainspec/local-testnet/state_transition_bytecode.wasm
Dentosal, line 1: Why is this removed?

xgreenx: It is not removed=) Because of simplification in the trait, the compiler removed 191 bytes from the bytecode. Bu the file still exists

Dentosal: image

GitHub says -100% which makes me thing the whole file is removed, but apparently this is not the case.

crates/fuel-core/src/graphql_api/view_extension.rs
```diff
@@ -37,7 +39,15 @@ impl Extension for ViewExtension {
...
     ) -> ServerResult<Request> {
         let database: &ReadDatabase = ctx.data_unchecked();
-        let view = database.view();
+        let view = database.view().map_err(|e| {
+            ServerError::new(
+                e.to_string(),
+                Some(Pos {
+                    line: line!() as usize,
+                    column: column!() as usize,
+                }),
```
Dentosal, line 48: This feels weird. Maybe it's actually supposed to work like this, but the mismatch between `line!` and `column!` evaluated separately seems wrong.

bin/fuel-core/chainspec/local-testnet/state_transition_bytecode.wasm
xgreenx: Yeah, it is the UI bug of the girhub=(

crates/storage/src/transactional.rs
```diff
@@ -33,23 +33,29 @@ use crate::{
...
-    /// Returns the view of the storage at the given `height`.
-    fn view_at(&self, height: &Self::Height) -> StorageResult<Self::View>;
-
     /// Returns the view of the storage for the latest block height.
-    fn latest_view(&self) -> Self::View;
+    fn latest_view(&self) -> StorageResult<Self::LatestView>;
+}
+
+/// Provides a view of the storage at the given height.
+pub trait HistoricalView: AtomicView {
```
Dentosal, line 53: Since `HistoricalView` is a subtrait of `AtomicView`, it'd like that the comment here (and possibly on `AtomicView` as well) would mention that `AtomicView` doesn't mean the latest version here.

crates/storage/src/transactional.rs
```diff
@@ -33,23 +33,29 @@ use crate::{
     kv_store::KVItem,
 };
 
-/// Provides a view of the storage at the given height.
+/// Provides a view of the storage at the latest height.
 /// It guarantees to be atomic, meaning the view is immutable to outside modifications.
 pub trait AtomicView: Send + Sync {
```
Dentosal, line 38: I find the wording/naming to be unclear. How can it provide a view of the latest height, but then not change when the latest height changes?

xgreenx: I also decided to move the `latest_height` to `HistoricalView` since it looks like a better place for this logic.

Dentosal (approved):

PR #1990, opened 2024-06-26
Use fresh `ReadView` for mutate queries

xgreenx:
When we manually produce blocks or `submit_and_wait` for the transaction, the `ReadView` has a view of the database at height X.

But when we return a response, it needs to use the view at height X + 1.

The change uses the latest view for mutate queries to solve this problem.

### Before requesting review


Dentosal:
Lots of conflicts with `master`.

xgreenx:
Fixed=)

crates/fuel-core/src/schema/chain.rs
```diff
@@ -816,20 +814,18 @@ impl ChainInfo {
...
 
     async fn da_height(&self, ctx: &Context<'_>) -> U64 {
-        let query: &ReadView = ctx.data_unchecked();
-
-        let height = query
-            .da_height()
-            .expect("The blockchain always should have genesis block");
+        let Ok(query) = ctx.read_view() else {
+            return 0.into();
+        };
```
Dentosal, line 826: Is this change of behavior intentional? Is returning zero correct?

xgreenx: If I change the output type to `Result`, it will change the schema, and it will be a breaking change.

Dentosal: My question is: should we rather panic or return zero? But all things considered I think zero is fine here.

Dentosal (approved):

PR #1991, opened 2024-06-26
Prepare the database to use different types than `Database` for atomic view

xgreenx:
## Change overview

The change implements the types for `AtomicView` and `HistoricalView` traits. It is only preparation for corresponding features without actual implementation.

In the past, we cloned the `Database` type and used it as a view without any atomic properties. In a new implementation, `Database` is only a view provider now, but the view itself implements all read functionality.

We have two views. Separation for `IterableView` and `KeyValueView` is needed because we can't iterate over the historical storage. We only have access by key to the values. It would be nice to only have one view that supports all functionality, but we don't need iteration functionality for the historical view(at least for now), and it is expensive to support(requires more storage or enormous computation resources).

## Implementation details

`Database`, `IterableView` and `KeyValueView` all are aliases to the `GenericDatabase`. The `GenericDatabase` implements all storage traits from the `fuel-storage` crate, deriving implementation from `StructuredStorage`. We need it reuse the storage trait implementation for all three types. They only require the inner storage to implement `KeyValueInspect` to derive implementation. 

`KeyValueViewWrapper` and `IterableViewWrapper` are wrappers around `Arc<dyn Trait>` that implement `KeyValueInspect` traits.

## Checklist

### Before requesting review


crates/fuel-core/src/service/adapters/consensus_module.rs
```diff
@@ -45,7 +48,7 @@ impl VerifierAdapter {
     }
 }
 
-impl fuel_core_poa::ports::Database for Database {
+impl fuel_core_poa::ports::Database for OnChainIterableView {
     fn block_header(&self, height: &BlockHeight) -> StorageResult<BlockHeader> {
         Ok(self.get_block(height)?.header().clone())
```
MitchTurner, line 53: I guess I'm confused why we are calling the database we use here `Iterable`. Since this doesn't really iterate.

crates/fuel-core/src/state/iterable_view.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+impl<Column> std::fmt::Debug for IterableViewWrapper<Column> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("IterableViewWrapper").finish()
+    }
+}
+
+impl<Column> IterableViewWrapper<Column> {
+    pub fn new(storage: Arc<dyn IterableStore<Column = Column> + Sync + Send>) -> Self {
```
MitchTurner, line 35: This concept of passing an `Arc<dyn Trait>` into `new` seems like a leak of the internal details. Same with returning it from `into_inner`. I'd expect `new` to take the inner trait object and wrap the `Arc` for you.

I don't know if I have a solution though. I feel like these helpers exist to solve specific situations, like converting to `KeyValueView` maybe we can just convert directly without exposing the inner `Arc`? I haven't dug into it enough to say and maybe we just need to settle with this.

crates/fuel-core/src/state/key_value_view.rs
```diff
@@ -0,0 +1,58 @@
...
+);
+
+impl<Column> std::fmt::Debug for KeyValueViewWrapper<Column> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("KeyValueViewWrapper").finish()
+    }
+}
+
+impl<Column> KeyValueViewWrapper<Column> {
+    pub fn new(storage: Arc<dyn KeyValueInspect<Column = Column> + Sync + Send>) -> Self {
```
MitchTurner, line 23: Same thing here.

MitchTurner (commented):
I don't see any big problems. Seems like generally the same structure in terms of architectural hierarchy, which is mostly what I'm concerned with wrt architecture.

I'm not sure I'm a fan of the term `IterableView`, since I'd expect it to give an `iter` to consume.

Also had some minor concerns with the interfaces for the different view types. I think there is room for improvement there.

crates/fuel-core/src/service/adapters/consensus_module.rs
MitchTurner: Well, `IterableStore::iter_store()` returns a `BoxedIter<KVItem>`. So that implies sometimes its purpose is iterating.

`BasicView` isn't bad and we can depend on `IterableStore` where we need to with an `impl IterableStore for BasicView` maybe?

crates/fuel-core/src/state/iterable_view.rs
MitchTurner: Do we need the `into_inner` to return `Arc`?

xgreenx: Yep, since we don't know what is type used inside of the `Arc`=) We can't extract it back.

crates/fuel-core/src/service/adapters/consensus_module.rs
xgreenx: The idea of `BasicView` is that it doesn't implement `IterableStore`. Only `View`(`IterableView`) will implement `IterableStore`.

MitchTurner: Oh. Add another view?

crates/fuel-core/src/state/iterable_view.rs
MitchTurner: That makes sense.

How do you feel about a

```
impl<C> From<IterableViewWrapper<C>> for Arc<dyn IterableStore<Column = C> + Sync + Send> 
```

xgreenx: I prefer to not overload `into` and have explicit `into_inner`. It makes clear that the type acts as a container.

crates/fuel-core/src/service/adapters/consensus_module.rs
xgreenx: I lost the line of conversation=D The `IterableView` is an extension of the `KeyValueView`. It is the same as `KeyValueView` but also allows iteration of the storage.

You don't like the name `IterableView`. I'm proposing to change it to `View` and change the name of `KeyValueView` to `BasicView`. Where is the third view here?=D

crates/fuel-core/src/state/iterable_view.rs
MitchTurner: Okay. I'm fine with that.

crates/fuel-core/src/service/adapters/consensus_module.rs
MitchTurner: That was the missing piece for my understanding. I get it now. 

In that case, I'm in favor of `IterableKeyValueView`. It's verbose but very clear.

MitchTurner (approved):

Issue #1581, opened 2024-01-05
Implement properly `AtomicView::latest_view`


xgreenx:
Implement `AtomicView::latest_view` using RockDB read transaction that guarantees state consistency. In the case of `InMemory` database we can simply create a clone of the database and store it inside of the `Arc`.

The change is better to implement after https://github.com/FuelLabs/fuel-core/pull/1576

Issue #1993, opened 2024-06-26
Add caching of the views on the `Database` level


xgreenx:
If the requested storage view doesn't exist, we need to create it and cache it. In the case when someone requests the view at exactly the same height, it should already be ready. The cache can be LRU or timeout-based.

PR #1994, opened 2024-06-26
Added the actual implementation for the `AtomicView::latest_view`

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1581

The change finally implements an atomic view of the RocksDB database via the snapshot feature.

We use a snapshot (if available) each time we read/iterate over the RocksDB. It guarantees a consistent view even if the data is removed/added/modified.

`InMemory` database also implements this feature by simply cloning all `BTreeMap`s. In the future, we need to optimize this behavior and cache views - https://github.com/FuelLabs/fuel-core/issues/1993.

## Checklist

### Before requesting review


crates/fuel-core/src/state/in_memory/memory_store.rs
```diff
@@ -62,6 +69,23 @@ impl<Description> MemoryStore<Description>
 where
     Description: DatabaseDescription,
 {
+    fn create_view(&self) -> MemoryView<Description> {
+        // Lock all tables at the same time to have consistent view.
+        let locks = self
+            .inner
+            .iter()
+            .map(|lock| lock.lock().expect("Poisoned lock"))
```
MitchTurner, line 77: Can we explain in the `expect` why we assume this isn't poisoned? I'm curious, but it's also best practice.

xgreenx: The code below works in the same way=D We could replace it with another lock that doesn't require expect in a separate RP.

MitchTurner: That would be even better than what I suggested.

MitchTurner (approved):
The tests really helped to understand a lot of the code.

Issue #1995, opened 2024-06-27
Support historical view for `MemoryStore`


xgreenx:

MitchTurner:
We don't have an immediate need for this, but it might be worth implementing if someone has time.

PR #1996, opened 2024-06-27
Implementation of the state rewind feature for the RocksDB

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/451

## Overview

Added support for the state rewind feature. The feature allows the execution of the blocks in the past and the same execution results to be received. Together with forkless upgrades, execution of any block from the past is possible if historical data exist for the target block height. The default size of historical/rewind window is 7 days. 

Also added support for rollback command when state rewind feature is enabled. The command allows the rollback of the state of the blockchain several blocks behind until the end of the historical window.

## Implementation details

The change adds a new `HistoricalRocksDB` type that is the wrapper around regular `RocksDB`. This type has inside another RocksDB instance that is used to duplicate all tables plus has one more column to store the reverse modifications at each block height. The reverse modification is the opposite to the operation that was done during transition from block height X to X + 1. The screenshot below should describe the idea:

image

The key of duplicated tables is extended with block height, and the value is the reverse operation to reach the state of entry at the previous height. Having the history of reverse operations, we can iterate back from the latest version of the entry to the previous one.

Using the main property of the RocksDB(sorting keys by default), lookup operations are fast and we don't need to iterate all modifications. It is just enough to find the nearest reverse operation to the target height.

## Checklist

### Before requesting review
  - https://github.com/FuelLabs/fuel-core/issues/1997
  - https://github.com/FuelLabs/fuel-core/issues/1995
  - https://github.com/FuelLabs/fuel-core/issues/1993


bin/fuel-core/src/cli/rollback.rs
```diff
@@ -0,0 +1,93 @@
...
+use anyhow::Context;
+use clap::Parser;
+use fuel_core::{
+    combined_database::CombinedDatabase,
+    service::genesis::NotifyCancel,
+    state::historical_rocksdb::StateRewindPolicy,
+};
+use std::path::PathBuf;
+
+/// Print a snapshot of blockchain state to stdout.
```
Voxelot, line 11: Is this rustdoc copied and pasted from some other command? It seems like this rewinds the database and doesn't print anything at all to std-out.

bin/fuel-core/src/cli/run.rs
```diff
@@ -114,6 +118,20 @@ pub struct Command {
...
+    /// The duration defines how many blocks back the rewind feature works.
+    /// Assuming each block requires one second to produce.
+    ///
+    /// The default value is 7 days = 604800 blocks.
+    ///
+    /// The `BlockHeight` is `u32`, meaning the maximum possible number of blocks
+    /// is less than 137 years. `2^32 / 24 / 60 / 60/ 365` = `136.1925195332` years.
+    /// If the value is 136 years or more, the rewind feature is enabled for all blocks.
+    #[clap(long = "state-rewind-duration", default_value = "7d", env)]
+    pub state_rewind_duration: humantime::Duration,
```
Voxelot, line 133: Should we do blocks instead of time?

What happens if the node is stuck for a while, do we start clearing historical changes even if no new changes are happening?

xgreenx: It works with blocks. We just convert time into blocks. Because `7days` is more readable than `604800`.

crates/fuel-core/src/database.rs
```diff
@@ -184,11 +189,25 @@ where
...
+    ) -> Result<Self> {
         use anyhow::Context;
-        let db = RocksDb::<Description>::default_open(path, capacity.into()).map_err(Into::<anyhow::Error>::into).with_context(|| format!("Failed to open rocksdb, you may need to wipe a pre-existing incompatible db e.g. `rm -rf {path:?}`"))?;
+        let db = RocksDb::<Description>::default_open(path, capacity.into())
+            .map_err(Into::<anyhow::Error>::into)
+            .with_context(|| {
+                format!(
+                    "Failed to open rocksdb, you may need to wipe a \
+                pre-existing incompatible db e.g. `rm -rf {path:?}`"
+                )
```
Dentosal, line 204: nit: Do will still want to suggest wiping the DB?

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
+- [#1996](https://github.com/FuelLabs/fuel-core/pull/1996): Added support for rollback command when state rewind feature is enabled. The command allows the rollback of the state of the blockchain several blocks behind until the end of the historical window. The default historical window it 7 days.
```
Dentosal, line 10: Looks like this might also have breaking changes, at least some `pub` fields were made `pub(crate)`.

MitchTurner: If they were added since the last release, then it should be fine.

crates/fuel-core/src/database.rs
xgreenx: I will remove it in a separate PR related to running Database

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -0,0 +1,787 @@
...
+impl<Description> HistoricalRocksDB<Description>
+where
+    Description: DatabaseDescription,
+{
+    pub fn new(
+        db: RocksDb<Description>,
+        state_rewind_policy: StateRewindPolicy,
+    ) -> DatabaseResult<Self> {
+        let path = db.path().join("history");
+        let history = RocksDb::default_open(path, None)?;
```
Voxelot, line 89: I see some risks to this proliferating pattern of creating many distinct rocksdb instances:

- We lose ACID properties, i.e. issues like the offchain and onchain db getting out of sync can now also happen with the historical changes getting out of sync. What protections are in place to safeguard us from a situation where the changes are committed to the historical database but not to the primary database? I'm concerned that we will introduce new complicated edge cases that are hard to debug by not properly encapsulating all the changes into atomic database transactions which could lead to brittleness under adverse conditions.
- Resource contention overhead such as too many open files and more complexity in configuring rocksdb cache sizes. For example, should all rocksdb instances have equal cache limits? It seems like to me, we should allocate more cache to the main databases vs the historical databases during regular block production etc. However, if an RPC node gets lots of heavy traffic for historical views compared to the amount of work it needs to do to sync, then more cache should be allocated for historical changes. It's hard to predict which db instances would need more caching as it will depend on the usecase. If we didn't have separate DB's then rocksdb would automatically make the most of the cache capacity allocated based on whatever usage scenario is thrown at it.

While I understand that the onchain and offchain updates are not atomic as they are driven by separate processes, I'm wondering if we could make the historical changes part of the same database that they are tracking the changes of. Given that the only way to recover from the offchain and onchain db's falling out of sync is to replay historical data, shouldn't we try to make the historical data as rock solid as possible given that it's part of the same commit_changes function which *should* be atomic?

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -0,0 +1,787 @@
...
+
+        let reverse_changes = self.reverse_history_changes(changes)?;
+        let old_changes = storage_transaction
+            .storage_as_mut::<ModificationsHistory<Description>>()
+            .insert(&height_u64, &reverse_changes)?;
+
+        // The existence of old changes at the current height means we
+        // try to override the current height(maybe because of crash).
+        // So, we need to remove all historical modifications from the past.
+        if let Some(old_changes) = old_changes {
```
Voxelot, line 201: We should probably do some kind of consistency check like this when the historical database is started. Otherwise, it may be possible to generate views at heights which shouldn't exist.

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -0,0 +1,787 @@
...
+    pub fn latest_view(&self) -> StorageResult<RocksDb<Description>> {
+        let db_view = self.db.create_snapshot();
+        Ok(db_view)
+    }
+
+    pub fn create_view_at(
+        &self,
+        height: &Description::Height,
+    ) -> StorageResult<ViewAtHeight<Description>> {
+        let latest_view = self.latest_view()?;
```
Voxelot, line 146: Should we do some kind of consistency check here? Could there be a race condition where the latest view snapshot is not in sync with the historical database when the snapshots are created?

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -0,0 +1,787 @@
...
+        })
+    }
+
+    fn reverse_history_changes(&self, changes: &Changes) -> StorageResult<Changes> {
+        let mut reverse_changes = Changes::default();
+
+        for (column, column_changes) in changes {
+            let results = self
+                .db
+                .multi_get(*column, column_changes.iter().map(|(k, _)| k))?;
```
Voxelot, line 103: Could there be potential race conditions here, ie such as many changes being committed to in parallel during synchronization?

xgreenx: If you work with pure `HistoricalRocksDB` - yes, multiple commits can produce different results.

If you work with `GenericDatabase` - it guarantees only one commit per time.

crates/fuel-core/src/state/historical_rocksdb.rs
xgreenx: This situation is handled on the `GenericDatabase` level. When we commit to the database, we lock the current block height.

image

The same is in the case of `view_at.` We lock the block height so the view can only be created outside of the commit.

image

crates/fuel-core/src/state/historical_rocksdb.rs
xgreenx: > We lose ACID properties, i.e. issues like the offchain and onchain db getting out of sync can now also happen with the historical changes getting out of sync. What protections are in place to safeguard us from a situation where the changes are committed to the historical database but not to the primary database? I'm concerned that we will introduce new complicated edge cases that are hard to resolve if we don't properly encapsulate all the changes into atomic database transactions, leading to brittleness under adverse conditions.

This situation is handled by committing to the historical database first and only afterward to the wrapped database. We can always roll back the `history` database to the same height as the wrapped database.

Below an example of how we process this case while committing new block height:

image


> Resource contention overhead such as too many open files and more complexity in configuring rocksdb cache sizes. For example, should all rocksdb instances have equal cache limits? It seems like to me, we should allocate more cache to the main databases vs the historical databases during regular block production or syncing etc. However, if an RPC node gets lots of heavy traffic for historical views compared to the amount of work it needs to do to sync, then more cache should be allocated for historical changes. It's hard to predict which db instances would need more caching as it will depend on the usecase. If we didn't have separate DB's then rocksdb would automatically make the most of the cache capacity allocated based on whatever usage scenario is thrown at it.

I agree with these concerns. We need to improve the configurability of these parameters. I think it should be possible to store historical data in the same database. I've split it into two databases to have better control over the database configuration and be able to prune historical data just by deleting the folder without additional tools.

I could try to use only one database.

Voxelot: Do you think merging into one db can be a follow-up? Or should we handle this before landing the feature since it'll be disruptive to the database schema with two back to back big db changes?

crates/fuel-core/src/state/in_memory/memory_store.rs
```diff
@@ -161,12 +159,25 @@ where
         Ok(())
     }
 
+    fn view_at_height(
+        &self,
+        _: &Description::Height,
+    ) -> StorageResult<KeyValueView<Self::Column>> {
+        // TODO: https://github.com/FuelLabs/fuel-core/issues/1995
+        unimplemented!("The historical view is not implemented for `MemoryStore`")
```
Voxelot, line 167: Should we return an error here instead?

Voxelot (dismissed):

Voxelot (approved):

Dentosal (approved):

Issue #1971, opened 2024-06-14
Unlimited inputs for GraphQL requests


xgreenx:
Currently, the GraphQL service accepts any data as input. It allows attackers to post a lot of trash data into the queries.

<img width="1293" alt="image" src="https://github.com/FuelLabs/fuel-core/assets/18346821/b7a47719-ef7b-41e9-9217-ccfd7dd13a81">

naino142:
great findings

PR #2001, opened 2024-07-01
Prevent GraphQL query body to be huge and cause OOM

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1971

## Checklist

### Before requesting review


crates/fuel-core/src/service/config.rs
```diff
@@ -129,6 +129,7 @@ impl Config {
                 max_queries_depth: 16,
                 max_queries_complexity: 20000,
                 max_queries_recursive_depth: 16,
+                max_queries_body_limit: 16 * 1024 * 1024,
```
Voxelot, line 132: Currently we support queries up to 7.5mb at the load balancer level. Should we match that here?

I think it's pretty common for RPC providers to support up to 10+ mb responses on other jsonrpc based chains.

cc @mchristopher

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [#1992](https://github.com/FuelLabs/fuel-core/pull/1992): Parse multiple relayer contracts, `RELAYER-V2-LISTENING-CONTRACTS` env variable using a `,` delimiter.
 
 #### Breaking
+- [#2001](https://github.com/FuelLabs/fuel-core/pull/2001): Prevent GraphQL query body to be huge and cause OOM. The default body size is `256KB`. The limit can be changed by the `graphql-max-queries-body-limit` CLI argument.
```
Voxelot, line 19: Currently we support queries up to 7.5mb at the load balancer level. Should we match that with our default here?

I think it's pretty common for RPC providers to support up to 10+ mb responses on other jsonrpc based chains.

cc @mchristopher

xgreenx: This limit applies only to the query itself; it doesn't affect the response.

But I realized that we use a hex representation of transactions, so we need to support twice 256 KB. Maybe it is better to set it to 1 MB.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [#1980](https://github.com/FuelLabs/fuel-core/pull/1980): Add `Transaction` to relayer 's event filter
 
 #### Breaking
+- [#2001](https://github.com/FuelLabs/fuel-core/pull/2001): Prevent GraphQL query body to be huge and cause OOM. The default body size is `1MB`. The limit can be changed by the `graphql-max-queries-body-limit` CLI argument.
```
mchristopher, line 21: Can we reference `bytes` in this variable name? As I read it, it sounds like this value influences the max size of the graph query, not the total byte size of the request. Maybe something like `graphql-max-request-body-bytes`?

bin/fuel-core/src/cli/run/graphql.rs
```diff
@@ -24,6 +24,14 @@ pub struct GraphQLArgs {
     #[clap(long = "graphql-max-recursive-depth", default_value = "16", env)]
     pub graphql_max_recursive_depth: usize,
 
+    /// The max body limit of the GraphQL query.
+    #[clap(
+        long = "graphql-max-queries-body-limit",
+        default_value = "1048576",
```
mchristopher, line 30: As a default, this seems a bit excessive. The built-in default axum limit is 2MB, so I'm not sure that halving that is the best default. I'd think somewhere in the 256KB range as a default, given how much data traversal (i.e., computational expensive queries) can be easily put into a single request.

xgreenx: The hex representation of the `Create` transaction can be `512KB`. I increased it to `1MB` just to allow additional data in the query or dry run two heavy transactions.

Before it was unlimited=)

mchristopher: Got it, makes sense. I misread that we disabled the max limit previously - in that case, 1MB seems reasonable to me.

Voxelot (approved):

PR #2002, opened 2024-07-01
Adapted the block producer to react on the outdated transactions from the TxPool

xgreenx:
Adapted the block producer to react to checked transactions that were using another version of consensus parameters during validation in the TxPool. After an upgrade of the consensus parameters of the network, TxPool could store invalid `Checked` transactions. This change fixes that by tracking the version that was used to validate the transactions.

## Checklist

### Before requesting review


Dentosal (approved):
I'd like to have better tests for this, but I cannot figure out how it would be done, so I think we can just merge this. It looks correct hand has at least some testing.

xgreenx:
I was thinking of creating an e2e test where the first transaction will be a consensus parameters upgrade and the second transaction should fail with upgraded parameters. But the problem is that I need to manipulate gas prices and gas costs to have the correct order. And it will be very fragile tests. So, I decided just to leave it on the executor's side.

I'm happy to add more tests if someone has ideas on how to test it in a better way=D

I will merge it for now, and we can add more tests in follow-up PR.

cc @MitchTurner

Do you have any ideas?=)

Issue #1584, opened 2024-01-05
Handle incorrect shutdown of the off-chain GraphQL worker


xgreenx:
It is possible that the node was shut down before we processed all imported blocks. It could lead to some missed blocks and the database's inconsistent state. Because the result of block execution is not stored on the chain, it is impossible to actualize the database without executing the block at the previous state of the blockchain. When `AtomicView<Storage>::view_at` is implemented, we can process all missed blocks and actualize the database here.

<img width="1488" alt="image" src="https://github.com/FuelLabs/fuel-core/assets/18346821/0994216d-e65c-4bfd-a698-bbafbbdd91b9">

Blocked by https://github.com/FuelLabs/fuel-core/issues/451

PR #2004, opened 2024-07-02
Added handling of incorrect shutdown of the off-chain GraphQL worker

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1584

More information about the problem in the issue.

The solution is to execute old blocks, and use produce `ImportResult` to index the off-chain database.

## Checklist

### Before requesting review


crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -429,27 +443,65 @@ where
...
+
+        let mut task = Task {
+            tx_pool,
+            block_importer,
+            database: off_chain_database,
+            chain_id,
+            continue_on_error,
+        };
+
+        while let Some(Some(block)) = task.block_importer.next().now_or_never() {
```
Voxelot, line 469: could you add comments explaining what this is doing?

crates/fuel-core/src/graphql_api/ports.rs
```diff
@@ -336,6 +344,14 @@ pub mod worker {
         fn block_events(&self) -> BoxStream<SharedImportResult>;
     }
 
+    pub trait ImportResultProvider: Send + Sync {
```
MitchTurner, line 347: I'm not a fan of the name `ImportResultProvider` here or on the adapter. It is referring to the type `ImportResult` which only has meaning inside the block importer context.

Why not just call it `ImportBlock` or something to that effect?

crates/fuel-core/src/graphql_api/worker_service.rs
MitchTurner: Or just move it to a helper function with a self-descriptive name.

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -429,27 +443,65 @@ where
...
+            task.process_block(block)?;
+        }
+
+        loop {
```
MitchTurner, line 473: This could be moved into a `sync_databases` helper or something to that effect.

crates/fuel-core/src/graphql_api/ports.rs
MitchTurner: Is it even appropriate for it to be using the importer type here? Looking at the impl, we are constructing the `SharedImportResult`.

crates/fuel-core/src/service/adapters/graphql_api.rs
```diff
@@ -180,3 +193,42 @@ impl ConsensusProvider for ConsensusParametersProvider {
         self.shared_state.latest_consensus_parameters()
     }
 }
+
+impl ImportResultProvider for ImportResultProviderAdapter {
+    fn result_at_height(
+        &self,
+        height: Option<BlockHeight>,
```
MitchTurner, line 200: This is confusing. Maybe create a new type:
```
enum BlockHeightQuery {
    Specific(BlockHeight),
    Genesis,
}
```
Same cardinality, but easier to understand.

xgreenx: This kind of change is not for this PR=) We use `None` as an identifier of the genesis in a lot of places.

MitchTurner: https://github.com/FuelLabs/fuel-core/issues/2005

crates/fuel-core/src/graphql_api/ports.rs
xgreenx: The second question partially answers the first one=D The block importer returns `SharedImportResult`, which we work with within the off-chain worker.

I will merge Block Importer port with ImportResultProvider

xgreenx:
Just want to hi light, that we need to review https://github.com/FuelLabs/fuel-core/pull/1996 first=)

Voxelot (approved):

Dentosal (approved):

PR #2007, opened 2024-07-03
Improved metrics for the database

xgreenx:
- Added database metrics per column.
- Added statistic about commit time of each database.
- Refactored how metrics are registered: Now, we use only one register shared between all metrics. This global register is used to encode all metrics.

### Before requesting review


Dentosal (approved):
The code looks good. I'm not sure if there's any performance impact here, but I guess we need these anyway.

Voxelot (approved):

PR #2009, opened 2024-07-03
Added shortlived cache to the executor for minimization of the storage access

xgreenx:
I haven't measured the performance of this change, so it is hard to say how useful it is. If, during the execution, we touched the same data, it could definitely help, but if each touched object was unique, then deallocation of all cached entries could be more expensive than saving from not touching the storage.

I will just leave the PR here open to gather thoughts.

### Before requesting review


xgreenx:
Close for now since it is not clear what the benefits are for now.

PR #2010, opened 2024-07-03
Updated block importer to allow more blocks to be queue

xgreenx:
Before, we required all services to process the import result before committing a new result. It was done to reduce the gap between on-chain and off-chain databases and minimize the damage from the https://github.com/FuelLabs/fuel-core/issues/1584 problem.

The https://github.com/FuelLabs/fuel-core/issues/1584 is no longer a problem with https://github.com/FuelLabs/fuel-core/pull/2004 fix. Because of that, we can allow committing more blocks in parallel while other services are processing old ones. It improves synchronization speed because we have a buffer before we wait for other services to catch up. It is very actual for cases when other services are busy right now with other work, but soon will be available to process `ImportResult`.

The default value size of the buffer is `1024`.

### Before requesting review


crates/services/importer/src/importer.rs
```diff
@@ -112,17 +110,21 @@ impl PartialEq for Error {
...
-    /// It is used to await until all receivers of the notification process the `SharedImportResult`
-    /// before starting committing a new block.
-    prev_block_process_result: Mutex<Option<oneshot::Receiver<()>>>,
     guard: tokio::sync::Semaphore,
+    /// The `mpsc` channel is used to notify about the end of the processing of
+    /// the previous `SharedImportResult`. The importer waits until all receivers
+    /// of the notification process  the `SharedImportResult` before starting
+    /// committing a new block.
+    prev_block_process_result_sender: mpsc::Sender<()>,
+    prev_block_process_result_receiver: ResultReceiver,
```
Dentosal, line 127: A better data type for limiting concurrency like this would be [Semaphore](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html), which is specifically the type to use to limit concurrency. Instead of holding `mpsc::Sender`, the processing party (i.e. `Awaiter`) could hold [SemaphorePermit](https://docs.rs/tokio/latest/tokio/sync/struct.SemaphorePermit.html), which automatically does the correct thing on deallocation.

Dentosal (changes requested):

Dentosal (dismissed):

Dentosal (approved):

PR #2013, opened 2024-07-05
Moved P2P database lookups into a separate thread

xgreenx:
When P2P service receives a large number of database related requests, it struggles to process them and p2p events. Database lookup is a heavy operation, and in the case of transactions, it requires a separate lookup for each transaction. It creates a queue of p2p events that will fail with timeouts. 

It causes the node to stop syncing, since it can't process new blocks from the network.

This change adds a separate thread to process database lookups to not affect the p2p communication.

## Checklist

### Before requesting review


crates/services/p2p/src/heavy_task_processor.rs
```diff
@@ -0,0 +1,189 @@
...
+            .build()
+            .map_err(|e| anyhow::anyhow!("Failed to create a rayon pool: {}", e))?;
+        let semaphore = Arc::new(Semaphore::new(number_of_pending_tasks));
+        Ok(Self {
+            rayon_thread_pool,
+            semaphore,
+        })
+    }
+
+    pub fn spawn<OP>(&self, op: OP) -> bool
```
Dentosal, line 25: Instead of `bool`, I'd like to use `Result<(), OutOfCapacity>` or something. This way it's not possible to forget to check the return value, as that would trigger `unused_must_use` lint.

Dentosal (approved):

PR #2015, opened 2024-07-05
Small fixes for the database

xgreenx:
- Fixed the name for historical columns - Metrics was working incorrectly for historical columns.
- Added recommended setting for the RocksDB - the source of recommendation is official documentation https://github.com/facebook/rocksdb/wiki/Setup-Options-and-Basic-Tuning#other-general-options.
- Removed repairing since it could corrupt the database if fails - several users reported about the corrupted state of the database after having a "Too many descriptors" error where in logs, repairing of the database also failed with this error creating a `lost` folder.

### Before requesting review


MitchTurner (approved):
LGTM. 

I'm not sure why we weren't using the inner value from the `HistoricalDuplicateColumn` enum before, this makes more sense. Was that just a mistake or are we reevaluating intended behavior here?

xgreenx:
It was a mistake=D

PR #1983, opened 2024-06-20
L2 Block Source & Metadata Storage implementations

MitchTurner:
Closes: https://github.com/FuelLabs/fuel-core/issues/1955

This actually implements the port adapters from the updater service to access our live databases while running.

Will follow up with another PR that replaces the current updater with the new one.

## Checklist
- [ ] Breaking changes are clearly marked as such in the PR description and changelog
- [ ] New behavior is reflected in tests
- [ ] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [ ] I have reviewed the code myself
- [ ] I have created follow-up issues caused by this PR and linked them here

### After merging, notify other teams

[Add or remove entries as needed]

- [ ] [Rust SDK](https://github.com/FuelLabs/fuels-rs/)
- [ ] [Sway compiler](https://github.com/FuelLabs/sway/)
- [ ] [Platform documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+) (for out-of-organization contributors, the person merging the PR will do this)
- [ ] Someone else?

PR #2025, opened 2024-07-05
Add V0 algorithm to actual services

MitchTurner:
Part of https://github.com/FuelLabs/fuel-core/issues/1624

The [previous PR](https://github.com/FuelLabs/fuel-core/pull/1983) added all the code for working with the dynamic gas price, but didn't plug it into our services. This PR does the actual dep injection. It also adds gas price estimation to AlgoirthmV0.

This change includes new flags for the CLI:
        - "starting-gas-price" - the starting gas price for the gas price algorithm
        - "gas-price-change-percent" - the percent change for each gas price update
        - "gas-price-threshold-percent" - the threshold percent for determining if the gas price will be increase or decreased
    And the following CLI flags are serving a new purpose
        - "min-gas-price" - the minimum gas price that the gas price algorithm will return

## Checklist

### Before requesting review

- [ ] Someone else?


bin/e2e-test-client/src/tests/script.rs
```diff
@@ -143,34 +139,12 @@ pub async fn run_contract_large_state(ctx: &TestContext) -> Result<(), Failed> {
 }
 
 pub async fn arbitrary_transaction(ctx: &TestContext) -> Result<(), Failed> {
-    const RAW_PATH: &str = "src/tests/test_data/arbitrary_tx.raw";
```
MitchTurner, line 146: I removed this because I don't see what it has to do with test and it was blocking my update.

If we want to add some roundtrip conversion test somewhere between raw/json, that should be separate from this logic.

bin/e2e-test-client/src/tests/test_data/arbitrary_tx.json
```diff
@@ -17,7 +17,7 @@
         0,
         0,
         0,
-        100000
+        1562622
```
MitchTurner, line 20: Unfortunately, these are fragile and will need to change any time we change the gas price algo.

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -173,7 +176,20 @@ pub fn init_sub_services(
...
-    let update_algo = StaticAlgorithmUpdater::new(config.static_gas_price);
+    let updater_metadata = UpdaterMetadata::V0(V0Metadata {
+        new_exec_price: config.starting_gas_price,
+        min_exec_gas_price: config.min_gas_price,
+        exec_gas_price_change_percent: config.gas_price_change_percent,
+        l2_block_height: last_height.into(),
+        l2_block_fullness_threshold_percent: config.gas_price_threshold_percent,
+    });
+    // TODO: This looks wrong, why are we getting this from an adapter for some other port?
+    let block_stream = importer_adapter.events_shared_result();
```
MitchTurner, line 187: Using something called `importer_adapter` as something other than an adapter seems like a smell, but that seems to be how we are using this elsewhere.

Ideally, we have a channel and we get the receivers from that before handing the sender to the adapter. I think this file needs to be cleaned up a bit anyway, so might just push that to a cleanup issue.

crates/fuel-gas-price-algorithm/src/v0.rs
```diff
@@ -85,15 +98,14 @@ impl AlgorithmUpdaterV0 {
             .unwrap_or(self.l2_block_fullness_threshold_percent);
 
         match fullness_percent.cmp(&self.l2_block_fullness_threshold_percent) {
-            std::cmp::Ordering::Greater => {
+            std::cmp::Ordering::Greater | std::cmp::Ordering::Equal => {
```
MitchTurner, line 101: This is arbitrary. It made the tests easier to write and we don't have any specs on how the threshold should behave.

bin/e2e-test-client/src/tests/test_data/arbitrary_tx.json
Dentosal: We should document these somehow, maybe a `README.md` in the `test_data` folder.

Dentosal (dismissed):
The code itself LGTM. My only nit is the note I left about documenting the magic values on test txs.

crates/fuel-core/src/service/sub_services.rs
xgreenx: Usually, the service defines its requirements to other services via ports. Importer is one of them. Usually, we pass the importer in the constructor of the service where it calls `get_steram_of_blocks` or something like that.

I don't think explicit work with the channel is better; in this case, a lot of implementation details are revealed. What if later we decide to use the `mpsc` channel or another channel? In this case, you need to update all places to use it. By using the importer and requiring that it provide some stream with blocks, we are more flexible and don't expose implementation details.

bin/fuel-core/src/cli/run.rs
```diff
@@ -165,10 +165,22 @@ pub struct Command {
...
+    #[arg(long = "starting-gas-price", default_value = "1000", env)]
+    pub starting_gas_price: u64,
+
+    /// The percentage change in gas price per block
+    #[arg(long = "gas-price-change-percent", default_value = "10", env)]
+    pub gas_price_change_percent: u64,
+
     /// The minimum allowed gas price
-    #[arg(long = "min-gas-price", default_value = "0", env)]
+    #[arg(long = "min-gas-price", default_value = "1000", env)]
```
xgreenx, line 177: Maybe our default behavior still should be to use a zero gas price to not break tooling and SDKs.  The same for the percent. @Voxelot What do you think?

bin/e2e-test-client/src/tests/test_data/arbitrary_tx.json
xgreenx: Why we can't use 5M or 30M instead?

bin/e2e-test-client/src/tests/script.rs
xgreenx: e2e binary is not a test. It is a binary that runs some tests on a real network. One of the tests is to run an arbitrary transaction where the transaction is hex-encoded from the canonical representation.

The source of the transaction should be `raw` file, not `json` file, because it is how the user sends transaction to the blockchain(in a `raw` representation). The idea of the tool is that you can take any transactions submitted to GraphQL and dry run it on a real network.

I agree that we need to document the use case of the `e2e` binary better and describe what it is responsible for. But we don't need to remove the logic that we need.

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - [1983](https://github.com/FuelLabs/fuel-core/pull/1983): Add adapters for gas price service for accessing database values
 
+### Breaking
```
xgreenx, line 12: Could you document new CLI arguments please?

tests/test-helpers/src/fuel_core_driver.rs
```diff
@@ -20,6 +20,35 @@ impl FuelCoreDriver {
...
     }
 
+    pub async fn spawn_feeless(extra_args: &[&str]) -> anyhow::Result<Self> {
+        let mut args = vec![
+            "--starting-gas-price",
+            "0",
+            "--gas-price-change-percent",
+            "0",
+            "--min-gas-price",
+            "0",
```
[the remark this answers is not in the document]
MitchTurner: The reason I added this is the `min-gas-price` was `1000` by default and it was changing the gas price to `1000` after the first block.

I agree with your other message that it should be `0`, in which case I can remove it here.

crates/services/producer/src/block_producer.rs
```diff
@@ -143,20 +143,9 @@ where
...
-    async fn calculate_gas_price(
-        &self,
-        header: &PartialBlockHeader,
-    ) -> anyhow::Result<u64> {
-        let consensus_params = self
-            .consensus_parameters_provider
-            .consensus_params_at_version(&header.consensus_parameters_version)?;
-        let gas_per_bytes = consensus_params.fee_params().gas_per_byte();
-        let max_gas_per_block = consensus_params.block_gas_limit();
-        let max_block_bytes = max_gas_per_block.checked_div(gas_per_bytes).ok_or(anyhow!(
```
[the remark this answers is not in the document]
MitchTurner: Exactly. It's a YAGNI thing... although I assume we will re-add it once we do the DA gas price.

bin/e2e-test-client/src/tests/test_data/arbitrary_tx.json
MitchTurner: We can. It is all arbitrary, so I just went with the least-arbitrary value. I guess we could avoid some fragility by doing a large value, but it won't be clear to engineers in the future why we used that value.

tests/tests/gas_price.rs
```diff
@@ -4,156 +4,278 @@ use crate::helpers::{
...
+            block_gas_limit,
+            ..Default::default()
+        }),
+        ..ChainConfig::local_testnet()
+    };
+    let mut node_config =
+        Config::local_node_with_configs(chain_config, StateConfig::local_testnet());
+    let starting_gas_price = 1_000_000_000;
+    let percent = 10;
+    let threshold = 50;
```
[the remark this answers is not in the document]
MitchTurner: This feels more "real" to have more or fewer txs submitted for the same threshold. I'd have to set the threshold to `5` to get it to fail with one tx. That's still a decent test, but I feel like there could be some edge cases where that passes but two txs don't.

5 vs 10 txs feels safer for some reason. 

I'm not dead set on that though and willing to try different things.

crates/fuel-gas-price-algorithm/src/v0.rs
```diff
@@ -16,12 +16,25 @@ pub enum Error {
...
     }
+
+    pub fn worst_case(&self, height: u32) -> u64 {
+        let mut price = self.new_exec_price;
+        for _ in self.for_height..height {
+            let change = price.saturating_mul(self.percentage).saturating_div(100);
+            price = price.saturating_add(change);
+        }
+        price
+    }
```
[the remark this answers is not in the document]
MitchTurner: We can use compounding interest formulas, but the problem is those will require floats:
https://www.investopedia.com/terms/c/compoundinterest.asp

Maybe that's okay?

MitchTurner:
> e2e binary is not a test. It is a binary that runs some tests on a real network. One of the tests is to run an arbitrary transaction where the transaction is hex-encoded from the canonical representation.

> The source of the transaction should be raw file, not json file, because it is how the user sends transaction to the blockchain(in a raw representation). The idea of the tool is that you can take any transactions submitted to GraphQL and dry run it on a real network.

> I agree that we need to document the use case of the e2e binary better and describe what it is responsible for. But we don't need to remove the logic that we need.

I reverted these changes. It's pretty bad how we have the override code built into the code. Seems like an abuse of the testing framework.

bin/e2e-test-client/src/tests/test_data/arbitrary_tx.json
MitchTurner: I just bumped the value up to 30M. Idk if we need documentation for it any more than the other values. Happy to add something if you still think I should @Dentosal I get your point if we are doing a specific number, but now it's just a tx with plenty allocated.

bin/e2e-test-client/src/tests/script.rs
```diff
@@ -158,8 +158,11 @@ pub async fn arbitrary_transaction(ctx: &TestContext) -> Result<(), Failed> {
         serde_json::from_str(dry_run_json.as_ref())
             .expect("Should be able do decode the Transaction from json representation");
 
+    let bytes = dry_run_tx_from_json.to_bytes();
+    dbg!(hex::encode(&bytes));
```
xgreenx, line 162: Do we need to keep this `dbg`?

bin/fuel-core/src/cli/run.rs
```diff
@@ -165,10 +165,22 @@ pub struct Command {
     #[arg(long = "native-executor-version", env)]
     pub native_executor_version: Option<StateTransitionBytecodeVersion>,
 
+    /// The starting gas price for the network
```
xgreenx, line 168: Taking into account that we also need some configuration for DA as well, maybe it makes sense to move all price-related fields into their config. What do yu think?

bin/fuel-core/src/cli/run.rs
```diff
@@ -165,10 +165,22 @@ pub struct Command {
...
+    #[arg(long = "starting-gas-price", default_value = "1000", env)]
+    pub starting_gas_price: u64,
```
xgreenx, line 170: To be backward compatible the default starting price also should be zero and `gas-price-change-percent` should be zero as well.

crates/types/src/blockchain/header.rs
```diff
@@ -76,7 +76,7 @@ impl BlockHeader {
     }
 
     /// Mutable getter for consensus portion of header
-    fn consensus_mut(&mut self) -> &mut ConsensusHeader<GeneratedConsensusFields> {
+    pub fn consensus_mut(&mut self) -> &mut ConsensusHeader<GeneratedConsensusFields> {
```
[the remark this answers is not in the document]
xgreenx: It seems you can move this function just into  section below=) 
image

crates/fuel-gas-price-algorithm/src/v0.rs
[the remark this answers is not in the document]
xgreenx: It is off-chain logic, so floats are not a problem. The only requirement is that the final gas price should not be less than if we calculated it via iterations. Maybe we could add a test to verify this behavior.

bin/fuel-core/src/cli/run.rs
MitchTurner: Are you thinking that we'd submit all the fields for the config in a single CLI argument? I'm not sure how that affects this specific piece of code.

crates/fuel-gas-price-algorithm/src/v0.rs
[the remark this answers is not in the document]
MitchTurner: K. I've made those changes and added a proptest to give us some confidence.

bin/fuel-core/src/cli/run.rs
Voxelot: Agree, the CLI defaults should make it easy as possible to start developing locally for the devex onboarding. Production settings should be configured on a per env basis, with the default catering to devs getting started.

MitchTurner: We've changed everything to `0` for now. Works for me.

tests/tests/gas_price.rs
```diff
@@ -4,156 +4,276 @@ use crate::helpers::{
...
+    let starting = driver.node.shared.config.starting_gas_price;
+    let arb_blocks_to_produce = 10;
+    for _ in 0..arb_blocks_to_produce {
+        driver.client.produce_blocks(1, None).await.unwrap();
+        tokio::time::sleep(Duration::from_millis(10)).await;
+    }
+    let latest_gas_price = driver.client.latest_gas_price().await.unwrap();
+    let LatestGasPrice { gas_price, .. } = latest_gas_price;
+    let expected = gas_price;
+    assert_ne!(expected, starting);
```
Voxelot, line 259: should we make this something other than the starting price? To ensure we aren't just defaulting to the starting price on restart.

MitchTurner: This is `ne` for that reason :) Maybe I'm missing something.

crates/fuel-gas-price-algorithm/src/v0.rs
```diff
@@ -16,12 +16,33 @@ pub enum Error {
...
     }
+
+    #[allow(clippy::cast_possible_truncation)]
+    pub fn worst_case(&self, height: u32) -> u64 {
+        let price = self.new_exec_price as f64;
+        let blocks = height.saturating_sub(self.for_height) as f64;
+        let percentage = self.percentage as f64;
+        let percentage_as_decimal = percentage / 100.0;
+        let multiple = (1.0f64 + percentage_as_decimal).powf(blocks);
+        let mut approx = price * multiple;
```
xgreenx, line 37: Are we sure that this function will not overflow during any of the operations?

xgreenx (approved):

tests/tests/gas_price.rs
Voxelot: ah i missed that

crates/fuel-gas-price-algorithm/src/v0.rs
MitchTurner: The prop test should ensure that it will never happen.

xgreenx (approved):

crates/fuel-gas-price-algorithm/src/v0.rs
xgreenx: But do we limit the block horizon that comes from the user? What if it is bigger than 10_000?
image

MitchTurner: You're right. We don't check that (because the test iterates over that entire size and I don't want to do `u32` max in a test).

I can add another test that makes sure it never overflows :)

xgreenx (approved):

Issue #1394, opened 2023-10-03
Return the reason WHY we can't generate the proof instead of `None`


xgreenx:
Yeah, I left this question=D I don't know why we want to return `None` instead of returning the reason WHY we can't generate the proof.

_Originally posted by @xgreenx in https://github.com/FuelLabs/fuel-core/pull/1392#discussion_r1342920044_

xgreenx:
Done in https://github.com/FuelLabs/fuel-core/pull/2258

PR #2028, opened 2024-07-06
Describe why a message proof couldn't be processed

c410-f3r:
Work towards #1394 without changing the public GraphQL API to avoid further breakage. Feel free to indicate if such thing is desirable.

* Changes the return type of `message_proof` from `StorageResult<Option<MessageProof>>` to `StorageResult<MessageProofEvaluation>`.

cc @xgreenx

### DIFF
--- a/crates/fuel-core/src/query/message.rs
+++ b/crates/fuel-core/src/query/message.rs
@@ -33,6 +33,8 @@ use fuel_core_types::{
         MerkleProof,
         Message,
         MessageProof,
+        MessageProofEvaluation,
+        MessageProofUnavailability,
         MessageStatus,
     },
     fuel_merkle::binary::in_memory::MerkleTree,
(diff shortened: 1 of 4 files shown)


crates/fuel-core/src/query/message.rs
```diff
@@ -205,13 +222,19 @@ pub fn message_proof<T: MessageProofData + ?Sized>(
...
+        }
     };
 
     let block_height = *commit_block_header.height();
     if block_height == 0u32.into() {
         // Cannot look beyond the genesis block
-        return Ok(None)
+        return Ok(MessageProofEvaluation::Unavailable(
+            MessageProofUnavailability::InvalidBlockHeaderHeight,
+        ));
```
c410-f3r, line 237: Not sure why it is not possible to see beyond the genesys block so I just copied and pasted the comment to the `InvalidBlockHeaderHeight` variant.

c410-f3r:
Closing due to the lack of feedback

PR #2033, opened 2024-07-16
refactor: remove `Option<BlockHeight>` and use new enum where applicable

### LINKED ISSUE #2005: Remove use of `Option<BlockHeight>` as a query for blocks
There are a number of places we use `Option<BlockHeight>` to represent the block, where `None` is the genesis block. This is confusing, as it's implicit what `None` means without looking at the code.

Instead, we should introduce some type with the same cardinality, but with explicit names:

something like
```
enum BlockQuery {
    Specific(BlockHeight),
    Genesis,
}
```

matt-user:
closes #2005 

I replaced `Option<BlockHeight>` with `BlockHeightQuery` where applicable.  IIUC there are `Option<BlockHeight>` which should remain.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/fuel-core/src/schema/tx.rs
```diff
@@ -132,7 +132,7 @@ impl TxQuery {
             |start: &Option<SortedTxCursor>, direction| {
                 let start = *start;
                 let block_id = start.map(|sorted| sorted.block_height);
-                let all_block_ids = query.compressed_blocks(block_id, direction);
+                let all_block_ids = query.compressed_blocks(block_id.into(), direction);
```
MitchTurner, line 135: Why is this `block_id`? Obviously this pre-existed your PR :)

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -156,42 +157,48 @@ impl DatabaseBlocks for ReadView {
...
 
-        if let Some(height) = height {
+        if let BlockHeightQuery::Specific(height) = height {
             match (height >= self.genesis_height, direction) {
-                (true, IterDirection::Forward) => {
-                    self.on_chain.blocks(Some(height), direction)
-                }
+                (true, IterDirection::Forward) => self
+                    .on_chain
+                    .blocks(BlockHeightQuery::Specific(height), direction),
```
MitchTurner, line 170: This seems like a smell. We are matching on the `BlockHeightQuery` then we reconstruct the `BlockHeightQuery`. I think `blocks` and `old_blocks` can probably just take `Option<BlockHeight>` and we can get rid of the `From` impl into `Option`./F

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -508,7 +508,7 @@ where
             off_chain_height.map(|height| BlockHeight::new(height.saturating_add(1)));
 
         let import_result =
-            import_result_provider.block_event_at_height(next_block_height)?;
+            import_result_provider.block_event_at_height(next_block_height.into())?;
```
MitchTurner, line 511: Is it possible to make `on_chain_database.latest_height()` return `BlockHeightQuery`? 

This makes me think that the name should be something different.

crates/fuel-core/src/schema/block.rs
```diff
@@ -292,7 +295,7 @@ impl BlockQuery {
         crate::schema::query_pagination(after, before, first, last, |start, direction| {
             Ok(blocks_query(
                 query.as_ref(),
-                start.map(Into::into),
+                start.map(Into::<BlockHeight>::into).into(),
```
MitchTurner, line 298: It doesn't look like we can avoid the `Option` here, so this is where we are giving the business-level context to a `None` I guess.

crates/fuel-core/src/graphql_api/worker_service.rs
matt-user: We could change `latest_height()` to return `BlockHeightQuery`.  I didn't do this because, IIUC, a database should be able to return None if it can't find the requested key.  Is the genesis block stored in the DB?  I agree the name `BlockHeightQuery` isn't quite right.  The only other name I can think of is `SpecificBlockOrGenesis`

crates/fuel-core/src/graphql_api/database.rs
matt-user: The only issue with this is that `blocks` is a port so it can either take `Option<BlockHeight>` or `BlockHeightQuery` so then this `blocks` function in `database.rs` would have to take a `Option<BlockHeight>`.  What's the difference between reconstructing `BlockHeightQuery` and reconstructing the `Option` like the previous implementation?

crates/fuel-core/src/schema/tx.rs
Voxelot: it's the cursor to a paginated query

crates/fuel-core/src/graphql_api/database.rs
MitchTurner: > What's the difference between reconstructing BlockHeightQuery and reconstructing the Option like the previous implementation?

There's no difference. I think it was a smell in the old code too.

How about
```
        if let BlockHeightQuery::Specific(inner) = height {
            match (inner >= self.genesis_height, direction) {
                (true, IterDirection::Forward) => self
                    .on_chain
                    .blocks(height, direction),
```

It would be nice if we had some type guarantee since the `inner >= self.genesis_height` is kinda redundant.

MitchTurner (dismissed):

rymnc:
@matt-user / @MitchTurner can I just regen the state transition bytecode, run tests again and then merge?

MitchTurner:
Ah dang. Sorry @matt-user I thought this got merged :o

AurelienFT:
Hello @matt-user ,

Sorry for letting your PR became stale like this. Are you ok to resolve the conflicts so that we can gett this merge soonish ? If you prefer we can take care of the conflicts for you, let me know :)
Sorry again.

matt-user:
> Hello @matt-user ,
> 
> Sorry for letting your PR became stale like this. Are you ok to resolve the conflicts so that we can gett this merge soonish ? If you prefer we can take care of the conflicts for you, let me know :) Sorry again.

Yeah sorry, I completely forgot about this.  I will resolve the conflicts

crates/types/src/blockchain/primitives.rs
```diff
@@ -179,3 +182,30 @@ impl TryFrom<&'_ [u8]> for BlockId {
         Ok(Self::from(TryInto::<[u8; 32]>::try_into(bytes)?))
     }
 }
+
+/// Represents either the Genesis Block or a block at a specific height
+#[derive(Copy, Clone, Debug, Display, PartialEq, Eq, Hash, Ord, PartialOrd)]
+pub enum BlockHeightQuery {
```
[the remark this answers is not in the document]
matt-user: I like `BlockAt` better.  Thanks for the suggestion

AurelienFT (dismissed):
LGTM

crates/fuel-core/src/service/adapters/import_result_provider.rs
```diff
@@ -37,38 +37,42 @@ impl ImportResultProvider {
 impl ImportResultProvider {
     pub fn result_at_height(
```
xgreenx, line 38: The mentioned issue was only related to this part of the code: https://github.com/FuelLabs/fuel-core/pull/2004#discussion_r1663230067

All other changes are not needed. All other places uses `None` because it means no prefix or not starting block

xgreenx (approved):

Issue #2030, opened 2024-07-08
Handle case where improper shutdown of client causes gas price algo to be out of sync


MitchTurner:
The gas price algorithm updater expects to get l2 blocks in order. If that gets out of sync, it won't know how to correct itself. Perhaps we can have it correct itself on initialization similar to what we are doing for GraphQL.

PR #2041, opened 2024-07-23
Add syncronization for Gas Price Database and On Chain Database on startup

MitchTurner:
Closes https://github.com/FuelLabs/fuel-core/issues/2030

In the unlikely situation that the metadata db is behind the on chain database, we should apply the values from each block to the algorithm updater to produce reasonable values.

For now, we should be able to use the updater deterministically, but this might become more difficult as we include the DA values in the updater.

We also define the behavior for when the gas price db is beyond the on-chain db. For now, we are assuming that the blocks produced with the gas prices in the db were either reverted or not properly committed to the on_chain db, and we just revert the gas price db to the height of the on chain db. i.e. **the on chain db is the source of truth.**

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/fuel-core/src/service/sub_services.rs
+++ b/crates/fuel-core/src/service/sub_services.rs
@@ -1,4 +1,5 @@
 #![allow(clippy::let_unit_value)]
+
 use super::{
     adapters::P2PAdapter,
     genesis::create_genesis_block,
@@ -7,14 +8,7 @@ use super::{
 use crate::relayer::Config as RelayerConfig;
 use crate::{
     combined_database::CombinedDatabase,
-    database::{
-        database_description::{
-            gas_price::GasPriceDatabase,
-            on_chain::OnChain,
-        },
-        Database,
-        RegularStage,
-    },
+    database::Database,
     fuel_core_graphql_api,
     fuel_core_graphql_api::Config as GraphQLConfig,
     schema::build_schema,
@@ -34,6 +28,7 @@ use crate::{
             TxPoolAdapter,
             VerifierAdapter,
         },
+        sub_services::algorithm_updater::get_synced_algorithm_updater,
         Config,
         SharedState,
         SubServices,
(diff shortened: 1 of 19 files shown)


crates/fuel-core/src/service/sub_services.rs
```diff
@@ -337,3 +372,97 @@ pub fn init_sub_services(
...
+        l2_block_fullness_threshold_percent: config.gas_price_threshold_percent,
+    });
+    if metadata_height < latest_block_height {
+        let mut metadata_storage = StructuredStorage::new(gas_price_db);
+        let metadata = metadata_storage
+            .get_metadata(&metadata_height.into())?
+            .unwrap_or(default_metadata);
+        let mut inner: AlgorithmUpdater = metadata.into();
+        let AlgorithmUpdater::V0(ref mut updater) = &mut inner;
+        for height in (metadata_height + 1)..=latest_block_height {
```
MitchTurner, line 414: This iterates over all the blocks and applies them to the algorithm. This should work for now, but could become confusing when we include the da price calculation in the future.

The confusing part is that the blocks already include the historical gas prices, so do we really need to re-calculate them via the updater?

crates/fuel-core/src/service/sub_services/algorithm_updater.rs
```diff
@@ -0,0 +1,209 @@
...
+                )?;
+                metadata_storage
+            }
+            Ordering::Greater => {
+                revert_gas_price_db_to_height(
+                    &mut gas_price_db,
+                    latest_block_height.into(),
+                )?;
+                StructuredStorage::new(gas_price_db)
+            }
```
MitchTurner, line 107: I don't like how complex and special-casey this code is. A lot of the complexity is coming from the fact that we don't store the metadata for the genesis block, so we have to manually pass the `genesis_metadata` around for certain cases.

I also think calling `revert_gas_price_db_to_height` in two different locations is a smell, so I might sleep on this code and see if a clear approach arises. Kinda been staring at it too long to think straight.

MitchTurner:
We can create an `InitializationTask` similar to the `graph_ql::Task` that will get run in a non-blocking thread on startup of the service.

This will help us contain the logic for advancing the gas price to the service and just use the OnChain DB to create a list of blocks that still need to be added.

Dentosal (dismissed):
Nice work with the tidy-up commit.

Dentosal (dismissed):

Dentosal (approved):

Voxelot (approved):

PR #2045, opened 2024-07-23
Ignore the message receipt if transaction is reverted

xgreenx:
Include withdrawal message only if transaction is executed successfully, otherwise it is possible to use same assets to create many withdrawals with failed transactions.

The network is patched already with a private fix.

## Checklist

### Before requesting review

### DIFF
--- a/crates/fuel-core/src/executor.rs
+++ b/crates/fuel-core/src/executor.rs
@@ -2573,7 +2573,7 @@ mod tests {
             nonce,
             data,
             ..
-        }) = result.tx_status[0].result.receipts().get(0).cloned()
+        }) = result.tx_status[0].result.receipts().first().cloned()
         else {
             panic!("Expected a MessageOut receipt");
         };
--- a/crates/fuel-core/src/executor.rs
+++ b/crates/fuel-core/src/executor.rs
@@ -54,7 +54,10 @@ mod tests {
             RegId,
         },
         fuel_crypto::SecretKey,
-        fuel_merkle::sparse,
+        fuel_merkle::{
+            common::empty_sum_sha256,
+            sparse,
+        },
         fuel_tx::{
             consensus_parameters::gas::GasCostsValuesV1,
             field::{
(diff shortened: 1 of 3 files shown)


crates/fuel-core/src/executor.rs
```diff
@@ -2528,6 +2531,84 @@ mod tests {
...
+            create_executor(Default::default(), Default::default())
+                .produce_and_commit(block)
+                .expect("block execution failed unexpectedly");
+        create_executor(Default::default(), Default::default())
+            .validate_and_commit(&block)
+            .expect("block validation failed unexpectedly");
+
+        // Then
+        let empty_root = empty_sum_sha256();
+        assert_ne!(block.header().message_outbox_root.as_ref(), empty_root)
```
MitchTurner, line 2570: Is it possible to check that the expected message is there explicitly?

crates/fuel-core/src/executor.rs
```diff
@@ -2528,6 +2531,84 @@ mod tests {
...
+            create_executor(Default::default(), Default::default())
+                .produce_and_commit(block)
+                .expect("block execution failed unexpectedly");
+        create_executor(Default::default(), Default::default())
+            .validate_and_commit(&block)
+            .expect("block validation failed unexpectedly");
+
+        // Then
+        let empty_root = empty_sum_sha256();
+        assert_eq!(block.header().message_outbox_root.as_ref(), empty_root)
```
MitchTurner, line 2609: Do we have enough test coverage elsewhere to show that an empty root does indeed not have any messages?

Dentosal: Yes, here: https://github.com/FuelLabs/fuel-vm/blob/00c5b61d0a51fb4c66fccf121231d63ecaf02ffb/fuel-merkle/src/binary/root_calculator.rs#L128

crates/fuel-core/src/executor.rs
Dentosal: Added the exact check in d119c4925

MitchTurner (approved):

PR #2046, opened 2024-07-23
Limited number of connections per IP

xgreenx:
It is possible to create many connections from the same subnetwork and occupy connection slots. This change should limit the number of connections allowed from the same remote IP.

## Checklist

### Before requesting review

### DIFF
--- a/bin/fuel-core/src/cli/run/p2p.rs
+++ b/bin/fuel-core/src/cli/run/p2p.rs
@@ -90,6 +90,10 @@ pub struct P2PArgs {
     #[clap(long = "max-peers-connected", default_value = "50", env)]
     pub max_peers_connected: u32,
 
+    /// Max number of unique peers connected with the same remote IP address.
+    #[clap(long = "max-peers-per-remote-ip", default_value = "5", env)]
+    pub max_peers_per_remote_ip: usize,
+
     /// Max number of connections per single peer
     /// The total number of connections will be `(max_peers_connected + reserved_nodes.len()) * max_connections_per_peer`
     #[clap(long = "max-connections-per-peer", default_value = "3", env)]
@@ -309,6 +313,7 @@ impl P2PArgs {
             reserved_nodes_only_mode: self.reserved_nodes_only_mode,
             enable_mdns: self.enable_mdns,
             max_peers_connected: self.max_peers_connected,
+            max_peers_per_remote_ip: self.max_peers_per_remote_ip,
             max_connections_per_peer: self.max_connections_per_peer,
             allow_private_addresses: self.allow_private_addresses,
             random_walk,
--- a/crates/services/p2p/src/config.rs
+++ b/crates/services/p2p/src/config.rs
@@ -91,6 +91,8 @@ pub struct Config<State = Initialized> {
     /// This number should be at least number of `mesh_n` from `Gossipsub` configuration.
     /// The total number of connections will be `(max_peers_connected + reserved_nodes.len()) * max_connections_per_peer`
     pub max_peers_connected: u32,
+    /// Max number of unique peers connected with the same remote IP address.
+    pub max_peers_per_remote_ip: usize,
     /// Max number of connections per single peer
     /// The total number of connections will be `(max_peers_connected + reserved_nodes.len()) * max_connections_per_peer`
     pub max_connections_per_peer: u32,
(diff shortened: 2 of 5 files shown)


bin/fuel-core/src/cli/run/p2p.rs
```diff
@@ -90,6 +90,10 @@ pub struct P2PArgs {
     #[clap(long = "max-peers-connected", default_value = "50", env)]
     pub max_peers_connected: u32,
 
+    /// Max number of unique peers connected with the same remote IP address.
+    #[clap(long = "max-peers-per-remote-ip", default_value = "5", env)]
```
matt-user, line 94:
```suggestion
    #[clap(long = "max-peers-per-remote-ip", default_value = "50", env)]
```
right?

xgreenx: No, 5=)

crates/services/p2p/src/config.rs
```diff
@@ -203,6 +206,7 @@ impl Config<NotInitialized> {
             bootstrap_nodes: vec![],
             enable_mdns: false,
             max_peers_connected: 50,
+            max_peers_per_remote_ip: 50,
```
AurelienFT, line 209:
```suggestion
            max_peers_per_remote_ip: 5,
```
The config default here doesn't match the default on clap

crates/services/p2p/src/peer_report.rs
```diff
@@ -94,24 +109,55 @@ impl Behaviour {
...
     fn handle_established_inbound_connection(
         &mut self,
         _connection_id: ConnectionId,
-        _peer: PeerId,
+        peer: PeerId,
         _local_addr: &Multiaddr,
-        _remote_addr: &Multiaddr,
+        remote_addr: &Multiaddr,
     ) -> Result<THandler<Self>, ConnectionDenied> {
+        if let Some(ip) = get_ip_addr(remote_addr) {
```
AurelienFT, line 140: Is it possible that `get_ip_addr` could return two different IP for an IPv4 and his IPv6 mapped or libp2p prevents IPv4-mapped in IPv6 ?

If it's the first case then they will fall in two separate entries and in this case we might want `get_ip_addr` to canonicalise the IP (https://doc.rust-lang.org/std/net/enum.IpAddr.html#method.to_canonical)

crates/services/p2p/src/peer_report.rs
```diff
@@ -71,15 +84,17 @@ pub struct Behaviour {
     connected_reserved_nodes: HashSet<PeerId>,
     pending_connections: HashSet<ConnectionId>,
     pending_events: VecDeque<ToSwarm<PeerReportEvent, Void>>,
+    ips_to_peers: HashMap<IpAddr, HashSet<PeerId>>,
```
AurelienFT, line 87: It seem that this is currently used only to track a number of connection per ip. If we don't want this to be usable for others use cases in the future we could use 
`HashMap<IpAddr, usize>`

It will simplify the current usage that we have by avoiding having to do `.len()` to measure it.

crates/services/p2p/src/peer_report.rs
```diff
@@ -94,24 +109,55 @@ impl Behaviour {
...
+                            "Max connections per IP reached",
+                        ))
+                    }
+
+                    set.insert(peer);
+                }
+                Entry::Vacant(_) => {
+                    // Do nothing
+                }
+            }
```
Dentosal, line 158:
```suggestion
            if let Some(peers) = self.ips_to_peers.get_mut(&ip) {
                if peers.len() >= self.max_connections_per_ip {
                    return Err(ConnectionDenied::new(
                        "Max connections per IP reached",
                    ))
                }
                peers.insert(peer);
            }
```

crates/services/p2p/src/peer_report.rs
```diff
@@ -160,6 +226,35 @@ impl NetworkBehaviour for Behaviour {
...
+                                }
+                                Entry::Vacant(_) => {
+                                    tracing::warn!(
+                                        "IP address not found in the \
+                                        map during disconnection of the peer {:?}",
+                                        peer_id
+                                    )
+                                }
+                            }
+                        }
```
Dentosal, line 255:
```suggestion
                        if let Some(ip) = get_ip_addr(send_back_addr) {
                            if let Some(peers) = self.ips_to_peers.get_mut(&ip) {
                                peers.remove(&peer_id);
                                if peers.is_empty() {
                                    self.ips_to_peers.remove(&ip);
                                }
                            } else {
                                tracing::warn!(
                                    "IP address not found in the \
                                    map during disconnection of the peer {:?}",
                                    peer_id
                                )
                            }
                        }
```

xgreenx:
We added limit on functional connections, allowing many discovery connections in https://github.com/FuelLabs/fuel-core/pull/2743

It should address the issue with many spamming requests.

Issue #2042, opened 2024-07-23
Remove Sparse Merkle tree usage from the `ContractsState` and `ContractsAssets` tables


xgreenx:
## Overview

Based on the latest research that we did, there exist other Merkle trees that show better performance, less touch the storage, and produce less footprint.

Because of that, we've decided to use a tree other than SMT. But their production-ready audited implementation will take some time. Until that moment, we just need to remove usage of SMT.

## Implementation details

 Use a `Plain` blueprint instead of the `Sparse` for `ContractsState` and `ContractsAssets` tables.

Since the `Plain` blueprint doesn't implement `MerkleRootStorage`, in the places where we used root, we need to use and root of the empty tree.

<img width="623" alt="image" src="https://github.com/user-attachments/assets/53a20930-2cdc-4ba9-9941-79f2fea44d4b">

PR #2048, opened 2024-07-24
Disable SMT for `ContractsAssets` and `ContractsState`

xgreenx:
Fixes https://github.com/FuelLabs/fuel-core/issues/2042

Added "smt" feature to the `fuel-core-storage` to enable old behavior. The SMT still is used in the tests and in benchmarks but disabled for the "production" mode.

## Checklist

### Before requesting review


Cargo.toml
```diff
@@ -139,3 +139,6 @@ itertools = { version = "0.12", default-features = false }
 insta = "1.8"
 tempfile = "3.4"
 tikv-jemallocator = "0.5"
+
+[patch.crates-io]
+fuel-vm = { path = "../fuel-vm/fuel-vm" }
```
Voxelot, line 144: ?

xgreenx: The PR requires https://github.com/FuelLabs/fuel-vm/pull/796 as well=)

Cargo.toml
```diff
@@ -139,3 +139,6 @@ itertools = { version = "0.12", default-features = false }
 insta = "1.8"
 tempfile = "3.4"
 tikv-jemallocator = "0.5"
+
+[patch.crates-io]
+fuel-vm = { git = "https://github.com/FuelLabs/fuel-vm/", branch = "feature/implement-for-merkle-root-for-reference" }
```
MitchTurner, line 144: We still need to cut a new release of `fuel-vm` for this. It's merged though.

crates/services/executor/src/refs/contract.rs
```diff
@@ -87,49 +88,101 @@ where
...
+    {
+        type InnerError = StorageError;
+    }
 }
 
 impl<'a, Database> ContractRef<&'a Database>
 where
-    Database: ContractStorageTrait,
-    anyhow::Error: From<Database::InnerError>,
+    &'a Database: ContractStorageTrait<InnerError = StorageError>,
```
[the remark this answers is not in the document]
MitchTurner: I think this one is fine. In reality, it looks like we're just using the `ContractStorageTrait` as a sort of alias for all the other traits. I don't think we're using it anywhere else. But I'm not going to nitpick too much. I guess I'm just not a huge fan of the sort of inheritance-ness going on here.

MitchTurner (approved):

PR #2051, opened 2024-07-31
AWS KMS block signing support and Rust 1.79

### LINKED ISSUE #2043: Add support for AWS KMS as a way to sign block in the `fuel-core`
## Overview

We need to add an ability to sign blocks with AWS KMS.

## Implementation details

Use [`aws-sdk-kms`](https://github.com/awslabs/aws-sdk-rust/tree/main/sdk/kms) crate for adding support for the PoA service to use AWS KMS to sign blocks.

It requires updating the CLI argument to pass a private key. It should be either a raw private key or credentials for KMS.

It would be nice to abstract the signing of the block via some trait and have different backed for signing. The sign function could be `async`.

## Testing details

Please check the https://fuellabs.slack.com/archives/C0298A7TT1C/p1720046999839619 thread for test credentials to test the change with a real cluster.

Dentosal:
Closes #2043

Adds support for using AWS KMS to sign blocks. Updates Rust version to 1.79 as the currently used 1.75.0 is too old for the latest aws crates.

## Testing inside localstack

```bash
localstack start -d
awslocal kms create-key --key-usage SIGN_VERIFY --customer-master-key-spec ECC_SECG_P256K1
AWS_DEFAULT_REGION=us-east-1 AWS_ENDPOINT_URL='http://127.0.0.1:4566'  AWS_ACCESS_KEY_ID=test AWS_SECRET_ACCESS_KEY=test  cargo run --bin fuel-core -- run --consensus-aws-kms KEY_ARN_COMES_HERE --poa-interval-period 1s
```

## Checklist

### Before requesting review

- [ ] Someone else?


bin/fuel-core/src/cli/run.rs
```diff
@@ -306,25 +311,53 @@ impl Command {
...
-        if consensus_key.is_some() && trigger == Trigger::Never {
+        let consensus_signer = if consensus_aws_kms {
+            let config = aws_config::load_from_env().await;
+            let client = aws_sdk_kms::Client::new(&config);
+
+            // Find if we have a key already
+            let key_id = client.list_keys().send().await?.keys.and_then(|keys| {
+                keys.first()
+                    .map(|key| key.key_id.clone().expect("Missing key_id"))
+            });
```
mchristopher, line 322: You won't have permission to do this in a live deployment (but works fine for localstack), this code should accept a `key_id` value as an environment variable (probably called `CONSENSUS_KEY_KMS_ARN` or something similar) and use that.

You'll still want to verify that if that env var is populated, the key can be accessed, so probably best to call GetPublicKey (https://docs.rs/rusoto_kms/latest/rusoto_kms/struct.GetPublicKeyRequest.html) with the provided ARN to make sure permissions are good in the live env.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -129,7 +116,7 @@ pub(crate) enum RequestType {
 }
 
 pub struct MainTask<T, B, I> {
-    signing_key: Option<Secret<SecretKeyWrapper>>,
+    signer: SignMode,
```
MitchTurner, line 119: This should be a port we depend on instead of taking a direct dependency on `aws_sdk_kms`. I'll work on adding that.

bin/fuel-core/src/cli/run.rs
Dentosal: Thanks for the heads-up. FIxed in 2c50c4be4.

bin/fuel-core/src/cli/run.rs
```diff
@@ -306,25 +311,32 @@ impl Command {
             info!("Block production disabled");
         }
 
-        let consensus_key = load_consensus_key(consensus_key)?;
-        if consensus_key.is_some() && trigger == Trigger::Never {
+        let consensus_signer = if let Some(key_id) = consensus_aws_kms {
+            let config = aws_config::load_from_env().await;
+            let client = aws_sdk_kms::Client::new(&config);
+            // Ensure that we can use the key
+            let _ = client.get_public_key().key_id(&key_id).send().await?;
```
mchristopher, line 318: This call will tell you if the key you're trying to access is valid & you can access the public key, but it won't tell you if you actually have permission to use it to sign payloads.

The only way to tell if you can sign payloads is to actually try to do that, which I would not recommend. But may be good to clarify the comment so future readers know what's going on here.

crates/services/consensus_module/poa/src/signer.rs
```diff
@@ -0,0 +1,84 @@
...
+                    .key_id(key_id)
+                    .signing_algorithm(SigningAlgorithmSpec::EcdsaSha256)
+                    .message_type(MessageType::Digest)
+                    .message(Blob::new(*message))
+                    .send()
+                    .await?;
+                let signature_der = reply
+                    .signature
+                    .ok_or_else(|| anyhow!("no signature returned from AWS KMS"))?
+                    .into_inner();
```
mchristopher, line 73: Will this also raise up errors from the SDK? If not, we should surface those for logging - it will make debugging permission or config issues _much_ easier.

Dentosal: We do capture the SDK errors. In this specific case though, there's no error avaible from the SDK. I'm not even sure if this branch can occur; the AWS SDK bindings are poorly documented. It is quite likely that this field can never be `None`, as any error would be captured by the `.send()?` above.

.github/workflows/ci.yml
```diff
@@ -183,6 +183,36 @@ jobs:
         env:
           RUSTFLAGS: -D warnings
 
+  cargo-test-kms:
+    runs-on: buildjet-4vcpu-ubuntu-2204
+    timeout-minutes: 45
```
mchristopher, line 188: One more thing - need to add permissions here:

```
    permissions:
      contents: read
      id-token: write
```

Dentosal: Yep, added in a174a8604. The CI workflow is still somewhat rough on the edges, but I'll have to iterate on that here.

tests/tests/poa.rs
```diff
@@ -85,6 +90,74 @@ async fn can_get_sealed_block_from_poa_produced_block() {
         .expect("failed to verify signature");
 }
 
+#[tokio::test]
+async fn can_get_sealed_block_from_poa_produced_block_when_signing_with_kms() {
+    // This test is only enabled if the environment variable is set
+    let Some(kms_arn) = option_env!("FUEL_CORE_TEST_AWS_KMS_ARN") else {
+        panic!("KMS not enabled");
+        // return;
```
Dentosal, line 98: This panic will be changed to the return when I have made made sure that the test actually fails without KMS.

Dentosal: removed in 4f14fda77

crates/services/consensus_module/poa/src/lib.rs
```diff
@@ -3,16 +3,20 @@
...
 #![deny(warnings)]
+#![allow(clippy::blocks_in_conditions)] // False positives with tracing macros
 
 mod deadline_clock;
 mod sync;
 
 #[cfg(test)]
 mod service_test;
 
+use aws_config as _; // todo
```
MitchTurner, line 14: `todo` what?

crates/services/consensus_module/poa/src/service_test.rs
```diff
@@ -148,21 +152,42 @@ impl TestContextBuilder {
...
             importer,
             p2p_port,
+            signer,
         );
         service.start().unwrap();
         TestContext { service }
     }
 }
 
+struct FakeBlockSigner;
```
MitchTurner, line 171: Looks like both the trait methods are both unimplemented. That's a smell to me.

Of course, it's there because we need to provide some `impl` for the trait, but maybe we should have tests that actually use this?

If not, then maybe we could use the `Unavailable` variant here?

tests/tests/poa.rs
```diff
@@ -85,6 +90,73 @@ async fn can_get_sealed_block_from_poa_produced_block() {
         .expect("failed to verify signature");
 }
 
+#[tokio::test]
+async fn can_get_sealed_block_from_poa_produced_block_when_signing_with_kms() {
```
MitchTurner, line 94: I think it would make sense to have a test for if you don't specify either a key or kms. Show that it fails to produce blocks somehow.

crates/services/consensus_module/poa/src/service_test.rs
Dentosal: Yep and it was also causing tests to fail. I had just forgotten that I actually have to implement those. Done in 225cfc3.

crates/services/consensus_module/poa/src/lib.rs
Dentosal: Figure out if we need the dependency at all, or behind a feature flag. Turns out tt's not needed at all. Removed in 3c0f4ccc9.

tests/tests/poa.rs
Dentosal: Seems too hard to test, not doing this for now.

crates/chain-config/src/config/snapshot_metadata.rs
```diff
@@ -19,6 +19,7 @@ pub enum TableEncoding {
     },
 }
 impl TableEncoding {
+    #[allow(clippy::assigning_clones)] // TODO: clean up the following code
```
MitchTurner, line 22: Let's create an issue for this.

crates/chain-config/src/config/snapshot_metadata.rs
```diff
@@ -80,6 +82,7 @@ impl SnapshotMetadata {
         Ok(snapshot)
     }
 
+    #[allow(clippy::assigning_clones)] // TODO: clean up the following code
```
MitchTurner, line 85: Here too.

crates/chain-config/src/config/snapshot_metadata.rs
Dentosal: https://github.com/FuelLabs/fuel-core/issues/2100

crates/services/consensus_module/poa/src/service_test.rs
```diff
@@ -166,23 +170,46 @@ impl TestContextBuilder {
...
+        if self.succeeds {
+            SignMode::Key(Secret::new(default_consensus_dev_key().into()))
+                .seal_block(block)
+                .await
+        } else {
+            Err(anyhow::anyhow!("failed to sign block"))
+        }
+    }
+
+    fn is_available(&self) -> bool {
```
MitchTurner, line 202: I think we should have some tests for this method.

Dentosal: https://github.com/FuelLabs/fuel-core/issues/2101

tests/tests/lib.rs
```diff
@@ -6,6 +6,11 @@ mod blob;
 mod blocks;
```
xgreenx, line 6: Why do we need changes in this file?

crates/services/consensus_module/poa/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 #![deny(unused_crate_dependencies)]
 #![deny(unused_must_use)]
 #![deny(warnings)]
+#![allow(clippy::blocks_in_conditions)] // False positives with tracing macros
```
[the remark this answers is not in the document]
Dentosal: I feel like a global allow like this is better, as we will probably have more of those blocks later. I don't think this allows any problematic code anyway.

tests/tests/lib.rs
Dentosal: We do not. A mistake with formatting. Fixed in e3d812b4c

bin/fuel-core/Cargo.toml
```diff
@@ -51,12 +56,20 @@ tempfile = { workspace = true }
 test-case = { workspace = true }
 
 [features]
-default = ["env", "relayer", "rocksdb"]
+default = ["env", "relayer", "rocksdb", "aws-kms"]
```
xgreenx, line 59:
```suggestion
default = ["env", "relayer", "rocksdb"]
```

crates/services/consensus_module/poa/Cargo.toml
```diff
@@ -12,10 +12,12 @@ version = { workspace = true }
 [dependencies]
 anyhow = { workspace = true }
 async-trait = { workspace = true }
+aws-sdk-kms = { version = "1.37.0", optional = true }
 fuel-core-chain-config = { workspace = true }
 fuel-core-services = { workspace = true }
 fuel-core-storage = { workspace = true }
 fuel-core-types = { workspace = true }
+k256 = { version = "0.13.3", features = ["ecdsa-core"] }
```
xgreenx, line 20: Why do we need `ecdsa-core` and why we didn't use it before?

xgreenx (commented):
LGTM!=)

crates/services/consensus_module/poa/Cargo.toml
Dentosal: It's needed for DER signature parsing. AWS returns signatures in DER format.

xgreenx (dismissed):

xgreenx (dismissed):

MitchTurner (approved):

xgreenx (dismissed):

AurelienFT:
On my PR that is based on this one : https://github.com/FuelLabs/fuel-core/pull/2102 the CI is failing in the new job added in this PR. I think it's because I'm making the PR from a fork.

Don't know if you think it's a problem or not but I prefer to, at least, mention it.

xgreenx (approved):

MitchTurner (approved):

PR #2060, opened 2024-08-08
Use `min-gas-price` as a starting point if `start-gas-price` is zero

xgreenx:
Use `min-gas-price` as a starting point if `start-gas-price` is zero to follow the old behavior.

## Checklist

### Before requesting review


MitchTurner (approved):
This is probably better/more explicit. It should behave exactly the same after the network starts, but maybe is different if you are checking values before the first block is produced.

MitchTurner (approved):

Issue #1724, opened 2024-03-04
Add malleable fields to transaction subscription


Dentosal:
From @nedsalk 

> Subscriptions currently return receipts, but they're not the only transaction-related data that's set on the node during execution; there are others like some transaction outputs, the receipts root, etc. Would it be possible for the subscriptions to return all the data that changes on a transaction after it's executed? Currently as it stands, the TS SDK creates a transaction summary that the frontend team uses in various use cases. Without e.g. the outputs on the subscription, we have to do another call to fetch them for the summary to be complete. This need for an additional fetch diminishes the benefits of the subscription even if we read the receipts from it.

Related to https://github.com/FuelLabs/fuels-ts/issues/1644

PR #2061, opened 2024-08-08
Allow querying transaction from the status

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1724

### Before requesting review


MitchTurner (commented):
Can we add a test for this?

xgreenx:
> Can we add a test for this?

Test requires updating the query. It means we need to add a new code that we don't plan to use=) I just tried to avoid useless code.

Added the test.

Voxelot (approved):

PR #2067, opened 2024-08-09
Return error from TxPool level if the `BlobId` is known

xgreenx:
Return error from TxPool level if the `BlobId` is known. Currently, we only return it from the block producer level.

## Checklist

### Before requesting review


crates/services/txpool/src/containers/dependency.rs
```diff
@@ -90,77 +96,24 @@ pub struct MessageState {
...
+            blobs: HashMap::new(),
             messages: HashMap::new(),
             max_depth,
             utxo_validation,
         }
     }
 
-    /// find all dependent Transactions that are inside txpool.
-    /// Does not check db. They can be sorted by gasPrice to get order of dependency
-    pub(crate) fn find_dependent(
```
xgreenx, line 106: It is not used anywhere, so I removed it=)

crates/services/txpool/src/containers/dependency.rs
```diff
@@ -268,7 +222,34 @@ impl Dependency {
...
+            let blob_id = checked_tx.transaction().blob_id();
+            if db
+                .blob_exist(blob_id)
+                .map_err(|e| Error::Database(format!("{:?}", e)))?
+            {
+                return Err(Error::NotInsertedBlobIdAlreadyTaken(*blob_id))
+            }
+
+            if let Some(state) = self.blobs.get(blob_id) {
+                if state.tip >= tx.tip() {
```
MitchTurner, line 238: This doesn't seem like a "collision" error, but we are returning `NotInsertedCollisionBlobId`

crates/services/txpool/src/containers/dependency.rs
```diff
@@ -663,6 +652,11 @@ impl Dependency {
             }
         }
 
+        if let PoolTransaction::Blob(checked_tx, _) = tx.as_ref() {
```
MitchTurner, line 655: This doesn't exactly match the name of the function. Maybe we can rename it `remove_all_dependencies`?

crates/services/txpool/src/containers/dependency.rs
xgreenx: If new transaction that deploys the same Blob has lower tip, it is a collision. We can't have two transactions that deploy the same Blob

crates/services/txpool/src/containers/dependency.rs
xgreenx: This function calls `recursively_remove_all_dependencies`, so the name is right=)

crates/services/txpool/src/containers/dependency.rs
MitchTurner: Sure. It's a collision if it's `<` as well though. What we're really doing here is choosing which tx we like more. To the end user the collision error is helpful, I guess, and not inaccurate. `NotInsertedCollisionBlobIdTipLowerThanOtherTx` would be more encompassing. 

I guess I'm having a hard time parsing this function. It would help if it was a little more declarative.

crates/services/txpool/src/containers/dependency.rs
```diff
@@ -461,7 +442,14 @@ impl Dependency {
             // collision of other outputs is not possible.
         }
 
-        Ok((max_depth, db_coins, db_contracts, db_messages, collided))
+        Ok((
```
MitchTurner, line 445: The name is `check_for_collision`. I'd expect a function named that to return `Result<()>` or maybe `Result<Vec<TxId>>, but it's returning a bunch of `HashMap`s as well. Ideally we don't have single function responsible for multiple things, but at the very least we could name the function something more accurate.

I'd prefer to move the db extension logic to separate functions.

crates/services/txpool/src/containers/dependency.rs
```diff
@@ -90,77 +96,24 @@ pub struct MessageState {
     tip: Word,
 }
 
+#[derive(Debug, Clone)]
+pub struct BlobState {
+    origin: TxId,
```
MitchTurner, line 101: Can we rename this to `origin_tx_id`?

crates/services/txpool/src/containers/dependency.rs
MitchTurner: Yeah. On a second look I think it's fine.

crates/services/txpool/src/txpool/tests.rs
```diff
@@ -1421,3 +1355,139 @@ async fn predicate_that_returns_false_is_invalid() {
         "unexpected error: {err}",
     )
 }
+
+#[tokio::test]
+async fn insert_blob_works() {
```
MitchTurner, line 1360:
```suggestion
async fn insert_single__works() {
```

crates/services/txpool/src/txpool/tests.rs
```diff
@@ -1421,3 +1355,139 @@ async fn predicate_that_returns_false_is_invalid() {
...
+    let mut txpool = context.build();
+
+    // Given
+    let tx = check_unwrap_tx(tx, &txpool.config).await;
+
+    // When
+    let result = txpool.insert_single(tx);
+
+    // Then
+    let _ = result.expect("Should insert blob");
```
MitchTurner, line 1384: Can we check that the tx is included rather than looking for an `Ok`?

crates/services/txpool/src/txpool/tests.rs
```diff
@@ -1421,3 +1355,139 @@ async fn predicate_that_returns_false_is_invalid() {
...
+}
+
+#[tokio::test]
+async fn insert_the_same_blob_fails_in_txpool_level() {
```
MitchTurner, line 1388:
```suggestion
async fn insert_single__fails_if_blob_already_inserted_and_lower_tip() {
```

crates/services/txpool/src/txpool/tests.rs
```diff
@@ -1421,3 +1355,139 @@ async fn predicate_that_returns_false_is_invalid() {
...
+
+    // When
+    let result = txpool.insert_single(same_blob_tx);
+
+    // Then
+    assert_eq!(result, Err(Error::NotInsertedCollisionBlobId(blob_id)));
+}
+
+#[tokio::test]
+async fn insert_the_same_blob_works_with_higher_tip() {
```
MitchTurner, line 1426:
```suggestion
async fn insert_single__succeeds_if_blob_already_inserted_but_higher_tip() {
```

crates/services/txpool/src/txpool/tests.rs
```diff
@@ -1421,3 +1355,139 @@ async fn predicate_that_returns_false_is_invalid() {
...
+
+    // When
+    let result = txpool.insert_single(same_blob_tx);
+
+    // Then
+    let _ = result.expect("Should insert transaction with the same blob but higher tip");
+}
+
+#[tokio::test]
+async fn insert_the_same_blob_fails_in_database_level() {
```
MitchTurner, line 1466:
```suggestion
async fn insert_single__fails_if_blob_already_exists_in_database() {
```

crates/services/txpool/src/containers/dependency.rs
```diff
@@ -268,7 +222,34 @@ impl Dependency {
         let mut max_depth = 0;
         let mut db_coins: HashMap<UtxoId, CoinState> = HashMap::new();
         let mut db_contracts: HashMap<ContractId, ContractState> = HashMap::new();
+        let mut db_blobs: HashMap<BlobId, BlobState> = HashMap::new();
```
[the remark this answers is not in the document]
MitchTurner: Not as relevant considering comment below.

crates/services/txpool/src/containers/dependency.rs
xgreenx: I just did it the same way it had worked before. I don't see a point in refactoring the code that will be removed soon during the refactoring of the TxPool.

crates/services/txpool/src/containers/dependency.rs
```diff
@@ -475,7 +463,7 @@ impl Dependency {
     where
         DB: TxPoolDb,
     {
-        let (max_depth, db_coins, db_contracts, db_messages, collided) =
+        let (max_depth, db_coins, db_contracts, db_blobs, db_messages, collided) =
             self.check_for_collision(txs, db, tx)?;
```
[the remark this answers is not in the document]
xgreenx: I don't know=) But let's leave it for the refactoring of the Txpool

crates/services/txpool/src/containers/dependency.rs
xgreenx: It would be nice to change all other places as well with a better name. But it is not part of this PR and will be addressed during the refactoring of the TxPool=) You can check https://github.com/FuelLabs/fuel-core/tree/feature/txpool-v2 it already uses better names.

MitchTurner (approved):

PR #2064, opened 2024-08-08
Allow gas price metadata values to be overridden with config

MitchTurner:
Closes https://github.com/FuelLabs/fuel-core/issues/2065

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog
- [x] New behavior is reflected in tests
- [ ] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [x] I have reviewed the code myself
- [ ] I have created follow-up issues caused by this PR and linked them here

### After merging, notify other teams

[Add or remove entries as needed]

- [ ] [Rust SDK](https://github.com/FuelLabs/fuels-rs/)
- [ ] [Sway compiler](https://github.com/FuelLabs/sway/)
- [ ] [Platform documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+) (for out-of-organization contributors, the person merging the PR will do this)
- [ ] Someone else?

PR #2059, opened 2024-08-07
Bugfix: Remove unwrap in Consensus Param conversion

MitchTurner:
Closes https://github.com/FuelLabs/fuel-core/issues/2058

## Checklist
- [ ] Breaking changes are clearly marked as such in the PR description and changelog
- [ ] New behavior is reflected in tests
- [ ] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [ ] I have reviewed the code myself
- [ ] I have created follow-up issues caused by this PR and linked them here

### After merging, notify other teams

[Add or remove entries as needed]

- [ ] [Rust SDK](https://github.com/FuelLabs/fuels-rs/)
- [ ] [Sway compiler](https://github.com/FuelLabs/sway/)
- [ ] [Platform documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+) (for out-of-organization contributors, the person merging the PR will do this)
- [ ] Someone else?

PR #2069, opened 2024-08-09
Version 0.32.1

MitchTurner:
### Added
- [2061](https://github.com/FuelLabs/fuel-core/pull/2061): Allow querying filled transaction body from the status.

### Changed
- [2067](https://github.com/FuelLabs/fuel-core/pull/2067): Return error from TxPool level if the `BlobId` is known.
- [2064](https://github.com/FuelLabs/fuel-core/pull/2064): Allow gas price metadata values to be overridden with config

### Fixes
- [2060](https://github.com/FuelLabs/fuel-core/pull/2060): Use `min-gas-price` as a starting point if `start-gas-price` is zero.
- [2059](https://github.com/FuelLabs/fuel-core/pull/2059): Remove unwrap that is breaking backwards compatibility
- [2063](https://github.com/FuelLabs/fuel-core/pull/2063): Don't use historical view during dry run.

## What's Changed
* Bugfix: Remove unwrap in Consensus Param conversion by @MitchTurner in https://github.com/FuelLabs/fuel-core/pull/2059
* Don't use historical view during dry run by @xgreenx in https://github.com/FuelLabs/fuel-core/pull/2063
* Weekly `cargo update` by @github-actions in https://github.com/FuelLabs/fuel-core/pull/2050
* Allow gas price metadata values to be overridden with config by @MitchTurner in https://github.com/FuelLabs/fuel-core/pull/2064
* Use `min-gas-price` as a starting point if `start-gas-price` is zero by @xgreenx in https://github.com/FuelLabs/fuel-core/pull/2060
* Allow querying transaction from the status by @xgreenx in https://github.com/FuelLabs/fuel-core/pull/2061
* Return error from TxPool level if the `BlobId` is known by @xgreenx in https://github.com/FuelLabs/fuel-core/pull/2067

**Full Changelog**: https://github.com/FuelLabs/fuel-core/compare/v0.32.0...v0.32.1

### DIFF
--- a/crates/services/upgradable-executor/src/executor.rs
+++ b/crates/services/upgradable-executor/src/executor.rs
@@ -137,7 +137,8 @@ impl<S, R> Executor<S, R> {
         ("0-29-0", 3),
         ("0-30-0", 4),
         ("0-31-0", 5),
-        ("0-32-0", LATEST_STATE_TRANSITION_VERSION),
+        ("0-32-0", 6),
+        ("0-32-1", LATEST_STATE_TRANSITION_VERSION),
     ];
 
     pub fn new(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [Version 0.32.1]
+
 ### Added
 - [2061](https://github.com/FuelLabs/fuel-core/pull/2061): Allow querying filled transaction body from the status.
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2784,7 +2784,7 @@ dependencies = [
 
 [[package]]
 name = "fuel-core"
-version = "0.32.0"
+version = "0.32.1"
 dependencies = [
  "anyhow",
  "assert_matches",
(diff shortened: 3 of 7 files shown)


crates/services/upgradable-executor/src/executor.rs
```diff
@@ -137,7 +137,8 @@ impl<S, R> Executor<S, R> {
         ("0-29-0", 3),
         ("0-30-0", 4),
         ("0-31-0", 5),
-        ("0-32-0", LATEST_STATE_TRANSITION_VERSION),
+        ("0-32-0", 6),
+        ("0-32-1", LATEST_STATE_TRANSITION_VERSION),
```
Voxelot, line 141: did any changes happen to onchain logic? maybe a new wasm isn't required?

MitchTurner: I don't think anything has changed, but does it hurt? I guess pushing the wasm up will be work.

MitchTurner: I feel like it's better to be safe than sorry and just always cut one but 🤷

xgreenx: We agreed in the past to do new WASM for each release to avoid any errors. So everything is correct and we need to bump it=)

Plus we upgraded dependencies, so WASM bytecode is different=)

xgreenx (approved):

Issue #1902, opened 2024-05-18
Support reapplying of the rollbacked blocks during the production of the blocks


xgreenx:
Should be used with https://github.com/FuelLabs/fuel-utils/issues/1

The block producer on start up will check if there are any predefined blocks that need to be produced first, then it will move to normal block production.

- Teach the block producer of `fuel-core` to support the production of the blocks from a pre-defined list of blocks. The blocks should be produced one by one without delays since the `time` is already known, and the blocks are from the past. Pre-defined blocks should be passed as a CLI argument via the file. The format of the file should match with the format from the https://github.com/FuelLabs/fuel-utils/issues/1. The tracking issue for this one is https://github.com/FuelLabs/fuel-core/issues/1902
    - During production we need to track the list of discarded transactions from pre-defined blocks for the statistic to report later in the block post about reorg.

Issue #1917, opened 2024-05-30
Investigate removing `DeadlineClock`


MitchTurner:
It got removed from one of its spots here:
https://github.com/FuelLabs/fuel-core/pull/1914

Might be worth looking at using `Interval` everywhere instead.

Dentosal:
Issue here is that if we error during block production, new blocks wont be produced anymore and the service hangs.

AurelienFT:
I don't have permission to assign myself yet, so placing this comment as an assignment placeholder

PR #2081, opened 2024-08-14
Allow producer to produce predefined blocks

MitchTurner:
## Linked Issues/PRs
Part of https://github.com/FuelLabs/fuel-core/issues/1902

## Description
Provide mechanism for taking predefined blocks and including them in block production.

Does not provide the mechanism for introducing those predefined blocks into the running system, e.g. parsing from file or something. That will come after.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/producer/src/block_producer.rs
```diff
@@ -88,6 +92,58 @@ pub struct Producer<ViewProvider, TxPool, Executor, GasPriceProvider, ConsensusP
...
+    where
+        Executor: ports::BlockProducer<Vec<Transaction>> + 'static,
+    {
+        let _production_guard = self.lock.lock().await;
+
+        let source = block.transactions().to_vec();
+
+        let height = block.header().consensus().height;
+
+        let header = block.header().into();
```
MitchTurner, line 115: We need to regenerate the header.

crates/services/producer/src/block_producer.rs
```diff
@@ -88,6 +92,60 @@ pub struct Producer<ViewProvider, TxPool, Executor, GasPriceProvider, ConsensusP
...
+    {
+        let _production_guard = self.lock.lock().await;
+
+        let source = predefined_block.transactions().to_vec();
+
+        let height = predefined_block.header().consensus().height;
+
+        let block_time = predefined_block.header().consensus().time;
+
+        let header = self.new_header(height, block_time).await?;
```
MitchTurner, line 117: This will fetch latest DA block height, but we want to use the predefined DA block height.

crates/services/producer/src/block_producer.rs
```diff
@@ -88,6 +92,60 @@ pub struct Producer<ViewProvider, TxPool, Executor, GasPriceProvider, ConsensusP
...
+            .ok_or(anyhow!(
+                "The last transaction in the block should be a mint transaction"
+            ))?;
+
+        let gas_price = *mint_tx.gas_price();
+
+        let component = Components {
+            header_to_produce: header,
+            transactions_source: source,
+            coinbase_recipient: self.config.coinbase_recipient.unwrap_or_default(),
```
MitchTurner, line 132: Should use the value from mint tx.

crates/services/producer/src/block_producer/tests.rs
```diff
@@ -534,6 +540,41 @@ mod produce_and_execute_block_txpool {
...
+    let executor = MockExecutorWithCapture::default();
+    let ctx = TestContext::default_from_executor(executor.clone());
+
+    let producer = ctx.producer();
+
+    // when
+    let result = producer.produce_and_execute_predefined(&block).await;
+
+    // then
+    assert!(result.is_ok());
```
MitchTurner, line 575: Verify the time, DA block height, gas price, and coinbase recipient are correct. We can do that via the capture.

tests/tests/snapshot.rs
```diff
@@ -28,10 +28,10 @@ async fn loads_snapshot() {
     // setup config
     let starting_state = StateConfig {
         last_block: Some(LastBlockConfig {
-            block_height: (u32::MAX - 1).into(),
+            block_height: (u32::MAX - 2).into(),
```
xgreenx, line 31: Why do we need to change this file?

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -480,53 +569,55 @@ where
...
+                    if let Some(request) = request {
+                        match request {
+                            Request::ManualBlocks((block, response)) => {
+                                let result = self.produce_manual_blocks(block).await;
+                                let _ = response.send(result);
+                            }
                         }
+                        should_continue = true;
+                    } else {
+                        tracing::error!("The PoA task should be the holder of the `Sender`");
```
Dentosal, line 598: I don't understand this message here. If I read this correctly, this branch is taken iff the PoA task drops the channel, which happens only if it panics. Maybe "The PoA task panicked!" or something like that would be appropriate? Maybe this function should panic as well?

crates/services/producer/src/block_producer.rs
```diff
@@ -88,6 +95,65 @@ pub struct Producer<ViewProvider, TxPool, Executor, GasPriceProvider, ConsensusP
...
+        };
+
+        // Store the context string in case we error.
+        let context_string =
+            format!("Failed to produce block {height:?} due to execution failure");
+        let result = self
+            .executor
+            .produce_without_commit(component)
+            .map_err(Into::<anyhow::Error>::into)
+            .context(context_string)?;
```
Dentosal, line 151: nit: This should probably use [`with_context`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#tymethod.with_context) instead, although the allocation on happy path is likely optimized away anyway.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -387,6 +402,70 @@ where
...
+            .commit_result(Uncommitted::new(
+                ImportResult::new_from_local(sealed_block.clone(), tx_status, events),
+                changes,
+            ))
+            .await?;
+
+        // Update last block time
+        self.last_height = *sealed_block.entity.header().height();
+        self.last_timestamp = sealed_block.entity.header().time();
+        self.last_block_created = last_block_created;
```
xgreenx, line 464: Because you don't set the timer here, I think after applying predefined blocks the server will halt in the case of `Interval` trigger more

Dentosal:
The whole mechanism should be documented somehow, somewhere. Probably in doc comments in some relevant trait or function.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -387,6 +402,70 @@ where
...
+            .await?
+            .into();
+
+        if !skipped_transactions.is_empty() {
+            tracing::error!("During block production got invalid transactions");
+            let txs = predefined_block.transactions();
+
+            for (tx_id, err) in skipped_transactions {
+                let maybe_tx = txs.iter().find(|tx| tx.id(chain_id) == tx_id);
+                if let Some(tx) = maybe_tx {
```
xgreenx, line 438: We should always be able to find the transaction. Maybe it is better to return an error

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -387,6 +402,70 @@ where
...
+            .produce_predefined_block(predefined_block)
+            .await?
+            .into();
+
+        if !skipped_transactions.is_empty() {
+            tracing::error!("During block production got invalid transactions");
+            let txs = predefined_block.transactions();
+
+            for (tx_id, err) in skipped_transactions {
+                let maybe_tx = txs.iter().find(|tx| tx.id(chain_id) == tx_id);
```
xgreenx, line 437: It is very performance-heavy. It is better to iterate over all transactions in the block and check that it is in the hash map of failed transactions.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -387,6 +402,70 @@ where
...
+                if let Some(tx) = maybe_tx {
+                    tracing::error!(
+                "During block production got invalid transaction {:?} with error {:?}",
```
xgreenx, line 440: We agreed that in the case of discarded predefined transactions we want to print them in JSON format with some `BEGIN {JSON_TX} END` marker.

crates/services/consensus_module/poa/src/service.rs
xgreenx: Maybe we could somehow reuse main part of the `produce_block` function

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -480,53 +569,55 @@ where
...
-                    match request {
-                        Request::ManualBlocks((block, response)) => {
-                            let result = self.produce_manual_blocks(block).await;
-                            let _ = response.send(result);
+        let next_height = self.next_height();
+        let chain_id = self.chain_id;
+        let maybe_block = self.predefined_blocks.get_block(&next_height);
+        if let Some(block) = maybe_block {
+            self.produce_predefined_block(&block, &chain_id).await?;
+            should_continue = true;
```
xgreenx, line 581: Could we please make an early return here? To avoid shifting a big part of the code into the `if` statement?

We can move the code below into the `else {` in a separate PR

crates/services/consensus_module/poa/src/service_test.rs
```diff
@@ -397,3 +425,124 @@ fn test_signing_key() -> Secret<SecretKeyWrapper> {
...
+    Block::new(header, transactions, Default::default(), Default::default()).unwrap()
+}
+
+#[tokio::test]
+async fn consensus_service__run__will_include_predefined_blocks_before_new_blocks() {
+    // given
+    let blocks: [(BlockHeight, Block); 3] = [
+        (2u32.into(), block_for_height(2)),
+        (3u32.into(), block_for_height(3)),
+        (4u32.into(), block_for_height(4)),
```
xgreenx, line 502: Could we also test the case when we produce some blocks first form the interval, then we have predefined blocks, please?

crates/services/consensus_module/poa/src/service_test.rs
```diff
@@ -397,3 +425,124 @@ fn test_signing_key() -> Secret<SecretKeyWrapper> {
...
+    let service = ServiceRunner::new(task);
+    service.start().unwrap();
+
+    // then
+    for (_, block) in blocks {
+        let expected = FakeProducedBlock::Predefined(block);
+        let actual = block_receiver.recv().await.unwrap();
+        assert_eq!(expected, actual);
+    }
+    let maybe_produced_block = block_receiver.recv().await.unwrap();
```
xgreenx, line 543: Could this test also verify the production of the block after pre-defined blocks, please? Like 3 predefined blocks, and 1-3 non predefined(from interval)

tests/tests/snapshot.rs
MitchTurner: Because the code internally calls `next_block` now. If it's `u32::Max` (as it becomes) then `next_block` will panic.

xgreenx: Then test shouldn't produce any blocks, maybe we need to disable block producer

crates/services/consensus_module/poa/src/service.rs
MitchTurner: This comment was here before. I agree it's not clear.

tests/tests/snapshot.rs
MitchTurner: Adding `Trigger::Never` doesn't seem to fix the issue; the producer is still being run.

crates/services/consensus_module/poa/src/service.rs
xgreenx: It will be removed during https://github.com/FuelLabs/fuel-core/issues/1917

xgreenx (approved):

Issue #1901, opened 2024-05-18
Add the logic that allows rotation of the PoA key


xgreenx:
The PoA key is defined by the `ChainConfig` and currently supports only one value. We need to support the key's rotation. As an example, we could have a new variant with listed PoA key rotations.

```rust
pub enum ConsensusConfig {
    PoA {
        signing_key: Address,
    },
    RotatedPoA {
        /// The key of the map is the block height at which the rotation occurs.
        /// The value is the new signing key.
        rotations: BTreeMap<BlockHeight, Address>,
    },
}
```

JoeruCodes:
For the default_poa function for the `RotatedPoA` Variant the `BTreeMap` would be generated from this `SecretKey::new_from_mnemonic_phrase_with_path(<PHRASE>, <PATH>)` where PATH is offsetted by `format!("m/44'/60'/0'/0/{}", nKeys)` would be the correct approach to generate the keys right? 

```rust
/// A default secret key to use for testing purposes only
pub fn default_consensus_dev_key() -> SecretKey {
    // Derived from:
    //  - Mnemonic phrase: "winner alley monkey elephant sun off boil hope toward boss bronze dish"
    //  - Path: "m/44'/60'/0'/0/0"
    // Equivalent to:
    //  `SecretKey::new_from_mnemonic_phrase_with_path(..)`
    let bytes: [u8; 32] = [
        0xfb, 0xe4, 0x91, 0x78, 0xda, 0xc2, 0xdf, 0x5f, 0xde, 0xa7, 0x4a, 0x11, 0xa9,
        0x0f, 0x99, 0x77, 0x62, 0x5f, 0xe0, 0x23, 0xcd, 0xf6, 0x41, 0x4b, 0xfd, 0x63,
        0x9d, 0x32, 0x7a, 0x2e, 0x9d, 0xdb,
    ];
    SecretKey::try_from(bytes.as_slice()).expect("valid key")
}

```

PR #2086, opened 2024-08-15
Added support for PoA key rotation

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1901
Closes https://github.com/FuelLabs/fuel-core/issues/1919

The `ConsensusConfig` now contains one more variant, `PoAV2`, that also overrides the PoA at specific block heights. It means that starting this block the node expects blocks signed by another PoA key.

Along with that, I've added additional features to automate the migration of the state so it is compatible with the new chain configuration:
- During start-up, the `fuel-core` binary checks if there are any changes in the chain config. If they are, the binary re-inserts the new `Conesnsus::Genesis` type in the `SealedBlockConsensus` table.  Now we can do https://github.com/FuelLabs/chain-configuration/issues/1.
- The `fuel-core` to react to `signing_key_overrides`. If the blocks at the corresponding overwritten block height have different signers, the `fuel-core` rollbacks to the height before overriding occurs.

Side changes:
- Renamed the methods of the `FuelService` to be more explicit on the logic.
- Moved business logic from `FuelService::Task` to `FuelService`.

## Checklist

### Before requesting review


benches/benches/block_target_gas.rs
```diff
@@ -337,17 +336,19 @@ fn service_with_many_contracts(
...
+            CombinedDatabase::new(
+                database,
+                Default::default(),
+                Default::default(),
+                Default::default(),
+            ),
+            config.clone(),
+        )
+        .await
+        .expect("Unable to start a FuelService")
```
Dentosal, line 350:
```suggestion
        .expect("Unable to start FuelService")
```

crates/fuel-core/src/combined_database.rs
```diff
@@ -229,6 +230,88 @@ impl CombinedDatabase {
...
+
+            if gas_price_chain_height > target_block_height {
+                self.gas_price().rollback_last_block()?;
+                tracing::info!(
+                    "Rolled back gas-price-chain database to height {:?}",
+                    gas_price_chain_height.pred()
+                );
+            }
+        }
+        Ok(())
```
Dentosal, line 307: Should we still return `Ok(())` when cancelled?

Dentosal (dismissed):

crates/fuel-core/src/service.rs
```diff
@@ -161,69 +207,167 @@ impl FuelService {
...
+        let mut rollback_height = None;
+        match start_up_consensus_config {
+            ConsensusConfig::PoA { .. } => {
+                // We don't support overriding of the heights for PoA version 1.
+            }
+            ConsensusConfig::PoAV2(poa) => {
+                let on_chain_view = combined_database.on_chain().latest_view()?;
+
+                for override_height in poa.get_all_overrides().keys() {
+                    let block_header = on_chain_view
```
xgreenx, line 230: Handle overrides fro mthe future

MitchTurner (approved):

MitchTurner (approved):

PR #2076, opened 2024-08-14
chore(iterators): replace iter_all with iter_all_keys where required

rymnc:
## Linked Issues

https://github.com/FuelLabs/fuel-core/issues/1997

## Description

Replaces usages of `iter_all` with `iter_all_keys` in instances where we only need to iterate over the keys. A follow up PR will be made to actually iterate over the keys without making allocations for the values, as described in https://github.com/FuelLabs/fuel-core/blob/a9e5e89f5fb38e3b3d6e6bdfe1ad339a01f2f3b9/crates/storage/src/iter.rs#L130

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog (no breaking changes)
- [x] New behavior is reflected in tests (existing test cases which test behaviour of iterators pass)
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (no changes to the spec)

### Before requesting review
- [ ] I have reviewed the code myself
- [ ] I have created follow-up issues caused by this PR and linked them here

### After merging, notify other teams

[Add or remove entries as needed]

- [ ] [Rust SDK](https://github.com/FuelLabs/fuels-rs/)
- [ ] [Sway compiler](https://github.com/FuelLabs/sway/)
- [ ] [Platform documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+) (for out-of-organization contributors, the person merging the PR will do this)
- [ ] Someone else?

Issue #1997, opened 2024-06-28
Optimize iteration over the keys in the database


xgreenx:
We have many cases when we need to iterate only over keys of the specific table/column. The current iteration logic loads both key and values into memory and later discards values. To avoid vesting of resources, we can add more functionality and provide iteration only over the keys.

As an example:
When we want to get the last block height, we load the whole `CompressedBlock` into the memory, when we don't need it.

vishnuc77:
Hi @xgreenx, I'm a first time contributor trying to get involved with the project.
I would like to know whether this task is something like making use of `iter_all_keys()` in place of (and other similar places) :
```
pub fn latest_height(&self) -> StorageResult<BlockHeight> {
        self.iter_all::<FuelBlocks>(Some(IterDirection::Reverse))
            .next()
            .ok_or(not_found!("BlockHeight"))?
            .map(|(height, _)| height)
 }
 ```
 so that CompressedBlock itself would not be loaded into memory, or is there something more to it?

rymnc:
taking this up 😁

rymnc:
should we mark this as closed with #2092?
cc: @xgreenx

xgreenx:
Closed via https://github.com/FuelLabs/fuel-core/pull/2092

PR #2092, opened 2024-08-16
feat(iterators): allow key-only iteration

rymnc:
## Linked Issues/PRs

- #1997
- #2076

## Description

Implements `iter_store_keys` and `iter_keys` appropriately for in memory stores and rocksdb. The previous PR made them call it, and this one changes the underlying implementation to avoid value allocations.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog (no breaking changes)
- [x] New behavior is reflected in tests (tests which test this behaviour under the hood still pass)
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (no changes to spec)

### Before requesting review

- [ ] Someone else?


crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -706,6 +785,89 @@ where
             }
         }
     }
+
+    fn iter_store_keys(
```
rymnc, line 789: taken *mostly* as-is from the `iter_store` fn defn, but only uses keys and uses `_iter_all_keys` under the hood.

I have an implementation which abstracts the functionality of using key prefixes and start index, but it seems verbose and might not provide much value.

crates/fuel-core/src/state/rocks_db_key_iterator.rs
```diff
@@ -0,0 +1,81 @@
...
+use rocksdb::{
+    DBAccess,
+    DBRawIteratorWithThreadMode,
+};
+
+pub enum OwnedIteratorMode {
+    Start,
+    End,
+    From(Vec<u8>, rocksdb::Direction),
+}
```
rymnc, line 17: Favouring "owning" the `IteratorMode` because our keys are relatively small, and to avoid all the lifetime annotations in places which use this to cut cognitive load.

crates/fuel-core/src/state.rs
```diff
@@ -157,4 +159,19 @@ where
...
+        prefix: Option<&[u8]>,
+        start: Option<&[u8]>,
+        direction: IterDirection,
+    ) -> BoxedIter<fuel_core_storage::kv_store::KeyItem> {
+        // We cannot define iter_store_keys appropriately for the `ChangesIterator`,
+        // because we have to filter out the keys that were removed, which are
+        // marked as `WriteOperation::Remove` in the value
+        self.iter_store(column, prefix, start, direction)
+            .map(|item| item.map(|(key, _)| key))
+            .into_boxed()
```
xgreenx, line 175: You could copy/paste everything from the function above and just remove `value.clone()`=)

crates/fuel-core/src/state/in_memory/memory_store.rs
```diff
@@ -129,6 +129,18 @@ where
...
     }
+
+    fn iter_store_keys(
+        &self,
+        column: Self::Column,
+        prefix: Option<&[u8]>,
+        start: Option<&[u8]>,
+        direction: IterDirection,
+    ) -> BoxedIter<fuel_core_storage::kv_store::KeyItem> {
+        self.iter_all(column, prefix, start, direction)
```
xgreenx, line 140: [nit] We could avoid `value.clone()` if we implemented that in some another way=) Maybe we could use generics  to minimize the code duplciation

crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -499,6 +505,55 @@ where
         }
     }
 
+    fn reverse_prefix_iter_keys(
```
xgreenx, line 508: The `reverse_prefix_iter` function is pretty complex on its own, and having two complex functions will be hard to support=D Maybe we could use the generic, which will define what type to return and defines the closure to get that type?

crates/fuel-core/src/state/rocks_db_key_iterator.rs
```diff
@@ -0,0 +1,81 @@
...
+
+impl<'a, D: DBAccess> Iterator for RocksDBKeyIterator<'a, D> {
+    // decoupling the Key type from crate::storage::KeyItem
+    type Item = DatabaseResult<Vec<u8>>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.done {
+            None
+        } else if let Some(key) = self.raw.key() {
+            let key = key.to_vec();
```
xgreenx, line 33: I'm curious. Is it possible to return a slice instead of the vector? In most cases we decode key and don't use the vector itself

rymnc: not without adding additional lifetime constraints :/

xgreenx: Okay, we can try to play with it in follow up PR=)

crates/fuel-core/src/state/rocks_db.rs
rymnc: totally agree, however, since the types we want to implement for, namely `KVItem` and `KeyItem` are a tuple and a distinct type, writing a generic for this becomes more involved. We will have to define a trait for both `KVItem` and `KeyItem` which provide the following helper fn -

```rust
trait StorageItem {
  type Key = Vec<u8>;

  fn key() -> Self::Key,
  fn key_starts_with(prefix: Self::Key) -> bool; 
}
```

which makes it easier to write the function, in the following way -


```rust
fn reverse_prefix_iter_generic<T, F>(
        &self,
        prefix: &[u8],
        column: Description::Column,
        iter_store_fn: F,
        iter_all_fn: impl Fn(
            &Self,
            Description::Column,
            ReadOptions,
            OwnedIteratorMode,
        ) -> BoxedIter<StorageResult<T>>,
    ) -> BoxedIter<StorageResult<T>>
    where
        T: 'static,
        F: Fn(
            &Self,
            Description::Column,
            Option<&[u8]>,
            Option<&[u8]>,
            IterDirection,
        ) -> BoxedIter<'static, StorageResult<T>>,
    {
        let maybe_next_item = next_prefix(prefix.to_vec())
            .and_then(|next_prefix| {
                iter_store_fn(
                    self,
                    column,
                    Some(next_prefix.as_slice()),
                    None,
                    IterDirection::Forward,
                )
                .next()
            })
            .and_then(|res| res.ok());

        let prefix_vec = prefix.to_vec(); 
        
        match maybe_next_item {
            Some(key_or_item) => {   
                let next_start_key_vec = key_or_item.key().to_vec();

                let iter_mode = OwnedIteratorMode::From(
                    next_start_key_vec.to_vec(),
                    rocksdb::Direction::Reverse,
                );

                iter_all_fn(self, column, self.read_options(), iter_mode)
                    .skip(1)
                    .take_while(move |item| match item {
                        Ok(key_or_item) => key_or_item.key_starts_with(&prefix_vec), 
                        Err(_) => true,
                    })
                    .into_boxed()
            }
            None => {
                iter_all_fn(self, column, self.read_options(), OwnedIteratorMode::End)
                    .take_while(move |item| match item {
                        Ok(key_or_item) => key_or_item.key_starts_with(&prefix_vec), 
                        Err(_) => true,
                    })
                    .into_boxed()
            }
        }
    }
```

the important line of code in the above fn being `let next_start_key_vec = key_or_item.key().to_vec();`

to avoid destructuring the tuple since the generic `T` can either be a KVItem (a tuple) or a KeyItem (Vec<u8>)

perhaps this can be addressed in a follow up PR to avoid a large diff :)

what are your thoughts?

crates/fuel-core/src/state.rs
rymnc: sure, addressed in 06bd0e962
thanks

crates/fuel-core/src/state/in_memory/memory_store.rs
rymnc: yeah, maybe we should change the fn signature of `iter_all_keys` in `IteratorOverTable` to include the column, prefix, start key. it is currently - 
```
    fn iter_all_keys<M>(
        &self,
        direction: Option<IterDirection>,
    ) -> BoxedIter<super::Result<M::OwnedKey>>
    where
        M: Mappable,
        Self: IterableTable<M>,
    {
        self.iter_all_filtered_keys::<M, [u8; 0]>(None, None, direction)
    
    ```

Comment (rymnc) [reply crates/fuel-core/src/state/in_memory/memory_store.rs]: nevermind, fixed in 1665b8606

thanks!

Comment (rymnc) [reply crates/fuel-core/src/state/in_memory/memory_store.rs]: will modify `keys_iterator` to do the same

File: crates/fuel-core/src/state/rocks_db_key_iterator.rs (hunk 6 above)
Comment (rymnc) [reply crates/fuel-core/src/state/rocks_db_key_iterator.rs:33]: https://github.com/FuelLabs/fuel-core/issues/2097#issue-2471520476

**Review by Dentosal — CHANGES_REQUESTED** [review]

File: crates/fuel-core/src/state/rocks_db.rs (hunk 5 above)
Comment (rymnc) [reply crates/fuel-core/src/state/rocks_db.rs:528]: @xgreenx check out the diff in e44976669, I added it there and if we want to include as a follow up pr, i can create it

File: CHANGELOG.md [hunk 7]
```diff
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - [2061](https://github.com/FuelLabs/fuel-core/pull/2061): Allow querying filled transaction body from the status.
+- [2092](https://github.com/FuelLabs/fuel-core/pull/2092): Allow iterating by keys in rocksdb, and other storages.
```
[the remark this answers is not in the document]
rymnc: addressed in [ad29d89](https://github.com/FuelLabs/fuel-core/pull/2092/commits/ad29d89427cf8309c40444c041df8f15259e201a)

crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -519,6 +525,55 @@ where
...
+                    column,
+                    Some(next_prefix.as_slice()),
+                    None,
+                    IterDirection::Forward,
+                )
+                .next()
+            })
+            .and_then(|res| res.ok());
+
+        let prefix_vec = prefix.to_vec(); // Convert `prefix` to an owned `Vec<u8>`
```
MitchTurner, line 545: This probably doesn't need to be `to_vec` you can just use the slice below.

MitchTurner: It looks like it will be painful trying to get the lifetimes to align here. It's probable that we can just use a constant sized array in the future, rather than a vec, in which case this shouldn't be a problem.

Not really a problem now either, but worth optimizing for sure.

xgreenx (dismissed):

Dentosal (approved):

Dentosal (approved):

xgreenx (approved):

PR #2096, opened 2024-08-16
feat: graphql endpoint to fetch the blob byte code by its blob ID

### LINKED ISSUE #2084: Add a GraphQL endpoint to fetch the blob byte code by its ID
Add a GraphQL endpoint to fetch the blob byte code using its ID. The endpoint should return a type that has a `bytecode` field and an `id` field.

If the bytecode doesn't exist, it should return an error.

netrome:
## Linked Issues/PRs
- Closes #2084 

## Description
This PR adds a graphql query to fetch blob bytecode by their ID

## Checklist

### Before requesting review

- [ ] [Platform documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - [2051](https://github.com/FuelLabs/fuel-core/pull/2051): Add support for AWS KMS signing for the PoA consensus module. The new key can be specified with `--consensus-aws-kms AWS_KEY_ARN`.
 - [2092](https://github.com/FuelLabs/fuel-core/pull/2092): Allow iterating by keys in rocksdb, and other storages.
+- [2096](https://github.com/FuelLabs/fuel-core/pull/2096): GraphQL endpoint to fetch blob byte code by its blob ID.
 
 ### Changed
 
@@ -23,7 +24,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2094](https://github.com/FuelLabs/fuel-core/pull/2094): Added `--predefined-blocks-path` CLI argument to pass the path to the predefined blocks.
 - [2081](https://github.com/FuelLabs/fuel-core/pull/2081): Enable producer to include predefined blocks.
 - [2079](https://github.com/FuelLabs/fuel-core/pull/2079): Open unknown columns in the RocksDB for forward compatibility.
-- [2096](https://github.com/FuelLabs/fuel-core/pull/2096): GraphQL endpoint to fetch blob byte code by its blob ID.
 
 ### Changed
 - [2076](https://github.com/FuelLabs/fuel-core/pull/2076): Replace usages of `iter_all` with `iter_all_keys` where necessary.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2094](https://github.com/FuelLabs/fuel-core/pull/2094): Added `--predefined-blocks-path` CLI argument to pass the path to the predefined blocks.
 - [2081](https://github.com/FuelLabs/fuel-core/pull/2081): Enable producer to include predefined blocks.
 - [2079](https://github.com/FuelLabs/fuel-core/pull/2079): Open unknown columns in the RocksDB for forward compatibility.
+- [2096](https://github.com/FuelLabs/fuel-core/pull/2096): GraphQL endpoint to fetch blob byte code by its blob ID.
 
 ### Changed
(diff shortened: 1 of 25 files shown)


crates/fuel-core/src/query/blob.rs
```diff
@@ -0,0 +1,38 @@
...
+impl<D: OnChainDatabase + OffChainDatabase + ?Sized> BlobQueryData for D {
+    fn blob_id(&self, id: BlobId) -> StorageResult<BlobId> {
+        let blob_exists = self.storage::<BlobData>().contains_key(&id)?;
+
+        if blob_exists {
+            Ok(id)
+        } else {
+            Err(not_found!(BlobData))
+        }
+    }
```
Dentosal, line 27: Feels a bit weird to have an identity function endpoint that tests inside if the blob exists. Typically things like this would return a boolean.

netrome: I agree. I did it this way to be consistent with the implementation in `crates/fuel-core/src/query/contract.rs`. I'm happy to change this to be an `blob_id_exists` method that returns a boolean, but then I would also change any other methods using this pattern including the `contract_id` method in `crates/fuel-core/src/query/contract.rs`, which would add some refactor-noise to this PR.

Do you prefer if I do this now or as a follow-up?

Dentosal: Hmmm I didn't realize the `contract_id` was like this. I'm feeling the backwards compatibility is more important than consistency here.

tests/tests/blob.rs
```diff
@@ -151,3 +168,19 @@ async fn blob__accessing_nonexitent_blob_panics_vm() {
...
+    // When
+    let bytecode: Vec<u8> = [op::ret(RegId::ONE)].into_iter().collect();
+    let (status, blob_id) = ctx.new_blob(bytecode.clone()).await.unwrap();
+    assert!(matches!(status, TransactionStatus::Success { .. }));
+
+    // Then
+    let result = ctx.query_blob(blob_id).await;
+    assert!(result.contains(&format!(r#""id":"0x{}"#, blob_id)));
+    assert!(result.contains(&format!(r#""bytecode":"0x{}""#, hex::encode(bytecode))));
+}
```
MitchTurner, line 186: [nit]: the SUT of this test would be the query.

```suggestion
#[tokio::test]
async fn blob__can_be_queried_if_uploaded() {
    // Given
    let mut ctx = TestContext::new().await;
    let bytecode: Vec<u8> = [op::ret(RegId::ONE)].into_iter().collect();
    let (status, blob_id) = ctx.new_blob(bytecode.clone()).await.unwrap();
    assert!(matches!(status, TransactionStatus::Success { .. }));

    // When
    let result = ctx.query_blob(blob_id).await;
    
    // Then
    assert!(result.contains(&format!(r#""id":"0x{}"#, blob_id)));
    assert!(result.contains(&format!(r#""bytecode":"0x{}""#, hex::encode(bytecode))));
}
```

netrome: Makes sense, thanks for suggesting!

crates/fuel-core/src/query/blob.rs
netrome: Offline discussion: backwards compatibility here is only relevant if the `contract_id` method would have been part of a public endpoint, which it is not - so I will do the refactor.

MitchTurner (dismissed):

xgreenx (changes requested):
We also need to add code to the `fuel-core-client` to query blobs and use it for e2e tests as well

netrome:
> We also need to add code to the `fuel-core-client` to query blobs and use it for e2e tests as well

Thanks for the pointer. I have added a query method in the client now and use it in the test instead of the previous custom query method.

xgreenx (approved):

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2094](https://github.com/FuelLabs/fuel-core/pull/2094): Added `--predefined-blocks-path` CLI argument to pass the path to the predefined blocks.
 - [2081](https://github.com/FuelLabs/fuel-core/pull/2081): Enable producer to include predefined blocks.
 - [2079](https://github.com/FuelLabs/fuel-core/pull/2079): Open unknown columns in the RocksDB for forward compatibility.
+- [2096](https://github.com/FuelLabs/fuel-core/pull/2096): GraphQL endpoint to fetch blob byte code by its blob ID.
```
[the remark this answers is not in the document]
netrome: Thanks for spotting, moved it.

Dentosal (approved):

MitchTurner (approved):

PR #2106, opened 2024-08-19
Handle the case when nodes with overriding start on the fresh network

xgreenx:
## Checklist

### Before requesting review

### DIFF
--- a/crates/fuel-core/src/database.rs
+++ b/crates/fuel-core/src/database.rs
@@ -133,9 +133,7 @@ impl OnChainIterableKeyValueView {
     }
 
     pub fn latest_height(&self) -> StorageResult<BlockHeight> {
-        Ok(self
-            .maybe_latest_height()?
-            .ok_or(not_found!("BlockHeight"))?)
+        self.maybe_latest_height()?.ok_or(not_found!("BlockHeight"))
     }
 
     pub fn latest_block(&self) -> StorageResult<CompressedBlock> {
--- a/crates/fuel-core/src/database.rs
+++ b/crates/fuel-core/src/database.rs
@@ -126,10 +126,16 @@ pub type ReyalerIterableKeyValueView = IterableKeyValueView<ColumnType<Relayer>>
 pub type GenesisDatabase<Description = OnChain> = Database<Description, GenesisStage>;
 
 impl OnChainIterableKeyValueView {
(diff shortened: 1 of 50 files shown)


Voxelot (approved):

crates/fuel-core/src/database.rs
```diff
@@ -126,10 +126,16 @@ pub type ReyalerIterableKeyValueView = IterableKeyValueView<ColumnType<Relayer>>
...
 impl OnChainIterableKeyValueView {
-    pub fn latest_height(&self) -> StorageResult<BlockHeight> {
+    pub fn maybe_latest_height(&self) -> StorageResult<Option<BlockHeight>> {
         self.iter_all_keys::<FuelBlocks>(Some(IterDirection::Reverse))
             .next()
-            .ok_or(not_found!("BlockHeight"))?
+            .transpose()
+    }
+
+    pub fn latest_height(&self) -> StorageResult<BlockHeight> {
```
MitchTurner, line 135: Can't we let the caller handle this case?

tests/tests/regenesis.rs
```diff
@@ -693,6 +693,47 @@ async fn starting_node_with_overwritten_old_poa_key_doesnt_rollback_the_state(
...
+            "--consensus-key",
+            original_secret_key.to_string().as_str(),
+        ],
+    )
+    .await;
+
+    // Then
+    let core = result.expect("Failed to start the node");
+    produce_block_with_tx(&mut rng, &core.client).await;
+
```
rymnc, line 733: shouldn't we assert here that it was a noop if the db was empty?

crates/fuel-core/src/database.rs
xgreenx: That method existed before, I don't want to affect other places=)

tests/tests/regenesis.rs
xgreenx: I didn't get your question=) Test verifies that we can start the node and produce one block=)

rymnc: my bad, thanks!

rymnc (dismissed):

Dentosal (dismissed):

MitchTurner (approved):

rymnc (approved):

PR #2109, opened 2024-08-20
Remove deadline clock in POA and replace with tokio time functions.

### LINKED ISSUE #1917: Investigate removing `DeadlineClock`

AurelienFT:
## Linked Issues/PRs
Closes #1917 

## Description
This PR removes the DeadlineClock structure and his usage in POA. It's replaced by a mechanism using Tokio features.
This changes avoid the service to hangs up on an error, it will now wait and retry a block production

## Checklist

### Before requesting review


Dentosal (dismissed):
Good cleanup. Removes code and adds tests. LGTM.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -517,12 +494,13 @@ where
...
-                    .set_timeout(block_time, OnConflict::Overwrite)
-                    .await;
+            Trigger::Interval { .. } => {
+                return Ok(Self {
+                    last_block_created: Instant::now(),
+                    ..self
+                })
             }
-        };
+        }
```
xgreenx, line 503: I think we even can remove this part since we have initialization for the `last_block_created` in the constructor=) 
image

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
```diff
@@ -194,6 +204,47 @@ async fn interval_trigger_produces_blocks_periodically() -> anyhow::Result<()> {
     Ok(())
 }
 
+// In case of error during block creation, the block should be retried after a delay of block time.
+#[tokio::test(start_paused = true)]
+async fn retry_block_creation_in_case_error() -> anyhow::Result<()> {
```
xgreenx, line 209: Thank you for the test!=) Often, tests with timeout fail because when we run all tests in parallel, the task scheduler could be busy. 

If we want to verify that we continue to produce blocks even in the case of error, maybe it is better to check that we called `produce_and_execute` method several times instead

crates/services/consensus_module/poa/src/service.rs
AurelienFT: It is not an issue that the constructor can be created long time ago before we make it into_task ?

xgreenx: I think it is okay, since the first block after restart could be produced immediately

AurelienFT: Ok done :)

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
AurelienFT: I have changed the test to be notify based instead of time based. I think we could follow this standard for the others test of POA that are in this file. Do you want me to make a separate issue ?

xgreenx: I think we can leave other tests for now as it is

xgreenx (dismissed):

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
```diff
@@ -194,6 +200,60 @@ async fn interval_trigger_produces_blocks_periodically() -> anyhow::Result<()> {
     Ok(())
 }
 
+// In case of error during block creation, the block should be retried after a delay of 1 second.
+#[tokio::test(start_paused = true)]
+async fn retry_block_creation_in_case_error() -> anyhow::Result<()> {
+    let config = Config {
+        trigger: Trigger::Interval {
+            block_time: Duration::new(2, 0),
```
MitchTurner, line 208: Shouldn't this test be that it triggers again after `2` seconds?

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
```diff
@@ -194,6 +200,60 @@ async fn interval_trigger_produces_blocks_periodically() -> anyhow::Result<()> {
     Ok(())
 }
 
+// In case of error during block creation, the block should be retried after a delay of 1 second.
+#[tokio::test(start_paused = true)]
+async fn retry_block_creation_in_case_error() -> anyhow::Result<()> {
```
MitchTurner, line 205: I know the other tests in this file don't do this yet, but can we add a `given`/`when`/`then` to clearly mark the SUT, the conditions, and the expected result?

It's not super clear here anyway, since we don't get to handle the service directly (which I would assume is the SUT here). But at least say:
`given`: a service that just failed to commit a block
`when`: we wait for 2 seconds
`then`: the service succeeds at committing a block

MitchTurner, line 205: [nit]: I don't think all the comments are necessary. 

For example, the comment above the test name could just be the test name. I don't mind verbose test names:
```rs
async fn service__if_commit_result_fails_then_retry_commit_result_after_trigger_duration
```

I think the comments above the `expect_commit_result`s are good. But a lot of the other comments are redundant.

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
AurelienFT: With @xgreenx we decided that we will not wait `block_time` to recreate a block but a 1 second hard number

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
AurelienFT: Ok, I removed the others comments !

AurelienFT: Reorganized it.

xgreenx (dismissed):

xgreenx (approved):

MitchTurner (approved):

Issue #2023, opened 2024-07-05
P2P is doing a lot of database lookups


xgreenx:
When we request transactions for the block, each transaction requires a separate lookup, which is very expensive, considering that transactions are randomly distributed throughout the database(because of randomness of `TxId`).

<img width="1657" alt="image" src="https://github.com/FuelLabs/fuel-core/assets/18346821/5c57b8dc-004a-4ceb-9b05-34b62c5fae14">

<img width="1667" alt="image" src="https://github.com/FuelLabs/fuel-core/assets/18346821/b280fd67-69d6-4acc-b452-a338b728f646">

Voxelot:
Without changing the structure of the db to store full blocks instead of compact blocks, the most practical improvement with minimal changes would be to change this to a multi-get operation.

AurelienFT:
I'm taking it as a side quest.

rymnc:
Taking this up :)

rymnc:
This is going to be partially solved with the caching layer in https://github.com/FuelLabs/fuel-core/pull/2352

Issue #808, opened 2022-12-01
Fuel-P2P Metrics


ControlCplusControlV:
Also look at gossipsub metrics config and use real parameters

**Fuel-p2p**
- [ ] - Track maybe amount of p2p messages
- [ ] - Unique peers connected to
- [ ] - Track Transactions from peers

AurelienFT:
@rafal-ch can we consider this as done or there is more things to add ?

rafal-ch:
> @rafal-ch can we consider this as done or there is more things to add ?

I'm not sure, tbh. @rymnc added quite a lot of P2M metrics in [this commit](https://github.com/FuelLabs/fuel-core/pull/2310/commits/76b10848a64cf40d3ca537a23ebb7c2a548ff9b1), but they seem to be more "low-level" than the ones mentioned in the description. For example, I don't see a metric for "Track Transactions from peers".

We should double-check with @rymnc or discuss on the grooming.

rymnc:
we can close this issue as p2p metrics have been added and transactions are tracked via txpool metrics

PR #2112, opened 2024-08-20
chore(sealed_blocks): optimize fetching of sealed block header at a given height

rymnc:
## Linked Issues/PRs

- #2023

## Description

We don't need to fetch the header *and* the consensus header from the storage if either one of them is `None`. If `consensus` evaluates to `None`, we can early return a `None`, but if the consensus header exists, we can then fetch the block header.

Performs another optimization in instances where we fetch the `SealedBlock` and only end up using the `entity` wrapped within it, and not the consensus headers. Previously, both the consensus headers and the block were fetched from the db and then returned as a `SealedBlock`, Now, we just check for existence of a given block height in the `SealedBlockConsensus` table, and avoid the allocation for the consensus headers. 

> note: also restricts how many blocks worth of transactions can be requested over p2p

## Checklist

### Before requesting review

- [ ] Someone else?


rymnc:
added block range restriction in 0a7b23af6 since the CI runners are struggling with all the open PRs

rymnc:
p2p metrics issue: https://github.com/FuelLabs/fuel-core/issues/808

crates/fuel-core/src/database/sealed_block.rs
```diff
@@ -51,6 +54,23 @@ impl OnChainIterableKeyValueView {
         }
     }
 
+    /// Returns a `Block`, which is sealed, by height
+    pub fn get_block_by_height(
+        &self,
+        height: &BlockHeight,
+    ) -> StorageResult<Option<Block>> {
+        if self
```
xgreenx, line 62: Why do we want to do an additional lookup into `SealedBlockConsensus` table? We can go directly to the `self.get_full_block`

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - [2106](https://github.com/FuelLabs/fuel-core/pull/2106): Remove deadline clock in POA and replace with tokio time functions.
+- [2112](https://github.com/FuelLabs/fuel-core/pull/2112): Alter the way the sealed blocks are fetched with a given height.
```
xgreenx, line 16: We need to move it to `## [Unreleased]` section

bin/fuel-core/src/cli/run/p2p.rs
```diff
@@ -66,6 +66,10 @@ pub struct P2PArgs {
     #[clap(long = "max-headers-per-request", default_value = "100", env)]
     pub max_headers_per_request: u32,
 
+    /// Max number of blocks in a transaction request response
+    #[clap(long = "max-blocks-per-tx-request", default_value = "100", env)]
+    pub max_blocks_per_tx_request: u32,
```
xgreenx, line 71: I think we can reuse `max_headers_per_request` to reduce number of arguments

crates/fuel-core/src/database/sealed_block.rs
```diff
@@ -51,6 +54,23 @@ impl OnChainIterableKeyValueView {
         }
     }
 
+    /// Returns a `Block`, which is sealed, by height
```
netrome, line 57: I find this docstring a bit confusing as we don't return a `SealedBlock` in this function. We chatted offline about this, and the important thing here is that the block has been sealed (i.e. it is signed) but downstream we only need the actual block.

```suggestion
    /// Returns a `Block` by it's height. Only returns a block if it has been sealed.
```

netrome: New information based on @xgreenx's comment below and an offline discussion. The block importer will only persist blocks that are sealed so we can omit this from the docstring altogether
```suggestion
    /// Returns a `Block` by its height.
```

crates/fuel-core/src/database/sealed_block.rs
rymnc: addressed in 5498c5fd9

bin/fuel-core/src/cli/run/p2p.rs
rymnc: addressed in https://github.com/FuelLabs/fuel-core/pull/2112/commits/06262b9b8c210ac8b485d96050f52a54d96d18f4

CHANGELOG.md
rymnc: addressed in https://github.com/FuelLabs/fuel-core/pull/2112/commits/da650f4ebe72bb42003d859750eb84cbf26a2271

crates/services/p2p/src/service.rs
```diff
@@ -437,6 +437,20 @@ where
...
+                if range.len() > max_len {
+                    tracing::error!("Requested range of blocks is too big. Requested length: {:?}, Max length: {:?}", range.len(), max_len);
+                    let response = None;
+                    let _ = self.p2p_service.send_response_msg(
+                        request_id,
+                        ResponseMessage::Transactions(response),
+                    );
+                    return Ok(())
+                }
+
```
netrome, line 453: Nit: This is a very long match arm. While it was pretty long already before this PR, I'd find the code easier to read if we could break this out to a separate function. If we use `tracing::instrument` on that function we could have a cleaner error message, where the lengths are automatically added to the span.

MitchTurner (approved):

crates/services/p2p/src/service.rs
rymnc: https://github.com/FuelLabs/fuel-core/issues/2121

netrome: Thanks for opening!

netrome (approved):
Just a few nits and enhancement suggestions from me, otherwise the PR looks good!

crates/services/p2p/src/service.rs
```diff
@@ -437,6 +437,20 @@ where
         let response_channel = self.request_sender.clone();
         match request_message {
             RequestMessage::Transactions(range) => {
+                let max_len = self
+                    .max_blocks_per_request
+                    .try_into()
+                    .expect("u32 should always fit into usize");
+                if range.len() > max_len {
+                    tracing::error!("Requested range of blocks is too big. Requested length: {:?}, Max length: {:?}", range.len(), max_len);
```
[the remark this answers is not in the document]
rymnc: addressed in ebca20dd5, with the other usage of this as well

MitchTurner (approved):

MitchTurner (approved):

crates/services/p2p/src/service.rs
```diff
@@ -437,6 +437,24 @@ where
         let response_channel = self.request_sender.clone();
         match request_message {
             RequestMessage::Transactions(range) => {
+                let max_len = self
```
xgreenx, line 440: We can use `u32` instead of `usize` to avoid `try_into`

xgreenx (approved):

Issue #2107, opened 2024-08-19
Failed to estimate gas price for block, algorithm not yet set: 10


arboleya:
It appears that after version 0.32, executing the GraphQL query `estimateGasPrice(10)` immediately after starting fuel-core will often throw the error below, making tests very flaky on [`fuels-ts`](https://github.com/FuelLabs/fuels-ts).

```console
Failed to estimate gas price for block, algorithm not yet set: 10
```

Quoting suspicions below:

> @MitchTurner: I believe this is a race condition where the graphql services task is started before the gas price service is fully initialized. Each is run in a separate thread and we don't have any extra synchronization to make sure one is started before the other.

> @Voxelot: hmm we need the graphql service up and running early otherwise the service may look unhealthy if other tasks take a while to initialize. We could split the graphql request handling from the health endpoint so they could be initialized at different times? (edited)

PR #2113, opened 2024-08-20
Remove option on shared algo, give default algo, reorder service startup

MitchTurner:
## Linked Issues/PRs
Closes: https://github.com/FuelLabs/fuel-core/issues/2107

## Description
This code will help ensure that some gas price algorithm is provided to the consumers. We do this by removing the `Option` and always initializing with our best value. We also prioritize the initialization of the gas price service.

## Checklist

### Before requesting review

- [ ] [TS SDK](https://github.com/FuelLabs/fuels-rs/)


crates/fuel-core/src/service/sub_services.rs
```diff
@@ -332,11 +332,10 @@ pub fn init_sub_services(
     #[allow(unused_mut)]
     // `FuelService` starts and shutdowns all sub-services in the `services` order
     let mut services: SubServices = vec![
-        // GraphQL should be shutdown first, so let's start it first.
+        Box::new(gas_price_service),
         Box::new(graph_ql),
```
xgreenx, line 336: We can move `graph_ql` to the off-chian worker and add it there.

xgreenx (commented):
LGTM!=)

rymnc (dismissed):
LGTM

xgreenx (approved):

rymnc (approved):

crates/services/gas_price_service/src/lib.rs
```diff
@@ -79,7 +79,7 @@ where
 }
 
 #[derive(Debug, Default)]
-pub struct SharedGasPriceAlgo<A>(Arc<RwLock<Option<A>>>);
+pub struct SharedGasPriceAlgo<A>(Arc<RwLock<A>>);
```
netrome, line 82: Nit (unrelated to this PR): I found it a bit confusing to read this struct, since the declaration has nothing to do with the fact that the content is a gas price algorithm. Since the type is very generic, I'd find it easier to read if it was just named `Shared`.

netrome (approved):
Looks mostly good to me.

However, I would prefer to avoid returning arbitrary default values in favor of using stronger typing. I.e. if a value can be unset we should return an Option as per the previous behavior.

I understand that we implicitly guarantee by the changed startup order of the sub services that the default values will never be exposed to a caller, this means that the default logic introduced in this PR is dead code - and it would be nice if we could get rid of it.

Since that may be a very complex change, and perhaps not worth prioritizing right now I am alright with leaving things as-is for now. But if anyone thinks it might be feasible to look into this, it would be nice to create a follow-up ticket for it.

PR #2119, opened 2024-08-20
feat: Add graphql query fields for retreiving information about upgrades

### LINKED ISSUE #2071: Add new GraphQL endpoints for retrieving information about upgrades
We support forkless upgrades via `Upgrade` and `Upload` transactions. When these transactions are submitted, they change the state of corresponding tables in the database in `fuel-core`(`StateTransitionBytecodeVersions`, `ConsensusParametersVersions`, `UploadedBytecodes`). Each block specifies the version of the consensus parameters and state transition function that was used to do a transition.

Currently, there is no way to check what consensus parameter or state transition function corresponds to the version except by going over all blocks and finding upgrade transactions.

We need to add new endpoints:
- Retrieve the consensus parameters by its version
- Retrieve the state transition bytecode and root by its version
- Retrieve the state transition bytecode by its root

netrome:
## Linked Issues/PRs
Closes #2071  

## Description
This PR adds the following three new query fields

- consensusParameters(version: Int!): ConsensusParameters!
- stateTransitionBytecodeByVersion(version: Int!): StateTransitionBytecode
- stateTransitionBytecodeByRoot(root: HexString!): StateTransitionBytecode!

which help provide more insights in previous upgrades on the chain.

## Checklist

### Before requesting review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2112](https://github.com/FuelLabs/fuel-core/pull/2112): Alter the way the sealed blocks are fetched with a given height.
 - [2115](https://github.com/FuelLabs/fuel-core/pull/2115): Add test for `SignMode` `is_available` method.
 
+### Added
+- [2119](https://github.com/FuelLabs/fuel-core/pull/2119): GraphQL query fields for retrieving information about upgrades.
 
 ## [Version 0.34.0]
 
@@ -18,7 +20,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2051](https://github.com/FuelLabs/fuel-core/pull/2051): Add support for AWS KMS signing for the PoA consensus module. The new key can be specified with `--consensus-aws-kms AWS_KEY_ARN`.
 - [2092](https://github.com/FuelLabs/fuel-core/pull/2092): Allow iterating by keys in rocksdb, and other storages.
 - [2096](https://github.com/FuelLabs/fuel-core/pull/2096): GraphQL query field to fetch blob byte code by its blob ID.
-- [2119](https://github.com/FuelLabs/fuel-core/pull/2119): GraphQL query fields for retrieving information about upgrades.
 
 ### Changed
 - [2106](https://github.com/FuelLabs/fuel-core/pull/2106): Remove deadline clock in POA and replace with tokio time functions.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - [2051](https://github.com/FuelLabs/fuel-core/pull/2051): Add support for AWS KMS signing for the PoA consensus module. The new key can be specified with `--consensus-aws-kms AWS_KEY_ARN`.
 - [2092](https://github.com/FuelLabs/fuel-core/pull/2092): Allow iterating by keys in rocksdb, and other storages.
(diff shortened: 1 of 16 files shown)


CHANGELOG.md
```diff
@@ -11,7 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - [2051](https://github.com/FuelLabs/fuel-core/pull/2051): Add support for AWS KMS signing for the PoA consensus module. The new key can be specified with `--consensus-aws-kms AWS_KEY_ARN`.
 - [2092](https://github.com/FuelLabs/fuel-core/pull/2092): Allow iterating by keys in rocksdb, and other storages.
-- [2096](https://github.com/FuelLabs/fuel-core/pull/2096): GraphQL endpoint to fetch blob byte code by its blob ID.
+- [2096](https://github.com/FuelLabs/fuel-core/pull/2096): GraphQL query field to fetch blob byte code by its blob ID.
```
netrome, line 14: Note: I just renamed "endpoint" to "query field" here as I realized these are technically not endpoints.

crates/client/assets/schema.sdl
```diff
@@ -1242,6 +1250,12 @@ scalar U64
 
 union UpgradePurpose = ConsensusParametersPurpose | StateTransitionPurpose
 
+type UploadedBytecode {
+	bytecode: HexString!
+	uploadedSubsectionsNumber: Int!
```
rymnc, line 1255: not too clear if this means - number of subsections uploaded, or the index of the uploaded subsections

crates/fuel-core/src/query/block.rs
```diff
@@ -46,7 +46,7 @@ pub trait BlockQueryData: Send + Sync + SimpleBlockData {
 
 impl<D> BlockQueryData for D
 where
-    D: OnChainDatabase + DatabaseBlocks + ?Sized,
```
rymnc, line 49: why are we removing DatabaseBlocks from here?

netrome: Two answers depending on the interpretation of the question:

_Why is it valid to remove DatabaseBlocks from here?_
Because the `OnChainDatabase` trait already requires `DatabaseBlocks` (see `crates/fuel-core/src/graphql_api/ports.rs`)

_Why are we removing DatabaseBlocks from here **in this PR**?_
Because of the "Boy Scout rule" - This was a small improvement that I preferred to clean up as I spotted it rather than create a standalone PR or issue for it.

Though looking at this again, I think `DatabaseBlocks` is the only required trait here so I think we can make this even cleaner by dropping the `OnChainDatabase` requirement here instead.

crates/client/assets/schema.sdl
netrome: Good point. I'll add a docstring so it is clear that it is the former interpretation we refer to.

netrome: Done. I also made this field optional since it should not be set if the upload is completed.

crates/fuel-core/src/query/block.rs
netrome: Changed this so we're removing `OnChainDatabase` instead of `DatabaseBlocks`.

rymnc: nice, didn't see that ~ thanks!

crates/client/assets/schema.sdl
rymnc: nice

Dentosal (dismissed):

Dentosal (approved):

rymnc (approved):

Issue #2118, opened 2024-08-20
Fetch L1 data from different RPC providers


xgreenx:
To resolve the https://github.com/FuelLabs/security-drills/issues/10, we need to create a service that can fetch L1 data from different sources.

The service should compare results from different sources and return an error if 51% can't be reached.

## Implementation details

The service should be independent and reusable. We should be able to use this service in the `fuel-core`, and we should be able to create a separate binary that uses this single service that implements the required methods of the Ethereum RPC.

xgreenx:
> A simpler option could be to use the QuorumProvider in ethers-rs
https://docs.rs/ethers/latest/ethers/providers/struct.QuorumProvider.html

AurelienFT:
As a first part of this issue, #2122 is changing the behaviour of the relayer to allow multiple URI and will accept the result of an RPC call only if 50% providers or more gives the same answer.

The next step would be to have a separate project that could be run as a service along with the core, block-committer, etc... that will implement this multiple RPC logic.

PR #2122, opened 2024-08-21
Change relayer URI address to be  vector and use quorum provider

AurelienFT:
## Linked Issues/PRs
Closes https://github.com/FuelLabs/fuel-core/issues/2118

## Description
This PR include the change to allow multiple RPC directly in our relayer code. It allows relayer users to pass multiple RPC addresses in CLI and the response to a RPC call will be accepted only if it's the same given by 50% or more of the RPC provider.

It shouldn't be a breaking change as the current format of the `--relayer` argument will still be valid and will result in a simple RPC call.

## Checklist

### Before requesting review

### DIFF
--- a/crates/services/relayer/src/service/test.rs
+++ b/crates/services/relayer/src/service/test.rs
@@ -143,7 +143,7 @@ async fn quorum__disagree_on_logs() {
     )
     .map_ok(|logs| logs.logs)
     .try_concat()
-    .await
+    .await;
     // Then
 
     match provider_error {


rymnc:
what is the behaviour of the quorum provider when a majority of the provided RPC are down? does it return an error, or just the minority response?

crates/services/relayer/src/config.rs
```diff
@@ -22,8 +22,8 @@ pub(crate) static ETH_FORCED_TX: Lazy<H256> =
 pub struct Config {
     /// The da block to which the contract was deployed.
     pub da_deploy_height: DaBlockHeight,
-    /// Uri address to ethereum client.
-    pub relayer: Option<url::Url>,
+    /// Uri addresses to ethereum client.
+    pub relayer: Option<Vec<url::Url>>,
```
xgreenx, line 26:
```suggestion
    pub relayer: Vec<url::Url>,
```

tests/tests/relayer.rs
```diff
@@ -143,6 +141,79 @@ async fn relayer_can_download_logs() {
     eth_node_handle.shutdown.send(()).unwrap();
 }
 
+#[tokio::test(flavor = "multi_thread")]
```
xgreenx, line 144: It would be nice to have a test when two endpoints return the different result for the same request

tests/tests/relayer.rs
```diff
@@ -143,6 +141,79 @@ async fn relayer_can_download_logs() {
...
+
+    let eth_node = Arc::new(eth_node);
+    let eth_node_handle = spawn_eth_node(eth_node).await;
+
+    relayer_config.relayer = Some(vec![
+        format!("http://{}", eth_node_handle.address)
+            .as_str()
+            .try_into()
+            .unwrap(),
+        "http://localhost:8080".try_into().unwrap(),
```
MitchTurner, line 159: [nit]: Maybe move this to a var with name `unresponsive_eth_node_address` or something like that.

tests/tests/relayer.rs
```diff
@@ -143,6 +141,79 @@ async fn relayer_can_download_logs() {
...
+        "http://localhost:8081".try_into().unwrap(),
```
MitchTurner, line 192: Same thing here.

tests/tests/relayer.rs
```diff
@@ -143,6 +141,79 @@ async fn relayer_can_download_logs() {
...
+    // Given
+    let db = Database::in_memory();
+    let srv = FuelService::from_database(db.clone(), config)
+        .await
+        .unwrap();
+
+    // When
+    srv.await_relayer_synced().await.unwrap();
+
+    // Then
```
MitchTurner, line 171: Maybe show that the sync `is_ok`, just so we don't have shutdown stuff confused as the expected result.

crates/services/relayer/src/config.rs
netrome: Can we rename this to `relayers`, or would that entail a lot of downstream work since we'd break potential existing configurations?

crates/services/relayer/src/service.rs
```diff
@@ -348,14 +351,21 @@ pub fn new_service<D>(database: D, config: Config) -> anyhow::Result<Service<D>>
...
+        .relayer
+        .clone()
+        .ok_or_else(|| {
+            anyhow::anyhow!(
+                "Tried to start Relayer without setting an eth_client in the config"
+            )
+        })?
+        .into_iter()
+        .map(|url| {
+            // TODO: Does this handle https?
```
netrome, line 364: Has this been resolved yet? Otherwise it sounds like we should create a follow-up issue to look into this.

tests/tests/relayer.rs
```diff
@@ -143,6 +141,79 @@ async fn relayer_can_download_logs() {
...
+    // When
+    tokio::select! {
+        _ = srv.await_relayer_synced() => {
+            panic!("Should not have synced");
+        }
+        _ = tokio::time::sleep(max_timeout) => {
+            // Then
+            // We should not have synced
+        }
+    }
```
netrome, line 211: Can we move this out so the "Then" section is on the same level as "Given" and "When"? Something like:

```suggestion
    // When
    has_synced = tokio::select! {
        _ = srv.await_relayer_synced() => {
            true
        }
        _ = tokio::time::sleep(max_timeout) => {
            false
        }
    }
    
    // Then
    assert!(!has_synced);
```

crates/services/relayer/src/config.rs
Voxelot: This command should probably should've been called something like `eth-rpc-urls` since theres only one relayer running inside of fuel-core. But it's a breaking change that would impact a lot of people.

netrome: Thanks, yeah that's what I was afraid of. A non breaking variant would be to add another arg with the intended name and mark this one as deprecated in its docstring. But I suppose we can await doing that until we have more fields to rename.

AurelienFT:
> what is the behaviour of the quorum provider when a majority of the provided RPC are down? does it return an error, or just the minority response?

It returns an error `NoQuorumReached` with the values and the errors gathered. I will try to change the tests to highlight more this behavior

tests/tests/relayer.rs
AurelienFT: I just added it in the new tests I just made :)

tests/tests/relayer.rs
AurelienFT: Closing because I have re-do all the tests

tests/tests/relayer.rs
AurelienFT: Closing because I have re-do all the tests

tests/tests/relayer.rs
AurelienFT: Closing because I have re-do all the tests

crates/services/relayer/src/service.rs
AurelienFT: @xgreenx are we already using https somewhere and we can validate this or do you want me to test it ?

xgreenx (dismissed):

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- [2122](https://github.com/FuelLabs/fuel-core/pull/2122): Change relayer URI address to be vector and use quorum provider
```
xgreenx, line 10:
```suggestion
- [2122](https://github.com/FuelLabs/fuel-core/pull/2122): Changed the relayer URI address to be a vector and use a quorum provider. The `relayer` argument now supports multiple URLs to fetch information from different sources.
```

xgreenx (approved):

crates/services/relayer/src/service.rs
Voxelot: yes we are already using https in prod

netrome: Nice, then I assume we can remove the comment.

AurelienFT: Done :)

crates/services/relayer/src/service/test.rs
```diff
@@ -45,6 +45,118 @@ async fn can_download_logs() {
...
+
+    match provider_error {
+        ProviderError::JsonRpcClientError(e) => {
+            assert_eq!(format!("{e}"), "No Quorum reached.");
+        }
+        _ => {
+            panic!("Expected a JsonRpcClientError")
+        }
+    }
+}
```
netrome, line 158: Nit: It's slightly awkward to have the "//Then" section begin in the middle of an expression. I'd suggest just matching on the error in the match clause:

```suggestion
    // Then

    match provider_error {
        Err(ProviderError::JsonRpcClientError(e)) => {
            assert_eq!(format!("{e}"), "No Quorum reached.");
        }
        _ => {
            panic!("Expected a JsonRpcClientError")
        }
    }
}
```

netrome (approved):
Just one nit, otherwise LGTM.

crates/services/relayer/src/service/test.rs
AurelienFT: I'm ok with this

xgreenx (approved):

netrome (approved):

xgreenx (approved):

PR #2127, opened 2024-08-26
Added a benchmark for the predicate with ed19 verification

xgreenx:
### Before requesting review

### DIFF
--- a/benches/benches/transaction_throughput.rs
+++ b/benches/benches/transaction_throughput.rs
@@ -8,6 +8,7 @@ use criterion::{
     Criterion,
     SamplingMode,
 };
+use ed25519_dalek::Signer;
 use fuel_core::service::config::Trigger;
 use fuel_core_benches::*;
 use fuel_core_storage::transactional::AtomicView;
@@ -49,6 +50,7 @@ use std::{
     time::Duration,
 };
 use test_helpers::builder::{
+    local_chain_config,
     TestContext,
     TestSetupBuilder,
 };
@@ -85,6 +87,7 @@ where
             // disable automated block production
             test_builder.trigger = Trigger::Never;
             test_builder.utxo_validation = true;
+            test_builder.gas_limit = Some(10_000_000_000);
 
             // spin up node
             let transactions: Vec<Transaction> =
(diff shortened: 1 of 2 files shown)


benches/benches/transaction_throughput.rs
```diff
@@ -281,17 +288,90 @@ fn predicate_transfers_eck1(c: &mut Criterion) {
...
             .finalize();
-        tx.estimate_predicates(&CheckPredicateParams::default(), MemoryInstance::new())
+        tx.estimate_predicates(&checked_parameters(), MemoryInstance::new())
             .expect("Predicate check failed");
         tx
     };
     bench_txs("predicate transfers eck1", c, generator);
 }
 
+fn predicate_transfers_ed19(c: &mut Criterion) {
```
netrome, line 298: This function looks very similar to `predicate_transfers_eck1`. I assume the intention is for them to be exactly the same but with different signing schemes. It would be nice to have the code reflect this by breaking out the shared logic to a separate function.

If you agree, I'd be happy to create a ticket and do that as a follow-up.

netrome (approved):
Looks good to me.

benches/benches/transaction_throughput.rs
MitchTurner: Exactly. The idea here is to benchmarks a bunch of simple transactions to get an idea of max throughput for our system. In this case we are including a predicate that includes
```
    op::ed19(0x20, 0x21, 0x22, 0x24),
```
The entry point for these benchmarks is the `criterion_group!` macro call below. I think the differences are moderately understandable looking there:
```
criterion_group!(
    benches,
    signed_transfers,
    predicate_transfers,
    predicate_transfers_eck1
    predicate_transfers_ed19
);
```
But I'll never argue with adding more clarity to code if things are unclear.

MitchTurner (approved):

benches/benches/transaction_throughput.rs
netrome: Right, thanks for clarifying. As long as it's just two signature schemes, a refactor of these functions may be a bit premature.

PR #2129, opened 2024-08-26
Use more realistic values in gas price analysis

MitchTurner:
## Checklist

### Before requesting review

- [ ] Someone else?


crates/fuel-gas-price-algorithm/gas-price-analysis/src/charts.rs
```diff
@@ -11,44 +11,44 @@ pub fn draw_gas_prices(
...
+    // chart
+    //     .draw_series(LineSeries::new(
+    //         exec_gas_prices.iter().enumerate().map(|(x, y)| (x, *y)),
+    //         EXEC_GAS_PRICE_COLOR,
+    //     ))
+    //     .unwrap()
+    //     .label("Exec Gas Price")
+    //     .legend(|(x, y)| {
+    //         PathElement::new(vec![(x, y), (x + 20, y)], EXEC_GAS_PRICE_COLOR)
+    //     });
```
netrome, line 51: Can we remove this? Or is it useful to be able to toggle this? Then perhaps these sections could be broken out to separate functions which could be conditionally invoked based on some flag for example.

crates/fuel-gas-price-algorithm/gas-price-analysis/src/main.rs
```diff
@@ -25,8 +26,21 @@ mod simulation;
...
+    let num = input
+        .to_string()
+        .as_bytes()
+        .rchunks(3)
+        .rev()
+        .map(std::str::from_utf8)
+        .collect::<Result<Vec<&str>, _>>()
+        .unwrap()
+        .join(","); // separator
+    num
```
netrome, line 39: Nit:
```suggestion
    input
        .to_string()
        .as_bytes()
        .rchunks(3)
        .rev()
        .map(std::str::from_utf8)
        .collect::<Result<Vec<&str>, _>>()
        .unwrap()
        .join(",") // separator
```

crates/fuel-gas-price-algorithm/gas-price-analysis/src/main.rs
```diff
@@ -35,35 +49,41 @@ fn main() {
...
         da_gas_prices,
         fullness,
+        bytes_and_costs,
         actual_profit,
         projected_profit,
         pessimistic_costs,
     } = best;
 
-    let plot_width = 640 * 2;
+    // let min_actual_profit = pretty(actual_profit.iter().min().unwrap());
```
netrome, line 58: Clean up?

crates/fuel-gas-price-algorithm/gas-price-analysis/src/optimisation.rs
```diff
@@ -1,14 +1,16 @@
...
-        })
-        .collect()
+    // let mut rng = StdRng::seed_from_u64(10902);
+    // (0usize..size)
+    //     .map(|_| {
+    //         let p = rng.gen_range(100_000..5_000_000);
+    //         let d = rng.gen_range(100_000..5_000_000);
+    //         (p, d)
+    //     })
+    //     .collect()
```
netrome, line 11: Can we clean this up? Or is this helpful to keep around as a reference?

crates/fuel-gas-price-algorithm/gas-price-analysis/src/optimisation.rs
```diff
@@ -20,7 +22,8 @@ pub fn naive_optimisation(
         .map(|(p, d)| (run_simulation(*p, *d, avg_window), (*p, *d, avg_window)))
         .min_by_key(|(results, _)| {
             let SimulationResults { actual_profit, .. } = results;
-            actual_profit.iter().map(|p| p.abs()).sum::<i64>()
+            let err = actual_profit.iter().map(|p| p.abs()).sum::<i64>();
```
netrome, line 25: I'm not following why we take the absolute value here? Are we interested in the total sum of the deviation between the reward and cost? Wouldn't we want an average here in that case?

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
```diff
@@ -196,14 +208,30 @@ fn fullness_and_bytes_per_block(size: usize, capacity: u64) -> Vec<(u64, u64)> {
...
+fn noisy_eth_price<T: TryInto<f64>>(input: T) -> f64
+where
+    <T as TryInto<f64>>::Error: core::fmt::Debug,
+{
+    const COMPONENTS: &[f64] = &[3.0, 4.0];
+    let input = input.try_into().unwrap();
+    gen_noisy_signal(input, COMPONENTS)
+}
+
+fn arb_cost_per_byte(size: usize, update_period: usize) -> Vec<u64> {
```
netrome, line 224: What is `arb` short for here? "arbitrary"?

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -66,45 +66,30 @@ pub struct AlgorithmV1 {
     /// The D component of the PID control for the DA gas price
     da_d_factor: i64,
     /// The average profit over the last `avg_window` blocks
-    avg_profit: i64,
-    /// The number of blocks to consider when calculating the average profit
-    avg_window: u32,
+    last_profit: i64,
+    /// the previous profit
+    last_last_profit: i64,
```
netrome, line 71: I find the naming a bit confusing here. How about something like:

```suggestion
    /// The average profit over the second last `avg_window` blocks
    second_last_profit: i64,
```

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -176,10 +163,10 @@ pub struct AlgorithmUpdaterV1 {
...
     /// The D component of the PID control for the DA gas price
     pub da_d_component: i64,
-    /// The average profit over the last `avg_window` blocks
-    pub profit_avg: i64,
-    /// The number of blocks to consider when calculating the average profit
-    pub avg_window: u32,
+    /// The last profit
+    pub last_profit: i64,
+    /// The profit before last
+    pub last_last_profit: i64,
```
netrome, line 169: Also I think we can improve the naming here:
```suggestion
    pub last_profit: i64,
    /// The second last profit
    pub second_last_profit: i64,
```

netrome (approved):
Some naming suggestions and a bunch of questions from me.

Unrelated to this particular PR I find it interesting that we're using integer arithmetic throughout the entire algorithm. I would have expected this type of algorithm to be computed using floats internally, and then discretize the resulting DA price change before returning it. This would give us more fine grained control over the dynamics if we want to further tune the P and D values (and a potential I value if we want to include that down the line).

crates/fuel-gas-price-algorithm/gas-price-analysis/src/optimisation.rs
MitchTurner: We want to keep this. It's how we generate new values. I have followup issues that will use CLI commands to activate this section.

crates/fuel-gas-price-algorithm/gas-price-analysis/src/charts.rs
MitchTurner: It's noise right now since, but I think we should keep it for long-term.

crates/fuel-gas-price-algorithm/gas-price-analysis/src/optimisation.rs
netrome: Nevermind, I misread this. We are never returning this value, as this is just used in `min_by_key` to find the parameters with the minimal profitability deviation.

netrome (approved):

netrome (dismissed):

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
```diff
@@ -180,7 +190,7 @@ fn fullness_and_bytes_per_block(size: usize, capacity: u64) -> Vec<(u64, u64)> {
     let bytes_scale: Vec<_> = std::iter::repeat(())
         .take(size)
         .map(|_| rng.gen_range(0.5..1.0))
-        .map(|x| x * 4.0)
+        .map(|x| x * 0.01)
```
rymnc, line 193:
```suggestion
        .map(|x| x * BYTES_SCALE_INC)
```
it would be nice to avoid use of magic numbers here

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
```diff
@@ -196,14 +206,29 @@ fn fullness_and_bytes_per_block(size: usize, capacity: u64) -> Vec<(u64, u64)> {
...
+    gen_noisy_signal(input, COMPONENTS)
+}
+
+fn arbitrary_cost_per_byte(size: usize, update_period: usize) -> Vec<u64> {
+    let actual_size = size.div_ceil(update_period);
+
+    (0u32..actual_size as u32)
         .map(noisy_eth_price)
-        .map(|x| x * 100. + 110.)
+        .map(|x| x * 5. + 5.)
```
rymnc, line 227: same as prev comment

MitchTurner:
> Unrelated to this particular PR I find it interesting that we're using integer arithmetic throughout the entire algorithm. I would have expected this type of algorithm to be computed using floats internally, and then discretize the resulting DA price change before returning it. This would give us more fine grained control over the dynamics if we want to further tune the P and D values (and a potential I value if we want to include that down the line).

You're right. We might have been able to get away with floats here and it would be a lot easier. Something I didn't realize when I started the algorithm, but I do now, Is that this algorithm will be completely off-chain. Meaning the PoA node can use whatever values she chooses and it won't be validated on-chain. 

In the future, we _will_ have gas price as part of the state transition function. This will allow it to be part of fraud proving and later, when/if we move to PoS, be verified by other nodes during consensus. Float arithmetic is not deterministic, and that is why I went with integer math because I assumed this need to be run deterministically as part of the state transition function.

So, technically, right now it is YAGNI. So we could move to floats. In the future, it needs to be fully deterministic, which floats are not. That's why it's integers.

rymnc (approved):

netrome (approved):

PR #2130, opened 2024-08-26
feat: build ARM and AMD in parallel

mchristopher:
## Description

Building the ARM and AMD Docker images using native runners in parallel. This will greatly speed up the CI runtime.

### Before requesting review


.github/workflows/ci.yml
```diff
@@ -332,17 +335,127 @@ jobs:
...
+          path: /tmp/digests/*
+          if-no-files-found: error
+          retention-days: 1
+
   publish-docker-image:
     needs:
+      - build-docker-images
       - verifications-complete
-    runs-on: buildjet-4vcpu-ubuntu-2204
+    runs-on: ubuntu-latest
```
xgreenx, line 440: @Voxelot Do we want to use `buildjet` or switch back to github runners?

Voxelot: we should probably stick with buildjet for CI performance reasons

cc @mchristopher

mchristopher: FYI, this moves back to basic runners for tasks that are not CPU or memory intensive (like downloading/uploading files). It's an optimization to make sure we're not spending more money than needed on buildjet.

deployment/Dockerfile
```diff
@@ -36,17 +36,28 @@ COPY --from=planner /build/recipe.json recipe.json
...
+RUN \
+  --mount=type=cache,target=/usr/local/cargo/registry/index \
+  --mount=type=cache,target=/usr/local/cargo/registry/cache \
+  --mount=type=cache,target=/usr/local/cargo/git/db \
+  --mount=type=cache,target=/build/target \
+  xx-cargo build --release --no-default-features --features "$BUILD_FEATURES" -p fuel-core-bin \
     && xx-verify ./target/$(xx-cargo --print-target-triple)/release/fuel-core \
-    && mv ./target/$(xx-cargo --print-target-triple)/release/fuel-core ./target/release/fuel-core \
-    && mv ./target/$(xx-cargo --print-target-triple)/release/fuel-core.d ./target/release/fuel-core.d
+    && cp ./target/$(xx-cargo --print-target-triple)/release/fuel-core /root/fuel-core \
```
xgreenx, line 59: Why do we want to copy if moving is faster?=)

xgreenx (approved):

deployment/Dockerfile
mchristopher: `mv` just changes the filesystem pointer, which is problematic when the `build` directory is cached. The cache is only mounted during this RUN command, so I need to copy the file to ensure it's available for Docker to copy to another image later in the run.

Voxelot (approved):

PR #2131, opened 2024-08-27
Send whole transaction pool upon subscription to gossip

AurelienFT:
## Linked Issues/PRs
Part of https://github.com/FuelLabs/fuel-core/issues/1049

## Description
This PR introduce a new workflow when we successfully subscribed to the TX topic in gossip we try ask for all the tx_ids in the pool's peer and then ask the full transactions that we don't have.
This PR also build the `SharedState` of `P2PService` independently to give more flexibility to the initialisation and avoid cyclic dependencies.

### Detailed workflow : 

- When a peer has subscribed to our topic send inform the TxPool
- The TxPool asks the tx ids of the peer
- The TxPool filter the one it already knows and ask for the full transactions of the others (a multi message send should be done if the number of txs asked is above the limit (currently 10k chosen randomly)).
- We verify and add them to our txpool

### Questions for review

- What's happens if the peer doesn't answer to a request that we await ?
- I placed Option in the request/response because there was in the other messages but why it's needed ?

## Checklist

### Before requesting review


AurelienFT:
@rymnc The PR is still in draft because I still need to make tests but you can start take a look the logic shouldn't change much.

rymnc:
> @rymnc The PR is still in draft because I still need to make tests but you can start take a look the logic shouldn't change much.
> 
> Thankssss <3

will look at it deeply tomorrow, hope that works!

rymnc:
looking good, however, I don't think we can call this "gossip" since its a direct p2p request/response protocol given that we make the request to the peer directly (`peer_id`) as part of the request.

> What's happens if the peer doesn't answer to a request that we await ?

we should have a timeout for waiting for the response, and then try another peer (cc: @xgreenx for confirmation)

> I placed Option in the request/response because there was in the other messages but why it's needed ?

for the other data we're sending over the req/res protocol (`sealed_block_headers` and `transactions`), there is a chance that the range provided is greater than the block height, and to handle that we send `None` for all blocks we don't have data for. We don't need Option usage for transaction pool transactions :)

crates/services/p2p/src/service.rs
```diff
@@ -439,6 +474,12 @@ where
             RequestMessage::SealedHeaders(range) => {
                 self.handle_sealed_headers_request(range, request_id)
             }
+            RequestMessage::AllTransactionsIds => {
+                self.handle_all_transactions_ids_request(request_id)
+            }
+            RequestMessage::FullTransactions(tx_ids) => {
```
rymnc, line 480: I think we should prefix the `RequestMessage`'s with `TxPool` to disambiguate between database transactions and txpool transactions :)

AurelienFT:
@rymnc Thanks for your overview and your comments 

About : 

> I don't think we can call this "gossip" since its a direct p2p request/response protocol

I don't think we ever call the request/response gossip, it's just that it's been trigger on the tx gossip topic being subscribed.

>  We don't need Option usage for transaction pool transactions :)

Will remove that

crates/services/p2p/src/service.rs
AurelienFT: Done

crates/fuel-core/src/service/adapters/txpool.rs
```diff
@@ -102,9 +118,48 @@ impl fuel_core_txpool::ports::PeerToPeer for P2PAdapter {
             Ok(())
         }
     }
+
+    async fn request_tx_ids(
+        &self,
+        peer_id: Vec<u8>,
+    ) -> anyhow::Result<Vec<fuel_core_txpool::types::TxId>> {
```
rymnc, line 125: maybe you want to import TxId at the top given you're using the fully qualified type in two places

crates/fuel-core/src/service/adapters/txpool.rs
```diff
@@ -91,6 +92,21 @@ impl fuel_core_txpool::ports::PeerToPeer for P2PAdapter {
...
+            wrappers::BroadcastStream,
+            StreamExt,
+        };
+        if let Some(service) = &self.service {
+            Box::pin(
+                BroadcastStream::new(service.subscribe_new_tx_subscription())
+                    .filter_map(|result| result.ok()),
+            )
+        } else {
+            fuel_core_services::stream::IntoBoxStream::into_boxed(tokio_stream::pending())
```
rymnc, line 106: can this be `fuel_core_services::stream::pending()` ?

crates/fuel-core/src/service/adapters/txpool.rs
```diff
@@ -126,6 +181,25 @@ impl fuel_core_txpool::ports::PeerToPeer for P2PAdapter {
...
+
+    fn new_tx_subscription(&self) -> BoxStream<Vec<u8>> {
+        Box::pin(fuel_core_services::stream::pending())
+    }
+
+    async fn request_tx_ids(
+        &self,
+        _peer_id: Vec<u8>,
+    ) -> anyhow::Result<Vec<fuel_core_txpool::types::TxId>> {
+        Ok(vec![])
```
rymnc, line 193: if we wrapped the vec in an option, we could just do `Ok(None)`. perhaps this is better for performance, but this will never be called if p2p isn't enabled, so it doesn't matter i guess

crates/services/p2p/src/p2p_service.rs
```diff
@@ -1594,6 +1627,46 @@ mod tests {
                                             }
                                         });
                                     }
+                                    RequestMessage::TxPoolAllTransactionsIds => {
```
rymnc, line 1630: imo, this is a lot of nesting within the same test, can we abstract some commonly used funcionality out?

crates/services/p2p/src/service.rs
```diff
@@ -903,19 +1054,58 @@ impl SharedState {
...
 }
 
-pub fn new_service<V, B>(
+pub fn build_shared_state(
+    config: Config<NotInitialized>,
+) -> (SharedState, Receiver<TaskRequest>) {
+    let (request_sender, request_receiver) = mpsc::channel(1024 * 10);
+    let (tx_broadcast, _) = broadcast::channel(1024 * 10);
+    let (new_tx_subscription_broadcast, _) = broadcast::channel(1024 * 10);
+    let (block_height_broadcast, _) = broadcast::channel(1024 * 10);
```
rymnc, line 1063: can we move the magic numbers out of this function and define it as -
a. a global, or
b. include it in the arguments to the function?

rymnc (commented):
In general LGTM, added some comments/questions :)

crates/fuel-core/src/service/adapters/txpool.rs
AurelienFT: This doesn't change much as it's never called and will bring a lot of overhead everywhere in the code to manage the option

rymnc: agreed!

crates/fuel-core/src/service/adapters/txpool.rs
AurelienFT: Done :)

crates/fuel-core/src/service/adapters/txpool.rs
AurelienFT: Done 👍

crates/services/p2p/src/service.rs
AurelienFT: The code was using magic number, idk why it's not in the config so I added a global.

crates/fuel-core/src/service/adapters/p2p.rs
```diff
@@ -49,3 +54,20 @@ impl BlockHeightImporter for BlockImporterAdapter {
         )
     }
 }
+
+impl TxPool for TxPoolAdapter {
+    fn get_all_tx_ids(&self) -> Vec<fuel_core_txpool::types::TxId> {
```
rymnc, line 59: nit: can we have a top level import for TxId please?
```suggestion
    fn get_all_tx_ids(&self) -> Vec<TxId> {
```

crates/fuel-core/src/service/adapters/p2p.rs
```diff
@@ -49,3 +54,20 @@ impl BlockHeightImporter for BlockImporterAdapter {
...
+        self.service.get_all_tx_ids()
+    }
+
+    fn get_full_txs(
+        &self,
+        tx_ids: Vec<fuel_core_txpool::types::TxId>,
```
rymnc, line 65:
```suggestion
        tx_ids: Vec<TxId>,
```

crates/fuel-core/src/service/adapters/p2p.rs
AurelienFT: Done

crates/fuel-core/src/service/adapters/p2p.rs
AurelienFT: Done

rymnc (approved):

crates/services/txpool/src/ports.rs
```diff
@@ -36,12 +37,17 @@ use std::{
...
+#[async_trait::async_trait]
 pub trait PeerToPeer: Send + Sync {
     type GossipedTransaction: NetworkData<Transaction>;
 
     // Gossip broadcast a transaction inserted via API.
     fn broadcast_transaction(&self, transaction: Arc<Transaction>) -> anyhow::Result<()>;
 
+    /// Creates a stream that is filled with the peer_id when they subscribe to
+    /// our transactions gossip.
+    fn new_tx_subscription(&self) -> BoxStream<Vec<u8>>;
```
Voxelot, line 49: should we use a PeerId newtype for Vec<u8>?

AurelienFT: I took example from :
https://github.com/FuelLabs/fuel-core/blob/95b373e80094b09ab5dc6c1e351deccf68ff7dd9/crates/services/p2p/src/service.rs#L842

I was thinking that this was the way you used `PeerId` outside of the P2P module. But I just saw that the PeerId type is on the type shared crate I will change this.

crates/fuel-core/src/service/adapters/txpool.rs
```diff
@@ -102,9 +119,45 @@ impl fuel_core_txpool::ports::PeerToPeer for P2PAdapter {
...
         }
     }
+
+    async fn request_tx_ids(&self, peer_id: Vec<u8>) -> anyhow::Result<Vec<TxId>> {
+        if let Some(service) = &self.service {
+            match service.get_all_transactions_ids_from_peer(peer_id).await {
+                Ok(txs) => Ok(txs),
+                Err(e) => {
+                    tracing::error!("Error getting tx ids from peer: {:?}", e);
+                    Ok(vec![])
```
Dentosal, line 129: Why we just log the error but do not propagate it upwards?

AurelienFT: I don't want the `txpool` service to stop because of the error if it's just that we can't communicate with this peer anymore. Maybe this function shouldn't return a result at all.

AurelienFT: Done

crates/services/txpool/src/ports.rs
AurelienFT: Done

tests/tests/tx_gossip.rs
```diff
@@ -199,7 +199,7 @@ async fn test_tx_gossiping_invalid_txs(
         .submit(valid_transaction.clone())
         .await
         .expect("Transaction is valid");
-    let block = tokio::time::timeout(Duration::from_secs(5), authority_blocks.next())
+    let block = tokio::time::timeout(Duration::from_secs(8), authority_blocks.next())
```
Dentosal, line 202: Looks like a [time dependency port](github.com/FuelLabs/fuel-core/pull/2145) would be a good thing to add here as well. As a follow-up, not in this PR.

crates/fuel-core/src/service/adapters/txpool.rs
rymnc: well, this function should return an error if sending the request to the peer fails, so I don't think we should remove the result from this.

AurelienFT: Done

rymnc (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
+- [2131](https://github.com/FuelLabs/fuel-core/pull/2131): Add flow in TxPool in order to ask to newly connected peers to share their transaction pool
```
rymnc, line 10: perhaps this should go after and not before the other entries :)

AurelienFT: Let's see when we will merge it it will propbably change again

crates/services/p2p/src/service.rs
```diff
@@ -339,7 +383,11 @@ pub struct Task<P, V, B> {
     request_sender: mpsc::Sender<TaskRequest>,
     database_processor: HeavyTaskProcessor,
     broadcast: B,
+    #[allow(dead_code)]
+    tx_pool: T,
     max_headers_per_request: usize,
+    #[allow(dead_code)]
```
xgreenx, line 389:
```suggestion
    tx_pool: T,
    max_headers_per_request: usize,
```

Why do we need dead code here?

crates/services/p2p/src/service.rs
```diff
@@ -354,43 +402,31 @@ pub struct HeartbeatPeerReputationConfig {
     low_heartbeat_frequency_penalty: AppScore,
 }
 
-impl<V> UninitializedTask<V, SharedState> {
+impl<V, T: TxPool> UninitializedTask<V, SharedState, T> {
```
xgreenx, line 405:
```suggestion
impl<V, T> UninitializedTask<V, SharedState, T> {
```

crates/services/p2p/src/service.rs
```diff
@@ -354,43 +402,31 @@ pub struct HeartbeatPeerReputationConfig {
...
-                block_height_broadcast,
-            },
+            broadcast: shared_state,
             config,
         }
     }
 }
 
-impl<P: TaskP2PService, V, B: Broadcast> Task<P, V, B> {
+impl<P: TaskP2PService, V, B: Broadcast, T: TxPool> Task<P, V, B, T> {
```
xgreenx, line 429:
```suggestion
impl<P: TaskP2PService, V, B: Broadcast, T> Task<P, V, B, T> {
```

crates/services/p2p/src/service.rs
```diff
@@ -756,6 +849,9 @@ where
                     Some(FuelP2PEvent::InboundRequestMessage { request_message, request_id }) => {
                         self.process_request(request_message, request_id)?
                     },
+                    Some(FuelP2PEvent::NewSubscription { peer_id, .. }) => {
+                        let _ = self.broadcast.new_tx_subscription_broadcast(FuelPeerId::from(peer_id.to_bytes()));
+                    },
```
xgreenx, line 854: This code currently works for all possible subscribers. I think we just want to know about subscribers for transactions. I think we either check the tag of subscription, or pass the tag along with peer id.

crates/services/p2p/src/service.rs
```diff
@@ -857,7 +955,55 @@ impl SharedState {
...
+        };
+        self.request_sender.send(request).await?;
+
+        let (response_from_peer, response) =
+            receiver.await.map_err(|e| anyhow!("{e}"))?;
+        assert_eq!(
+            peer_id.as_ref(),
+            response_from_peer.to_bytes(),
+            "Bug: response from non-requested peer"
+        );
```
xgreenx, line 984:
```suggestion
        debug_assert_eq!(
            peer_id.as_ref(),
            response_from_peer.to_bytes(),
            "Bug: response from non-requested peer"
        );
```

crates/services/p2p/src/service.rs
```diff
@@ -857,7 +955,55 @@ impl SharedState {
...
+        let request = TaskRequest::TxPoolGetFullTransactions {
+            tx_ids,
+            from_peer,
+            channel: sender,
+        };
+        self.request_sender.send(request).await?;
+
+        let (response_from_peer, response) =
+            receiver.await.map_err(|e| anyhow!("{e}"))?;
+        assert_eq!(
```
xgreenx, line 1005:
```suggestion
        debug_assert_eq!(
```

crates/services/txpool/src/service.rs
```diff
@@ -271,6 +273,53 @@ where
                 }
             }
 
+            // TODO: Refactor this to not block on calls to the peer and verification (need to open issue before merging)
+            new_peer_subscribed = self.new_tx_gossip_subscription.next() => {
```
xgreenx, line 277: I think we need to do this change in this PR since this branch of Tokyo select can be very heavy and long. I think we need spawn a task for it.

We can create queue of task by using rayon for example. The same like we did with `crates/services/p2p/src/heavy_task_processor.rs` just to process this use case.

If you want, it can be another PR on top of this one, we will review and merge into this branch

crates/services/p2p/src/service.rs
AurelienFT: Done

crates/services/p2p/src/service.rs
AurelienFT: Done

crates/services/p2p/src/service.rs
AurelienFT: Done

crates/services/p2p/src/service.rs
AurelienFT: Outdated spec. Removed.

CHANGELOG.md
AurelienFT: Done

crates/fuel-core/src/service/adapters/p2p.rs
```diff
@@ -49,3 +55,17 @@ impl BlockHeightImporter for BlockImporterAdapter {
...
+impl TxPool for TxPoolAdapter {
+    fn get_all_tx_ids(&self) -> Vec<TxId> {
+        self.service.get_all_tx_ids()
+    }
+
+    fn get_full_txs(&self, tx_ids: Vec<TxId>) -> Vec<Option<Transaction>> {
+        self.service
+            .find(tx_ids)
+            .into_iter()
+            .map(|tx_info| tx_info.map(|tx| tx.tx().as_ref().into()))
```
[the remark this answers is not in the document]
xgreenx: I see. Yeah, we can solve it by using `Arc<Transaction>` on ports and adapter levels. Instead of using `Arc<PoolTransaction>` in the TxPool, we could use `struct PoolTransaction(Arc<Transaction>)`. Then, in this case, we could easily get `Arc<Transaction>`.

It is okay to do this change in a separate PR. It would be nice to make this change before merging this PR, if possible. But it is up to you=)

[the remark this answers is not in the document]
AurelienFT: I can change this in the transaction pool v2. Can we wait tx pool v2 before merging this ?

crates/services/txpool/src/service.rs
AurelienFT: Why do we we want to use a rayon dedicated thread pool instead of spawning a call tokio task ?

xgreenx: We can spawn tokyo task as well, we need to limit the number of active tasks to be one. Either we can write own queue of tasks and manage them, or reuse `rayon::spawn_fifo`.

crates/fuel-core/src/service/adapters/p2p.rs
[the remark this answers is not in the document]
xgreenx: I think it is better to do in a separate small PR. This branch and TxPoolV2 could use this branch as a base for this change

crates/services/txpool/src/service.rs
AurelienFT: Implemented in : https://github.com/FuelLabs/fuel-core/pull/2131/commits/268d794bb0a70f0ccde2e0ad807f5a997c867538

crates/fuel-core/src/service/adapters/p2p.rs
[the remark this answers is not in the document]
AurelienFT: After discussion I decided to go for a special type in P2P that allow to serialize a pool transaction or simple transaction as transaction and deserialize only in transactiton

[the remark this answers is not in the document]
AurelienFT: Here is a diff showing the changes that implement this : https://github.com/FuelLabs/fuel-core/compare/268d794...5e16d87

crates/services/txpool/src/service.rs
```diff
@@ -271,6 +275,54 @@ where
...
+                            params.as_ref(),
+                            &shared_state.gas_price_provider,
+                            shared_state.memory_pool.clone()
+                        ).await.into_iter().flatten().collect::<Vec<_>>();
+
+                        shared_state.txpool.lock().insert(
+                            &tx_status_change,
+                            version,
+                            checked_transactions
+                        );
```
xgreenx, line 320: I think you can just call `SharedState::insert`

crates/services/p2p/src/service.rs
```diff
@@ -857,14 +956,74 @@ impl SharedState {
...
+        );
+
+        let txs = response.map_err(|e| anyhow!("Invalid response from peer {e:?}"))?;
+        if txs.len() > self.max_txs_per_request {
+            return Err(anyhow!("Too many transactions requested: {}", txs.len()));
+        }
+        Ok(txs.into_iter().map(|tx| tx.map(|tx_enum| match tx_enum {
+            NetworkableTransactionPool::Transaction(tx) => tx,
+            NetworkableTransactionPool::PoolTransaction(_) => panic!("A transaction that comes from the network must always be deserialized as a Transaction"),
+        })).collect())
```
xgreenx, line 1024: It would be nice to separate iterator part with the `Ok` part.

Also, it would be better to avoid panic, even if it is not possible in the current code. Or implement `From<NetworkableTransactionPool>` to `Transaction`. Another question, is that possible to have one type for serialization and another type for deseriliazation? Maybe we can avoid this check at all

crates/services/p2p/src/service.rs
```diff
@@ -857,14 +956,74 @@ impl SharedState {
...
+        tx_ids: Vec<TxId>,
+    ) -> anyhow::Result<Vec<Option<Transaction>>> {
+        let (sender, receiver) = oneshot::channel();
+        let from_peer = PeerId::from_bytes(peer_id.as_ref()).expect("Valid PeerId");
+        let request = TaskRequest::TxPoolGetFullTransactions {
+            tx_ids,
+            from_peer,
+            channel: sender,
+        };
+        self.request_sender.send(request).await?;
```
xgreenx, line 1007: It would be nice to use `try_send`

crates/services/p2p/src/service.rs
```diff
@@ -857,14 +956,74 @@ impl SharedState {
...
+        &self,
+        peer_id: FuelPeerId,
+    ) -> anyhow::Result<Vec<TxId>> {
+        let (sender, receiver) = oneshot::channel();
+        let from_peer = PeerId::from_bytes(peer_id.as_ref()).expect("Valid PeerId");
+        let request = TaskRequest::TxPoolGetAllTxIds {
+            from_peer,
+            channel: sender,
+        };
+        self.request_sender.send(request).await?;
```
xgreenx, line 977: It would be nice to use `try_send`

crates/services/p2p/src/service.rs
```diff
@@ -555,8 +595,38 @@ where
...
+        request_id: InboundRequestId,
+    ) -> anyhow::Result<()> {
+        let response_channel = self.request_sender.clone();
+
+        let txs = self.tx_pool.get_full_txs(tx_ids);
+        response_channel.try_send(TaskRequest::TxPoolFullTransactions {
+            response: txs,
+            request_id,
+        })?;
+        Ok(())
```
xgreenx, line 628: Potentially TxPool can be very busy with current work and answering can take a lot of time. It is better to use `handle_request` as well

xgreenx (commented):
I have small pack of comments, but the change is almost done!=)

crates/services/p2p/src/service.rs
AurelienFT: Done

crates/fuel-core/src/service/adapters/p2p.rs
```diff
@@ -49,3 +57,21 @@ impl BlockHeightImporter for BlockImporterAdapter {
...
+impl TxPool for TxPoolAdapter {
+    fn get_all_tx_ids(&self) -> Vec<TxId> {
+        self.service.get_all_tx_ids()
+    }
+
+    fn get_full_txs(&self, tx_ids: Vec<TxId>) -> Vec<Option<NetworkableTransactionPool>> {
+        self.service
+            .find(tx_ids)
+            .into_iter()
+            .map(|tx_info| {
```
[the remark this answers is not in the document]
AurelienFT: I just want to re-use an existitng method and not make another one that do the same but `.map()` inside the txpool method

crates/services/p2p/src/service.rs
AurelienFT: I don't really know what you mean by your first sentence. Also in your second setence you want me to use the thread pool to make the ask to the TxPool ? (I can't directly reuse `handle_request` as it's specifically done for range asks (I can rework the function)

crates/services/p2p/src/service.rs
AurelienFT: Done.

crates/services/p2p/src/service.rs
AurelienFT: Done.

crates/services/p2p/src/service.rs
AurelienFT: I have changed to propagate the result.

I didn't find any way to serialize in a type and deserialize to another with the libp2p enum and serde that's why I ended up with this enum.

crates/services/txpool/src/service.rs
AurelienFT: Yeah good catch I was duplicating the verif. All the boilerplate and generics makes it very hard to read.

crates/services/txpool/src/service.rs
```diff
@@ -271,6 +275,54 @@ where
                 }
             }
 
+            new_peer_subscribed = self.new_tx_gossip_subscription.next() => {
+                // If we are out of capacity, we will skip this event.
+                let _ = self.heavy_async_processor.spawn({
+                    let shared_state = self.tx_pool_shared_state.clone();
+                    let tx_status_change = shared_state.tx_status_sender.clone();
+                    async move {
+                    if let Some(peer_id) = new_peer_subscribed {
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/services/txpool/src/service.rs
```diff
@@ -271,6 +275,54 @@ where
                 }
             }
 
+            new_peer_subscribed = self.new_tx_gossip_subscription.next() => {
+                // If we are out of capacity, we will skip this event.
+                let _ = self.heavy_async_processor.spawn({
+                    let shared_state = self.tx_pool_shared_state.clone();
+                    let tx_status_change = shared_state.tx_status_sender.clone();
+                    async move {
```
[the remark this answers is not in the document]
AurelienFT: Done

crates/services/p2p/src/config.rs
```diff
@@ -49,6 +49,9 @@ pub const MAX_RESPONSE_SIZE: usize = 18 * 1024 * 1024;
 /// Maximum number of blocks per request.
 pub const MAX_HEADERS_PER_REQUEST: usize = 100;
 
+/// Maximum number of transactions ids asked per request.
+pub const MAX_TXS_PER_REQUEST: usize = 1000000;
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/fuel-core/src/service/adapters/txpool.rs
```diff
@@ -91,6 +94,21 @@ impl fuel_core_txpool::ports::PeerToPeer for P2PAdapter {
         }
     }
 
+    fn new_tx_subscription(&self) -> BoxStream<PeerId> {
```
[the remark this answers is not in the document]
AurelienFT: Done.

bin/fuel-core/src/cli/run/p2p.rs
```diff
@@ -66,6 +66,10 @@ pub struct P2PArgs {
     #[clap(long = "max-headers-per-request", default_value = "100", env)]
     pub max_headers_per_request: usize,
 
+    /// Max number of txs in a single txs request response
+    #[clap(long = "max-txs-per-request", default_value = "1000000", env)]
```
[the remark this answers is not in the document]
AurelienFT: Changed. We need to becareful that this value is under the maximum size of the pool as right now there is no "send in parts" protocol.

crates/services/p2p/src/service.rs
AurelienFT: Done in https://github.com/FuelLabs/fuel-core/pull/2131

crates/services/p2p/src/service.rs
```diff
@@ -858,14 +1000,85 @@ impl SharedState {
...
+            response_from_peer.to_bytes(),
+            "Bug: response from non-requested peer"
+        );
+
+        let Some(txs) =
+            response.map_err(|e| anyhow!("Invalid response from peer {e:?}"))?
+        else {
+            return Ok(vec![]);
+        };
+        if txs.len() > self.max_txs_per_request {
```
xgreenx, line 1070: Maybe we need to not send more than `self.max_txs_per_request` as well=)

xgreenx (approved):

crates/services/p2p/src/ports.rs
```diff
@@ -28,3 +32,9 @@ pub trait BlockHeightImporter: Send + Sync {
     /// Creates a stream of next block heights
     fn next_block_height(&self) -> BoxStream<BlockHeight>;
 }
+
+pub trait TxPool: Send + Sync + Clone {
+    fn get_all_tx_ids(&self) -> Vec<TxId>;
+
+    fn get_full_txs(&self, tx_ids: Vec<TxId>) -> Vec<Option<NetworkableTransactionPool>>;
+}
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/services/p2p/src/service.rs
AurelienFT: Done

xgreenx: I meant it would be nice from TxPool not to request more than our own config allows=)

xgreenx (dismissed):

crates/services/p2p/src/service.rs
AurelienFT: Oopsie, Changed.

xgreenx (approved):

rymnc (approved):

Voxelot:
Do we reconcile the txs before or after subscribing? If we start processing gossipped txs while we are still syncing the txpool, we may end up dropping them if they depend on earlier txs that havent been synced yet.

AurelienFT:
> Do we reconcile the txs before or after subscribing? If we start processing gossipped txs while we are still syncing the txpool, we may end up dropping them if they depend on earlier txs that havent been synced yet.

The trigger to make this reconciliation is the subscription so yes it happens just after suscribing. You are right it could lead to invalid gossiped transactions from the point of view of the receiver.
Do you have an idea of when we should make the trigger ?
Because if we activate the gossip after the reconciliation there the possibility that we have miss a child during a period where the gossip wasn't activate too.

Possible solution that I see is, given the tx pool v2 is that all gossiped transactions are queued until we finished at least one peer reconcialition and then we start to process them.
It's pretty easy to do given the fact that in TxPool v2 the verif and insert is done in parrallel manged by a threadpool and a queue

PR #2134, opened 2024-08-27
Secp256k1 RecoveryID correction for KMS-genrated signatures

Dentosal:
AWS KMS doesn't insert recovery ids into otherwise-recoverable secp256k1 signatures. This PR patches the process [in the same way we already do in fuel-vm](https://github.com/FuelLabs/fuel-vm/blob/6e137db6387bd291b9505d17e15e0f2edda7957e/fuel-crypto/src/secp256/backend/k1/k256.rs#L55).

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog (I guess. Fixing the incorrect behavior is a fix, not a breaking change, although it changes how the code behaves.)

### Before requesting review


CHANGELOG.md
```diff
@@ -16,9 +18,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2115](https://github.com/FuelLabs/fuel-core/pull/2115): Add test for `SignMode` `is_available` method.
 - [2124](https://github.com/FuelLabs/fuel-core/pull/2124): Generalize the way p2p req/res protocol handles requests.
 
-### Added
-- [2119](https://github.com/FuelLabs/fuel-core/pull/2119): GraphQL query fields for retrieving information about upgrades.
-- [2116](https://github.com/FuelLabs/fuel-core/pull/2116): Replace `H160` in config and cli options of relayer by `Bytes20` of `fuel-types`
+### Fixed
+- [2134](https://github.com/FuelLabs/fuel-core/pull/2134): Perform RecoveryID normalization for AWS KMS -generated signatures.
```
rymnc, line 22:
```suggestion
- [2134](https://github.com/FuelLabs/fuel-core/pull/2134): Perform RecoveryID normalization for AWS KMS generated signatures.
```

rymnc (approved):
LGTM

CHANGELOG.md
Dentosal: Oh good point I ddn't realize the space-before-dash compound-noun-modifiers are nonstandard English. After a long consultation with ChatGPT and couple of style guides, I'm still convinced that the form I used clearner than the alternative without the dash. If you feel strongly about this, the sentence could be reformulated as 

> Perform RecoveryID normalization for signatures generated by AWS KMS.

crates/services/consensus_module/poa/src/signer.rs
```diff
@@ -90,8 +102,40 @@ async fn sign_with_kms(
...
-    Ok(Signature::from_bytes(sig_bytes))
+
+    // This is a hack to get the recovery id. The signature should be normalized
+    // before computing the recovery id, but aws kms doesn't support this, and
+    // instead always computes the recovery id from non-normalized signature.
+    // So instead the recovery id is determined by checking which variant matches
+    // the original public key.
+
+    let recid1 = RecoveryId::new(false, false);
+    let recid2 = RecoveryId::new(true, false);
```
netrome, line 113: Why are we only testing two out of four possible recovery ID:s? Shouldn't this be
```suggestion
    let recid1 = RecoveryId::new(false, false);
    let recid2 = RecoveryId::new(true, false);
    let recid3 = RecoveryId::new(true, true);
    let recid4 = RecoveryId::new(false, true);
```

netrome: I see the assertion "reduced-x form coordinates are never generated by this", but I don't understand why we can assume that this is the case. Is that a guarantee from KMS?

netrome (approved):
I don't fully understand why we can guarantee that the reduced x form is never generated, but given the explicit assertion this seems intentional - hence I'll approve.

crates/services/consensus_module/poa/src/signer.rs
Dentosal: Most implementations, e.g. ethereum, disallow reduced-x form signatures. See [EIP-155](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md). While it's true that reduced-x signatures can be genenerated, it's exceedingly rare, approximately once per 2e36 signatures. Still, it's likely not smart to panic here and we should indeed return an error so that the caller can retry. However, no retry logic is currently implemented, so the distinction doesn't matter right now. Still feels like it makes sense to fix this, at least the panic message, so I'll do so.

Dentosal: Also, the current fuel-crypto interface pretends to be infallible, and will indeed panic on such signatures.

netrome: Thanks for the explanation. Yeah I'm not a fan of hidden panics even if they are unlikely.

Dentosal: Unlikely is an understatement here: This is less likely than winning lottery five times in a row. I wouldn't usually prepare for uuid collisions either, and this is way less likely.

Fixed in 8ed3df7ea.

netrome (approved):

crates/services/consensus_module/poa/src/signer.rs
netrome: Fair enough.

rymnc (approved):

PR #2135, opened 2024-08-28
chore(p2p_service): add metrics for number of blocks requested over p2p req/res protocol

rymnc:
## Linked Issues/PRs

- #2023
- #2112

## Description

We bubble up the usage of a new function `log_metrics` to clean up how metrics are being logged in the p2p module. additionally, we also specify a `Gauge` for how many blocks worth of data has been requested, so we can re-use those numbers in our simulations to find the most optimal lookup method.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/p2p/src/p2p_service.rs
```diff
@@ -271,6 +271,15 @@ impl FuelP2PService {
         }
     }
 
+    pub fn log_metrics<T>(&self, cb: T)
```
netrome, line 274: Can we use a more descriptive name for `cb` here? Is it a callback? What do we assume the callback does?

crates/metrics/src/p2p_metrics.rs
```diff
@@ -28,3 +42,11 @@ static P2P_METRICS: OnceLock<P2PMetrics> = OnceLock::new();
 pub fn p2p_metrics() -> &'static P2PMetrics {
     P2P_METRICS.get_or_init(P2PMetrics::new)
 }
+
+pub fn inc_unique_peers() {
```
netrome, line 46: Nit: I'd find this easier to read with a more descriptive name could we call this `increment_unique_peers`?

crates/services/p2p/src/p2p_service.rs
netrome: Also, I guess we're not really logging anything here. Given how this is used, I'd suggest renaming the function to something like `update_metrics`, where the callback could be named `update_function`. I'm open to other names as well of course.

crates/services/p2p/src/service.rs
```diff
@@ -427,6 +439,13 @@ where
     V: AtomicView + 'static,
     V::LatestView: P2pDb,
 {
+    fn log_metrics<T>(&self, cb: T)
+    where
+        T: FnOnce(),
+    {
+        self.p2p_service.log_metrics(cb)
+    }
```
netrome, line 447: Same here

netrome (approved):
Approved with some naming suggestions.

crates/services/p2p/src/p2p_service.rs
rymnc: good point about the naming, will address.
the callback should be something that updates metrics

rymnc: addressed in 22f7693a5

crates/services/p2p/src/service.rs
rymnc: addressed in 22f7693a5

crates/services/p2p/src/service.rs
```diff
@@ -196,9 +197,20 @@ pub trait TaskP2PService: Send {
     ) -> anyhow::Result<()>;
 
     fn update_block_height(&mut self, height: BlockHeight) -> anyhow::Result<()>;
+
+    fn update_metrics<T>(&self, update_fn: T)
```
MitchTurner, line 201: This seems off to me. I don't feel like the `update_metrics` method should be on this trait. (And the method effectivly has nothing to do with metrics, it's more like `run_closure`)

It looks like you put it here because the `metrics: bool` is on `FuelP2PService`, but that is only used in one spot currently.

Metrics are handled globally. Maybe we consider handling capture of metrics globally as well? What do you think?

rymnc: I discussed this with Green today, and it looks like only 3/n modules use the metrics flag, and p2p being one of them. the others export metrics regardless of the flag being enabled or disabled (which imo is incorrect behaviour). 

I have a task on my to-do list to create an issue to make sure the metrics flag is propagated to all modules and exporting them only if it's enabled ~ perhaps this warrants a change or a trait from the metrics crate, but I'm holding off on that on account of premature optimization. 

happy to make it global if that makes it cleaner (which will need removal of the metrics field from P2pConfig)

MitchTurner: I'm fine with leaving this as is if we create a follow up issue that specifically includes removing the trait method and the metrics Boolean.  It sounds like we need to unify metrics architecture for all our services, in which case it could happen as part of the same issue/feature.

rymnc: https://github.com/FuelLabs/fuel-core/issues/2144 :)

rymnc: > feature flag at compile time

maybe operators would want to turn metrics collection on/off using the cli argument, so I am in favour of keeping the flag with a sensible default :)

netrome: Makes sense :+1:

netrome (approved):

MitchTurner (approved):

PR #2142, opened 2024-08-28
chore: add benchmarks for varied forms of lookups

rymnc:
This PR introduces benchmarks for varied methods of database lookups for blocks and included transactions, namely - 
1. `header_and_tx_lookup`
2. `full_block_lookup`
3. `multi_get_lookup`

We can use these benchmarks to ascertain which form is best for this use-case, and then implement it.

It is suggested to alter the bench matrix before running and using a beefy machine to do so.

Run it with `cargo bench --bench db_lookup_times`

## Linked Issues/PRs

- #2023 

## Description

- moved the implementation of `KeyValueMutate` for rocksdb out of the test module.
- each benchmark is located in `benches/benches/db_lookup_times.rs`
- stores `FullFuelBlocks` along with the `CompressedBlocks` and `Transactions` table to not affect single header/tx lookup times in other usages.
- implemented various traits for the `FullFuelBlocks` table
- the [notion doc](https://www.notion.so/fuellabs/DB-Analysis-for-Lookups-f62aa0cd5cdf4c91b9575b8fb45683f7) has more information about the analysis we did.
- each benchmark randomizes the block height that is queries *per iteration* to ensure randomness and fair comparison
- rocksdb caching has been `disabled` for these benchmarks

## Checklist

### Before requesting review

- [ ] Someone else?


benches/benches/db_lookup_times.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    seed_full_block_matrix();
+    let mut group = c.benchmark_group(method);
+
+    for (block_count, tx_count) in matrix() {
+        let database = open_db(block_count, tx_count, method);
+        let height = get_random_block_height(block_count);
+        group.bench_function(format!("{block_count}/{tx_count}"), |b| {
+            b.iter(|| {
+                let view = database.latest_view().unwrap();
```
xgreenx, line 74: I think it is better to create view before the benchmark=)

benches/benches/db_lookup_times.rs
```diff
@@ -0,0 +1,88 @@
...
+
+pub fn header_and_tx_lookup(c: &mut Criterion) {
+    let method = "header_and_tx";
+
+    seed_compressed_blocks_and_transactions_matrix(method);
+    let mut group = c.benchmark_group(method);
+
+    for (block_count, tx_count) in matrix() {
+        let database = open_db(block_count, tx_count, method);
+        let height = get_random_block_height(block_count);
```
xgreenx, line 32: What if we choose a random height on each iteration? We could create rng before the iterations and just pass it between iterations

benches/benches/db_lookup_times.rs
rymnc: its draft, im still cleaning up !

rymnc: addressed!

benches/benches/db_lookup_times_utils/matrix.rs
```diff
@@ -0,0 +1,13 @@
+pub const BLOCK_COUNT_MATRIX: [u32; 2] = [10, 100];
+pub const TX_COUNT_MATRIX: [u32; 2] = [100, 1000];
+
+// todo: we can make this lazy loaded
+pub fn matrix() -> Box<dyn Iterator<Item = (u32, u32)>> {
```
netrome, line 5: What is the reason for boxing the iterator here instead of just returning `impl Iterator`?

benches/benches/db_lookup_times_utils/matrix.rs
```diff
@@ -0,0 +1,13 @@
+pub const BLOCK_COUNT_MATRIX: [u32; 2] = [10, 100];
+pub const TX_COUNT_MATRIX: [u32; 2] = [100, 1000];
+
+// todo: we can make this lazy loaded
```
netrome, line 4: Is this a stale comment? I assume since the matrix() function returns an iterator that it is lazy loaded already.

benches/benches/db_lookup_times_utils/utils.rs
```diff
@@ -0,0 +1,101 @@
...
+    Rng,
+};
+use std::{
+    borrow::Cow,
+    path::Path,
+};
+
+pub fn thread_rng() -> ThreadRng {
+    rand::thread_rng()
+}
```
netrome, line 43: What's the purpose of wrapping the `rand::thread_rng()` function here? This seems unnecessary to me.

netrome: Also more generally, do we want to use the thread rng here? I tend to prefer to use seeded rngs for reproducibility in tests, but I'm not sure what's best practice for benchmarks.

benches/benches/db_lookup_times.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    seed_compressed_blocks_and_transactions_matrix(method);
+    let mut group = c.benchmark_group(method);
+
+    for (block_count, tx_count) in matrix() {
+        let database = open_raw_rocksdb(block_count, tx_count, method);
+        group.bench_function(format!("{block_count}/{tx_count}"), |b| {
+            b.iter(|| {
+                let height = get_random_block_height(&mut rng, block_count);
+                // todo: clean up
```
netrome, line 60: It's not obvious to me which code this comment refers to. Is this a stale todo that was addressed by factoring out the `multi_get_block` function or are there any other cleanups you intend?

netrome (changes requested):
The code is not the prettiest as you mentioned in our call yesterday, but it seems to be doing what it's supposed to do.

I'd like to see the remaining todo comments removed or clarified with follow-up issues before approving.

benches/benches/db_lookup_times.rs
rymnc: i want to assert here that the multi_get_block lookup succeeded :)

benches/benches/db_lookup_times_utils/matrix.rs
rymnc: we call matrix() multiple times, and it was my understanding that it is *not* lazy loaded, like using `OnceCell` perhaps. It should be okay for now given our matrix is small, and will not grow

benches/benches/db_lookup_times_utils/matrix.rs
rymnc: addressed in c619b937b

benches/benches/db_lookup_times_utils/matrix.rs
netrome: Oh then I see what you mean. Yeah that's probably not the bottleneck. Even if the matrix grows, the iteration should still not be significant in comparison to the benchmarks being run at each iteration. So from that perspective I'd prefer if we just skipped this comment altogether then.

rymnc: agreed 🤝

benches/benches/db_lookup_times_utils/utils.rs
netrome: Right and you want to run the benchmarks multiple times with different randomness. Fair enough.

> I don't want to import rand::thread_rng into the db_lookup_times.rs benchmark file and hence wrapped over it.

If you want to avoid importing it, you can keep the namespace when calling it in the code like you do in this function. I often prefer that. Especially for a small crate like `rand`.

I don't see the benefit of wrapping the function compared to simply importing it. What is the problem with importing it?

rymnc: eh, just how I prefer to write code. will address.

rymnc: addressed in 61d866828

netrome: > eh, just how I prefer to write code. will address.

Okay, interesting pattern. I'm not used to wrapping functions as an alternative to importing them. Personally I tend to keep the namespaces to a great extent, but this code base seems to prefer a lot of explicit imports. Thank you for accommodating this.

benches/benches/db_lookup_times_utils/matrix.rs
netrome: Nice, let me know when you've updated this and I'll approve :rocket:

netrome (changes requested):
Just one more todo comment to handle.

benches/benches/db_lookup_times_utils/matrix.rs
rymnc: addressed in 0af85338c

netrome (dismissed):

netrome (approved):

crates/storage/src/column.rs
```diff
@@ -71,6 +71,10 @@ pub enum Column {
     // TODO: Remove this column and use `Metadata` column instead.
     /// Table for genesis state import progress tracking.
     GenesisMetadata = 21,
+
+    /// See [`FullFuelBlocks`](crate::tables::FullFuelBlocks)
+    /// we don't use this table at the moment, only used for benchmarks
+    FullFuelBlocks = 22,
```
MitchTurner, line 77: [nit]: If we're gonna keep it, maybe put it above `GenesisMetadata`.

crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -863,6 +867,33 @@ fn next_prefix(mut prefix: Vec<u8>) -> Option<Vec<u8>> {
     None
 }
 
+impl<Description> KeyValueMutate for RocksDb<Description>
```
MitchTurner, line 870: I wonder if this should be behind a feature flag, since it wasn't exposed publicly before.

benches/benches/db_lookup_times_utils/seed.rs
```diff
@@ -0,0 +1,153 @@
...
+    block
+}
+
+fn insert_full_block(database: &mut RocksDb<OnChain>, height: u32, tx_count: u32) {
+    let block = insert_compressed_block(database, height, tx_count);
+
+    let height_key = height_key(height);
+    let raw_full_block = postcard::to_allocvec(&block).unwrap().to_vec();
+
+    // 3. insert into FullFuelBlocks table
```
MitchTurner, line 124: This `1.`, `2.`, and `3.` in separate functions is confusing. Are these comments left behind from a previous iteration?

benches/benches/db_lookup_times_utils/seed.rs
```diff
@@ -0,0 +1,153 @@
...
+    let mut blocks = vec![];
+    for block_number in 0..block_count {
+        blocks.push(insert_compressed_block(database, block_number, tx_count));
+    }
+    blocks
+}
+
+fn seed_full_blocks(database: &mut RocksDb<OnChain>, block_count: u32, tx_count: u32) {
+    // we seed compressed blocks and transactions to not affect individual
+    // lookup times
```
MitchTurner, line 148: This comment doesn't make sense to me inside the context of `seed_full_blocks`. I'm also not sure what it means to "affect individual lookup times".

rymnc: the context is that if we do include `FullFuelBlocks` in the codebase at some point and use it, we want to use it in tandem with the `FuelBlocks` and `Transactions` tables so as to not affect their individual lookup times, i.e cases where we make individual block header calls (to get the block height, for example), and where we make individual tx lookups.

crates/storage/src/column.rs
rymnc: `GenesisMetadata` is 21, so logically this should come after, right? or is it the general preference to keep everything before the `GenesisMetadata` table?

crates/fuel-core/src/state/rocks_db.rs
rymnc: it should be behind the `rocksdb` feature flag, and it doesn't affect any other api's :)

benches/benches/db_lookup_times_utils/seed.rs
rymnc: addressed in 432c7605f

crates/storage/src/column.rs
MitchTurner: Yeah. If you look at previous additions we've always kept `GenesisMetadata` on the bottom.

benches/benches/db_lookup_times_utils/seed.rs
MitchTurner: I understand now. I think this comment should be above the `insert_full_block` in that case.

rymnc: addressed in 7cca27818

benches/benches/db_lookup_times_utils/seed.rs
```diff
@@ -0,0 +1,151 @@
...
+            .unwrap();
+    }
+
+    block
+}
+
+// we seed compressed blocks and transactions to not affect individual
+// lookup times
+fn insert_full_block(database: &mut RocksDb<OnChain>, height: u32, tx_count: u32) {
+    let block = insert_compressed_block(database, height, tx_count);
```
MitchTurner, line 121: Sorry. I know I said here, but I think this is best.
```suggestion
fn insert_full_block(database: &mut RocksDb<OnChain>, height: u32, tx_count: u32) {
    // we seed compressed blocks and transactions to not affect individual
    // lookup times
    let block = insert_compressed_block(database, height, tx_count);
```

rymnc: addressed in 20184a415

netrome (approved):

crates/fuel-core/src/state/rocks_db.rs
xgreenx: It should be behind the "test-helpers" feature. We shouldn't allow modification of the Rocksdb outside of the `commit` method

crates/storage/src/column.rs
```diff
@@ -67,6 +67,9 @@ pub enum Column {
     UploadedBytecodes = 19,
     /// See [`Blobs`](fuel_vm_private::storage::BlobData)
     Blobs = 20,
+    /// See [`FullFuelBlocks`](crate::tables::FullFuelBlocks)
+    /// we don't use this table at the moment, only used for benchmarks
+    FullFuelBlocks = 22,
```
xgreenx, line 72: Benchmarks shouldn't affect the columns that are used in the real logic. RocksDB allows opening it with a custom set of columns.

crates/storage/src/structured_storage/blocks.rs
```diff
@@ -47,6 +60,21 @@ impl TableWithBlueprint for FuelBlocks {
     }
 }
 
+impl TableWithBlueprint for FullFuelBlocks {
+    type Blueprint = Merklized<
```
xgreenx, line 64: We don't need to create a Merkle tree over the blocks used only by P2P

benches/benches/db_lookup_times_utils/seed.rs
```diff
@@ -0,0 +1,151 @@
...
+    };
+    let txes = generate_bench_transactions(tx_count);
+    let block = PartialFuelBlock::new(header, txes);
+    block.generate(&[], Default::default()).unwrap()
+}
+
+fn generate_bench_transactions(tx_count: u32) -> Vec<Transaction> {
+    let mut txes = vec![];
+    for _ in 0..tx_count {
+        txes.push(Transaction::default_test_tx());
```
xgreenx, line 69: [nit] It would be nice to make transactions random to be sure that compression will not affect it too much

crates/storage/src/column.rs
rymnc: continuing work in #2159

crates/storage/src/structured_storage/blocks.rs
rymnc: addressed in #2159

benches/benches/db_lookup_times.rs
```diff
@@ -0,0 +1,96 @@
...
+
+pub fn header_and_tx_lookup(c: &mut Criterion) {
+    let method = "header_and_tx";
+    let mut rng = thread_rng();
+
+    seed_compressed_blocks_and_transactions_matrix(method);
+    let mut group = c.benchmark_group(method);
+
+    for (block_count, tx_count) in matrix() {
+        let database = open_db(block_count, tx_count, method);
```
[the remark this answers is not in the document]
rymnc: addressed in #2159

benches/benches/db_lookup_times_utils/seed.rs
rymnc: under the hood the transactions are randomized ~ not sure if I understand correctly!
image

crates/fuel-core/src/state/rocks_db.rs
rymnc: addressed in #2159

crates/storage/src/column.rs
rymnc: precisely in https://github.com/FuelLabs/fuel-core/pull/2159/files#diff-0e430672a8d94fbbca929879774abd2f09dd2a86eca77c32c77615e100b51831R14-R34

crates/storage/src/structured_storage/blocks.rs
rymnc: precisely in https://github.com/FuelLabs/fuel-core/pull/2159/files#diff-0e430672a8d94fbbca929879774abd2f09dd2a86eca77c32c77615e100b51831R57-R64

benches/benches/db_lookup_times.rs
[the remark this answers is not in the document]
rymnc: precisely in https://github.com/FuelLabs/fuel-core/pull/2159/files#diff-0e0c7132cb01e4397b764c20e78657d53e9daef5be6ecd41680f45ae6aac77e6R4-R16

benches/benches/db_lookup_times_utils/seed.rs
xgreenx: `add_random_fee_input` always return the same result=D

rymnc: lol, okay ~ will alter

PR #2145, opened 2024-08-30
feat: Introduce time port in PoA service

### LINKED ISSUE #2091: We are using `Instant::now()` and `Tai64::now()` directly in the block producer, rather than having some time interface
Time is unreliable IO. We should be able to isolate an manipulate it using a fake/mock rather than being dependent on the system clock.
```rs
    fn extract_block_info(last_block: &BlockHeader) -> (BlockHeight, Tai64, Instant) {
        let duration_since_last_block =
        let last_block_created = Instant::now()
            .checked_sub(duration_since_last_block)
            .unwrap_or(Instant::now());
        let last_height = *last_block.height();
        (last_height, last_timestamp, last_block_created)
    }
 ```

### PR DESCRIPTION
**Pull request by netrome**
## Linked Issues/PRs
Closes #2091 

## Description
This PR introduces a `GetTime` port in the PoA service. The port is implemented in a `SystemTime` adapter matching the current behavior of the PoA service when running live, and in a `TestTime` implementation allowing fine-grained control of how time elapses in tests.

Three additional tests have been added for previously untestable (but correct) behavior:
- `service_test::trigger_tests::interval_trigger_produces_blocks_in_the_future_when_time_is_lagging`
- `service_test::trigger_tests::interval_trigger_produces_blocks_with_current_time_when_block_production_is_lagging`
- `service_test::trigger_tests::interval_trigger_produces_blocks_in_the_future_when_time_rewinds`

## Checklist
- [x] Old behavior is reflected in tests

### Before requesting review

### CODE REVIEW

File: crates/services/consensus_module/poa/src/service_test/trigger_tests.rs [hunk 1]
```diff
@@ -104,7 +104,27 @@ impl DefaultContext {
...
+                block.header_mut().set_time(time);
+                block.header_mut().recalculate_metadata();
+                Ok(UncommittedResult::new(
+                    ExecutionResult {
+                        block,
+                        skipped_transactions: Default::default(),
+                        tx_status: Default::default(),
+                        events: Default::default(),
+                    },
+                    Default::default(),
```
rymnc, line 122: maybe we want to implement `Default` for `ExecutionResult` and convert to the following -
```suggestion
                    ExecutionResult {
                        block,
                        ..Default::default(),
                    },
                    Default::default(),
```

rymnc (commented):
Nice PR, I've left some questions/comments :)

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -199,13 +203,17 @@ where
...
-    fn extract_block_info(last_block: &BlockHeader) -> (BlockHeight, Tai64, Instant) {
+    fn extract_block_info(
+        clock: &C,
+        last_block: &BlockHeader,
+    ) -> (BlockHeight, Tai64, Instant) {
         let last_timestamp = last_block.time();
         let duration_since_last_block =
-            Duration::from_secs(Tai64::now().0.saturating_sub(last_timestamp.0));
+            Duration::from_secs(clock.now().0.saturating_sub(last_timestamp.0));
         let last_block_created = Instant::now()
```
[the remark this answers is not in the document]
netrome: Nope. Unfortunately, tokio time and clock time are two different things that can progress independently of each other. `clock.now()` returns a timestamp that can be communicated with the external world, whereas `Instant::now()` returns an opaque type which can only be used internally to compare against other instants.

I considered making instants a part of the `GetTime` port as well, but since tokio time is already testable (because we can run `tokio::time::pause()`, `tokio::time::advance()` etc.) I figured it would be overkill to do that.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -199,13 +203,17 @@ where
...
 
-    fn extract_block_info(last_block: &BlockHeader) -> (BlockHeight, Tai64, Instant) {
+    fn extract_block_info(
+        clock: &C,
+        last_block: &BlockHeader,
+    ) -> (BlockHeight, Tai64, Instant) {
         let last_timestamp = last_block.time();
         let duration_since_last_block =
-            Duration::from_secs(Tai64::now().0.saturating_sub(last_timestamp.0));
+            Duration::from_secs(clock.now().0.saturating_sub(last_timestamp.0));
```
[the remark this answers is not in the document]
netrome: > thoughts on saturating_sub vs checked_sub ? I think we should error out instead of saturating the result since it might be erroneous.

In general I agree with you. In this case I'm leaning towards thinking `checked_sub` is acceptable here because:

1. This is only relevant if the last block happened in the future, which is a pretty unlikely scenario.
2. The current behavior here is to tolerate previous blocks in the future, but "pretending" that they happened right now. Changing this breaks existing functionality.

I do think there is a lot of room for improvement in how the current code is structured, including this point. I'd be happy to create a follow-up ticket to address this if you think it's worth looking into.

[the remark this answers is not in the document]
netrome: > additionally, we should try to avoid using inner fields explicitly for methods coming from trait impls. maybe you can impl CheckedSub for GetTime,.

Unfortunately, the orphan rule prevents me from implementing `CheckedSub` for `Tai64` (which GetTime::now outputs). The alternative would be to use a wrapper type which feels overkill in the current context. Especially since this is equivalent to the previous behavior.

Although I agree that using a wrapper type and implementing `CheckedSub` would be an improvement to the code.

crates/services/consensus_module/poa/src/service_test.rs
```diff
@@ -82,13 +86,17 @@ use tokio::{
...
 mod trigger_tests;
 
+use test_time::TestTime;
+
 struct TestContextBuilder {
     config: Option<Config>,
     txpool: Option<MockTransactionPool>,
     importer: Option<MockBlockImporter>,
     producer: Option<MockBlockProducer>,
+    start_time: Option<Tai64N>,
```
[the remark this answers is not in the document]
netrome: This is fed to the constructor of `TestTime` which takes a `Tai64N` as input:

```
    pub fn new(start_time: Tai64N) -> Self {
```

so this is correct, but thanks for noting.

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
netrome: Nice suggestion, will do.

crates/services/consensus_module/poa/src/service.rs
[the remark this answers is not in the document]
rymnc: agree, a follow up issue will suffice here :)

crates/services/consensus_module/poa/src/service.rs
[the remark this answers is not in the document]
rymnc: I see! perhaps this can be included in the follow up issue you mentioned :)

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
netrome: Done in [5fed46e](https://github.com/FuelLabs/fuel-core/pull/2145/commits/5fed46e1eb06ba58face9351e99d5f5e5ea78fba)

crates/services/consensus_module/poa/src/service.rs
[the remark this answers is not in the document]
netrome: Follow-up issue created: https://github.com/FuelLabs/fuel-core/issues/2146

crates/services/consensus_module/poa/src/service.rs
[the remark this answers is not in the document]
netrome: Nice, mentioned in https://github.com/FuelLabs/fuel-core/issues/2146#issuecomment-2320487089

crates/services/consensus_module/poa/src/service_test/test_time.rs
```diff
@@ -0,0 +1,81 @@
...
+
+impl TestTime {
+    pub fn at_unix_epoch() -> Self {
+        Self::new(Tai64N::UNIX_EPOCH)
+    }
+
+    pub fn new(start_time: Tai64N) -> Self {
+        tokio::time::pause();
+        let time = watch::Sender::new(start_time);
+        let last_observed_tokio_time = Instant::now();
```
MitchTurner, line 27: IS there a scenario we'd want this to also be at the `start_time`? Are we okay with this always having a relative dependency on system time `now`?

I guess it's a YAGNI thing. We have what we need for the current tests.

crates/services/consensus_module/poa/src/service_test/test_time.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    pub fn watch(&self) -> Watch {
+        let time = self.time.subscribe();
+        Watch { time }
+    }
+}
+
+impl Drop for TestTime {
+    fn drop(&mut self) {
+        tokio::time::resume();
```
MitchTurner, line 69: Interesting that `tokio::time::pause()` doesn't generate some kind of handle that we can drop. 
🐢
🐢
🐢
🐢
🐢
🐢
We're still dealing with some global state :)

crates/services/consensus_module/poa/src/service_test/test_time.rs
```diff
@@ -0,0 +1,81 @@
...
+    Tai64N,
+};
+use tokio::{
+    sync::watch,
+    time::Instant,
+};
+
+use crate::ports::GetTime;
+
+pub struct TestTime {
```
MitchTurner, line 14: How do you feel about naming this `TokioTime`?

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
```diff
@@ -299,3 +317,144 @@ async fn interval_trigger_doesnt_react_to_full_txpool() -> anyhow::Result<()> {
...
+
+    // Then
+
+    // We should only have produced the three blocks.
+    assert!(matches!(
+        ctx.block_import.try_recv(),
+        Err(broadcast::error::TryRecvError::Empty)
+    ));
+
+    // Even though real time is frozen, the blocks should advance into the future.
```
MitchTurner, line 352: I feel like these tests might be confusing domains. 

Maybe we could have tests for the `TestTime` to show that it works independent of "real time", but I feel like the point of the `GetTime` port is that there is no such thing as "real time".

MitchTurner: i.e. This test should just be
```rs
async fn interval_trigger__when_advance_time_three_blocks_worth_then_produces_three_blocks() {}
```
then there can be a test for `TestTime` that is
```rs
async fn now__whe_use_tokio_sleep_then_current_time_advances()
```

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
```diff
@@ -299,3 +317,144 @@ async fn interval_trigger_doesnt_react_to_full_txpool() -> anyhow::Result<()> {
...
+
+    // When
+
+    // We produce the first block in real time.
+    time::sleep(block_time + offset).await;
+    ctx.test_ctx.time.advance_with_tokio();
+    let first_block_time = ctx.block_import.try_recv().unwrap().entity.header().time();
+
+    // And we rewind time before attempting to produce the next block.
+    ctx.test_ctx.time.rewind(block_time);
```
MitchTurner, line 442: `ctx` should just provide a method for this IMO. 

Either that, or we have a separate handle for manipulating time.

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
```diff
@@ -299,3 +317,144 @@ async fn interval_trigger_doesnt_react_to_full_txpool() -> anyhow::Result<()> {
...
+        ..Default::default()
+    });
+    ctx.status_sender.send_replace(Some(TxId::zeroed()));
+    let start_time = ctx.test_ctx.time.watch().now();
+
+    // When
+
+    // We produce the first block in real time.
+    time::sleep(block_time + offset).await;
+    ctx.test_ctx.time.advance_with_tokio();
```
MitchTurner, line 438: Same here.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -199,13 +203,17 @@ where
             predefined_blocks,
             trigger,
             sync_task_handle,
+            clock,
         }
     }
 
-    fn extract_block_info(last_block: &BlockHeader) -> (BlockHeight, Tai64, Instant) {
+    fn extract_block_info(
+        clock: &C,
```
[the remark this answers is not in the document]
netrome: Sure thing, will update :+1:

crates/services/consensus_module/poa/src/service_test/test_time.rs
netrome: The crux here is that there is that `Instant` is only useful for relative time comparisons. Looking at the implementation of `Instant::now()` it boils down to calling the libc `clock_gettime` function with `CLOCK_UPTIME_RAW`, which can't be compared to a timestamp.

I would love to get rid of the system time dependency, but I am not currently aware of a feasible way to do it.

crates/services/consensus_module/poa/src/service_test/test_time.rs
netrome: Yeah, I'm not too happy about managing tokio time from this struct because it is global mutable state which essentially turns this type into a singleton. Though only within the context of a tokio runtime, hence why we can still run the tests in parallel.

Ideally I'd like to abstract away all tokio time management behind a controllable interface, but given our other priorities that felt like over-doing it right now.

crates/services/consensus_module/poa/src/service_test/test_time.rs
netrome: I appreciate the name suggestion, but I don't think it is appropriate here because while this struct tracks tokio time, it can still be advanced and rewinded independently of the tokio time, and the tokio time may be advanced without this struct following it.

The purpose of observing tokio time in this struct is to allow this time to follow the tokio time, but it cannot control it unfortunately due to both the auto-advancing feature of tokio and the availability of any other function to advance or rewind the tokio time.

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
netrome: Sure, I can add a convenience function here.

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
netrome: Sure thing, will do.

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
netrome: > Maybe we could have tests for the TestTime to show that it works independent of "real time", but I feel like the point of the GetTime port is that there is no such thing as "real time".

I guess I could use clearer terminology here. I refer to the tokio time as the "real time", because that's what it is meant to represent, whereas the timestamps are a read from the system clock (which can advance and rewind independently of the tokio time even when running live).

netrome: I could refer to "tokio time" instead of "real time" in the coments.

netrome: > `async fn now__whe_use_tokio_sleep_then_current_time_advances()`

Unfortunately, this is not necessarily the case. Or perhaps I'm misunderstanding you. What would you like to test in this scenario?

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
netrome: Done.

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
netrome: Done.

crates/services/consensus_module/poa/src/service.rs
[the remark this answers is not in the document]
netrome: Done.

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
netrome: For reference, we just had a chat about the concept of time and what we'd like to achieve and arrived at some conclusions:

1. Time in software systems are essentially (at least) two different but related concepts, we can call them the **wall clock** and the **stop watch**. The **wall clock** gives us a comparable time stamp that is understandable by other systems, but doesn't make many other guarantees. It may rewind or fast forward, making it unreliable for timing execution. For example if I want to make a request to an external resource with a 5s timeout at the same time as the wall clock fast forwards, the request may instantly time out. The **stop watch** instead provides us with pretty accurate relative times, allowing us to delay or pause execution of certain tasks for a certain amount of seconds or milliseconds. However, the stop watch does not provide us with a timestamp that we can send to another system.

2. Our current system uses `Tai64` for the wall clock and `tokio::time` as the stop watch. We use both concepts for different parts of the current business logic. The `GetTime` interface introduced in this PR captures the **wall clock** concept, thus making timestamps testable. However, I have not added any stop watch related functionality in the trait because 1. that would be pretty complex and result in a very substantial change and 2. the time functionality provided by tokio is already controllable in our tests.

3. Ideally we would like to reduce these two to a single concept of time, captured in a single trait which can both return a monotonically increasing somewhat accurate timestamp and provide decent control over execution time. Doing this is non-trivial, and will require making tradeoffs between execution time predictability and timestamp accuracy. This would also alter the current system behavior.

As exciting as it would be to dig into 3. we agree that it is beyond the scope of what is reasonable to do right now given our other priorities.

netrome: @MitchTurner please let me know if there's anything you'd like to add or change in the above summary.

MitchTurner (approved):

MitchTurner:
Sorry about the slow review.

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
```diff
@@ -299,3 +333,144 @@ async fn interval_trigger_doesnt_react_to_full_txpool() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn interval_trigger_produces_blocks_in_the_future_when_time_is_lagging() {
+    // Given
```
acerone85, line 339: nit: something missing here?

crates/services/consensus_module/poa/src/service_test.rs
```diff
@@ -168,18 +177,26 @@ impl TestContextBuilder {
 
         let predefined_blocks = HashMap::new().into();
 
+        let time = match self.start_time {
+            Some(start_time) => TestTime::new(start_time),
+            None => TestTime::at_unix_epoch(),
+        };
```
acerone85, line 183: if you give TestTime a Default implementation that returns `TestTime::at_unix_epoch()`, you can reduce this to 

```rust
        let time = self.start_time.map(TestTime::new).unwrap_or_default();
```

crates/services/consensus_module/poa/src/service_test/trigger_tests.rs
netrome: Not sure what you refer to here. The test is broken down into three sections "Given, When, Then", and this is the beginning of the first one.

crates/services/consensus_module/poa/src/service_test.rs
netrome: Good point, will do.

netrome: Done in https://github.com/FuelLabs/fuel-core/pull/2145/commits/288d9031ec8207c1375a8882ca733c41fc5c9486

rymnc (commented):
lgtm if tests pass 😎

MitchTurner (approved):

rymnc (approved):

Issue #1560, opened 2023-12-18
Add `connection_limits` behavior to `FuelBehavior`


MitchTurner:
Before upgrading `libp2p` we managed our connection limits through the `SwarmBuilder`. That functionality has since been removed from the `SwarmBuilder` and moved to a dedicated `connection_limit` behavior:
https://github.com/libp2p/rust-libp2p/blob/master/misc/connection-limits/src/lib.rs

We need to add tests that verify desired connection limiting and then add the behavior to fulfill our requirements :)

Issue #904, opened 2023-01-13
Add config options for max pending connections, for both outgoing and incoming


Voxelot:
... this defines the number of max established "incoming" connections, I arbitrarily set it to be 50% of total connections allowed, so 50:50 between incoming and outgoing connections. But we can have this configurable in p2p config if we want, plus maybe we should add config options for max pending connections, for both outgoing and incoming.

_Originally posted by @leviathanbeak in https://github.com/FuelLabs/fuel-core/pull/876#discussion_r1065933670_

MitchTurner:
We have an open PR that has been inactive for a while.

Issue #1968, opened 2024-06-14
Improve P2P DoS resilience


xgreenx:
Currently, the external sentries in the network are very vulnerable to external attackers. The attacker can spam nodes with many connections, requests, and heavy data-consuming sentry resources.

It can halt the node or cause resource exhaustion or OOM. Following issue highlights part of the vectors of atackes:
- https://github.com/FuelLabs/fuel-core/issues/904
- https://github.com/FuelLabs/fuel-core/issues/943
- https://github.com/FuelLabs/fuel-core/issues/1567
- https://github.com/FuelLabs/fuel-core/issues/1677
- https://github.com/FuelLabs/fuel-core/issues/1774 - Without it attacker can consume all available connections

PR #2150, opened 2024-09-01
feat: add pending connection limits + bump libp2p to `0.54.1`

maschad:
## Linked Issues/PRs

- Closes https://github.com/FuelLabs/fuel-core/issues/1560
- Closes https://github.com/FuelLabs/fuel-core/issues/904

## Description

This is apart of the broader initiative outlined in https://github.com/FuelLabs/fuel-core/issues/1968 to improve the DoS resilience 

This PR doesn't make the pending connection limits configurable for a reason, it is an easy way for the operator to footgun and mess up the node configuration. We've used some defaults for now, but there should be some analysis done on optimal connection count for latency to propagate messages.

## Checklist

### Before requesting review


bin/fuel-core/src/cli/run/p2p.rs
```diff
@@ -95,6 +95,15 @@ pub struct P2PArgs {
...
     pub max_connections_per_peer: u32,
 
+    /// Max number of concurrent pending incoming connections
+    /// Useful in mitigating against DDoS attacks
+    #[clap(long = "max-pending-incoming-connections", default_value = "100", env)]
+    pub max_pending_incoming_connections: u32,
+
+    /// Max number of concurrent pending outgoing connections
+    #[clap(long = "max-pending-outgoing-connections", default_value = "100", env)]
+    pub max_pending_outgoing_connections: u32,
```
maschad, line 105: Even though I've added this parameter, I'm not fully sold that it should be present. The rationale is we only dial a peer if: 

a) We are below the desired amount of peers in a gossipsub mesh (which would imply we would be below any potential `max_pending_outgoing_connections` anyway) and we determined that the peer in question would have a reasonable peer score. So essentially our gossipsub config should handle our scenario.

AND 

b) We are relatively certain that peer is reachable i.e. the `multiaddr` for that peer can be successfully dialled - if that is so, we want to prioritize dialling it.

But I may be missing something about the current p2p setup.

MitchTurner: You are unsure if we should put a limit on outgoing connections?

maschad: Correct, more specifically on outgoing pending connections (as opposed to established outgoing connections)

AurelienFT (commented):
For me that's an interesting add but it's doesn't close : https://github.com/FuelLabs/fuel-core/issues/1560. IMO, in order to close this issue we also need to add parameters for the number of established connections (in/out)

AurelienFT (dismissed):
Looks good to me. Little nit on the CHANGELOG. Can you update it now that you have a new parameter added ?

maschad:
> Looks good to me. Little nit on the CHANGELOG. Can you update it now that you have a new parameter added ?

Good catch @AurelienFT  , I've updated it in 1127c7200

AurelienFT (dismissed):

crates/services/p2p/src/p2p_service.rs
```diff
@@ -180,7 +180,7 @@ impl FuelP2PService {
 
         // configure and build P2P Service
         let (transport_function, connection_state) = build_transport_function(&config);
-        let tcp_config = tcp::Config::new().port_reuse(true);
+        let tcp_config = tcp::Config::new();
```
Dentosal, line 183: What's the reason for removing `.port_reuse(true)`?

xgreenx:
We shouldn't merge this PR before https://github.com/FuelLabs/fuel-core/pull/2131

Plus, let's wait for my review as well, since maybe we need to clean up some other logic

crates/services/p2p/src/p2p_service.rs
maschad: This option does nothing now, since the port reuse policy is now decided on a per-connection basis by the behaviour.  It has [deprecated](https://github.com/libp2p/rust-libp2p/blob/master/transports/tcp/src/lib.rs#L185-L189) since `0.42.0`

MitchTurner:
How's this coming?

maschad:
> How's this coming?

Thanks for resurfacing this @MitchTurner  it got _lost in the ether_ with the mainnet whirlwind of changes.

I've resolved the conflicts so it's ready for review again. I know @xgreenx wanted to look into it again but the changes imo are relatively low surface - I've upgraded our libp2p version + introduced connection limit options.

AurelienFT:
Hello @maschad,

As you mentionned that @xgreenx wanted to look again we will wait him before merging this but I can already re-approve as a first approver. However before approving the conflicts should be resolved, are you ok to resolve them ? Or we can do it for you, if you prefer :)

Thanks again for this PR.

maschad:
Thanks @AurelienFT I've gone ahead and resolved the conflicts.

AurelienFT (approved):

rymnc:
@maschad is it okay if I take over this PR? there seem to be some formatting issues

AurelienFT (dismissed):
LGTM. Ty @rymnc for takeover

Issue #1970, opened 2024-06-14
Unlimited dry running of the transactions


xgreenx:
## Problem overview

The executor ignores the block gas limit during the dry running of transactions. It allows the attacker to submit a lot of transactions to dry run in one request, consuming the endpoint entirely.

## Solution

1. Update the executor's dry run logic to respect the block gas limit.
1. And upgrade the `dry_run` endpoint to deserialize transactions one by one and check that cumulative gas less than block gas limit.
<img width="1524" alt="image" src="https://github.com/FuelLabs/fuel-core/assets/18346821/c1a6e5a9-6403-442e-8c49-73d901a217ae">

PR #2151, opened 2024-09-02
Add verification on transaction dry_run that they don't spend more than block gas limit

AurelienFT:
## Linked Issues/PRs
Close https://github.com/FuelLabs/fuel-core/issues/1970

## Description
Add check that transactions submitted to the graphql endpoint `dry_run` doesn't spend too much gas. Also add this check in the executor level.

### Still todo
Add tests

## Checklist

### Before requesting review


crates/fuel-core/src/schema/tx.rs
```diff
@@ -276,16 +277,43 @@ impl TxMutation {
...
+                    max_gas_usable += tx.max_gas(consensus_params.gas_costs(), fee)
+                }
+                FuelTx::Upload(tx) => {
+                    max_gas_usable += tx.max_gas(consensus_params.gas_costs(), fee)
+                }
+                FuelTx::Blob(tx) => {
+                    max_gas_usable += tx.max_gas(consensus_params.gas_costs(), fee)
+                }
+            };
+            if max_gas_usable > block_gas_limit {
```
xgreenx, line 313: I was thinking maybe to add this logic directly into the `dry_run` method of the executor=)

AurelienFT: In the issue, I understand that you wanted both. I also made the modification in the executor I can remove the graphql part.

AurelienFT: Ok I was thinking you wanted in the API AND in the executor, will change that :)

tests/tests/tx.rs
```diff
@@ -145,6 +145,44 @@ async fn dry_run_create() {
...
+    let tx = TransactionBuilder::script(script, vec![])
+        .script_gas_limit(gas_limit)
+        .maturity(maturity)
+        .add_random_fee_input()
+        .finalize_as_transaction();
+
+    // When
+    client.dry_run(&[tx.clone()]).await
+    // Then
+    .expect_err("Dry run should return an error because transaction is using too much gas");
```
netrome, line 183: Can we explicity match on the error variant just to be clear here? Otherwise this test case could pas on a completely irrelevant error.

crates/fuel-core/src/schema/tx.rs
```diff
@@ -276,16 +277,48 @@ impl TxMutation {
...
+                    max_gas_usable = max_gas_usable
+                        .saturating_add(tx.max_gas(consensus_params.gas_costs(), fee));
+                }
+                FuelTx::Create(tx) => {
+                    max_gas_usable = max_gas_usable
+                        .saturating_add(tx.max_gas(consensus_params.gas_costs(), fee));
+                }
+                FuelTx::Mint(_) => {
+                    // Mint transactions doesn't consume gas and so we will add a flat 10% of the block gas limit
+                    max_gas_usable = max_gas_usable.saturating_add(block_gas_limit / 10);
```
xgreenx, line 303: People shouldn't submit `Mint` transaction. It is only generated by a block producer. So we need to return an error=)

crates/fuel-core/src/schema/tx.rs
xgreenx: In the case of dry run we currently ignore `remaining_gas_limit`

crates/fuel-core/src/schema/tx.rs
```diff
@@ -276,16 +277,48 @@ impl TxMutation {
...
+                }
+                FuelTx::Blob(tx) => {
+                    max_gas_usable = max_gas_usable
+                        .saturating_add(tx.max_gas(consensus_params.gas_costs(), fee));
+                }
+            };
+            if max_gas_usable > block_gas_limit {
+                return Err(anyhow::anyhow!("The sum of the gas usable by the transactions is greater than the block gas limit").into());
+            }
+            transaction.precompute(&consensus_params.chain_id())?;
```
xgreenx, line 321: You can calculate it before `max_gas`

crates/services/producer/src/block_producer.rs
```diff
@@ -280,6 +280,7 @@ where
         gas_price: Option<u64>,
     ) -> anyhow::Result<Vec<TransactionExecutionStatus>> {
         let view = self.view_provider.latest_view()?;
+
```
xgreenx, line 283:
```suggestion
```

tests/tests/tx.rs
```diff
@@ -145,6 +145,44 @@ async fn dry_run_create() {
...
+    let script = [
+        op::addi(0x10, RegId::ZERO, 0xca),
+        op::addi(0x11, RegId::ZERO, 0xba),
+        op::log(0x10, 0x11, RegId::ZERO, RegId::ZERO),
+        op::ret(RegId::ONE),
+    ];
+    let script: Vec<u8> = script
+        .iter()
+        .flat_map(|op| u32::from(*op).to_be_bytes())
+        .collect();
```
xgreenx, line 172:
```suggestion
    let script = script.into_iter().collect();
```

crates/fuel-core/src/schema/tx.rs
AurelienFT: @xgreenx Could you point me to this ? Because I don't see where is the behavior difference

crates/fuel-core/src/schema/tx.rs
AurelienFT: Why do we prefer to precompute first ? Because if we reach the maximum gas it will avoid us making a precompute for nothing

xgreenx: Yeah, it makes sense, let's keep it at the end then

tests/tests/tx.rs
AurelienFT: Done

crates/fuel-core/src/schema/tx.rs
```diff
@@ -276,16 +277,48 @@ impl TxMutation {
...
+                }
+                FuelTx::Upload(tx) => {
+                    max_gas_usable = max_gas_usable
+                        .saturating_add(tx.max_gas(consensus_params.gas_costs(), fee));
+                }
+                FuelTx::Blob(tx) => {
+                    max_gas_usable = max_gas_usable
+                        .saturating_add(tx.max_gas(consensus_params.gas_costs(), fee));
+                }
+            };
```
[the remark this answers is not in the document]
AurelienFT: Done

netrome (dismissed):

xgreenx (commented):
I still think that we need to handle the gas before `match self.execute_transaction_and_commit(` since the `TxSource` can lie/ignore remaining gas.
image

crates/fuel-core/src/schema/tx.rs
AurelienFT: Done.

xgreenx (approved):

crates/services/executor/src/ports.rs
```diff
@@ -54,6 +66,53 @@ impl MaybeCheckedTransaction {
     }
 }
 
+pub trait TransactionExt {
+    fn max_gas(&self, consensus_params: &ConsensusParameters) -> ExecutorResult<u64>;
+}
+
```
netrome, line 72: Nit: Since this is a one-method trait it would be easier to read the intention of it if it was named after the function: `pub trait MaxGas`.

netrome (approved):
LGTM, one nit.

Issue #2049, opened 2024-07-25
Run benchmarks for fuel-vm 0.56


Dentosal:
New opcodes were introduced, and `LDC` has been changed as well.

PR #2153, opened 2024-09-02
Update default gas costs based on 0.35.0 benchmarks

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/2049

Results from the benchmark with `fuel-core 0.35.0`. The gas cost decreased for all that somehow touched storage or balances because of optimization for SMT not to request siblings.

## Checklist


benches/src/lib.rs
```diff
@@ -418,6 +453,12 @@ impl TryFrom<VmBench> for VmBenchPrepared {
             db.deploy_contract_with_id(&slots, &contract, &id)?;
         }
 
+        if let Some(BlobCode { code, id }) = blob {
+            db.storage_as_mut::<BlobData>()
+                .insert(&id, &code.0)
+                .unwrap();
+        }
```
netrome, line 460: Any reason why we unwrap here instead of just returning an error?

```suggestion
        if let Some(BlobCode { code, id }) = blob {
            db.storage_as_mut::<BlobData>()
                .insert(&id, &code.0)?;
        }
```

netrome (approved):
Looks good to me, though there's still a lot I feel I cannot fully verify. One thing that stands out to me though is how much larger the benchmarks file has become. Is that due to the extra blobcode benchmarks?

image

xgreenx:
> Looks good to me, though there's still a lot I feel I cannot fully verify. One thing that stands out to me though is how much larger the benchmarks file has become. Is that due to the extra blobcode benchmarks?

Github just shows it incorrectly=) If you open removed file, you can see that this file had the same number of lines(almost)

netrome:
> > Looks good to me, though there's still a lot I feel I cannot fully verify. One thing that stands out to me though is how much larger the benchmarks file has become. Is that due to the extra blobcode benchmarks?
> 
> Github just shows it incorrectly=) If you open removed file, you can see that this file had the same number of lines(almost)

Ah interesting. Thanks for the clarification! 🙏

netrome (approved):

netrome (approved):

benches/benches/vm_set/blockchain.rs
```diff
@@ -441,6 +483,33 @@ pub fn run(c: &mut Criterion) {
 
     csiz.finish();
 
+    let mut bsiz = c.benchmark_group("bsiz");
```
MitchTurner, line 486: How were we getting the values for `bldd` and `bsiz` before? It looks like they didn't exist before.

benches/benches/vm_set/mem.rs
```diff
@@ -77,7 +86,8 @@ pub fn run(c: &mut Criterion) {
         cfei_linear.push(1_000_000);
         cfei_linear.push(10_000_000);
         for i in cfei_linear {
-            let bench = VmBench::new(op::cfei(i as u32));
+            let memory = MemoryInstance::from(vec![123; MEM_SIZE]);
+            let bench = VmBench::new(op::cfei(i as u32)).with_memory(memory);
```
MitchTurner, line 90: Why did we need the `memory` added here?/Why didn't we have it before?

benches/src/bin/collect.rs
```diff
@@ -642,7 +642,7 @@ fn linear_regression(x_y: Vec<(u64, u64)>) -> f64 {
 }
 
 fn dependent_cost(name: &String, x_y: Vec<(u64, u64)>) -> DependentCost {
-    const NEAR_LINEAR: f64 = 0.1;
+    const NEAR_LINEAR: f64 = 0.2;
```
rafal-ch, line 645: Shall we also change the comment in [this line](https://github.com/FuelLabs/fuel-core/pull/2153/files#diff-4ce720017e878bdefcfe9d40e80692be37fb307f5fcff3b12d271c6dad0ed694R650) to:
```
/// and the last points are almost the same(The difference is < 0.2).
```
or maybe even
```
/// and the last points are almost the same(The difference is < NEAR_LINEAR).
```

MitchTurner (commented):
LGTM. Just some quick questions to understand everything that happened with this PR.

benches/benches/vm_set/mem.rs
```diff
@@ -77,7 +86,8 @@ pub fn run(c: &mut Criterion) {
         cfei_linear.push(1_000_000);
         cfei_linear.push(10_000_000);
         for i in cfei_linear {
-            let bench = VmBench::new(op::cfei(i as u32));
+            let memory = MemoryInstance::from(vec![123; MEM_SIZE]);
```
rafal-ch, line 89: `123` is just an arbitrary value, yes?

rafal-ch (approved):
LGTM :+1: 

However, it's not clear for me how exactly the numbers in `GasCostsValuesV4::default_gas_costs()` were calculated.

benches/benches/vm_set/blockchain.rs
xgreenx: We had a task to add benchamrks for these upcodes: https://github.com/FuelLabs/fuel-core/issues/2049

Before we just copied values from similar contract opcodes

benches/benches/vm_set/mem.rs
xgreenx: Yep

MitchTurner (approved):

rafal-ch (approved):

PR #2157, opened 2024-09-03
Get DA costs from predefined data

MitchTurner:
## Linked Issues/PRs
https://github.com/FuelLabs/fuel-core/issues/2138

### Before requesting review


crates/fuel-gas-price-algorithm/gas-price-analysis/src/main.rs
```diff
@@ -70,83 +58,141 @@ enum Mode {
...
+    let eth = *max_actual_profit as f64 / (10_f64).powf(18.);
+    println!("max actual profit: {} ETH at {}", eth, index);
+
+    // Max projected profit
+    let (index, max_projected_profit) = projected_profit
+        .iter()
+        .enumerate()
+        .max_by(|(_, a), (_, b)| a.cmp(b))
+        .unwrap();
+    let eth = *max_projected_profit as f64 / (10_f64).powf(18.);
```
rymnc, line 142: nit: I see this used in multiple places. worth an inline fn?
```suggestion
#[inline]
fn fuel_to_eth(fuel_units: i128) -> f64 {
  (fuel_units as f64) / (10_f64).powf(18.)
}
...
    let eth = fuel_to_eth(*max_projected_profit);
```

crates/fuel-gas-price-algorithm/gas-price-analysis/src/main.rs
```diff
@@ -70,83 +58,141 @@ enum Mode {
...
+        .min_by(|(_, a), (_, b)| a.cmp(b))
+        .unwrap();
+    let eth = *min_da_gas_price as f64 / (10_f64).powf(18.);
+    println!(
+        "min DA gas price: {} Wei ({} ETH) at {}",
+        min_da_gas_price, eth, index
+    );
 }
 
+pub fn pretty<T: ToString>(input: T) -> String {
```
rymnc, line 188:
```suggestion
pub fn prettify_numbers<T: Display>(input: T) -> String {
```

nit: improved naming and usage of `Display` since everyone probably implements Display and gets ToString as a side-effect, ToString is okay though, not a hard req

crates/fuel-gas-price-algorithm/gas-price-analysis/src/optimisation.rs
```diff
@@ -11,13 +11,13 @@ fn da_pid_factors(size: usize) -> Vec<(i64, i64)> {
         .collect()
 }
 
-pub fn naive_optimisation(iterations: usize, size: usize) -> (SimulationResults, (i64, i64)) {
+pub fn naive_optimisation(simulator: Simulator, iterations: usize, da_recording_rate: usize) -> (SimulationResults, (i64, i64)) {
```
rymnc, line 14:
```suggestion
pub fn naive_optimisation(simulator: &Simulator, iterations: usize, da_recording_rate: usize) -> (SimulationResults, (i64, i64)) {
```

can be a reference? acc to the signature: https://github.com/FuelLabs/fuel-core/pull/2157/files#diff-a5477e1747a3035f9dcb256499945866dd5643c5988a0e5534de0752652b4bdbR39

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
```diff
@@ -1,162 +1,220 @@
...
+            // latest_da_cost_per_byte: *starting_da_gas_price as u128,
+            latest_da_cost_per_byte: 0,
+            projected_total_da_cost: 0,
+            latest_known_total_da_cost: 0,
+            unrecorded_blocks: vec![],
+            da_p_component,
+            da_d_component,
+            last_profit: 0,
+            second_to_last_profit: 0,
+        };
```
rymnc, line 99: for people with not much context, maybe a few comments about these fields and how they affect the gas price calc would be nice, i.e how the gas price scales with the increase/decrease in the given value

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation/da_cost_per_byte.rs
```diff
@@ -0,0 +1,95 @@
...
+        }
+        costs.push(cost);
+    }
+    println!("Max cost: {}", pretty(max_cost));
+    costs
+}
+
+fn noisy_eth_price<T: TryInto<f64>>(input: T) -> f64
+where
+    <T as TryInto<f64>>::Error: core::fmt::Debug,
```
rymnc, line 74: curious, is this constraint required?

rymnc (commented):
nice PR, added some comments/questions in-line.
in general, I think we should try sticking to one type for all calculations, there are a lot of places where there are explicit casts and conversions, can be an enhancement :)

MitchTurner:
@rymnc definitely. The casts are new and kinda ruin the checked integer math. We have the follow-up issue to deal with the overflow problems, which is related to why we have the casts now.

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
```diff
@@ -1,162 +1,220 @@
...
-    SimulationResults {
-        gas_prices,
-        exec_gas_prices,
-        da_gas_prices,
-        fullness,
-        bytes_and_costs,
-        actual_profit,
-        projected_profit,
-        pessimistic_costs,
+    fn zip_l2_blocks_with_da_blocks(
```
rafal-ch, line 187: A nit, but this could use some basic UT to see if it produces expected data.

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
```diff
@@ -1,162 +1,220 @@
...
+        let updater = AlgorithmUpdaterV1 {
+            min_exec_gas_price: 10,
+            min_da_gas_price: 10,
+            new_scaled_exec_price: 10 * gas_price_factor,
+            // last_da_gas_price: *starting_da_gas_price,
+            last_da_gas_price: 100,
+            gas_price_factor: NonZeroU64::new(gas_price_factor).unwrap(),
+            l2_block_height: 0,
+            l2_block_fullness_threshold_percent: 50,
+            exec_gas_price_change_percent: 2,
```
rafal-ch, line 86: It would be nice to be able to read more of those parameters from cmd line. Could be in a different PR, though.

crates/fuel-gas-price-algorithm/gas-price-analysis/src/charts.rs
```diff
@@ -181,9 +243,9 @@ pub fn draw_bytes_and_cost_per_block(
 
 pub fn draw_profit(
     drawing_area: &DrawingArea<BitMapBackend, Shift>,
-    actual_profit: &[i64],
-    projected_profit: &[i64],
-    pessimistic_block_costs: &[u64],
+    actual_profit: &[i128],
```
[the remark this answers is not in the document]
MitchTurner: Will look into this with followup issue.

crates/fuel-gas-price-algorithm/gas-price-analysis/src/main.rs
MitchTurner: It's a good idea, but not all of the types are `i128`. And `Into<f64>` doesn't exist for `i128`. Have to rely on the cast for now. Will get better when I clean up these types in the followup issue.

(Unless there is a way to constrain by casts?)

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation/da_cost_per_byte.rs
MitchTurner: Which part. I'm using it for more than one `From` type, hence the `TryInto`, and the `Debug` is for the `unwrap` :P

image


I guess we're moving away from `unwrap` in the charts. Maybe worth doing it for the simulation code too.

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
MitchTurner: As a rule of thumb, I typically don't test internal/private functions. I understand that having some more confidence in this simulation code would be good though.

I don't think I want to hold this PR for this, but I'll think of ways I can test this all better.

rafal-ch: Makes sense, thanks for explanation.

rafal-ch (approved):
LGTM :+1:

crates/fuel-gas-price-algorithm/gas-price-analysis/src/main.rs
rymnc: yup, idt there is a way to constrain by casts. resolving.

crates/fuel-gas-price-algorithm/gas-price-analysis/src/charts.rs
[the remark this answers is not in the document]
rymnc: sounds good! resolving

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation/da_cost_per_byte.rs
rymnc: eh, yeah ~ feel free to ignore

rymnc (approved):
lgtm, thanks for addressing my comments.

rafal-ch (approved):

PR #2155, opened 2024-09-03
feat(gas_price_service): include trait declaration for block committer data

rymnc:
> [!WARNING]
> 🏗️ This PR is 1/2 of including block committer "da"ta in v1 of the gas price algo. We have a dependency on the api of the block committer, which can be included in a follow up PR when we have more details 🚧

## Linked Issues/PRs

- https://github.com/FuelLabs/fuel-core/issues/2139

## Description

This PR includes the following changes - 
- new trait `DaCommitSource` which specifies `get_da_commit_details` that returns the latest da commit details from the block committer
- removed usages of std lib and replaced with `core::*`
- refactored to make functions shorter and readable

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog
- [x] New behavior is reflected in tests
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [x] I have reviewed the code myself
- [x] I have created follow-up issues caused by this PR and linked them here

### After merging, notify other teams

[Add or remove entries as needed]

- [ ] [Rust SDK](https://github.com/FuelLabs/fuels-rs/)
- [ ] [Sway compiler](https://github.com/FuelLabs/sway/)
- [ ] [Platform documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+) (for out-of-organization contributors, the person merging the PR will do this)
- [ ] Someone else?

Issue #2139, opened 2024-08-28
Create committer DA block provider for V1 gas price updater


MitchTurner:
The V1 gas price algorithm is going to need DA block data from our committer. We need to create an adapter for that.

rymnc:
addressed in https://github.com/FuelLabs/fuel-core/pull/2265

PR #2163, opened 2024-09-05
feat(gas_price_service): create runnable task for expensive background polling for da metadata

rymnc:
> [!WARNING]
> 🏗️ This PR is 2/3 of including block committer "da"ta in v1 of the gas price algo. We have a dependency on the api of the block committer, which will warrant tweaking of the `DaMetadataResponse` type. The next PR will hook it up with cli args 🚧

## Linked Issues/PRs

- follow up to: https://github.com/FuelLabs/fuel-core/pull/2155
- https://github.com/FuelLabs/fuel-core/issues/2139

## Description

- specified a Service that polls an ingestor at specified intervals to fetch the da metadata in the background
- specified a "sync"/"non-expensive" api for the data, which marks it as stale/None after fetching it

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/gas_price_service/src/fuel_gas_price_updater.rs
```diff
@@ -113,16 +117,22 @@ pub trait L2BlockSource: Send + Sync {
     async fn get_l2_block(&mut self, height: BlockHeight) -> Result<BlockInfo>;
 }
 
-#[derive(Debug, Default)]
-pub struct DaCommitDetails {
+#[derive(Debug, Default, Clone)]
+pub struct DaGasPrice {
```
MitchTurner, line 121: I think this should be named something like `DaBlockCosts` or something like that. We use this value to calculate the "Price" but it isn't the price yet.

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/dummy_source.rs
```diff
@@ -0,0 +1,14 @@
+use crate::fuel_gas_price_updater::{
+    service::Result as DaGasPriceSourceResult,
+    DaGasPrice,
+    DaGasPriceSource,
+};
+
+pub struct DummyDaGasPriceSource;
```
MitchTurner, line 7: Maybe `DummyDaBlockCosts` would be better.

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
```diff
@@ -1,16 +1,145 @@
...
+        }
+    }
+
+    type TestValidService =
+        DaGasPriceProviderService<DummyDaGasPriceSource, DaGasPriceProvider>;
+    type TestErroringService =
+        DaGasPriceProviderService<ErroringSource, DaGasPriceProvider>;
+
+    #[tokio::test]
+    async fn test_service_sets_cache_when_request_succeeds() {
```
MitchTurner, line 91: nit: I think we should be a little clearer on what the SUT of these tests is. I think it's safe to say that it's actually `run`, not `start`? The service runner is just calling `run` for us.
```suggestion
    async fn run__when_da_block_cost_source_gives_value_shared_value_is_updated() {
```

Here is a WIP doc on testing standards that talks about this a little:
https://www.notion.so/fuellabs/Engineering-Style-Guide-83ba6fd81f12443796c814ecfc746b39?pvs=4#f3232125ea964e959a66b9f41267d96e

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
```diff
@@ -1,16 +1,145 @@
...
+        sleep(Duration::from_millis(10)).await;
+        service.stop();
+
+        // then
+        let da_gas_price_opt = shared_state.get().unwrap();
+        assert!(da_gas_price_opt.is_some());
+    }
+
+    #[tokio::test]
+    async fn test_service_invalidates_cache() {
```
MitchTurner, line 110: This could use a rename like above.

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
```diff
@@ -1,16 +1,145 @@
...
+        service.stop();
+        let _ = shared_state.get().unwrap();
+
+        // then
+        let da_gas_price_opt = shared_state.get().unwrap();
+        assert!(da_gas_price_opt.is_none());
+    }
+
+    #[tokio::test]
+    async fn test_service_does_not_set_cache_when_request_fails() {
```
MitchTurner, line 129: This could use a rename like above.

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_source.rs
```diff
@@ -0,0 +1,60 @@
...
+use anyhow::anyhow;
+use reqwest::Url;
+use serde::{
+    Deserialize,
+    Serialize,
+};
+
+/// This struct is used to denote the block committer da gas price source,,
+/// which receives data from the block committer (only http api for now)
+pub struct BlockCommitterDaGasPriceSource {
```
MitchTurner, line 15: Maybe `BlockCommitterDaBlockCosts` or something like that would be better

crates/services/gas_price_service/src/fuel_gas_price_updater.rs
```diff
@@ -113,16 +117,22 @@ pub trait L2BlockSource: Send + Sync {
...
-pub trait DaCommitSource: Send + Sync {
-    fn get_da_commit_details(&mut self) -> Result<Option<DaCommitDetails>>;
+pub trait GetDaGasPriceFromSink: Send + Sync + Default + Clone {
+    fn get(&mut self) -> Result<Option<DaGasPrice>>;
+}
+
+// ensures that the da gas price is set to the sink only by
+// selected impls. we don't want the algo updater to have
+// access to this method.
+pub trait SetDaGasPriceToSink: Send + Sync + Default + Clone {
```
MitchTurner, line 134: I don't think this trait needs to exist. 

(Committer <- CommitterDaBlockCostSource Adapter -> (DaBlockCostSource Port <- DaBlockCostService -> `Arc<Mutex<_>>` <- DaGasPriceProvider Adapter -> (GetDaBlockCost Port <- GasPriceUpdaterService)))

Arrows are in the direction of dependencies. **The `DaBlockCostService` is part of the `DaGasPriceProvider` adapter.** We don't need to isolate `DaBlockCostService` from `Arc<Mutex<_>>` since it's job is to get data from some `DaBlockCostSource` and write to an `Arc<Mutex<_>>`. We insulate ourselves from the dependency on `Arc<Mutex<_>>` on the get side.

I've added the "((()))" around the design to show why Ports and Adapters is often called "onion architecture": https://medium.com/expedia-group-tech/onion-architecture-deed8a554423 
It's because the dependencies go outward, not inward, from the domain. Also related:
https://en.wikipedia.org/wiki/Dependency_inversion_principle

crates/services/gas_price_service/src/fuel_gas_price_updater.rs
```diff
@@ -113,16 +117,22 @@ pub trait L2BlockSource: Send + Sync {
...
+pub struct DaGasPrice {
     pub l2_block_range: core::ops::Range<u32>,
     pub blob_size_bytes: u32,
     pub blob_cost_wei: u32,
-    pub partial_block_heights: Option<[u32; 2]>,
 }
 
-pub trait DaCommitSource: Send + Sync {
-    fn get_da_commit_details(&mut self) -> Result<Option<DaCommitDetails>>;
+pub trait GetDaGasPriceFromSink: Send + Sync + Default + Clone {
```
MitchTurner, line 127: This name leaks implementation details. Doesn't `Sink` imply something like `Arc<Mutex<_>>`. Either way, `GetDaBlockCosts` would be plenty context, don't you think?

MitchTurner (commented):
Looking good. I've made some general suggestions wrt architecture and nomenclature that I hope will help.

crates/services/gas_price_service/src/fuel_gas_price_updater.rs
rymnc: addressed in https://github.com/FuelLabs/fuel-core/pull/2163/commits/74fc9bf4e08cb0d4a1a5def23f31d2c4fbd31c9f#diff-a87f94b204ea2e176921aba2703aeedb00fad9254c2157038d735d9d53dc889dR95-R98

crates/services/gas_price_service/src/fuel_gas_price_updater.rs
rymnc: addressed in https://github.com/FuelLabs/fuel-core/pull/2163/commits/74fc9bf4e08cb0d4a1a5def23f31d2c4fbd31c9f#diff-d0929c84f197be22ea7a8f1bc32981a49731a6ec12250819e9c86ebba1373ffaL43-L54

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_source.rs
rymnc: addressed in https://github.com/FuelLabs/fuel-core/pull/2163/commits/74fc9bf4e08cb0d4a1a5def23f31d2c4fbd31c9f#diff-d6d135691e84a45a0d07553a66a5a6db1bf760fcfdd0b626c15a6e966ca3e09cR15-R48

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
rymnc: addressed in https://github.com/FuelLabs/fuel-core/pull/2163/commits/74fc9bf4e08cb0d4a1a5def23f31d2c4fbd31c9f#diff-d0929c84f197be22ea7a8f1bc32981a49731a6ec12250819e9c86ebba1373ffaR71

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
rymnc: addressed in https://github.com/FuelLabs/fuel-core/pull/2163/commits/74fc9bf4e08cb0d4a1a5def23f31d2c4fbd31c9f#diff-d0929c84f197be22ea7a8f1bc32981a49731a6ec12250819e9c86ebba1373ffaR90

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/dummy_source.rs
rymnc: addressed in https://github.com/FuelLabs/fuel-core/pull/2163/commits/74fc9bf4e08cb0d4a1a5def23f31d2c4fbd31c9f#diff-3985ff8ef4e8622f1508b6665a9bf061b8a02fed5c53b1abfaecf072e88ceaa2R7

crates/services/gas_price_service/src/fuel_gas_price_updater.rs
rymnc: addressed in https://github.com/FuelLabs/fuel-core/pull/2163/commits/74fc9bf4e08cb0d4a1a5def23f31d2c4fbd31c9f#diff-f8482408ef6bf8c1120f3a3a5844ee9cdc5281443f3eea46611822ababd3c7a9R121

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
```diff
@@ -1,16 +1,145 @@
...
+    impl DaGasPriceSource for ErroringSource {
+        async fn get(&mut self) -> DaGasPriceSourceResult<DaGasPrice> {
+            Err(anyhow!("boo!"))
+        }
+    }
+
+    type TestValidService =
+        DaGasPriceProviderService<DummyDaGasPriceSource, DaGasPriceProvider>;
+    type TestErroringService =
+        DaGasPriceProviderService<ErroringSource, DaGasPriceProvider>;
```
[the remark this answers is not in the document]
rymnc: addressed in https://github.com/FuelLabs/fuel-core/pull/2163/commits/74fc9bf4e08cb0d4a1a5def23f31d2c4fbd31c9f#diff-d0929c84f197be22ea7a8f1bc32981a49731a6ec12250819e9c86ebba1373ffaR74

MitchTurner (approved):

MitchTurner (dismissed):

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_costs.rs
```diff
@@ -0,0 +1,60 @@
...
+/// This struct is used to denote the block committer da gas price source,,
+/// which receives data from the block committer (only http api for now)
+pub struct BlockCommitterDaBlockCosts {
+    client: reqwest::Client,
+    url: Url,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone, Default, PartialEq)]
+struct RawDaBlockCosts {
+    pub l2_block_range: core::ops::Range<u32>,
```
xgreenx, line 22: I think we just want to know what is the last L2 height is and what the L1 height used to commit the last blob.

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/service.rs
```diff
@@ -0,0 +1,118 @@
...
+            source,
+        }
+    }
+}
+
+/// This trait is implemented by the sources to obtain the
+/// da block costs in a way they see fit
+#[async_trait::async_trait]
+pub trait DaBlockCostsSource: Send + Sync {
+    async fn get(&mut self) -> Result<DaBlockCosts>;
```
xgreenx, line 51:
```suggestion
    async fn request_da_cost(&mut self) -> Result<DaBlockCosts>;
```

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
```diff
@@ -1,16 +1,125 @@
...
+    #[tokio::test]
+    async fn run__when_da_block_cost_source_errors_shared_value_is_not_updated() {
+        // given
+        let service =
+            DaBlockCostsService::new(ErroringSource, Some(Duration::from_millis(1)));
+        let mut shared_state = service.shared_data();
+        let service = ServiceRunner::new(service);
+
+        // when
+        service.start().unwrap();
```
xgreenx, line 117:
```suggestion
        service.start_and_await().await.unwrap();
```

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
```diff
@@ -1,16 +1,125 @@
...
+        sleep(Duration::from_millis(10)).await;
+        service.stop();
```
xgreenx, line 119:
```suggestion
        service.stop_and_await().await;
```

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
xgreenx: it would be nice if you waited in other places as well

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
xgreenx: it would be nice if you waited in other places as well

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
```diff
@@ -1,16 +1,125 @@
...
+                "Failed to lock shared gas price state: {:?}",
+                err
+            ))
+        })?;
+
+        let da_block_costs = da_block_costs_guard.clone();
+
+        // now mark it as consumed because we don't want to serve the same data
+        // multiple times
+        *da_block_costs_guard = None;
```
xgreenx, line 36: I would suggest implementing the DA source service as a stream that guarantees the uniqueness of events. It is the responsibility of the gas provider to decide what to do with these unique events.

xgreenx (changes requested):

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -335,6 +345,7 @@ pub fn init_sub_services(
     #[allow(unused_mut)]
     // `FuelService` starts and shutdowns all sub-services in the `services` order
     let mut services: SubServices = vec![
+        Box::new(da_block_costs_service),
```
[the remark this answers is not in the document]
rymnc: I discussed this with @MitchTurner but it was his opinion that it should run as a sub-service (it was my opinion that the gas price service takes this service as a dependency and should be defined internally)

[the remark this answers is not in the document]
xgreenx: If we define the new service inside of the same crate and it is only used by this crate, I don't see the point in exposing this service outside. We could keep it as an internal detail of the implementation.

@MitchTurner  what are your thoughts?

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
```diff
@@ -1,16 +1,125 @@
...
+pub use dummy_costs::DummyDaBlockCosts;
+pub use service::*;
+
+pub const POLLING_INTERVAL_MS: u64 = 10_000;
+
+pub type DaBlockCostsProvider = Arc<Mutex<Option<DaBlockCosts>>>;
+
+impl GetDaBlockCosts for DaBlockCostsProvider {
+    fn get(&mut self) -> GasPriceUpdaterResult<Option<DaBlockCosts>> {
+        let mut da_block_costs_guard = self.try_lock().map_err(|err| {
```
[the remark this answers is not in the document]
rymnc: we don't want it to be an expensive operation in case the lock is held by the function that sets data into it, and we always want to be able to return instantly. as per discussion with @MitchTurner

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_costs.rs
```diff
@@ -0,0 +1,60 @@
...
+    pub blob_size_bytes: u32,
+    pub blob_cost: u32,
```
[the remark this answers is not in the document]
rymnc: this will change eventually when we use bigint's so I will make this change to `u128` for now

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_costs.rs
```diff
@@ -0,0 +1,60 @@
...
+/// which receives data from the block committer (only http api for now)
+pub struct BlockCommitterDaBlockCosts {
+    client: reqwest::Client,
+    url: Url,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone, Default, PartialEq)]
+struct RawDaBlockCosts {
+    pub l2_block_range: core::ops::Range<u32>,
+    pub blob_size_bytes: u32,
```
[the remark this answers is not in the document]
rymnc: the blob size will always be < 128kb

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_costs.rs
rymnc: the block committer will return unordered data to us. they will commit a set of L2 blocks at once to the L1, and return the range, the cost and the size of the blob :)

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/service.rs
```diff
@@ -0,0 +1,118 @@
...
+        let continue_running;
+
+        tokio::select! {
+            biased;
+            _ = state_watcher.while_started() => {
+                continue_running = false;
+            }
+            _ = self.poll_interval.tick() => {
+                let da_block_costs = self.source.get().await?;
+                let mut da_block_costs_guard = self.block_cost_provider.try_lock().map_err(|err| {
```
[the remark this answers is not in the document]
rymnc: this can be a lock, good idea

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_costs.rs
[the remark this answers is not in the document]
xgreenx: The block committer will return cumulative value, not the blob size of the last committed blob.

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_costs.rs
xgreenx: It looks like it would be simple if they just returned the last DA block height and the last committed L2 block height. Since they commit block only in order

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_costs.rs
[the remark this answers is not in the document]
rymnc: don't think so, we discussed with @MujkicA and I don't think it will return cumulative value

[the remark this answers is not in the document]
xgreenx: I don't see how they can return anything else with the current API. Either they return cumulative values, or the committer provides another API like `get_cost_at_l1_height(block_height: DaBlockHeight)` or `get_cost_at_l2_height(block_height: BlockHeight)`(or instead of the height it uses range).


Because the committer can't know for what range it should return data based on the current API.

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -204,13 +208,19 @@ pub fn init_sub_services(
     let settings = consensus_parameters_provider.clone();
     let block_stream = importer_adapter.events_shared_result();
 
+    // todo(#2139): replace dummy source with the block committer source
+    let da_block_costs_service = DaBlockCostsService::new(DummyDaBlockCosts, None);
+    let da_block_costs = da_block_costs_service.shared_data();
+    let da_block_costs_service = ServiceRunner::new(da_block_costs_service);
```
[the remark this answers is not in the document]
rymnc: addressed in 3cd2dce945d59563a7483ace7ae5368f9bfc9e22

crates/services/gas_price_service/src/fuel_gas_price_updater.rs
```diff
@@ -21,14 +21,18 @@ pub use fuel_gas_price_algorithm::{
 mod tests;
 
 pub mod fuel_core_storage_adapter;
+
 pub mod fuel_da_source_adapter;
 
-pub struct FuelGasPriceUpdater<L2, Metadata, DaSource> {
+pub use fuel_core_storage_adapter::*;
+pub use fuel_da_source_adapter::*;
```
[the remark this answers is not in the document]
rymnc: addressed in 1ab0d9a7a

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
```diff
@@ -1,16 +1,125 @@
...
+pub mod service;
 
-impl DaCommitSource for FuelDaSource {
-    fn get_da_commit_details(
-        &mut self,
-    ) -> GasPriceUpdaterResult<Option<DaCommitDetails>> {
-        todo!() // TODO(#2139): pending research on how to get the data from the block committer
+pub use block_committer_costs::BlockCommitterDaBlockCosts;
+pub use dummy_costs::DummyDaBlockCosts;
+pub use service::*;
```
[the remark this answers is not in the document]
rymnc: addressed in 1ab0d9a7a

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
```diff
@@ -1,16 +1,125 @@
...
+
+pub const POLLING_INTERVAL_MS: u64 = 10_000;
+
+pub type DaBlockCostsProvider = Arc<Mutex<Option<DaBlockCosts>>>;
```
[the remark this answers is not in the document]
rymnc: addressed in 18249b099

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
```diff
@@ -1,16 +1,125 @@
...
+impl GetDaBlockCosts for DaBlockCostsProvider {
+    fn get(&mut self) -> GasPriceUpdaterResult<Option<DaBlockCosts>> {
+        let mut da_block_costs_guard = self.try_lock().map_err(|err| {
+            CouldNotFetchDARecord(anyhow!(
+                "Failed to lock shared gas price state: {:?}",
+                err
+            ))
+        })?;
+
+        let da_block_costs = da_block_costs_guard.clone();
```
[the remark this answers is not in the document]
rymnc: addressed in 3e777fe89

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_costs.rs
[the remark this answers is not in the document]
rymnc: addressed in ac72e4bca

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/service.rs
rymnc: addressed in fa96bede0

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
rymnc: addressed in 751324ddf

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_costs.rs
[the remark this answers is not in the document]
rymnc: the block committer will create a new api for us :)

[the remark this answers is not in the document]
xgreenx: If they plan to create a new API, then we need to agree on it in advance and implement logic based on this agreement=)

But another question: do we need a new API? It seems for us is enough to receive cumulative values with latest L1 and L2 block heights and it should fulfill our requirements

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_costs.rs
xgreenx: If the committer does not commit in order I don't understand how fraud proving will work and how L1 contract can know which header belongs to which block.

crates/fuel-core/src/service/sub_services.rs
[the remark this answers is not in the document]
MitchTurner: My opposition was with taking a dependency on `RunnableService` inside the `GasPriceService` domain. 

I'm fine with having the adapter manage its own service, OR having a `RunnableService` spun up in integration. But `RunnableService` is an integration level abstraction and I'd prefer it not get used everywhere/anywhere we want to run something in a separate thread.

I honestly don't have problem with it getting spun up on startup just like the `GasPriceService` itself--we already know concrete types when we run `init_sub_services`. I initially suggested we keep it internal to the adapter and ditch the `RunnableService` impl.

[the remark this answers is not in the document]
xgreenx: I like the idea of using `RunnableService` to fetch data from an external source. It looks like what service should be responsible for, since it has some internal logic of how often to fetch data and how to provide this data to its consumers. And this service should support graceful shutdown as well.

I'm okay with having this service as an independent and non-internal sub-service, but then it should live outside of the gas provider crate(currently, it lives there).

I also like the idea of having this service as an internal sub-service of the gas price provider since its usage only belongs to the gas provider. I don't see any problem with having sub-services in the service. The main idea is that each service depends on `fuel-core-services` and is already aware of `RunnableService`. 

At the end of the day, `fuel-core` is also a service with many sub-services =)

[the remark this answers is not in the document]
MitchTurner: If you guys both want to use RunnableService inside the other service, then I'll defer to your judgment.

I think the ServiceRunner library is a bit of a Swiss Army knife and that's a design smell for me. But on the other hand, I like having some reusable library for managing services ran in separate threads. I just wish it wasn't trying to solve so many problems at once and the more places we use it the worse that's going to get.

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_costs.rs
MujkicA: We've also solidified some design decisions that I might've not communicated since my last discussion with @MitchTurner and @rymnc .

The workflow for posting L2 data will roughly look like this:
we accumulate some amount of blocks from L2 into a bundle -> we apply compression over this bundle -> the resulting compressed data is then fragmented to fit into blob transactions (+ some metadata) and sent to L1.

The final order of those transactions shouldn't matter for the following reasons
- the data is compressed and the order can be reconstructed from the metadata
- the endpoint for the gas algorithm will only provide data for a full range of posted blocks

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_costs.rs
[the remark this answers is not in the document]
MujkicA: Since there are two layers of abstraction between a block and what is actually posted inside blobs, we can only provide a cumulative value over some block range. We can tailor the api to your needs given that constraint.

See other [comment](https://github.com/FuelLabs/fuel-core/pull/2163#discussion_r1750668108)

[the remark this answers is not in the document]
xgreenx: From my perspective, it should be simpler just to use cumulative value for all time=) But it is up to @MitchTurner  and @rymnc to use a small range or the whole range. Based on the definition of `RawDaBlockCosts,` I see that we don't have all the information we need right now. We also need to know the DA block height of latest committed blob

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_costs.rs
xgreenx: > the data is compressed and the order can be reconstructed from the metadata

What if the first half of the block at height X in another blob and you haven't submitted it yet? How can we reconstruct block without the first half?

> we apply compression over this bundle

Do you mean GZIP? Because if you mean DA compression that we work on, I think we will just provide a new endpoint for you to fetch compressed blocks=)

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/service.rs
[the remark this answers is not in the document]
rymnc: addressed in 7a967d95e

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter.rs
rymnc: okay, agreed ~ I'll use a channel for this purpose. I don't think there is much value in adding the stream to the sources (block committer, dummy etc) because they are relatively simple in logic, and don't need to know that we are polling the endpoints at regular intervals.

rymnc: uniqueness handled in 8ff00388e

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_da_source_adapter/block_committer_costs.rs
MujkicA: > What if the first half of the block at height X in another blob and you haven't submitted it yet? How can we reconstruct block without the first half?

The idea is that you always reconstruct in bundles. The last fragment of the bundle, submitted in the last tx for that bundle can contain metadata like the number of fragments, compression metadata, the block height at which we sent the first fragment etc.
For reconstruction you would need to fetch all the fragments belonging to the bundle and decompress

> Do you mean GZIP? Because if you mean DA compression that we work on, I think we will just provide a new endpoint for you to fetch compressed blocks=)

I wasn't aware that there will be an endpoint for compressed blocks. Can we still expect to make gains if we also do something like GZIP on our end?

crates/fuel-core/src/service/sub_services.rs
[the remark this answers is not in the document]
rymnc: integrated in https://github.com/FuelLabs/fuel-core/pull/2163/commits/2f803f9296a96a303f6e5808e7ccc170d55c277b :)

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -204,13 +208,19 @@ pub fn init_sub_services(
...
         config.clone(),
         genesis_block_height,
         settings,
         block_stream,
         database.gas_price().clone(),
         database.on_chain().clone(),
+        da_block_costs.clone(),
     )?;
     let next_algo = gas_price_init.shared_data();
     let gas_price_service = ServiceRunner::new(gas_price_init);
```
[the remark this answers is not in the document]
rymnc: https://github.com/FuelLabs/fuel-core/issues/2184

rymnc:
please note that all changes to the block committer api will be handled in a follow up PR :)

crates/services/gas_price_service/src/fuel_gas_price_updater/da_source_adapter.rs
```diff
@@ -0,0 +1,116 @@
...
+        let mut shared_state = service.shared.clone();
+
+        // when
+        service.start_and_await().await.unwrap();
+        sleep(Duration::from_millis(10)).await;
+        service.stop_and_await().await.unwrap();
+
+        // then
+        let da_block_costs_opt = shared_state.get().unwrap();
+        assert!(da_block_costs_opt.is_some());
```
MitchTurner, line 81: WE should modify `DummyDaBlockCosts` to return a specified value (no default) and then we can check that the values match here, rather than checking that is `Some`.

You can have it return a `Result` to for more flexibility.

```rs
pub struct DummyDaBlockCosts {
    value: anyhow::Result<DaBlockCosts>,
}

#[async_trait::async_trait]
impl DaBlockCostsSource for DummyDaBlockCosts {
    async fn request_da_block_cost(&mut self) -> DaBlockCostsResult<DaBlockCosts> {
        self.value.clone()
    }
}
```
Then you can get rid of `ErroringSource`.

crates/services/gas_price_service/src/fuel_gas_price_updater/da_source_adapter/service.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+impl<Source> DaBlockCostsService<Source>
+where
+    Source: DaBlockCostsSource,
+{
+    pub fn new(source: Source, poll_interval: Option<Duration>) -> Self {
+        #[allow(clippy::arithmetic_side_effects)]
+        let (sender, receiver) = tokio::sync::mpsc::channel(
+            core::mem::size_of::<DaBlockCosts>() * CACHE_COUNT,
```
MitchTurner, line 50: I think the buffer is count, not memory size. Right? And maybe this `const` could be named `CHANNEL_BUFFER_SIZE`. It doesn't seem to have anything to do with the `self.cache`

crates/services/gas_price_service/src/lib.rs
```diff
@@ -25,6 +31,8 @@ pub struct GasPriceService<A, U> {
     next_block_algorithm: SharedGasPriceAlgo<A>,
     /// The code that is run to update your specific algorithm
     update_algorithm: U,
+    // The da block cost sub service
+    da_block_costs_task_handle: ServiceRunner<DaBlockCostsService<DummyDaBlockCosts>>,
```
MitchTurner, line 35: I think this handle can just be on the adapter. 

```
#[derive(Clone)]
pub struct DaBlockCostsProvider {
    service_handle: ServiceRunner<DaBlockCostsService<DummyDaBlockCosts>>,
    receiver: Arc<Mutex<mpsc::Receiver<DaBlockCosts>>>,
}
```
Why not?

CHANGELOG.md
```diff
@@ -8,7 +8,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - [2135](https://github.com/FuelLabs/fuel-core/pull/2135): Added metrics logging for number of blocks served over the p2p req/res protocol.
+<<<<<<< feat/da-source-implementation
+- [2163](https://github.com/FuelLabs/fuel-core/pull/2163): Added runnable task for fetching block committer data.
+=======
 - [2151](https://github.com/FuelLabs/fuel-core/pull/2151): Added limitations on gas used during dry_run in API.
+>>>>>>> master
```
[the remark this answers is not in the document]
rymnc: bzzz, it was supposed to remove it, sec

crates/services/gas_price_service/src/fuel_gas_price_updater/da_source_adapter/service.rs
rymnc: addressed in 0240ba73d

crates/services/gas_price_service/src/fuel_gas_price_updater/da_source_adapter.rs
rymnc: addressed in fee8fd145

CHANGELOG.md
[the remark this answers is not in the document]
rymnc: addressed in cbab3559f

crates/services/gas_price_service/src/lib.rs
rymnc: The primary reason not to place the `service_handle` directly on the adapter is that it violates separation of concerns and single responsibility principle. The adapter’s role should be strictly focused on fetching and providing the da block costs. wdyt?

crates/services/gas_price_service/src/fuel_gas_price_updater/da_source_adapter/service.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+#[async_trait::async_trait]
+impl<Source> RunnableService for DaBlockCostsService<Source>
+where
+    Source: DaBlockCostsSource,
+{
+    const NAME: &'static str = "DaBlockCostsService";
+
+    type SharedData = DaBlockCostsProvider;
```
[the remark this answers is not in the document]
rymnc: why not? we have to create the channel internally and if we have no `SharedData` we would have to create the channel externally

crates/services/gas_price_service/src/lib.rs
MitchTurner: I don't think it's a violation of either of those.

It would be violation of SoC if we had the client also making calls for the L2block info. Those things can vary independently, and it would be wrong to couple them. The concern here is getting info from the committer. Our _implementation_ is having a service that runs in another thread that eagerly buffers the data for us. If we were doing it lazily--without the service thread--everything would be in the adapter as well, and I assume you wouldn't have any issue with that.

The responsibility here is getting the costs from DA. It's not a misdirection at all for something called `DaBlockCostsProvider` to have a handle to task that is providing us with the latest `DaBlockCosts`. 

I definitely think the `GasPriceService` is the _wrong_ place for it. It has no concept of L2 or DA at that level. That's the job of our specific `update_algorithm` (not the best name, mb)

The only question I have about if this is the "right" pace to hold the handle is how we want to handle errors from the service.

crates/services/gas_price_service/src/fuel_gas_price_updater/da_source_adapter/service.rs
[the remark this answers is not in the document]
MitchTurner: Oh. It's `Clone`. 

Okay. You can have it return the `Arc<Mutex<Receiver>>` and then the `DaBlockCostsProvider` can be constructed from that :)

MitchTurner (approved):
LGTM!

MitchTurner (approved):

crates/services/gas_price_service/src/fuel_gas_price_updater/da_source_adapter/service.rs
```diff
@@ -0,0 +1,130 @@
...
+/// This struct houses the shared_state, polling interval
+/// and a source, which does the actual fetching of the data
+pub struct DaBlockCostsService<Source>
+where
+    Source: DaBlockCostsSource,
+{
+    poll_interval: Interval,
+    source: Source,
+    sender: Sender<DaBlockCosts>,
+    cache: HashSet<DaBlockCosts>,
```
xgreenx, line 34: It would be nice to clean it up periodically:) Or even not to use a cache. We can just track the last range/da height of the bundle

xgreenx (approved):
Based on our discussions we need to change a lot of stuff, but we can do that in follow up PRs

Let's merge it and start work on next changes:)

crates/services/gas_price_service/src/fuel_gas_price_updater/da_source_adapter/service.rs
rymnc: this will be omitted anyway in the next pr when we restructure the gas price service

Issue #2164, opened 2024-09-05
Bug: It appears the V1 gas price algorithm gives up after a certain point


MitchTurner:
<img width="993" alt="image" src="https://github.com/user-attachments/assets/01c5bd1d-b4d9-47fb-9944-35ce93ea7d7c">

May be related to: https://github.com/FuelLabs/fuel-core/issues/2147

Issue #2147, opened 2024-08-30
V1 Total Cost and Total Reward are subject to overflow


MitchTurner:
We either need to use bigger integers (e.g. `u128`) or trim the totals down occasionally

Nithin-Varma:
yes, use bigger integers and always use a library for better checks and asserts, similar to openzeppelin safemath.sol for solidity contracts.
if possible, please provide me the exact file, i want to look into it once.

PR #2173, opened 2024-09-06
bug: fix algorithm overflow issues

MitchTurner:
## Linked Issues/PRs
Closes https://github.com/FuelLabs/fuel-core/issues/2164
Closes https://github.com/FuelLabs/fuel-core/issues/2147

## Description
The main change with this code is "normalizaing" the costs and rewards instead of keeping a total over all time. i.e. every time we receive a DA block, we see if the reward is greater than the costs, or vice versa. If the reward is higher, we set the reward to the difference and set the the last known cost to `0` and adjust the projected cost accordingly.

In addition, we were using a random set of types for the algorithm and also used casts in many places. This PR should fix a lot of those problems.

Bonus: This fix prompted me to run the optimization again. Since the set is much bigger now, I decided to enable running the simulation in parallel tasks to speed up the code.

## Checklist

### Before requesting review


crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
```diff
@@ -70,29 +63,24 @@ impl Simulator {
         da_p_component: i64,
         da_d_component: i64,
     ) -> AlgorithmUpdaterV1 {
-        // Scales the gas price internally, value is arbitrary
         let gas_price_factor = 100;
         let updater = AlgorithmUpdaterV1 {
             min_exec_gas_price: 10,
             min_da_gas_price: 10,
-            // Change to adjust where the gas price starts on block 0
```
MitchTurner, line 78: Accidentally removed these on merge.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -82,36 +82,51 @@ impl AlgorithmV1 {
...
     }
 
-    fn assemble_price(&self, change: i64) -> u64 {
-        let last_da_gas_price = self.last_da_price as i128;
-        let maybe_new_da_gas_price = last_da_gas_price
-            .saturating_add(change as i128)
-            .try_into()
-            .unwrap_or(self.min_da_gas_price);
+    fn assemble_price(&self, change: i128) -> u64 {
+        // let last_da_gas_price = self.last_da_price as i128;
```
MitchTurner, line 114: Remove

crates/fuel-gas-price-algorithm/src/v1/tests/update_da_record_data_tests.rs
```diff
@@ -283,3 +284,54 @@ fn update__da_block_updates_projected_total_cost_with_known_and_guesses_on_top()
     let expected = new_known_total_cost + guessed_part;
     assert_eq!(actual, expected as u128);
 }
+
+#[test]
+fn update_da_record_data__when_reward_is_greater_than_cost_will_zero_reward_and_subtract_from_cost(
```
MitchTurner, line 289: Need to check the projected cost as well, specifically when it is ahead of DA still after the update.

crates/fuel-gas-price-algorithm/src/v1/tests/update_da_record_data_tests.rs
```diff
@@ -283,3 +290,126 @@ fn update__da_block_updates_projected_total_cost_with_known_and_guesses_on_top()
...
+    fn update_da_record_data__when_cost_is_greater_than_reward_will_zero_reward_and_subtract_from_cost(
+        (cost, reward, blocks) in cost_greater_than_reward_with_da_blocks()
+    ) {
+        _update_da_record_data__when_cost_is_greater_than_reward_will_zero_reward_and_subtract_from_cost(
+            cost,
+            reward,
+            blocks
+        )
+    }
+}
```
MitchTurner, line 384:
```suggestion
}

```

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -60,17 +60,17 @@ pub struct AlgorithmV1 {
...
     latest_da_cost_per_byte: u128,
     /// The cumulative reward from the DA portion of the gas price
-    total_rewards: u64,
+    total_rewards: u128,
     /// The cumulative cost of recording L2 blocks on the DA chain as of the last recorded block
     total_costs: u128,
     /// The P component of the PID control for the DA gas price
     da_p_factor: i64,
     /// The D component of the PID control for the DA gas price
     da_d_factor: i64,
```
rymnc, line 69: is my understanding correct that these values are `i64`'s because we need them to be for float calculation?

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -82,36 +82,44 @@ impl AlgorithmV1 {
...
             .last_da_price
-            .saturating_mul(self.max_change_percent as u64)
-            .saturating_div(100) as i64;
+            .saturating_mul(upcast_percent)
+            .saturating_div(100)
+            .into();
         let sign = pd_change.signum();
         let signless_da_change = min(max_change, pd_change.abs());
         sign.saturating_mul(signless_da_change)
     }
```
rymnc, line 111:
```suggestion
        let clamped_change = pd_change.abs().min(max_change);
        pd_change.signum().saturating_mul(clamped_change)
    }
```

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: I'm not sure what you mean by "float calculation". We don't have any floats in this code.

These actually probably don't need to be `i` and could be `u`. When I added them originally, I was open to the idea of using negative values. Just haven't reevaluated.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -82,36 +82,44 @@ impl AlgorithmV1 {
...
+            let new_price_oversized = upcast_last_da_price.saturating_add(change);
+            new_price_oversized.try_into().unwrap_or(u64::MAX)
+        } else {
+            let upcast_last_da_price: i128 = self.last_da_price.into();
+            let new_price_oversized = upcast_last_da_price.saturating_add(change);
+            new_price_oversized.try_into().unwrap_or(0)
+        };
         let new_da_gas_price = max(self.min_da_gas_price, maybe_new_da_gas_price);
         self.new_exec_price.saturating_add(new_da_gas_price)
     }
```
rymnc, line 125:
```suggestion
fn assemble_price(&self, change: i128) -> u64 {
    let upcast_last_da_price: i128 = self.last_da_price.into();
    let new_price_oversized = upcast_last_da_price.saturating_add(change);

    let maybe_new_da_gas_price: u64 = new_price_oversized
        .try_into()
        .unwrap_or(if change.is_positive() { u64::MAX } else { 0 });

    let new_da_gas_price = max(self.min_da_gas_price, maybe_new_da_gas_price);
    self.new_exec_price.saturating_add(new_da_gas_price)
}
```

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -336,12 +347,34 @@ impl AlgorithmUpdaterV1 {
...
+            self.latest_known_total_da_cost_excess = 0;
+        } else {
+            let excess = total_costs.saturating_sub(total_rewards);
+            let projected_cost_excess =
+                self.projected_total_da_cost.saturating_sub(total_rewards);
+            self.projected_total_da_cost = projected_cost_excess;
+            self.total_da_rewards_excess = 0;
+            self.latest_known_total_da_cost_excess = excess;
+        }
+    }
```
rymnc, line 379:
```suggestion
  fn normalize_rewards_and_costs(&mut self) {
      let (excess, projected_cost_excess) = if self.total_da_rewards_excess > self.latest_known_total_da_cost_excess {
          (
              self.total_da_rewards_excess.saturating_sub(self.latest_known_total_da_cost_excess),
              self.projected_total_da_cost.saturating_sub(self.latest_known_total_da_cost_excess),
          )
      } else {
          (
              self.latest_known_total_da_cost_excess.saturating_sub(self.total_da_rewards_excess),
              self.projected_total_da_cost.saturating_sub(self.total_da_rewards_excess),
          )
      };
  
      self.projected_total_da_cost = projected_cost_excess;
      if self.total_da_rewards_excess > self.latest_known_total_da_cost_excess {
          self.total_da_rewards_excess = excess;
          self.latest_known_total_da_cost_excess = 0;
      } else {
          self.total_da_rewards_excess = 0;
          self.latest_known_total_da_cost_excess = excess;
      }
  }
```
extracting common logic out of the branches

rymnc: can we also use `checked_sub` here instead?

rymnc (commented):
LGTM in general, left some suggestions/questions

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -153,11 +161,11 @@ pub struct AlgorithmUpdaterV1 {
     /// The maximum percentage that the DA portion of the gas price can change in a single block
     pub max_da_gas_price_change_percent: u8,
```
Dentosal, line 159: nit: I feel like introducing a newtype to enforce the percentage invariant (in range `0..=100`) might be a good idea. Up to you.

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: Can you be more specific about why we'd prefer `checked_sub`?

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: I'm open to it! I'd just have to introduce fallibility into the constructor.

Also. In theory, this could be set to something larger than 100%. So even `u8` is an arbitrary ceiling to how high it could be set to. I'm realizing that's why `exec_gas_price_change_percent` wasn't `u8` before. Both could be higher in theory.

The `l2_block_fullness_threshold_percent` on the otherhand would be a better candidate for a newtype since there is no such thing as a 110% full block hopefully.

Dentosal: The arbitrary u8 ceiling feels weird, if there's ever any reason to set it to 200%, surely there's also reasons to set it to 300% as well? In any case, it wasn't clear to me from the doc comment, so maybe explicitly writing it there would be a good first step, and the newtype can be introduced later when the implementation has settled down.

MitchTurner: Changed both to `u16` and added explanation.

Dentosal (dismissed):
LGTM; although I was mostly looking at the changeset and not the whole picture

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: Added a newtype for the threshold percentage. I called it `ClampedPercent` to communicate that it's infallible and just clamps the value.

xgreenx (commented):
I hope @rymnc tried to dive deeper in the formulas and checked them=D

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation/da_cost_per_byte.rs
```diff
@@ -25,6 +25,7 @@ pub fn get_da_cost_per_byte_from_source(
     }
 }
 
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
MitchTurner: It's used in CSV reader.

crates/fuel-gas-price-algorithm/src/v1.rs
rymnc: ah okay. as far as i remember there were a bunch of floats used so i might be wrong

crates/fuel-gas-price-algorithm/src/v1.rs
rymnc: well, we should have an error if the value becomes negative, right?

MitchTurner: In theory it can't, so maybe it would be most correct to do an `expect` here with a message explaining why it'll never happen.

MitchTurner: I'd rather the function be infallible. Hopefully we can prove to ourselves that the algorithm can't get into that state with tests rather than adding `Result` here.

rymnc: okay, sounds good

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
```diff
@@ -75,7 +75,7 @@ impl Simulator {
             gas_price_factor: NonZeroU64::new(gas_price_factor).unwrap(),
             l2_block_height: 0,
             // Choose the ideal fullness percentage for the L2 block
-            l2_block_fullness_threshold_percent: 50,
+            l2_block_fullness_threshold_percent: 50u8.into(),
```
rymnc, line 78: why not `50u64`?

MitchTurner:
```rs
impl From<u8> for ClampedPercentage {
    fn from(value: u8) -> Self {
        Self::new(value)
    }
}
```
I only implemented `into` for `u8`. I considered doing more, but YAGNI.

rymnc (approved):

xgreenx (approved):

PR #2179, opened 2024-09-09
chore: Add initial CODEOWNERS file

### LINKED ISSUE #1982: Add code owner and enhance approval process

netrome:
Simple CODEOWNERS file setting @xgreenx, @Dentosal and @MitchTurner as code owners for this repo. This initially will add them to any PR within the repo, and also allows us to require one approval from either of them to pass a PR.

As the team matures, we should replace the root CODEOWNERS with just the `FuelLabs/Client` team.

Closes #1982

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# Default code owners for fuel-core
+* @xgreenx @Dentosal @MitchTurner


.github/CODEOWNERS
```diff
@@ -0,0 +1,2 @@
+# Default code owners for fuel-core
+* @xgreenx @Dentosal @MitchTurner
```
rymnc, line 2: perhaps the older team members want more granularity on which parts they've written majorly ?

Dentosal (approved):

.github/CODEOWNERS
Dentosal: I think this is fine as-is, no reason to complicate this for now. If I want someone else to do review for a specific part of code, I'll just ask them to review instead.

netrome: Yeah I figured let's start simple and extend it when we want to be more specific. The long term vision imo is that we have the client team as default code owners for the repo, and more well-defined sub-teams for any specific part of the code we want to have more granular control over.

MitchTurner (approved):

Issue #2114, opened 2024-08-20
Add more graceful handling of the huge number of transactions during production


xgreenx:
The total number of allowed transactions per block is `65535` due to the `TxPointer` limitations.

Our current network configuration handles it.

But since we plan to have many custom rollups that use `fuel-core`. It would be nice to add more graceful handling of a huge number of transitions.

We have several places where we need to handle it:
- Executor fetches transactions from the `TxPool` based on the remaining gas. We need to also take into account the remaining number of transaction slots. Note: FTI transactions also occupy slots.
- When the block producer increases DA height, only gas used by FTI is taken into account. We also need to be sure that all transactions can fit into one block.
- Maybe increase the minimal gas on the L1 side to be more than `1` https://github.com/FuelLabs/fuel-bridge/pull/215/files#diff-cf9b39d4822d2f60928e44960317ad86812ccf644704cfdd364279ba02e673acR10. Maybe `10_000` is not bad as a default value.

acerone85:
I'll start working on this one, will probably add more comments with questions to get context

acerone85:
What is the expected way this should be handled? Should we have a consensus parameter that mentions the maximum number of transactions per block (in which case this is related to what @rafal-ch is doing with [#2133](https://github.com/FuelLabs/fuel-core/issues/2133))? Or is there any other approach we can take?

acerone85:
What is a FTI transaction?

xgreenx:
FTI - Force transaction inclusion. You can submit the transaction to L1 instead of L2(maybe someone censorships your transaction), and it will be included when the block producer increases the DA block height. FTI - is a regular transaction that just comes from L1, instead of TxPool.

acerone85:
Reopening, Still need to update the maximal gas price

PR #2182, opened 2024-09-09
TransactionSource: specify maximum number of transactions to be fetched

acerone85:
## Linked Issues/PRs

Part of #2114 

## Description

- Add a new argument `transactions_limit: u16` to the function `TxSource::next()`. 
- Changes the logic of the block production to never exceed `u16::MAX` transactions (included FTI transactions and mint transactions) when selecting L2 transacitons. The number of transactions can still exceed `u16::MAX` when selecting transactions from L1. This is addressed in #2189.
- Changes the implementation of all TxSource adapters to fetch a number of transactions below or equal to the `transaction_limit`. 
- Introduces a new WASM host function to peek the next transactions size for a number of transactions below a specified limit.

Breaks forward compatibility of the genesis transition function. See https://github.com/FuelLabs/fuel-core/blob/8eeae5d6730d34ae06390b7797478bb98fd07987/version-compatibility/forkless-upgrade/README.md

## TODO:
- [x] Check that requirements are met
- [x] Implementation for the WasmTxSource is missing (only the argument has been added) 
- [ ] Integration Tests

## Checklist
- [ x] Breaking changes are clearly marked as such in the PR description and changelog

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2158](https://github.com/FuelLabs/fuel-core/pull/2158): Log the public address of the signing key, if it is specified
 - [2188](https://github.com/FuelLabs/fuel-core/pull/2188): Upgraded the `fuel-vm` to `0.57.0`. More information in the [release](https://github.com/FuelLabs/fuel-vm/releases/tag/v0.57.0).
 
-
 ## [Version 0.35.0]
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - [2131](https://github.com/FuelLabs/fuel-core/pull/2131): Add flow in TxPool in order to ask to newly connected peers to share their transaction pool
+- [2182](https://github.com/FuelLabs/fuel-core/pull/2151): Limit number of transactions that can be fetched via TxSource::next
 
 ### Changed
 
@@ -37,6 +38,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2142](https://github.com/FuelLabs/fuel-core/pull/2142): Added benchmarks for varied forms of db lookups to assist in optimizations.
 - [2158](https://github.com/FuelLabs/fuel-core/pull/2158): Log the public address of the signing key, if it is specified
 - [2188](https://github.com/FuelLabs/fuel-core/pull/2188): Upgraded the `fuel-vm` to `0.57.0`. More information in the [release](https://github.com/FuelLabs/fuel-vm/releases/tag/v0.57.0).
+
 
 ## [Version 0.35.0]
 
--- a/crates/services/executor/src/executor.rs
+++ b/crates/services/executor/src/executor.rs
@@ -589,6 +589,7 @@ where
         let mut regular_tx_iter = l2_tx_source
             .next(remaining_gas_limit, remaining_tx_count, remaining_size)
             .into_iter()
+            .take(remaining_tx_count as usize)
             .peekable();
         while regular_tx_iter.peek().is_some() {
             for transaction in regular_tx_iter {
@@ -619,6 +620,7 @@ where
             regular_tx_iter = l2_tx_source
                 .next(remaining_gas_limit, remaining_tx_count, remaining_size)
                 .into_iter()
+                .take(remaining_tx_count as usize)
                 .peekable();
         }
 
--- a/crates/services/executor/src/executor.rs
+++ b/crates/services/executor/src/executor.rs
@@ -162,6 +162,8 @@ use alloc::{
     vec::Vec,
 };
 
+pub const MAX_TX_COUNT: u16 = u16::MAX.saturating_sub(1);
+
 pub struct OnceTransactionsSource {
     transactions: ParkingMutex<Vec<MaybeCheckedTransaction>>,
 }
@@ -580,8 +582,7 @@ where
         // When processing l2 transactions, we must take into account transactions from the l1
         // that have been included in the block already (stored in `data.tx_count`), as well
         // as the final mint transaction.
-        let max_tx_count = u16::MAX.saturating_sub(1);
-        let mut remaining_tx_count = max_tx_count.saturating_sub(data.tx_count);
+        let mut remaining_tx_count = MAX_TX_COUNT.saturating_sub(data.tx_count);
 
         let mut regular_tx_iter = l2_tx_source
             .next(remaining_gas_limit, remaining_tx_count, remaining_size)
@@ -610,7 +611,7 @@ where
                     }
                 }
                 remaining_gas_limit = block_gas_limit.saturating_sub(data.used_gas);
-                remaining_tx_count = max_tx_count.saturating_sub(data.tx_count);
+                remaining_tx_count = MAX_TX_COUNT.saturating_sub(data.tx_count);
             }
 
             regular_tx_iter = l2_tx_source
--- a/crates/services/executor/src/executor.rs
+++ b/crates/services/executor/src/executor.rs
@@ -162,6 +162,8 @@ use alloc::{
     vec::Vec,
 };
 
+/// The maximum amount of transactions that can be included in a block,
+/// excluding the mint transaction.
 pub const MAX_TX_COUNT: u16 = u16::MAX.saturating_sub(1);
 
 pub struct OnceTransactionsSource {
@@ -578,7 +580,7 @@ where
         // TODO: Handle `remaining_size` https://github.com/FuelLabs/fuel-core/issues/2133
         let remaining_size = u32::MAX;
 
-        // We allow at most u64::MAX transactions in a block, including the mint transaction.
+        // We allow at most u16::MAX transactions in a block, including the mint transaction.
         // When processing l2 transactions, we must take into account transactions from the l1
         // that have been included in the block already (stored in `data.tx_count`), as well
         // as the final mint transaction.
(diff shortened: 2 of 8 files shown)


crates/services/executor/src/ports.rs
```diff
@@ -55,8 +55,13 @@ impl MaybeCheckedTransaction {
...
 pub trait TransactionsSource {
-    /// Returns the next batch of transactions to satisfy the `gas_limit`.
-    fn next(&self, gas_limit: u64) -> Vec<MaybeCheckedTransaction>;
+    /// Returns the next batch of transactions to satisfy the `gas_limit` and the
+    /// without selecting more than `max_transactions` transactions. Because
+    /// `max_transactions` has type `u16`, no more than `u16::MAX` transactions
+    /// will be included in a block, which enforce the property that all transaction
+    /// outputs will be indexable in the future.
+    fn next(&self, gas_limit: u64, max_transactions: u16)
+        -> Vec<MaybeCheckedTransaction>;
```
acerone85, line 64: IIUC Here we also want to take the number of slots to reserve for FTI transactions?

crates/services/txpool/src/txpool.rs
```diff
@@ -126,12 +126,22 @@ impl<ViewProvider, WasmChecker> TxPool<ViewProvider, WasmChecker> {
...
+        max_transactions: u16,
+    ) -> impl Iterator<Item = ArcPoolTx> + '_ {
         self.by_ratio_gas_tip
             .sort
             .iter()
             .rev()
-            .map(|(_, tx)| tx.clone())
+            .enumerate()
+            .map_while(move |(i, (_, tx))| {
+                if i <= (max_transactions as usize) {
```
acerone85, line 139: IIUC, here I need to check against `max_transactions - 1` to allow for the mint transaction to be included last. Probably need to consider also the slots reserved for FTI transactions (is the mint transaction considered to be a FTI transaction?)

crates/services/upgradable-executor/wasm-executor/src/ext.rs
```diff
@@ -148,7 +148,10 @@ pub fn input(size: usize) -> anyhow::Result<InputDeserializationType> {
 }
 
 /// Gets the next transactions by using the host function.
-pub fn next_transactions(gas_limit: u64) -> anyhow::Result<Vec<MaybeCheckedTransaction>> {
+pub fn next_transactions(
+    gas_limit: u64,
+    _max_transactions: u16,
```
acerone85, line 153: Transactions are fetched via a host function here. I am not sure how to retrieve up to `max_transactions` of them without changing the host function interface (assuming that under the hood the host function implementation will change the host state and remove the transactions)

crates/services/upgradable-executor/src/instance.rs
```diff
@@ -176,7 +190,64 @@ impl Instance<Created> {
...
+                .entry(encoded_size)
+                .or_default()
+                .push(encoded_txs);
+            Ok(encoded_size)
+        };
+
+        Func::wrap(&mut self.store, closure)
+    }
+
+    fn peek_next_txs_size_with_limit<TxSource>(
```
acerone85, line 230: Ideally I would like to factor the closure to a dedicated function, and call it both from `peek_next_txs_size` and `peek_next_txs_size_with_limits`, but this causes the type of the closure to become `FnOnce` (instead of `Fn`), and `Func::wrap` cannot be applied to it.

CHANGELOG.md
```diff
@@ -26,6 +27,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2158](https://github.com/FuelLabs/fuel-core/pull/2158): Log the public address of the signing key, if it is specified
 - [2188](https://github.com/FuelLabs/fuel-core/pull/2188): Upgraded the `fuel-vm` to `0.57.0`. More information in the [release](https://github.com/FuelLabs/fuel-vm/releases/tag/v0.57.0).
 
+
```
xgreenx, line 30:
```suggestion
```

crates/services/executor/src/executor.rs
```diff
@@ -565,11 +573,16 @@ where
...
-        // TODO: Handle `remaining_tx_count` https://github.com/FuelLabs/fuel-core/issues/2114
-        let remaining_tx_count = u16::MAX;
         // TODO: Handle `remaining_size` https://github.com/FuelLabs/fuel-core/issues/2133
         let remaining_size = u32::MAX;
 
+        // We allow at most u64::MAX transactions in a block, including the mint transaction.
+        // When processing l2 transactions, we must take into account transactions from the l1
+        // that have been included in the block already (stored in `data.tx_count`), as well
+        // as the final mint transaction.
+        let max_tx_count = u16::MAX.saturating_sub(1);
```
xgreenx, line 583: We can store this value in some constant=)

tests/tests/blocks.rs
```diff
@@ -423,4 +424,95 @@ mod full_block {
...
+            .expect("Second last block should be defined");
+        let last_block = client
+            .block_by_height(BlockHeight::from(2))
+            .await
+            .unwrap()
+            .expect("Last Block should be defined");
+        assert_eq!(second_last_block.transactions.len(), 65_535);
+        assert_eq!(
+            last_block.transactions.len(),
+            (tx_count as usize - 65_534) + 1
```
xgreenx, line 515:
```suggestion
            (tx_count as usize - u16::MAX) + 2 /* mint transaction from two blocks */
```

tests/tests/blocks.rs
```diff
@@ -423,4 +424,95 @@ mod full_block {
...
+            .block_by_height(BlockHeight::from(1))
+            .await
+            .unwrap()
+            .expect("Second last block should be defined");
+        let last_block = client
+            .block_by_height(BlockHeight::from(2))
+            .await
+            .unwrap()
+            .expect("Last Block should be defined");
+        assert_eq!(second_last_block.transactions.len(), 65_535);
```
xgreenx, line 512:
```suggestion
        assert_eq!(second_last_block.transactions.len(), u16::MAX);
```

xgreenx (commented):
LGTM!=)

tests/tests/blocks.rs
```diff
@@ -423,4 +424,95 @@ mod full_block {
...
+    use rand::{
+        rngs::StdRng,
+        Rng,
+        SeedableRng,
+    };
+
+    #[tokio::test]
+    async fn too_many_transactions_are_split_in_blocks() {
+        let max_gas_limit = 50_000_000;
+        let mut rng = StdRng::seed_from_u64(2322);
```
netrome, line 445: Just curious, why 2322? This is what I got from googling it https://www.sosyncd.com/the-meaning-of-the-2322-angel-number/ :blush:

tests/tests/blocks.rs
```diff
@@ -423,4 +424,95 @@ mod full_block {
...
+                    1000 + i,
+                    Default::default(),
+                    Default::default(),
+                )
+                .add_output(Output::Change {
+                    amount: 0,
+                    asset_id: Default::default(),
+                    to: rng.gen(),
+                })
+                .finalize_as_transaction()
```
netrome, line 492: It would be nice to break this out to a helper function for readability.

tests/tests/blocks.rs
```diff
@@ -423,4 +424,95 @@ mod full_block {
...
+    };
+
+    use rand::{
+        rngs::StdRng,
+        Rng,
+        SeedableRng,
+    };
+
+    #[tokio::test]
+    async fn too_many_transactions_are_split_in_blocks() {
```
netrome, line 443: Can you divide this test into "Given", "When" "Then" sections? It makes it nicer to read.

netrome (approved):
Nice stuff. Some minor thoughts from me.

tests/tests/blocks.rs
acerone85: Not sure, I just copy-pasted from other tests :)

acerone85: but I like the `inner wisdom and strength` argument from the website :D

netrome: Yeah, perhaps that's the origin of the number :sweat_smile:

netrome (dismissed):
Nice!

netrome (approved):

crates/services/executor/src/executor.rs
```diff
@@ -565,11 +577,15 @@ where
...
         // TODO: Handle `remaining_size` https://github.com/FuelLabs/fuel-core/issues/2133
         let remaining_size = u32::MAX;
 
+        // We allow at most u16::MAX transactions in a block, including the mint transaction.
+        // When processing l2 transactions, we must take into account transactions from the l1
+        // that have been included in the block already (stored in `data.tx_count`), as well
+        // as the final mint transaction.
+        let mut remaining_tx_count = MAX_TX_COUNT.saturating_sub(data.tx_count);
+
         let mut regular_tx_iter = l2_tx_source
```
xgreenx, line 589: If the `next` returns more data than you requested, you will not catch a situation in which you have more transitions than allowed. Unit test is passing since `OnceTransactionsSource` handles it. But in the case of another `TxSource` it may not work.

acerone85: Good point. All TxSources implement next to never return more than `remaining_tx_count` transactions, but I have added a `.take` to avoid this scenario: 

Two remarks: 
1. The new behaviour of the function means that if we use a "bad" transaction source, transactions over the limit will be removed and not be included in a block. If we want to be completely safe we could have a new `SafeTransactionSource` trait with the same signature of `TransactionsSource`, and implement a struct `CacheableTransactionSource<TxSource>` which is implemented whenever `TxSource: TransactionsSource`, 
and keeps transactions over the limit in memory. 
Given that we control the implementations of `TransactionsSource`, this is probably an overkill.

2. The new change is not tested, I'd need to implement a `BadTransactionSource` and modify the commit_without_execute function to run with that transaction source. I can implement that if necessary,  but I'd prefer to open a dedicated PR for it.

xgreenx: Yeah, `BadTransactionSource` sounds good=D

xgreenx (approved):

netrome (approved):

Issue #2184, opened 2024-09-11
chore(gas_price_service): refactor service initialisation in sub_services module


rymnc:
ref: https://github.com/FuelLabs/fuel-core/pull/2163#discussion_r1750191150

need to re-architect the gas price service initialisation by hiding gas price algorithm details etc.

`InitializeTask` can be a `RunnableService`, but `GasPriceService` should only implement `RunnableTask`. Try to use `TaskParams` for initialization, and handle sub service lifecycle within `GasPriceService`

## Definition of done
- [x] gas price service is initialised in the sub_services module, and all its internal details are abstracted away.

rymnc:
addressed in #2256

Issue #2166, opened 2024-09-05
Use up-to-date L1 blob costs to keep the gas price algorithm up-to-date


MitchTurner:
Even when it hasn't received data from committed blocks, we can still tell the algorithm what the true `latest_da_cost_per_byte`. This will need to be adjusted to address compression rates however, so we will need to track some rolling compression rate.

PR #2192, opened 2024-09-13
chore(gas_price_service): refactor gas price service

rymnc:
## Linked Issues/PRs

- https://github.com/FuelLabs/fuel-core/issues/2184
- https://github.com/FuelLabs/fuel-core/pull/2163

## Description

This PR moves the L2 block source and DA block cost source out of the algorithm updater and into the `GasPriceService` for improved handling of lifecycle and reducing 1 layer of abstraction. There will be a follow up PR to sunset `InitializeTask`. I did not include it in this PR to reduce overhead since there were quite a few significant changes needed to be made to the l2 block source.

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/fuel-core/src/service/sub_services/algorithm_updater.rs
+++ b/crates/fuel-core/src/service/sub_services/algorithm_updater.rs
@@ -15,11 +15,7 @@ use crate::{
 
 use fuel_core_gas_price_service::{
     fuel_gas_price_updater::{
-        da_source_adapter::{
-            dummy_costs::DummyDaBlockCosts,
-            DaBlockCostsProvider,
-            DaBlockCostsSharedState,
-        },
+        da_source_adapter::dummy_costs::DummyDaBlockCosts,
         fuel_core_storage_adapter::{
             storage::GasPriceMetadata,
             FuelL2BlockSource,
@@ -39,7 +35,6 @@ use fuel_core_gas_price_service::{
 use fuel_core_services::{
     stream::BoxStream,
     RunnableService,
-    Service,
     StateWatcher,
 };
 use fuel_core_storage::{
@@ -60,11 +55,7 @@ use fuel_core_types::{
     services::block_importer::SharedImportResult,
 };
 
-type Updater = FuelGasPriceUpdater<
-    FuelL2BlockSource<ConsensusParametersProvider>,
-    MetadataStorageAdapter,
-    DaBlockCostsSharedState,
->;
+type Updater = FuelGasPriceUpdater<MetadataStorageAdapter>;
 
 pub struct InitializeTask {
     pub config: Config,
@@ -74,13 +65,17 @@ pub struct InitializeTask {
     pub on_chain_db: Database<OnChain, RegularStage<OnChain>>,
     pub block_stream: BoxStream<SharedImportResult>,
     pub shared_algo: SharedGasPriceAlgo<Algorithm>,
-    pub da_block_costs_provider: DaBlockCostsProvider<DummyDaBlockCosts>,
 }
 
 type MetadataStorageAdapter =
     StructuredStorage<Database<GasPriceDatabase, RegularStage<GasPriceDatabase>>>;
 
-type Task = GasPriceService<Algorithm, Updater>;
+type Task = GasPriceService<
+    Algorithm,
+    Updater,
+    FuelL2BlockSource<ConsensusParametersProvider>,
+    DummyDaBlockCosts,
+>;
 
 impl InitializeTask {
     pub fn new(
@@ -98,12 +93,6 @@ impl InitializeTask {
         let default_metadata = get_default_metadata(&config, latest_block_height);
         let algo = get_best_algo(&gas_price_db, default_metadata)?;
         let shared_algo = SharedGasPriceAlgo::new_with_algorithm(algo);
-        // there's no use of this source yet, so we can safely return an error
-        let da_block_costs_source =
-            DummyDaBlockCosts::new(Err(anyhow::anyhow!("Not used")));
-        let da_block_costs_provider =
-            DaBlockCostsProvider::new(da_block_costs_source, None);
-
         let task = Self {
             config,
             genesis_block_height,
@@ -112,7 +101,6 @@ impl InitializeTask {
             on_chain_db,
             block_stream,
             shared_algo,
-            da_block_costs_provider,
         };
         Ok(task)
     }
@@ -170,19 +158,26 @@ impl RunnableService for InitializeTask {
         let updater = get_synced_gas_price_updater(
             self.config,
             self.genesis_block_height,
-            self.settings,
+            self.settings.clone(),
             self.gas_price_db,
             self.on_chain_db,
-            self.block_stream,
-            self.da_block_costs_provider.shared_state,
         )?;
 
-        self.da_block_costs_provider
-            .service
-            .start_and_await()
-            .await?;
-        let inner_service =
-            GasPriceService::new(starting_height, updater, self.shared_algo).await;
+        let l2_block_source = FuelL2BlockSource::new(
+            self.genesis_block_height,
+            self.settings,
+            self.block_stream,
+        );
+        let da_block_cost_source = DummyDaBlockCosts::new(Err(anyhow::anyhow!("unused")));
+
+        let inner_service = GasPriceService::new(
+            starting_height,
+            updater,
+            self.shared_algo,
+            l2_block_source,
+            da_block_cost_source,
+        )
+        .await;
         Ok(inner_service)
     }
 }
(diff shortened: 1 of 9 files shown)


crates/services/gas_price_service/src/lib.rs
```diff
@@ -152,21 +186,43 @@ where
...
                 should_continue = false;
             }
-            new_algo = self.update_algorithm.next() => {
-                let new_algo = new_algo?;
+            da_block_costs = self.da_block_costs_provider.recv() => {
+                self.da_block_costs = match da_block_costs {
+                    Ok(da_block_costs) => Some(da_block_costs),
+                    Err(err) => {
+                        tracing::error!("Failed to get da block costs: {:?}", err);
+                        None
```
xgreenx, line 194: If you have some value already, but the next request failed, you will set it to be `None`, while old value still is good=)

crates/services/gas_price_service/src/lib.rs
```diff
@@ -152,21 +186,43 @@ where
...
+                    }
+                };
+                should_continue = true;
+            }
+            l2_block = self.l2_block_source.get_l2_block() => {
+                let l2_block = l2_block?;
+                let da_block_costs = self.da_block_costs.clone();
```
xgreenx, line 201: Maybe you want to do `take` instead of `clone`=)

crates/services/gas_price_service/src/fuel_gas_price_updater/da_source_adapter/service.rs
```diff
@@ -13,39 +11,93 @@ use std::{
...
 pub use anyhow::Result;
+use tokio::sync::broadcast::Receiver;
+
+pub struct DaBlockCostsProvider<Source: DaBlockCostsSource + 'static> {
+    handle: ServiceRunner<DaBlockCostsService<Source>>,
+    subscription: Receiver<DaBlockCosts>,
+}
+
+#[async_trait::async_trait]
+pub trait DaBlockCostsProviderPort {
```
xgreenx, line 30: I don't see the reason of having trait for `DaBlockCostsProvider`. It looks like we can just implement methods directly on the `DaBlockCostsProvider`.

crates/services/gas_price_service/src/fuel_gas_price_updater/da_source_adapter/service.rs
```diff
@@ -13,39 +11,93 @@ use std::{
...
     time::{
         interval,
         Interval,
     },
 };
 
 pub use anyhow::Result;
+use tokio::sync::broadcast::Receiver;
+
+pub struct DaBlockCostsProvider<Source: DaBlockCostsSource + 'static> {
```
xgreenx, line 24: Why do we want to wrap `DaBlockCostsService` into `DaBlockCostsProvider`? The gas price service already knows about `DaBlockCostsService` type. We can just work directly with this type.

The gas price service can subscribe in `new_service` from the shared state of the `DaBlockCostsService`

crates/services/gas_price_service/src/fuel_gas_price_updater/da_source_adapter/service.rs
```diff
@@ -13,39 +11,93 @@ use std::{
...
+
+impl DaBlockCostsServiceSharedState {
+    pub fn subscribe(&self) -> Receiver<DaBlockCosts> {
+        self.sender.subscribe()
+    }
+
+    fn send(&self, da_block_costs: DaBlockCosts) -> Result<()> {
+        self.sender.send(da_block_costs)?;
+        Ok(())
+    }
```
xgreenx, line 74: [nit] You have access to private fields of this structure, so maybe we don't need this helper function

crates/fuel-core/src/service/sub_services/algorithm_updater.rs
```diff
@@ -193,8 +188,6 @@ pub fn get_synced_gas_price_updater(
     settings: ConsensusParametersProvider,
```
xgreenx, line 188: I think that most of the logic of this function should live inside the service itself. Only a small part of the `sync_metadata_storage_with_on_chain_storage` logic related to the on-chain database should live here

crates/services/gas_price_service/src/lib.rs
```diff
@@ -60,18 +84,23 @@ pub trait UpdateAlgorithm {
     fn start(&self, for_block: BlockHeight) -> Self::Algorithm;
 
     /// Wait for the next algorithm to be available
-    async fn next(&mut self) -> anyhow::Result<Self::Algorithm>;
+    async fn next(
```
xgreenx, line 87: Do we still need `async` here?

crates/services/gas_price_service/src/lib.rs
```diff
@@ -152,21 +186,43 @@ where
...
+            if let Some(new_algo) = self
+                .update_algorithm
+                .next(l2_block, da_block_costs)
+                .now_or_never()
+            {
+                let new_algo = new_algo?;
+                tracing::debug!("Updating gas price algorithm");
+                self.update(new_algo).await;
+            }
         }
```
xgreenx, line 226: It should be a loop, since we can have more blocks than one. Plus it would be nice to avoid logic duplication, could you move the same logic into the function and reuse it here and in `run` function?

crates/services/gas_price_service/src/fuel_gas_price_updater/da_source_adapter/service.rs
rymnc: my initial impl did this, it seemed cleaner to use the `DaBlockCostsProvider` to wrap over the service, and have only one field on the `GasPriceService`, which is responsible to get new block costs and the service handle :)

crates/services/gas_price_service/src/fuel_gas_price_updater/da_source_adapter/service.rs
rymnc: addressed in d68b02d83

crates/services/gas_price_service/src/lib.rs
rymnc: addressed in 2f33f7b0a

crates/services/gas_price_service/src/lib.rs
rymnc: addressed in f0bbf4974

crates/services/gas_price_service/src/lib.rs
rymnc: addressed in 00e0d0600

crates/services/gas_price_service/src/fuel_gas_price_updater/da_source_adapter/service.rs
rymnc: addressed in 69651a53a

crates/fuel-core/src/service/sub_services/algorithm_updater.rs
rymnc: I agree ~ this is part of the follow up PR though, which sunsets `InitializeTask` appropriately.

crates/services/gas_price_service/src/fuel_gas_price_updater/da_source_adapter/service.rs
rymnc: I would prefer if we maintained this structure instead :)

crates/services/gas_price_service/src/lib.rs
rymnc: avoided duplication in 911c06f71

i didn't understand why this should be a loop since we're shutting down here :)

crates/services/gas_price_service/src/fuel_gas_price_updater.rs
```diff
@@ -24,12 +24,9 @@ pub mod fuel_core_storage_adapter;
 
 pub mod da_source_adapter;
 
-pub struct FuelGasPriceUpdater<L2, Metadata, DaBlockCosts> {
+pub struct FuelGasPriceUpdater<Metadata> {
```
MitchTurner, line 27: Hmm. I'm not sure how much we are actually gaining from this. I was kinda thinking that we could get rid of the `UpdateAlgorithm` abstraction completely, including removing this struct.

But I don't think the `gas-price-algorithm` should depend directly on some algorithm, in which case we want to keep the `UpdateAlgorithm` trait.

This approach commits us to a specific trait interface with
```rs
        l2_block: BlockInfo,
        da_block_costs: Option<DaBlockCosts>,
```
whereas before the `gas-price-service` didn't know anything about the dependencies of the updater.

MitchTurner: We are going to want to add another value here soon with this issue:
https://github.com/FuelLabs/fuel-core/issues/2166

rymnc: okay. then I would propose to move the L2 source and DA back into the `FuelGasPriceUpdater` and delegate sub-service startup and shutdown to it. does that sound okay?

rymnc: (it also means we turn `FuelGasPriceUpdater` into a RunnableTask)

rymnc:
closing in favor of another approach :)

Issue #2190, opened 2024-09-12
Gas Price Analyzer isn't part of workspace, so it isn't formated or linted


MitchTurner:
We should either include it in the workspace or include something in CI to check the code. I'd prefer just including it in the workspace so that tooling just works with it out of the box and we get formatting for free. 

Maybe we should considering putting it into its own repo eventually, or in `fuel-utils`? All worth considering, but the workflow isn't fun currently.

MitchTurner:
We might want to move it to the `fuel-utils` crate once we "stabilize" the algorithm.

Adding it to the `fuel-core` workspace will bloat the cargo.lock and the number of dependencies we take.

PR #2194, opened 2024-09-13
improvement: Include in workspace, fix resulting lint warnings

MitchTurner:
## Linked Issues/PRs
https://github.com/FuelLabs/fuel-core/issues/2190

## Description
Improve developer experience by moving the algo analyzer to the workspace. This allows it to receive all the global linting and formatting configurations and helps the IDE work in the analyzer and algorithm dirs at the same time.

### Before requesting review


Cargo.lock
```diff
@@ -1250,6 +1250,12 @@ version = "1.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c3ac9f8b63eca6fd385229b3675f6cc0dc5c8a5c8a54a59d4f52ffd670d87b0c"
 
+[[package]]
```
xgreenx, line 1253: Moving this tool into workspace adds a lot of dependencies into `Cargo.lock`. Plus it now requires compilation of this crate during `cargo test`.

I still prefer not to include it into workspace. Can we have it in a separate repository? Maybe in `fuel-utils`?

MitchTurner: That's where I feel it should live long-term, but I find that I'm constantly making changes to the algorithm code when I'm working on this and it would be nice to have them in the same place.

xgreenx:
I think we can close it for now=)

Issue #1887, opened 2024-05-08
Handle `ExecutionResult` and `ValidationResult` separately in Wasm executor


MitchTurner:
Because of the way we are doing serialization, we are returning an `ExecutionResult` from "validate" on the wasm executor. We should be able to handle more cases.

See convo:
https://github.com/FuelLabs/fuel-core/pull/1886#discussion_r1593153448

Include good integ tests to show that this works :)

Issue #2198, opened 2024-09-15
Add forward compatibility test for `fuel-core 0.36.0`


xgreenx:
After release of `fuel-core 0.36.0` we need to add forward compatibility test for it.

<img width="1148" alt="image" src="https://github.com/user-attachments/assets/63f0e447-81c9-4935-abd7-2b32f5f63422">

acerone85:
I'm happy to have a look at this one after 0.36.0 is released

rafal-ch:
Closed with https://github.com/FuelLabs/fuel-core/pull/2271

Issue #2133, opened 2024-08-27
Add support block size limit


xgreenx:
Currently, the block is limited by the gas, and with the `63` gas per byte, the maximum size of the block can be `30000000/63 = 476_190` byte.

But since we want to have a hard limit 128kb or 256kb, we either need to increase the gas per byte, or add a new block size limit variable.

Support the block size limit requires:
- Add a new `block_size_limit` field to the consensus parameters.
- Updating the executor's logic to get transactions based on the remaining size.
- Update the transaction pool to support the limitation by size.

xgreenx:
This feature requires breaking change into the WASM executor since we need to pass the size of the block when fetching transactions from the TxPool. It would be nice to release this breaking change along with https://github.com/FuelLabs/fuel-core/issues/2114 and https://github.com/FuelLabs/fuel-core/issues/1887

rafal-ch:
A few comments w.r.t to implementation:
* new consensus parameter – `block_transaction_size_limit`
  * not `block_size_limit` due to what we’re gonna be actually limiting is the set of L2 transactions
* we’re ok if occasionally the block will get bigger than a limit due to L1 transactions, mint transaction or a header. Hence, if we want block to be limited to roughly 128kb, we coule set the `block_transaction_size_limit` to 125-126kb
* size of a transaction = size of the result of canonical serialization of transaction

PR #2199, opened 2024-09-15
Applying several breaking changes to the WASM interface from backlog

xgreenx:
Closes https://github.com/FuelLabs/fuel-core/issues/1887
Closes https://github.com/FuelLabs/fuel-vm/issues/797
Related https://github.com/FuelLabs/fuel-core/issues/2114
Related https://github.com/FuelLabs/fuel-core/issues/2133

This PR contains several small breaking changes to the WASM interface:
- Get the module to execute WASM byte code from the storage first, an fallback to the built-in version in the case of the `FUEL_ALWAYS_USE_WASM`.
- Added `host_v1` with a new `peek_next_txs_size` method, that accepts `tx_number_limit` and `size_limit`.
- Added new variant of the return type to pass the validation result. It removes block serialization and deserialization and should improve performance.
- Added a V1 execution result type that uses `JSONError` instead of postcard serialized error. It adds flexibility of how variants of the error can be managed. More information about it in https://github.com/FuelLabs/fuel-vm/issues/797. The change also moves `TooManyOutputs` error to the top. It shows that `JSONError` works as expected.
- Use a WASM generated bytecode for the local testnet since we always want to use the latest version compatible with native execution.

## Checklist

### Before requesting review
    - https://github.com/FuelLabs/fuel-core/issues/2198
    - https://github.com/FuelLabs/fuel-core/issues/2200


rafal-ch (dismissed):
LGTM :+1:

crates/services/upgradable-executor/src/instance.rs
```diff
@@ -108,9 +167,23 @@ impl Instance {
 }
 
 impl<Stage> Instance<Stage> {
-    const LATEST_HOST_MODULE: &'static str = "host_v0";
+    const V0_HOST_MODULE: &'static str = "host_v0";
+    const LATEST_HOST_MODULE: &'static str = "host_v1";
```
acerone85, line 171: In general I would avoid using a `LATEST_HOST_MODULE` to refer to the latest module, as it is more error prone when introducing new versions. E.g. when we introduce V2, a new `add_method_v1` function will need to be added and the latest `peek_next_txs_size` will need to be refactored to be in `v1`.

crates/services/upgradable-executor/wasm-executor/src/utils.rs
```diff
@@ -80,16 +84,166 @@ pub enum WasmDeserializationBlockTypes<TxSource> {
...
+/// versions of the return without introducing new host functions.
+#[cfg(not(feature = "std"))]
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub enum ReturnType {
+    ExecutionV0,
+    ExecutionV1(Result<Uncommitted<ExecutionResult<JSONError>, Changes>, JSONError>),
+    Validation(Result<Uncommitted<ValidationResult, Changes>, JSONError>),
+}
+
+/// Converts the latest execution result to the v1 execution result.
```
acerone85, line 143: what's the difference between the latest execution result and the v1 execution result? (I thought v1 is the latest version)

crates/services/upgradable-executor/wasm-executor/src/utils.rs
```diff
@@ -80,16 +84,166 @@ pub enum WasmDeserializationBlockTypes<TxSource> {
...
+    ExecutionV1(Result<Uncommitted<ExecutionResult<JSONError>, Changes>, JSONError>),
+    Validation(Result<Uncommitted<ValidationResult, Changes>, JSONError>),
+}
+
+/// The return type for the WASM executor. Enum allows handling different
+/// versions of the return without introducing new host functions.
+#[cfg(not(feature = "std"))]
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub enum ReturnType {
+    ExecutionV0,
```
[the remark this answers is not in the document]
xgreenx: WASM executor doesn't use `ExecutionV0` variant, so from its perspective it is empty. While from an upgradable perspective it has a type.

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -426,7 +434,12 @@ where
             match &self.execution_strategy {
                 ExecutionStrategy::Native => self.native_validate_inner(block, options),
                 ExecutionStrategy::Wasm { module } => {
-                    self.wasm_validate_inner(module, block, options)
+                    let maybe_blocks_module = self.get_module(block_version);
```
MitchTurner, line 437: nit: same thing here

version-compatibility/forkless-upgrade/src/tests_helper.rs
```diff
@@ -42,7 +44,7 @@ macro_rules! define_core_driver {
             /// This must be before the _db_dir as the drop order matters here.
             pub node: $service,
             pub _db_dir: tempfile::TempDir,
-            #[allow(dead_code)] // Only some of the clients generated by this macro are used
+            // Only some of the clients generated by this macro are used
```
MitchTurner, line 47: Do you still need this comment without the `allow`?

MitchTurner (commented):
I'll finish reviewing later. I've read through most of it, but I want to give it another pass.

crates/fuel-core/src/service/adapters.rs
```diff
@@ -122,6 +132,23 @@ impl ExecutorAdapter {
...
+        component: Components<Vec<Transaction>>,
+    ) -> ExecutorResult<UncommittedResult<Changes>> {
+        let new_components = Components {
+            header_to_produce: component.header_to_produce,
+            transactions_source: OnceTransactionsSource::new(
+                component.transactions_source,
+            ),
+            gas_price: component.gas_price,
+            coinbase_recipient: component.coinbase_recipient,
+        };
```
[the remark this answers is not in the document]
xgreenx: We wrap the vector of transaction into `OnceTransactionsSource`. I just move this code from block producer port to the adapter directly to be able to reuse in the tests=)

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -383,7 +386,12 @@ where
                     self.native_produce_inner(block, options, dry_run)
                 }
                 ExecutionStrategy::Wasm { module } => {
-                    self.wasm_produce_inner(module, block, options, dry_run)
+                    let maybe_blocks_module = self.get_module(block_version);
```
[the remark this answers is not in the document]
xgreenx: It is a bug, that we used `WASM_BYTECODE` from the binary itself instead of using value from the database(if we have any). For example, if you run the node with `FUEL_ALWAYS_USE_WASM`, it will use the built-in(into `fuel-core` binary) WASM byte code instead of the byte code that comes with the upgrade of the network. If you compiled your binary with wrong features flag, it can stop you form syncing with the network even with WASM executor

crates/services/upgradable-executor/src/instance.rs
```diff
@@ -53,6 +56,17 @@ use wasmtime::{
 trait CallerHelper {
     /// Writes the encoded data to the memory at the provided pointer.
     fn write(&mut self, ptr: u32, encoded: &[u8]) -> anyhow::Result<()>;
+
+    /// Peeks the next transactions from the source and returns the size of the encoded transactions.
+    fn peek_next_txs_size<Source>(
```
MitchTurner, line 61: Oh. Can we specify that this is `size_in_bytes` or maybe just `peek_next_txs_bytes`? It wasn't obvious that this was bytes and not # of txs.

I guess within the wasm context it would kinda have to be bytes... but it wasn't obvious immediately to me.

MitchTurner (dismissed):
Okay. LGTM. Reading through everything again today it clicked what everything here is doing. 


It felt like a bunch of random changes and I was going to say you should have broken it into separate PRs, but it's all kinda related.

I added a nitty comment, but other than that I'm okay with this.

MitchTurner (dismissed):

MitchTurner (dismissed):

MitchTurner (approved):

Dentosal (approved):

rafal-ch (approved):
LGTM :+1:

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -383,7 +386,12 @@ where
                     self.native_produce_inner(block, options, dry_run)
                 }
                 ExecutionStrategy::Wasm { module } => {
-                    self.wasm_produce_inner(module, block, options, dry_run)
+                    let maybe_blocks_module = self.get_module(block_version).ok();
+                    if let Some(blocks_module) = maybe_blocks_module {
+                        self.wasm_produce_inner(&blocks_module, block, options, dry_run)
+                    } else {
+                        self.wasm_produce_inner(module, block, options, dry_run)
+                    }
```
acerone85, line 394: question: is this needed in the case that the node does not have the most recent version of the bytecode available, hence it will fallback to the one specified in the `ExecutionStrategy::Wasm` value?

xgreenx: It is for the case of the `FUEL_ALWAYS_USE_WASM=true`. We use it in the cases of tests, benchmarks, or if you just want to run WASM executor instead of the native execution.

The ideas of this change is to just use the write state transition instead of WASM version of the native state transition logic, since it can be different if you compiled the binary with different list of feature(it affects the native execution code and the WASM version as well).

Issue #2202, opened 2024-09-16
feat(p2p): Handle `dnsaddr` resolution


rymnc:
When we accept a top-level `dnsaddr` of the form as an entry to bootstrap nodes cli arg, like `/dnsaddr/domain.ext` we do not perform resolution of the dnsaddr protocol, and therefore do not include those peers in our dht. 

start with https://github.com/FuelLabs/fuel-core/blob/da1f69fa1ef16212473fda4b46d7f32cbb46f255/crates/services/p2p/src/lib.rs#L37-L44, and add a match case for `DnsAddr`, similar to https://github.com/libp2p/rust-libp2p/blob/a2a281609a0a64b211f7917aa856924983b63200/transports/dns/src/lib.rs#L540-L577 because they drop them in https://github.com/libp2p/rust-libp2p/blob/a2a281609a0a64b211f7917aa856924983b63200/transports/dns/src/lib.rs#L338

## Definition of Done

- We can start `fuel-core` with a dnsaddr that does not have a `PeerId` included and it should manage the `_dnsaddr` dns txt lookup, recursively fetch all the associated multiaddresses and include them in the dht, and start syncing.

PR #2204, opened 2024-09-16
chore(p2p): dnsaddr recursive resolution

rymnc:
## Linked Issues/PRs

- fixes https://github.com/FuelLabs/fuel-core/issues/2202

## Description

- creates a new module `dnsaddr_resolution` which handles recursive dnsaddr resolution to add to the DHT *without* suffix matching. This way we can connect to all peers behind a domain without specifying the exact `PeerId`, like `/dnsaddr/bootstrap.libp2p.io`.
- asyncifies some of the methods related to mounting the p2p service
- resultifies some of the methods related to mounting the p2p service 

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/p2p/Cargo.toml
```diff
@@ -50,6 +50,7 @@ thiserror = "1.0.47"
 tokio = { workspace = true, features = ["sync"] }
 tracing = { workspace = true }
 void = "1"
+hickory-resolver = "0.25.0-alpha.2"
```
xgreenx, line 53: Can we use resolver from `libp2p`?

rymnc: do you mean the same version as libp2p or to use their resolver?

their resolver is in libp2p-dns (https://docs.rs/libp2p-dns/latest/libp2p_dns/tokio/type.Transport.html) and doesn't export any helper functions to resolve dnsaddr's.

xgreenx: I see, they use the same library to do it. Yeah. we need to use the same version to not bloat the `Cargo.lock`

rymnc: addressed in [4f4b662](https://github.com/FuelLabs/fuel-core/pull/2204/commits/4f4b662c4f9c6e0ac22d381075c58a58723329a5)

crates/services/p2p/src/discovery/discovery_config.rs
```diff
@@ -167,13 +162,42 @@ impl Config {
...
+                None
+            }
+        })
+        .collect::<Vec<_>>();
+
+    let dns_resolver = DnsResolver::new().await?;
+    let mut dnsaddr_multiaddrs = vec![];
+
+    for dnsaddr in dnsaddr_urls {
+        let multiaddrs = dns_resolver.lookup_dnsaddr(dnsaddr.as_ref()).await?;
```
xgreenx, line 194: What if `dnsaddr` will have another `dnsaddr`, will it work?

crates/services/p2p/src/discovery/dnsaddr_resolution.rs
```diff
@@ -0,0 +1,101 @@
...
+use std::pin::Pin;
+
+/// The prefix for `dnsaddr` protocol TXT record lookups.
+const DNSADDR_PREFIX: &str = "_dnsaddr.";
+/// The maximum number of DNS lookups when dialing.
+/// This limit is for preventing malicious or misconfigured DNS records from causing infinite recursion.
+const MAX_DNS_LOOKUPS: usize = 10;
+
+#[async_trait::async_trait]
+pub trait DnsLookup {
```
xgreenx, line 13: Why do we need trait?

crates/services/p2p/src/discovery/discovery_config.rs
```diff
@@ -97,7 +102,7 @@ impl Config {
         self
     }
 
-    pub fn finish(self) -> Behaviour {
+    pub async fn finish(self) -> anyhow::Result<Behaviour> {
```
xgreenx, line 105: Just curious, have you considered implementation that doesn't require `async` during construction? Maybe it is possible to resolved addresses in the `fn start`?

crates/services/p2p/src/discovery/discovery_config.rs
rymnc: yup, resolution is recursive. the test case handles that :)

crates/services/p2p/src/discovery.rs
```diff
@@ -265,7 +268,10 @@ mod tests {
                 .with_bootstrap_nodes(bootstrap_nodes)
                 .with_random_walk(Duration::from_millis(500));
 
-            config.finish()
+            async move { config.finish().await }
+                .now_or_never()
+                .unwrap()
+                .unwrap()
```
netrome, line 274: I don't see how we can guarantee this future to be immediately ready. What happens if the `DnsResolver::new().await?` is pending for example? As I read this, we'd panic in that scenario. That feels quite brittle to me, in which case it would feel safer to just use the synchronous resolver instead.

My preferred option though would be to allow `build_behavior_fn` to return a function which returns a future, so we can return this future without having to panic if it isn't ready here.

crates/services/p2p/src/discovery/discovery_config.rs
rymnc: do you mean here - https://github.com/FuelLabs/fuel-core/blob/1d1560f2fd05b2d28f7df14c98bf65d98fbbcf53/crates/services/p2p/src/p2p_service.rs#L238 ?

crates/services/p2p/src/discovery/discovery_config.rs
```diff
@@ -167,13 +162,42 @@ impl Config {
...
+    let dnsaddr_urls = multiaddrs
+        .iter()
+        .filter_map(|node| {
+            if let Protocol::Dnsaddr(multiaddr) = node.iter().next()? {
+                Some(multiaddr.clone())
+            } else {
+                None
+            }
+        })
+        .collect::<Vec<_>>();
```
netrome, line 188: Nit: This reads a bit funky to me. Wouldn't the singular of `multiaddrs` be `multiaddr` rather than `node`? I'd suggest this naming instead:

```suggestion
    let dnsaddr_urls = multiaddrs
        .iter()
        .filter_map(|multiaddr| {
            if let Protocol::Dnsaddr(dnsaddr_url) = multiaddr.iter().next()? {
                Some(dnsaddr_url.clone())
            } else {
                None
            }
        })
        .collect::<Vec<_>>();
```

crates/services/p2p/src/discovery/dnsaddr_resolution.rs
```diff
@@ -0,0 +1,101 @@
...
+use libp2p::Multiaddr;
+use std::pin::Pin;
+
+/// The prefix for `dnsaddr` protocol TXT record lookups.
+const DNSADDR_PREFIX: &str = "_dnsaddr.";
+/// The maximum number of DNS lookups when dialing.
+/// This limit is for preventing malicious or misconfigured DNS records from causing infinite recursion.
+const MAX_DNS_LOOKUPS: usize = 10;
+
+#[async_trait::async_trait]
```
netrome, line 12: Async traits have been stabilized since [rust 1.75](https://blog.rust-lang.org/2023/12/28/Rust-1.75.0.html), so I'd prefer not to add this declaration for new traits. Instead just have the functions return `impl Future<Output = ...>` and use `async fn` in the implementations.

crates/services/p2p/src/discovery/dnsaddr_resolution.rs
```diff
@@ -0,0 +1,101 @@
...
+
+impl DnsResolver {
+    /// Internal method to handle recursive DNS lookups.
+    fn resolve_recursive<'a>(
+        &'a self,
+        addr: &'a str,
+        depth: usize,
+    ) -> Pin<
+        Box<dyn std::future::Future<Output = anyhow::Result<Vec<Multiaddr>>> + Send + 'a>,
+    > {
```
netrome, line 42: Do we need to do the `Pin<Box<...>>` here? That feels like an artifact of `async_trait` which should preferably be removed (as per my other suggestion), or if you want to keep the `async_trait`, this should happen in the trait implementation and not in this private helper method.

crates/services/p2p/src/discovery/dnsaddr_resolution.rs
```diff
@@ -0,0 +1,101 @@
...
+    async fn dns_resolver__parses_all_multiaddresses_from_ipfs_bootstrap_nodes() {
+        // given
+        let resolver = DnsResolver::new().await.unwrap();
+        // when
+        let multiaddrs = resolver
+            .lookup_dnsaddr("bootstrap.libp2p.io")
+            .await
+            .unwrap();
+        // then
+        assert!(!multiaddrs.is_empty());
```
netrome, line 99: Do we assume here that any machine running this test will always be able to look up the "bootstrap.libp2p.io" address in their environment?

For example, what happens if I'm flushing my local DNS cache and disconnect from the internet before running this test? As I read this, the test case would fail in that scenario.

I think we need to introduce a a port to manage the DNS lookup to make this logic testable.

netrome (changes requested):
I hate to block this PR, but there are two pretty substantial issues I can see so far that I would like to have addressed.

1. I don't think we should use `.now_or_never()` to assert that `config.finish().await` is ready in `build_behavior_fn`. It is not clear to me that this future will always be ready when this is invoked.
2. The DNS resolution tests should not be dependent on the host machines local DNS cache or network connection. Instead I'd suggest creating a port for this, which would allow us to create more tests for different DNS configurations.

Moreover, if a port is introduced to allow the DNS lookup logic to be tested I'd be interested in seeing a test case covering recursive lookups to make sure they work as expected.

crates/services/p2p/src/discovery/dnsaddr_resolution.rs
xgreenx: I think introduction of the port will make this tests useless, since it will just test how port itself works. Plus I'm not sure how hard is it I think the introduction of the port will make these tests useless since they will just test how the port itself works. Plus I'm not sure how hard it is to write your own DNS resolver.

I'm okay with the idea that this test fails without a connection or local DNS resolution.

Having a test with real DNS proves that it works in a real environment. The libp2p library tests it in the same way=)

netrome: While I see your point, I have to respectfully disagree. I think the port will help us make the tests **even more useful**. Right now we can't assert much more than `!multiaddrs.is_empty()`, because there's no way for us to control which DNS records we get when we look up `bootstrap.libp2p.io`. With the port, we for example can test that our logic parses the TXT records correctly.

As for another example, there's currently no way to know from this test if this test is doing a recursive lookup or just a plain single lookup - and this can vary depending on how `libp2p.io` configures their records. With the trait we can write multiple test cases to test and make clear assertions about different scenarios.

I don't think it would be too hard to put a trait between our code and the `TokioAsyncResolver`, because as far as I can see, we're only calling `TokioAsyncResolver::txt_lookup` so we only need to mock one method for the tests.

crates/services/p2p/src/discovery/dnsaddr_resolution.rs
rymnc: recursion needs us to pin the future

netrome: Ah right, makes sense :+1:

crates/services/p2p/src/discovery/discovery_config.rs
rymnc: addressed in 5b4671f1f

crates/services/p2p/src/discovery.rs
rymnc: deprecated in 5b4671f1f

crates/services/p2p/src/discovery/dnsaddr_resolution.rs
rymnc: removed :)

crates/services/p2p/src/discovery/dnsaddr_resolution.rs
xgreenx: Okay, I will try to put that in another way=)

The main functionality that I would like to see tested is that we can resolve the real addresses from ipfs(or fuel testnet/mainnet). So, I still want to see the test that uses real `dnsaddr` and verifies that `TokioAsyncResolver` works as expected.

The functionality of `resolve_recursive` and how this function works are not so important to me. 
The approach with the port for internal resolver can help to test the behaviour of the `resolve_recursive` function in different use cases. And it looks like the right call to do it.

It is up to @rymn. I'm okay with doing that in a separate PR just dedicated to better test coverage of the `resolve_recursive` since the main feature is implemented and we have an integration test that covers this feature=)

crates/services/p2p/src/discovery/discovery_config.rs
rymnc: addressed in 0482f8023

crates/services/p2p/src/discovery/dnsaddr_resolution.rs
rymnc: removed the trait

crates/services/p2p/src/discovery/dnsaddr_resolution.rs
rymnc: how about this? I have some dnsaddr's on my personal domain anyway, we can just reuse those with recursion and assert strictly about the multiaddrs this function spits out?

crates/services/p2p/src/discovery.rs
netrome: Nice, thank you!

netrome (commented):
I'll re-review tonight, but my main concerns have been addressed. Thank you!

crates/services/p2p/src/discovery/dnsaddr_resolution.rs
rymnc: addressed in [8c77c1d](https://github.com/FuelLabs/fuel-core/pull/2204/commits/8c77c1d418b19c39a0d7b25c5d7f5a5b8abc6f7c)

i won't touch the records, i promise 😂

netrome: Hahah pinky promise :joy: Nice comment also, super helpful thank you! :pray:

crates/services/p2p/src/dnsaddr_resolution.rs
```diff
@@ -79,17 +79,29 @@ impl DnsResolver {
...
+        let expected_multiaddrs: HashSet<Multiaddr> = vec![
+            "/dns4/sjc-1.bootstrap.libp2p.io/tcp/443/wss/p2p/QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN",
+        ].iter().map(|s| s.parse().unwrap()).collect();
+
         // when
-        let multiaddrs = resolver
-            .lookup_dnsaddr("bootstrap.libp2p.io")
-            .await
-            .unwrap();
+        // run a `dig +short txt rymnc.com` to get the TXT records
```
rymnc, line 93: err, this should be _dnsaddr.rymnc.com. patching.

crates/services/p2p/src/p2p_service.rs
```diff
@@ -168,20 +169,52 @@ pub enum FuelP2PEvent {
...
+    for dnsaddr in dnsaddr_urls {
+        let multiaddrs = dns_resolver.lookup_dnsaddr(dnsaddr.as_ref()).await?;
+        dnsaddr_multiaddrs.extend(multiaddrs);
+    }
+
+    Ok(multiaddrs
+        .into_iter()
+        .filter(|multiaddr| !multiaddr.iter().any(|p| matches!(p, Protocol::Dnsaddr(_))))
+        .chain(dnsaddr_multiaddrs.into_iter())
+        .collect::<Vec<_>>())
```
xgreenx, line 196:
```suggestion
    let resolved_multiaddrs = multiaddrs
        .into_iter()
        .filter(|multiaddr| !multiaddr.iter().any(|p| matches!(p, Protocol::Dnsaddr(_))))
        .chain(dnsaddr_multiaddrs.into_iter())
        .collect();
    Ok(resolved_multiaddrs)
```

crates/services/p2p/src/dnsaddr_resolution.rs
```diff
@@ -0,0 +1,107 @@
...
+        let expected_multiaddrs: HashSet<Multiaddr> = vec![
+            "/dns4/sjc-1.bootstrap.libp2p.io/tcp/443/wss/p2p/QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN",
+        ].iter().map(|s| s.parse().unwrap()).collect();
+
+        // when
+        // run a `dig +short txt _dnsaddr.rymnc.com` to get the TXT records
+        // notice that it contains -
+        // `dnsaddr=/dnsaddr/zone-1.rymnc.com/tcp/4001/p2p/QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN`
+        // which is a recursive call
+        let multiaddrs = resolver.lookup_dnsaddr("rymnc.com").await.unwrap();
```
xgreenx, line 97: Maybe it is better to use main net dnsaddr=)

xgreenx (commented):
LGTM=)

crates/services/p2p/src/dnsaddr_resolution.rs
rymnc: addressed in [92f5f22](https://github.com/FuelLabs/fuel-core/pull/2204/commits/92f5f220f77321cc2a5de2b04a20f92c58b4b5f0)

xgreenx (dismissed):

xgreenx (approved):

crates/services/p2p/src/dnsaddr_resolution.rs
```diff
@@ -0,0 +1,111 @@
...
+            "/dns/p2p-mainnet.fuel.network/tcp/30337/p2p/16Uiu2HAmQunK6Dd81BXh3rW2ZsszgviPgGMuHw39vv2XxbkuCfaw",
+            "/dns/p2p-mainnet.fuel.network/tcp/30333/p2p/16Uiu2HAkuiLZNrfecgDYHJZV5LoEtCXqqRCqHY3yLBqs4LQk8jJg",
+            "/dns/p2p-mainnet.fuel.network/tcp/30334/p2p/16Uiu2HAkzYNa6yMykppS1ij69mKoKjrZEr11oHGiM5Mpc8nKjVDM",
+            "/dns/p2p-mainnet.fuel.network/tcp/30335/p2p/16Uiu2HAm5yqpTv1QVk3SepUYzeKXTWMuE2VqMWHq5qQLPR2Udg6s"
+        ].iter().map(|s| s.parse().unwrap()).collect();
+
+        // when
+        // run a `dig +short txt _dnsaddr.mainnet.fuel.network` to get the TXT records
+        let multiaddrs = resolver
+            .lookup_dnsaddr("mainnet.fuel.network")
```
mchristopher, line 99: Can we switch this to `core-test.fuellabs.net`? The records on the `mainnet.fuel.network` record can and will change over time, but I just setup the `core-test` DNS record to be static.

crates/services/p2p/src/lib.rs
```diff
@@ -5,6 +5,7 @@ pub mod behavior;
 pub mod codecs;
 pub mod config;
 pub mod discovery;
+mod dnsaddr_resolution;
```
netrome, line 8: Nit: I'd move this declaration to a separate block than the public modules.

netrome (approved):
Looks good to me :+1: It would be nice to use the test records Mike suggested, but that can be done as a follow-up if you want to get this merged promptly for release.

crates/services/p2p/src/dnsaddr_resolution.rs
netrome: Thanks for setting this up. I created [this follow-up issue](https://github.com/FuelLabs/fuel-core/issues/2212) to use the `core-test.fuellabs.net` hostname in the test, since this PR got auto-merged after my approval.

Theghost256 (approved):

PR #2205, opened 2024-09-16
Use single heavy input in the `transaction_throughput.rs` benchmarks

xgreenx:
@SilentCicero 

>Benchmark Adjustment
As of the latest benchmarks for TPS on the FuelVM/Client, we seem to do benchmarks each with two signatures per transaction.
This is highly unusual for blockchain benchmarks, namely because there is usually only a single signature payload to verify per transaction in the models for all other blockchains.
We have conservatively, for almost no reason/benefit, halved the actual tps capability of Fuel in our public benchmarks by making each benchmark do two identical signature verifications.
This is because the signature payload being checked twice is identical and can be trivially cached and would be cached in an optimized real world transactional scenario. It would be far more realistic to do fuel client benchmarks with one signature predicate, one non signature predicate and two inputs and two outputs (one fee and one transfer), for realism.

### Before requesting review


Voxelot (approved):

benches/benches/transaction_throughput.rs
```diff
@@ -179,16 +181,22 @@ fn signed_transfers(c: &mut Criterion) {
...
+                Default::default(),
+                predicate.clone(),
+                vec![],
+            ))
             .add_output(Output::coin(rng.gen(), 50, AssetId::default()))
             .add_output(Output::change(rng.gen(), 0, AssetId::default()))
-            .finalize()
+            .finalize();
+        tx.estimate_predicates(&checked_parameters(), MemoryInstance::new())
+            .expect("Predicate check failed");
```
netrome, line 198: What's the purpose of estimating the predicates here?

benches/benches/transaction_throughput.rs
```diff
@@ -179,16 +181,22 @@ fn signed_transfers(c: &mut Criterion) {
...
+        tx
     };
     bench_txs("signed transfers", c, generator);
```
netrome, line 201: Should we rename this to "predicate transfer" now when we use a predicate instead of a signed coin input?

netrome (approved):
A few questions from me but otherwise this looks good.

benches/benches/transaction_throughput.rs
xgreenx: We don't use predicate instead. We use one signed coin along with one predicate. All benchmarks now use at least one predicate that returns `true`. The second input is different and the name reflects it

rafal-ch (approved):
LGTM :+1:

Issue #2062, opened 2024-08-08
Support dry running of transactions in the past


xgreenx:
~Requires first https://github.com/FuelLabs/fuel-core/issues/1995 feature to avoid error in the case of the `InMemory` storage.~

Scope draft:
* [X] Feature is "roughly" implemented on Green's branch (patch files + branch `hotfix/dry_run_may_return`)
* [X] Merge master into above 
* [ ] Write test which proves that a transaction can't be executed on the tip of the blockchain, but works on block from the past
* [ ] Consider enabling this feature by a flag different than `--debug` (_allow_historical_dry_run_, or smth)
* [ ] Option<Overrides> (?)
* [ ] Clean-up code and cover with tests (if needed)

PR #2206, opened 2024-09-16
feat: Use timestamp of last block when dry running transactions

### LINKED ISSUE #2128: Manually produces blocks in test and simulates a call the timestamp is not correct
Maybe related to https://github.com/FuelLabs/fuel-core/issues/2091

https://fuellabs.slack.com/archives/C01L1HUJSL8/p1724242020509839

netrome:
## Description
Closes #2128 

This PR adds the option to specify a timestamp in BlockProducer::dry_run. When no timestamp is provided, the block producer will use the time stamp of the previous block, instead of using `Tai64::now()` (previous behavior).

## Checklist

### Before requesting review


crates/services/producer/src/block_producer/tests.rs
```diff
@@ -543,6 +543,86 @@ mod produce_and_execute_block_txpool {
     }
 }
 
+// Tests for the `dry_run` method.
+mod dry_run {
+    use super::*;
+
+    #[tokio::test]
+    async fn dry_run__executes_with_given_timestamp() {
```
MitchTurner, line 551: What if time is earlier than the last block? Is that valid?

crates/services/producer/src/mocks.rs
```diff
@@ -227,6 +229,20 @@ impl BlockProducer<Vec<Transaction>> for MockExecutorWithCapture<Transaction> {
...
 
+impl DryRunner for MockExecutorWithCapture<Transaction> {
+    fn dry_run(
+        &self,
+        block: Components<Vec<Transaction>>,
+        _utxo_validation: Option<bool>,
+    ) -> ExecutorResult<Vec<TransactionExecutionStatus>> {
+        *self.captured.lock().unwrap() = Some(block);
+
+        Err(ExecutorError::Other(
```
xgreenx, line 240: Why do we want to return an error? It seems will be more clear to retune `Ok`, since you verify the timestamp from the `block`

crates/services/producer/src/block_producer/tests.rs
```diff
@@ -865,67 +956,49 @@ impl TestContextBuilder {
...
+        let block_gas_limit = self.block_gas_limit.unwrap_or_default();
 
         let mock_relayer = MockRelayer {
             latest_block_height: self.latest_block_height,
             latest_da_blocks_with_costs: self.blocks_with_gas_costs.clone(),
             ..Default::default()
         };
 
+        let db = MockDb {
+            blocks: self.pre_existing_blockss(),
```
xgreenx, line 994:
```suggestion
            blocks: self.pre_existing_blocks(),
```

tests/tests/tx.rs
```diff
@@ -252,6 +261,64 @@ async fn submit_and_await_status() {
...
+        .result
+    else {
+        panic!("unexpected execution result from dry run")
+    };
+
+    // Then
+    assert_eq!(
+        start_time.0
+            + block_production_interval_seconds
+                * number_of_blocks_to_produce_manually.saturating_sub(1) as u64,
```
xgreenx, line 317: Can we move it to a separate variable?

crates/services/producer/src/mocks.rs
netrome: My thinking here is that this mock **only provides a side channel to capture the block** right now, and it **does not simulate any execution**. Because of this, I prefer to return an error here to clarify for the caller that the simulated execution failed because it's not implemented. This should prevent any future misconceptions if someone uses this mock and assumes it actually simulates execution.

If you still prefer to have this return an `Ok` result, I'm also up for doing that. Let me know what you think.

crates/services/producer/src/block_producer/tests.rs
netrome: Hahah, sounds like a snake talking :sweat_smile: I'll fix. Thanks for spotting!

tests/tests/tx.rs
netrome: Sure thing, will do.

crates/services/producer/src/block_producer/tests.rs
netrome: Good question, I'd assume that it's valid because it would allow for testing contrafactuals. I.e. "what if this transaction had been submitted in the past, would it have been valid?". Though we do error on previous block heights (which we shouldn't do though https://github.com/FuelLabs/fuel-core/issues/2062).

netrome: I'll add a test case for it.

tests/tests/tx.rs
netrome: Done in https://github.com/FuelLabs/fuel-core/pull/2206/commits/c8f3ee1b540bf35a0cf2fc3cd8ed9fa48bb9223e

crates/services/producer/src/block_producer/tests.rs
netrome: Fixed in https://github.com/FuelLabs/fuel-core/pull/2206/commits/c8f3ee1b540bf35a0cf2fc3cd8ed9fa48bb9223e

crates/services/producer/src/block_producer/tests.rs
netrome: Added in https://github.com/FuelLabs/fuel-core/pull/2206/commits/c8f3ee1b540bf35a0cf2fc3cd8ed9fa48bb9223e and it works.

crates/services/producer/src/mocks.rs
MitchTurner: I think the `Err` is noise. It's harder to understand the tests since an error feels like it should be an edge case.

Admittedly, any "happy" value will also feel arbitrary, but since you aren't inspecting them it won't feel like extra context.

something like `arb_succeeding` instead of `default` might help communicate, but is probably overkill.

netrome: Alright, fair enough. I'll change to return an `Ok` result then.

netrome: Done in https://github.com/FuelLabs/fuel-core/pull/2206/commits/c3661969e331692399d064f2d446b2c4bd594dc3

crates/services/producer/src/block_producer.rs
```diff
@@ -268,26 +268,33 @@ where
...
+
+        let simulated_height = height.unwrap_or_else(|| {
+            latest_height
                 .succ()
                 .expect("It is impossible to overflow the current block height")
         });
 
-        let header = self._new_header(height, Tai64::now())?;
+        let simulated_time = time.unwrap_or_else(|| {
+            view.get_block(&latest_height)
```
rymnc, line 292: qq: should this use the latest height or the simulated height?

netrome: Simulated height might not have a block, so defaulting to latest makes sense here to me. If the caller specifies a block height, and not a block time it is likely that the thing they are testing isn't using the block time. If someone needs to simulate a block with a specific height and timestamp, they can provide both.

crates/services/producer/src/block_producer/tests.rs
```diff
@@ -543,6 +543,110 @@ mod produce_and_execute_block_txpool {
...
+        let simulated_block_time = Tai64::from_unix(1337);
+        let executor = MockExecutorWithCapture::default();
+        let ctx = TestContext::default_from_executor(executor.clone());
+
+        // When
+        let _err = ctx
+            .producer()
+            .dry_run(vec![], None, Some(simulated_block_time), None, None)
+            .await
+            .expect("dry run failed");
```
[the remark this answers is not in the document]
netrome: Since I changed the executor to return `Ok`, this is a normal `expect` so it should only panic if the dry run has failed. Though thinking about this we don't care about whether this returns an error or not here so I'll change this to just ignore the results instead.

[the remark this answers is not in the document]
netrome: Changed in https://github.com/FuelLabs/fuel-core/pull/2206/commits/24d509552fadf33412bb3bc488ebac738c839f96

crates/services/producer/src/block_producer/tests.rs
```diff
@@ -543,6 +543,110 @@ mod produce_and_execute_block_txpool {
...
+        let ctx = TestContextBuilder::new()
+            .with_prev_time(last_block_time)
+            .build_with_executor(executor.clone());
+
+        // When
+        let _err = ctx
+            .producer()
+            .dry_run(vec![], None, None, None, None)
+            .await
+            .expect("dry run failed");
```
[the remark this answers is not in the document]
netrome: Changed to ignore the result in https://github.com/FuelLabs/fuel-core/pull/2206/commits/24d509552fadf33412bb3bc488ebac738c839f96

crates/services/producer/src/block_producer/tests.rs
```diff
@@ -543,6 +543,110 @@ mod produce_and_execute_block_txpool {
...
+        let ctx = TestContextBuilder::new()
+            .with_prev_time(last_block_time)
+            .build_with_executor(executor.clone());
+
+        // When
+        let _err = ctx
+            .producer()
+            .dry_run(vec![], None, Some(simulated_block_time), None, None)
+            .await
+            .expect("dry run failed");
```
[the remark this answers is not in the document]
netrome: Changed to ignore the result in https://github.com/FuelLabs/fuel-core/pull/2206/commits/24d509552fadf33412bb3bc488ebac738c839f96

xgreenx (dismissed):

xgreenx (dismissed):

rymnc (dismissed):

rafal-ch (approved):
LGTM :+1:

crates/services/producer/src/block_producer.rs
```diff
@@ -268,26 +268,33 @@ where
...
+                .map(|block| block.header().time())
+                .unwrap_or(Tai64::UNIX_EPOCH)
+        });
+
+        let header = self._new_header(simulated_height, simulated_time)?;
```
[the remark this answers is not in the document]
netrome: Yeah I'm wondering the same thing :sweat_smile:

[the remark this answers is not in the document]
netrome: Seems to be to avoid a collision with the existing `new_header` function but yeah those methods could be named better.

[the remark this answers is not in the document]
netrome: Renamed to better reflect what the functions seem to be doing in https://github.com/FuelLabs/fuel-core/pull/2206/commits/98c32af83bedc1135ec204ab25235241108c44b4
```
new_header -> new_header_with_new_da_height
_new_header -> new_header
```

rafal-ch (approved):

MitchTurner (approved):

Issue #2187, opened 2024-09-11
Add all mechanisms needed for the internal modules to connect to TXPoolV2


AurelienFT:
- Block importer
- P2P
- Block producer

xgreenx:
Done in https://github.com/FuelLabs/fuel-core/pull/2263

PR #2216, opened 2024-09-18
Txpool v2 features for futures connections with other modules

AurelienFT:
REVIEWER NOTE: Review https://github.com/FuelLabs/fuel-core/pull/2193 first for context
## Linked Issues/PRs
Closes https://github.com/FuelLabs/fuel-core/issues/2187

## Description

(2k of additions is the tests from TxPool v1 which has been instantiated again)

This PR adds all the logic for the service and shared state of the TxPool to be able to be used by all the others modules and to make all the auxiliary mechanisms working.

The PR changes the way the `SharedState` interact with the pool and the service. Now the `SharedState` only sends commands to the service that will answer him whenever he is time to do depending on the priority of the request (managed through a tokio select biaised). This makes all methods to return async even `find` which I'm not a big fan of but didn't found any better solution. 
**EDIT: I have made some tests especially tries to run `execute_suite` tests and it seems that blocking can cause "double blocking forever" and so I'm up for an alternative**

Here is a list of all interactions made by the different modules and their new name/behavior/reason : 

### Block producer
| Existing | New |
| - | - |
| select_transactions(max_gas) | select_transactions(max_gas) |

### PoA
| Existing | New |
| - | - |
| total_consumable_gas() | Nothing (unused) |
| transaction_status_events() | Replaced by new_txs_notifier [reason](https://github.com/FuelLabs/fuel-core/blob/d5ba082264e1e3db54c6c81f6036eb06b9d3044c/crates/services/consensus_module/poa/src/service.rs#L590) |
| pending_number() | Not used anymore after transaction_status_events() change |
| remove_txs() | broadcast_txs_skipped_reason() |

### Block Importer
| Existing | New |
| - | - |
| Stream with blockEvents | Stream with block events |

### P2P
| Existing | New |
| - | - |
| gossiped_tx_stream | tx_from_p2p_stream |
| new_tx_gossip_subscription | new_peers_subscribed_stream |
| broadcast_transaction | broadcast_transaction |
| get_tx_ids| get_tx_ids |
| get_txs | get_txs |
| notify_gossip_transaction_validity | notify_gossip_transaction_validity |

### GraphQL API

| Existing | New |
| - | - |
| transaction | transaction |
| tx_update_subscribe | tx_update_subscribe |

### Others changes

- The PR also changes the error type to help separate subdomains of errors and try to be more clear about the moment errors can happens.
- The TTL management of the transactions is also in place as the less priority task for the service.
- A new thread pool has been added to manage P2P sync that could take a long time (same behavior as txpool v1)
- The transactions are now stored as Arc because they need to be shareable with others modules (API, P2P)
- All the tests from the previous pool has been instantiated and pass. 
- The TxPool don't have anymore unused authorization and match the correct linting requirements.

## Checklist

### Before requesting review


AurelienFT:
Making this draft given the conflicts.

AurelienFT:
Readyf ro review again

AurelienFT:
Change this to draft to add gas price and manage delete of skipped transactions dependents

AurelienFT:
@xgreenx I have added the check of gas price in selection and the remove of coin dependents on skipped txs.

xgreenx (approved):

crates/services/txpool_v2/src/service.rs
```diff
@@ -1,234 +1,157 @@
...
+pub struct Task<View> {
+    chain_id: ChainId,
+    subscriptions: Subscriptions,
+    verification: Verification<View>,
+    p2p: Arc<dyn P2PRequests>,
+    transaction_verifier_process: SyncProcessor,
+    p2p_sync_process: AsyncProcessor,
+    pruner: TransactionPruner,
+    pool: Shared<TxPool>,
+    current_height: Arc<RwLock<BlockHeight>>,
```
acerone85, line 141: is there any reason why you use the `Shared` type alias for pool and not for current_height? 
 Overall I think `Arc<RwLock<_>>` is more clear

crates/services/txpool_v2/src/service.rs
```diff
@@ -1,234 +1,157 @@
...
-        ConsensusParamsProvider,
-        GasPriceProvider,
-        WasmChecker,
-        MemoryPool,
-    >,
+pub struct Task<View> {
+    chain_id: ChainId,
+    subscriptions: Subscriptions,
+    verification: Verification<View>,
+    p2p: Arc<dyn P2PRequests>,
```
netrome, line 136: Do we need to hold a trait object here? Can't we use a generic and save a heap allocation?

```suggestion
pub struct Task<View. P2PRequests> {
    chain_id: ChainId,
    subscriptions: Subscriptions,
    verification: Verification<View>,
    p2p: P2PRequests,
```

AurelienFT: The choice has been made to reduce the generics and use the Arc even if it's less performant because they are mainly used in tasks that consume a lot and so they will not be our bottleneck. + we win code readability

crates/services/txpool_v2/src/service.rs
```diff
@@ -1,234 +1,157 @@
...
-    persistent_storage_provider: Arc<PSProvider>,
-    current_height: Arc<RwLock<BlockHeight>>,
-    consensus_parameters_provider: Arc<ConsensusParamsProvider>,
-    gas_price_provider: Arc<GasPriceProvider>,
-    wasm_checker: Arc<WasmChecker>,
-    memory: Arc<MemoryPool>,
-    heavy_async_processor: Arc<HeavyAsyncProcessor>,
-    utxo_validation: bool,
-}
+pub type Shared<T> = Arc<RwLock<T>>;
```
netrome, line 99: While this is a pretty nice type-alias, I don't see any point in making it public. Any other modules that would benefit from the type alias can redefine it, which would be more straight-forward to read and just as concise as importing it without the extra indirection.

```suggestion
type Shared<T> = Arc<RwLock<T>>;
```

crates/services/txpool_v2/src/service/verifications.rs
```diff
@@ -0,0 +1,303 @@
...
+    error::Error,
+    ports::{
+        ConsensusParametersProvider,
+        GasPriceProvider,
+        TxPoolPersistentStorage,
+        WasmChecker,
+    },
+    service::{
+        memory::MemoryPool,
+        Shared,
```
netrome, line 11: For example here I would find it more readable if we instead of importing `Shared` redefine the alias in this module.

crates/services/txpool_v2/src/service.rs
netrome: Optionally if this is used in a lot of places we should move it out to another module, since the alias has nothing to do with this particular service.

netrome (commented):
Some opinions on the `Shared` type, continuing my review...

crates/services/src/lib.rs
```diff
@@ -74,3 +78,10 @@ pub use sync::{
     Shared,
     SharedMutex,
 };
+#[cfg(feature = "sync-processor")]
+pub use sync_processor::SyncProcessor;
+
+// For tests
+use crate as fuel_core_services;
+#[allow(unused_imports)]
+use fuel_core_services as _;
```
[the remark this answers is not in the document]
AurelienFT: @xgreenx

crates/services/txpool_v2/src/service.rs
AurelienFT: I changed it to pub(crate) because it's used in the tests

crates/services/txpool_v2/src/service/verifications.rs
AurelienFT: will let @xgreenx  give his advice on this as it's a type he added

crates/services/src/lib.rs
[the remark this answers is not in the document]
xgreenx: image


It was done to enable feature by default during testing

crates/services/txpool_v2/src/service/verifications.rs
xgreenx: Up to @AurelienFT )

AurelienFT: Noted on my note will think about it in the next days to not block the merge.

acerone85 (approved):

xgreenx (approved):

crates/services/txpool_v2/src/service.rs
netrome: Makes sense :+1:

crates/services/txpool_v2/src/service.rs
```diff
@@ -246,36 +169,81 @@ where
...
             _ = watcher.while_started() => {
                 should_continue = false;
             }
+
+            block_result = self.subscriptions.imported_blocks.next() => {
+                if let Some(result) = block_result {
+                    self.import_block(result);
+                    should_continue = true;
+                } else {
+                    should_continue = false;
```
netrome, line 190: I know the pattern was established before this PR, but it would make more sense to me to just return here instead of declaring and assigning the variable to later return it.

```suggestion
                    return Ok(true);
                } else {
                    return Ok(false);
```

(and the same for all other branches)

netrome (commented):
Another suggestion (not blocking merge)

crates/services/txpool_v2/src/service.rs
netrome: Nevermind, I see you already merged! 🎉

AurelienFT: It's Ok you can leave comment and I will address in patch in the future. I note this and let this open in the meantime

crates/services/txpool_v2/src/service.rs
```diff
@@ -284,55 +252,471 @@ where
...
-    PSProvider: AtomicView<LatestView = PSView> + Clone + Send + Sync,
+    P2P: P2PSubscriptions<GossipedTransaction = TransactionGossipData>,
+    P2P: P2PRequests + Send + Sync + 'static,
+    PSProvider: AtomicView<LatestView = PSView> + Send + Sync + 'static,
     PSView: TxPoolPersistentStorage,
     ConsensusParamsProvider: ConsensusParametersProvider + Send + Sync,
     GasPriceProvider: GasPriceProviderTrait + Send + Sync,
     WasmChecker: WasmCheckerTrait + Send + Sync,
-    MemoryPool: MemoryPoolTrait + Send + Sync,
+    BlockImporter: BlockImporterTrait + Send + Sync,
```
rafal-ch, line 622: These trait bounds look too restrictive, please check if the `Send + Sync` are needed. Also unsure about `'static`.

crates/services/txpool_v2/src/pool.rs
```diff
@@ -224,30 +227,65 @@ where
...
+                        tracing::warn!(
+                            "Dependent storage data not found for \
+                            the transaction during `remove_transaction`."
+                        );
+                        continue
+                    };
+                    self.selection_algorithm
+                        .new_executable_transaction(dependent, storage_data);
+                }
+                self.update_components_and_caches_on_removal(iter::once(&transaction));
```
rafal-ch, line 286: Just noticed that all tests in `crates/services/txpool_v2/src/tests/mod.rs` are still green if I remove these lines. Seems like tests don't ensure that we promote dependents to executables.

Maybe there are tests for that in different place.

crates/services/txpool_v2/src/service.rs
AurelienFT: Static is necessary on the persistent storage because we create an Arc from this. Removed the one from PSProvider.
For the `Sync` and `Send` they were added in `ports.rs` and don't removed from here. I have removed them from here, ty :) They all have `Send` `Sync` and `'static` in `ports.rs` to be usable in arcs and in the future that is defined in `async fn run`.

Changes are available here : https://github.com/FuelLabs/fuel-core/pull/2263/commits/1c345c0bc7799e8e64cece36261f105d896323af

crates/services/txpool_v2/src/pool.rs
```diff
@@ -224,30 +227,65 @@ where
...
+
+    pub fn iter_tx_ids(&self) -> impl Iterator<Item = &TxId> {
+        self.tx_id_to_storage_id.keys()
+    }
+
+    /// Remove transaction but keep its dependents.
+    /// The dependents become executables.
+    pub fn remove_transaction(&mut self, tx_ids: Vec<TxId>) {
+        for tx_id in tx_ids {
+            if let Some(storage_id) = self.tx_id_to_storage_id.remove(&tx_id) {
```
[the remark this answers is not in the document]
AurelienFT: We are ok to try to delete the others if they are not already been removed yet. I'm not sure about the warning because this function can be called from a lot of differents place/modules around the code and even if it's suboptimal to call again this function (with an other module for example) on already deleted functions I don't think it's worth warning to the user about this has it will not make his node gone badly or break anything.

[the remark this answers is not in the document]
rafal-ch: If this doesn't break anything there's no need to log, indeed.

Thanks for explanation! :+1:

Issue #2203, opened 2024-09-16
Add `worst_case` function to V1 algorithm


MitchTurner:
It will work just like the `worst_case` on V0, but with the `da_gas_price` calculated as well

PR #2219, opened 2024-09-18
Implement worst case scenario for price algorithm v1

rafal-ch:
## Linked Issues/PRs
Closes https://github.com/FuelLabs/fuel-core/issues/2203

## Description
This PR adds the `worst_case()` calculation to the `V1` price algorithm.

## Checklist

### Before requesting review


crates/fuel-gas-price-algorithm/src/lib.rs
```diff
@@ -4,3 +4,24 @@
 
 pub mod v0;
 pub mod v1;
+
+#[allow(clippy::cast_possible_truncation)]
+pub(crate) fn cumulative_percentage(
```
MitchTurner, line 9: I think we should name this "cumulative_percentage_change" or something like that instead

rafal-ch: Updated in https://github.com/FuelLabs/fuel-core/pull/2219/commits/adffd995e4949ba2e025d88ea6c51660187e1d7d

MitchTurner (dismissed):

crates/fuel-gas-price-algorithm/src/lib.rs
```diff
@@ -4,3 +4,24 @@
...
+    let blocks = height.saturating_sub(for_height) as f64;
+    let percentage_as_decimal = percentage as f64 / 100.0;
+    let multiple = (1.0f64 + percentage_as_decimal).powf(blocks);
+    let mut approx = new_exec_price as f64 * multiple;
+    // account for rounding errors and take a slightly higher value
+    const ARB_CUTOFF: f64 = 16948547188989277.0;
+    if approx > ARB_CUTOFF {
+        const ARB_ADDITION: f64 = 2000.0;
+        approx += ARB_ADDITION;
+    }
```
Dentosal, line 24: I don't understand what happens here. What's the meaning of `ARB_CUTOFF`?

MitchTurner: It's a hack for dealing with really large floats. The rounding error will be on the order of `2000` and based on the prop tests will always result in a value that is larger than the `actual` worst case value. 

So `ARB_CUTOFF` is where we start seeing the error and `ARB_ADDITION` is the the amount we add to account for that error. This is the cost of using floats to represent these values, maybe there is a cleaner way to represent it :).

netrome: It doesn't sound arbitrary then. Can we rename it to something like this?
```suggestion
    // account for rounding errors and take a slightly higher value
    const ROUNDING_ERROR_CUTOFF: f64 = 16948547188989277.0;
    if approx > ROUNDING_ERROR_CUTOFF {
        const ROUNDING_ERROR_COMPENSATION: f64 = 2000.0;
        approx += ROUNDING_ERROR_COMPENSATION;
    }
```

MitchTurner: Background:

https://github.com/FuelLabs/fuel-core/pull/2025#discussion_r1681725245

crates/fuel-gas-price-algorithm/src/lib.rs
```diff
@@ -4,3 +4,24 @@
 
 pub mod v0;
 pub mod v1;
+
+#[allow(clippy::cast_possible_truncation)]
+pub(crate) fn cumulative_percentage_change(
```
netrome, line 9: It feels unclean to have this function on the crate root. How about putting it in a `utils` module or something similar?

crates/fuel-gas-price-algorithm/src/v1/tests/algorithm_v1_tests.rs
```diff
@@ -28,3 +32,60 @@ fn calculate__returns_sum_of_da_and_exec_gas_prices() {
...
+
+    let expected = expected_exec_price.saturating_add(expected_da_gas_price);
+
+    dbg!(actual, expected);
+    assert!(actual >= expected);
+}
+
+proptest! {
+    #[test]
+    fn worst_case__correctly_calculates_value(
```
MitchTurner, line 81: We should add another test to check for overflows. See the V0 test and this convo:
https://github.com/FuelLabs/fuel-core/pull/2025#discussion_r1683965151

netrome (dismissed):
Some suggestions, otherwise LGTM

crates/fuel-gas-price-algorithm/src/lib.rs
MitchTurner: > It doesn't sound arbitrary then. Can we rename it to something like this?

There isn't a "correct" value to choose here, that's why I called it `ARB`, but you're right that the reason we chose it isn't completely arbitrary and your suggestion is much clearer and helpful anyway.

rafal-ch: Variable names updated in https://github.com/FuelLabs/fuel-core/pull/2219/commits/4337b14502861275c52aabedc4d9396b6f9e5147

crates/fuel-gas-price-algorithm/src/lib.rs
rafal-ch: Updated in https://github.com/FuelLabs/fuel-core/pull/2219/commits/d1672d32199c4d29da6c2430bfef9f1d47937e8c

crates/fuel-gas-price-algorithm/src/v1/tests/algorithm_v1_tests.rs
rafal-ch: Test added in https://github.com/FuelLabs/fuel-core/pull/2219/commits/6cb31617facd5b55ed99b649146c2750654eb133

crates/fuel-gas-price-algorithm/src/v1/tests/algorithm_v1_tests.rs
```diff
@@ -28,3 +32,114 @@ fn calculate__returns_sum_of_da_and_exec_gas_prices() {
...
+        let target_height = starting_height.saturating_add(block_horizon);
+        let _ = algorithm.worst_case(target_height);
+
+        // then
+        // doesn't panic with an overflow
+    }
+}
+
+#[test]
+fn worst_case__same_block_gives_new_exec_price() {
```
MitchTurner, line 122: This name is wrong, it's misssing the da portion in the name. I think instead just saying it gives the same as calculate() would be best, for both the name and how it calculates the expected value.

rafal-ch: Names updated in https://github.com/FuelLabs/fuel-core/pull/2219/commits/c55aa6569f1a001544c96562dd4fd16561b9be26

MitchTurner (approved):

netrome (approved):

PR #2224, opened 2024-09-19
chore(gas_price_service): define port for L2 data

rymnc:
> [!NOTE]
> This is PR 1/n in cleaning up the gas price service. This PR is first of few that moves the `algorithm_updater` from the `sub_services` module of `fuel-core` to `fuel-core-gas-price-service`.
>
> To do so, we have to cut dependencies on `fuel-core`, which means we have to remove the direct dependency on `Database`, `ConsensusParametersProvider` etc.

## Description

We define a new port in the `fuel-core-gas-price-service` crate that is meant to cut the dependency of `algorithm_updater` from `fuel-core`.

## Checklist
- [x] New behavior is reflected in tests (existing tests already work)

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/fuel-core/src/service/adapters/gas_price_adapters.rs
+++ b/crates/fuel-core/src/service/adapters/gas_price_adapters.rs
@@ -24,14 +24,10 @@ use fuel_core_storage::{
 };
 use fuel_core_types::{
     blockchain::{
-        block::{
-            Block,
-        },
+        block::Block,
         header::ConsensusParametersVersion,
     },
-    fuel_tx::{
-        Transaction,
-    },
+    fuel_tx::Transaction,
     fuel_types::BlockHeight,
 };
 
--- a/crates/fuel-core/src/service/adapters/gas_price_adapters.rs
+++ b/crates/fuel-core/src/service/adapters/gas_price_adapters.rs
@@ -13,15 +13,7 @@ use fuel_core_gas_price_service::{
     },
     ports::L2Data,
 };
-use fuel_core_storage::{
-    not_found,
-    tables::{
-        FuelBlocks,
-        Transactions,
-    },
-    Result as StorageResult,
-    StorageAsRef,
-};
+use fuel_core_storage::Result as StorageResult;
 use fuel_core_types::{
     blockchain::{
         block::Block,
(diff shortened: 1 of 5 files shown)


MitchTurner (dismissed):
LGTM

crates/fuel-core/src/service/sub_services/algorithm_updater.rs
```diff
@@ -290,31 +298,33 @@ fn sync_metadata_storage_with_on_chain_storage(
...
-            .storage::<FuelBlocks>()
-            .get(&height.into())?
+            .get_block(&height.into())?
             .ok_or(anyhow::anyhow!("Expected block to exist"))?;
         let last_tx_id = block.transactions().last().ok_or(anyhow::anyhow!(
             "Expected block to have at least one transaction"
         ))?;
         let param_version = block.header().consensus_parameters_version;
         let params = settings.settings(&param_version)?;
         let mint = view
```
xgreenx, line 326: Hmm, the logic below `let block_gas_used = mint.mint_amount();` looks incorrect. In another places we calculate this value from several paramterers.

image

I think we don't need ability to fetch the last transaction. I think we need to process the whole block as we do in the service itself. Sofrom port perspective we need only `latest_height` and `get_block` where `get_block` return whole block with all transactions.

It is an expensive operation than just fetching last transaction, but it will avoid potential issue and code will be easier to read, since ew use the same update logic during initialization, during `run` function and during `shutdown`.

rymnc: addressed in d48e3744b8, qq: (maybe @MitchTurner) what behaviour should we have for this line? - https://github.com/FuelLabs/fuel-core/blob/d48e3744b84b59dc400482d9fb88d04cfe89e9e0/crates/fuel-core/src/service/sub_services/algorithm_updater.rs#L332

I marked it as an error, since we have the `saturating_add` above.

crates/fuel-core/src/service/adapters/gas_price_adapters.rs
```diff
@@ -1,17 +1,64 @@
...
+        for tx_id in block.transactions() {
+            let tx = self
+                .storage::<Transactions>()
+                .get(tx_id)
+                .map(|tx| tx.map(|tx| tx.as_ref().clone()))?
+                .ok_or(not_found!("Transaction"))?;
+            transactions.push(tx);
+        }
+        Ok(block.uncompress(transactions))
+    }
```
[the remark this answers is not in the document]
rymnc: addressed in abbf480bf3

crates/fuel-core/src/service/sub_services/algorithm_updater.rs
MitchTurner: Yeah. That seems fine to me.

MitchTurner (approved):

crates/fuel-core/src/service/sub_services/algorithm_updater.rs
```diff
@@ -290,41 +301,47 @@ fn sync_metadata_storage_with_on_chain_storage(
...
         let block = view
-            .storage::<FuelBlocks>()
-            .get(&height.into())?
-            .ok_or(anyhow::anyhow!("Expected block to exist"))?;
-        let last_tx_id = block.transactions().last().ok_or(anyhow::anyhow!(
-            "Expected block to have at least one transaction"
-        ))?;
+            .get_block(&height.into())?
+            .ok_or(not_found!("FullBlock"))?;
         let param_version = block.header().consensus_parameters_version;
```
xgreenx, line 324: When `InitializeTask` will be moved into `gas-price-service`, instead of duplicating logic we can call `GasPriceTaskV0::process_l2_block(block)` inside of `InitializeTask::into_task`. Just a reminder to not forget

xgreenx (approved):

crates/fuel-core/src/service/sub_services/algorithm_updater.rs
rymnc: yes, that's on the roadmap :)

PR #2226, opened 2024-09-20
chore(gas_price_service): define port for GasPriceDatabase

rymnc:
> [!NOTE]
> This is PR 2/n in cleaning up the gas price service. This PR is first of few that moves the `algorithm_updater` from the `sub_services` module of `fuel-core` to `fuel-core-gas-price-service`. 
>
> Please review https://github.com/FuelLabs/fuel-core/pull/2224 before getting to this one ;) 

- **chore(gas_price_service): define port for gas price db lookups**

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/fuel-core/src/combined_database.rs
+++ b/crates/fuel-core/src/combined_database.rs
@@ -311,6 +311,18 @@ impl CombinedDatabase {
 
         Ok(())
     }
+
+    pub fn sync_aux_db_heights<S>(&self, shutdown_listener: &mut S) -> anyhow::Result<()>
+    where
+        S: ShutdownListener,
+    {
+        let on_chain_height = self
+            .on_chain()
+            .latest_height_from_metadata()?
+            .ok_or(anyhow::anyhow!("on-chain database doesn't have height"))?;
+
+        self.rollback_to(on_chain_height, shutdown_listener)
+    }
 }
 
 /// A trait for listening to shutdown signals.
--- a/crates/fuel-core/src/combined_database.rs
+++ b/crates/fuel-core/src/combined_database.rs
@@ -316,12 +316,11 @@ impl CombinedDatabase {
     where
         S: ShutdownListener,
     {
-        let on_chain_height = self
-            .on_chain()
-            .latest_height_from_metadata()?
-            .ok_or(anyhow::anyhow!("on-chain database doesn't have height"))?;
+        if let Some(on_chain_height) = self.on_chain().latest_height_from_metadata()? {
+            self.rollback_to(on_chain_height, shutdown_listener)?;
+        };
 
-        self.rollback_to(on_chain_height, shutdown_listener)
+        Ok(())
     }
 }
 
--- a/crates/fuel-core/src/combined_database.rs
+++ b/crates/fuel-core/src/combined_database.rs
@@ -317,7 +317,11 @@ impl CombinedDatabase {
         S: ShutdownListener,
     {
         if let Some(on_chain_height) = self.on_chain().latest_height_from_metadata()? {
-            let _ = self.rollback_to(on_chain_height, shutdown_listener);
+            // todo(https://github.com/FuelLabs/fuel-core/issues/2239): This is a temporary fix
+            let res = self.rollback_to(on_chain_height, shutdown_listener);
+            if res.is_err() {
+                tracing::warn!("Failed to rollback auxiliary databases to on-chain database height: {:?}", res);
+            }
         };
 
         Ok(())
--- a/crates/fuel-core/src/combined_database.rs
+++ b/crates/fuel-core/src/combined_database.rs
@@ -317,7 +317,7 @@ impl CombinedDatabase {
         S: ShutdownListener,
     {
         if let Some(on_chain_height) = self.on_chain().latest_height_from_metadata()? {
-            self.rollback_to(on_chain_height, shutdown_listener)?;
+            let _ = self.rollback_to(on_chain_height, shutdown_listener);
         };
 
         Ok(())
--- a/crates/fuel-core/src/service/adapters/gas_price_adapters.rs
+++ b/crates/fuel-core/src/service/adapters/gas_price_adapters.rs
@@ -67,11 +67,11 @@ impl GasPriceData for Database<GasPriceDatabase> {
             .map(|metadata| metadata.map(|metadata| metadata.as_ref().clone()))
     }
 
-    fn set_metadata(&mut self, metadata: UpdaterMetadata) -> StorageResult<()> {
+    fn set_metadata(&mut self, metadata: &UpdaterMetadata) -> StorageResult<()> {
         let height = metadata.l2_block_height();
         let mut tx = self.write_transaction();
         tx.storage_as_mut::<GasPriceMetadata>()
-            .insert(&height, &metadata)?;
+            .insert(&height, metadata)?;
         tx.commit()?;
         Ok(())
     }
--- a/crates/fuel-core/src/service/adapters/gas_price_adapters.rs
+++ b/crates/fuel-core/src/service/adapters/gas_price_adapters.rs
@@ -16,6 +16,7 @@ use fuel_core_gas_price_service::{
     ports::{
         GasPriceData,
         L2Data,
+        MetadataStorage,
     },
 };
 use fuel_core_storage::{
@@ -58,30 +59,39 @@ impl L2Data for OnChainIterableKeyValueView {
 }
 
 impl GasPriceData for Database<GasPriceDatabase> {
-    fn get_metadata(
-        &self,
-        block_height: &BlockHeight,
-    ) -> StorageResult<Option<UpdaterMetadata>> {
-        self.storage::<GasPriceMetadata>()
-            .get(block_height)
-            .map(|metadata| metadata.map(|metadata| metadata.as_ref().clone()))
-    }
-
-    fn set_metadata(&mut self, metadata: &UpdaterMetadata) -> StorageResult<()> {
-        let height = metadata.l2_block_height();
-        let mut tx = self.write_transaction();
-        tx.storage_as_mut::<GasPriceMetadata>()
-            .insert(&height, metadata)?;
-        tx.commit()?;
-        Ok(())
-    }
-
     fn latest_height(&self) -> Option<BlockHeight> {
         HistoricalView::latest_height(self)
     }
+}
 
-    fn rollback_last_block(&self) -> StorageResult<()> {
-        self.rollback_last_block()
+impl MetadataStorage for Database<GasPriceDatabase> {
+    fn get_metadata(
+        &self,
+        block_height: &BlockHeight,
+    ) -> GasPriceResult<Option<UpdaterMetadata>> {
+        let metadata = self
+            .storage::<GasPriceMetadata>()
+            .get(block_height)
+            .map_err(|err| GasPriceError::CouldNotFetchMetadata {
+                source_error: err.into(),
+            })?;
+        Ok(metadata.map(|inner| inner.into_owned()))
+    }
+
+    fn set_metadata(&mut self, metadata: &UpdaterMetadata) -> GasPriceResult<()> {
+        let block_height = metadata.l2_block_height();
+        let mut tx = self.write_transaction();
+        tx.storage_as_mut::<GasPriceMetadata>()
+            .insert(&block_height, metadata)
+            .map_err(|err| GasPriceError::CouldNotSetMetadata {
+                block_height,
+                source_error: err.into(),
+            })?;
+        tx.commit().map_err(|err| GasPriceError::CouldNotSetMetadata {
+            block_height,
+            source_error: err.into(),
+        })?;
+        Ok(())
     }
 }
 
--- a/crates/fuel-core/src/service/adapters/gas_price_adapters.rs
+++ b/crates/fuel-core/src/service/adapters/gas_price_adapters.rs
@@ -5,15 +5,28 @@ use crate::{
 use fuel_core_gas_price_service::{
     fuel_gas_price_updater::{
         fuel_core_storage_adapter::{
+            storage::GasPriceMetadata,
             GasPriceSettings,
             GasPriceSettingsProvider,
         },
         Error as GasPriceError,
         Result as GasPriceResult,
+        UpdaterMetadata,
     },
-    ports::L2Data,
+    ports::{
+        GasPriceData,
+        L2Data,
+    },
 };
-use fuel_core_storage::Result as StorageResult;
+use fuel_core_storage::{
+    transactional::{
+        HistoricalView,
+        WriteTransaction,
+    },
+    Result as StorageResult,
+    StorageAsMut,
+    StorageAsRef,
+};
 use fuel_core_types::{
     blockchain::{
         block::Block,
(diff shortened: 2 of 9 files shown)


crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_core_storage_adapter/metadata_tests.rs
```diff
@@ -36,23 +40,30 @@ fn database() -> StorageTransaction<InMemoryStorage<GasPriceColumn>> {
     InMemoryStorage::default().into_transaction()
 }
```
rymnc, line 41: hey @MitchTurner do you know why we return a `StorageTransaction` here?

MitchTurner: IIRC just because it met the requirements and was simple to work with. I don't think it's particularly important or like we're attached to it.

`Database` itself is a `fuel-core` concept, so we don't have access to it from `gas_price_service`.

MitchTurner (dismissed):

MitchTurner:
I guess this is still in draft. So maybe shouldn't have improved, but since this is all being done incrementally... I'm good with what I'm seeing.

netrome (commented):
Looks good to me. Just two todos that are missing a reference to an issue.

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_core_storage_adapter.rs
```diff
@@ -51,41 +34,6 @@ mod l2_source_tests;
 
 pub mod storage;
 
-impl<Storage> MetadataStorage for StructuredStorage<Storage>
```
xgreenx, line 54: The main idea of this code was that gas price crate defined default implementation of the `MetadataStorage` port. I don't see why we don't want to follow that idea=)

xgreenx: Having default implementation allows us test it within this crate, and inherit it later in the `fuel-core`, if we want to, of course=)

crates/fuel-core/src/service/adapters/gas_price_adapters.rs
```diff
@@ -39,6 +57,34 @@ impl L2Data for OnChainIterableKeyValueView {
...
+            .insert(&height, metadata)?;
+        tx.commit()?;
+        Ok(())
+    }
+
+    fn latest_height(&self) -> Option<BlockHeight> {
+        HistoricalView::latest_height(self)
+    }
+
+    fn rollback_last_block(&self) -> StorageResult<()> {
```
xgreenx, line 83: I think we don't need to leak this kind of functionality to the service.

I think it would be better if `FuelService` were responsible for database integrity. It is okay if other databases are behind than on-chain database. It is the responsibility of the service to catch up(since only the service knows how to advance the state). But if the on-chain database is behind other databases, then `FuelService` should roll back them since only `FuelService` knows that the database supports the state rewind feature.

MitchTurner: Hmmmm. Kinda confused. InitializationTask currently handles that and I thought we wanted to get rid of it and move that code into `GasPriceService`.

I think @xgreenx  right that it's leaking something, but at the same time we have to leak the idea that they can be out of sync to include the syncing logic into `GasPriceService`.

MitchTurner: i.e. I'm not sure how the `FuelService` could handle it without accessing the metadata and updater code.

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_core_storage_adapter.rs
rymnc: retained in [100b5b8](https://github.com/FuelLabs/fuel-core/pull/2226/commits/100b5b86e092e00e466befb1070d570fa0935f8e)

crates/fuel-core/src/service/adapters/gas_price_adapters.rs
rymnc: I think @xgreenx here means that the `FuelService` *should* be aware of rollback strategy for each of the db's included within `CombinedDatabase`, which isn't that complex at the moment where we're leaking business logic, its just rolling back to the same height as the `OnChainDb`

rymnc: addressed in 9575a21b48
seems cleaner than before :)

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_core_storage_adapter/metadata_tests.rs
```diff
@@ -36,23 +40,30 @@ fn database() -> StorageTransaction<InMemoryStorage<GasPriceColumn>> {
...
+            .insert(&metadata.l2_block_height(), metadata)?;
+        self.commit_changes(self.changes().clone())?;
+        Ok(())
+    }
 
-    // then
-    let expected = Some(metadata);
-    assert_eq!(expected, actual);
+    fn latest_height(&self) -> Option<BlockHeight> {
+        todo!()
```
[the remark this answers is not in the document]
rymnc: outdated, this code was pruned :)

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_core_storage_adapter/metadata_tests.rs
```diff
@@ -36,23 +40,30 @@ fn database() -> StorageTransaction<InMemoryStorage<GasPriceColumn>> {
...
+    }
+
+    fn rollback_last_block(&self) -> StorageResult<()> {
+        todo!()
```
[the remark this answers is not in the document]
rymnc: outdated, this code was pruned :)

crates/fuel-core/src/service/adapters/gas_price_adapters.rs
rymnc: slightly unsure about where you'd like to place the rewind/rollback call ~ i've placed it before the `init_sub_services` call :)

crates/fuel-core/src/combined_database.rs
```diff
@@ -317,7 +317,7 @@ impl CombinedDatabase {
         S: ShutdownListener,
     {
         if let Some(on_chain_height) = self.on_chain().latest_height_from_metadata()? {
-            self.rollback_to(on_chain_height, shutdown_listener)?;
+            let _ = self.rollback_to(on_chain_height, shutdown_listener);
```
rymnc, line 320: https://github.com/FuelLabs/fuel-core/issues/2239

rymnc: definitely don't like this

netrome: Why can't you return the error here? Or at least log it?

netrome: Okay I see there are a few small cases to consider. It would be nice with a todo link referencing this from the code as well.

rymnc: addressed in 295301924b

netrome: Thank you! :pray:

crates/services/gas_price_service/src/ports.rs
```diff
@@ -12,3 +16,13 @@ pub trait L2Data: Send + Sync {
...
     ) -> StorageResult<Option<Block<Transaction>>>;
 }
+
+pub trait MetadataStorage: Send + Sync {
+    fn get_metadata(&self, block_height: &BlockHeight)
+        -> Result<Option<UpdaterMetadata>>;
+    fn set_metadata(&mut self, metadata: &UpdaterMetadata) -> Result<()>;
+}
+
+pub trait GasPriceData: Send + Sync {
```
acerone85, line 26: the name of the port is `GasPriceData`, but its only function returns a `BlockHeight`.

Would it be possible to have a comment that explains how the two are related?

crates/services/gas_price_service/src/fuel_gas_price_updater/tests.rs
```diff
@@ -38,15 +38,13 @@ impl FakeMetadata {
 }
 
 impl MetadataStorage for FakeMetadata {
-    fn get_metadata(
-        &self,
-        _block_height: &BlockHeight,
-    ) -> Result<Option<UpdaterMetadata>> {
-        Ok(self.inner.lock().unwrap().clone())
+    fn get_metadata(&self, _: &BlockHeight) -> Result<Option<UpdaterMetadata>> {
```
acerone85, line 41: maybe not related to this MR, but is there any reason why FakeMetadata::inner is wrapped around an `Arc`? Mutex implements `Send and Sync` already, and

crates/fuel-core/src/service/sub_services/algorithm_updater.rs
```diff
@@ -189,18 +190,22 @@ where
...
     da_block_costs: DaBlockCostsSharedState,
-) -> anyhow::Result<Updater>
+) -> anyhow::Result<Updater<GasPriceStore>>
 where
     L2DataStore: L2Data,
     L2DataStoreView: AtomicView<LatestView = L2DataStore>,
+    GasPriceStore: GasPriceData
+        + MetadataStorage
+        + KeyValueInspect<Column = GasPriceColumn>
+        + Modifiable,
```
MitchTurner, line 208: We should be careful about these all being bound together. Obviously this is still in an intermediate state, so maybe goes without saying, but ultimately the different ports should be handled independently. If they are the same thing under the hood, that shouldn't be leaked to like this.

crates/fuel-core/src/service/sub_services/algorithm_updater.rs
```diff
@@ -339,23 +335,8 @@ where
...
         updater.update_l2_block_data(height, block_gas_used, block_gas_capacity)?;
         let metadata = AlgorithmUpdater::V0(updater.clone()).into();
-        metadata_storage.set_metadata(metadata)?;
+        metadata_storage.set_metadata(&metadata)?;
     }
 
     Ok(())
 }
-
-fn revert_gas_price_db_to_height(
```
acerone85, line 348: why is this fragment of code removed? Is it related to the temporary fix in `combined_database.rs`?

MitchTurner (dismissed):
Sometimes it's hard to know what I should give feedback on or just accept as a stopgap. With that in mind, I don't see any immediate need to hold this up.

crates/fuel-core/src/service/sub_services/algorithm_updater.rs
rymnc: yes we don't need this code anymore since we delegate rolling back of db to `FuelService`

crates/services/gas_price_service/src/fuel_gas_price_updater/tests.rs
rymnc: @mitchturner perhaps?

crates/services/gas_price_service/src/ports.rs
```diff
@@ -12,3 +16,13 @@ pub trait L2Data: Send + Sync {
         height: &BlockHeight,
     ) -> StorageResult<Option<Block<Transaction>>>;
 }
+
+pub trait MetadataStorage: Send + Sync {
```
xgreenx, line 20: Why do we want to move this trait here? I thought that each service(v0 and v1) would have its own ports. It is the same for storage traits as well.

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_core_storage_adapter/metadata_tests.rs
```diff
@@ -36,25 +30,6 @@ fn database() -> StorageTransaction<InMemoryStorage<GasPriceColumn>> {
     InMemoryStorage::default().into_transaction()
 }
 
-#[tokio::test]
-async fn get_metadata__can_get_most_recent_version() {
```
xgreenx, line 40: Why do we want to remove this test?

crates/fuel-core/src/service/sub_services/algorithm_updater.rs
xgreenx: I also don't know the reason for making it generic in this PR; I think we can do it later when we switch from the V0 to the V1 algorithm.

crates/fuel-core/src/service/adapters/gas_price_adapters.rs
```diff
@@ -39,6 +58,44 @@ impl L2Data for OnChainIterableKeyValueView {
     }
 }
 
+impl GasPriceData for Database<GasPriceDatabase> {
+    fn latest_height(&self) -> Option<BlockHeight> {
+        HistoricalView::latest_height(self)
+    }
+}
+
+impl MetadataStorage for Database<GasPriceDatabase> {
```
xgreenx, line 67: Why do we want to duplicate the implementation here instead of deriving the blanket implementation from gas price service crate?

crates/services/gas_price_service/src/fuel_gas_price_updater/tests.rs
MitchTurner: So it's clone-able and we can access a copy of the inner value outside of the task it's sent to.

Quickly looking through the code I don't see anyplace we are using that functionality, but I think we probably should be in the `next__new_l2_block_saves_old_metadata` test.

crates/services/gas_price_service/src/ports.rs
rymnc: The metadata storage would remain the same since we discussed that we will keep it versioned right

xgreenx: It was up to you=D I was thinking that you will choose the approach with two different columns, since it is more flexible.

crates/services/gas_price_service/src/fuel_gas_price_updater/fuel_core_storage_adapter/metadata_tests.rs
MitchTurner: No. They are testing different functions. It doesn't matter if they are redundant, the point isn't coverage.

rymnc: okay

rymnc: addressed in 62343d76a8

crates/services/gas_price_service/src/ports.rs
rymnc: addressed in d51d494f5b

xgreenx: The `gas_price_service` crate doesn't know about the `Database<GasPriceDatabase>`=) 


The comment and name of the trait should be more clear regarding what expectation to the type that implements it, and maybe some description why

crates/fuel-core/src/service/adapters/gas_price_adapters.rs
rymnc: addressed in 417459d415

crates/services/gas_price_service/src/ports.rs
rymnc: clearer in 64818a6c94

xgreenx (approved):

MitchTurner (approved):

rymnc:
already merged into master

Issue #2165, opened 2024-09-05
Fuel uses Gwei as smallest units, but L1 uses Wei


MitchTurner:
For the V1 gas price, where we are taking into account the cost per byte of blobs, we need to make sure we are using the correct units when we integrate the algorithm.

This creates an interesting problem. If the L1 costs drop below 1 Gwei (they are usually 1 Wei!!!!!) we are going to either need to set our fee to 0 or 1. If we choose 1Gwei, then we are going to build up a large profit pillow that will further push the gas price down.

How do we want to handle this?

PR #2229, opened 2024-09-20
Use fees to calculate DA reward and avoid issues with Gwei/Wei conversions

MitchTurner:
## Linked Issues/PRs
Closes https://github.com/FuelLabs/fuel-core/issues/2165
Closes https://github.com/FuelLabs/fuel-core/issues/2243

## Description
- Modify the calculation of rewards to use the block fee instead of block gas.
- Update the charts to be displayed in Gwei rather than Wei to reduce noise and more closely resemble what we will see as an outside observer of the Fuel L2 gas prices.

## Checklist

### Before requesting review


Dentosal (approved):

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -208,7 +208,7 @@ impl AlgorithmUpdaterV1 {
         used: u64,
         capacity: NonZeroU64,
         block_bytes: u64,
-        _gas_price: u64,
+        fee_wei: u64,
```
xgreenx, line 211: The fee that comes from the block is in Gwei

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -243,11 +236,44 @@ impl AlgorithmUpdaterV1 {
...
+    fn update_projected_cost(&mut self, block_bytes: u64) {
+        let block_projected_da_cost =
+            (block_bytes as u128).saturating_mul(self.latest_da_cost_per_byte);
+        self.projected_total_da_cost = self
+            .projected_total_da_cost
+            .saturating_add(block_projected_da_cost);
+    }
+
+    // Take the `fee_wei` and return the portion of the fee that should be used for paying DA costs
+    fn da_portion_of_fee(&self, fee_wei: u64) -> u64 {
```
xgreenx, line 255: Do we need to add the same function for execution price?

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: Yeah. But we don't need to know that here. The caller can convert it for us. In the spirit of what you said:
> I think the algorithm should work with Wei, and all logic around gas price should work with Wei. We always can convert L2 gas price into Wei by using gas price and gas price factor.

It really doesn't matter either way (either wei :P), we just need to pick one and be clear to the caller which one it is.

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: The execution price uses the gas fullness to adjust. I don't know what part of the algorithm would be interested in the gathered fee from exec.

rafal-ch (approved):
LGTM :+1:

crates/fuel-gas-price-algorithm/gas-price-analysis/src/charts.rs
```diff
@@ -284,27 +308,27 @@ pub fn draw_profit(
...
-            0..actual_profit.len(),
-            0..*pessimistic_block_costs.iter().max().unwrap(),
+            0..actual_profit_gwei.len(),
+            0..*pessimistic_block_costs_gwei.iter().max().unwrap(),
         );
 
     chart
         .configure_mesh()
-        .y_desc("Profit")
+        .y_desc("Profit (GWei)")
```
rymnc, line 320: nit: 
```suggestion
        .y_desc("Profit (Gwei)")
```

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -243,11 +236,44 @@ impl AlgorithmUpdaterV1 {
...
+        }
+    }
+
     // We are assuming that the difference between u128::MAX and i128::MAX is negligible
     fn projected_cost_as_i128(&self) -> i128 {
         i128::try_from(self.projected_total_da_cost).unwrap_or(i128::MAX)
     }
 
+    fn rewards_as_i128(&self) -> i128 {
+        i128::try_from(self.total_da_rewards_excess).unwrap_or(i128::MAX)
```
rymnc, line 274: qq: is there ever a chance that the `total_da_rewards_excess` will not be converted successfully?

crates/fuel-gas-price-algorithm/gas-price-analysis/src/charts.rs
```diff
@@ -76,19 +76,29 @@ pub fn draw_gas_prices(
     da_gas_prices: &[u64],
     title: &str,
 ) -> anyhow::Result<()> {
+    let gas_prices_gwei: Vec<_> =
+        gas_prices.into_iter().map(|x| x / 1_000_000_000).collect();
+    let _exec_gas_prices_gwei: Vec<_> = _exec_gas_prices
+        .into_iter()
+        .map(|x| x / 1_000_000_000)
```
rymnc, line 83: nit: suggestion to not use magic numbers, but this works fine as well :)
```suggestion
        .map(|x| x / ONE_GWEI)
```

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: It just depends on what we define as "successful". Essentially, the `i128` is way way bigger than we'll ever need. 

Above I have a comment that isn't accurate:
```rs
     // We are assuming that the difference between u128::MAX and i128::MAX is negligible
```
in reality
```
u128::MAX: 340282366920938463463374607431768211455
i128::MAX:     170141183460469231731687303715884105727
```

But I don't know that any other behavior is better. i.e. is there a more explicit way we could define these types to allow the arithmetic but also capture the intended range. I'm not sure.

Maybe we could rename this `clamped_rewards_as_i128` or something?

crates/fuel-gas-price-algorithm/src/v1.rs
xgreenx: It would be nice to know just total execution profit for statistic, plus later we can transfer part of the execution profit to the da profit.

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: I went with that for now.

crates/fuel-gas-price-algorithm/src/v1.rs
xgreenx: Then I think we need to introduce two new types `GWei` and `Wei` into our codebase and use it everywhere for fee and gas price(Even in state transition function).

In this case, we can be sure that we haven't forgotten to do the conversion.

It is a separate concern for a separate issue=) Could you create an issue please for it?

xgreenx (approved):

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: Maybe? Fuel can be an L2 for other chains than Ethereum. I hesitated to put `wei` in the param name `fee_wei`, but I didn't know how else to specify--and it's much easier to change the one param name here than everywhere in our code. "This is just the Ethereum algorithm" 😬

MitchTurner: https://github.com/FuelLabs/fuel-core/issues/2248

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: When we add any sort of metric that will consume the method, then we can add it.

rymnc (approved):

Issue #2095, opened 2024-08-16
The `ModificationsHistory` table is not ordered


xgreenx:
The ordering of entries is currently broken for the `ModificationsHistory` because it uses LE instead of BE.

<img width="783" alt="image" src="https://github.com/user-attachments/assets/a93bafc0-02d5-41d2-80ff-9d3eaafe51d1">

The fix is simple, but the problem is that it invalidates the old history. We need to migrate from version 1 to version 2 of the historical database and change all keys to be BE.

MuhtasimTanmoy:
On it.

acerone85:
Some questions related to this one: 

* What's the size of the data that needs to be migrated? 
* What do the keys represent? Are collisions likely when changing the encoding from little endian to big endian, or are collisions not a problem (e.g. when the keys are hashes of some object such as blocks or transactions)?
* Are we allowed any downtime?

Depending on the answer, there are two possible solutions: 

Solution 1. - pre-migrate before start

* Iterate over the key-value pairs in the modifications_history column 
* For each key-pair, copy it to a new table `modifications_history_v2` using the right key format. 
* Delete each key-pair from `modifications_history` as soon as it has been copied to `modifications_history_v2` (need to check if iterators over key-pairs allow it). 
* Once the iterator has finished the traversal, proceed with starting the node. 

Pros: 
* All data is migrated before the node is started, we can eventually drop support for the all key encoding
* Idempotent, if the service stops abruptly during the migration we might need to 
* Collision-resistant. Even if we get collisions

Cons:
* If the size of the data to be migrated is large, there might be too much of a downtime
* The rocksDB module will need to be changed to read modifications from a new column, this process might be error prone.

Solution 2.

* Every time we need to fetch a key-value pair, we encode the key with the V2 encoding. If we find a value, then we are done. 
* Otherwise, we read from the same column-family using the V1 encoding for the key, and if we find a value we write it to the same column-family using the V2 encoding. 
* Writes of new data are done directly using the new V2 encoding for the key

Pros: 

* No downtime, 
* Changes should be local to the modifications_history module,

Cons: 
* Not suitable if collisions are a problem e.g. we have one key whose V1 encoding is the same as the V2 encoding for another key, 
* Keys that are never read will never be migrated, we might have to support V1 for a long time

If we do not have collisions and we are happy with supporting V1 key encoding for the foreseeable future, then I'd go with option 2.

xgreenx:
[...]

PR #2233, opened 2024-09-20
Versioned Storage for Modifications History

acerone85:
## Linked Issues/PRs

Related to https://github.com/FuelLabs/fuel-core/issues/2095

## Description

- [ ] Adds a ModificationsHistoryV2 column in the Historical RocksDB, with keys encoded in big endian
- [ ] Adds a flag to the historical RocksDB that keeps track whether the migration is still in progress

- [ ] Change HistoricalRocksDB to alway insert historical changes from V2 table. When the migration is in progress, delete key-value pairs from both V2 and V2, read key-value pairs from V2 falling back to V1 if no value in V2 is found

#TODO (either this or next MR): 
- [x] Small task that performs the migration of key-value pairs from V1 to V2 (Done in #2260)

## Checklist

### Before requesting review

- [ ] Someone else?


crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -300,12 +355,95 @@ where
...
+        }
+    }
+}
+
+// Try to take the value from `ModificationsHistoryV2` first.
+// If the migration is still in progress, remove the value from
+// `ModificationsHistoryV1` and return it if no value for `ModificationsHistoryV2`
+// was found. This is necessary to avoid scenarios where it is possible to
+// roll back twice to the same block height
+fn multiversion_take<Description>(
```
rafal-ch, line 394: `multiversion_take()` and `multiversion_replace()` are very similar, please consider some deduplication. We should be able to provide, let's say, `multiversion_op()` which also takes a generic `F` argument.

Then, on the callsite we can use something like this:
```rust
        let old_changes = multiversion_op(
            storage_transaction,
            height_u64,
            self.migration_in_progress,
            |storage| storage.replace(&height_u64, &reverse_changes),
        )?;
```
or
```rust
        let last_changes = multiversion_op(
            &mut storage_transaction,
            height_to_rollback,
            self.migration_in_progress,
            |storage| storage.take(&height_to_rollback),
        )?
        .ok_or(not_found!(ModificationsHistoryV1<Description>))?;
```

However, due to the complex types, the definition of `F` may become a mess, so I'm leaving it up to you if this is worth changing.
```rust
    F: FnOnce(
        StorageMut<
            '_,
            StructuredStorage<
                InMemoryTransaction<&rocks_db::RocksDb<Historical<Description>>>,
            >,
            ModificationsHistoryV2<Description>,
        >,
    ) -> StorageResult<Option<Changes>>,
```

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -735,11 +875,195 @@ mod tests {
...
+            .iter_all::<ModificationsHistoryV2<OnChain>>(None)
+            .collect::<Vec<_>>();
+        let v1_entries = historical_rocks_db
+            .db
+            .iter_all::<ModificationsHistoryV1<OnChain>>(None)
+            .collect::<Vec<_>>();
+        // then
+
+        assert_eq!(v2_entries.len(), 1);
+        assert_eq!(v1_entries.len(), 0);
```
rafal-ch, line 915: No need to change anything, I like the way you did it, it's more readable. 

I was just surprised that Clippy doesn't complain with `#[warn(clippy::len_zero)]`.

`assert!(v1_entries.len() == 0);` will probably trigger this lint.

crates/fuel-core/src/state/historical_rocksdb/modifications_history.rs
```diff
@@ -3,18 +3,29 @@ use crate::{
...
     transactional::Changes,
     Mappable,
 };
 
-pub struct ModificationsHistory<Description>(core::marker::PhantomData<Description>)
+/// Versioned modification history. The `const N: usize` generic parameter
+/// is used to specify the version.
+/// This allows to define different [`TableWithBlueprint`]
+/// implementations for V1 and V2 of the modification history.
+pub struct ModificationsHistoryVersion<Description, const N: usize>(
```
rafal-ch, line 21:
```suggestion
pub struct ModificationsHistoryVersion<Description, const VERSION: usize>(
```

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -77,10 +80,16 @@ pub enum StateRewindPolicy {
...
+/// Implementation of a database
 #[derive(Debug)]
 pub struct HistoricalRocksDB<Description> {
+    /// The [`StateRewindPolicy`] used by the historical rocksdb
     state_rewind_policy: StateRewindPolicy,
+    /// The Description of the database.
     db: RocksDb<Historical<Description>>,
+    /// Indicates whether the [`Column::ModificationsHistory`]
+    /// is being migrated to [`Column::ModificationsHistoryV2`]
+    migration_in_progress: bool,
```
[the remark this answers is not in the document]
acerone85: Yep, 44c8571384

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -246,28 +268,59 @@ where
...
-            .transpose()?;
+            .iter_all_keys::<ModificationsHistoryV2<Description>>(Some(direction))
+            .next();
+        let v1_changes = if check_v1 {
+            self.db
+                .iter_all_keys::<ModificationsHistoryV1<Description>>(Some(direction))
+                .next()
+        } else {
+            None
+        };
```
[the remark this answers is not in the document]
acerone85: TIL `.then`. 8ca2093f32

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -246,28 +268,59 @@ where
...
+        );
+
+        let latest_height = match (v2_latest_height, v1_latest_height) {
+            (None, None) => Err(DatabaseError::ReachedEndOfHistory)?,
+            (None, Some(v1_res)) => v1_res?,
+            (Some(v2_res), None) => v2_res?,
+            (Some(Err(_)), Some(v1_res)) => v1_res?,
+            (Some(v2_res), Some(Err(_))) => v2_res?,
+            (Some(Ok(v1)), Some(Ok(v2))) => v1.min(v2),
+        };
```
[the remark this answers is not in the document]
acerone85: done in 55c4528cc9

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -300,12 +355,95 @@ where
 
         Ok(())
     }
+
+    /// Migrates a ModificationHistory key-value pair from V1 to V2.
+    /// The migration fails if other
```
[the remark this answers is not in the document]
acerone85: Changed: 87ace1b3bf. Although I'll wait also for @xgreenx to check if my comment is correct, because from the code it does not look like we use the `ConflictPolicy::Fail` flag when writing directly to rocksDB.

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -300,12 +355,95 @@ where
...
+        );
+
+        let v1_changes = migration_transaction
+            .storage_as_mut::<ModificationsHistoryV1<Description>>()
+            .take(&height)?;
+        if let Some(v1_changes) = v1_changes {
+            migration_transaction
+                .storage_as_mut::<ModificationsHistoryV2<Description>>()
+                .insert(&height, &v1_changes)
+                .unwrap();
```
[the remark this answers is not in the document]
acerone85: Nice catch: fe3e453fec

crates/fuel-core/src/state/historical_rocksdb.rs
acerone85: Tried, the signature itself is not a problem, but I spent quite a bit of time fighitng with lifetimes and eventually gave up.

crates/fuel-core/src/state/historical_rocksdb/modifications_history.rs
acerone85: Done: fe3e453fec

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -300,12 +354,96 @@ where
...
+    /// The migration fails if a concurrent transaction which rollbacks
+    /// to the `height` being migrated commits before this
+    /// migration transaction commits.
+    pub fn migrate_modifications_history_at_height(
+        &self,
+        height: u64,
+    ) -> StorageResult<()> {
+        let mut migration_transaction = StorageTransaction::transaction(
+            &self.db,
+            ConflictPolicy::Fail,
```
xgreenx, line 368: Conflict policy works only when you try to commit one in-memory transaction into another. It doesn't work with the Rocksdb.

The migration process should accumulate changes in the memory and dump them inside of the `commit_changes` function(when upper services like block importer or relayer call this function).

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -94,6 +103,7 @@ where
         Ok(Self {
             state_rewind_policy,
             db,
+            modifications_history_migration_in_progress: true,
```
xgreenx, line 106: It would be nice to add logic that checks do we have any entries in the V1. If not, set this to false

xgreenx (commented):
It also would be nice to have a tests that inserts 1000 blocks, and tries to rollback them by calling `HistoricalRocksDB::rollback_last_block`

crates/fuel-core/src/state/historical_rocksdb.rs
rafal-ch: Please see here or how I imagined it :)
https://github.com/FuelLabs/fuel-core/pull/2277

But again, up to you. Feel free to close the PR and leave it implemented with two separate functions if you find it more convenient.

acerone85: Thanks, I had a look :) 
I changed this PR quite a lot, so still fighting with typechecking when trying to rebase. 
But I definitely like it how you have done it

acerone85:
> It also would be nice to have a tests that inserts 1000 blocks, and tries to rollback them by calling `HistoricalRocksDB::rollback_last_block`

Done. There is a new test that checks that commits are rollback consistently to the latest height, when we use V2. 0d2c4908d84b6291f961b1944104164ec7e5ef1f

crates/fuel-core/src/state/historical_rocksdb.rs
acerone85: Thanks. In the latest version of the MR I switched to having cumulative changes in the rocksDB state, and I handle race conditions by locking on them for the whole duration of the migration/rollback/commit_changes function. Unfortunately I cannot use more fine-grained guards because of potential consistency issues (explanation in the comments).

crates/fuel-core/src/state/historical_rocksdb.rs
acerone85: Yes, the latest version of the PR does not have this check anymore, and instead we check directly if we are able to fetch at least one change for the v1 history.

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -295,20 +416,160 @@ where
...
+            .commit_changes(&std::mem::take(&mut *self.migration_changes.lock()))
+    }
+}
+
+// Try to take the value from `ModificationsHistoryV2` first.
+// If the migration is still in progress, remove the value from
+// `ModificationsHistoryV1` and return it if no value for `ModificationsHistoryV2`
+// was found. This is necessary to avoid scenarios where it is possible to
+// roll back twice to the same block height
+fn multiversion_take<Description, T>(
```
rafal-ch, line 514: I'd still suggest deduplicating these functions with:
```
fn multiversion_op<Description, T, F>(
    storage_transaction: &mut StorageTransaction<T>,
    height: u64,
    modifications_history_migration_in_progress: bool,
    f: F,
) -> StorageResult<Option<Changes>>
where
    T: KeyValueInspect<Column = Column<Description>>,
    F: FnOnce(
        StorageMut<'_, StorageTransaction<T>, ModificationsHistoryV2<Description>>,
    ) -> StorageResult<Option<Changes>>,
{
    let v2_last_changes =
        f(storage_transaction.storage_as_mut::<ModificationsHistoryV2<Description>>())?;

    if modifications_history_migration_in_progress {
        let v1_last_changes = storage_transaction
            .storage_as_mut::<ModificationsHistoryV1<Description>>()
            .take(&height)?;
        Ok(v2_last_changes.or(v1_last_changes))
    } else {
        Ok(v2_last_changes)
    }
}
```

crates/fuel-core/src/state/historical_rocksdb/modifications_history.rs
```diff
@@ -39,3 +57,18 @@ where
...
+/// Blueprint for Modifications History V2. Keys are stored in big endian order
+/// using the `Column::HistoryColumnV2` column family.
+impl<Description> TableWithBlueprint for ModificationsHistoryV2<Description>
+where
+    Description: DatabaseDescription,
+{
+    type Blueprint = Plain<Primitive<8>, Postcard>;
+    type Column = Column<Description>;
+
+    // We store key-value pairs using the same column family as for V1
```
rafal-ch, line 70: I'm a bit confused here, since I see that the column is different, not the same.

crates/fuel-core/src/state/historical_rocksdb/modifications_history.rs
```diff
@@ -26,11 +38,17 @@ where
...
+
+/// Blueprint for Modifications History V1. Keys are stored in little endian order
+/// using the `Column::HistoryColumn` column family.
+impl<Description> TableWithBlueprint for ModificationsHistoryV1<Description>
 where
     Description: DatabaseDescription,
 {
-    // TODO: The Blueprint should be `Plain<Primitive<8>, Postcard>` to sort
     //  the keys in the database. https://github.com/FuelLabs/fuel-core/issues/2095
     type Blueprint = Plain<Postcard, Postcard>;
```
rafal-ch, line 53: Postcard uses "little endian" by default, making this is the actual root-cause, right?

crates/fuel-core/src/state/historical_rocksdb/modifications_history.rs
```diff
@@ -26,11 +38,17 @@ where
...
+    ModificationsHistoryVersion<Description, 1>;
+
+/// Blueprint for Modifications History V1. Keys are stored in little endian order
+/// using the `Column::HistoryColumn` column family.
+impl<Description> TableWithBlueprint for ModificationsHistoryV1<Description>
 where
     Description: DatabaseDescription,
 {
-    // TODO: The Blueprint should be `Plain<Primitive<8>, Postcard>` to sort
     //  the keys in the database. https://github.com/FuelLabs/fuel-core/issues/2095
```
rafal-ch, line 52: We might as well remove the second line of the comment.

crates/fuel-core/src/state/historical_rocksdb/modifications_history.rs
```diff
@@ -3,18 +3,30 @@ use crate::{
...
+        postcard::Postcard,
+        primitive::Primitive,
+    },
     structured_storage::TableWithBlueprint,
     transactional::Changes,
     Mappable,
 };
 
-pub struct ModificationsHistory<Description>(core::marker::PhantomData<Description>)
+/// Versioned modification history. The `const N: usize` generic parameter
```
rafal-ch, line 17:
```suggestion
/// Versioned modification history. The `const VERSION: usize` generic parameter
```

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -439,15 +703,50 @@ where
...
+        );
+
         if let Some(height) = height {
             self.store_modifications_history(&mut storage_transaction, &height)?;
         }
 
-        self.db
-            .commit_changes(&storage_transaction.into_changes())?;
-        Ok(())
+        // This cannot fail
```
rafal-ch, line 731: This comment just caught my attention :)
How can we be sure it doesn't fail?

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -439,15 +703,50 @@ where
...
+            cumulative_changes.clone(),
+        );
+        let mut storage_transaction = StorageTransaction::transaction(
+            &mut migration_transaction,
+            ConflictPolicy::Overwrite,
+            changes,
+        );
+
         if let Some(height) = height {
             self.store_modifications_history(&mut storage_transaction, &height)?;
```
rafal-ch, line 728: In case we fail here, could it be that we lose the changes?

crates/fuel-core/src/state/historical_rocksdb.rs
acerone85: thanks, I need to become better at generic programming in Rust :) 
I tried to incorporate your changes from a previous PR of yours, but failed miserably while refactoring 😅

AurelienFT (commented):
I agree with some comments of rafal and added some

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -246,41 +275,133 @@ where
...
     //  https://github.com/FuelLabs/fuel-core/issues/2095
     fn rollback_last_block(&self) -> StorageResult<u64> {
-        let latest_height = self
-            .db
-            .iter_all_keys::<ModificationsHistory<Description>>(Some(
-                IterDirection::Reverse,
-            ))
-            .next()
-            .ok_or(DatabaseError::ReachedEndOfHistory)??;
+        let modifications_history_migration_in_progress = self.is_migration_in_progress();
```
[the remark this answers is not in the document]
acerone85: The method would still not work properly after this PR, because the migration is not triggered at start. 
I would prefer to remove the todo in #2260

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -246,41 +275,133 @@ where
...
+    }
+
     fn rollback_block_to(&self, height_to_rollback: u64) -> StorageResult<()> {
-        let mut storage_transaction = self.db.read_transaction();
+        let (cumulative_changes, maybe_cumulative_changes_guard) =
+            self.take_migration_changes();
+
+        // Will clone an empty set of changes if the migration is not in progress, which should
+        // not impact performance. However, when a migration is in progress, this operation could
+        // be expensive if many changes have been accumulated.
```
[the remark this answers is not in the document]
acerone85: Short answer: it is possible, but the logic becomes complicated and I'm happy with keeping the lock for a considerably long time. 

Long answer: At a logical level, we have three kind of transactions: "rollback", "commit" and "migrate". The first two commit all the cumulative changes from memory into the underlying rocksdb, while the latter only adds its changes the cumulative changes in memory.  

We can technically keep the lock guard for a short period of time, only when taking and pushing cumulative changes in the shared memory changes, but there is some extra care to be taken: 
* Changes from "commit" and "rollback" transactions are considered to be newer, hence they should take priority and overwrite changes from migraiton transactions. When we want to add new changes to the in-memory shared cumulative changes, we simply overwrite them on top of any change that we might have. 
* Migration changes can be computed independently, but once we want to commit them we will need to acquire the lock on the shared portion of cumulative changes, create a new transaction with the cumulative changes as base, and commit the migration changes on top of it. We must make sure that the migration changes do not overwrite any "newer" changes from other transactions, so a `ConflictPolicy = Fail` must be used for migration transactions. Even if the transaction fails, we know that another transaction has updated the key to a newer value (which will be stored as V2). 
* There is still a chance of a race where the migration changes of a key are written to the shared memory after newer V2 data for that key has been committed to the DB. In this case the migration transaction will succeed despite having a `ConflictPolicy = Fail`. If we want to avoid that altogether, we must either keep the lock while changes are flushed to the database (which basically invalidates any performance benefit), or we must keep the cumulative changes for longer after they have been committed to RocksDB, to prevent migrating transactions from committing stale data. This now poses the problem to have to remove cumulative changes from memory, which I'm not 100% how to do safely.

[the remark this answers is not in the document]
acerone85: (Btw, another way around it is to use a `RwLock`. Migrating transactions can acquire a `Read` lock, and other transactions can acquire a `Write` lock. But I'm afraid having a read lock for something that writes to memory might raise some eyebrows)

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -295,20 +416,160 @@ where
...
+            Changes::default(),
+        );
+
+        let v1_changes = migration_transaction
+            .storage_as_mut::<ModificationsHistoryV1<Description>>()
+            .take(&height)?;
+        if let Some(v1_changes) = v1_changes {
+            migration_transaction
+                .storage_as_mut::<ModificationsHistoryV2<Description>>()
+                .insert(&height, &v1_changes)?; // TODO: If this is an error we want to at least write back the old changes
```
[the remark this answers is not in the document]
acerone85: Stale Todo, I'll check all of them to make sure they have been addressed. For this one here, This function cannot fail.

The long explanation: 
This one here is a leftover from a previous iteration. `InMemoryTransaction<S>` implements `KeyValueMutate<_>::put` (https://github.com/FuelLabs/fuel-core/blob/f5240cb8baa8210c7b744a49d96695df0c84b2f1/crates/storage/src/transactional.rs#L413), and this is what ultimately gets called by `StructuredStorage<InMemoryTransaction<S>>::insert()`, https://github.com/FuelLabs/fuel-core/blob/773226cb326c2607d274717348ff50aab30b0828/crates/storage/src/structured_storage.rs#L271, via `TableWithBlueprint::Blueprint::put()` (https://github.com/FuelLabs/fuel-core/blob/ce72eeaf5c9490f87aa3a39ae7f92087908e7d79/crates/storage/src/blueprint/plain.rs#L47). 
If you follow the implementation you will see that the call to insert here fails if and only if the inner call to `KeyValueInspect::put` fails, but this is not the case for `InMemoryStorage`.

crates/fuel-core/src/state/historical_rocksdb.rs
acerone85: Code digging :) 
This is because the transaction targets another transaction as the storage into which to commit, and the `ConflictPolicy` is specified to be `Overwrite`. 
When `.commit` is called, the `.commit_changes` function from the underlying inner storage is called (https://github.com/FuelLabs/fuel-core/blob/f5240cb8baa8210c7b744a49d96695df0c84b2f1/crates/storage/src/transactional.rs#L280). 
This is implemented here: https://github.com/FuelLabs/fuel-core/blob/f5240cb8baa8210c7b744a49d96695df0c84b2f1/crates/storage/src/transactional.rs#L287. If you follow the code, you will see that the function can fail only when `ConflictPolicy = Fail`. 

I'll clarify this in the comment

crates/fuel-core/src/state/historical_rocksdb/modifications_history.rs
acerone85: That is my understanding, yes :)

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -735,11 +1034,245 @@ mod tests {
         assert_eq!(result, Ok(1));
         let entries = historical_rocks_db
             .db
-            .iter_all::<ModificationsHistory<OnChain>>(None)
+            .iter_all::<ModificationsHistoryV2<OnChain>>(None)
             .collect::<Vec<_>>();
         assert_eq!(entries.len(), 0);
     }
 
+    #[test]
```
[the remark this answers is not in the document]
acerone85: Done: 9f050e5aea60aa5b7d2678a45366715d7a2b949f 
(Although for these tests `Given/When/Then` seem not to work well since we test multiple things pre, mid, and post migration

crates/fuel-core/src/state/historical_rocksdb/description.rs
```diff
@@ -82,6 +90,7 @@ where
                 Some(Description::prefix(c).unwrap_or(0).saturating_add(8)) // `u64::to_be_bytes`
             }
             Column::HistoryColumn => Some(8),
+            Column::HistoryV2Column => Some(8),
```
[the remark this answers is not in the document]
acerone85: Done: a5cc0bbd6ccd4ff720b4fb4727ea5adda53eb5a5

crates/fuel-core/src/state/historical_rocksdb.rs
acerone85: Good catch. 2739d6162aa9e7e3fbc6e4c56cb546379f6f3146

crates/fuel-core/src/state/historical_rocksdb.rs
acerone85: Clarified in 4dc2643fc248d656ed772fb55efc8a77f2b9ca99

crates/fuel-core/src/state/historical_rocksdb/modifications_history.rs
acerone85: Done: d56f351f9ba666fadd507ce3839b79c766f9345e

crates/fuel-core/src/state/historical_rocksdb/modifications_history.rs
acerone85: yep: 315172c9aad7b3e87f510308546579f85469c14e

crates/fuel-core/src/state/historical_rocksdb/modifications_history.rs
acerone85: yes, stale comment. Removed in f162887b0c

crates/fuel-core/src/state/historical_rocksdb.rs
acerone85: Added your change: 5ea1bbcb0c668d7d49aeb049001a8f3648bee6e0

crates/fuel-core/src/state/historical_rocksdb.rs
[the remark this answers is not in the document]
acerone85: Todo removed in f996101ba5d4472aa64f190d2bb1d3a2a9006b50

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -295,20 +416,160 @@ where
...
+    /// the migration from ModificationHistoryV1 to ModificationHistoryV2
+    /// are simply recorded, and will be flushed to the database when it
+    /// commits or rollbacks to a new height.
+    #[cfg(test)]
+    fn migrate_modifications_history_at_height(&self, height: u64) -> StorageResult<()> {
+        // We need to keep the lock to the migration changes until the end of the function.
+        // This is to avoid a scenario where the migration changes overwrites changes to the modification history
+        // coming from other transactions.
+        // If we were to release the lock as soon as we take the cumulative changes, and then acquiring it again when
+        // writing the updatet set of changes, then we could have the following schedule of events.
```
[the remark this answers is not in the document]
acerone85: 07d43e2a8ca6d556a7c9db564805e18579fc2c00

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -248,41 +278,133 @@ where
...
+
+                Ok(())
+            }
+        }
+    }
+
     fn rollback_block_to(&self, height_to_rollback: u64) -> StorageResult<()> {
-        let mut storage_transaction = self.db.read_transaction();
+        let (cumulative_changes, maybe_cumulative_changes_guard) =
+            self.take_migration_changes();
```
xgreenx, line 385: Could you elaborate more on why we need to change how `rollback_block_to` should work?

How migration is going shouldn't affect the work of the `rollback_block_to` method. Migration may have a state higher than the chain's height (after rollback), and it can be a problem. But we can handle this case in the `commit_changes` function and remove modifications that are higher than blockchain height.

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -297,20 +419,171 @@ where
...
+        }
+    }
+
+    /// Migrates a ModificationHistory key-value pair from V1 to V2.
+    /// The migration is lazy, in that the changes necessary to perform
+    /// the migration from ModificationHistoryV1 to ModificationHistoryV2
+    /// are simply recorded, and will be flushed to the database when it
+    /// commits or rollbacks to a new height.
+    #[cfg(test)]
+    fn migrate_modifications_history_at_height(&self, height: u64) -> StorageResult<()> {
```
xgreenx, line 450: Side question: Why do we want to implement a function in this PR? It can be part of the migration PR.


I think the migration process should work with the iterator of the historical modification table instead of fetching each modification height by height. In this case, you can easily fetch butch of modifications up to some limit, migrate them, and store them in one `Changes` type inside of the `migration_changes` by just locking it once. Then the migration process can wait until we take changes from `migration_changes` during `commit_changes`, and continue after.

With this solution, you only lock the `migration_changes` changes once for very fast operation. Plus, the iteration over the storage will be much faster(and it will work even if we do rollback or alter the history somehow because while you are holding an iterator over RocksDB, it will not lose the data).

crates/fuel-core/src/state/historical_rocksdb.rs
acerone85: I follow only partially here. I was under the impression that `rollback_block_to` reverts all the changes from the current height to the one specified in the function, but this does not seem to be the case. This leaves me with the question: why is it okay to revert only the changes at the block height specified in input? Do we use this function only to rollback to the last block (which seems to be the case from the codebase)? 

Regarding your question, my flow of thought is as follows: Migration changes are piggybacked into transactions whenever they are committed to the rocksDB instance. We do this in two places: `rollback_block_to` and `<HistoricalRocksDB as TransactableStorage>::commit_changes`. In the first case we compute the last_changes by taking into account the migration changes from the in memory transaction, so even if the key for that height is being migrated from V1 to V2, the migrated changes will be captured in `last_changes` and removed. Please let me know if i am missing something. 

If the cumulative changes were not flushed here, they would be kept in memory and flush when we later commit a block. Now we could have a scenario where we rollbacked a block several time, but we did not commit the migration changes for blocks at a higher height.

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -441,15 +717,58 @@ where
         height: Option<Description::Height>,
         changes: Changes,
     ) -> StorageResult<()> {
-        let mut storage_transaction =
-            StorageTransaction::transaction(&self.db, ConflictPolicy::Overwrite, changes);
+        // cumulative_changes_lock_guard is defined to be Some only when the migration is in progress.
```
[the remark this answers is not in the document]
acerone85: For clarity: do you mean the function `take_migration_changes` should return only `Option<Changes>`?
If we do not care about keeping the migration changes I can remove keeping the lock guard here. 
I also don't need to readd the migration changes later on.

Note that this works under the assumption that `commit_changes` always commit changes at a `height` that is not present already in the underlying db, so I want to validate this first. If this is not the case, then there is the risk of having a stale version from a migration  overwriting the newer version that has been written into the DB by `commit_changes`.

crates/fuel-core/src/state/historical_rocksdb.rs
acerone85: Sounds good. I can make the change, although probably it makes sense to do it in the migration PR. 
Btw in this PR that function is only for testing purposes, so if that's okay I would keep it and change the migration process using the rocksDB iterator in the follow-up.

crates/fuel-core/src/state/historical_rocksdb.rs
xgreenx: > This leaves me with the question: why is it okay to revert only the changes at the block height specified in input? Do we use this function only to rollback to the last block (which seems to be the case from the codebase)?

It is a temporary workaround because `rollback_last_block` doesn't work as expected.

> Migration changes are piggybacked into transactions whenever they are committed to the rocksDB instance.

For us it will be enough to commit changes from the migration only in `commit_changes`. 

> If the cumulative changes were not flushed here, they would be kept in memory and flush when we later commit a block. Now we could have a scenario where we rollbacked a block several time, but we did not commit the migration changes for blocks at a higher height.

I don't see any problems in keeping them in the memory until we start importing new blocks(it is okay if we need to rollback a lot of blocks. Usually it is how we use `rollback_block_to`).

crates/fuel-core/src/state/historical_rocksdb.rs
xgreenx: It seems this function is only used inside of the `migrate_modifications_history_works` tests that is dedicated to this function=D I think it is better to test in the PR where we really plan to have final migration

crates/fuel-core/src/state/historical_rocksdb.rs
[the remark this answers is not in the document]
xgreenx: > do you mean the function take_migration_changes should return only Option<Changes>?

Yes, it seems should be enough.

> Note that this works under the assumption that commit_changes always commit changes at a height that is not present already in the underlying db, so I want to validate this first. If this is not the case, then there is the risk of having a stale version from a migration overwriting the newer version that has been written into the DB by commit_changes.

Why we can't apply changes from the `store_modifications_history` on top of the changes from migration? In this case  we will not lose changes from the `store_modifications_history`.

If you worry that later migration will override modification history, then it shouldn't happen because you migrate only v1 entries, while `store_modifications_history` uses v2. Only in the case of rollback can you face the situation where migration with old modifications can override new ones. But as I mentioned in another comment, we can handle this situation just during the migration process by tracking what is the maximum height of the migration(and decreasing this height during the rollback event)

crates/fuel-core/src/state/historical_rocksdb.rs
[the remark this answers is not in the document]
rafal-ch: Thanks for elaboration. I think if we can live with the lock being locked for some longer time, there's no need to complicate the code even more.

crates/fuel-core/src/state/historical_rocksdb.rs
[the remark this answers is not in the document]
acerone85: > Why we can't apply changes from the store_modifications_history on top of the changes from migration? In this case we will not lose changes from the store_modifications_history. 

The scenario I was concerned about goes as follows (As I mention this can happen only if we commit twice at the same height, which does not seem something that can happen without rollbacking first)

- You start the function to commit at a height for which there is a V1 key present already, with a different set of changes. This function takes the migration changes to commit, and find that they are empty
- You migrate the same key from V1 to V2. The changes (which are different from the ones from V1) are persisted in memory
- The commit transaction commits into rocksDB. Later on, a second commit transaction at a different height gets the stale migration changes that have been persisted in memory. 

So if the following above cannot happen then I am convinced that it should be fine to not keep the lock for long

crates/fuel-core/src/state/historical_rocksdb.rs
acerone85: > I don't see any problems in keeping them in the memory until we start importing new blocks(it is okay if we need to rollback a lot of blocks. Usually it is how we use rollback_block_to).

My main concern in this case is a scenario that would go as follows (note that unless I have misunderstood the rollback and commit functions, this can happen): 

- Say the max height of the modification history is 100. We rollback the last block 50 times, using a strategy where we do not commit the migration changes in rocksDB. 
- During this time the migration process for all blocks completes and the changes are persisted in memory, 
- The next time the changes for a block height are committed, all the migration changes (including those at height > 50) will be committed, leaving the DB in an inconsistent state. 

You mentioned to keep track of the maximum height of the migration, but if rollbacks are not frequent a simpler alternative would be to wipe all migration changes before committing a transaction. Though this would require the lock to be kept while the commit to rocksDB happens to avoid data races.

acerone85: (Or we can simply wipe all migration changes higher than the block height to be committed, when committing)

rafal-ch (dismissed):

rafal-ch:
Approved with the assumption that the additional comments from @xgreenx will be addressed/resolved (could be in follow-up PR).
cc: @acerone85

xgreenx (approved):

rafal-ch (dismissed):

Cargo.lock
```diff
@@ -6958,7 +6958,7 @@ dependencies = [
  "pest_meta",
  "proc-macro2",
  "quote",
- "syn 2.0.87",
+ "syn 2.0.89",
```
xgreenx, line 6961:
```suggestion
 "syn 2.0.87",
```

xgreenx (approved):

rafal-ch (approved):

Issue #2210, opened 2024-09-17
Bug: It is possible for `abs` to overflow in V1 algorithm


MitchTurner:
This is a bug that was pointed out during the Immunefi event.

Similar to in this reduced piece of code:

```rs
fn main() {
    let num: i64 = -9223372036854775808;
    assert_eq!(num, i64::MIN);
    dbg!(&num.abs());
}
```

or here:

```rs
#[test]
#[should_panic]
fn algo_v1_panics() {
    let algo = AlgorithmV1 {
        min_da_gas_price: 0,
        new_exec_price: 0,
        last_da_price: 0,
        max_change_percent: 0,
        latest_da_cost_per_byte: 0,
        total_rewards: 0,
        total_costs: 0,
        da_p_factor: 0,
        da_d_factor: 0,
        avg_profit: 0,
        avg_window: 0,
    };

    algo.change(i64::MIN / 2, i64::MIN / 2);
}
```

PR #2241, opened 2024-09-23
Protect against passing `i128::MIN` to `abs()` which causes overflow

rafal-ch:
## Linked Issues
Closes https://github.com/FuelLabs/fuel-core/issues/2210

## Description
This PR uses `saturating_abs()` instead of "raw" `abs()` inside the `da_change()` to prevent overflow on `i128::MIN`.

### Before requesting review

### DIFF
--- a/crates/fuel-gas-price-algorithm/src/v1.rs
+++ b/crates/fuel-gas-price-algorithm/src/v1.rs
@@ -4,7 +4,10 @@ use std::{
     ops::Div,
 };
 
-use crate::utils::cumulative_percentage_change;
+use crate::utils::{
+    cumulative_percentage_change,
+    safe_signed_abs,
+};
 
 #[cfg(test)]
 mod tests;
@@ -325,19 +328,13 @@ impl AlgorithmUpdaterV1 {
 
     fn da_change(&self, p: i128, d: i128) -> i128 {
         let pd_change = p.saturating_add(d);
-        let pd_change = if pd_change == i128::MIN {
-            pd_change + 1
-        } else {
-            pd_change
-        };
         let upcast_percent = self.max_da_gas_price_change_percent.into();
         let max_change = self
             .new_scaled_da_gas_price
             .saturating_mul(upcast_percent)
             .saturating_div(100)
             .into();
-        debug_assert!(pd_change != i128::MIN);
-        let clamped_change = pd_change.abs().min(max_change);
+        let clamped_change = safe_signed_abs(pd_change).min(max_change);
         pd_change.signum().saturating_mul(clamped_change)
     }
 
--- a/crates/fuel-gas-price-algorithm/src/v1.rs
+++ b/crates/fuel-gas-price-algorithm/src/v1.rs
@@ -325,12 +325,18 @@ impl AlgorithmUpdaterV1 {
 
     fn da_change(&self, p: i128, d: i128) -> i128 {
         let pd_change = p.saturating_add(d);
+        let pd_change = if pd_change == i128::MIN {
+            pd_change + 1
+        } else {
+            pd_change
+        };
         let upcast_percent = self.max_da_gas_price_change_percent.into();
         let max_change = self
             .new_scaled_da_gas_price
             .saturating_mul(upcast_percent)
             .saturating_div(100)
             .into();
+        debug_assert!(pd_change != i128::MIN);
         let clamped_change = pd_change.abs().min(max_change);
         pd_change.signum().saturating_mul(clamped_change)
     }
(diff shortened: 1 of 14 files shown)


acerone85 (commented):
Looks like what you want is `i32::saturating_abs()`

https://doc.rust-lang.org/std/primitive.i32.html#method.saturating_abs

netrome:
> Looks like what you want is `i32::saturating_abs()`
> 
> https://doc.rust-lang.org/std/primitive.i32.html#method.saturating_abs

Tiny correction, we want `i128::saturating_abs()`. But yeah seems like that one should be used instead.

https://doc.rust-lang.org/std/primitive.i128.html#method.saturating_abs

rafal-ch:
> Looks like what you want is `i32::saturating_abs()`
> 
> https://doc.rust-lang.org/std/primitive.i32.html#method.saturating_abs

Nice one, TIL :+1: 

I've been playing with `unsigned_abs()` but this only moved the problem to different part of the code :)

Updated in: https://github.com/FuelLabs/fuel-core/pull/2241/commits/056ba1e9592955650536c851a68c978cb5b58445

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -331,7 +331,7 @@ impl AlgorithmUpdaterV1 {
             .saturating_mul(upcast_percent)
             .saturating_div(100)
             .into();
-        let clamped_change = pd_change.abs().min(max_change);
+        let clamped_change = pd_change.saturating_abs().min(max_change);
```
MitchTurner, line 334: Oh. Nice find.

crates/fuel-gas-price-algorithm/src/v1/tests/algorithm_v1_tests.rs
```diff
@@ -143,3 +147,30 @@ fn worst_case__same_block_gives_the_same_value_as_calculate() {
...
+        projected_total_da_cost: 0,
+        da_p_component: 0,
+        da_d_component: 0,
+        last_profit: 0,
+        second_to_last_profit: 0,
+        latest_da_cost_per_byte: 0,
+        unrecorded_blocks: vec![],
+    };
+
+    updater.da_change(i128::MIN / 2, i128::MIN / 2);
```
MitchTurner, line 175: Let's not test private functions. We should be able to test this with `update_l2_block_data`.

acerone85 (dismissed):
LGTM

crates/fuel-gas-price-algorithm/src/v1/tests/algorithm_v1_tests.rs
rafal-ch: Updated in https://github.com/FuelLabs/fuel-core/pull/2241/commits/3047c7e2024081212514b0bd3304ca37986fe732, however, I'm not fully convinced this fits better here.

_Rationale:_
we need to carefully select the values in order to trigger the "what used to be" an overflow inside `da_change()`. Now, if we, for example, change how the `total_da_rewards_excess` is updated [here](https://github.com/FuelLabs/fuel-core/blob/master/crates/fuel-gas-price-algorithm/src/v1.rs#L224-L227) or how we calculate P and D [here](https://github.com/FuelLabs/fuel-core/blob/master/crates/fuel-gas-price-algorithm/src/v1.rs#L311-L324) we may cause the `da_change()` to be invoked with values that do not cause overflow and the test will pass with both `abs()` and `saturating_abs()`.

I'm open to suggestions on how we can do this better.

MitchTurner: I'd be okay with removing the test since there are many places in the code that "could" panic if programmed poorly.

It might be worth adding some more aggressive prop testing for those cases.

rafal-ch: I think this boils down to the fact that people (me, at least) do not associate an "innocent" `abs()` call with a possible panic. Think `unwrap`s or `expect`s - these are usually caught during the review process and removed from the production code + after removing them, we do not add a test.

So yes, I agree that this test is kinda redundant - I removed it in https://github.com/FuelLabs/fuel-core/pull/2241/commits/7359bd7a850e6a48732c4bc887526dd39093f196

What we could do is:

1. Spawn a tech-debt issue to review the codebase against unwraps, expects, **abs** and similar and update relevant places
2. Rely on a clippy lint to detect those, but as far as I know there is no such lint. Maybe we can contribute to clippy and add "potential_panic_in_std"? :)

Wdyt?

acerone85 (approved):

crates/fuel-gas-price-algorithm/src/v1/tests/algorithm_v1_tests.rs
MitchTurner: Currently we have
```rs
#![deny(warnings)]
```

But I'd be open to adding more for sure.

MitchTurner (approved):

netrome (approved):

Issue #2225, opened 2024-09-19
 Test that executor is resilient against bad transaction sources


acerone85:
The new change is not tested, I'd need to implement a `BadTransactionSource` and modify the commit_without_execute function to run with that transaction source. I can implement that if necessary,  but I'd prefer to open a dedicated PR for it.

_Originally posted by @acerone85 in https://github.com/FuelLabs/fuel-core/pull/2182#discussion_r1767166815_

PR #2242, opened 2024-09-23
Executor: test Tx_count limit with incorrect tx source

acerone85:
## Linked Issues/PRs

Closes https://github.com/FuelLabs/fuel-core/issues/2225

## Description

As per discussion, we should check that the current Executor logic is resilient against transactions source that ignore the limit on MAX_TX_COUNT. 

## Checklist

### Before requesting review

### DIFF
--- a/crates/fuel-core/src/executor.rs
+++ b/crates/fuel-core/src/executor.rs
@@ -4,17 +4,22 @@
 #[cfg(test)]
 mod tests {
 
+    use std::{
+        cell::RefCell,
+        sync::Mutex,
+    };
+
     use crate as fuel_core;
     use fuel_core::database::Database;
     use fuel_core_executor::{
         executor::{
-            BadTransactionsSource,
             OnceTransactionsSource,
             MAX_TX_COUNT,
         },
         ports::{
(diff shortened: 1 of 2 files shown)


crates/fuel-core/src/executor.rs
```diff
@@ -183,6 +190,32 @@ mod tests {
         }
     }
 
+    /// Bad transaction source: ignores the limit of `u16::MAX -1` transactions
+    /// that should be returned by [`TransactionsSource::next()`].
+    /// It is used only for testing purposes
+    pub struct BadTransactionsSource {
+        transactions: Mutex<RefCell<Vec<MaybeCheckedTransaction>>>,
```
xgreenx, line 197: You either need to use `RefCell` or `Mutex`=D Both is too much=D

acerone85: Ah yes, I can access the data as mutable from the Lock, my bad. Fixed in 8d5e04c9de

xgreenx (approved):

xgreenx (approved):

rafal-ch (approved):
LGTM :+1:

Issue #2244, opened 2024-09-24
Use true bytes not compressed bytes for calculated `cost_per_byte` in algorithm


MitchTurner:
The committer isn't aware of the original byte size of each block, just the compressed bytes. We need this information to calculate the real value of `cost_per_byte`. 

We should record the bytes for each  block in `unrecorded_blocks` and use that value when updating the `latest_cost_per_byte` and ensure that is the case for any place we are using the block bytes.

PR #2252, opened 2024-09-25
Use bytes from `unrecorded_blocks` rather from the block from DA

MitchTurner:
## Linked Issues/PRs
Closes https://github.com/FuelLabs/fuel-core/issues/2244

## Description
Essentially, we would be getting the compressed bytes from the committer, and we already have access to the true bytes in the `unrecorded_blocks`, so this PR changes the algo to use the values on those `unrecorded_blocks`

Uncovered some problems with the simulator in the mean time and created a followup issue for dealing with that.

image

## Checklist

### Before requesting review

### Followup Issue
https://github.com/FuelLabs/fuel-core/issues/2264


crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -372,19 +368,18 @@ impl AlgorithmUpdaterV1 {
...
-    ) -> Result<(), Error> {
+    fn da_block_update(&mut self, height: u32, block_cost: u64) -> Result<(), Error> {
         let expected = self.da_recorded_block_height.saturating_add(1);
         if height != expected {
             Err(Error::SkippedDABlock {
                 expected: self.da_recorded_block_height.saturating_add(1),
                 got: height,
             })
         } else {
+            let block_bytes = self
```
xgreenx, line 379: Maybe we want to remove all blocks until `height`?

MitchTurner: > Maybe we want to remove all blocks until height?

That might be more performant (and kinda what we were doing before). We'd want to do that before we called `da_block_update` and pair each `block_bytes` with the corresponding `RecordedBlock`.

I honestly don't know the performance of `remove` for `HashMap`. `get` is O(1), but obviously doesn't `mut` so `remove` probably does re-scaling and other heap garbage. The most performant would be a `VecDeque` and `split_off` probably? The order is now an issue then. I was going to say we don't need to include the `height` (just the `bytes`) if we trust the order, but just in case we probably should and throw an error if it doesn't match the `recorded_block` it is paired with.

MitchTurner: > Anyway, leaving this comment for consideration as we might think about protecting unrecorded_blocks from growing indefinitely in case of some unexpected flow.

Yeah. It's directly relevant to what we're talking about. We definitely make assumption about the order.

Talking to the rollup team guys, it sounds like they might not want to guarantee order in the long run, if that's the case then the `HashMap` approach might be the best. We could even get rid of `SkippedL2Block` I think actually.

xgreenx: I meant: Block committer submits blocks in the bundles. So bundle can have several blocks inside. When you receive the notification from the block commuter about DA submission, you can have `unrecorded_blocks = vec![block_height-5, block_height-4, ... block_height]`. Then instead of removing only one entry, you need to remove all entries up to `block_height`.

If that is something that we want, then `BTreeMap` is better=)

MitchTurner: We are removing all in the bundle. We are just iterating over the recorded blocks:
```rs
    pub fn update_da_record_data(
        &mut self,
        blocks: &[RecordedBlock],
    ) -> Result<(), Error> {
        for block in blocks {
            self.da_block_update(block.height, block.block_cost)?;
        }
        self.recalculate_projected_cost();
        self.normalize_rewards_and_costs();
        Ok(())
    }
```
What I was saying is we could get them all first in some efficient way:
```rs
    pub fn update_da_record_data(
        &mut self,
        blocks: &[RecordedBlock],
    ) -> Result<(), Error> {
        let bytes = self.get_bytes_for_recorded_blocks(&blocks)?;
        for (block, bytes) in blocks.iter().zip(bytes) {
            self.da_block_update(block.height, block.block_cost, bytes)?;
        }
        self.recalculate_projected_cost();
        self.normalize_rewards_and_costs();
        Ok(())
    }
``` 
or something similar without `zip`.

Yeah maybe `BTreeMap` makes the most sense for the ordered case.

xgreenx: `RecordedBlock` data is linked to each other. It just has the latest block height 

cc @rymnc Since he wrote down the interface that we agreed with Rollup team. Maybe I remember it incorrectly)

MitchTurner: Yes. I was part of those conversations and we've discussed what is required for the algo.

rymnc: This is what I have -
```rust
struct Bundle {
	sequence_number: u64,
	blocks_range: Range<BlockHeight>,
	// The DA block height of the last transaciton in the bundle.
	da_block_height: DaBlockHeight
	// Total cost of all bundles for the whole history.
	total_cost: u256,
	// Total size of all bundles for the whole history.
	total_size: u256,
}

trait CommitterAPI {
	fn get_n_last_bundle(&self, number: u64) -> Result<Bundle>;

	// Range is based on `sequence_number`
	fn get_bundles_by_range(&self, range: Range<u64>) -> Result<Vec<Bundle>>;
}
```

crates/fuel-gas-price-algorithm/src/v1/tests.rs
```diff
@@ -14,6 +11,12 @@ mod update_da_record_data_tests;
 #[cfg(test)]
 mod update_l2_block_data_tests;
 
+#[derive(Debug, Clone)]
+pub struct BlockBytes {
```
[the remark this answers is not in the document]
MitchTurner: The one in V0 is not being used anymore. Thanks I'll remove it.

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: Right. Okay. So in that case we can just sum all the block bytes for that range and do a single cost calculation. Sounds like maybe that's what you were suggesting, @xgreenx . I was still under the impression that more processing would happen on the service side, but this actually works fine.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -374,43 +367,57 @@ impl AlgorithmUpdaterV1 {
...
                 .latest_known_total_da_cost_excess
-                .saturating_add(block_cost as u128);
+                .saturating_add(range_cost);
             self.latest_known_total_da_cost_excess = new_block_cost;
             self.latest_da_cost_per_byte = new_cost_per_byte;
             Ok(())
         }
     }
 
+    fn drain_l2_block_bytes_for_range(
```
MitchTurner, line 399: So we can consider improving performance here still.

MitchTurner: Switched to `BTreeMap`

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -374,43 +366,60 @@ impl AlgorithmUpdaterV1 {
...
+        &mut self,
+        height_range: Range<u32>,
+    ) -> Result<u128, Error> {
+        let mut total: u128 = 0;
+        for expected_height in height_range {
+            let (actual_height, bytes) = self
+                .unrecorded_blocks
+                .pop_first()
```
rafal-ch, line 406: Could `split_off()` be even more performant here? Not a big deal, because the ranges will most likely be of reasonable size, but worth having a look at.

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
```diff
@@ -215,7 +213,16 @@ impl Simulator {
                     }
                 },
             );
-        l2_blocks_with_no_da_blocks.chain(da_blocks).collect()
+        let da_block_ranges = da_blocks.into_iter().map(|maybe_recorded_blocks| {
+            maybe_recorded_blocks.map(|list| {
+                let heights_iter = list.iter().map(|(height, _, _)| *height);
+                let min = heights_iter.clone().min().unwrap();
+                let max = heights_iter.max().unwrap();
```
rafal-ch, line 220: Nit:
We could potentially get `min` and `max` with a single pass over the list. But since it's only in the simulator, not the actual production code, we can leave this as is for clarity.

rafal-ch (approved):
LGTM :+1: 

Left two small nits.

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: `split_off` wouldn't do checks that all the values we expect would be there. 

It also gives us the wrong values, so we'd have to swap them every time. i.e. it removes the elements after the split point from the map, and we want it to remove the elements before the split point. It would be nice if there was a `split_off_front`. Probably would still be more performant with `split_off` though.

rafal-ch: `pop_first()` is acceptable.

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
rafal-ch: Can be left as is.

rafal-ch (approved):
LGTM :+1: 

One possible issue for consideration: https://github.com/FuelLabs/fuel-core/pull/2252#discussion_r1775937609

This could be a non-issue or a real deal that we might consider handling as an follow-up issue.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -176,29 +184,17 @@ impl core::ops::Deref for ClampedPercentage {
...
-        blocks: &[RecordedBlock],
+        height_range: Range<u32>,
+        range_cost: u128,
     ) -> Result<(), Error> {
-        for block in blocks {
-            self.da_block_update(block.height, block.block_bytes, block.block_cost)?;
+        if !height_range.is_empty() {
+            self.da_block_update(height_range, range_cost)?;
+            self.recalculate_projected_cost();
+            // self.normalize_rewards_and_costs();
```
Dentosal, line 196: Is this commented-out on purpose?

crates/fuel-gas-price-algorithm/src/v1/tests/update_da_record_data_tests.rs
```diff
@@ -291,69 +278,99 @@ fn update_da_record_data__da_block_updates_projected_total_cost_with_known_and_g
     assert_eq!(actual, expected as u128);
 }
 
+#[derive(Debug, Clone)]
+pub struct RecordedBlock {
+    pub height: u32,
+    // pub block_bytes: u64,
```
Dentosal, line 284: Do we want to keep this?

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: Yes. I added the TODO with the related issue.

rafal-ch:
> > This could be a non-issue or a real deal that we might consider handling as an follow-up issue.
> 
> @rafal-ch It's possible that we get into a situation that this is never shrunk. I'm not sure what the correct way to handle it is.

It could be as simple as having a hardcoded or configured upper limit on unrecorded blocks and reject to accept more.

> I don't think it should be the worry of the gas price algorithm though. If the committer isn't responding for days or weeks, then the issue is going to need to be addressed anyway.

True.
It could also happen due to a bug or an edge case in wiring up of the function calls, even if the committer is responding.

But yeah, since it's unlikely and of rather low impact we can move forward as is and observe how it behaves in production.

rafal-ch (approved):

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
```diff
@@ -104,8 +104,7 @@ impl Simulator {
         capacity: u64,
         max_block_bytes: u64,
         fullness_and_bytes: Vec<(u64, u64)>,
-        // blocks: Enumerate<Zip<Iter<(u64, u64)>, Iter<Option<Vec<RecordedBlock>>>>>,
-        blocks: impl Iterator<Item = (usize, ((u64, u64), &'a Option<Vec<RecordedBlock>>))>,
+        blocks: impl Iterator<Item = (usize, ((u64, u64), &'a Option<(Range<u32>, u128)>))>,
```
Dentosal, line 107: nit: this type seems complicated enough that a named type should make it more readable

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
```diff
@@ -137,13 +136,13 @@ impl Simulator {
...
             // Update DA blocks on the occasion there is one
-            if let Some(da_blocks) = &da_block {
-                let mut total_cost = updater.latest_known_total_da_cost_excess;
-                for block in da_blocks {
-                    total_cost += block.block_cost as u128;
-                    actual_costs.push(total_cost);
+            if let Some((range, cost)) = da_block {
+                for height in range.to_owned() {
+                    updater
+                        .update_da_record_data(height..(height + 1), *cost)
```
Dentosal, line 142:
```suggestion
                        .update_da_record_data(height..=height, *cost)
```

Dentosal (dismissed):

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
MitchTurner: Unfortunately, `RangeInclusive` isn't the same type as `Range`, I went back and forth on whether to do a generic for the few traits I needed or just do `Range` and I figured this was less noise.

crates/services/gas_price_service/src/fuel_gas_price_updater.rs
```diff
@@ -0,0 +1,354 @@
+use crate::{
```
xgreenx, line 1: Do we need changes from this file? Looks like @rymnc already migrated them and V1 will define own updater/service?

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
MitchTurner: Willing to reconsider if generics seems better.

crates/fuel-gas-price-algorithm/gas-price-analysis/src/simulation.rs
MitchTurner: Done.

xgreenx (approved):

rafal-ch (approved):

PR #2254, opened 2024-09-26
Minor improvements to the compression crate

xgreenx:
Minor improvements to the https://github.com/FuelLabs/fuel-core/pull/1609 PR=) Requires decision from @Dentosal

### DIFF
--- a/crates/compression/src/eviction_policy.rs
+++ b/crates/compression/src/eviction_policy.rs
@@ -9,14 +9,14 @@ use crate::ports::EvictorDb;
 pub(crate) struct CacheEvictor<T> {
     /// Set of keys that must not be evicted
     keep_keys: HashSet<RegistryKey>,
-    _marker: std::marker::PhantomData<T>,
+    _keyspace_marker: std::marker::PhantomData<T>,
 }
 
 impl<T> CacheEvictor<T> {
     pub fn new(keep_keys: HashSet<RegistryKey>) -> Self {
         Self {
             keep_keys,
-            _marker: std::marker::PhantomData,
+            _keyspace_marker: std::marker::PhantomData,
         }
     }
 
--- a/crates/compression/src/eviction_policy.rs
+++ b/crates/compression/src/eviction_policy.rs
@@ -7,14 +7,19 @@ use crate::ports::EvictorDb;
 /// Evictor for a single keyspace
 #[derive(Debug)]
 pub(crate) struct CacheEvictor<T> {
-    pub keyspace: std::marker::PhantomData<T>,
     /// Set of keys that must not be evicted
-    pub keep_keys: HashSet<RegistryKey>,
+    keep_keys: HashSet<RegistryKey>,
+    _marker: std::marker::PhantomData<T>,
 }
 
 impl<T> CacheEvictor<T> {
-    /// Get a key, evicting an old value if necessary
-    #[allow(non_snake_case)] // Match names of types exactly
+    pub fn new(keep_keys: HashSet<RegistryKey>) -> Self {
+        Self {
+            keep_keys,
+            _marker: std::marker::PhantomData,
+        }
+    }
+
     pub fn next_key<D>(&mut self, db: &mut D) -> anyhow::Result<RegistryKey>
     where
         D: EvictorDb<T>,
--- a/crates/compression/src/ports.rs
+++ b/crates/compression/src/ports.rs
@@ -18,18 +18,44 @@ use fuel_core_types::{
 /// must be committed atomically, after which block height must be incremented.
 pub trait TemporalRegistry<T> {
     /// Reads a value from the registry at its current height.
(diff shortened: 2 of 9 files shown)


crates/compression/src/eviction_policy.rs
```diff
@@ -7,14 +7,19 @@ use crate::ports::EvictorDb;
 /// Evictor for a single keyspace
 #[derive(Debug)]
 pub(crate) struct CacheEvictor<T> {
-    pub keyspace: std::marker::PhantomData<T>,
     /// Set of keys that must not be evicted
-    pub keep_keys: HashSet<RegistryKey>,
+    keep_keys: HashSet<RegistryKey>,
+    _marker: std::marker::PhantomData<T>,
```
Dentosal, line 12: The type already says "marker", I'd rather describe what it does.

```suggestion
    _keyspace_marker: std::marker::PhantomData<T>,
```

crates/compression/src/eviction_policy.rs
```diff
@@ -7,14 +7,19 @@ use crate::ports::EvictorDb;
...
 }
 
 impl<T> CacheEvictor<T> {
-    /// Get a key, evicting an old value if necessary
-    #[allow(non_snake_case)] // Match names of types exactly
+    pub fn new(keep_keys: HashSet<RegistryKey>) -> Self {
+        Self {
+            keep_keys,
+            _marker: std::marker::PhantomData,
```
Dentosal, line 19:
```suggestion
            _keyspace_marker: std::marker::PhantomData,
```

crates/compression/src/ports.rs
```diff
@@ -18,20 +18,46 @@ use fuel_core_types::{
 /// must be committed atomically, after which block height must be incremented.
 pub trait TemporalRegistry<T> {
     /// Reads a value from the registry at its current height.
-    fn read_registry(&self, key: RegistryKey) -> anyhow::Result<T>;
+    fn read_registry(&self, key: &RegistryKey) -> anyhow::Result<T>;
 
     /// Reads a value from the registry at its current height.
-    fn write_registry(&mut self, key: RegistryKey, value: &T) -> anyhow::Result<()>;
+    fn write_registry(&mut self, key: &RegistryKey, value: &T) -> anyhow::Result<()>;
```
Dentosal, line 24: I don't understand why `RegistryKey` should be passed by reference. It's small enough to fit into a register so performance is likely not the reason here.

crates/compression/src/ports.rs
```diff
@@ -18,20 +18,46 @@ use fuel_core_types::{
...
 
     /// Lookup registry key by the value.
     fn registry_index_lookup(&self, value: &T) -> anyhow::Result<Option<RegistryKey>>;
 }
 
+impl<D, T> TemporalRegistry<T> for &mut D
+where
+    D: TemporalRegistry<T>,
```
Dentosal, line 32: Why?

crates/compression/src/ports.rs
```diff
@@ -56,10 +82,22 @@ pub struct MessageInfo {
     pub data: Vec<u8>,
 }
 
-/// Temporal registry evictor state storage,
-/// currently backed by a `DaCompressionTemporalRegistryEvictor*`
-/// columns in the offchain database.
```
Dentosal, line 61: Not sure why this is removed? It's quite helpful in finding the backing storage. I understand that it breaks the illusion of not depending on the other crate, but in reality they are tightly coupled.

crates/compression/src/ports.rs
xgreenx: Just to avoid `*key` and `&key` in the code=) You can see that now everything just `key`

crates/compression/src/ports.rs
xgreenx: It allows to use `&mut Database` as an input to `CompressCtx` instead of `Database`

crates/compression/src/ports.rs
xgreenx: The trait should be defined without implementation details. The name of the table that is used under the hood inside of the `fuel-core` an be different in the future, or implementation can be different=) We don't need this information here.

We need to describe how this trait could be used, but not how it is used.

crates/compression/src/tables.rs
```diff
@@ -141,47 +230,59 @@ macro_rules! tables {
...
 }
 
-tables!(Address, AssetId, ContractId, ScriptCode, PredicateCode,);
+tables!(
+    address: Address,
+    asset_id: AssetId,
+    contract_id: ContractId,
+    script_code: ScriptCode,
+    predicate_code: PredicateCode
+);
```
Dentosal, line 285: I removed this so the macros are easier to work with. Do you think it makes sense to introduce the `snake_case` names just to get rid of the `#[allow(non_snake_case)]`? Or is there another reason?

PR #2257, opened 2024-09-26
Decouple compression, decompression and registry

xgreenx:
With this approach, we have more macro rules, but the logic is split between different modules, and we can see the hierarchy of dependency between modules(instead of everyone depending on everyone).


crates/compression/src/tables.rs
```diff
@@ -1,299 +0,0 @@
...
-{
-    async fn compress_with(
-        &self,
-        ctx: &mut CompressCtx<D>,
-    ) -> anyhow::Result<CompressedUtxoId> {
-        ctx.db.lookup(*self)
-    }
-}
-
-macro_rules! tables {
```
xgreenx, line 119: I renamed this file to `registry.rs` and it broke diff=(

Dentosal (approved):

PR #2259, opened 2024-09-26
Minimized the number of the tables for metadata and reverse index

xgreenx:
- Minimized the number of tables for metadata and reverse index since we don't need control over this information.
- Use explicitly the `VersionedCompressedBlock` type as the input type for compression and decompression. It is up to the `fuel-core` decide how to represent the final compressed block(postcard or something else).
- Added initialization of the `next_key`.
- Use a more performant codec for each table of the type instead of a Postcard codec.


crates/fuel-core/src/graphql_api/storage/da_compression/script_code_codec.rs
```diff
@@ -0,0 +1,28 @@
...
+    Encode,
+};
+use fuel_core_types::fuel_tx::ScriptCode;
+use std::{
+    borrow::Cow,
+    ops::Deref,
+};
+
+// TODO: Remove this codec when the `ScriptCode` implements
+//  `AsRef<[u8]>` and from `TryFrom<[u8]>` and use `Raw` codec instead.
```
Dentosal, line 12:
```suggestion
// TODO: Remove this codec when the `ScriptCode` implements
//  `AsRef<[u8]>` and `TryFrom<[u8]>` and use `Raw` codec instead.
```

crates/fuel-core/src/graphql_api/storage/da_compression/predicate_code_codec.rs
```diff
@@ -0,0 +1,28 @@
...
+    Encode,
+};
+use fuel_core_types::fuel_tx::input::PredicateCode;
+use std::{
+    borrow::Cow,
+    ops::Deref,
+};
+
+// TODO: Remove this codec when the `PredicateCode` implements
+//  `AsRef<[u8]>` and from `TryFrom<[u8]>` and use `Raw` codec instead.
```
Dentosal, line 12:
```suggestion
// TODO: Remove this codec when the `PredicateCode` implements
//  `AsRef<[u8]>` and `TryFrom<[u8]>` and use `Raw` codec instead.
```

crates/fuel-core/src/graphql_api/storage/da_compression/reverse_key.rs
```diff
@@ -0,0 +1,82 @@
...
+
+impl From<&ContractId> for ReverseKey {
+    fn from(contract_id: &ContractId) -> Self {
+        Self::ContractId(*contract_id)
+    }
+}
+
+impl From<&ScriptCode> for ReverseKey {
+    fn from(script_code: &ScriptCode) -> Self {
+        let hash = fuel_core_types::fuel_crypto::Hasher::hash(script_code.deref());
```
Dentosal, line 57: Why are we using on `fuel-crypto` for this hash? The hash here is implementation detail of the database, and has nothing to do with hashes used in the vm or the protocol.

xgreenx: You are right, it is up to the database to decide how better to hash it=) 

What do you think about using blake2b instead of sha2?

PR #2261, opened 2024-09-27
Replace TxPoolv1 by TxPoolv2

AurelienFT:
WIP


AurelienFT:
Don't like how it shows on github this way, I will go for using v2 everywhere and remove v1 in a next pr.

Issue #2239, opened 2024-09-23
chore(CombinedDatabase): handle errors on rolling back auxiliary dbs appropriately


rymnc:
On https://github.com/FuelLabs/fuel-core/pull/2226/commits/b2aa337905669638a23a2be336eb28c34142dbbf#diff-c9ec92d6a50c589b6cf389155213049e62dffd865cb2aaf85ebf35a1c74298eaR319-R321

we attempt to sync auxiliary databases with the `OnchainDb` however, we do not handle errors while doing so. 

## Definition of Done

- [ ] Rolling back errors are propagated and if on_chain_height / gas_price_height / off_chain_height is unset, swallow the errors

PR #2272, opened 2024-10-01
fix(combined_database): syncing auxiliary databases on startup with custom behaviour

rymnc:
## Linked Issues/PRs

Fixes https://github.com/FuelLabs/fuel-core/issues/2239

## Description

- Doesn't swallow the error blindly now and errors where necessary, this should fix the flakiness as well

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/fuel-core/src/combined_database.rs
+++ b/crates/fuel-core/src/combined_database.rs
@@ -318,26 +318,38 @@ impl CombinedDatabase {
     /// we leave it to the caller to decide how to bring them up to date.
     /// We don't rollback the on-chain database as it is the source of truth.
     /// The target height of the rollback is the latest height of the on-chain database.
-    pub fn sync_aux_db_heights(&self) -> anyhow::Result<()> {
-        let on_chain_height = match self.on_chain().latest_height_from_metadata()? {
-            Some(height) => height,
-            None => return Ok(()), // Exit loop if on-chain height is None
-        };
+    pub fn sync_aux_db_heights<S>(&self, shutdown_listener: &mut S) -> anyhow::Result<()>
+    where
+        S: ShutdownListener,
+    {
+        while !shutdown_listener.is_cancelled() {
+            let on_chain_height = match self.on_chain().latest_height_from_metadata()? {
+                Some(height) => height,
+                None => break, // Exit loop if on-chain height is None
+            };
 
-        let off_chain_height = self.off_chain().latest_height_from_metadata()?;
-        let gas_price_height = self.gas_price().latest_height_from_metadata()?;
+            let off_chain_height = self.off_chain().latest_height_from_metadata()?;
+            let gas_price_height = self.gas_price().latest_height_from_metadata()?;
 
-        // Handle off-chain rollback if necessary
-        if let Some(off_height) = off_chain_height {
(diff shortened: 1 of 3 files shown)


crates/fuel-core/src/combined_database.rs
```diff
@@ -312,17 +312,46 @@ impl CombinedDatabase {
...
+                None => break, // Exit loop if on-chain height is None
+            };
+
+            let off_chain_height = self.off_chain().latest_height_from_metadata()?;
+            let gas_price_height = self.gas_price().latest_height_from_metadata()?;
+
+            // Handle off-chain rollback if necessary
+            if let Some(off_height) = off_chain_height {
+                if off_height > on_chain_height {
+                    self.off_chain().rollback_last_block()?;
```
MitchTurner, line 337: What if the `rollback_last_block` fails after rolling back a couple blocks. Are we okay with mutating the state without correcting it?

MitchTurner (approved):
Approved but with a thought.

crates/fuel-core/src/combined_database.rs
rymnc: good question ~ if we include this refactor, we will also have to ensure the existing function on `CombinedDatabase::rollback_to` will have to be refactored - 
https://github.com/FuelLabs/fuel-core/blob/103ecf6f49db2c04b3c4f88f3d87d8f6ff2d47c2/crates/fuel-core/src/combined_database.rs#L234-L313

a simple way would perhaps be to use a `WriteTransaction` and then commit once done.

rymnc: how does this look @MitchTurner? - 54457c21e3

rymnc: eh, looks like the internal function doesn't actually rollback atomically. we'll have to keep the loop. it rolls back the blocks one-by-one, which makes the `rollback_to` quite misleading as a function name in https://github.com/FuelLabs/fuel-core/blob/b1f2bf445a2e9a270f1c0032f20c680d8e15d251/crates/fuel-core/src/state.rs#L61

cc: @xgreenx for confirmation

rymnc: I'm going to revert [54457c2](https://github.com/FuelLabs/fuel-core/pull/2272/commits/54457c21e3c2cac4a00a4f3c5d01fe07ce4a0ef2) for now

xgreenx: I think we are okay with mutating state since during next launch we continue to rollback the state. It doesn't live as in corrupted state.

xgreenx (approved):

xgreenx:
I'm curious: why don't we want to continue the use of the `rollback_to`? Looks like it does what we need

rymnc:
> I'm curious: why don't we want to continue the use of the rollback_to? Looks like it does what we need

I tried using it and a lot of the tests break because there are instances where the height of the auxiliary databases are lower than onchain height and `rollback_to` throws an error. we explicitly define this behaviour in the docstring :)

xgreenx (approved):

MitchTurner (approved):

PR #2295, opened 2024-10-04
Add decompression traits and a test case

Dentosal:
https://github.com/FuelLabs/fuel-core/pull/1609#discussion_r1783545359

Add a test case for decompressing DA-compressed blocks. FIx `CombinedDb::from_config` to respects `state_rewind_policy` with tmp RocksDB.


crates/fuel-core/src/graphql_api/da_compression.rs
```diff
@@ -65,14 +72,23 @@ where
     Ok(())
 }
 
-struct CompressTx<'a, Tx> {
-    db_tx: &'a mut Tx,
+pub struct DbTx<'a, Tx> {
+    pub db_tx: &'a mut Tx,
+}
```
netrome, line 77: What's the benefit we get from this wrapper? Couldn't we just use `&'a mut Tx` directly in `DecompressDbTx` in the same way it used to be in `CompressTx`?

netrome (approved):
LGTM, one question from me.

crates/fuel-core/src/graphql_api/da_compression.rs
Dentosal: `DbTx` is used to share `TemporalRegistry` implementation between `CompressDbTx` and `DecompressDbTx`. If we'd be using `&'a mut Tx` directly, then the `TemporalRegistry` implentation would have to be duplicated for both of them.

netrome: But couldn't we implement `TemporalRegistry` directly on `&'a mut Tx` instead?

netrome: Ah right but yeah that gets more convoluted since `Tx` is generic, so we'd have to make a blanket implementation or battle with a bunch of extra `where` clauses in any `impl` blocks using this. Sure I see why this might be simpler then.

tests/tests/da_compression.rs
```diff
@@ -68,7 +110,19 @@ async fn can_fetch_da_compressed_block_from_graphql() {
...
+    let mut tx_inner = db.off_chain().clone().into_transaction();
+    let db_tx = DecompressDbTx {
+        db_tx: DbTx {
+            db_tx: &mut tx_inner,
+        },
+        onchain_db: db.on_chain().view_at(&0u32.into()).unwrap(),
+    };
+    let decompressed = decompress(compression_config, db_tx, block).await.unwrap();
+
+    assert!(decompressed.transactions.len() == 2);
```
MitchTurner, line 125: Might as well check that the second tx matches the one submitted.

tests/tests/da_compression.rs
```diff
@@ -68,7 +110,19 @@ async fn can_fetch_da_compressed_block_from_graphql() {
 
     let block = client.da_compressed_block(block_height).await.unwrap();
     let block = block.expect("Unable to get compressed block");
-    let _: VersionedCompressedBlock = postcard::from_bytes(&block).unwrap();
+    let block: VersionedCompressedBlock = postcard::from_bytes(&block).unwrap();
```
MitchTurner, line 113: Why not:
```rs
    let serialized_block = client
        .da_compressed_block(block_height)
        .await
        .unwrap()
        .expect("Unable to get compressed block");
    let block: VersionedCompressedBlock = postcard::from_bytes(&serialized_block).unwrap();
```

MitchTurner (commented):
WIP. I want to give it another, deeper pass.

crates/fuel-core/src/graphql_api/da_compression.rs
```diff
@@ -210,3 +298,78 @@ where
...
+    }
+
+    fn coin(
+        &self,
+        utxo_id: fuel_core_types::fuel_tx::UtxoId,
+    ) -> anyhow::Result<fuel_core_compression::ports::CoinInfo> {
+        let coin = self
+            .onchain_db
+            .storage_as_ref::<fuel_core_storage::tables::Coins>()
+            .get(&dbg!(utxo_id))?
```
MitchTurner, line 350: We probably shouldn't leave the `dbg!`.

tests/tests/da_compression.rs
```diff
@@ -68,7 +110,19 @@ async fn can_fetch_da_compressed_block_from_graphql() {
...
+
+    let db = &srv.shared.database;
+    let mut tx_inner = db.off_chain().clone().into_transaction();
+    let db_tx = DecompressDbTx {
+        db_tx: DbTx {
+            db_tx: &mut tx_inner,
+        },
+        onchain_db: db.on_chain().view_at(&0u32.into()).unwrap(),
+    };
+    let decompressed = decompress(compression_config, db_tx, block).await.unwrap();
```
MitchTurner, line 123: I guess I'm a little confused by using the shared db for decompressing the info. Shouldn't we be able to decompress it with an isolated, clean db?

Maybe that's not the point of this test.

rafal-ch (approved):
LGTM :+1:

tests/tests/da_compression.rs
rymnc: decompression requires some previous indexation to be done on the data, it can't be done from an empty database afaiu

crates/fuel-core/src/graphql_api/da_compression.rs
Dentosal: Definitely. Why isn't the CI catching this?

crates/fuel-core/src/combined_database.rs
```diff
@@ -94,6 +94,19 @@ impl CombinedDatabase {
...
+    /// A test-only temporary rocksdb database with given rewind policy.
+    #[cfg(feature = "rocksdb")]
+    pub fn from_state_rewind_policy(
+        state_rewind_policy: StateRewindPolicy,
+    ) -> DatabaseResult<Self> {
+        Ok(Self {
+            on_chain: Database::rocksdb_temp(state_rewind_policy)?,
+            off_chain: Database::rocksdb_temp(state_rewind_policy)?,
+            relayer: Database::rocksdb_temp(state_rewind_policy)?,
+            gas_price: Default::default(),
```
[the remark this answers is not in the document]
Dentosal: @xgreenx  might know, I have no idea

crates/fuel-core/src/combined_database.rs
```diff
@@ -94,6 +94,19 @@ impl CombinedDatabase {
         })
     }
 
+    /// A test-only temporary rocksdb database with given rewind policy.
+    #[cfg(feature = "rocksdb")]
+    pub fn from_state_rewind_policy(
+        state_rewind_policy: StateRewindPolicy,
+    ) -> DatabaseResult<Self> {
+        Ok(Self {
+            on_chain: Database::rocksdb_temp(state_rewind_policy)?,
```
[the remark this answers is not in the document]
Dentosal: What's the reason behind this?

crates/fuel-core/src/graphql_api/da_compression.rs
```diff
@@ -210,3 +298,78 @@ where
...
+                .copy_from_slice(&tx_index.to_be_bytes());
+            return Ok(fuel_core_types::fuel_tx::UtxoId::new(
+                fuel_core_types::fuel_tx::TxId::from(bytes),
+                0,
+            ));
+        }
+
+        let block_info = self
+            .onchain_db
+            .storage_as_ref::<fuel_core_storage::tables::FuelBlocks>()
```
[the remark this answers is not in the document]
Dentosal: a4c891e62b

tests/tests/da_compression.rs
Dentosal: a4c891e62b

crates/fuel-core/src/combined_database.rs
```diff
@@ -94,6 +94,19 @@ impl CombinedDatabase {
         })
     }
 
+    /// A test-only temporary rocksdb database with given rewind policy.
+    #[cfg(feature = "rocksdb")]
```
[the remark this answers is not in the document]
Dentosal: Done in 5251318e5e

tests/tests/da_compression.rs
Dentosal: Some fields will differ, as the returned tx would have to be executed to generate full values

crates/fuel-core/src/combined_database.rs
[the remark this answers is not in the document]
Dentosal: I reverted this in beaee56e2e since it seems like we're using it with when empty path is provided, and I'm not sure if we need that outside tests as well.

crates/fuel-core/src/combined_database.rs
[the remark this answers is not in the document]
rymnc: because we have a way to clean up the directory after shutdown.. not sure if this method deletes the db after we are done with it.

netrome (approved):

crates/fuel-core/src/combined_database.rs
```diff
@@ -105,6 +105,19 @@ impl CombinedDatabase {
...
+    pub fn from_state_rewind_policy(
```
xgreenx, line 110:
```suggestion
    pub fn temp_database_with_state_rewind_policy(
```

crates/fuel-core/src/combined_database.rs
```diff
@@ -105,6 +105,19 @@ impl CombinedDatabase {
...
+        state_rewind_policy: StateRewindPolicy,
+    ) -> DatabaseResult<Self> {
+        Ok(Self {
+            on_chain: Database::rocksdb_temp(state_rewind_policy)?,
+            off_chain: Database::rocksdb_temp(state_rewind_policy)?,
+            relayer: Database::rocksdb_temp(state_rewind_policy)?,
```
xgreenx, line 116:
```suggestion
            relayer: Default::default(),
```

We don't need state rewind for the relayer database because all data can be restored from L1 and we can't alter the data itself=)

crates/fuel-core/src/combined_database.rs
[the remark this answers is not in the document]
xgreenx: Because gas price table and relayer table doesn't need to support state rewind in the case of the temporary database(right now at least). Because we don't have any logic around it.

tests/tests/da_compression.rs
```diff
@@ -66,9 +108,24 @@ async fn can_fetch_da_compressed_block_from_graphql() {
...
+        .expect("Unable to get compressed block");
+    let block: VersionedCompressedBlock = postcard::from_bytes(&block).unwrap();
+
+    let db = &srv.shared.database;
+    let mut tx_inner = db.off_chain().clone().into_transaction();
+    let db_tx = DecompressDbTx {
+        db_tx: DbTx {
+            db_tx: &mut tx_inner,
+        },
+        onchain_db: db.on_chain().view_at(&0u32.into()).unwrap(),
```
xgreenx, line 124: It would be nice if in the test you highlighted somehow that we use the previous state of the database because it is important detail=)

crates/fuel-core/src/combined_database.rs
Dentosal: aca530adcb

crates/fuel-core/src/combined_database.rs
Dentosal: aca530adcb

tests/tests/da_compression.rs
Dentosal: Added a comment to that effect in aca530adcb

xgreenx (approved):

xgreenx (approved):

netrome (approved):

PR #2299, opened 2024-10-04
Support blobs in predicates

Voxelot:
## Linked Issues/PRs
https://github.com/FuelLabs/fuel-vm/pull/848
https://github.com/FuelLabs/fuel-vm/pull/849

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog
- [x] New behavior is reflected in tests
- [ ] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [x] I have reviewed the code myself

PR #2300, opened 2024-10-05
feat: blob exists query

segfault-magnet:
So that the sdk doesn't get the full blob when it is only interested in whether the blob exists.


crates/client/assets/schema.sdl
```diff
@@ -871,6 +871,12 @@ type Query {
 		"""
 		id: BlobId!
 	): Blob
+	blobExists(
```
Voxelot, line 874: this shouldn't require any changes to the schema

you can just add a method to fuel-core-client that does the same kind of existence query as the ts-sdk

Voxelot: Screenshot 2024-10-05 at 11 01 20 AM

When you query a blob by id and ask for the id back, it will return null if it doesn't exist

xgreenx (commented):
Could you add 2 tests please, for `true` and `false`?=)

xgreenx (approved):

crates/client/src/client/schema/blob.rs
```diff
@@ -26,3 +26,20 @@ pub struct Blob {
...
+
+#[derive(cynic::QueryFragment, Clone, Debug)]
+#[cynic(
+    schema_path = "./assets/schema.sdl",
+    graphql_type = "Query",
+    variables = "BlobByIdArgs"
+)]
+pub struct BlobExistsQuery {
+    #[arguments(id: $id)]
+    pub blob: Option<BlobIdFragment>,
```
rymnc, line 44: is there a reason we don't just return a bool here instead of returning the fragment?
we also need to dimension the query cost for `blob_exists`, as I can see we do it for the `bytecode`: https://github.com/FuelLabs/fuel-core/blob/fa1a049fd0cdde7bc7da6f2164d4a9aef8103d49/crates/fuel-core/src/schema/blob.rs#L31

rymnc (commented):
I'm going to cc @netrome on this PR since he has worked on the blob graphql api

crates/client/src/client/schema/blob.rs
AurelienFT: Complexity added in #2299

rymnc: noticed!

PR #2304, opened 2024-10-06
Add initialization for the genesis base asset contract

xgreenx:
This match initializes storage slots for the genesis base asset contract(because we forgot to add it).
In the next release we will remove it form our codebase=)

Side change:
- Moves `ChangesIterator` to the `fuel-core-storage` crate to allow usage in the tests.

## Checklist

### Before requesting review

### DIFF
--- a/crates/services/executor/src/hacks.rs
+++ b/crates/services/executor/src/hacks.rs
@@ -1,0 +1,133 @@
+use fuel_core_storage::{
+    column::Column,
+    kv_store::KeyValueInspect,
+    tables::ContractsState,
+    transactional::StorageTransaction,
+    ContractsStateKey,
+    StorageAsMut,
+};
+use fuel_core_types::{
+    fuel_tx::Input,
+    fuel_types::{
+        Bytes32,
+        ContractId,
+    },
+    services::executor::Result as ExecutorResult,
+};
+
+struct Slot {
+    key: &'static str,
+    value: &'static str,
+}
+
+const CONTRACT_ID_WITH_BAD_STATE: &str =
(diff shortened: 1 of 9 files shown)


Voxelot (approved):

crates/services/executor/src/hacks.rs
```diff
@@ -0,0 +1,133 @@
...
+    let bad_contract_id: ContractId = CONTRACT_ID_WITH_BAD_STATE.parse().unwrap();
+
+    if contract_id == &bad_contract_id {
+        for slot in STATE.iter() {
+            let storage_key: Bytes32 = slot.key.parse().unwrap();
+
+            let key = ContractsStateKey::new(contract_id, &storage_key);
+            let contains = storage
+                .storage_as_mut::<ContractsState>()
+                .contains_key(&key)?;
```
rymnc, line 76: qq: what was the previous value stored in the keys defined by `STATE`? does zeroing them out fix it?

crates/services/executor/src/hacks.rs
```diff
@@ -0,0 +1,133 @@
...
+        // When
+        maybe_fix_storage(&[input], &mut storage).unwrap();
+
+        // Then
+        let changes = storage.into_changes();
+        let view = ChangesIterator::new(&changes);
+        let entries = view
+            .iter_all::<ContractsState>(None)
+            .map(|r| r.unwrap())
+            .collect::<Vec<_>>();
```
rymnc, line 130: `iter_all_keys` perhaps? but then again, we don't test the values and this length would be the same before and after running `maybe_fix_contract` right?

rymnc (commented):
left a couple qs

PR #2309, opened 2024-10-07
Limit number of concurrent queries to the graphql service

### LINKED ISSUE #2303: Limit number of concurrent queries the GraphQL service
## Definition of done:

The number of queries in GraphQL service is limited.

netrome:
## Linked Issues/PRs
- Closes #2303

## Description
This PR adds a concurrency limit to the GraphQL service using the `ConcurrencyLimitLayer` middleware provided by tower.

## Checklist

### Before requesting review


bin/fuel-core/src/cli/run/graphql.rs
```diff
@@ -28,6 +28,14 @@ pub struct GraphQLArgs {
     #[clap(long = "graphql-max-directives", default_value = "10", env)]
     pub max_queries_directives: usize,
 
+    /// The max number of directives in the query.
+    #[clap(
+        long = "graphql-max-concurrent-queries",
+        default_value = "usize::MAX",
```
Voxelot, line 34: Maybe the default should be something reasonable like 1024?

netrome: Sure, I'm curious why 1024 is a reasonable number here though? Is it your gut feeling or do we have any insights in the expected load on nodes that we can base this number on?

My reasoning about setting `usize::MAX` (though I have corrected it to `Semaphore::MAX_PERMITS` now) is that it is less intrusive for the node operators compared to having an arbitrary limit imposed on them which they may not be aware of.

I can also imagine that some node operators may want to configure this type of limit directly on their load balancer, in which case it's nice not to have a default limit set on our end.

That being said, you have much more insight than I in how this is going to be run once released so I trust your judgement. What are your thoughts?

Voxelot: we should probably base it on a worst case estimate of query load against our recommended node specs in the docs:

https://docs.fuel.network/guides/running-a-node/#hardware-requirements

@xgreenx should be able to help with coming up with the worst case ball-park figure

netrome: Alright, I've set it to 1024 for now and added two integration tests. Let me see if I can find some numbers anywhere, otherwise I agree that if we want to be conservative 1024 feels like a good number. But I guess we're looking for something like $\frac{8\text{GB}}{m}$ where $m$ is the maximum amount of memory we expect a single query to require.

netrome: So I've set the default as 1024, and also marked this change as breaking now.

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Version 0.37.1]
 
+### Changed
+
+#### Breaking
+- [2309](https://github.com/FuelLabs/fuel-core/pull/2309): Limit number of concurrent queries to the graphql service.
```
xgreenx, line 17: We need to move it up

tests/tests/dos.rs
```diff
@@ -378,3 +385,91 @@ async fn complex_queries__2_state_transition_bytecode__query_to_complex() {
...
+    // In an idealized model we should see c = s * n / 2
+    // where
+    //   c = average concurrent request time
+    //   s = single request time
+    //   n = number of request
+    //   2 = the first even natural non-zero number ;)
+    //
+    // However, since this is inherently flaky we divide by 4 instead of 2 to have some margin,
+    // while still maintaining our ability to assert a large deviation between the two measurements.
+    assert!(avg_concurrent_request_time > avg_request_time * num_samples / 4);
```
xgreenx, line 474: Maybe we want to increase number of samples to be `10_000` to be sure that values will be big and remove any divide

xgreenx (commented):
LGTM=)

bin/fuel-core/src/cli/run/graphql.rs
```diff
@@ -28,6 +28,10 @@ pub struct GraphQLArgs {
     #[clap(long = "graphql-max-directives", default_value = "10", env)]
     pub max_queries_directives: usize,
 
+    /// The max number of directives in the query.
```
rymnc, line 31:
```suggestion
    /// The max number of concurrent queries.
```

tests/tests/dos.rs
```diff
@@ -1,11 +1,18 @@
...
+};
+
+use ethers::abi::Tokenizable;
 use fuel_core::service::{
     Config,
     FuelService,
     ServiceTrait,
 };
 use fuel_core_types::blockchain::header::LATEST_STATE_TRANSITION_VERSION;
+use spki::der::asn1::PrintableStringRef;
```
rymnc, line 15: curious where this is being used

tests/tests/dos.rs
```diff
@@ -1,11 +1,18 @@
 #![allow(warnings)]
 
+use std::time::{
+    Duration,
+    Instant,
+};
+
+use ethers::abi::Tokenizable;
```
rymnc, line 8: curious where this is being used

crates/fuel-core/Cargo.toml
```diff
@@ -62,6 +62,7 @@ tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 tokio-rayon = { workspace = true }
 tokio-stream = { workspace = true, features = ["sync"] }
 tokio-util = { workspace = true }
+tower = { version = "0.4", features = ["limit"] }
```
rymnc, line 65: maybe we want to stick to the same version as `tower-http` here? not a blocker

rymnc (commented):
left a few questions

tests/tests/dos.rs
netrome: 10_000 takes a long time. The test already takes a couple of seconds with number of samples being `100`. Though I checked now for a couple of runs and the deviation between the average concurrent request time and the average serial request time multiplied by the number of samples divided by two (s * n / 2) never deviates more than one percent. 

I could instead make an assertion about the relative difference between these two.

netrome: I also use nanoseconds here to reduce the impact of integer division errors.

netrome: Actually using relative difference has more problems in case we get extra delays not accounted for in the model etc. I'd like to stick with the simple equality check for now as the current setup does these two things:

1. It proves that we get a significant slowdown with the concurrency limit in place. This assertion would fail if we don't see a huge slowdown in the same order of magnitude as the expected slowdown caused by the concurrency limit.
2. It has enough margin to hopefully not become one of those annoying flaky tests which occasionally fails in CI.

bin/fuel-core/src/cli/run/graphql.rs
netrome: Thanks for spotting!

tests/tests/dos.rs
netrome: Weird, don't know how that got there :sweat_smile:

tests/tests/dos.rs
netrome: Same

crates/fuel-core/Cargo.toml
netrome: Sure thing. I've bumped `tower-http` in https://github.com/FuelLabs/fuel-core/pull/2309/commits/83103c0834321de0c28d584ff5da873159a4c8a9

tests/tests/dos.rs
netrome: Fixed in https://github.com/FuelLabs/fuel-core/pull/2309/commits/83103c0834321de0c28d584ff5da873159a4c8a9

tests/tests/dos.rs
netrome: Fixed in https://github.com/FuelLabs/fuel-core/pull/2309/commits/83103c0834321de0c28d584ff5da873159a4c8a9

bin/fuel-core/src/cli/run/graphql.rs
netrome: Fixed in https://github.com/FuelLabs/fuel-core/pull/2309/commits/a39da820cb8f5a94724838b4180fd41630c9e3a1

xgreenx (approved):

AurelienFT (approved):
LGTM, ty for the test

Voxelot (approved):

PR #2316, opened 2024-10-08
fix(txpool): Error in GossipsubMessageAcceptance variant docstrings

netrome:
## Description
The documentation for the `Ignore` and `Reject` variants has been mixed up. This PR corrects it.

### DIFF
--- a/crates/types/src/services/p2p.rs
+++ b/crates/types/src/services/p2p.rs
@@ -44,17 +44,14 @@ pub struct GossipsubMessageInfo {
     pub peer_id: PeerId,
 }
 
-// TODO: Maybe we can remove most of types from here directly into P2P
-
-/// Reporting levels on the status of a message received via Gossip
+/// Status of a message received via Gossip
 #[derive(Clone, Copy, PartialEq, Eq, Debug, Hash)]
 pub enum GossipsubMessageAcceptance {
-    /// Report whether the gossiped message is valid and safe to rebroadcast
+    /// The gossiped message is valid and safe to rebroadcast.
     Accept,
-    /// Ignore the received message and prevent further gossiping
+    /// The gossiped message is invalid and should be ignored.
     Ignore,
-    /// Punish the gossip sender for providing invalid
-    /// (or malicious) data and prevent further gossiping
+    /// The gossiped message is invalid and anyone relaying it should be penalized.
     Reject,
 }
 
--- a/crates/types/src/services/p2p.rs
+++ b/crates/types/src/services/p2p.rs
@@ -52,10 +52,10 @@ pub enum GossipsubMessageAcceptance {
     /// Report whether the gossiped message is valid and safe to rebroadcast
     Accept,
     /// Ignore the received message and prevent further gossiping
-    Reject,
+    Ignore,
     /// Punish the gossip sender for providing invalid
     /// (or malicious) data and prevent further gossiping
-    Ignore,
+    Reject,
 }
 
 /// A gossipped message from the network containing all relevant data.


MitchTurner (commented):
And we are certain this isn't swapped in usage elsewhere?

netrome:
> And we are certain this isn't swapped in usage elsewhere?

I'm not certain, but the place that caused my eyebrows to shift is correct. I'll go over all references to these variants and see if I find any other error. But the documentation is incorrect, and this PR fixes that at least.

crates/types/src/services/p2p.rs
```diff
@@ -52,10 +52,10 @@ pub enum GossipsubMessageAcceptance {
     /// Report whether the gossiped message is valid and safe to rebroadcast
     Accept,
     /// Ignore the received message and prevent further gossiping
-    Reject,
+    Ignore,
     /// Punish the gossip sender for providing invalid
     /// (or malicious) data and prevent further gossiping
```
MitchTurner, line 57: I think the comment should mention the message. Also, we can't guarantee that they will be punished by
choosing this type, but we can "indicate" or mark them so our peer reputation can.
```suggestion
    /// Reject received message and indicate that the gossip sender has provided 
    /// invalid (or malicious) data
```

crates/types/src/services/p2p.rs
```diff
@@ -52,10 +52,10 @@ pub enum GossipsubMessageAcceptance {
     /// Report whether the gossiped message is valid and safe to rebroadcast
```
MitchTurner, line 52: "Report whether" is weird wording here. Doesn't it just "Accept" the message?

MitchTurner (commented):
Some nits on the existing comments while we are here.

netrome:
> And we are certain this isn't swapped in usage elsewhere?

Sweeped all references now. There seems to only be two places where the individual variants are referenced outside of test code (and this is extensively referenced from test code). The first place is the construction in `process_insertion_result`, which is correct, and the other place is just a simple mapping between this and the equivalent `MessageAcceptance` struct which is just a 1:1 mapping.

So yes, I am confident that this is correct now.

crates/types/src/services/p2p.rs
netrome: Sounds good to me :+1:

rafal-ch (approved):
LGTM :+1:

crates/types/src/services/p2p.rs
netrome: Agreed. I wonder if we should even use verbs in these comments like "Report". This is a data structure, and not a behavior we're documenting.

I've reworked the comments a bit now. What do you think of the following?
```rust
/// Status of a message received via Gossip
#[derive(Clone, Copy, PartialEq, Eq, Debug, Hash)]
pub enum GossipsubMessageAcceptance {
    /// The gossiped message is valid and safe to rebroadcast.
    Accept,
    /// The gossiped message is invalid and should be ignored.
    Ignore,
    /// The gossiped message is invalid and anyone relaying it should be penalized.
    Reject,
}
```

netrome: Changed in https://github.com/FuelLabs/fuel-core/compare/6c589b8ed39d4cb4e628b08a7c72a2543df61102..7cf9472ca0fa01793a9c025850c778dbd7bf15ff

xgreenx (approved):

crates/types/src/services/p2p.rs
```diff
@@ -44,18 +44,15 @@ pub struct GossipsubMessageInfo {
     pub peer_id: PeerId,
 }
 
-// TODO: Maybe we can remove most of types from here directly into P2P
```
netrome, line 47: Removed this comment because it's not clear exactly what it refers to, and we don't have an issue to track it.

MitchTurner (approved):

PR #2317, opened 2024-10-08
fix(gas_price_service): service name and unused trait impl

rymnc:
## Linked Issues/PRs

- none

## Description

Unused trait impl `RunnableService` moved to tests where it's actually used.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/gas_price_service/src/v0/service.rs
```diff
@@ -194,17 +167,48 @@ mod tests {
...
         Service,
         ServiceRunner,
+        StateWatcher,
     };
     use fuel_core_types::fuel_types::BlockHeight;
     use std::sync::Arc;
     use tokio::sync::mpsc;
 
+    #[async_trait::async_trait]
+    impl<L2, Metadata> RunnableService for GasPriceServiceV0<L2, Metadata>
```
MitchTurner, line 187: Can we run the task in the tests without the `ServiceRunner`?

rymnc: https://github.com/FuelLabs/fuel-core/blob/6979ce6f45a3813ca698f57607db932c8b43f222/crates/services/src/service.rs#L298 needs a `ServiceRunner` to `run_task`, which is private.

MitchTurner: Ah. But wait. Can't we just test `UninitializedTask`? That's the true interface that we deal with, not the task anyway.

rymnc: yes, but we want to test the behaviour of the task itself here

rymnc: there's a pending PR which will have merge conflicts if I add tests for `UninitializedTask` in this PR. let's treat it as a follow up.

MitchTurner: That works for me. 

> yes, but we want to test the behaviour of the task itself here

Just if there is no way to run a task without a service, then it's not a public interface and we should treat the service as the interface for that task.

MitchTurner (approved):

crates/services/gas_price_service/src/v0/service.rs
rymnc: agreed

netrome (approved):

rymnc:
waiting on @xgreenx to approve, please don't merge until then

crates/services/gas_price_service/src/v0/service.rs
xgreenx, line 187: I don't see the point of testing code that is not used anywhere

rymnc: we're not testing the `RunnableService` it's just a helper for testing the logic within the task.

xgreenx: I still don't see why do you need to implement `RunnableService`. The type already implements `RunnableTask`, and you need to test this one=)

rymnc: ref https://github.com/FuelLabs/fuel-core/pull/2317#discussion_r1792035909

xgreenx (approved):

acerone85 (approved):

PR #2322, opened 2024-10-09
fix(genesis): set salt of contract on execution of genesis state configuration

### LINKED ISSUE #2296: It is not possible to request the salt of the genesis contract
## Problem

Requesting salt of the genesis contracts fails because we forgot to update salt table and store this information. We need to add handling of it on the `fuel-core` side. 

Run next query on the testnet:

```
query {
  contract(id: "0x7e2becd64cd598da59b4d1064b711661898656c6b1f4918a787156b8965dc83c") {
    id
    salt
    bytecode
  }
}
```

## Definition of done

The query shouldn't fail for the testnet and any network.
The genesis process should also index contracts in the off chain database. Not only coins and messages.

rymnc:
## Linked Issues/PRs

fixes #2296

## Description

- Implements `AsTable`, `ImportTable` for `graphql::ContractInfo` . 
- Added test to ensure that the query resolves and returns 0 for genesis contracts.

image

question: this doesn't work with an existing database, can anyone point me to the right place to include it?

I'm not sure if 8d0464e651 is a good idea at all.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/fuel-core/src/graphql_api/storage/contracts.rs
```diff
@@ -31,6 +40,24 @@ impl TableWithBlueprint for ContractsInfo {
...
+            })
+            .collect()
+    }
+}
+
+impl AddTable<ContractsInfo> for StateConfig {
+    fn add(&mut self, _entries: Vec<TableEntry<ContractsInfo>>) {
+        // Do not include these
+    }
+}
```
netrome, line 59: All other `AddTable` implemetnations are for `StateConfigBuilder` and not `StateConfig`, I suppose we'd want to do the same here

```suggestion
impl AddTable<ContractsInfo> for StateConfigBuilder {
    fn add(&mut self, _entries: Vec<TableEntry<ContractsInfo>>) {
        // Do not include these
    }
}
```

netrome: Also I don't see where this is used. What do we need this implementation for?

netrome (commented):
I have a few questions so far. While this PR is pretty small, I feel like I don't understand the storage traits well enough to comprehend what's going on here - but I have tried running this and the query works on my local node.

I'll approve when I've acquired more knowledge of this.

crates/fuel-core/src/graphql_api/storage/contracts.rs
rymnc: eh, was left over from my testing, removed in 717ba42ceb

netrome: Thanks!

netrome: I took the liberty to remove this from the description as well

crates/fuel-core/src/service/genesis.rs
```diff
@@ -150,6 +149,23 @@ pub async fn execute_genesis_block(
     Ok(result)
 }
 
+pub async fn recover_missing_tables_from_genesis_state_config(
+    watcher: StateWatcher,
+    config: &Config,
+    db: &CombinedDatabase,
+) -> anyhow::Result<()> {
+    let genesis_block = create_genesis_block(config);
+    let db = db.clone().unsafe_into_genesis();
```
netrome, line 158: Alright, so if I understand this correctly the reason we need to do this "unsafe" conversion is because we're patching a missing genesis state right? So this code path is only useful for existing nodes with the missing genesis tables, but once this is patched this code path is no longer needed?

Do we have a flow for defining database migrations? Having node operators run a separate migration against their persisted databases seems like an alternative approach here.

netrome (dismissed):
Okay this looks good to me. I have one consideration about the general approach but that is more for my own understanding.

crates/fuel-core/src/service/genesis/importer.rs
```diff
@@ -109,6 +109,21 @@ impl SnapshotImporter {
...
+    pub async fn repopulate_maybe_missing_tables(
+        db: CombinedGenesisDatabase,
+        genesis_block: Block,
+        snapshot_reader: SnapshotReader,
+        watcher: StateWatcher,
+    ) -> anyhow::Result<()> {
+        let mut importer = Self::new(db, genesis_block, snapshot_reader, watcher);
+
+        // the below tables were not populated from the genesis snapshot on older versions
+        importer.spawn_worker_off_chain::<ContractsInfo, ContractsInfo>()?;
```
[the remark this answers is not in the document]
netrome: What if someone rolls back their node to a historic state and performs this query? It sounds like what you have implemented in this PR is more correct than just writing to `ContractsInfo` directly.

crates/fuel-core/src/service/genesis.rs
rymnc: yeah the insertion is idempotent though. tradeoff between one read-write and one write

crates/fuel-core/src/service/genesis/importer.rs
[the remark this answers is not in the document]
rymnc: I agree, that's why I did it 😅

netrome (approved):

rymnc:
@xgreenx added https://github.com/FuelLabs/fuel-core/issues/2329 and linked in [11a8a72](https://github.com/FuelLabs/fuel-core/pull/2322/commits/11a8a72d611b8187ea7fc62635fbb6715867002f)

crates/fuel-core/src/database.rs
```diff
@@ -227,6 +227,13 @@ where
         );
         GenesisDatabase::new(self.into_inner().data)
     }
+
+    /// !!!! WARNING !!!!
+    /// This method is unsafe because it doesn't check if the height is already set.
+    /// This allows overriding the genesis state with a new one.
+    pub fn unsafe_into_genesis(self) -> GenesisDatabase<Description> {
+        GenesisDatabase::new(self.into_inner().data)
+    }
```
xgreenx, line 236: We need to add a todo to rework this. It breaks all procession that we have for database right now.

rymnc: already done #2329

netrome (dismissed):

rymnc:
I think we can just hold off on this PR until we have a dedicated strategy of how to deal with genesis insertions. this might be doing more harm than good by introducing tech debt.

netrome:
> I think we can just hold off on this PR until we have a dedicated strategy of how to deal with genesis insertions. this might be doing more harm than good by introducing tech debt.

I was under the impression that we needed to patch this before mainnet. If not, I agree it sounds wise to hold it off.

crates/fuel-core/src/database.rs
xgreenx: I meant in the code itself

Dentosal (approved):

xgreenx (approved):

PR #2323, opened 2024-10-09
Log executor metrics only if enabled in config

rafal-ch:
Draft attempt to use "metrics" config param in the executor.

Not sure if this is the way to go because it adds [metrics](https://github.com/FuelLabs/fuel-core/compare/807_add_more_metrics...807_use_metrics_config_param?expand=1#diff-e95b4e1265412154a4e0238710d201a8cc36af50df083a6055a1730e20d1c4d2R251) field to `ExecutionOptions`, which in turn is a member of the **versioned** `InputSerializationType` and `InputDeserializationType`. This could mean a breaking change.

---
Alternative approach would be to log the metrics in higher layer, ie.: modify the `fn produce_inner<TxSource>()` to get additional `metrics` parameter (we have access to `Config` in all call-sites). Then after either of `native_produce_inner()` or `wasm_produce_inner()` finishes, the metric values could be calculated from the `ExecutionResult`.

### DIFF
--- a/crates/fuel-core/src/executor.rs
+++ b/crates/fuel-core/src/executor.rs
@@ -161,6 +161,8 @@ mod tests {
         pub consensus_parameters: ConsensusParameters,
         /// Print execution backtraces if transaction execution reverts.
         pub backtrace: bool,
+        /// Enables prometheus metrics for execution.
+        pub metrics: bool,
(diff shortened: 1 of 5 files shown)


rafal-ch:
Not doing this, will update the metrics in the importer.

PR #2330, opened 2024-10-10
Reject queries that are recursive during the resolution

xgreenx:
If during resolution of the query we hit the same type, reject further execution. The code to do this check uses `Visitor` logic from the `async-graphql` library, but this logic is private, so I duplicated it in our branch, Later it would be nice to create PR to them in make it public. Or add a way to pass our own query validators into the  `async-graphql`(which will be more performant).

## Checklist

### Before requesting review


Cargo.toml
```diff
@@ -143,3 +143,9 @@ itertools = { version = "0.12", default-features = false }
 insta = "1.8"
 tempfile = "3.4"
 tikv-jemallocator = "0.5"
+
+# Add caching of the RocksDB for local development
+librocksdb-sys = "0.11.0+8.1.1"
+
+[patch.crates-io]
+librocksdb-sys = { git = "https://github.com/FuelLabs/rust-rocksdb/", branch = "release/v0.21.0" }
```
netrome, line 151: Nit: I assume we don't want to commit this part.
```suggestion
```

crates/fuel-core/src/graphql_api/validation_extension.rs
```diff
@@ -0,0 +1,97 @@
...
+
+impl ExtensionFactory for ValidationExtension {
+    fn create(&self) -> Arc<dyn Extension> {
+        Arc::new(ValidationInner::new(self.recursion_limit))
+    }
+}
+
+struct ValidationInner {
+    recursion_limit: usize,
+    visited: parking_lot::Mutex<HashMap<String, usize>>,
```
netrome, line 41: Is this so performance critical that we need the parking lot mutex instead of the std one? If we introduce parking lot, I assume we'd likely want to follow up with using their synchronization primitives everywhere (which could be a separate issue).

crates/fuel-core/src/graphql_api/validation_extension.rs
```diff
@@ -0,0 +1,97 @@
...
+            recursion_limit,
+            visited: parking_lot::Mutex::new(HashMap::new()),
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl Extension for ValidationInner {
+    async fn resolve(
+        &self,
```
netrome, line 56: Houston, we have a problem. This extension is shared between all concurrent query tasks. So if we have multiple requests for the same field, we may accidentally reject one query as recursive.

I tested it with this scenario:

```rust
#[tokio::test]
async fn submitting_multiple_queries_should_not_fail_but_it_does() {
    let query = r#"
    query {
      blocks(last: 1) {
        edges {
          cursor
          node {id}
        }
      }
    }"#;

    let node = FuelService::new_node(Config::local_node()).await.unwrap();
    let url = format!("http://{}/v1/graphql", node.bound_address);
    let client = FuelClient::new(url.clone()).unwrap();

    client
        .submit_and_await_commit(&Transaction::default_test_tx())
        .await
        .unwrap();

    for _ in 0..2 {
        let url = url.clone();
        let handle = tokio::spawn(async move {
            let result = send_graph_ql_query(&url, query).await;
            assert!(!result.contains("Recursion detected"), "{:?}", result);
        });
    }
}
```

which fails when inserting a ```tokio::time::sleep(std::time::Duration::from_millis(100)).await;``` before the `next.run` function.

Unfortunately I don't see a simple fix for this. We could perhaps do something with `tokio::task::id` ([link](https://docs.rs/tokio/latest/tokio/task/fn.id.html)) but that is an unstable feature.

Alternatively, this seems pretty unlikely (I could not trigger the error without inserting the artificial `sleep` call), so we could accept the risk for now and create a follow-up issue for the proper fix.

netrome (commented):
Spotted a race condition that can cause false positive recursion detections. It's not obvious to me whether this is a showstopper or not depending on how pragmatic we want to be. Let me know your thoughts. Meanwhile, I'll think about potential ways to handle this problem.

crates/fuel-core/src/graphql_api/validation_extension.rs
xgreenx: It fails indeed, but not for the reason that you've mentioned=D

Each query has its own copy of the extension, so parallel query will have its own `visited: parking_lot::Mutex<HashMap<String, usize>>`.

But within one query, resolvers for other fields start to run when the resolver future is paused. And it is why test fails=)

```rust
#[tokio::test]
async fn submitting_multiple_queries_should_not_fail_but_it_does() {
    let query = r#"
    query {
      blocks(last: 1) {
        edges {
          cursor
          node {id}
        }
      }
    }"#;

    let node = FuelService::new_node(Config::local_node()).await.unwrap();
    let url = format!("http://{}/v1/graphql", node.bound_address);
    let client = FuelClient::new(url.clone()).unwrap();

    client
        .submit_and_await_commit(&Transaction::default_test_tx())
        .await
        .unwrap();
}
```

netrome: Ah yeah you're right. There's no Arc, so there is no way the Mutex is shared between tasks. But still it's concurrently accessed like you say.

xgreenx: Fixed with one time analyze during parsing of the query

crates/fuel-core/src/graphql_api/validation_extension/visitor.rs
```diff
@@ -0,0 +1,914 @@
+//! The code below is copy from `<https://github.com/async-graphql/async-graphql/blob/master/src/validation/visitor.rs>`.
```
netrome, line 1:
```suggestion
//! The code below is copy from `<https://github.com/async-graphql/async-graphql/blob/da5fea4023b8b043c5295f0a13ba76180d9e2c53/src/validation/visitor.rs>`.
```

crates/fuel-core/src/graphql_api/validation_extension.rs
```diff
@@ -0,0 +1,109 @@
...
+
+impl ExtensionFactory for ValidationExtension {
+    fn create(&self) -> Arc<dyn Extension> {
+        Arc::new(ValidationInner::new(self.recursion_limit))
+    }
+}
+
+struct ValidationInner {
+    recursion_limit: usize,
+    errors: Mutex<Option<Vec<RuleError>>>,
```
netrome, line 50: Do we need the option here? I.e. is it meaningful to differentiate between an empty vec of errors and no errors? Otherwise I think we can simplify this to
```suggestion
    errors: Mutex<Vec<RuleError>>,
```

crates/fuel-core/src/graphql_api/validation_extension/visitor.rs
netrome: Also, ugh. Would be really nice to expose this upstream!

netrome (approved):
Nice stuff, thanks for fixing!

rymnc (commented):
can we use the image generated by this pr on devnet before merging?

rymnc (dismissed):
maybe we need a review from @Dentosal to be able to merge.

approving so the new release isn't blocked

Dentosal (approved):
The code looks fine. I'm not fan of this idea at all, as having GraphQL without recursive ("graph") queries is nonsensical and just worse than a normal REST API. Have we really carried all this complexity for nothing? But given the timelines, this is the practical solution so I'm approving it anyway.

rymnc (approved):

Dentosal (approved):

PR #2324, opened 2024-10-09
Fixing the issue with duplicate transaction synchronization processes

xgreenx:
Ignore peer if we already are syncing transactions from it.
Added metrics for sync and async processor.
Added metrics for all GraphQL queries.

## Checklist
- [x] New behavior is reflected in tests

### Before requesting review
- [x] I have reviewed the code myself

PR #2320, opened 2024-10-08
fix(api_service): exclude health checks from throttling with ConcurrencyLimitLayer

rymnc:
## Linked Issues/PRs

closes #2319

## Description

Uses axum's `merge` chaining operator to add the `ConcurrencyLimit Layer` only on subsets of the routes we serve, and exclude `/health` from it.

We retain `/metrics` to be throttled, because it acquires a lock on the `GLOBAL_REGISTER`, which may prevent Prometheus from reading metrics. This is a double-edged sword though i think.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog
- [x] New behavior is reflected in tests
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [x] I have reviewed the code myself
- [ ] I have created follow-up issues caused by this PR and linked them here

### After merging, notify other teams

[Add or remove entries as needed]

- [ ] [Rust SDK](https://github.com/FuelLabs/fuels-rs/)
- [ ] [Sway compiler](https://github.com/FuelLabs/sway/)
- [ ] [Platform documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+) (for out-of-organization contributors, the person merging the PR will do this)
- [ ] Someone else?

PR #2314, opened 2024-10-08
fix(dummy_da_block_costs): remove dependency on polling_interval and use channel instead

rymnc:
## Linked Issues/PRs

fixes https://github.com/FuelLabs/fuel-core/issues/2312

## Description

- removes dependency on time in da source service tests.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog
- [x] New behavior is reflected in tests
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [x] I have reviewed the code myself
- [ ] I have created follow-up issues caused by this PR and linked them here

### After merging, notify other teams

[Add or remove entries as needed]

- [ ] [Rust SDK](https://github.com/FuelLabs/fuels-rs/)
- [ ] [Sway compiler](https://github.com/FuelLabs/sway/)
- [ ] [Platform documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+) (for out-of-organization contributors, the person merging the PR will do this)
- [ ] Someone else?

PR #2311, opened 2024-10-07
chore(executor): instrument errors to be more meaningful

rymnc:
## Linked Issues/PRs

Issues with `forc deploy` not too relevant without more meaningful errors. should help us debug.

## Description

- Updated the error type `ExecutorError::GasOverflow` with more details on what failed.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog
- [x] New behavior is reflected in tests
- [ ] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [x] I have reviewed the code myself
- [ ] I have created follow-up issues caused by this PR and linked them here

### After merging, notify other teams

[Add or remove entries as needed]

- [ ] [Rust SDK](https://github.com/FuelLabs/fuels-rs/)
- [ ] [Sway compiler](https://github.com/FuelLabs/sway/)
- [ ] [Platform documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+) (for out-of-organization contributors, the person merging the PR will do this)
- [ ] Someone else?

PR #2333, opened 2024-10-11
Release v0.39.0

rafal-ch:
### Added
- [2324](https://github.com/FuelLabs/fuel-core/pull/2324): Added metrics for sync, async processor and for all GraphQL queries.
- [2320](https://github.com/FuelLabs/fuel-core/pull/2320): Added new CLI flag `graphql-max-resolver-recursive-depth` to limit recursion within resolver. The default value it "1".

## Fixed
- [2320](https://github.com/FuelLabs/fuel-core/issues/2320): Prevent `/health` and `/v1/health` from being throttled by the concurrency limiter.
- [2322](https://github.com/FuelLabs/fuel-core/issues/2322): Set the salt of genesis contracts to zero on execution.
- [2324](https://github.com/FuelLabs/fuel-core/pull/2324): Ignore peer if we already are syncing transactions from it.

#### Breaking

- [2320](https://github.com/FuelLabs/fuel-core/pull/2330): Reject queries that are recursive during the resolution of the query.

#### Breaking
- [2311](https://github.com/FuelLabs/fuel-core/pull/2311): Changed the text of the error returned by the executor if gas overflows.

## What's Changed
* chore(executor): instrument errors to be more meaningful by @rymnc in https://github.com/FuelLabs/fuel-core/pull/2311
* fix(dummy_da_block_costs): remove dependency on polling_interval and use channel instead by @rymnc in https://github.com/FuelLabs/fuel-core/pull/2314
* fix(txpool): Error in GossipsubMessageAcceptance variant docstrings by @netrome in https://github.com/FuelLabs/fuel-core/pull/2316
* refactor: Eager returns in txpool_v2::service::Task::run by @netrome in https://github.com/FuelLabs/fuel-core/pull/2325
* fix(api_service): exclude health checks from throttling with ConcurrencyLimitLayer by @rymnc in https://github.com/FuelLabs/fuel-core/pull/2320
* Remove the `normalize_rewards_and_costs()` function by @rafal-ch in https://github.com/FuelLabs/fuel-core/pull/2293
* fix(genesis): set salt of contract on execution of genesis state configuration by @rymnc in https://github.com/FuelLabs/fuel-core/pull/2322
* Fixing the issue with duplicate transaction synchronization processes by @xgreenx in https://github.com/FuelLabs/fuel-core/pull/2324
* Reject queries that are recursive during the resolution by @xgreenx in https://github.com/FuelLabs/fuel-core/pull/2330

**Full Changelog**: https://github.com/FuelLabs/fuel-core/compare/v0.38.0...v0.39.0

### DIFF
--- a/crates/services/upgradable-executor/src/executor.rs
+++ b/crates/services/upgradable-executor/src/executor.rs
@@ -153,7 +153,8 @@ impl<S, R> Executor<S, R> {
         ("0-36-0", 11),
         ("0-37-0", 12),
         ("0-37-1", 13),
-        ("0-38-0", LATEST_STATE_TRANSITION_VERSION),
+        ("0-38-0", 14),
+        ("0-39-0", LATEST_STATE_TRANSITION_VERSION),
     ];
 
     pub fn new(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [Version 0.39.0]
+
 ### Added
 - [2324](https://github.com/FuelLabs/fuel-core/pull/2324): Added metrics for sync, async processor and for all GraphQL queries.
 - [2320](https://github.com/FuelLabs/fuel-core/pull/2320): Added new CLI flag `graphql-max-resolver-recursive-depth` to limit recursion within resolver. The default value it "1".
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3168,7 +3168,7 @@ dependencies = [
 
 [[package]]
 name = "fuel-core"
-version = "0.38.0"
+version = "0.39.0"
 dependencies = [
  "anyhow",
  "assert_matches",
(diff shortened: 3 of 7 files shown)


xgreenx (approved):

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -153,7 +153,8 @@ impl<S, R> Executor<S, R> {
         ("0-36-0", 11),
         ("0-37-0", 12),
         ("0-37-1", 13),
-        ("0-38-0", LATEST_STATE_TRANSITION_VERSION),
+        ("0-38-0", 14),
+        ("0-39-0", LATEST_STATE_TRANSITION_VERSION),
```
acerone85, line 157: Question mostly for me to understand: do we update the genesis state transition version at every release? Or would there be cases when this does not need to be done (e.g. if we don't change the executor at all between two releases)?

acerone85 (approved):

xgreenx (approved):

crates/services/upgradable-executor/src/executor.rs
xgreenx: We do this for each release, but yeah, it is a problem in the case when we need to update only off-chain logic. 

It would be nice if all off-chain logic lived in another place, in this case we don't need to release `fuel-core` often itself

rymnc (approved):

PR #2334, opened 2024-10-11
Prepare the GraphQL service for the switching to `async` methods

xgreenx:
To avoid blocking all threads with heavy sync requests to the database, we decided to move the main database lookups to a separate thread pool.

After we introduced `ReadView` in the GraphQL service, we duplicated all ports for this type to avoid breaking many places in the codebase(plus simplifying the review process of the change that caused the creation of `ReadView`). But after adding the `ReadView` type, we don't need ports on this type anymore since it is an internal database wrapper type of the GraphQL service, and it has full control over how it works. So, it was an old tech debt to remove the usage of ports. The time is come. To simplify the introduction of the `async` and not affect ports with it, this PR decouples ports and `ReadView`.

This PR prepares the switching from `sync` to `async` logic(it will be a follow-up PR).

### Before requesting review


netrome (dismissed):
Nice simplifications. It's good to get rid of all these extra traits and instead define the getters directly on the `ReadView`.

rymnc (dismissed):
ty 🙏

rymnc (approved):
LGTM, can we test this with wallets/sdk in devnet to ensure that any behaviour they have won't change? I'm aware this is an internal change, but we need to see if there is a performance regression/improvement with consumers of the api

PR #2335, opened 2024-10-11
feat: CLI arguments for configuring GraphQL query costs.

### LINKED ISSUE #2318: Make all limits of the of the complexity to be configurable via the CLI
image

netrome:
## Linked Issues/PRs
- Closes #2318 

## Description
This PR adds CLI options to configure graphql query costs dynamically.

## Checklist

### Before requesting review


crates/fuel-core/src/graphql_api/api_service.rs
```diff
@@ -296,6 +298,10 @@ where
     ))
 }
 
+fn initialize_query_costs(config: Config) {
+    QUERY_COSTS.get_or_init(|| config.config.costs);
+}
+
```
rymnc, line 304: would suggest to move this into the part where the `OnceLock` is defined.

tests/tests/dos.rs
```diff
@@ -245,6 +245,23 @@ async fn complex_queries__41_full_block__query_to_complex() {
...
 
+#[tokio::test]
+async fn complex_queries__increased_block_cost__query_too_complex() {
+    let query = FULL_BLOCK_QUERY.to_string();
+    let query = query.replace("$NUMBER_OF_BLOCKS", "1");
+
+    let mut config = Config::local_node();
+    config.graphql_config.costs.block_header *= 41;
+    config.graphql_config.costs.block_transactions *= 41;
+    config.graphql_config.costs.block_header *= 41;
```
rymnc, line 256: do we need all 3 of these to result in the error that we expect below? shouldn't 1 with a very high value work as well?

rymnc, line 256:
```suggestion
    config.graphql_config.costs.block_header *= 1000;
```
passes the test

rymnc (commented):
left some comments and queries

crates/fuel-core/src/graphql_api/api_service.rs
netrome: Yeah that's better. Will do.

crates/fuel-core/src/graphql_api.rs
```diff
@@ -54,16 +58,18 @@ pub struct Costs {
     pub da_compressed_block_read: usize,
 }
 
-pub const QUERY_COSTS: Costs = Costs {
-    // balance_query: 4000,
+impl Default for Costs {
```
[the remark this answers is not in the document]
netrome: Nice, thanks for suggesting!

crates/fuel-core/src/graphql_api/api_service.rs
netrome: Addressed in https://github.com/FuelLabs/fuel-core/pull/2335/commits/58790851f446e19150d1541f850498fe1d0871cd

tests/tests/dos.rs
netrome: Addressed in https://github.com/FuelLabs/fuel-core/pull/2335/commits/58790851f446e19150d1541f850498fe1d0871cd

netrome: Addressed in https://github.com/FuelLabs/fuel-core/pull/2335/commits/58790851f446e19150d1541f850498fe1d0871cd

Changed to use `config.graphql_config.max_queries_complexity` so this test doesn't suddenly fail if that value is modified later as well.

crates/fuel-core/src/graphql_api.rs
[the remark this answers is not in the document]
netrome: Addressed in https://github.com/FuelLabs/fuel-core/pull/2335/commits/58790851f446e19150d1541f850498fe1d0871cd

netrome:
> wonder why the ci is failing for the test included in this pr. lgtm otherwise, the test passes locally.

Yeah, I'll investigate it. It was the same with the previous parameters so something else is going on here but I have no idea what.

netrome:
> Yeah, I'll investigate it. It was the same with the previous parameters so something else is going on here but I have no idea what.

Reproduced locally now. The test only fails for me with `libtest` and not `nextest` when running `cargo test complex_queries` - so it seems to be some sort of interference between the tests. Running it in isolation succeeds with both `libtest` and `nextest`. I suspect that we're querying a node initialized in another test.

netrome:
> Reproduced locally now. The test only fails for me with `libtest` and not `nextest` when running `cargo test complex_queries` - so it seems to be some sort of interference between the tests. Running it in isolation succeeds with both `libtest` and `nextest`. I suspect that we're querying a node initialized in another test.

Alright, so the problem is that libtest will initialize `QUERY_COSTS` once for the entire test suite.

crates/fuel-core/src/graphql_api.rs
```diff
@@ -87,14 +87,14 @@ pub const DEFAULT_QUERY_COSTS: Costs = Costs {
     da_compressed_block_read: 4000,
 };
 
-pub fn query_costs() -> &'static Costs {
-    QUERY_COSTS.get().unwrap_or(&DEFAULT_QUERY_COSTS)
+pub fn query_costs(extract_field: impl FnOnce(&Costs) -> usize) -> usize {
+    extract_field(&QUERY_COSTS.read().unwrap())
```
rymnc, line 91: I don't think we should sunset the `OnceLock` usage, because in regular node operation, we do not expect these values to change once the node has started. either hide it behind a `RwLock` only in the `test-helpers` feature, and if not, it should still use `OnceLock`. you can also force initialization.

- https://github.com/FuelLabs/fuel-core/blob/a952ff44e0c73323f87f0669112f4a76db6dfc61/benches/src/utils.rs#L150 we run these sequentially to prevent env conflicts for example
- https://doc.rust-lang.org/beta/std/sync/struct.OnceLock.html#method.get_mut get_mut

at the start of the test, mutably alter the QUERY_COSTS, and then after set it back to its initial value, this *might* work ?

netrome: We've synced on this and agreed to:

1. Use the `OnceLock` solution.
2. Disallow tests from initializing non-default values.

I've implemented this in https://github.com/FuelLabs/fuel-core/pull/2335/commits/8f93f4f85dcb67774b061d107dd591e2164b5d5a now.

crates/fuel-core/src/graphql_api/api_service.rs
```diff
@@ -220,6 +221,8 @@ where
     OnChain::LatestView: OnChainDatabase,
     OffChain::LatestView: OffChainDatabase,
 {
+    graphql_api::initialize_query_costs(config.config.costs.clone())?;
```
rymnc, line 224: a comment here about not moving this line will be nice. if it is moved later, calls to `query_costs()` will yield the default value instead of the user's configuration.

rymnc: additionally, I have a question about the graphql query complexity annotations, are they evaluated at compile time or runtime?

answered my own question - they are evaluated at runtime within a closure (expanded macro) -

image

I assume here that for every query it will run the closure, which might be a red flag for graphql performance, because of the atomic operations done by `OnceLock::get()` in the `query_costs()` function. 

will defer to @xgreenx here, but we should load test this PR to ensure there isn't a regression.

rymnc: the load testing can happen with the image generated by this pr on devnet and testnet, let's do that first

xgreenx: We don't currently have any load tests for GraphQL. Of course, we need to add them, but I think for now, it is okay. There is only one available option for implementing this change for now.

Alternative way will require unsafe code. If we want to improve performance, we can create a PR into `async-graphql` with proposed solution.

xgreenx (approved):

crates/fuel-core/src/graphql_api/api_service.rs
rymnc: @netrome if you can create a follow up issue for this will gladly approve

AurelienFT (approved):

crates/fuel-core/src/graphql_api/api_service.rs
netrome: Created https://github.com/FuelLabs/fuel-core/issues/2355

PR #2337, opened 2024-10-12
Updated all pagination queries to work with the `Stream` instead of `Iterator`

xgreenx:
This change updates all pagination queries to work with the async stream instead of the sync iterator.

It is preparation for the next PR to move the fetching of transactions, coins, and messages into a separate thread and await the return of the result from this thread. 

Also, this PR limits the maximum number of requested coins from the `coins_to_spend` query to `max_inputs`.

## Checklist

### Before requesting review

### DIFF
--- a/crates/fuel-core/src/graphql_api/database.rs
+++ b/crates/fuel-core/src/graphql_api/database.rs
@@ -61,6 +61,7 @@ use fuel_core_types::{
         txpool::TransactionStatus,
     },
 };
+use futures::Stream;
 use std::{
     borrow::Cow,
     sync::Arc,
@@ -249,8 +250,8 @@ impl ReadView {
         &self,
         start_message_id: Option<Nonce>,
         direction: IterDirection,
-    ) -> BoxedIter<'_, StorageResult<Message>> {
-        self.on_chain.all_messages(start_message_id, direction)
+    ) -> impl Stream<Item = StorageResult<Message>> + '_ {
+        futures::stream::iter(self.on_chain.all_messages(start_message_id, direction))
     }
 
     pub fn message_exists(&self, nonce: &Nonce) -> StorageResult<bool> {
@@ -269,9 +270,12 @@ impl ReadView {
         contract: ContractId,
         start_asset: Option<AssetId>,
         direction: IterDirection,
-    ) -> BoxedIter<StorageResult<ContractBalance>> {
-        self.on_chain
-            .contract_balances(contract, start_asset, direction)
+    ) -> impl Stream<Item = StorageResult<ContractBalance>> + '_ {
+        futures::stream::iter(self.on_chain.contract_balances(
+            contract,
+            start_asset,
+            direction,
+        ))
     }
 
     pub fn da_height(&self) -> StorageResult<DaBlockHeight> {
@@ -306,18 +310,23 @@ impl ReadView {
         owner: &Address,
         start_coin: Option<UtxoId>,
         direction: IterDirection,
-    ) -> BoxedIter<'_, StorageResult<UtxoId>> {
-        self.off_chain.owned_coins_ids(owner, start_coin, direction)
+    ) -> impl Stream<Item = StorageResult<UtxoId>> + '_ {
+        let iter = self.off_chain.owned_coins_ids(owner, start_coin, direction);
+
+        futures::stream::iter(iter)
     }
 
-    pub fn owned_message_ids(
-        &self,
-        owner: &Address,
+    pub fn owned_message_ids<'a>(
+        &'a self,
+        owner: &'a Address,
         start_message_id: Option<Nonce>,
         direction: IterDirection,
-    ) -> BoxedIter<'_, StorageResult<Nonce>> {
-        self.off_chain
-            .owned_message_ids(owner, start_message_id, direction)
+    ) -> impl Stream<Item = StorageResult<Nonce>> + 'a {
+        futures::stream::iter(self.off_chain.owned_message_ids(
+            owner,
+            start_message_id,
+            direction,
+        ))
     }
 
     pub fn owned_transactions_ids(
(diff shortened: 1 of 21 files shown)


crates/fuel-core/src/query/balance.rs
```diff
@@ -28,29 +29,28 @@ use std::{
...
+    pub async fn balance(
         &self,
         owner: Address,
         asset_id: AssetId,
         base_asset_id: AssetId,
     ) -> StorageResult<AddressBalance> {
         let amount = AssetQuery::new(
             &owner,
-            &AssetSpendTarget::new(asset_id, u64::MAX, usize::MAX),
+            &AssetSpendTarget::new(asset_id, u64::MAX, u16::MAX),
```
rymnc, line 40: there's quite a bit of difference between usize::MAX (18446744073709551615) and u16::MAX (65535)
nice optimization

crates/fuel-core/src/query/balance.rs
```diff
@@ -28,29 +29,28 @@ use std::{
...
-
-            // Increase the balance
-            balance = balance.saturating_add(amount);
-
-            Ok(balance)
-        })?;
+        .try_fold(0u64, |balance, amount| {
+            async move {
+                // Increase the balance
+                Ok(balance.saturating_add(amount))
```
rymnc, line 50: this behaviour would be incorrect if the user has a balance of > 18.45 trillion (possible for memecoin projects/?)

rymnc: but it's a major edge case so feel free to ignore

crates/fuel-core/src/query/balance/asset_query.rs
```diff
@@ -196,7 +213,14 @@ impl<'a> AssetQuery<'a> {
 
     /// Returns the iterator over all valid(spendable, allowed by `exclude`) coins of the `owner`
     /// for the `asset_id`.
-    pub fn coins(&self) -> impl Iterator<Item = StorageResult<CoinType>> + '_ {
+    pub fn coins(self) -> impl Stream<Item = StorageResult<CoinType>> + 'a {
         self.query.coins()
     }
 }
+
+fn has_asset(assets: &Option<HashSet<&AssetId>>, asset_id: &AssetId) -> bool {
```
rymnc, line 221: hmm, maybe the signature of this function should be -
```suggestion
fn has_asset(assets: &HashSet<&AssetId>, asset_id: &AssetId) -> bool {
```

and leave it to the caller to decide whether to call or not based on `assets` having a value inside the `Option`.

crates/fuel-core/src/query/balance/asset_query.rs
```diff
@@ -196,7 +213,14 @@ impl<'a> AssetQuery<'a> {
...
+    assets
+        .as_ref()
+        .map(|assets| assets.contains(asset_id))
+        .unwrap_or(true)
```
rymnc, line 225: shouldn't this be false if the Option is `None`?

crates/storage/src/kv_store.rs
```diff
@@ -17,13 +17,8 @@ use core::ops::Deref;
...
 /// The key of the storage.
 pub type Key = Vec<u8>;
-#[cfg(feature = "std")]
 /// The value of the storage. It is wrapped into the `Arc` to provide less cloning of massive objects.
-pub type Value = std::sync::Arc<Vec<u8>>;
-
-#[cfg(not(feature = "std"))]
-/// The value of the storage. It is wrapped into the `Rc` to provide less cloning of massive objects.
-pub type Value = alloc::rc::Rc<Vec<u8>>;
+pub type Value = alloc::sync::Arc<Vec<u8>>;
```
rymnc, line 21: is this change intended to be in this PR? if so, should we use `alloc::sync::Arc` from this point onwards instead of stdlib?

rymnc (commented):
left a few comments and suggestions :)

crates/storage/src/kv_store.rs
xgreenx: `alloc::sync::Arc` and `std::sync::Arc` are the same in the context of `std`. Yes, without this change, it is not possible to compile code because I require `Send` on all `BoxedStreams` now

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -306,28 +310,35 @@ impl ReadView {
...
+
+        futures::stream::iter(iter)
     }
 
-    pub fn owned_message_ids(
-        &self,
-        owner: &Address,
+    pub fn owned_message_ids<'a>(
+        &'a self,
+        owner: &'a Address,
```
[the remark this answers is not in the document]
xgreenx: Have you tried to compile with the suggested change? I remember before it was failing. Plus usually Rust says, when we can remove lifetime during ocmpilation

crates/fuel-core/src/query/balance.rs
xgreenx: It is how it worked before=)

crates/fuel-core/src/graphql_api/database.rs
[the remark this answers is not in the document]
rymnc: yup tried it and ran cargo check too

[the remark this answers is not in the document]
rymnc: image

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -306,28 +310,35 @@ impl ReadView {
...
         start_message_id: Option<Nonce>,
         direction: IterDirection,
-    ) -> BoxedIter<'_, StorageResult<Nonce>> {
-        self.off_chain
-            .owned_message_ids(owner, start_message_id, direction)
+    ) -> impl Stream<Item = StorageResult<Nonce>> + 'a {
```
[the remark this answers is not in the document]
xgreenx: Done in follow up PR https://github.com/FuelLabs/fuel-core/pull/2340/commits/d982fbc2ba14f3ac72249c0fad7ec129db33f727

crates/fuel-core/src/graphql_api/database.rs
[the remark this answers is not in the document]
xgreenx: Done in follow up PR https://github.com/FuelLabs/fuel-core/pull/2340/commits/d982fbc2ba14f3ac72249c0fad7ec129db33f727

crates/fuel-core/src/query/balance/asset_query.rs
xgreenx: https://github.com/FuelLabs/fuel-core/pull/2341/commits/c355cd1569c785379018912aa3ab1a1661d7ee8b

crates/fuel-core/src/query/balance/asset_query.rs
xgreenx: https://github.com/FuelLabs/fuel-core/pull/2341/commits/c355cd1569c785379018912aa3ab1a1661d7ee8b

PR #2340, opened 2024-10-13
Avoid long heavy tasks in the GraphQL service

xgreenx:
This PR adds chunking to the stream in GraphQL and requests data in batches.

The current implementation is simple and executes batches in the same runtime. But at the end of batch fetching, it yields, allowing other tasks to be processed.

## Checklist

### Before requesting review


crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -252,6 +273,13 @@ impl ReadView {
         direction: IterDirection,
     ) -> impl Stream<Item = StorageResult<Message>> + '_ {
         futures::stream::iter(self.on_chain.all_messages(start_message_id, direction))
+            .chunks(self.batch_size)
+            .filter_map(|chunk| async move {
+                // Give a chance to other tasks to run.
+                tokio::task::yield_now().await;
+                Some(futures::stream::iter(chunk))
+            })
```
MitchTurner, line 281: Why not just `map`?

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -134,7 +141,7 @@ impl ReadView {
     pub fn transaction(&self, tx_id: &TxId) -> StorageResult<Transaction> {
         let result = self.on_chain.transaction(tx_id);
         if result.is_not_found() {
-            if let Some(tx) = self.old_transaction(tx_id)? {
+            if let Some(tx) = self.off_chain.old_transaction(tx_id)? {
```
MitchTurner, line 144: I prefer the old helper... although the meaning of "old" here isn't clear to me.

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -252,6 +273,13 @@ impl ReadView {
         direction: IterDirection,
     ) -> impl Stream<Item = StorageResult<Message>> + '_ {
         futures::stream::iter(self.on_chain.all_messages(start_message_id, direction))
+            .chunks(self.batch_size)
+            .filter_map(|chunk| async move {
+                // Give a chance to other tasks to run.
+                tokio::task::yield_now().await;
```
MitchTurner, line 279: Does the stream get treated like one giant future instead of a serious of separate futures? Then I'd understand why we need this, but seems kinda hacky for a situation that seems pretty innocuous.

crates/fuel-core/src/query/block.rs
```diff
@@ -24,5 +27,12 @@ impl ReadView {
         direction: IterDirection,
     ) -> impl Stream<Item = StorageResult<CompressedBlock>> + '_ {
         futures::stream::iter(self.blocks(height, direction))
+            .chunks(self.batch_size)
+            .filter_map(|chunk| async move {
```
MitchTurner, line 31: ditto

crates/fuel-core/src/graphql_api/database.rs
MitchTurner: I've pulled the code and I see this is the way to make it compile. It's annoying and adds misdirection though.

xgreenx: Yes, only `filter_map` allows you to resolve the future=)

bin/fuel-core/src/cli/run/graphql.rs
```diff
@@ -12,6 +12,10 @@ pub struct GraphQLArgs {
     #[clap(long = "port", default_value = "4000", env)]
     pub port: u16,
 
+    /// The size of the batch fetched from the database by GraphQL service.
+    #[clap(long = "graphql-database-batch-size", default_value = "100", env)]
+    pub database_batch_size: usize,
```
rymnc, line 17: does this *have* to be configurable? can we just set it to a global now to reduce the config surface? we have too many options rn :)

xgreenx: I decided to make it configurable for now to be able to adjust our performance on the fly. Later we can remove it when we have better ideas about the best value

crates/fuel-core/src/database/block.rs
```diff
@@ -67,7 +67,7 @@ impl OnChainIterableKeyValueView {
         let db_block = self.storage::<FuelBlocks>().get(height)?;
         if let Some(block) = db_block {
             // fetch all the transactions
-            // TODO: optimize with multi-key get
+            // TODO: Use multiget when it's implemented.
```
netrome, line 70: What's the tracking issue for multiget? Can we reference it here?

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -144,6 +151,20 @@ impl ReadView {
         }
     }
 
+    pub async fn transactions(
+        &self,
+        tx_ids: Vec<TxId>,
+    ) -> Vec<StorageResult<Transaction>> {
+        // TODO: Use multiget when it's implemented.
```
netrome, line 158: Same here

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -144,6 +151,20 @@ impl ReadView {
...
+        let result = tx_ids
+            .iter()
+            .map(|tx_id| self.transaction(tx_id))
+            .collect::<Vec<_>>();
+        // Give a chance to other tasks to run.
+        tokio::task::yield_now().await;
```
netrome, line 164: Nit: I don't think this comment is needed.

```suggestion
        tokio::task::yield_now().await;
```

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -252,6 +273,13 @@ impl ReadView {
         direction: IterDirection,
     ) -> impl Stream<Item = StorageResult<Message>> + '_ {
         futures::stream::iter(self.on_chain.all_messages(start_message_id, direction))
+            .chunks(self.batch_size)
+            .filter_map(|chunk| async move {
+                // Give a chance to other tasks to run.
```
netrome, line 278: Nit: Same here.
```suggestion
```

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -276,6 +304,13 @@ impl ReadView {
             start_asset,
             direction,
         ))
+        .chunks(self.batch_size)
+        .filter_map(|chunk| async move {
+            // Give a chance to other tasks to run.
```
netrome, line 309:
```suggestion
```

crates/fuel-core/src/query/balance.rs
```diff
@@ -106,5 +106,12 @@ impl ReadView {
             })
             .map_ok(|stream| stream.map(Ok))
             .try_flatten()
+            .chunks(self.batch_size)
+            .filter_map(|chunk| async move {
+                // Give a chance to other tasks to run.
```
netrome, line 111:
```suggestion
```

crates/fuel-core/src/query/balance/asset_query.rs
```diff
@@ -1,4 +1,5 @@
 use crate::graphql_api::database::ReadView;
+use async_graphql::futures_util::TryStreamExt;
```
netrome, line 2: Do we need to use the `async_graphql` version of this trait? Otherwise I'd suggest using the one from `futures` directly (which just re-exports the same trait from `futures_util`)

```suggestion
use futures::TryStreamExt;
```

crates/fuel-core/src/query/balance/asset_query.rs
```diff
@@ -78,7 +79,9 @@ impl<'a> AssetsQuery<'a> {
 
     fn coins_iter(mut self) -> impl Stream<Item = StorageResult<CoinType>> + 'a {
         let allowed_assets = self.allowed_assets.take();
-        self.database
+        let database = self.database;
```
netrome, line 82: Nit: I don't see the need for this assignment
```suggestion
```

crates/fuel-core/src/query/balance/asset_query.rs
```diff
@@ -99,12 +102,25 @@ impl<'a> AssetsQuery<'a> {
...
                     };
-                    // TODO: Fetch coin in a separate thread
-                    let coin = self.database.coin(id)?;
-
-                    Ok(CoinType::Coin(coin))
+                    Ok(id)
                 })
+            });
+
+        futures::stream::StreamExt::chunks(stream, database.batch_size)
```
netrome, line 109: We could just refer to `self.database` here
```suggestion
        futures::stream::StreamExt::chunks(stream, self.database.batch_size)
```

crates/fuel-core/src/query/balance/asset_query.rs
```diff
@@ -99,12 +102,25 @@ impl<'a> AssetsQuery<'a> {
...
+            .map(|chunk| {
+                use itertools::Itertools;
+
+                let chunk = chunk.into_iter().try_collect::<_, Vec<_>, _>()?;
+                Ok::<_, StorageError>(chunk)
             })
+            .try_filter_map(move |chunk| async move {
+                let chunk = database
```
netrome, line 117: And here
```suggestion
                let chunk = self.database
```

crates/fuel-core/src/query/coin.rs
```diff
@@ -29,18 +31,35 @@ impl ReadView {
         Ok(coin.uncompress(utxo_id))
     }
 
+    pub async fn coins(
+        &self,
+        utxo_ids: Vec<UtxoId>,
+    ) -> impl Iterator<Item = StorageResult<Coin>> + '_ {
+        // TODO: Use multiget when it's implemented.
+        let coins = utxo_ids.into_iter().map(|id| self.coin(id));
+        // Yield to the runtime to allow other tasks to run.
```
netrome, line 40:
```suggestion
```

crates/fuel-core/src/query/block.rs
```diff
@@ -24,5 +27,12 @@ impl ReadView {
...
+                // Give a chance to other tasks to run.
```
netrome, line 32:
```suggestion
```

crates/fuel-core/src/query/message.rs
```diff
@@ -81,19 +81,33 @@ impl ReadView {
             .map(Cow::into_owned)
     }
 
+    pub async fn messages(
+        &self,
+        ids: Vec<Nonce>,
+    ) -> impl Iterator<Item = StorageResult<Message>> + '_ {
+        let messages = ids.into_iter().map(|id| self.message(&id));
+        // Yield to the runtime to allow other tasks to run.
```
netrome, line 89:
```suggestion
```

netrome (approved):
A bunch of nits, but otherwise looks good to me.

crates/fuel-core/src/database/block.rs
xgreenx: Done

crates/fuel-core/src/graphql_api/database.rs
xgreenx: I think comments are useful if you are not familiar with `tokio::task::yield_now()`. I would prefer to keep them

netrome: Sure, then let's keep them. My thinking is that this is the only purpose of the `yeild_now()` function, so if anyone is unfamiliar with it they'd look it up and then understand that this is the intention. But the comment speeds that process up, so it can be helpful in that sense.

crates/fuel-core/src/graphql_api/database.rs
MitchTurner: Can we just write an extension trait for these spots? :P So we actually have a method that declares what it's doing rather than happening implicitly?

MitchTurner: i.e. roughly:
```rs
trait ResolveStreamFutures {
    type Item;
    fn resolve_each(&self) -> Item;
}
```

and then do a blanket impl of that for `Stream` or `stream::Iter` or whatever.

xgreenx: I think `ResolveStreamFutures` is a bad name, we resolve nothing here, and stream itself already resolves futures=)

crates/fuel-core/src/query/balance/asset_query.rs
MitchTurner: Or add a getter for `batch_size` :)

crates/fuel-core/src/graphql_api/database.rs
MitchTurner: Yes, it could have a better name. I was just putting something down. 

I don't want to have to write a whole alternative to `FilterMap` right now. It's kinda involved. So I'll just let it be. It's hard to understand though.

xgreenx: https://github.com/FuelLabs/fuel-core/pull/2340/commits/727b2da6e460dc2fdb03e56ac01bef7f04f4379d

crates/fuel-core/src/schema/tx.rs
```diff
@@ -188,6 +207,7 @@ impl TxQuery {
         before: Option<String>,
     ) -> async_graphql::Result<Connection<TxPointer, Transaction, EmptyFields, EmptyFields>>
     {
+        use futures::stream::StreamExt;
```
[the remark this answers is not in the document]
rymnc: so that it doesn't collide with other `StreamExt`'s, it's imported within the fn

rymnc (approved):

crates/services/src/yield_stream.rs
```diff
@@ -0,0 +1,164 @@
...
+    },
+};
+
+pin_project_lite::pin_project! {
+    /// Stream that yields each `batch_size` items.
+    #[derive(Debug)]
+    #[must_use = "streams do nothing unless polled"]
+    pub struct YieldStream<St: Stream> {
+        #[pin]
+        stream: Fuse<St>,
```
netrome, line 23: Is there a reason to fuse the stream here or are we just extra cautious? From what I can see, we guarantee that we won't call `poll_next` on this stream after it has returned `Poll::Ready(None)`, unless our `poll_next` is called after it itself has completed - which feels like it should be acceptable behavior, since we're just wrapping the behavior of the underlying stream.

netrome (approved):
Nice stuff! Just one question about fusing the underlying stream in `YieldStream`.

crates/services/src/yield_stream.rs
xgreenx: I tried to duplicate the logic of the `Chunk` iterator and, for stability, decided to continue return `None` if we did that in the past. In this case, the stream is predictable.

crates/fuel-core/src/graphql_api/database.rs
Voxelot: Screenshot 2024-10-14 at 2 31 18 PM

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -144,6 +151,21 @@ impl ReadView {
...
+        tx_ids: Vec<TxId>,
+    ) -> Vec<StorageResult<Transaction>> {
+        // TODO: Use multiget when it's implemented.
+        //  https://github.com/FuelLabs/fuel-core/issues/2344
+        let result = tx_ids
+            .iter()
+            .map(|tx_id| self.transaction(tx_id))
+            .collect::<Vec<_>>();
+        // Give a chance to other tasks to run.
+        tokio::task::yield_now().await;
```
Voxelot, line 165: what is the point of this yield? It's already finished using the database by now.

xgreenx: The idea is that in the future, we will have an async caching mechanism where we will wait for notification when the cache fetches a new desired value(which potentially can be used by several queries in the last block with transactions).

This idle `yield_now` it this behavior also allowing tokio to work with other tasks.

crates/services/src/yield_stream.rs
netrome: Oh I see the chunk iterator now [(link](https://docs.rs/futures-util/0.3.31/src/futures_util/stream/stream/chunks.rs) for reference). Yeah it's very similar. I wonder why they decided to fuse the stream there 🤔

crates/fuel-core/src/query/balance/asset_query.rs
```diff
@@ -99,12 +104,25 @@ impl<'a> AssetsQuery<'a> {
...
                     };
-                    // TODO: Fetch coin in a separate thread
-                    let coin = self.database.coin(id)?;
-
-                    Ok(CoinType::Coin(coin))
+                    Ok(id)
                 })
+            });
+
+        futures::stream::StreamExt::chunks(stream, database.batch_size)
```
Voxelot, line 111: could this all be cleaned up with the new `yield_each` to avoid the need to map chunks and flatten results?

xgreenx: No, here we actually send a batch request to the database like: "Please fetch the all coins for these list of UtxoIds".

Later we can optimize it with multi get + caching

crates/fuel-core/src/query/coin.rs
```diff
@@ -29,18 +31,36 @@ impl ReadView {
         Ok(coin.uncompress(utxo_id))
     }
 
+    pub async fn coins(
```
Voxelot, line 34: is this function really needed? Why not just use yield_each on the stream?

crates/fuel-core/src/query/message.rs
```diff
@@ -81,19 +81,35 @@ impl ReadView {
             .map(Cow::into_owned)
     }
 
+    pub async fn messages(
```
Voxelot, line 84: is this function really needed? Why not just use yield_each on the stream?

crates/fuel-core/src/query/tx.rs
```diff
@@ -32,24 +34,29 @@ impl ReadView {
...
         owner: Address,
         start: Option<TxPointer>,
         direction: IterDirection,
     ) -> impl Stream<Item = StorageResult<(TxPointer, Transaction)>> + '_ {
         self.owned_transactions_ids(owner, start, direction)
-            .map(|result| {
-                result.and_then(|(tx_pointer, tx_id)| {
-                    // TODO: Fetch transactions in a separate thread (https://github.com/FuelLabs/fuel-core/pull/2340)
-                    let tx = self.transaction(&tx_id)?;
+            .chunks(self.batch_size)
```
Voxelot, line 44: why not just use `yield_each`?

crates/fuel-core/src/query/balance/asset_query.rs
Voxelot: Oic, since it might be multiget later?

Voxelot (approved):

crates/fuel-core/src/query/balance/asset_query.rs
xgreenx: Yep

PR #2342, opened 2024-10-14
Do not require build of docker images to pass CI

xgreenx:
Do not require the build of docker images to pass CI. It seems we have some problems with the cache, and right now, CI doesn't use the cache for the docker build process. Until @mchristopher fixes that, I disable it as a requirement to pass CI.


rymnc (approved):
thanks for this

AurelienFT (approved):

Issue #2274, opened 2024-10-01
Add metrics for the graphql complexity


xgreenx:
Track the complexity of each query in the histogram divided by 1k, 5k, 10k, 20k, 50k, 100k, 250k, 500k, 1M, 5M, 10M

tusharnagar17:
Hi @xgreenx,
I would like to work on this. 
Would appreciate any clarification or guidance you can provide ?

AurelienFT:
Done in #2349

PR #2349, opened 2024-10-14
Add graphql query complexity histogram metric

AurelienFT:
## Linked Issues/PRs
Closes https://github.com/FuelLabs/fuel-core/issues/2274

## Checklist

### Before requesting review


xgreenx (approved):

crates/fuel-core/src/graphql_api/metrics_extension.rs
```diff
@@ -118,4 +121,14 @@ impl Extension for MetricsExtInner {
 
         res
     }
+
+    async fn validation(
```
MitchTurner, line 125: It doesn't look like this gets used by anything in this PR... where is it used?

MitchTurner (commented):
One little question. Everything else looks good.

crates/fuel-core/src/graphql_api/metrics_extension.rs
AurelienFT: I'm not very familiar with async-graphql but the trait seems to have a default implementation and I override it. For me this is call by async-graphql itself when triggering validation step of the query

xgreenx: It is how `async-graphql` library works. It has a bunch of default methods that are called during the processing of the request. Valuation is one of the phases there.
image

MitchTurner: Ah. I missed it was a trait impl. That makes sense now.

MitchTurner (approved):

PR #2352, opened 2024-10-14
fix(p2p): cache responses to serve without roundtrip to db

rymnc:
## Linked Issues/PRs

- the intermittent outages on testnet

## Description

~When we request transactions for a given block range, we shouldn't only keep using the same peer causing pressure on it. we should pick a random one with the same height and try to get the transactions from that instead.~

This PR caches p2p responses (ttl 10 seconds by default) and serves requests from cache falling back to db for others.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/p2p/src/service.rs
```diff
@@ -485,12 +498,87 @@ impl<P: TaskP2PService, V, B: Broadcast, T> Task<P, V, B, T> {
...
+}
+
+impl CachedView {
+    fn new(metrics: bool) -> Self {
+        Self {
+            sealed_block_headers: DashMap::new(),
+            transactions_on_blocks: DashMap::new(),
+            metrics,
+        }
+    }
```
rymnc, line 514: we probably want to also support sub-ranges or even partial ranges here, but can be in the future :)

crates/services/p2p/src/cached_view.rs
```diff
@@ -0,0 +1,278 @@
...
+    blockchain::SealedBlockHeader,
+    services::p2p::Transactions,
+};
+use std::ops::Range;
+
+pub struct CachedView {
+    sealed_block_headers: DashMap<Range<u32>, Vec<SealedBlockHeader>>,
+    transactions_on_blocks: DashMap<Range<u32>, Vec<Transactions>>,
+    metrics: bool,
+}
```
netrome, line 18: I'm a bit hesitant to the current approach of storing everything and clearing on a regular interval. Right now, there is no memory limit of the cache, and we use ranges as keys. So if someone queries the ranges (1..=4, 1..=2, 3..=4), we'd store all blocks in the 1..=4 range twice - and this could theoretically grow quadratically for larger ranges.

I would assume that the most popular queries at a given time are quite similar. Why not use a normal LRU cache with fixed memory size? Alternatively just maintain a cache over the last $N$ block headers and their transactions, evicting old ones as new ones gets populated?

netrome (changes requested):
Thanks for implementing this. I Hate to be annoying here, but to approve this I need to:

1. See the Changelog updated.
2. Understand the reasoning behind the current caching strategy and the benefits/drawbacks over an LRU cache.
3. Be certain that we don't open the door to OOM attacks by allowing our cache to be overloaded.

Let me know your thoughts on 2 and 3. I'm happy to jump on a call to discuss this and figure out a good path forward.

crates/services/p2p/src/service.rs
```diff
@@ -781,6 +802,11 @@ where
             AsyncProcessor::new("P2P_TxPoolLookUpProcessor", tx_pool_threads, 32)?;
         let request_sender = broadcast.request_sender.clone();
 
+        let cache_reset_interval = Duration::from_millis(10_000);
+        let next_cache_reset_time = Instant::now()
+            .checked_add(cache_reset_interval)
+            .expect("The cache reset interval should be small enough to do frequently");
```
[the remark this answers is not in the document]
rymnc: the thing with tokio::time::Interval is its burst mode which we don't want to use. I tried to use the other modes but it still triggered the tokio::select branch of the code even when it was not time for it to be triggered. this is more stable.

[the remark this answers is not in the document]
rymnc: this error is qed, it will never throw an error because we control the `cache_reset_interval`

CHANGELOG.md
```diff
@@ -30,6 +30,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2310](https://github.com/FuelLabs/fuel-core/pull/2310): The `metrics` command-line parameter has been replaced with `disable-metrics`. Metrics are now enabled by default, with the option to disable them entirely or on a per-module basis.
 - [2341](https://github.com/FuelLabs/fuel-core/pull/2341): The maximum number of processed coins from the `coins_to_spend` query is limited to `max_inputs`.
 
+### Fixed
+
+- [2352](https://github.com/FuelLabs/fuel-core/pull/2352): Fetches transactions during sync phase from any node that can provide it instead of just 1.
```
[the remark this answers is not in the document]
rymnc: yeah this PR is still in draft haha, working out some things

crates/services/p2p/src/cached_view.rs
rymnc: yup, its still wip.

crates/services/p2p/src/service.rs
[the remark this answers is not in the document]
netrome: Interesting, I would expect `MissedTickBehavior::Delay` to be exactly what we're looking for here. But sure. Yeah the error is basically unreachable, so I can live with an `expect` here. I just prefer^1 errors to be propagated normally, even if they are very esoteric.

[^1]: My preference, not a blocker for merging this PR.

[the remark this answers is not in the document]
rymnc: > to be exactly what we're looking for here 

i understand, went with that approach initially though 😉

crates/services/p2p/src/cached_view.rs
netrome: Ah right, I see this PR is still a draft :)

crates/services/p2p/src/service.rs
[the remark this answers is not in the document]
netrome: Cool, sad it didn't work out 🫠 - though not sure if having the cache cleared based on elapsed time makes sense at all as I've commented in the other thread so let's sort that out first. If you still think the time interval approach makes sense, then I can live with the `sleep_until` and the `expect`.

CHANGELOG.md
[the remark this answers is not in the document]
netrome: Right yeah, no problem holding the changelog update until everything is sorted out 😅

rymnc:
synced a testnet node and had 2 local nodes sync from it at the same time -
image

crates/services/p2p/src/cached_view.rs
rymnc: we now use block height as the key in https://github.com/FuelLabs/fuel-core/pull/2352/commits/64222101e1ace05b7bf4c3b23f44303651f1e0bd

we will retain the time-based eviction strategy for now

crates/services/p2p/src/cached_view.rs
```diff
@@ -0,0 +1,320 @@
...
+        let transactions = view.get_transactions(missing_ranges.clone())?;
+        if let Some(transactions) = &transactions {
+            for (block_height, transactions_per_block) in
+                missing_ranges.zip(transactions.iter())
+            {
+                self.transactions_on_blocks
+                    .insert(block_height, transactions_per_block.clone());
+            }
+        }
+        Ok(transactions)
```
netrome, line 117: Nit: This code is very similar to the one in `get_sealed_headers`. Perhaps it would be nice to factor it out to a separate helper function. Though it's only repeated once so if the helper function gets overly complex with generics and function pointers (e.g. for the metrics functions) then feel free to keep it as-is.

netrome (approved):
Nice stuff! Just a few nits/suggestions from me, but nothing blocking.

crates/services/p2p/Cargo.toml
```diff
@@ -13,6 +13,7 @@ description = "Fuel client networking"
 [dependencies]
 anyhow = { workspace = true }
 async-trait = { workspace = true }
+dashmap = "6.1.0"
```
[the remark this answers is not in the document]
rymnc: papaya looks cool, but lets stick with dashmap for now. if we detect unusual behaviour, then we can test with papaya.

crates/services/p2p/src/cached_view.rs
rymnc: i agree about refactoring it, but its more readable this way. if it's required in the future, (i.e serving more types of data over req/res), we can make it generic.

crates/services/p2p/Cargo.toml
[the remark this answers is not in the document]
netrome: :100:

crates/services/p2p/src/cached_view.rs
netrome: Makes sense :+1:

netrome (approved):

crates/services/p2p/src/cached_view.rs
```diff
@@ -0,0 +1,319 @@
...
+        self.update_metrics(increment_p2p_req_res_cache_misses);
+        let missing_headers = view.get_sealed_headers(missing_ranges.clone())?;
+        if let Some(missing_headers) = &missing_headers {
+            for header in missing_headers.iter() {
+                self.sealed_block_headers
+                    .insert((*header.entity.height()).into(), header.clone());
+                headers.push(header.clone());
+            }
+        }
+        Ok(missing_headers)
```
acerone85, line 78: do you want to return only the missing headers, or all headers (including the ones cached?)

crates/services/p2p/src/cached_view.rs
```diff
@@ -0,0 +1,319 @@
...
+        let transactions = view.get_transactions(missing_ranges.clone())?;
+        if let Some(transactions) = &transactions {
+            for (block_height, transactions_per_block) in
+                missing_ranges.zip(transactions.iter())
+            {
+                self.transactions_on_blocks
+                    .insert(block_height, transactions_per_block.clone());
+            }
+        }
+        Ok(transactions)
```
acerone85, line 116: same here, it looks like only the transactions fetched because of a cache miss are returned?

crates/services/p2p/src/cached_view.rs
```diff
@@ -0,0 +1,319 @@
...
+        &self,
+        view: &V,
+        block_height_range: Range<u32>,
+    ) -> StorageResult<Option<Vec<Transactions>>>
+    where
+        V: P2pDb,
+    {
+        let mut transactions = Vec::new();
+        let mut missing_ranges = block_height_range.clone();
+        for block_height in block_height_range.clone() {
```
acerone85, line 91:
```suggestion
        for block_height in block_height_range {
```
clone here seems unnecessary

crates/services/p2p/src/cached_view.rs
```diff
@@ -0,0 +1,319 @@
...
+            if let Some(header) = self.sealed_block_headers.get(&block_height) {
+                headers.push(header.clone());
+            } else {
+                // for the first block not in the cache, start a new range
+                missing_ranges.start = block_height;
+                break;
+            }
+        }
```
acerone85, line 62: if you are okay with making the iterator mutable, you can simplify this part of the function:

```suggestion
        let mut headers = Vec::new();
        let mut next_header_to_check = block_height_range.start;

        while let Some(header) = block_height_range
            .next()
            .and_then(|block_height| self.sealed_block_headers.get(&block_height))
        {
            headers.push(header.clone());
            next_header_to_check = next_header_to_check.saturating_add(1);
        }

        // If no headers are missing, this range will be empty.
        let missing_ranges = next_header_to_check..block_height_range.end;

```

crates/services/p2p/src/cached_view.rs
```diff
@@ -0,0 +1,319 @@
...
+            for header in missing_headers.iter() {
+                self.sealed_block_headers
+                    .insert((*header.entity.height()).into(), header.clone());
+                headers.push(header.clone());
+            }
+        }
+        Ok(missing_headers)
+    }
+
+    pub(crate) fn get_transactions<V>(
```
acerone85, line 81: get_sealed_headers and get_transactions look very similar. Is there any way to have a single function that takes the dashmap and the fetch from the db functions in input?

crates/services/p2p/src/cached_view.rs
```diff
@@ -0,0 +1,319 @@
...
+            self.update_metrics(increment_p2p_req_res_cache_hits);
+            return Ok(Some(headers))
+        }
+
+        self.update_metrics(increment_p2p_req_res_cache_misses);
+        let missing_headers = view.get_sealed_headers(missing_ranges.clone())?;
+        if let Some(missing_headers) = &missing_headers {
+            for header in missing_headers.iter() {
+                self.sealed_block_headers
+                    .insert((*header.entity.height()).into(), header.clone());
```
acerone85, line 74: In practice this is a contention point in case of concurrent requests. 
If several requests to overlapping ranges happen, all requests might be blocking waiting for a write lock held by another request. 

An alternative can be to read the value first, and then inserting it if `None` is found. Concurrent requests will block on the first writer, but will then be able to proceed concurrently as they will all get a `Read` lock guard.

crates/services/p2p/src/cached_view.rs
rymnc: seeing that even @netrome requested this, addressed in 5ad509354d

crates/services/p2p/src/cached_view.rs
```diff
@@ -0,0 +1,319 @@
...
+        &self,
+        view: &V,
+        block_height_range: Range<u32>,
+    ) -> StorageResult<Option<Vec<SealedBlockHeader>>>
+    where
+        V: P2pDb,
+    {
+        let mut headers = Vec::new();
+        let mut missing_ranges = block_height_range.clone();
+        for block_height in block_height_range.clone() {
```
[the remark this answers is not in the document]
rymnc: actually needed the clone, something was wrong with the cache hit check below. see 5ad509354d

crates/services/p2p/src/cached_view.rs
rymnc: actually needed the clone, something was wrong with the cache hit check below. see https://github.com/FuelLabs/fuel-core/commit/5ad509354d91f94dcf56713f71350138905819f0

crates/services/p2p/src/cached_view.rs
rymnc: addressed in https://github.com/FuelLabs/fuel-core/commit/5ad509354d91f94dcf56713f71350138905819f0

crates/services/p2p/src/cached_view.rs
rymnc: I believe we can defer this to dashmap, since it utilizes sharding under the hood, write contention is kept low. I believe concurrent reads are possible as long as they don't touch the same bucket ~ which we are yet to ascertain the probability of. happy to make the change, but we may want to ignore for now

crates/services/p2p/src/cached_view.rs
```diff
@@ -41,41 +41,64 @@ impl CachedView {
...
             } else {
-                // for the first block not in the cache, start a new range
-                missing_ranges.start = block_height;
+                missing_start = Some(height);
                 break;
             }
         }
 
-        if missing_ranges.is_empty() {
+        if missing_start.is_none() {
```
netrome, line 68: We can remove the sequential coupling (i.e. the safety of the `unwrap` below depending on this check) here by using a `let else` statement.

```suggestion
        let Some(missing_start) = missing_start else {
```

crates/services/p2p/src/cached_view.rs
```diff
@@ -41,41 +41,64 @@ impl CachedView {
...
             self.update_metrics(increment_p2p_req_res_cache_hits);
-            return Ok(Some(headers))
+            return Ok(Some(items));
         }
 
+        let missing_range = missing_start.unwrap()..range.end;
```
netrome, line 73: ...and then we can just skip the unwrap here

```suggestion
        let missing_range = missing_start..range.end;
```

(had to make this two separate comments because GitHub doesn't let me make suggestions for deleted lines)

netrome (approved):
Added a suggestion to get rid of an unwrap that's bugging me right now, otherwise looks good to me.

crates/services/p2p/src/cached_view.rs
rymnc: addressed in [f16cb2f](https://github.com/FuelLabs/fuel-core/pull/2352/commits/f16cb2f0435fa3b478ae615ea4ba908acdabbf2e)

crates/services/p2p/src/cached_view.rs
rymnc: addressed in [f16cb2f](https://github.com/FuelLabs/fuel-core/pull/2352/commits/f16cb2f0435fa3b478ae615ea4ba908acdabbf2e)

crates/services/p2p/src/cached_view.rs
netrome: Wonderful. thanks!

netrome (approved):

rafal-ch (commented):
Partial review

crates/services/p2p/src/service.rs
```diff
@@ -1691,6 +1728,9 @@ pub mod tests {
             heartbeat_max_time_since_last,
             next_check_time: Instant::now(),
             heartbeat_peer_reputation_config: heartbeat_peer_reputation_config.clone(),
+            cached_view: Arc::new(CachedView::new(false)),
+            cache_reset_interval: Duration::from_secs(0),
+            next_cache_reset_time: Instant::now(),
```
rafal-ch, line 1733: I see that this is handled in the same way as pre-existing [heartbeat reputation check](https://github.com/FuelLabs/fuel-core/blob/master/crates/services/p2p/src/service.rs#L963-L973), but I'm wondering if we should allow `0` as the interval.

Is there a risk that the cache will never be cleared?

crates/services/p2p/src/service.rs
```diff
@@ -1601,6 +1635,9 @@ pub mod tests {
             heartbeat_max_time_since_last,
             next_check_time: Instant::now(),
             heartbeat_peer_reputation_config: heartbeat_peer_reputation_config.clone(),
+            cached_view: Arc::new(CachedView::new(false)),
+            cache_reset_interval: Duration::from_secs(0),
```
rafal-ch, line 1639: We could consider adding a test when the reset interval is >0 to check if it is respected.

crates/services/p2p/src/service.rs
rymnc: in https://github.com/FuelLabs/fuel-core/blob/f597ac9db0602caed7e20eaeafed0a1bd2e96ca7/crates/services/p2p/src/service.rs#L828 we define it to be 10sec, we set it to 0 just in the tests for reduced complexity.
the cache *may* not be cleared since it is the least priority future in the `tokio::select` expansion ~ but it will eventually happen if it was missed.

rymnc: good point though, had to reorder the priorities to ensure it clears :)

netrome (approved):

rafal-ch (dismissed):
LGTM :+1:

crates/services/p2p/src/cached_view.rs
```diff
@@ -0,0 +1,320 @@
...
+impl CachedView {
+    pub fn new(metrics: bool) -> Self {
+        Self {
+            sealed_block_headers: DashMap::new(),
+            transactions_on_blocks: DashMap::new(),
+            metrics,
+        }
+    }
+
+    pub(super) fn clear(&self) {
```
xgreenx, line 30: Maybe we could use some LRU cache instead of wiping the whole cache every 10 seconds? This is just a thought since the current approach should also work.

crates/services/p2p/src/service.rs
```diff
@@ -412,6 +413,11 @@ pub struct Task<P, V, B, T> {
     heartbeat_max_time_since_last: Duration,
     next_check_time: Instant,
     heartbeat_peer_reputation_config: HeartbeatPeerReputationConfig,
+    // cached view
+    cached_view: Arc<CachedView>,
+    // milliseconds wait time between cache reset
+    cache_reset_interval: Duration,
+    next_cache_reset_time: Instant,
```
xgreenx, line 420: Looks like it could be an internal logic of the `CachedView` and on each insert/get we can clean it.

crates/services/p2p/src/cached_view.rs
```diff
@@ -0,0 +1,320 @@
...
+        V: P2pDb,
+        T: Clone,
+        F: Fn(&V, Range<u32>) -> StorageResult<Option<Vec<T>>>,
+    {
+        let mut items = Vec::new();
+        let mut missing_start = None;
+
+        for height in range.clone() {
+            if let Some(item) = cache.get(&height) {
+                items.push(item.clone());
```
xgreenx, line 61: It can be follow up PR, but it would be nice if we avoid heavy clone here and used `Arc` instead

crates/services/p2p/src/cached_view.rs
```diff
@@ -0,0 +1,320 @@
...
+    ) -> StorageResult<Option<Vec<T>>>
+    where
+        V: P2pDb,
+        T: Clone,
+        F: Fn(&V, Range<u32>) -> StorageResult<Option<Vec<T>>>,
+    {
+        let mut items = Vec::new();
+        let mut missing_start = None;
+
+        for height in range.clone() {
```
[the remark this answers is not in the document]
acerone85: I think you can go around it by copying the range bounds before consuming the range itself, since those are `Copy`.  

e.g. 

```rust 
let range_copy = range.start..range.end
for height in range { /* ... */ } // No need to clone anymore
```

In this case you only need `range.end`, so it might be better to copy that value only

crates/services/p2p/src/cached_view.rs
rymnc: Yeah, in the future we can use an LRU :) we discussed it somewhere above too

crates/services/p2p/src/cached_view.rs
```diff
@@ -0,0 +1,320 @@
...
+            values: true,
+        };
+        let cached_view = CachedView::new(false);
+
+        // when
+        let notified = sender.notified();
+        let block_height_range = 0..100;
+        let sealed_headers = default_sealed_headers(block_height_range.clone());
+        let result = cached_view
+            .get_sealed_headers(&db, block_height_range.clone())
```
acerone85, line 215: I would expect the cache to be linked to the DB at the time it is created, rather than having to specify the DB when invoking the function `get_sealed_headers` or `get_transactions`. Just curious to know what's the reason behind this choice?

rymnc: you will notice that the view of the current tip of the db (LatestView) is passed into the CachedView while making calls

crates/services/p2p/src/cached_view.rs
rymnc: added comment here - d897cba69c

rymnc: associated issue: https://github.com/FuelLabs/fuel-core/issues/2436

crates/services/p2p/src/service.rs
rymnc: it could be, but I wanted to not make the `get_from_cache_or_db` require a mutable reference to `Self` because it's just a getter. no strong opinion here, so if you want it that way, i can move it around

acerone85:
LGTM. I have a side question of whether the cache should be cleared in case of a DB rollback, to avoid inconsistencies?

rymnc:
> LGTM. I have a side question of whether the cache should be cleared in case of a DB rollback, to avoid inconsistencies?

that's a good question! i wonder if we have a hook from the db to be notified when it gets rolled back.

xgreenx (commented):
> that's a good question! i wonder if we have a hook from the db to be notified when it gets rolled back.

We only can do rollback before services started. We don't need to handle the case when we do rollback during running of the node.

crates/services/p2p/src/service.rs
```diff
@@ -444,7 +450,7 @@ impl<V, T> UninitializedTask<V, SharedState, T> {
     }
 }
 
-impl<P: TaskP2PService, V, B: Broadcast, T> Task<P, V, B, T> {
+impl<P: TaskP2PService, V: AtomicView, B: Broadcast, T> Task<P, V, B, T> {
```
xgreenx, line 453: [nit]: Would be better to move all constraints to `where`

xgreenx (commented):
I think the usage of the concurrent LRU cache would solve all race conditions between updating cache and actually usage of it. Plus it will remove the usage of a separate timer.

Could we try to look into it in this PR? 

The most of the logic of PR remains the same, it will just remove "cleanup" logic.

crates/services/p2p/src/service.rs
rymnc: addressed in b677c01

rymnc:
> I think the usage of the concurrent LRU cache would solve all race conditions between updating cache and actually usage of it. Plus it will remove the usage of a separate timer.

added in [496071f](https://github.com/FuelLabs/fuel-core/pull/2352/commits/496071f27230210e9ef448245faf22e0f6b2c885)

there is still the explicit clone that is done before sending the value as a response though

crates/services/p2p/src/cached_view.rs
```diff
@@ -0,0 +1,432 @@
...
+        if self.cache.len() >= self.capacity {
+            if let Some(least_used) = order.pop_front() {
+                self.cache.remove(&least_used);
+            }
+        }
+
+        self.cache.insert(key.clone(), Arc::new(value));
+
+        // Update the access order.
+        order.retain(|k| k != &key);
```
xgreenx, line 54: I meant to use already existing implementation of the LRU from crates.io. The current implementation is too slow. and operation here is expensive. Plus, usage of the `Mutex` to manage `order` destroyed all benefits of using `DashMap`.

xgreenx (changes requested):

crates/services/p2p/src/cached_view.rs
rymnc: without using `Dashmap`, we will need a mutable reference to the underlying LruCache, from https://crates.io/crates/lru for example. and for every p2p request that comes through, we will need to pass a mutable reference of the `CachedView` https://github.com/FuelLabs/fuel-core/blob/496071f27230210e9ef448245faf22e0f6b2c885/crates/services/p2p/src/service.rs#L581, which isn't possible without a mutex (which is what I assume you want to avoid). This is why I implemented the lru cache using a dashmap.

rymnc: we could use something like https://docs.rs/crossbeam-skiplist/latest/crossbeam_skiplist/ to order the insertions of elements

xgreenx: I don't want lock-free LRU. The `Dashmap` has locks inside, and I'm trying to say that usage of the `Mutex` on upper level removes benefits of using `Dashmap` that uses RW lock inside.

We don't need to re-invent LRU implementation, we just can reuse already existing implementation that works with `&self` without requiring usage of the mutex/rwlock by us explicitly.

rymnc: replaced dashmap with quick_cache which supports LRU-style eviction - 1bdcf0094f

xgreenx (approved):

xgreenx (dismissed):

Issue #2357, opened 2024-10-14
`fuel-core-sync` should cache the result of responses instead of throwing them away


xgreenx:
## Overview

The `fuel-core-sync` service has a single stream of events where processes result sequentially. In several parts of the stream, we request data from another peer in parallel to speed up synchronization.

If any requests fails, it shuts down the stream and cleans everything. The problem is that we throw away results that maybe be useful when we start this stream next time.

## Definition of done

The service doesn't fetch data that already was known and valid and keeps it instead.

AurelienFT:
When you are talking about any requests fails. I don't see a lot of error returned etc except here : https://github.com/FuelLabs/fuel-core/blob/da1f69fa1ef16212473fda4b46d7f32cbb46f255/crates/services/sync/src/import.rs#L321

Is that what you meant by "request fails" ?

xgreenx:
If any of the requests fails, we return `None` and `into_scan_none` will stop the stream, even if we have next 999 blocks by the first block failed.
<img width="1048" alt="image" src="https://github.com/user-attachments/assets/5e238f95-5e29-488d-ad0b-c0bb71082e90">

AurelienFT:
@xgreenx If you have some time this evening I would love a confirmation that this test test the future expected behavior : https://github.com/FuelLabs/fuel-core/blob/704a5837ec4677d40ee4f68d427ef15c9a8ccdda/crates/services/sync/src/import/tests.rs#L251

PR #2361, opened 2024-10-15
Keep data in fails cases in sync service

AurelienFT:
## Linked Issues/PRs
Closes https://github.com/FuelLabs/fuel-core/issues/2357

## Description

This pull request introduces a caching mechanism to the sync service to avoid redundant data fetching from the network. The most important changes include adding a cache module, modifying the `Import` struct to include a cache, and updating related methods to utilize this cache.

### Caching Mechanism:

* [`crates/services/sync/src/import.rs`](diffhunk://#diff-08dae2906fc83b4bb6a03ea474b9e751ef7e290343f66859a669a101c0844a9fR5-R8): Added a new `cache` module and integrated it into the `Import` struct. Updated methods to use the cache for fetching and storing headers and blocks. 
* Cache mechanism allow use to retrieve a stream of batches of either cached headers, cached full blocks, or range to fetch data.

### Test Updates:

* Update the P2P port in mocks to use async to simulate more complex tests needed for this feature.

This PR contains 50% of changes in the tests and addition of tests in the cache.

## Checklist

### Before requesting review


crates/services/sync/src/import.rs
```diff
@@ -98,6 +99,26 @@ pub struct Import<P, E, C> {
     executor: Arc<E>,
     /// Consensus port.
     consensus: Arc<C>,
+    /// A cache of already validated header or blocks.
+    cache: SharedMutex<HashMap<Range<u32>, CachedData>>,
```
netrome, line 103: suggestion: Perhaps we should consider using a [`DashMap`](https://crates.io/crates/dashmap) here instead like in https://github.com/FuelLabs/fuel-core/pull/2352

netrome: Alternatively since we're storing ranges, a `BTreeMap` could be useful. I'm a bit hesitant towards using `Range<u32>` as key here. It's not obvious to me that we'll process exactly the same ranges if the stream is restarted. It would be more robust to instead maintain a map of the block height to the cached data at that height.

Or can we be sure that we're
1. Not storing overlapping ranges and
2. will query the same ranges that we have cached?

crates/services/sync/src/import.rs
```diff
@@ -364,28 +399,72 @@ fn get_block_stream<
     params: Config,
     p2p: Arc<P>,
     consensus: Arc<C>,
+    cache: SharedMutex<HashMap<Range<u32>, CachedData>>,
 ) -> impl Stream<Item = impl Future<Output = SealedBlockBatch>> {
-    let header_stream = get_header_batch_stream(range.clone(), params, p2p.clone());
-    header_stream
+    let ranges = range_chunks(range, params.header_batch_size);
+    futures::stream::iter(ranges)
+        .map({
```
netrome, line 406: While the pattern was established before this PR, I think it would be nice to use `then` instead of `map` here, and skip the `.awaits`. We'd be able to return just a `Stream<Item = SealedBlockBatch>` instead of having the nested futures in the returned stream.

crates/services/sync/src/import.rs
```diff
@@ -118,6 +139,7 @@ impl<P, E, C> Import<P, E, C> {
             p2p,
             executor,
             consensus,
+            cache: SharedMutex::new(HashMap::new()),
```
netrome, line 142: So the cache is created here and lives as long as the `Import` task. How long does that task live? I don't see any clearing logic anywhere. Do we risk OOM errors if this runs for too long?

netrome (commented):
I don't understand the import task well enough to approve right now. I need clarification on the following points:

1. How do we ensure this cache doesn't grow forever? Is the `Import` task short-lived? While the import task launches short-lived streams, it seems like a long-living task to me.
2. How can we be sure we'll query exactly the same ranges as we have cached? Where is that invariant maintained.

Let me know if you want to jump on a call to chat about this, or just write if I'm missing something obvious here.

AurelienFT:
@netrome Thanks for taking the time to review this Regarding your interrogations : 
1 - Yes for me it will leave a long time but all asked data should be ok at some point and so be cleared otherwise we will only have batch_size as number of element in the cache. But I'm not very sure about this that's why I placed a comment about this in "Interrogation" in the PR. Maybe we need a pruning management.
2 - I was thinking that we re-ask all the same ranges because the batch_size doesn't change but the starting point can change to the last synced block and so ranges can change. I will change this to save only block heights

crates/services/sync/src/import.rs
AurelienFT: There is a clear in the `launch_stream` function that should, from my understanding clear all data but i'm not sure (cf https://github.com/FuelLabs/fuel-core/pull/2361#issuecomment-2417970089)

crates/services/sync/src/import.rs
AurelienFT: I can use `DashMap` didn't know about it, seems interesting.

About storing range, I answered here: https://github.com/FuelLabs/fuel-core/pull/2361#issuecomment-2417970089 , however I'm trying to implement it now but the problem is that p2p only works with range and so it's a problem if we have a cache in the middle of the range

crates/services/sync/src/import.rs
netrome: Oh I see now, thanks. Yeah then we pop the ranges once they are consumed, that makes sense. Then we only need to sort out if the ranges are guaranteed to be the same. Otherwise it seems like we'll leak memory.

crates/services/sync/src/import.rs
netrome: Yeah that's why I think a `BTreeMap` is pretty nice since they allow for querying ranges of keys directly.

AurelienFT: YYeah but this doesn't solve the problem that we will have to split ranges asked to p2p and multiply the number of requests

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- [2361](https://github.com/FuelLabs/fuel-core/pull/2361): Add caches to the syync service to not reask for data it already fetched from the network.
```
rafal-ch, line 10: Nit:
```suggestion
- [2361](https://github.com/FuelLabs/fuel-core/pull/2361): Add caches to the sync service to not reask for data it already fetched from the network.
```

crates/services/sync/src/import.rs
```diff
@@ -364,28 +399,72 @@ fn get_block_stream<
     params: Config,
     p2p: Arc<P>,
     consensus: Arc<C>,
+    cache: SharedMutex<HashMap<Range<u32>, CachedData>>,
 ) -> impl Stream<Item = impl Future<Output = SealedBlockBatch>> {
-    let header_stream = get_header_batch_stream(range.clone(), params, p2p.clone());
-    header_stream
+    let ranges = range_chunks(range, params.header_batch_size);
```
rafal-ch, line 404: Stylistic nit, because this function is now too much of a blob for me.

Please consider transforming it into something like the example below, as long as cloning all the arcs and mutexes isn't too much of a hassle.

```
futures::stream::iter(ranges)
        .map({
            move |range| {
                fetch_headers(Arc::clone(&p2p), SharedMutex::clone(&cache), range)
            }
        })
        .map({
            move |header_batch| {
                check_headers(
                    Arc::clone(&p2p),
                    Arc::clone(&consensus),
                    SharedMutex::clone(&cache),
                    header_batch,
                )
            }
        })
        .map({
            move |headers| {
                fetch_blocks(headers, consensus, p2p, cache)
                    .instrument(tracing::debug_span!("consensus_and_transactions"))
                    .in_current_span()
            }
        })
```

crates/services/sync/src/import/back_pressure_tests.rs
```diff
@@ -50,7 +50,7 @@ struct Input {
         block_stream_buffer_size: 10,
         header_batch_size: 5,
     }
-    => is less_or_equal_than Count{ headers: 10, consensus: 10, transactions: 10, executes: 1, blocks: 50 }
+    => is less_or_equal_than Count{ headers: 10, consensus: 10, transactions: 10, executes: 1, blocks: 51 }
```
rafal-ch, line 53: I'm wondering why this change?

rafal-ch (commented):
So far looks good, I need to have a deeper look at the tests though.

AurelienFT:
Convert to draft because of big refacto.

crates/services/sync/src/import.rs
AurelienFT: DashMap isn't an option anymore as we are using BTreeMap

crates/services/sync/src/import.rs
AurelienFT: I agree and there is a lot more things to improve on this service I don't wanna make this PR even bigger and so I created an issue for that : https://github.com/FuelLabs/fuel-core/issues/2370

crates/services/sync/src/import.rs
AurelienFT: I agree and there is a lot more things to improve on this service I don't wanna make this PR even bigger and so I created an issue for that : https://github.com/FuelLabs/fuel-core/issues/2370

crates/services/sync/src/import/back_pressure_tests.rs
AurelienFT: Changed to add more wait on the header verification. From my understanding it was going too fast now and so it was making too much blocks verified.

AurelienFT:
Now everything is cached one by one but there is an issue that I'm having hard time to find a solution. When we successfully had the header but never had the transactions, we need the peer_id to ask the transactions again. However if I cache the peer_id that we used to get the header and failed to give us transactions it will ask him again and I don't think we want to re-ask to someone that returned a fail. But I don't have any ways to find a peer that I know have the transactions.

On top of that the range that I build from cached data could have been fetched from multiple peers.
The only solution I see that simplify everything but cache less things is to cache only full blocks.
Any ideas @netrome @xgreenx @rafal-ch ?

netrome:
> Now everything is cached one by one but there is an issue that I'm having hard time to find a solution. When we successfully had the header but never had the transactions, we need the peer_id to ask the transactions again. However if I cache the peer_id that we used to get the header and failed to give us transactions it will ask him again and I don't think we want to re-ask to someone that returned a fail. But I don't have any ways to find a peer that I know have the transactions.
> 
> On top of that the range that I build from cached data could have been fetched from multiple peers. The only solution I see that simplify everything but cache less things is to cache only full blocks. Any ideas @netrome @xgreenx @rafal-ch ?

Had a chat about this. @xgreenx proposed we change the p2p interface to not require any peer ID when requesting transactions, but instead leave it up to the p2p implementation to decide which peer to request them from and return that peer ID in the response.

crates/services/sync/src/import/cache.rs
```diff
@@ -0,0 +1,533 @@
...
+            }
+            (CachedDataBatch::None(_), CachedData::Block(data)) => {
+                CachedDataBatch::Blocks(Batch::new(
+                    None,
+                    height..height.saturating_add(1),
+                    vec![data],
+                ))
+            }
+            (CachedDataBatch::Headers(mut batch), CachedData::Header(data)) => {
+                batch.range = batch.range.start..height.saturating_add(1);
```
xgreenx, line 147:
```suggestion
                debug_assert_eq!(batch.range.end, height);
                batch.range = batch.range.start..batch.range.end.saturating_add(1);
```

crates/services/sync/src/import/cache.rs
```diff
@@ -0,0 +1,533 @@
...
+                batch.results.push(data);
+                CachedDataBatch::Headers(batch)
+            }
+            (CachedDataBatch::Blocks(mut batch), CachedData::Block(data)) => {
+                batch.range = batch.range.start..height.saturating_add(1);
```
xgreenx, line 152:
```suggestion
                debug_assert_eq!(batch.range.end, height);
                batch.range = batch.range.start..batch.range.end.saturating_add(1);
```

crates/services/sync/src/import/cache.rs
```diff
@@ -0,0 +1,533 @@
...
+            (CachedDataBatch::Headers(headers_batch), CachedData::Block(block)) => {
+                chunks.push(CachedDataBatch::Headers(headers_batch));
+                CachedDataBatch::Blocks(Batch::new(
+                    None,
+                    height..height.saturating_add(1),
+                    vec![block],
+                ))
+            }
+            (CachedDataBatch::Blocks(blocks_batch), CachedData::Header(header)) => {
+                chunks.push(CachedDataBatch::Blocks(blocks_batch));
```
xgreenx, line 165:
```suggestion
                debug_assert_eq!(blocks_batch.range.end, height);
                chunks.push(CachedDataBatch::Blocks(blocks_batch));
```

crates/services/sync/src/import/cache.rs
```diff
@@ -0,0 +1,533 @@
...
+                batch.results.push(data);
+                CachedDataBatch::Headers(batch)
+            }
+            (CachedDataBatch::Blocks(mut batch), CachedData::Block(data)) => {
+                batch.range = batch.range.start..height.saturating_add(1);
+                batch.results.push(data);
+                CachedDataBatch::Blocks(batch)
+            }
+            (CachedDataBatch::Headers(headers_batch), CachedData::Block(block)) => {
+                chunks.push(CachedDataBatch::Headers(headers_batch));
```
xgreenx, line 157:
```suggestion
                debug_assert_eq!(headers_batch.range.end, height);
                chunks.push(CachedDataBatch::Headers(headers_batch));
```

crates/services/sync/src/import/cache.rs
```diff
@@ -0,0 +1,533 @@
...
+        current_chunk: CachedDataBatch,
+        max_chunk_size: usize,
+        chunks: &mut Vec<CachedDataBatch>,
+        current_height: u32,
+    ) -> CachedDataBatch {
+        match current_chunk {
+            CachedDataBatch::Headers(batch) => {
+                if batch.results.len() >= max_chunk_size {
+                    chunks.push(CachedDataBatch::Headers(batch));
+                    CachedDataBatch::None(current_height..current_height)
```
xgreenx, line 186: I don't see why we want to return `None` instead of new `Headers` with remaining elements.

I see that it was extracted from the loop and there it makes sense, because `None` is a default value to start the next iteration of the loop. But here, it looks strange.

I think if we had function with name `truncate_chunk` and did something like `current_chunk = truncate_chunk(current_chunk, &mut chunks)`, it would be simpler to understand=)

crates/services/sync/src/import/cache.rs
```diff
@@ -0,0 +1,533 @@
...
+    }
+
+    fn push_current_chunk(
+        &self,
+        chunks: &mut Vec<CachedDataBatch>,
+        current_chunk: CachedDataBatch,
+    ) {
+        if !current_chunk.is_range_empty() {
+            chunks.push(current_chunk);
+        }
```
xgreenx, line 210: I think moving it to a separate function is overkill, it would be easier to read it in the loop and after the loop, because it is logical check and pretty simple=)

crates/services/sync/src/import/tests.rs
```diff
@@ -227,6 +247,204 @@ async fn import__signature_fails_on_header_5_only() {
...
+        .times(1)
+        .in_sequence(&mut seq)
+        .returning(|range| {
+            Box::pin(async move {
+                let peer = random_peer();
+                let headers = Some(range.map(empty_header).collect());
+                let headers = peer.bind(headers);
+                Ok(headers)
+            })
+        });
```
xgreenx, line 305: Why the sequence is [fail, success, success]? Based on the comments I will expect either [fail, success] or [success(for first 3 blocks), fail, success]. The same question for get transactions.

crates/services/sync/src/import/tests.rs
```diff
@@ -14,7 +14,11 @@ use crate::{
     },
```
xgreenx, line 14: It would be nice to see test where execution fails, and we see that we will not call `p2p` because all data is fetched already.

xgreenx (commented):
The change looks really good=)

crates/services/sync/src/import.rs
```diff
@@ -364,28 +392,60 @@ fn get_block_stream<
...
+                                        header,
+                                        peer.clone(),
+                                        &p2p,
+                                        &consensus,
+                                    )
+                                })
+                                .collect::<Vec<_>>();
+                            let batch = Batch::new(peer, range.clone(), checked_headers);
+                            if !batch.is_err() {
+                                cache.insert_headers(batch.clone());
```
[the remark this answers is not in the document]
AurelienFT: Done : https://github.com/FuelLabs/fuel-core/issues/2461

crates/services/sync/src/import.rs
```diff
@@ -516,27 +563,47 @@ where
...
+            else {
+                report_peer(
+                    p2p,
+                    Some(peer_id.clone()),
+                    PeerReportReason::MissingTransactions,
+                );
+                return None;
+            };
+            Some(SourcePeer { peer_id, data: txs })
+        }
```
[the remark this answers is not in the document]
AurelienFT: I think so because if we are in the case where we don't use cache and we already fetched the header to a particular peer and we have his peer_id it's more optimize to directly asks him the transactions instead of running computation to find someone with these infos (and probably end-up on him also)

crates/services/sync/src/import/tests.rs
AurelienFT: I made a test locally that is failing because this is not the behavior we decided together. I think we said that if execution fails we should remove the data from the cache because it would probably fails again. The line that clears the cache : https://github.com/FuelLabs/fuel-core/blob/a566ac569581447cbcc24cc4677bdb84f9c1c9be/crates/services/sync/src/import.rs#L340

crates/services/sync/src/import/tests.rs
AurelienFT: I have explicit the comments the expected is indeed : [fail (4), success(5), success(6)] and then : [success(4)] only for both of the tests. Tell me if it's more clear :)

crates/services/sync/src/import/cache.rs
AurelienFT: Done

crates/services/sync/src/import/cache.rs
AurelienFT: Done

crates/services/sync/src/import/cache.rs
AurelienFT: Done

crates/services/sync/src/import/cache.rs
AurelienFT: Done

crates/services/sync/src/import/cache.rs
```diff
@@ -0,0 +1,533 @@
...
+                self.push_current_chunk(&mut chunks, current_chunk);
+                self.push_missing_chunks(
+                    &mut chunks,
+                    current_height,
+                    height,
+                    max_chunk_size,
+                    chunk_size_u32,
+                    end,
+                );
+                current_chunk = self.prepare_next_chunk(data, height);
```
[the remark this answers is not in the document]
AurelienFT: I applied you logic but changed `height = current_height` to `current_height = height` and it work and I think that what you meant because `height` is immutable and represent the last cache object we found and it's managed by the iterator

crates/services/sync/src/import/cache.rs
```diff
@@ -0,0 +1,533 @@
...
+        let cache_iter = self.collect_cache_data(range.clone());
+        let mut current_height = *range.start();
+        let mut chunks = Vec::new();
+        let mut current_chunk = CachedDataBatch::None(0..0);
+
+        for (height, data) in cache_iter {
+            if height == current_height {
+                current_chunk =
+                    self.handle_current_chunk(current_chunk, data, height, &mut chunks);
+                current_chunk = self.check_chunk_limit(
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/services/sync/src/import/cache.rs
```diff
@@ -0,0 +1,533 @@
...
+
+        let cache_iter = self.collect_cache_data(range.clone());
+        let mut current_height = *range.start();
+        let mut chunks = Vec::new();
+        let mut current_chunk = CachedDataBatch::None(0..0);
+
+        for (height, data) in cache_iter {
+            if height == current_height {
+                current_chunk =
+                    self.handle_current_chunk(current_chunk, data, height, &mut chunks);
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/services/sync/src/import.rs
```diff
@@ -395,24 +455,39 @@ fn get_block_stream<
...
+                                    SealedHeaderBatch::new(peer, range.clone(), results);
+                                let batch = get_blocks(&p2p, headers).await;
+                                if !batch.is_err() {
+                                    cache.insert_blocks(batch.clone());
+                                }
+                                batch
+                            }
+                        }
+                        BlockHeaderData::Cached(CachedDataBatch::None(_)) => {
+                            unreachable!()
```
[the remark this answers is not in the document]
AurelienFT: I have added a log and I returned a malformed batch which is used as error in this whole process. I don't want to change the whole architecture of the module for this error. (the other solution is to panic like it's done here : https://github.com/FuelLabs/fuel-core/blob/99135e39c0853813bcb0b535252d91e8d90825d9/crates/services/sync/src/import.rs#L475)

crates/services/sync/src/import/cache.rs
AurelienFT: I refactored this to split only when the chunk is inserted in the accumulator and it really simplify the whole code. I added some comments also.

AurelienFT:
@xgreenx Thanks for the kind comment and I have addressed all of your concerns some may still need some answers :)

crates/services/sync/src/import/cache.rs
```diff
@@ -0,0 +1,541 @@
...
+                blocks
+                    .range
+                    .zip(blocks.results)
+                    .chunks(max_chunk_size)
+                    .into_iter()
+                    .map(|chunk| {
+                        let (range, blocks): (Vec<u32>, Vec<SealedBlock>) = chunk.unzip();
+                        CachedDataBatch::Blocks(Batch::new(
+                            None,
+                            range[0]..range[range.len() - 1].saturating_add(1),
```
netrome, line 144: As clippy noted, you'll need to fix the inner subtraction to be saturating, and use `.get()` instead of array slicing here `[]` to avoid potential unexpected panics here.

netrome (changes requested):
Missed this previously, but saw the clippy error in CI regarding the arithmetic operation and want to ensure we avoid the array slicing as well since that can also panic.

crates/services/sync/src/import/tests.rs
xgreenx: It would be nice if the closure inside of the `get_sealed_block_headers` verified the block height like `height == 3`. In this case it will be clear which block is `success` and which is `fail`

crates/services/sync/src/ports.rs
```diff
@@ -34,8 +34,8 @@ pub enum PeerReportReason {
     InvalidTransactions,
 }
 
-#[cfg_attr(any(test, feature = "benchmarking"), mockall::automock)]
 #[async_trait::async_trait]
+#[cfg_attr(any(test, feature = "benchmarking"), mockall::automock)]
```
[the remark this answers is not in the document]
AurelienFT: Yeah it's realted to the PR even if it's not very clear. It's because previously as the code was synchronous the request wasn't in parallel and so it was impossible to test if some work happens while answers from another peer hasn't arrive yet (and end-up with incorrect data but because other job has been done in the same time we have cached data). Now we can sleep during a p2p call : https://github.com/FuelLabs/fuel-core/blob/d9ef69ea61d1f8fe1647092e23ed135cb6109404/crates/services/sync/src/import/tests.rs#L391 that stimulate someone taking a bit more time to answers, the correct one answers directly and after the failed one answers and we re-ask the correct one has been cached.

crates/services/sync/src/import/cache.rs
```diff
@@ -0,0 +1,541 @@
...
+            }
+            (CachedDataBatch::None(_), CachedData::Block(data)) => {
+                CachedDataBatch::Blocks(Batch::new(
+                    None,
+                    height..height.saturating_add(1),
+                    vec![data],
+                ))
+            }
+            (CachedDataBatch::Headers(mut batch), CachedData::Header(data)) => {
+                debug_assert_eq!(batch.range.end, height);
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/services/sync/src/import/cache.rs
```diff
@@ -0,0 +1,541 @@
...
+    ] ; "multiple empty chunks")]
+    #[test_case(&[
+        create_header(0)
+    ], &[], 3, 0..=10 => vec![
+        CachedDataBatch::Headers(Batch::new(None, 0..1, vec![create_header(0)])),
+        CachedDataBatch::None(1..4),
+        CachedDataBatch::None(4..7),
+        CachedDataBatch::None(7..10),
+        CachedDataBatch::None(10..11),
+    ]; "one header and empty ranges")]
```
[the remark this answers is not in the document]
AurelienFT: The max batch 0 created a panic and it makes no sens so I changed to `NonZeroUsize` however I didn't propagate up to the config to not cause too much changes I opened an issue for that :  https://github.com/FuelLabs/fuel-core/issues/2462
For the empty range it's not really possible to do it here because the type on test is `RangeInclusive` which force me to have 1 element at least.

crates/services/sync/src/import/cache.rs
AurelienFT: Done.

netrome (approved):
Thanks for the update :pray:

crates/services/sync/src/import/cache.rs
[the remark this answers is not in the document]
netrome: Ah right, nice :+1:

crates/services/sync/src/import/cache.rs
```diff
@@ -0,0 +1,550 @@
...
+        let mut lock = self.0.lock();
+        for header in batch.results {
+            lock.insert(**header.entity.height(), CachedData::Header(header));
+        }
+    }
+
+    pub fn get_chunks(
+        &self,
+        range: RangeInclusive<u32>,
+        max_chunk_size: NonZeroUsize,
```
xgreenx: We could use `NonZeroU32` and avoid `expect`=)

xgreenx (approved):

AurelienFT:
@xgreenx Why you prefer to make the check inside the loop instead of the split at the end to simplify the logic in the loop ?

xgreenx:
> Why you prefer to make the check inside the loop instead of the split at the end to simplify the logic in the loop ?

It is much easier to read. It is more performant, because the old implementation was doing `into_iter`, `chunks`, `collect` each `batch_size` blocks.

netrome (approved):

Issue #1311, opened 2023-08-21
Return better error messages from P2P requests


MitchTurner:
Over P2P, when a party requests a value and the sender can't fulfill the request, we aren't providing adequate error reporting.

For example, if you request a `Block` by block height and that block doesn't exist, you will receive a `None` with no explanation. It would be better to receive an error explaining that the block for requested height does not exist.

MitchTurner:
We can try using error codes and just make sure we have graceful handling of unknown error codes.

PR #2362, opened 2024-10-15
Add versioning to request response protocols

acerone85:
## Linked Issues/PRs

Related issue: #1311 

## Context
The P2P service uses the [request_response](https://docs.rs/libp2p-request-response/latest/libp2p_request_response/) protocol to let peer request information (e.g. block headers, or transactions in the tx pool) from other peers. 
Currently, the P2P service provides a protocol `/fuel/req_res/0.0.1`, which defines responses to have 
an `Option` value. When a request from a peer is unsuccessful, we send a `None` value as a response, wrapped around the relevant response type. 

In practice, we want to be able to send more meaningful information in responses when requests are unsuccessful, in the form of error codes. This requires implementing a new version of the protocol `/fuel/req_res/0.0.2` where Response messages are changed to contain error codes. 
Because we want to be backward compatible with peers that will only use the version `0.0.1` of the protocol, we must account for requests sent by, and responses sent to, those peers.

## Description
- Rename the `ResponseMessage` type in the P2P service to `LegacyResponseMessage`
- Add a `ResponseMessageErrorCode` enum, which for now only contains a `LegacyNodeEmptyResponse` enum. 
- Implement a new `ResponseMessage` type that returns a `Result<_, ResponseMessageErrorCode>` wrapped around enum variants
- Change P2P service to use the neq `ResponseMessage` type. Responses previously returned as `None` are now returned as `Err<LegacyNodeEmptyResponse>`.
- Change the `PostcardCodec` to serialize and deserialize according to two different versions of the `/fuel/req_res` protocol: `/fuel/req_res/0.0.1` and `/fuel/req_res/0.0.2`. 
- Serialization and deserialization of responses using `/fuel/req_res/0.0.1` are performed by first converting a `ResponseMessage` into a `LegacyResponseMessage` and vicecersa, thus preserving backward-compatibility
- Change the req_res behaviour used by the `P2P` service to use both version of the req_res protocol.

### TODO (here or in different PRs)
- [ ] Investigate how protocols are exchanged during peer handshakes. We don't want two nodes that both understand `/fuel/req_res/0.0.2` to communicate using `/fuel/req_res/0.0.1`. 
- [ ] Add more meaningful error codes.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/p2p/src/request_response/messages.rs
```diff
@@ -32,14 +34,77 @@ pub enum RequestMessage {
     TxPoolFullTransactions(Vec<TxId>),
 }
 
+// TODO: Do we want explicit status codes or an Error type?
```
AurelienFT, line 37: I think it's not a big problem to have more detailed error (with specific block id or thing like that) as soon as we add a limit on the size of the error.

AurelienFT (approved):

crates/services/p2p/src/codecs/postcard.rs
```diff
@@ -160,30 +176,200 @@ impl GossipsubCodec for PostcardCodec {
...
+        codec
+            .write_response(&PostcardProtocol::V1, &mut buf, response.clone())
+            .await
+            .expect("Valid Vec<SealedBlockHeader> is serialized using v1");
+
+        let deserialized_as_legacy =
+            // We cannot access the codec trait from an old node here, 
+            // so we deserialize directly using the `LegacyResponseMessage` type.
+            deserialize::<LegacyResponseMessage>(&buf).expect("Deserialization as LegacyResponseMessage should succeed");
+        match deserialized_as_legacy {
```
MitchTurner, line 347: I'd prefer a `matches!` macro here (and probably on all the tests) because when I'm reading a test I'm always looking for an explicit expected condition, not an implicit one like here.

crates/services/p2p/src/codecs/postcard.rs
```diff
@@ -160,30 +176,200 @@ impl GossipsubCodec for PostcardCodec {
...
+            LegacyResponseMessage::SealedHeaders(None) => {}
+            other => {
+                panic!("Deserialized to {other:?}, expected {response:?}")
+            }
+        }
+    }
```
MitchTurner, line 353: Another example of how to structure the tests:
```suggestion
    #[tokio::test]
    async fn codec__write_response_is_backwards_compatible_with_v1() {
        // given
        let response = ResponseMessage::SealedHeaders(Err(
            ResponseMessageErrorCode::ProtocolV1EmptyResponse,
        ));
        let mut codec = PostcardCodec::new(1024);
        let mut buf = Vec::with_capacity(1024);
        
        // when
        codec
            .write_response(&PostcardProtocol::V1, &mut buf, response.clone())
            .await
            .expect("Valid Vec<SealedBlockHeader> is serialized using v1");
            
        let deserialized_as_legacy =
            // We cannot access the codec trait from an old node here, 
            // so we deserialize directly using the `LegacyResponseMessage` type.
            deserialize::<LegacyResponseMessage>(&buf).expect("Deserialization as LegacyResponseMessage should succeed");
            
        // then
        match deserialized_as_legacy {
            LegacyResponseMessage::SealedHeaders(None) => {}
            other => {
                panic!("Deserialized to {other:?}, expected {response:?}")
            }
        }
    }
```

MitchTurner (commented):
WIP. I'm gonna give this another pass soon.

crates/services/p2p/src/codecs/postcard.rs
```diff
@@ -160,30 +176,200 @@ impl GossipsubCodec for PostcardCodec {
     }
 }
 
+// TODO: Remove this NetworkCodec
```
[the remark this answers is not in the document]
acerone85: Yes, issue moved to the trait definition and referenced in b710d8d3199722f9eec428d17280c48648c4d07f

crates/services/p2p/src/codecs/postcard.rs
```diff
@@ -160,30 +176,200 @@ impl GossipsubCodec for PostcardCodec {
...
+
+        match deserialized {
+            ResponseMessage::SealedHeaders(Ok(sealed_block_headers_response)) => {
+                assert_eq!(sealed_block_headers, sealed_block_headers_response);
+            }
+            other => {
+                panic!("Deserialized to {other:?}, expected {sealed_block_headers:?}")
+            }
+        }
+    }
```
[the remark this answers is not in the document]
acerone85: Given/When/Then sections added in cec899c7a91fafb79eeea00e8ab2f5d303b3e05d

crates/services/p2p/src/codecs/postcard.rs
acerone85: nice :) I have refactored the tests to use matches! in 4280781c9c6fd39db7fde51b4fa643ca67ba1b94

crates/services/p2p/src/codecs/postcard.rs
acerone85: I have changed the names of the tests to follow your suggestions in 6276af58621d120f5b4e4526b5c0d2fa19f93f58

MitchTurner (approved):

AurelienFT (approved):

crates/services/p2p/src/codecs/postcard.rs
```diff
@@ -161,29 +177,214 @@ impl GossipsubCodec for PostcardCodec {
...
 }
 
-#[derive(Default, Debug, Clone)]
-pub struct MessageExchangePostcardProtocol;
+#[derive(Debug, Default, Clone, EnumIter)]
+pub enum PostcardProtocol {
+    #[default]
+    V1,
+    V2,
+}
```
xgreenx, line 195: Why `V1` is default if we prefer `V2`?

crates/services/p2p/src/p2p_service.rs
```diff
@@ -686,8 +688,8 @@ impl FuelP2PService {
                             }
                         },
                         ResponseSender::Transactions(c) => match response {
-                            ResponseMessage::Transactions(v) => {
-                                c.send((peer, Ok(v))).is_ok()
+                            V2ResponseMessage::Transactions(v) => {
+                                c.send((peer, Ok(v.ok()))).is_ok()
```
xgreenx, line 692: It would be nice to have the same TODO here

crates/services/p2p/src/p2p_service.rs
```diff
@@ -710,8 +712,8 @@ impl FuelP2PService {
                             }
                         },
                         ResponseSender::TxPoolFullTransactions(c) => match response {
-                            ResponseMessage::TxPoolFullTransactions(v) => {
-                                c.send((peer, Ok(v))).is_ok()
+                            V2ResponseMessage::TxPoolFullTransactions(v) => {
+                                c.send((peer, Ok(v.ok()))).is_ok()
```
xgreenx, line 716: It would be nice to have the same TODO here

crates/services/p2p/src/codecs/postcard.rs
acerone85: Thanks, this is a leftover for when I was doing protocol selection manually. In fact, we don't even need a Default version. 
I have removed it in the follow-up MR: https://github.com/FuelLabs/fuel-core/pull/2377 (commit ea4842c3c6)

Issue #2347, opened 2024-10-14
Create a stop in the case that DA gas price should go up but there are no txs to pay for it.


MitchTurner:
The way the gas price algorithm is designed, we increase the DA gas price when the cost of posting  L2 blocks to DA exceeds the reward received from charging L2 users the DA gas price.

In conjunction with this, as the DA gas price increases, there is a negative incentive for new txs to be submitted on L2, and this will lower the reward received, thus increasing the DA gas price further. This forms a negative feedback loop. If no txs are submitted, then the DA gas price could increase indefinitely, essentially bricking the network.

We need to introduce some kind of stop into the algorithm to prevent this negative feedback loop. Perhaps taking into account somehow the fullness of each L2 block before increasing the DA gas price.

This alone will only solve the symptom of the problem: DA gas price is high; and neglect the cause of the problem: a low profit (reward - cost). We need to also accommodate gradually working paying off the cost if there isn't an appetite from the market to pay the extra fee.

MitchTurner:
**One possible solution:**
We already have the `max_da_gas_price_change_percent` as part of the updater. What if we also include an `activity` parameter and a `activity_fullness_threshold_percent` parameter. Let's say the `activity` parameter is a value between `1` and `100` as is the `activity_fullness_threshold_percent`. Each time an L2 block fullness exceeds the `activity_fullness_threshold_percent`, the updater increases the `activity` by `1`. 

The `activity` parameter is then used in conjunction with the `max_da_gas_price_change_percent` when _increasing_ the gas price (not when decreasing). It would be an additional fraction of the `max_da_gas_price_change_percent`, so if `activity` is `0`, the `max_da_gas_price_change_percent` would be multiplied by `0` and the DA gas price wouldn't change.

This would have the effect of, as the network gets less and less used, the price will get increased less and less, even if the profit is negative and decreasing. _hopefully_ this will prevent the network from ever getting into a state where the prices prevent activity from lowering the prices.

PR #2364, opened 2024-10-16
Deal with negative feed back loop in DA gas price

MitchTurner:
## Linked Issues/PRs
#2347 

## Description
```rs
/// Because the DA gas price can increase even when no-one is using the network, there is a potential
/// for a negative feedback loop to occur where the gas price increases, further decreasing activity
/// and increasing the gas price. The `L2ActivityTracker` is used to moderate changes to the DA
/// gas price based on the activity of the L2 chain.
///
/// For each L2 block, the activity is calculated as a percentage of the block capacity used. If the
/// activity is below a certain threshold, the activity is decreased. The activity exists on a scale
/// between 0 and the sum of the increase, hold, and decrease buffers.
///
/// e.g. if the decrease range is 50, the hold range is 50, and the increase range is 50:
///
/// 0<-- decrease range -->50<-- hold range -->100<-- increase range -->150
///
/// The current activity determines the behavior of the DA gas price.
///
/// For healthy behavior, the activity should be in the `increase` range.
```

## Checklist

### Before requesting review


crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -151,13 +146,128 @@ pub struct AlgorithmUpdaterV1 {
...
+                .decrease_range_size
+                .saturating_add(self.hold_range_size)
+        {
+            DAGasPriceSafetyMode::Normal
+        } else if self.activity > self.decrease_range_size {
+            DAGasPriceSafetyMode::Hold
+        } else {
+            DAGasPriceSafetyMode::Decrease
+        }
+    }
```
MitchTurner, line 250: I think this should be tested a little better to make sure it chooses the expected "mode" based on given parameters.

netrome (approved):
Nice stuff! A few suggestions from my end.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -151,13 +146,128 @@ pub struct AlgorithmUpdaterV1 {
...
+/// and increasing the gas price. The `L2ActivityTracker` is used to moderate changes to the DA
+/// gas price based on the activity of the L2 chain.
+///
+/// For each L2 block, the activity is calculated as a percentage of the block capacity used. If the
+/// activity is below a certain threshold, the activity is decreased. The activity exists on a scale
+/// between 0 and the sum of the increase, hold, and decrease buffers.
+///
+/// e.g. if the decrease range is 50, the hold range is 50, and the increase range is 50:
+///
+/// 0<-- decrease range -->50<-- hold range -->100<-- increase range -->150
```
rafal-ch, line 166: Nit, for consistency.

```suggestion
/// 0<-- decrease range -->50<-- hold range -->100<-- normal range -->150
```

rafal-ch, line 166: Maybe a little variancy in sizes could make the example more clear.

For example, something like below?

```suggestion
/// e.g. if the decrease range is 20, the hold range is 40, and the increase range is 60:
///
/// 0<-- decrease range -->20<-- hold range -->60<-- increase range -->120
```

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -151,13 +146,128 @@ pub struct AlgorithmUpdaterV1 {
...
+///
+/// The current activity determines the behavior of the DA gas price.
+///
+/// For healthy behavior, the activity should be in the `increase` range.
 #[derive(serde::Serialize, serde::Deserialize, Debug, Clone, PartialEq)]
+pub struct L2ActivityTracker {
+    /// The _size_ of the range of the activity buffer that will increase the gas price
```
rafal-ch, line 173: Hmm, does it really mean that it will increase? I was interpreting Normal ([here](https://github.com/FuelLabs/fuel-core/pull/2364/files#diff-552dc8c5652389164f12c99ac314ae961e6e24684ad154f31ff4614eeb023991R188)) as "unrestricted changes to DA gas price", meaning it could increase, decrease or hold, but I may be wrong.

rafal-ch, line 173: I would also suggest consistent usage of "DA gas price" and "L2 gas price". I was even thinking at some point about newtypes to avoid mistakes - both in implementation and in attempts to understand the algo :-)

crates/fuel-gas-price-algorithm/src/v1.rs
rafal-ch: Yes, seems to be well covered. However, maybe this will version is more clear?

```
        let increase_threshold = self
            .decrease_range_size
            .saturating_add(self.hold_range_size);
        let hold_threshold = self.decrease_range_size;

        if self.activity > increase_threshold {
            return DAGasPriceSafetyMode::Normal;
        }
        if self.activity > hold_threshold {
            return DAGasPriceSafetyMode::Hold;
        }
        DAGasPriceSafetyMode::Decrease
```

We could even precalculate the thresholds in the constructor if we don't plan to change them in runtime.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -326,6 +449,18 @@ impl AlgorithmUpdaterV1 {
         );
     }
 
+    fn da_change_accounting_for_activity(&self, maybe_da_change: i128) -> i128 {
+        if maybe_da_change > 0 {
+            match self.l2_activity.safety_mode() {
+                DAGasPriceSafetyMode::Normal => maybe_da_change,
+                DAGasPriceSafetyMode::Hold => 0,
+                DAGasPriceSafetyMode::Decrease => self.max_change().saturating_mul(-1),
```
rafal-ch, line 457: Is it possible that `self.max_change() -> i128` gives a negative value and in such case we'll actually increase the price?

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -151,13 +146,128 @@ pub struct AlgorithmUpdaterV1 {
...
+    normal_range_size: u16,
+    /// The _size_ of the range of the activity buffer that will hold the gas price
+    hold_range_size: u16,
+    /// The _size_ of the range of the activity buffer that will decrease the gas price
+    decrease_range_size: u16,
+    /// The current activity of the L2 chain
+    activity: u16,
+    /// The threshold of activity below which the activity will be decreased
+    block_activity_threshold: ClampedPercentage,
+}
```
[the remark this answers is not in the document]
MitchTurner: I took your suggestions and made a few additional changes.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -151,13 +146,128 @@ pub struct AlgorithmUpdaterV1 {
...
+        } else {
+            DAGasPriceSafetyMode::Decrease
+        }
+    }
+
+    pub fn update(&mut self, block_usage: ClampedPercentage) {
+        let cap = self
+            .decrease_range_size
+            .saturating_add(self.hold_range_size)
+            .saturating_add(self.normal_range_size);
```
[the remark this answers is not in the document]
MitchTurner: We are pre-computing a lot of the values now.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -151,13 +146,128 @@ pub struct AlgorithmUpdaterV1 {
...
+            normal_range_size,
+            hold_range_size,
+            decrease_range_size,
+            activity,
+            block_activity_threshold,
+        }
+    }
+
+    pub fn new_always_normal() -> Self {
+        Self::new_full(1, 0, 0, ClampedPercentage::new(0))
```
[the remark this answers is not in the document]
MitchTurner: I named the inputs so that it's clearer, but I like using the constructor when possible.

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: Improved the wording here with the changes.

MitchTurner: Not a bad idea with the types.

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: I'm pre-computing them now.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -151,13 +146,133 @@ pub struct AlgorithmUpdaterV1 {
...
+    /// The maximum value the chain activity can hit
+    max_activity: u16,
+    /// The threshold if the block activity is below, the DA gas price will be held when it would otherwise be increased
+    hold_activity_threshold: u16,
+    /// If the chain activity falls below this value, the DA gas price will be decreased when it would otherwise be increased
+    decrease_activity_threshold: u16,
+    /// The current activity of the L2 chain
+    chain_activity: u16,
+    /// The threshold of block activity below which the chain activity will be decreased
+    block_activity_threshold: ClampedPercentage,
```
AurelienFT, line 186: Nit: maybe could be nice to specify that otherwise the chain activity is increased.

AurelienFT (commented):
Do you have an idea of the value you want to put in the threshold. You are saying that we should aim to be in `normal` range but right now on Ignition there is only 0 or 1 tx per block (excluding mint tx) and so the threshold seems hard to find given this granularity (0 or 1) and I'm having hard times to understand how we can maintain `normal` range in these conditions

AurelienFT (dismissed):

crates/fuel-gas-price-algorithm/src/v1.rs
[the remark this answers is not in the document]
netrome: Nice!

crates/fuel-gas-price-algorithm/src/v1.rs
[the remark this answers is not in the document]
netrome: Fair enough!

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -151,13 +146,134 @@ pub struct AlgorithmUpdaterV1 {
...
+///
+/// The chain activity is a cumulative measure, updated whenever a new block is processed.
+/// For each L2 block, the block usage is a percentage of the block capacity used. If the
+/// block usage is below a certain threshold, the chain activity is decreased, if above the threshold,
+/// the activity is increased The chain activity exists on a scale
+/// between 0 and the sum of the normal, hold, and decrease buffers.
+///
+/// e.g. if the decrease buffer size is 20, the hold buffer size is 60, and the increase buffer size is 40:
+///
+/// 0<-- decrease buffer -->20<-- hold buffer -->80<-- normal buffer -->120
```
netrome, line 170:
```suggestion
/// e.g. if the decrease activity threshold is 20, the hold activity threshold is 80, and the max activity is 120,
/// we'd have the following ranges:
///
/// 0 <-- decrease range -->20<-- hold range -->80<-- normal range -->120
```

netrome (changes requested):
The `L2ActivityTracker` needs a small update to reflect the new parameters. Otherwise this looks good to me.

AurelienFT (approved):

netrome (dismissed):

crates/fuel-gas-price-algorithm/src/v1.rs
rafal-ch: Yes, definitely not in this PR, if at all.

rafal-ch (dismissed):
LGTM :+1:

crates/fuel-gas-price-algorithm/src/v1.rs
rafal-ch: btw: would you mind leaving a `TODO` or creating a dedicated (good first) issue?

netrome: Just don't leave any `TODOs` without referencing a ticket. I'm okay with a ticket without a TODO though.

acerone85: > Just don't leave any TODOs without referencing a ticket. I'm okay with a ticket without a TODO though. 
We should do this consistently :) 
It looks like there is a rustfmt setting that would be useful to enforce this in the CI: https://github.com/rust-lang/rustfmt/blob/23f01eddacccf4c9b871b0fd84d4344fd67bf668/src/config.rs#L386-L389. But I suspect it would be a massive job to go through the whole code-base and add issues for any todo.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -151,13 +146,135 @@ pub struct AlgorithmUpdaterV1 {
...
+    chain_activity: u16,
+    /// The threshold of block activity below which the chain activity will be decreased,
+    /// above or equal it will always increase
+    block_activity_threshold: ClampedPercentage,
+}
+
+/// Designates the intended behavior of the DA gas price based on the activity of the L2 chain
+pub enum DAGasPriceSafetyMode {
+    /// Should increase DA gas price freely
+    Normal,
```
acerone85, line 194: I find `Normal` confusing, should this not be `Increase` instead?

MitchTurner: The idea is that it should increase and decrease at the expected times. "Increase" might accidentally imply that it should always increase. That was the thought around choosing "Normal" at least.

netrome: Perhaps the source of the confusion is the `Decrease` name. Maybe it would be more clear to rename it to `AlwaysDecrease`.

MitchTurner: In that case, maybe `Hold` isn't completely accurate either. Maybe it should be `Cap` or something like that, since it _can_ still decrease.

netrome: Oh right good point!

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: https://github.com/FuelLabs/fuel-core/issues/2382

Added the TODO with the issue attached.

AurelienFT (approved):
Even more test youhouuu

netrome (approved):

Dentosal (approved):

PR #2365, opened 2024-10-16
Fixed the error during dry run

xgreenx:
Because of race condition it was possible that dry run will fail. I moved the check to production, where we really need to require this behavior. But in the case of the dry run we only care about previous block existence.

## Checklist

### Before requesting review


crates/services/consensus_module/poa/src/service.rs
```diff
@@ -359,8 +359,6 @@ where
             consensus: seal,
         };
 
-        block.entity.header().time();
-
```
rymnc, line 363: nice catch

AurelienFT (approved):

acerone85 (approved):
Did not go in depth with the review, but I can follow the main idea behind the fix.
LGTM

AurelienFT (dismissed):

rafal-ch (approved):
LGTM :+1:

AurelienFT (approved):

crates/services/producer/src/block_producer/tests.rs
```diff
@@ -649,27 +649,33 @@ mod dry_run {
     }
 
     #[tokio::test]
-    async fn dry_run__errors_early_if_height_is_lower_than_chain_tip() {
```
MitchTurner, line 652: I did a quick scan but couldn't see a test for this on the produce or produce_predefined. Maybe I missed it. Do you think we need that coverage?

xgreenx: We have test case for `BlockHeightShouldBeHigherThanPrevious` error. 

image

But after introduction of the state rewind feature, I don't think that `BlockHeightShouldBeHigherThanPrevious` error makes a lot of sense. Basically importer already is responsible and upgradable executor do this kind of check inside, some maybe we can just remove it from the block producer

Issue #1834, opened 2024-04-17
Check Nonce of messages received by relayer


MitchTurner:
Our code assumes that each message received by the relayer has a unique nonce, but doesn't check that this is true.

A collision could be very bad because it would override the previous message. 

For now, maybe we halt the relayer if this occurs? We could also handle specific cases, like if the old message has already been spent?

Voxelot:
we should probably just skip nonces that have already been seen before

MitchTurner:
Had a PR that tried to solve the problem, but closed because the approach wasn't right:
https://github.com/FuelLabs/fuel-core/pull/2375

PR #2375, opened 2024-10-21
Ignore messages with duplicated nonces

MitchTurner:
## Linked Issues/PRs
Closes https://github.com/FuelLabs/fuel-core/issues/1834

## Description
Ignore new messages that share a nonce with an old message.

This isn't the only way to solve the issue but it was the [solution proposed](https://github.com/FuelLabs/fuel-core/issues/1834#issuecomment-2236877197)
> we should probably just skip nonces that have already been seen before

## Checklist

### Before requesting review


MitchTurner:
Is this actually a breaking change? The contract is kinda changing, but only on the code side. It was implicitly the same before.

crates/fuel-core/src/executor.rs
```diff
@@ -3285,6 +3285,70 @@ mod tests {
...
+            let changes = producer
+                .produce_without_commit(block.into())
+                .unwrap()
+                .into_changes();
+
+            // then
+            let changes_contains_message = ChangesIterator::<Column>::new(&changes)
+                .iter_all::<Messages>(None)
+                .count()
+                > 0;
```
AurelienFT, line 3348: Maybe testing directly that it's 0 could be more readable (and imaginary negative numbers in `count` return (if prototype changes (never happens but who knows)) will not pass)

AurelienFT (approved):
LGTM, little proposition in test.

crates/fuel-core/src/executor.rs
MitchTurner: I agree. That's easier to read. I'm not a big fan of the `!` operator anyway.

crates/services/executor/src/executor.rs
```diff
@@ -916,9 +916,15 @@ where
...
+                            .storage_as_ref::<Messages>()
+                            .contains_key(message_nonce)?
+                        {
+                            block_storage_tx
+                                .storage_as_mut::<Messages>()
+                                .insert(message_nonce, &message)?;
+                        }
                         execution_data
                             .events
                             .push(ExecutorEvent::MessageImported(message));
```
acerone85, line 930: should this also be conditional to the message_nonce not being already present in the block storage?

MitchTurner: Good catch. I've moved it.

This makes me wonder if we should be including an event in the case that a message is skipped. Or add it to a list of skipped messages, similar to the skipped tx.

MitchTurner: Added.

crates/types/src/services/executor.rs
```diff
@@ -131,6 +131,13 @@ impl<DatabaseTransaction> UncommittedResult<DatabaseTransaction> {
 pub enum Event {
     /// Imported a new spendable message from the relayer.
     MessageImported(Message),
+    /// Ignored a message from the relayer.
+    MessageIgnored {
+        /// Ignored message
+        message: Message,
+        /// The reason for ignoring the message
+        reason: String,
+    },
```
xgreenx, line 140: We shouldn't add a new event at the middle since it is a breaking change for the WASM interface, basically you break all nodes in the network with this update and forces them to use a new `fuel-core` node.

Plus, I don't think that we even need `MessageIgnored` event since it doesn't change the state of the chain.

crates/services/executor/src/executor.rs
```diff
@@ -916,12 +916,21 @@ where
...
+                        {
+                            block_storage_tx
+                                .storage_as_mut::<Messages>()
+                                .insert(message_nonce, &message)?;
+                            ExecutorEvent::MessageImported(message)
+                        } else {
+                            let reason =
+                                "Message with the same nonce already exists".to_string();
+                            ExecutorEvent::MessageIgnored { message, reason }
+                        };
```
xgreenx, line 932: I'm not sure even how do we want to handle this case.

Right now you don't handle the situation mentioned by @Voxelot .

> we should probably just skip nonces that have already been seen before

The `Messages` table contains only the list of unspent messages. If a message was seen before but spent after, it can still be inserted into the state in this solution.

xgreenx (commented):
I think it is a bad idea to just ignore the error. If the message is emitted by the contract, it means that someone deposited some funds. Ignoring the messages means that funds are lost forever.

I think we need to return an error if the message already exists in the table and insert it otherwise. It is not our responsibility to track the validity of how messages are emitted on the L1. We just need to be sure that it will not override the state.

crates/fuel-core/src/executor.rs
```diff
@@ -3285,6 +3285,78 @@ mod tests {
...
+        fn on_chain_db_containing_message(
+            message: &Message,
+            da_block_height: u64,
+        ) -> Database<OnChain> {
+            let mut db = add_consensus_parameters(
+                Database::default(),
+                &ConsensusParameters::default(),
+            );
+            db.storage_as_mut::<Messages>()
+                .insert(message.id(), message)
```
xgreenx, line 3297:
```suggestion
                .insert(message.nonce(), message)
```

May I ask you to rename this method please?=D `message.id()` for me means `MessageId` return type, while it is not true.

xgreenx (changes requested):
Also will block this PR until we agreed on what is expected behavior

crates/fuel-core/src/executor.rs
MitchTurner: There is already a `nonce` method that returns the same thing as `id`. Very silly.

crates/services/executor/src/executor.rs
MitchTurner: >  If a message was seen before but spent after, it can still be inserted into the state in this solution.

You mean if a message with the same nonce gets spent before we check for nonce collisions, it will get added still. Right?

crates/types/src/services/executor.rs
MitchTurner: That's fair. It was an afterthought and added a breaking section the the CHANGELOG to flag for this reason.

crates/services/executor/src/executor.rs
xgreenx: Yes. And I think it is okay behavior if L1 contracts emit the message with the same nonce. It is not our business to decide is it bad or not in this situation=D

MitchTurner: Then maybe we want a data structure that allows us to store multiple messages with colliding nonces?

xgreenx: No, only one message should be active at the time.

MitchTurner:
Gonna close this since we need to re-evaluate the solution. We can keep the closed PR as a reference in the issue.

PR #2376, opened 2024-10-21
Add a way to fetch transactions without specifying a peer in P2P

AurelienFT:
## Linked Issues/PRs
This is a requirement for https://github.com/FuelLabs/fuel-core/pull/2361

## Description

This PR adds a way to fetch transactions with p2p but without giving a specific peer and let p2p choose the one they prefer.
This will be used in #2361

## Checklist

### Before requesting review

### DIFF
--- a/crates/services/p2p/src/service.rs
+++ b/crates/services/p2p/src/service.rs
@@ -116,6 +116,10 @@ pub enum TaskRequest {
     },
     GetTransactions {
         block_height_range: Range<u32>,
+        channel: OnResponse<Option<Vec<Transactions>>>,
+    },
+    GetTransactionsFromPeer {
+        block_height_range: Range<u32>,
         from_peer: PeerId,
         channel: OnResponse<Option<Vec<Transactions>>>,
     },
@@ -164,6 +168,9 @@ impl Debug for TaskRequest {
             }
             TaskRequest::GetTransactions { .. } => {
                 write!(f, "TaskRequest::GetTransactions")
+            }
+            TaskRequest::GetTransactionsFromPeer { .. } => {
+                write!(f, "TaskRequest::GetTransactionsFromPeer")
             }
             TaskRequest::TxPoolGetAllTxIds { .. } => {
                 write!(f, "TaskRequest::TxPoolGetAllTxIds")
@@ -856,7 +863,16 @@ where
                             tracing::warn!("No peers found for block at height {:?}", height);
                         }
                     }
-                    Some(TaskRequest::GetTransactions { block_height_range, from_peer, channel }) => {
+                    Some(TaskRequest::GetTransactions {block_height_range, channel }) => {
+                        let channel = ResponseSender::Transactions(channel);
+                        let request_msg = RequestMessage::Transactions(block_height_range.clone());
+                        let height = BlockHeight::from(block_height_range.end.saturating_sub(1));
+                        let peer = self.p2p_service.get_peer_id_with_height(&height);
+                        if self.p2p_service.send_request_msg(peer, request_msg, channel).is_err() {
+                            tracing::warn!("No peers found for block at height {:?}", block_height_range.end);
+                        }
+                    }
+                    Some(TaskRequest::GetTransactionsFromPeer { block_height_range, from_peer, channel }) => {
                         let channel = ResponseSender::Transactions(channel);
                         let request_msg = RequestMessage::Transactions(block_height_range);
                         self.p2p_service.send_request_msg(Some(from_peer), request_msg, channel).expect("We always a peer here, so send has a target");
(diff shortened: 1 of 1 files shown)


crates/services/p2p/src/service.rs
```diff
@@ -115,6 +115,10 @@ pub enum TaskRequest {
         channel: OnResponse<Option<Vec<SealedBlockHeader>>>,
     },
     GetTransactions {
+        block_height_range: Range<u32>,
+        channel: OnResponse<Option<Vec<Transactions>>>,
+    },
+    GetTransactionsFromPeer {
```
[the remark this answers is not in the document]
AurelienFT: I think it's useful in case of syncing when you already received the header from someone in the current syncing process and so you can directly ask him without having to redo peer selection. Maybe there is also othhers use cases but this is the only one I see.

crates/services/p2p/src/service.rs
```diff
@@ -115,6 +115,10 @@ pub enum TaskRequest {
         channel: OnResponse<Option<Vec<SealedBlockHeader>>>,
     },
     GetTransactions {
+        block_height_range: Range<u32>,
+        channel: OnResponse<Option<Vec<Transactions>>>,
```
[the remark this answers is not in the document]
AurelienFT: The problem that I have with this solution is that in case of the p2p that select the peer I want p2p to return the peer so that I can report things about his results afterwards. Something that I don't need when I give the peer because I already know it.

crates/services/p2p/src/service.rs
```diff
@@ -856,7 +863,16 @@ where
...
                     }
-                    Some(TaskRequest::GetTransactions { block_height_range, from_peer, channel }) => {
+                    Some(TaskRequest::GetTransactions {block_height_range, channel }) => {
+                        let channel = ResponseSender::Transactions(channel);
+                        let request_msg = RequestMessage::Transactions(block_height_range.clone());
+                        let height = BlockHeight::from(block_height_range.end.saturating_sub(1));
+                        let peer = self.p2p_service.get_peer_id_with_height(&height);
+                        if self.p2p_service.send_request_msg(peer, request_msg, channel).is_err() {
+                            tracing::warn!("No peers found for block at height {:?}", block_height_range.end);
+                        }
```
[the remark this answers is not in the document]
AurelienFT: Done. I followed the pattern from the headers should I change the headers one ?

acerone85 (dismissed):
LGTM

Dentosal (dismissed):

crates/services/p2p/src/service.rs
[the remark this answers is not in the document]
xgreenx: It seems that after your PR we don't need this variant anymore. But we can address it in your PR

crates/services/p2p/src/service.rs
```diff
@@ -880,7 +887,16 @@ where
                             tracing::warn!("No peers found for block at height {:?}", height);
                         }
                     }
-                    Some(TaskRequest::GetTransactions { block_height_range, from_peer, channel }) => {
+                    Some(TaskRequest::GetTransactions {block_height_range, channel }) => {
+                        let channel = ResponseSender::Transactions(channel);
+                        let request_msg = RequestMessage::Transactions(block_height_range.clone());
+                        let height = BlockHeight::from(block_height_range.end.saturating_sub(1));
+                        let peer = self.p2p_service.get_peer_id_with_height(&height);
+                        if self.p2p_service.send_request_msg(peer, request_msg, channel).is_err() {
```
xgreenx, line 895: We should return error to the `channel` if `peer` is `None`, to allow proper handling of it by the caller.

AurelienFT: Before doing it I want to ask if we want the same in `GetSealedHeaders` (just above in the code) ? It has the same behavior and doesn't return the error for now.

crates/services/p2p/src/service.rs
```diff
@@ -1044,6 +1060,31 @@ impl SharedState {
...
+                "Cannot retrieve transactions for an empty range of block heights"
+            ));
+        }
+
+        self.request_sender
+            .send(TaskRequest::GetTransactions {
+                block_height_range: range,
+                channel: sender,
+            })
+            .await?;
```
rymnc, line 1080: perhaps we should continue to use the `TaskRequest::GetTransactionsFromPeer` and perform peer selection for a given block height instead - https://github.com/FuelLabs/fuel-core/blob/f597ac9db0602caed7e20eaeafed0a1bd2e96ca7/crates/services/p2p/src/service.rs#L264-L266

AurelienFT: I personally think that there is still use case where it's interesting to already give a peer that we know have the information, we use this un sync service.

rymnc: right, i agree ~ but this information is also valid for a peer returned by `get_peer_id_with_height`. perhaps we should not duplicate valid peers for a given height then.

AurelienFT: I don't get your comment. My flow is that I ask for an header and so the peer is returned along with the header then I can directly ask to this peer about the transactions because in our network if you have the header we assume you have the transactions. It avoid running this `get_peer_id_with_height` function that can be costly.
Maybe I miss your point and in this case we can go over more voice discussion :)

xgreenx: > perhaps we should continue to use the TaskRequest::GetTransactionsFromPeer and perform peer selection for a given block height instead

If we do peer selection outside of the `run` loop, we can face race condition were peer is disconnected already

crates/services/p2p/src/service.rs
AurelienFT: Done. It complexify a bit the type in the channel because we have a two-stage error type but with unification of return type etc that we want to do in p2p maybe we could improve this in the future.

xgreenx (approved):

rymnc (approved):

PR #2377, opened 2024-10-21
Fuel/Request_Response v0.0.2: More meaningful error messages

### LINKED ISSUE #1311: Return better error messages from P2P requests

acerone85:
## Linked Issues/PRs

Closes #1311

## Description

This PR follows #2362, which introduces a new version of the protocol `/fuel/req_response/0.0.2` which allows to return Errors in the response. This PR makes the following changes: 
- [X] the different errors that can be returned in a response are implemented.  
- [X] The type of `ResponseSender` used by a peer to propagate received responses upstream is changed to allow the rrors to be propagated 
- [X] Error Responses received by a peer following a request are displayed with as log messages with `warning` level.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/p2p/src/p2p_service.rs
```diff
@@ -677,7 +677,7 @@ impl FuelP2PService {
                             V2ResponseMessage::SealedHeaders(v) => {
                                 // TODO: https://github.com/FuelLabs/fuel-core/issues/1311
```
AurelienFT, line 678: Outdated todo :)

crates/services/p2p/src/service.rs
```diff
@@ -585,7 +583,7 @@ where
         // TODO: https://github.com/FuelLabs/fuel-core/issues/1311
```
AurelienFT, line 583: Outdated todo

AurelienFT (commented):
LGTM. I think you can run a search over : "https://github.com/FuelLabs/fuel-core/issues/1311" in codebase and remvoe all todo related and make sure they are all addressed

crates/services/p2p/src/request_response/messages.rs
```diff
@@ -38,6 +38,12 @@ pub enum ResponseMessageErrorCode {
     /// The peer sent an empty response using protocol `/fuel/req_res/0.0.1`
     #[error("Empty response sent by peer using legacy protocol /fuel/req_res/0.0.1")]
     ProtocolV1EmptyResponse = 0,
+    #[error("The requested range is too large")]
+    RequestedRangeTooLarge = 1,
+    #[error("Timeout while processing request")]
+    Timeout = 2,
+    #[error("Sync processor is out of capacity")]
+    SyncProcessorOutOfCapacity = 3,
```
xgreenx, line 46: Also, maybe we need to add `Unknown(u32)` for backward compatibility in the case if we can't deserialize the error.

crates/services/p2p/src/service.rs
acerone85: done: fd2ba043a2f389cc206a906d9fa66c5029004abb

crates/services/p2p/src/request_response/messages.rs
acerone85: Tried something in 5ea9b30e4d409b85f76aa40cc60c52d409b11ec9. There is an unknown variant, but the error code is erased away. The reason for this is that deserializing unknown values in serde has been recently implemented via the attribute `serde(untagged)`, but this is not supported by postcard (fails with error `WontImplement`).  See 5ea9b30e4d409b85f76aa40cc60c52d409b11ec9 for the discussion on `serde(untagged)`. There are other alternatives proposed, such as nested enums or custom deserialization functions for a variant, but they won't either work with postcard or are too complicated imho. 

Adding an `Unknown` variant which abstracts away the error code does at least ensure backwards compatibility.

xgreenx (dismissed):

AurelienFT (approved):

AurelienFT (approved):

xgreenx (approved):

PR #2321, opened 2024-10-08
Add metrics to TxPool

acerone85:
## Linked Issues/PRs

Closes #1937

## Description

 - [x] Number of all transactions in pool
 - [x] Number of immediately executable transactions in pool
 - [x] Number of transactions pending verification
 - [x] (Txs pending verification) how many tasks are created in sync processor
 - [x] (Txs pending verification) and for how long they were active in the sync processor)
 - [x] Size of transactions (when transaction is inserted (meaning: it was accepted for inclusion) - histogram)
 - [x] Time to insert transaction (from when received until it is included in a block)
 - [x] Select transaction for block (how long it took for the selection algorithm to select transactions)

TODO: 
- [x] Enable metrics only if `config.metrics = true`

## Checklist
- [ ] Breaking changes are clearly marked as such in the PR description and changelog
- [ ] New behavior is reflected in tests
- [ ] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [ ] I have reviewed the code myself
- [ ] I have created follow-up issues caused by this PR and linked them here

### After merging, notify other teams

[Add or remove entries as needed]

- [ ] [Rust SDK](https://github.com/FuelLabs/fuels-rs/)
- [ ] [Sway compiler](https://github.com/FuelLabs/sway/)
- [ ] [Platform documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+) (for out-of-organization contributors, the person merging the PR will do this)
- [ ] Someone else?

Issue #1937, opened 2024-06-05
Add metrics for TxPool


xgreenx:
Metrics for:
- [x] Number of all transactions in pool
- [x] Number of immediately executable transactions in pool
- [x] Number of transactions pending verification 
- [x] (Txs pending verification) how many tasks are created in sync processor 
- [x] (Txs pending verification) and for how long they were active in the sync processor)
- [x] Size of transactions (when transaction is inserted (meaning: it was accepted for inclusion) - histogram)
- [x] Time to insert transaction (from when received until it is included in a block)
- [x] Select transaction for block (how long it took for the selection algorithm to select transactions)

Possibly for follow-up, we first need to create an enum describing the lifetime of the transaction inside TxPool
- [ ] Removed transactions (during which stage)
- [ ] Number of rejected transactions (during verification)
- [ ] Number of squeezed out transactions

MitchTurner:
Completed for now. Might reopen in the future :)

PR #2385, opened 2024-10-23
Txpool metrics update

rafal-ch:
## Linked Issues/PRs
This PR addresses the outstanding comments to the TxPool metrics PR: https://github.com/FuelLabs/fuel-core/pull/2321

## Description
There are a couple of changes to how and when metrics are registered and some histogram buckets are redefined.

Additionally there is also a small clean-up of the changelog file.

### Before requesting review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
-- [2321](https://github.com/FuelLabs/fuel-core/pull/2321): New metrics for the txpool:
+- [2321](https://github.com/FuelLabs/fuel-core/pull/2321): New metrics for the TxPool:
     - The size of transactions in the txpool (`txpool_tx_size`)
     - The time spent by a transaction in the txpool in seconds (`txpool_tx_time_in_txpool_seconds`)
     - The number of transactions in the txpool (`txpool_number_of_transactions`)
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -6,18 +6,24 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- [2321](https://github.com/FuelLabs/fuel-core/pull/2321): New metrics for the txpool:
```
acerone85, line 10: should this mention both #2321 and #2385?

acerone85 (approved):
LGTM

AurelienFT (approved):
LGTM

crates/services/txpool_v2/src/service.rs
```diff
@@ -432,7 +430,8 @@ where
             };
 
             if metrics {
-                record_tx_size(&checked_tx)
+                let size = checked_tx.metered_bytes_size();
+                txpool_metrics().tx_size.observe(size as f64);
             };
```
xgreenx, line 435: Do we want to record the size of inserted transaction or size of all transactions that entered the pool? If the first one, then maybe it makes sense to move this metric into `TxPool::insert` function. If the second, then we need to do it along with `.number_of_transactions_pending_verification.inc()`.

CHANGELOG.md
rafal-ch: I think it should. I wanted to keep the "one PR per line" rule, so this is how I have changed it:
https://github.com/FuelLabs/fuel-core/pull/2385/commits/dadaaa6214160b9a331c55ac8bb13bc366a44433

crates/services/txpool_v2/src/service.rs
rafal-ch: As discussed, we're interested in logging transactions that has been _inserted_ into the TxPool, meaning: every transaction that passed all validations and is actually being inserted.

Other cases (i.e. log rejected transactions) are still to be handled, as per the original issue: https://github.com/FuelLabs/fuel-core/issues/1937 - this will be delivered as a separate PR.

xgreenx (approved):

AurelienFT (approved):

CHANGELOG.md
```diff
@@ -6,22 +6,29 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
```
acerone85, line 9: Something wrong with the changelog?

acerone85: ah no, entries have been moved.

rafal-ch: Yes, it's good to have `Added` at the top.

crates/metrics/src/buckets.rs
```diff
@@ -76,6 +62,44 @@ fn initialize_buckets() -> HashMap<Buckets, Vec<f64>> {
                 256.0 * 1024.0,
             ]
         ),
+        (
+            Buckets::TransactionInsertionTimeInThreadPool,
```
acerone85, line 66: not for this PR, but I think having the name of the buckets implying their use is not optimal. That said, we discussed this already and IIRC we could not find a good naming convention :)

acerone85 (approved):

xgreenx (approved):

Issue #2093, opened 2024-08-16
RocksDB Opens too many descriptors by default


xgreenx:
The RocksDB opens a lot of files for its work. Currently, we open an unlimited number of files that could hit at some point 2M descriptors.

Not all machines have huge descriptor limits by default. To avoid errors on the user's side, we could set the limit for open files based on OS limit. 

Let's say the limit on the OS is `X`. Then, we could use `X/2` as a limit for the RocksDB.

Also, we need to update the Readme file to highlight that property and that low descriptors limit could decrease performance.

mvares:
@xgreenx, where can I the limit of a OS?

MitchTurner:
We could add a cli flag that would allow the user to control this behavior. i.e. if it's set it uses the specified value, if it's not set it uses system limit.

PR #2386, opened 2024-10-23
Add new flag for maximum file descriptors in rocksdb.

AurelienFT:
## Linked Issues/PRs
Resolves: https://github.com/FuelLabs/fuel-core/issues/2093

## Description

Add a flag to define the maximum number of file descriptors that RocksDB can use. By default it's half of the OS limit.

## Checklist

### Before requesting review


MitchTurner (commented):
Looking good. Left some thoughts.

bin/fuel-core/src/cli/run.rs
```diff
@@ -126,6 +130,19 @@ pub struct Command {
...
+    #[cfg(feature = "rocksdb")]
+
+    /// Defines a specific number of file descriptors that RocksDB can use.
+    ///
+    /// If defined as -1 no limit will be applied and will use the OS limits.
+    /// If not defined the system default divided by two is used.
+    #[clap(
+        long = "rocksdb-max-fds",
+        env,
+        default_value = getrlimit(Resource::NOFILE).map(|(_, hard)| i32::try_from(hard.saturating_div(2)).unwrap_or(i32::MAX)).unwrap().to_string()
```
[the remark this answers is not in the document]
AurelienFT: Done :)

crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -279,6 +288,7 @@ where
         }
         opts.set_max_background_jobs(6);
         opts.set_bytes_per_sync(1048576);
+        opts.set_max_open_files(max_fds);
 
         #[cfg(feature = "test-helpers")]
         opts.set_max_open_files(512);
```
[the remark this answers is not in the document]
AurelienFT: Yep maybe just to be sure I changed to 512.

bin/fuel-core/src/cli/rollback.rs
```diff
@@ -30,7 +30,7 @@ pub struct Command {
     #[clap(
         long = "rocksdb-max-fds",
         env,
-        default_value = getrlimit(Resource::NOFILE).map(|(_, hard)| i32::try_from(hard.saturating_div(2)).unwrap_or(i32::MAX)).unwrap().to_string()
+        default_value = getrlimit(Resource::NOFILE).map(|(_, hard)| i32::try_from(hard.saturating_div(2)).unwrap_or(i32::MAX)).expect("Failed to get OS max file descriptors.").to_string()
```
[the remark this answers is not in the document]
AurelienFT: Sorry I didn't correctly understood you first message and didn't know about this convention. Happy to learn :D Ty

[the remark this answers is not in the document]
netrome: Oh didn't know this. I've definitely done a bunch of `.expect("failed to blah blah")` 😅 TIL

bin/fuel-core/src/cli/rollback.rs
```diff
@@ -19,6 +23,17 @@ pub struct Command {
     )]
     pub database_path: PathBuf,
 
+    /// Defines a specific number of file descriptors that RocksDB can use.
+    ///
+    /// If defined as -1 no limit will be applied and will use the OS limits.
+    /// If not defined the system default divided by two is used.
```
netrome, line 29: What's the reason for using `-1` to mean no limit as opposed to take an `Option<u32>`?

bin/fuel-core/src/cli/rollback.rs
```diff
@@ -19,6 +23,17 @@ pub struct Command {
...
+    #[clap(
+        long = "rocksdb-max-fds",
+        env,
+        default_value = getrlimit(Resource::NOFILE).map(|(_, hard)| i32::try_from(hard.saturating_div(2)).unwrap_or(i32::MAX)).expect("Our supported platforms should return max FD.").to_string()
```
netrome, line 33: This is quite long to be in an attribute. How about breaking this out to a helper function?

bin/fuel-core/src/cli/run.rs
```diff
@@ -126,6 +130,19 @@ pub struct Command {
...
+    #[cfg(feature = "rocksdb")]
+
+    /// Defines a specific number of file descriptors that RocksDB can use.
+    ///
+    /// If defined as -1 no limit will be applied and will use the OS limits.
+    /// If not defined the system default divided by two is used.
+    #[clap(
+        long = "rocksdb-max-fds",
+        env,
+        default_value = getrlimit(Resource::NOFILE).map(|(_, hard)| i32::try_from(hard.saturating_div(2)).unwrap_or(i32::MAX)).expect("Our supported platforms should return max FD.").to_string()
```
netrome, line 142: Same here

bin/fuel-core/src/cli/snapshot.rs
```diff
@@ -29,6 +33,17 @@ pub struct Command {
...
     pub database_path: PathBuf,
 
+    /// Defines a specific number of file descriptors that RocksDB can use.
+    ///
+    /// If defined as -1 no limit will be applied and will use the OS limits.
+    /// If not defined the system default divided by two is used.
+    #[clap(
+        long = "rocksdb-max-fds",
+        env,
+        default_value = getrlimit(Resource::NOFILE).map(|(_, hard)| i32::try_from(hard.saturating_div(2)).unwrap_or(i32::MAX)).expect("Our supported platforms should return max FD.").to_string()
```
netrome, line 43: And here

netrome (dismissed):
Looks correct to me. I think this would be even better if we used an `Option<u32>` instead of an `i32` where `-1` has a special meaning.

bin/fuel-core/src/cli/rollback.rs
AurelienFT: It is the default value used by RocksDB for no limit that's why they are a i32.

bin/fuel-core/src/cli/rollback.rs
MitchTurner: Good idea.

MitchTurner:
Ready to approve after those last changes are made.

bin/fuel-core/src/cli/rollback.rs
AurelienFT: And we want to have different behaviour between : 
- No defined value which will be using OS divided by two
- and max value of the system which is -1

netrome: Alright, fair enough.

bin/fuel-core/src/cli/rollback.rs
AurelienFT: Done

bin/fuel-core/src/cli/run.rs
AurelienFT: Done

bin/fuel-core/src/cli/snapshot.rs
AurelienFT: Done

benches/benches/vm_set/blockchain.rs
```diff
@@ -75,6 +75,7 @@ impl BenchDb {
             tmp_dir.path(),
             None,
             Default::default(),
+            -1,
```
MitchTurner, line 78: There are a few places that we're still using `-1`. Is that okay?

bin/fuel-core/src/cli/rollback.rs
```diff
@@ -19,11 +23,28 @@ pub struct Command {
     )]
     pub database_path: PathBuf,
 
+    /// Defines a specific number of file descriptors that RocksDB can use.
+    ///
+    /// If defined as -1 no limit will be applied and will use the OS limits.
```
MitchTurner, line 28: What is the behavior if they use -2 or any other negative value?

bin/fuel-core/src/cli/rollback.rs
rafal-ch: I agree that this limit should generally be `Option<u32>`, but this can only represent two states, not the three we need. Let's keep it like that for the sake of simplicity. But we should maybe sanitize this value and not allow -2, -3, etc.?

benches/benches/vm_set/blockchain.rs
AurelienFT: I have let in benches because benches didn't used the test value before and I don't want them to be slowed down by a new limit (maybe it don't change anything but there is no risk for benches I think)

bin/fuel-core/src/cli/rollback.rs
AurelienFT: Didn't found anything about it in rocksdb source code in Rust or C, so I tried myself in a test and it seems to work without issues. Maybe this is the code : https://github.com/facebook/rocksdb/blob/8b38d4b4006ca9fd49432ccc16d9911919870dd2/db/db_impl/db_impl_open.cc#L57 I'm not sure that I fully understand this code but fore me if the value is -2 it will be clamped to 20.

bin/fuel-core/src/cli/rollback.rs
AurelienFT: RocksDB doesn't seems to consider these values as error : https://github.com/FuelLabs/fuel-core/pull/2386#discussion_r1814657998

But we can do. I don't think it's very useful because it doesn't creates any problem and the only reason to tryy a -2 is to tryy to make a problem.

MitchTurner (approved):

netrome (approved):
Nice!

PR #2388, opened 2024-10-23
Remove p2pservice networkcodec trait

### LINKED ISSUE #2368: Remove P2PService NetworkCodec trait
## Context

In the P2P service, we define a CodecTrait that can be defined for any struct implementing both GossipsubCodec and RequestResponse::Codec. See 
See https://github.com/FuelLabs/fuel-core/blob/f50a4d3874c6b1d464ad83c529956a6c7a473819/crates/services/p2p/src/codecs.rs#L33. 

However, this trait only defines functions that are relevant to the `RequestResponse` protocol, namely iterating over all protocols available. Furthermore, it constrains the types  of messages sent via `GossipSub` or `Request/Response` to be the same, which is unnecessary coupling between the two. 

## Acceptance criteria

The `NetworkCodec` trait is removed from the codebase, and the `FuelService` is refactored to not use this codec.

## Relevant files: 
- Definition of the `NetworkCodec` trait: https://github.com/FuelLabs/fuel-core/blob/release/v0.40.0/crates/services/p2p/src/codecs.rs
- Usage of the `NetworkCodec::get_req_res_protocol` function in the `FuelService`: https://github.com/FuelLabs/fuel-core/blob/eea8c605258d79d5243f3f822c0e1205b0802f3c/crates/services/p2p/src/behavior.rs#L116
- Implementation of the `NetworkCodec` trait for the `PostcardCodec`: https://github.com/FuelLabs/fuel-core/blob/eea8c605258d79d5243f3f822c0e1205b0802f3c/crates/services/p2p/src/codecs/postcard.rs#L163

acerone85:
## Linked Issues/PRs

Closes #2368 

## Description

In the P2P service, we couple together GossipSub and RequestResponse codecs in a `NetworkCodec` trait. In practice, this coupling is not necessary as the two codecs use different Messages and different logics for encoding/deconding messages. The only thing that they share in common is using Postcard as the data format to serialize and deserialize data, but this is not made clear in the network trait.  

This PR is a rework of Codecs in the libp2p service, which deletes the `NetworkCodec` trait and therefore the coupling between `GossipSubCodec` and `RequestResponseCodec`. In particular: 
- [ ] A new DataFormat trait is defined for serializing deserializing any SerDe datatype, and the helper functions for postcard serialization and deserialization` are moved to a struct `Postcard` that implements the `DataFormat` trait, 
- [ ] A new `BoundedCodec` struct that implements the `RequestResponse::Codec` trait for any `DataFormat` supersedes the old `PostcardCodec`. 
- [ ] A new `UnboundedCodec` struct that implements the `GossipsubCodec trait for any `DataFormat` supersedes the old `PostcardCodec`, 
- [ ] 
- [ ] Parts specific to the Request/Response protocol have been moved to a new `request_response::protocols` module`, 
- [ ] The FuelBehaviour and LibP2P service have been refactored to use different codecs for `Gossipsub` and `RequestResponse`. 

TODO: 
- [ ] Better Documentation
- [x] Better names for modules
- [ ] Maybe move DataFormats to a different part of the codebase?

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/p2p/src/codecs.rs
```diff
@@ -1,19 +1,24 @@
...
 use std::io;
 
+trait DataFormatCodec {
+    type Error;
+    fn deserialize<'a, R: Deserialize<'a>>(
+        encoded_data: &'a [u8],
+    ) -> Result<R, Self::Error>;
+
+    fn serialize<D: Serialize>(data: &D) -> Result<Vec<u8>, Self::Error>;
+}
```
netrome, line 20: Are we sure there are no data formats that take parameters? I'd err on the cautious side and accept a `self` parameter to these methods. This would for example allow a dynamically configurable formalt such as:

```
enum MyFormat {
    Postcard,
    Json
}

While I don't know by heart when or where we'd like to use this, it feels overly restrictive to explicitly rule it out. This also eliminates the need for `PhantomData` markers in our structs, which is a bit of code smell to me.
```

```suggestion
trait DataFormatCodec {
    type Error;
    fn deserialize<'a, R: Deserialize<'a>>(
        &self,
        encoded_data: &'a [u8],
    ) -> Result<R, Self::Error>;

    fn serialize<D: Serialize>(&self, data: &D) -> Result<Vec<u8>, Self::Error>;
}
```

crates/services/p2p/src/codecs/bounded.rs
```diff
@@ -0,0 +1,156 @@
...
+};
+
+use super::{
+    DataFormatCodec,
+    RequestResponseProtocols,
+};
+
+#[derive(Debug, Clone)]
+pub struct BoundedCodec<Format> {
+    _format: PhantomData<Format>,
```
netrome, line 31: Following my other suggestion I'd suggest skipping the `PhantomData` here. If the `Format` is an empty struct, there's no cost to holding it directly.

```suggestion
    _format: Format,
```

crates/services/p2p/src/codecs/bounded.rs
```diff
@@ -0,0 +1,156 @@
...
+
+/// Since Postcard does not support async reads or writes out of the box
+/// We prefix Request & Response Messages with the length of the data in bytes
+/// We expect the substream to be properly closed when response channel is dropped.
+/// Since the request protocol used here expects a response, the sender considers this
+/// early close as a protocol violation which results in the connection being closed.
+/// If the substream was not properly closed when dropped, the sender would instead
+/// run into a timeout waiting for the response.
+#[async_trait]
+impl<Format> request_response::Codec for BoundedCodec<Format>
```
netrome, line 60: The naming feels off here. A trait named `Codec` should have encode/decode or serialize/deserialize methods. This trait contains read/write methods and seem to be representing primarily I/O operations. I get that this comes from p2p, but perhaps we should find another name for `BoundedCodec`. Right now I'd expect some sort of similarity between the `BoundedCodec` and `UnboundedCodec` types, but they feel very different in what they do with the current implementations.

Perhaps I'm just not understanding this well enough right now. I'd love to hear your thoughts on this.

crates/services/p2p/src/p2p_service.rs
```diff
@@ -211,7 +213,8 @@ impl FuelP2PService {
     pub async fn new(
         reserved_peers_updates: broadcast::Sender<usize>,
         config: Config,
-        codec: PostcardCodec,
+        gossipsub_codec: UnboundedCodec<PostcardDataFormat>,
+        request_response_codec: BoundedCodec<PostcardDataFormat>,
```
netrome, line 217: Given that these are the only two instances of using these codecs, perhaps we could simplify the implementation with more specific traits. What if we just renamed `UnboundedCodec` -> `GossipsubCodec` (perhaps we only need one trait and not two traits here), and renamed `BoundedCodec` -> `RequestResponseReadWrite` (or something similar)?

crates/services/p2p/src/codecs.rs
```diff
@@ -1,19 +1,24 @@
...
-};
+use crate::gossipsub::messages::GossipTopicTag;
 use libp2p::request_response;
+use serde::{
+    Deserialize,
+    Serialize,
+};
 use std::io;
 
+trait DataFormatCodec {
```
netrome, line 13: How about just renaming this `DataFormat`?

crates/services/p2p/src/codecs.rs
```diff
@@ -1,19 +1,24 @@
...
+    type Error;
+    fn deserialize<'a, R: Deserialize<'a>>(
+        encoded_data: &'a [u8],
```
netrome, line 16: This could be generalized to a `impl std::io::Read`.

crates/services/p2p/src/codecs.rs
```diff
@@ -1,19 +1,24 @@
...
+    ) -> Result<R, Self::Error>;
+
+    fn serialize<D: Serialize>(data: &D) -> Result<Vec<u8>, Self::Error>;
```
netrome, line 19: This could also be generalized by taking a `impl std::io::Write` input which it writes to. Then a convenience function `serialize_to_vec` could be provided with this signature.

netrome (commented):
Generally I like the decoupling you're achieving here. I think we could do some simplifications by reducing the number of traits, and for example just having one `GossipSubCodec` trait and one`RequestResponseReadWrite` trait. Both of these could be generic over the data format if needed.

crates/services/p2p/src/codecs.rs
acerone85: Thanks, I like the suggestion. I have made the change here: c564941d769d1524d04dfec9675b6f7da7d5b5f6. Note that the `DataFormatCodec` trait has been removed in favour of two `Encode` and `Decode` traits, following what has been done in the `storage` crate.

crates/services/p2p/src/codecs/bounded.rs
acerone85: Agreed, I had a lot of troubles trying to find the right terminology (I also did not like `BoundedCodec` and `UnboundedCodec`. In the last version I have made the following changes: 
* `DataFormatCodec` has been replaced by `Encode` and `Decode` traits, 
* `BoundedCodec` is now a `RequestResponseMessageHandler`. I still don't like the Handler part in the name as it might make people think that it's something akin to a task, but I could not find anything better. The `data_format` field has been renamed to `codec`
* `UnboundedCodec` is now `GossipsubMessageHandler`,  and the `data_format` field has been renamed to `codec`, 
* The `Format` type parameter has been renamed to `codec`.
* I did not rename the `GossipsubCodec` trait, suggestions for better names are very welcome.

acerone85: Commits are ead761a08a1a103bdd103f8f7e66da077fb92186 and c5aa3622bb56ef69ce7aee44231ae20a9d05d2c3

crates/services/p2p/src/p2p_service.rs
acerone85: Followed the suggestion: c5aa3622bb56ef69ce7aee44231ae20a9d05d2c3 and ead761a08a1a103bdd103f8f7e66da077fb92186

crates/services/p2p/src/codecs.rs
acerone85: I'm not sure I follow completely here, do you mean changing the type of `econded_data`? If we do so, we need to allocate a second buffer where to read the value into. Or do you have something else in mind?

crates/services/p2p/src/codecs.rs
acerone85: This has been changed to reflect what's been done in the Storage crate. We now have an `Encoder` trait that is used as the type where the result is stored, and this is implemented for `Cow<'_, [u8]>`. I could in theory change this to have the input where the serialized data is written passed into the function, but this would create me trouble when I will try to merge the traits in the storage trait and this PR together (see new issue https://github.com/FuelLabs/fuel-core/issues/2403)

crates/services/p2p/src/codecs.rs
netrome: > I'm not sure I follow completely here, do you mean changing the type of econded_data? 
Yes, the suggestion taking `encoded_data: impl std::io::Read + 'a`, instead of explicitly taking a slice here since it allows users to pass in any reader, which is more flexible. However, I see that we use `AsyncRead` and not `Read` for the network communication which would be a bit more messy so let's not do this.
> If we do so, we need to allocate a second buffer where to read the value into. Or do you have something else in mind?
That shouldn't be necessary since `&[u8]` also implements `Read`, so all current calls should still work. Or perhaps I'm misunderstanding you here.

crates/services/p2p/src/codecs/gossipsub.rs
```diff
@@ -0,0 +1,53 @@
...
+        + Decode<Transaction, Error = io::Error>
+        + Send,
+{
+    type RequestMessage = GossipsubBroadcastRequest;
+    type ResponseMessage = GossipsubMessage;
+
+    fn encode(&self, data: Self::RequestMessage) -> Result<Vec<u8>, io::Error> {
+        match data {
+            GossipsubBroadcastRequest::NewTx(tx) => {
+                Ok(self.codec.encode(&*tx)?.as_bytes().into_owned())
```
netrome, line 35: Nit: The dereference is redundant here.

```suggestion
                Ok(self.codec.encode(&tx)?.as_bytes().into_owned())
```

crates/services/p2p/src/codecs/postcard.rs
```diff
@@ -1,219 +1,85 @@
...
-            GossipTopicTag::NewTx => GossipsubMessage::NewTx(deserialize(encoded_data)?),
-        };
-
-        Ok(decoded_response)
+impl<T> Encode<T> for PostcardCodec
+where
+    T: ?Sized + serde::Serialize,
+{
+    type Encoder<'a> = Cow<'a, [u8]> where T: 'a;
+    type Error = io::Error;
```
netrome, line 43: Why do we return an IO error here instead of just propagating the postcard error?

crates/services/p2p/src/codecs/request_response.rs
```diff
@@ -0,0 +1,150 @@
...
+    RequestResponseProtocols,
+};
+
+#[derive(Debug, Clone)]
+pub struct RequestResponseMessageHandler<Codec> {
+    pub(crate) codec: Codec,
+    /// Used for `max_size` parameter when reading Response Message
+    /// Necessary in order to avoid DoS attacks
+    /// Currently the size mostly depends on the max size of the Block
+    pub(crate) max_response_size: usize,
```
netrome, line 33: We seem to convert this one to a `u64` wherever used. How about defining it as such?

```suggestion
    pub(crate) max_response_size: u64,
```

crates/services/p2p/src/codecs/request_response.rs
```diff
@@ -0,0 +1,150 @@
...
+        &mut self,
+        _protocol: &Self::Protocol,
+        socket: &mut T,
+    ) -> io::Result<Self::Request>
+    where
+        T: AsyncRead + Unpin + Send,
+    {
+        let mut response = Vec::new();
+        socket
+            .take(self.max_response_size as u64)
```
netrome, line 68: This `as` conversion can theoretically error. Personally I'd prefer to use `u64::try_from` explicitly instead, or as suggested before - redefine `max_response_size` as a `u64`.

crates/services/p2p/src/codecs/request_response.rs
```diff
@@ -0,0 +1,150 @@
...
+impl<Codec> RequestResponseProtocols for Codec
+where
+    Codec: request_response::Codec<Protocol = RequestResponseProtocol>,
+{
+    fn get_req_res_protocols(
+        &self,
+    ) -> impl Iterator<Item = <Self as request_response::Codec>::Protocol> {
+        // TODO: Iterating over versions in reverse order should prefer
+        // peers to use V2 over V1 for exchanging messages. However, this is
+        // not guaranteed by the specs for the `request_response` protocol.
```
netrome, line 147: When should we do this? Can you create a follow-up?

netrome (changes requested):
Overall looks good to me. There's a dangling todo that needs to reference an issue, and a few potentially dangerous `as` conversions I'd prefer to see removed. Other than that, great stuff!

crates/services/p2p/src/codecs/postcard.rs
acerone85: Previously the implementation of serialize was 

```rust 
fn serialize<D: Serialize>(data: &D) -> Result<Vec<u8>, io::Error> {
    postcard::to_stdvec(&data)
        .map_err(|e| io::Error::new(io::ErrorKind::Other, e.to_string()))
}
```
so I left the same error type. 
We can change the error type, but I would not do it in this PR. I usually avoid introducing any non-functional changes in refactoring PRs (that also includes fixing bugs), as any functional change should be handled separately (or at least in a separate commit).

crates/services/p2p/src/codecs/request_response.rs
acerone85: Same comment as above, I avoid functional changes in refactoring PRs. 

However in this case I have created a reference issue #2459, referenced it in the code (see b94c4519c2), and opened a PR for it (https://github.com/FuelLabs/fuel-core/pull/2460).

crates/services/p2p/src/codecs/postcard.rs
netrome: Fair enough :+1:

crates/services/p2p/src/codecs/request_response.rs
netrome: Thanks, makes sense :+1:

crates/services/p2p/src/codecs/request_response.rs
netrome: Wonderful thanks!

netrome (approved):
Thanks for updating! :pray:

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
 - [2429](https://github.com/FuelLabs/fuel-core/pull/2429): Introduce custom enum for representing result of running service tasks
 - [2377](https://github.com/FuelLabs/fuel-core/pull/2377): Add more errors that can be returned as responses when using protocol `/fuel/req_res/0.0.2`. The errors supported are `ProtocolV1EmptyResponse` (status code `0`) for converting empty responses sent via protocol `/fuel/req_res/0.0.1`, `RequestedRangeTooLarge`(status code `1`) if the client requests a range of objects such as sealed block headers or transactions too large, `Timeout` (status code `2`) if the remote peer takes too long to fulfill a request, or `SyncProcessorOutOfCapacity` if the remote peer is fulfilling too many requests concurrently.
+- [2388](https://github.com/FuelLabs/fuel-core/pull/2388): Rework the P2P service codecs to avoid unnecessary coupling between components. The refactoring makes it explicit that the Gossipsub and RequestResponse codecs only share encoding/decoding functionalities from the Postcard codec. It also makes handling Gossipsub and RequestResponse messages completely independent of each other.
```
xgreenx, line 40: Sorry for keeping PR un reviewed for long time. Could you please move it to relevant section please?=D

crates/services/p2p/src/codecs.rs
```diff
@@ -1,18 +1,60 @@
...
+use libp2p::request_response as libp2p_request_response;
+
+use std::{
+    borrow::Cow,
+    io,
 };
-use libp2p::request_response;
-use std::io;
+
+// TODO: https://github.com/FuelLabs/fuel-core/issues/2403
```
xgreenx, line 13: We closed the issue, so I think you can remove todos

crates/services/p2p/src/codecs/gossipsub.rs
```diff
@@ -0,0 +1,53 @@
...
+#[derive(Debug, Clone, Default)]
+pub struct GossipsubMessageHandler<Codec> {
+    pub(crate) codec: Codec,
+}
+
+impl<Codec> GossipsubCodec for GossipsubMessageHandler<Codec>
+where
+    Codec: Encode<Transaction, Error = io::Error>
+        + Decode<Transaction, Error = io::Error>
+        + Send,
```
xgreenx, line 27:
```suggestion
    Codec: Encode<Transaction, Error = io::Error>
        + Decode<Transaction, Error = io::Error>,
```

`Encode` already is `Send`

crates/services/p2p/src/codecs/gossipsub.rs
```diff
@@ -0,0 +1,53 @@
...
+{
+    type RequestMessage = GossipsubBroadcastRequest;
+    type ResponseMessage = GossipsubMessage;
+
+    fn encode(&self, data: Self::RequestMessage) -> Result<Vec<u8>, io::Error> {
+        match data {
+            GossipsubBroadcastRequest::NewTx(tx) => {
+                Ok(self.codec.encode(&tx)?.as_bytes().into_owned())
```
xgreenx, line 35: If you convert it into owned, I think it will be better to return vector from the `Encoder` trait itself. Because right now you create two vectors.

crates/services/p2p/src/codecs/request_response.rs
xgreenx: `usize -> u64` can't error until people switched to `u128` as a base=D

xgreenx (commented):
Looks good to me, small comments=)

crates/services/p2p/src/behavior.rs
```diff
@@ -70,13 +70,16 @@ pub struct FuelBehaviour {
...
+    /// The Behaviour to manage connection limits.
+    connection_limits: connection_limits::Behaviour,
 }
 
 impl FuelBehaviour {
     pub(crate) fn new(
         p2p_config: &Config,
         request_response_codec: RequestResponseMessageHandler<PostcardCodec>,
-    ) -> anyhow::Result<Self> {
+    ) -> anyhow::Result<Self, anyhow::Error> {
```
netrome, line 82: This looks wrong. `anyhow::Result` should only take the return type as a parameter.

netrome (changes requested):
Just a weird anyhow result that should be updated. Otherwise still LGTM

crates/services/p2p/src/behavior.rs
netrome: I'm surprised this isn't caught in CI :thinking:

crates/services/p2p/src/codecs/request_response.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    async fn write_response<T>(
+        &mut self,
+        protocol: &Self::Protocol,
+        socket: &mut T,
+        res: Self::Response,
+    ) -> io::Result<()>
+    where
+        T: futures::AsyncWrite + Unpin + Send,
```
rafal-ch, line 120: Nit, for consistency with the `AsyncRead` above.

```suggestion
        T: AsyncWrite + Unpin + Send,
```

crates/services/p2p/src/codecs/request_response.rs
```diff
@@ -0,0 +1,153 @@
...
+impl<Codec> RequestResponseProtocols for Codec
+where
+    Codec: request_response::Codec<Protocol = RequestResponseProtocol>,
+{
+    fn get_req_res_protocols(
+        &self,
+    ) -> impl Iterator<Item = <Self as request_response::Codec>::Protocol> {
+        // TODO: https://github.com/FuelLabs/fuel-core/issues/2458.
+        // Iterating over versions in reverse order should prefer
+        // peers to use V2 over V1 for exchanging messages. However, this is
```
rafal-ch, line 149: Question, why is `V1` marked as default [here](https://github.com/FuelLabs/fuel-core/pull/2388/files#diff-df0c3b6a4b16bbcc620b334f6a644a054bd7d99f792c2c484b3cceeef87237c8R8-R9) if we want to prefer `V2`.

rafal-ch (dismissed):
Nice one 👍 
Left just some nitpics and a question

CHANGELOG.md
acerone85: Done as part of conflict handling during merge: fa021c53fc84ff0c88501dfcfef5ba2c5cfdb03a

crates/services/p2p/src/codecs.rs
acerone85: All todos related to the issue removed: a9c9eca86f743b57c24ecb8588a4ea7265d033de

crates/services/p2p/src/behavior.rs
acerone85: Amended in cbfd9a7b609ce024be25f57706d7d58bb9827ffe

crates/services/p2p/src/codecs/request_response.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    async fn write_request<T>(
+        &mut self,
+        _protocol: &Self::Protocol,
+        socket: &mut T,
+        req: Self::Request,
+    ) -> io::Result<()>
+    where
+        T: futures::AsyncWrite + Unpin + Send,
```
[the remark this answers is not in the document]
acerone85: refactored in 5063decfe6

crates/services/p2p/src/codecs/request_response.rs
acerone85: refactored in 5063decfe6

crates/services/p2p/src/p2p_service.rs
```diff
@@ -123,8 +125,8 @@ pub struct FuelP2PService {
     /// to the peer that requested it.
     inbound_requests_table: HashMap<InboundRequestId, ResponseChannel<V2ResponseMessage>>,
 
-    /// NetworkCodec used as `<GossipsubCodec>` for encoding and decoding of Gossipsub messages
-    network_codec: PostcardCodec,
+    /// `UboundedCodec` as GossipsubCodec for encoding and decoding of Gossipsub messages
```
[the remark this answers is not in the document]
acerone85: yeah, this is supposed to be `PostcardCodec`. Changed in 07924f6d12

crates/services/p2p/src/codecs/request_response.rs
acerone85: That's a good question, and I don't remember the answer. But de facto we never invoke default, and i have removed the #[derive(Default)] clause in ad57a6bbe2. I can revert the commit if we want to keep the default.

rafal-ch: I'm ok without the unused default, thanks.

rafal-ch (approved):

crates/services/p2p/src/codecs/gossipsub.rs
netrome: I prefer the Cow over the Vec for these low-level traits. A Cow can be converted to a vec without cloning using `Vec::from`. See https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-From%3CCow%3C'a,+%5BT%5D%3E%3E-for-Vec%3CT%3E

So the problem here is the `.as_bytes().into_owned()` rather than the use `Cow`.

netrome (dismissed):
Approved, but I'd prefer to see https://github.com/FuelLabs/fuel-core/commit/d4e21988d8ffc7863f2f6a3fc9069af6197c0043 reverted.

xgreenx (approved):

crates/services/p2p/src/codecs.rs
```diff
@@ -1,18 +1,48 @@
...
+/// The trait decodes the type from the bytes.
+pub trait Decode<T> {
+    type Error;
+    /// Decodes the type `T` from the bytes.
+    fn decode(&self, bytes: &[u8]) -> Result<T, Self::Error>;
+}
+
+impl<'a> Encoder for Cow<'a, [u8]> {
+    fn to_bytes(self) -> Vec<u8> {
+        self.into()
```
xgreenx, line 43:
```suggestion
        self.into_owned()
```

netrome (dismissed):

xgreenx (approved):

netrome (approved):

Issue #1188, opened 2023-05-25
Add unit test for p2p to check that node handles appropriately the case of reserved nodes containing self multiaddr


xgreenx:
Maybe it could be a root cause to https://github.com/FuelLabs/fuel-core/issues/1190

PR #2390, opened 2024-10-24
Add test node herself in reserved nodes.

AurelienFT:
## Linked Issues/PRs
Closes https://github.com/FuelLabs/fuel-core/issues/1188

## Description
Add a test to verfiy that we don't connect to ourself if we are in the reserved nodes.

## Checklist

### Before requesting review

### DIFF
--- a/crates/services/p2p/src/p2p_service.rs
+++ b/crates/services/p2p/src/p2p_service.rs
@@ -1063,7 +1063,7 @@ mod tests {
         // to have the same PeerId
         let node_b = {
             // Given
-            p2p_config.reserved_nodes = vec![node_a.multiaddrs().pop().unwrap()];
+            p2p_config.reserved_nodes = node_a.multiaddrs();
             let max_block_size = p2p_config.max_block_size;
             let (sender, _) =
                 broadcast::channel(p2p_config.reserved_nodes.len().saturating_add(1));
--- a/crates/services/p2p/src/p2p_service.rs
+++ b/crates/services/p2p/src/p2p_service.rs
@@ -897,6 +897,7 @@ mod tests {
         net::{
             IpAddr,
             Ipv4Addr,
+            TcpListener,
         },
         ops::Range,
         sync::Arc,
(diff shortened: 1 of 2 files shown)


AurelienFT:
Seems like the test is timing out on CI. Changing this to draft

crates/services/p2p/src/p2p_service.rs
```diff
@@ -1036,6 +1041,31 @@ mod tests {
         stop_sender.send(()).unwrap();
     }
 
+    #[tokio::test]
+    #[instrument]
+    async fn our_node_in_reserved_nodes() {
+        let mut p2p_config = Config::default_initialized("own_node_in_reserved_nodes");
+        p2p_config.address = IpAddr::V4(Ipv4Addr::from([0, 0, 0, 0]));
+        p2p_config.tcp_port = 11111;
```
xgreenx, line 1049: When we run many tests, this port can be already used and then test will be incorrect. Maybe it is better to request the port from the OS and start a node on this port. It may require modification of the `build_service_from_config` to allow bypassing of the port binding.

AurelienFT: I found a solution that is IMO the best solution (that doesn't require to change all of the code just for this test) and as it seems a bit hacky I added a big comment that, I hope, explain all decisions taken.

crates/services/p2p/src/p2p_service.rs
```diff
@@ -1036,6 +1042,63 @@ mod tests {
...
+    #[tokio::test]
+    #[instrument]
+    async fn our_node_in_reserved_nodes() {
+        let mut retries = 10;
+        // We use bind to get a random port for the node to listen on.
+        // We use a loop because the port might be taken between the time we drop the listener and it's used by libp2p.
+        // We don't use `build_service_from_config` because we want to drop the tcp_listener at the last moment.
+        // and we don't want to `.unwrap()` if the port is taken but we prefer retry.
+        let mut node = loop {
+            let tcp_listener = TcpListener::bind("127.0.0.1:0").unwrap();
```
xgreenx, line 1054: Maybe it is somehow possible pass the binding into the `FuelP2PService` and use in the p2p?

xgreenx (commented):
The task is more about the same `peer_id`, not the port or address.

crates/services/p2p/src/p2p_service.rs
AurelienFT: I tried to find a method to give something else than a `MultiAddr` and for instance a listener. Didn't found any. If you find one I will happy to use it

AurelienFT:
> The task is more about the same `peer_id`, not the port or address.

You mean two different peer with two different port but the same peer id ? This simplify a lot ahah

xgreenx:
> You mean two different peer with two different port but the same peer id ? This simplify a lot ahah

Yeah=D

crates/services/p2p/src/p2p_service.rs
```diff
@@ -1053,6 +1053,47 @@ mod tests {
...
+    #[instrument]
+    async fn dont_connect_to_node_with_same_peer_id() {
+        let mut p2p_config =
+            Config::default_initialized("dont_connect_to_node_with_same_peer_id");
+        let mut node_a = build_service_from_config(p2p_config.clone()).await;
+        // We don't use build_service_from_config here, because we want to use the same keypair
+        // to have the same PeerId
+        let node_b = {
+            // Given
+            p2p_config.reserved_nodes = vec![node_a.multiaddrs().pop().unwrap()];
```
xgreenx, line 1066:
```suggestion
            p2p_config.reserved_nodes = node_a.multiaddrs();
```

xgreenx (approved):

rafal-ch (approved):
LGTM :+1:

PR #2396, opened 2024-10-25
perf: Add support for multi get operation for database queries

### LINKED ISSUE #2344: Add support for multi-get operation in the Database

netrome:
### Linked Issues
Closes #2344 

### Description
This PR adds support for getting batches of values from the storage through a new `StorageBatchInspect` trait in `fuel-core-storage`.

This trait is implemented for `StructuredStorage` and `GenericDatabase` through a new `get_batch` method added to the `KeyValueInspect` and `BluePrintInspect` traits. The `get_batch` method is implemented with the `multi-get` operation for the RocksDB based storage implementations, and uses a default blanket implementation for in-memory implementations.

## Checklist

### Before requesting review

### DIFF
--- a/crates/client/src/client/schema/primitives.rs
+++ b/crates/client/src/client/schema/primitives.rs
@@ -177,7 +177,7 @@ impl LowerHex for TxPointer {
     }
 }
 
-#[derive(cynic::Scalar, Debug, Clone)]
+#[derive(cynic::Scalar, Debug, Clone, PartialEq, Eq)]
 pub struct HexString(pub Bytes);
 
 impl From<HexString> for Vec<u8> {
@@ -194,7 +194,7 @@ impl Deref for HexString {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Bytes(pub Vec<u8>);
 
 impl FromStr for Bytes {
--- a/crates/fuel-core/src/graphql_api/database.rs
+++ b/crates/fuel-core/src/graphql_api/database.rs
@@ -151,8 +151,6 @@ impl ReadView {
         &self,
         tx_ids: Vec<TxId>,
     ) -> Vec<StorageResult<Transaction>> {
-        // TODO: Use multiget when it's implemented.
-        //  https://github.com/FuelLabs/fuel-core/issues/2344
         let on_chain_results: BTreeMap<_, _> = tx_ids
             .iter()
             .enumerate()
@@ -178,11 +176,7 @@ impl ReadView {
         let mut results = on_chain_results;
         results.extend(off_chain_results);
 
-        // let result = tx_ids
-        //    .iter()
-        //    .map(|tx_id| self.transaction(tx_id))
-        //    .collect::<Vec<_>>();
-        // Give a chance to other tasks to run.
+        // Give a chance for other tasks to run.
         tokio::task::yield_now().await;
 
         results.into_values().collect()
--- a/crates/fuel-core/src/graphql_api/database.rs
+++ b/crates/fuel-core/src/graphql_api/database.rs
@@ -165,7 +165,7 @@ impl ReadView {
         }
     }
 
-    pub async fn extend_with_off_chain_results(
+    async fn extend_with_off_chain_results(
         &self,
         on_chain_results: BTreeMap<(usize, &TxId), StorageResult<Transaction>>,
     ) -> Vec<StorageResult<Transaction>> {
--- a/crates/fuel-core/src/graphql_api/database.rs
+++ b/crates/fuel-core/src/graphql_api/database.rs
@@ -9,8 +9,8 @@ use fuel_core_services::yield_stream::StreamYieldExt;
 use fuel_core_storage::{
     iter::{
         BoxedIterSend,
-        IntoBoxedIterSend,
         IntoBoxedIter,
+        IntoBoxedIterSend,
         IterDirection,
     },
     transactional::AtomicView,
--- a/crates/fuel-core/src/graphql_api/database.rs
+++ b/crates/fuel-core/src/graphql_api/database.rs
@@ -147,12 +147,26 @@ impl ReadView {
     }
 
     pub async fn transactions(&self, tx_ids: &[TxId]) -> Vec<StorageResult<Transaction>> {
-        let on_chain_results: BTreeMap<_, _> = tx_ids
-            .iter()
-            .enumerate()
-            .zip(self.on_chain.transactions(tx_ids.iter().into_boxed()))
+        let transactions: Vec<_> = self
+            .on_chain
+            .transactions(tx_ids.iter().into_boxed())
             .collect();
 
+        // Give a chance for other tasks to run.
+        tokio::task::yield_now().await;
+
+        if transactions.iter().any(|result| result.is_not_found()) {
+            let on_chain_results = tx_ids.iter().enumerate().zip(transactions).collect();
+
+            self.extend_with_off_chain_results(on_chain_results).await
+        } else {
+            transactions
+        }
+    }
+    pub async fn extend_with_off_chain_results(
+        &self,
+        on_chain_results: BTreeMap<(usize, &TxId), StorageResult<Transaction>>,
+    ) -> Vec<StorageResult<Transaction>> {
         let off_chain_indexed_txids: Vec<_> = on_chain_results
             .iter()
             .filter_map(|(indexed_tx_id, result)| {
--- a/crates/fuel-core/src/graphql_api/database.rs
+++ b/crates/fuel-core/src/graphql_api/database.rs
@@ -147,10 +147,7 @@ impl ReadView {
     }
 
     // TODO: Test case to ensure order is preserved
-    pub async fn transactions(
-        &self,
-        tx_ids: Vec<TxId>,
-    ) -> Vec<StorageResult<Transaction>> {
+    pub async fn transactions(&self, tx_ids: &[TxId]) -> Vec<StorageResult<Transaction>> {
         let on_chain_results: BTreeMap<_, _> = tx_ids
             .iter()
             .enumerate()
--- a/crates/fuel-core/src/graphql_api/database.rs
+++ b/crates/fuel-core/src/graphql_api/database.rs
@@ -163,6 +163,7 @@ impl ReadView {
             transactions
         }
     }
+
     pub async fn extend_with_off_chain_results(
         &self,
         on_chain_results: BTreeMap<(usize, &TxId), StorageResult<Transaction>>,
--- a/crates/fuel-core/src/graphql_api/database.rs
+++ b/crates/fuel-core/src/graphql_api/database.rs
@@ -12,8 +12,6 @@ use fuel_core_storage::{
         IntoBoxedIter,
         IterDirection,
     },
-    not_found,
-    tables::Transactions,
     transactional::AtomicView,
     Error as StorageError,
     IsNotFound,
@@ -65,6 +63,7 @@ use fuel_core_types::{
 use futures::Stream;
 use std::{
     borrow::Cow,
+    collections::BTreeMap,
     sync::Arc,
 };
 
@@ -141,29 +140,52 @@ impl ReadView {
     pub fn transaction(&self, tx_id: &TxId) -> StorageResult<Transaction> {
         let result = self.on_chain.transaction(tx_id);
         if result.is_not_found() {
-            if let Some(tx) = self.off_chain.old_transaction(tx_id)? {
-                Ok(tx)
-            } else {
-                Err(not_found!(Transactions))
-            }
+            self.off_chain.old_transaction(tx_id)
         } else {
             result
         }
     }
 
+    // TODO: Test case to ensure order is preserved
     pub async fn transactions(
         &self,
         tx_ids: Vec<TxId>,
     ) -> Vec<StorageResult<Transaction>> {
         // TODO: Use multiget when it's implemented.
         //  https://github.com/FuelLabs/fuel-core/issues/2344
-        let result = tx_ids
+        let on_chain_results: BTreeMap<_, _> = tx_ids
             .iter()
-            .map(|tx_id| self.transaction(tx_id))
-            .collect::<Vec<_>>();
+            .enumerate()
+            .zip(self.on_chain.transactions(tx_ids.iter().into_boxed()))
+            .collect();
+
+        let off_chain_indexed_txids: Vec<_> = on_chain_results
+            .iter()
+            .filter_map(|(indexed_tx_id, result)| {
+                result.is_not_found().then_some(*indexed_tx_id)
+            })
+            .collect();
+
+        let off_chain_results = off_chain_indexed_txids.iter().copied().zip(
+            self.off_chain.old_transactions(
+                off_chain_indexed_txids
+                    .iter()
+                    .map(|(_, tx_id)| *tx_id)
+                    .into_boxed(),
+            ),
+        );
+
+        let mut results = on_chain_results;
+        results.extend(off_chain_results);
+
+        // let result = tx_ids
+        //    .iter()
+        //    .map(|tx_id| self.transaction(tx_id))
+        //    .collect::<Vec<_>>();
         // Give a chance to other tasks to run.
         tokio::task::yield_now().await;
-        result
+
+        results.into_values().collect()
     }
 
     pub fn block(&self, height: &BlockHeight) -> StorageResult<CompressedBlock> {
--- a/crates/fuel-core/src/graphql_api/database.rs
+++ b/crates/fuel-core/src/graphql_api/database.rs
@@ -8,7 +8,8 @@ use crate::fuel_core_graphql_api::{
 use fuel_core_services::yield_stream::StreamYieldExt;
 use fuel_core_storage::{
     iter::{
-        BoxedIter,
+        BoxedIterSend,
+        IntoBoxedIterSend,
         IntoBoxedIter,
         IterDirection,
     },
@@ -205,7 +206,7 @@ impl ReadView {
         &self,
         height: Option<BlockHeight>,
         direction: IterDirection,
-    ) -> BoxedIter<'_, StorageResult<CompressedBlock>> {
+    ) -> BoxedIterSend<'_, StorageResult<CompressedBlock>> {
         // Chain together blocks from the off-chain db and the on-chain db
         // The blocks in off-chain db, if any, are from time before regenesis
 
@@ -218,12 +219,12 @@ impl ReadView {
                     .on_chain
                     .blocks(Some(height), direction)
                     .chain(self.off_chain.old_blocks(None, direction))
-                    .into_boxed(),
+                    .into_boxed_send(),
                 (false, IterDirection::Forward) => self
                     .off_chain
                     .old_blocks(Some(height), direction)
                     .chain(self.on_chain.blocks(None, direction))
-                    .into_boxed(),
+                    .into_boxed_send(),
                 (false, IterDirection::Reverse) => {
                     self.off_chain.old_blocks(Some(height), direction)
                 }
@@ -234,12 +235,12 @@ impl ReadView {
                     .off_chain
                     .old_blocks(None, direction)
                     .chain(self.on_chain.blocks(None, direction))
-                    .into_boxed(),
+                    .into_boxed_send(),
                 IterDirection::Reverse => self
                     .on_chain
                     .blocks(None, direction)
                     .chain(self.off_chain.old_blocks(None, direction))
-                    .into_boxed(),
+                    .into_boxed_send(),
             }
         }
     }
--- a/crates/fuel-core/src/graphql_api/database.rs
+++ b/crates/fuel-core/src/graphql_api/database.rs
@@ -146,7 +146,6 @@ impl ReadView {
         }
     }
 
-    // TODO: Test case to ensure order is preserved
     pub async fn transactions(&self, tx_ids: &[TxId]) -> Vec<StorageResult<Transaction>> {
         let on_chain_results: BTreeMap<_, _> = tx_ids
             .iter()
--- a/crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
+++ b/crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
@@ -41,6 +41,7 @@ use fuel_core_storage::{
     Error as StorageError,
     Result as StorageResult,
     StorageAsRef,
+    StorageBatchInspect,
 };
 use fuel_core_types::{
     blockchain::{
@@ -166,10 +167,21 @@ impl OffChainDatabase for OffChainIterableKeyValueView {
             .into_owned())
     }
 
-    fn old_transaction(&self, id: &TxId) -> StorageResult<Option<Transaction>> {
+    fn old_transaction(&self, id: &TxId) -> StorageResult<Transaction> {
         self.storage_as_ref::<OldTransactions>()
-            .get(id)
-            .map(|tx| tx.map(|tx| tx.into_owned()))
+            .get(id)?
+            // Note: Breaking change (previously we returned not_found!(Transactions))
+            .ok_or(not_found!(OldTransactions))
+            .map(|tx| tx.into_owned())
+    }
+
+    fn old_transactions<'a>(
+        &'a self,
+        ids: BoxedIter<'a, &'a TxId>,
+    ) -> BoxedIter<'a, StorageResult<Transaction>> {
+        <Self as StorageBatchInspect<OldTransactions>>::get_batch(self, ids)
+            .map(|result| result.and_then(|opt| opt.ok_or(not_found!(OldTransactions))))
+            .into_boxed()
     }
 
     fn relayed_tx_status(
--- a/crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
+++ b/crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
@@ -26,10 +26,10 @@ use fuel_core_storage::{
     blueprint::BlueprintInspect,
     codec::Encode,
     iter::{
-        BoxedIterSend,
         BoxedIter,
-        IntoBoxedIterSend,
+        BoxedIterSend,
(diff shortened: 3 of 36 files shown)


crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -141,29 +140,42 @@ impl ReadView {
...
 
-    pub async fn transactions(
-        &self,
-        tx_ids: Vec<TxId>,
-    ) -> Vec<StorageResult<Transaction>> {
-        // TODO: Use multiget when it's implemented.
-        //  https://github.com/FuelLabs/fuel-core/issues/2344
-        let result = tx_ids
+    pub async fn transactions(&self, tx_ids: &[TxId]) -> Vec<StorageResult<Transaction>> {
+        let on_chain_results: BTreeMap<_, _> = tx_ids
```
xgreenx, line 150: It is nice that this function handles this hard use case.

In most cases, we have only `on_chain` data unless we did regenesis. It would be nice if this did not affect performance for the primary use case.

Can we first request values from the on-chain database, and if any of them are not found, we will fall back into the logic with `enumerate` and `BTreeMap`?

crates/storage/src/lib.rs
```diff
@@ -176,6 +177,17 @@ pub trait StorageBatchMutate<Type: Mappable>: StorageMutate<Type> {
...
 
+/// Allows getting batches of values from the storage,
+/// which can be faster in certain implementations than
+/// getting values one by one.
+pub trait StorageBatchInspect<Type: Mappable> {
+    /// Get a batch of values associated with the provided keys.
+    fn get_batch<'a>(
+        &'a self,
+        keys: BoxedIter<'a, &'a Type::Key>,
+    ) -> BoxedIter<'a, Result<Option<Type::OwnedValue>>>;
```
xgreenx, line 188: Why we can't use generic for the iterator here and don't use `StorageBatchInspect` in the `den` context?

crates/storage/src/blueprint.rs
```diff
@@ -111,6 +139,7 @@ where
     fn delete(storage: &mut S, key: &M::Key, column: S::Column) -> StorageResult<()>;
 }
 
+// TODO: Rename
```
xgreenx, line 142: Could you add more description to the todo please?=D

crates/storage/src/lib.rs
xgreenx: Also curios, is that possible to return `impl Iterator<Item = Result<Option<Type::OwnedValue>>>`?=D

crates/storage/src/kv_store.rs
```diff
@@ -62,6 +66,15 @@ pub trait KeyValueInspect {
     /// Returns the value from the storage.
     fn get(&self, key: &[u8], column: Self::Column) -> StorageResult<Option<Value>>;
 
+    /// Returns multiple values from the storage.
+    fn get_batch<'a>(
+        &'a self,
+        keys: BoxedIter<'a, Vec<u8>>,
```
xgreenx, line 72: It is very sad that we need to convert `&[u8]` into a vector. Are you use that we can't pass `Cow<'a, [u8]>`?

crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -712,20 +748,36 @@ where
...
+            .map_err(|e| StorageError::Other(DatabaseError::Other(e.into()).into()));
 
-        if let Some(value) = &value {
-            self.metrics.bytes_read.inc_by(value.len() as u64);
-        }
+        self.register_read(result, column.id())
+            .map(|opt| opt.map(Arc::new))
+    }
+
+    fn get_batch<'a>(
```
xgreenx, line 760: We already have `multi_get` function that you can use, you don't need to add a new one=)

crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -712,20 +748,36 @@ where
...
+    ) -> BoxedIter<'a, StorageResult<Option<Value>>> {
+        let column_family = self.cf(column);
+        let keys = keys.map(|key| (&column_family, key));
 
-        Ok(value.map(Arc::new))
+        self.db
+            .multi_get_cf_opt(keys, &self.read_options)
+            .into_iter()
+            .map(move |result| {
+                self.register_read(
```
xgreenx, line 772: I like the old approach more, because we called `self.metrics.columns_read_statistic.get(&column)` once, while in you case you will call it for each key, that could be expensive since it uses a lock inside.


image

crates/fuel-core/src/graphql_api/database.rs
netrome: Good idea, I'll do it.

crates/storage/src/lib.rs
netrome: Yeah that's what I wanted to do initially, relating to this discussion https://github.com/FuelLabs/fuel-core/issues/2344#issuecomment-2433432356

The problem is that we need a big refactor to implement it for `StructuredStorage` or `GenericDatabase` because they are currently used in trait objects in several places.

There could be something I'm missing here, but ultimately when we use this in the GraphQL API the database is held through a trait object one way or another. I'd love to hear any suggestions if you have any ideas about how to get around this, but anything I have thought of so far would require refactoring the graphql API to use generics.

I would be happy to do the GraphQL API refactor to allow us to work around the trait objects here, but last time we discussed it my understanding was that getting the multi-get implementation with boxed iterators in place was higher priority and that we should hold off any refactors of the GraphQL API since we have bigger refactors planned for it.

crates/storage/src/blueprint.rs
netrome: Oh missed this todo... In previous iterations I added another trait next to this, in which case it would be nice to rename this one. That's no longer needed. I'll remove the todo comment.

crates/storage/src/kv_store.rs
netrome: Thanks for suggesting. Yeah I think a `Cow` can be used here. Originally I wanted to take `impl AsRef<[u8]>` but again that's a problem with object safety.

I'll see if I hit any walls with the `Cow` solution.

crates/fuel-core/src/state/rocks_db.rs
netrome: Ah I see. I'll revert to the old solution then. I missed the other `multi_get` implementation that already had metrics when I implemented this.

crates/fuel-core/src/state/rocks_db.rs
netrome: Oh right. One problem with the existing function is that it returns a `Result<Vec<...>>` whereas our use cases expect a `Vec<Result<...>>`. This should be easy to refactor as I just need to remove the `try_collect` and map the error, so I'll do that.

crates/storage/src/blueprint.rs
netrome: Removed in https://github.com/FuelLabs/fuel-core/pull/2396/commits/8b35460e2c60b3de75e0d32f58928aca264884c5

crates/fuel-core/src/state/rocks_db.rs
netrome: Reverted in https://github.com/FuelLabs/fuel-core/pull/2396/commits/f6977073d73ff04110a102bf5a972b4b3a71c316

crates/fuel-core/src/state/rocks_db.rs
netrome: Done in https://github.com/FuelLabs/fuel-core/pull/2396/commits/3d0a71f59870fc787e1e85d16ce8fcb76864054e

crates/fuel-core/src/graphql_api/database.rs
netrome: Done in https://github.com/FuelLabs/fuel-core/pull/2396/commits/5de80b77b587eed062f3dcfc5769f0d878f7dc3b

crates/storage/src/lib.rs
xgreenx: I meant something like this https://github.com/FuelLabs/fuel-core/pull/2419/files

netrome: Interesting, thanks. I'll have a look 👀

netrome: Ah I see. We don't need the `StorageBatchInspect` constraint on the GraphQL ports, which means that the whole implementation can be with iterators and we only need the boxes at the port traits. Thats muuuuch better, thanks for showing!

netrome: Ah right but we still need the boxed iterators for the `KeyValueInspect` trait. That's why I missed this. Because I tried there and I tried at the graphql level. But it's nice to skip the boxes for this trait in the middle at least.

netrome: Merged your proposal.

crates/storage/src/codec.rs
```diff
@@ -16,9 +16,12 @@ pub mod primitive;
 pub mod raw;
 
 /// The trait is usually implemented by the encoder that stores serialized objects.
-pub trait Encoder {
+pub trait Encoder<'a> {
     /// Returns the serialized object as a slice.
     fn as_bytes(&self) -> Cow<[u8]>;
+
+    /// Consumes the encoder, returning the serialized object as a slice.
+    fn into_bytes(self) -> Cow<'a, [u8]>;
```
xgreenx, line 24: https://github.com/FuelLabs/fuel-core/pull/2420

crates/storage/src/kv_store.rs
```diff
@@ -41,7 +44,7 @@ pub trait StorageColumn: Copy + core::fmt::Debug {
 
 /// The definition of the key-value inspection store.
 #[impl_tools::autoimpl(for<T: trait> &T, &mut T, Box<T>)]
-pub trait KeyValueInspect {
+pub trait KeyValueInspect: Send + Sync {
```
xgreenx, line 47: It is very sad that we need to require `Send + Sync` only because `BoxedIter` requires `Send`. Maybe we can define non Send boxed iterator instead and use it(because it seems we don't need `Send` feature, but maybe I'm wrong)?

crates/fuel-core/src/service/adapters/graphql_api/on_chain.rs
```diff
@@ -103,11 +119,54 @@ impl DatabaseMessages for OnChainIterableKeyValueView {
...
+
+        Ok(coin.uncompress(utxo_id))
+    }
+
+    fn coins<'a>(
+        &'a self,
+        utxo_ids: BoxedIter<'a, &'a UtxoId>,
+    ) -> BoxedIter<'a, StorageResult<fuel_core_types::entities::coins::coin::Coin>> {
+        let utxo_ids_1: Vec<_> = utxo_ids.collect();
+        let utxo_ids_2 = utxo_ids_1.clone();
```
xgreenx, line 155: Instead of collecting ids here, you could past `Vec<_>`.

Plus code allows you to do that=) 
image

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
```diff
@@ -166,10 +167,20 @@ impl OffChainDatabase for OffChainIterableKeyValueView {
...
+            .map(|tx| tx.into_owned())
+    }
+
+    fn old_transactions<'a>(
+        &'a self,
+        ids: BoxedIter<'a, &'a TxId>,
+    ) -> BoxedIter<'a, StorageResult<Transaction>> {
+        <Self as StorageBatchInspect<OldTransactions>>::get_batch(self, ids)
+            .map(|result| result.and_then(|opt| opt.ok_or(not_found!(OldTransactions))))
+            .into_boxed()
```
xgreenx, line 183: If you want, you can add the same syntax sugars that we did with `storage_as_ref` to avoid `<Self as StorageInspect<M>>` usage.

xgreenx (commented):
It would be nice to get some benchmarks(I guess you need to add new one that will use GraphQL) =)

crates/storage/src/kv_store.rs
netrome: Yeah this bound is just to satisfy the `BoxedIter` requirement. I can see if I can define a non-send one.

crates/fuel-core/src/service/adapters/graphql_api/on_chain.rs
netrome: Right that's better. I've simplified the port in https://github.com/FuelLabs/fuel-core/pull/2396/commits/df33ef2bde75eb7a445232371d1a87926756487d

crates/storage/src/codec.rs
netrome: Nice, approved and merged :)

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
netrome: Oh interesting, that would be nice. I'll look into it.

netrome:
> It would be nice to get some benchmarks(I guess you need to add new one that will use GraphQL) =)

Sure thing. I'm not sure if it makes sense to use criterion/cargo bench for end-to-end performance testing though as it is more geared towards micro-benchmarking, and we need to do something like the following.

1. Populate the database with a bunch of coins/transactions/messages etc.
2. Call the `DatabaseMessages::message_batch`, `DatabaseCoins::coins` etc. methods and record the lookup times.

I'll look into our options for this type of workload. Let me know if you have any opinions or thoughts on this.

crates/storage/src/kv_store.rs
netrome: Hitting some walls with this implementation. Will try again tomorrow with fresh eyes.

netrome: Alright my mistake was trying to change some iterators that we turn into streams, and they need to be `Send`. Now I have managed to get rid of these trait bounds https://github.com/FuelLabs/fuel-core/pull/2396/commits/cbb6efc4332490122e8de70d082c7048cc97e0a4.

Now we have two boxed iterators: `BoxedIter` which doesn't require Send, and `BoxedIterSend` for the cases when you need Send.

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
netrome: Turns out this is not as trivial as I had thought, as we're hitting some lifetime issues. 

### The problem
`get_batch` returns an iterator bound to the lifetime of the `self` parameter. I.e. the storage we call it on. This is necessary since we do `self.get(...)` within the `KeyValueInspect::get_batch` implementation.

So while we can implement `StorageBatchInspect` for the `StorageRef` type as:

```rust
impl<'a, S, Type> StorageBatchInspect<Type> for StorageRef<'a, S, Type>
where
    S: StorageBatchInspect<Type>,
    Type: Mappable,
{
    #[inline(always)]
    fn get_batch<'b, Iter>(
        &'b self,
        _keys: Iter,
    ) -> impl Iterator<Item = Result<Option<Type::OwnedValue>>> + 'b
    where
        Iter: 'b + IntoIterator<Item = &'b Type::Key>,
        Type::Key: 'b,
    {
        None.into_iter() // Note that we'd need access to `self.0` here which requires VM changes, or copying the `StorageRef` type into this crate. But that's a separate, and very manageable issue
    }
}
```

when we try to use it as

```rust
    fn old_transactions<'a>(
        &'a self,
        ids: BoxedIter<'a, &'a TxId>,
    ) -> BoxedIter<'a, StorageResult<Transaction>> {
        self.storage::<OldTransactions>()
            .get_batch(ids)
            .map(|result| result.and_then(|opt| opt.ok_or(not_found!(OldTransactions))))
            .into_boxed()
    }
```

we're hitting this issue
```
error[E0716]: temporary value dropped while borrowed
   --> crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs:183:9
    |
179 |       fn old_transactions<'a>(
    |                           -- lifetime `'a` defined here
...
183 |           self.storage::<OldTransactions>()
    |           -^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |           |
    |  _________creates a temporary value which is freed while still in use
    | |
184 | |             .get_batch(ids)
    | |___________________________- argument requires that borrow lasts for `'a`
...
187 |       }
    |       - temporary value is freed at the end of this statement
```

### Going forward
For this PR I suggest we don't go down this rabbit hole, but I'd be happy to do it as a follow-up if you think it's possible/interesting to explore it. Seems like a quite low prio though.

netrome:
> > It would be nice to get some benchmarks(I guess you need to add new one that will use GraphQL) =)
> 
> Sure thing. I'm not sure if it makes sense to use criterion/cargo bench for end-to-end performance testing though as it is more geared towards micro-benchmarking, and we need to do something like the following.
> 
>     1. Populate the database with a bunch of coins/transactions/messages etc.
> 
>     2. Call the `DatabaseMessages::message_batch`, `DatabaseCoins::coins` etc. methods and record the lookup times.
> 
> 
> I'll look into our options for this type of workload. Let me know if you have any opinions or thoughts on this.

I'll aim at defining these workloads as a stand-alone binary, which would allow us to use [hyperfine](https://github.com/sharkdp/hyperfine) to actually execute the benchmarks and interpret the results.

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -141,29 +141,57 @@ impl ReadView {
...
+        if transactions.iter().any(|result| result.is_not_found()) {
+            let on_chain_results = tx_ids.iter().enumerate().zip(transactions).collect();
+
+            self.extend_with_off_chain_results(on_chain_results).await
+        } else {
+            transactions
+        }
+    }
+
+    pub async fn extend_with_off_chain_results(
```
rafal-ch, line 168:
```suggestion
    async fn extend_with_off_chain_results(
```

crates/storage/src/codec.rs
```diff
@@ -16,19 +16,22 @@ pub mod primitive;
...
 }
 
 /// The trait encodes the type to the bytes and passes it to the `Encoder`,
 /// which stores it and provides a reference to it. That allows gives more
 /// flexibility and more performant encoding, allowing the use of slices and arrays
 /// instead of vectors in some cases. Since the [`Encoder`] returns `Cow<[u8]>`,
 /// it is always possible to take ownership of the serialized value.
-pub trait Encode<T: ?Sized> {
+pub trait Encode<T: ?Sized>: 'static {
```
rafal-ch, line 32: I'm not exactly sure why we need the `'static` bound here.

crates/client/src/client/schema/primitives.rs
```diff
@@ -177,7 +177,7 @@ impl LowerHex for TxPointer {
     }
 }
 
-#[derive(cynic::Scalar, Debug, Clone)]
+#[derive(cynic::Scalar, Debug, Clone, PartialEq, Eq)]
```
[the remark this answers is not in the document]
netrome: Right, it doesn't seem like these are needed anymore. I probably added them while iterating on the code, but I can cut them out. Though since this is a public type, I'd argue that we still should proactively add these common traits as per https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits

I can still cut them out if you want me to. Let me know what you think.

crates/client/src/client/schema/primitives.rs
```diff
@@ -194,7 +194,7 @@ impl Deref for HexString {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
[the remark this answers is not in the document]
netrome: Same reply :)

crates/storage/src/codec.rs
netrome: Right no this is just a leftover from a previous attempt to fix the lifetime issues with the `Encode` trait, but in the current solution this is redundant. Thanks for noticing!

netrome: Updated in https://github.com/FuelLabs/fuel-core/pull/2396/commits/c3b72642dacdbbe44fff3c6a3a95d68beb3ff6c5

crates/storage/src/blueprint.rs
```diff
@@ -75,6 +76,32 @@ where
             })
             .transpose()
     }
+
+    /// Returns multiple values from the storage.
+    fn get_batch<'a, Iter>(
+        storage: &'a S,
+        keys: Iter,
+        column: S::Column,
+    ) -> impl Iterator<Item = crate::Result<Option<M::OwnedValue>>> + 'a
```
acerone85, line 85: nit: 
```suggestion
    ) -> impl IntoIterator<Item = crate::Result<Option<M::OwnedValue>>> + 'a
```

`Iterator`s implement `IntoIterator`, so if you change the signature to return a `IntoIterator` then `StorageInspect` becomes slightly more flexible, as implementations can return either an iterator or a collection.

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -141,29 +141,57 @@ impl ReadView {
...
             result
         }
     }
 
-    pub async fn transactions(
+    pub async fn transactions(&self, tx_ids: &[TxId]) -> Vec<StorageResult<Transaction>> {
+        let transactions: Vec<_> = self
+            .on_chain
+            .transactions(tx_ids.iter().into_boxed())
+            .collect();
```
acerone85, line 154: I suspect you can avoid collecting here since later you iterate on transactions again.

netrome: Unforutnately I need to both iterate over the transactions and zip them with the on_chain_results - so we need to iterate twice over the tersults. We could potentially do something fancy with [Itertools::tee](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.tee), but I think this solution is simpler and easier on the eyes.

crates/storage/src/blueprint.rs
netrome: As a general principle, I like interfaces to be very general in what they accept but prefer them to be specific in what they return. (see [Robusness principle](https://en.wikipedia.org/wiki/Robustness_principle)). That's why I typically accept `IntoIterator` in parameters, because as you say then the caller doesn't have to care about doing `.into_iter()´ explicitly if they have a collection or anything.

Therefore, for return values it's better in my opinion to return the `Iterator` directly since it is more specific. Otherwise, all callers are basically forced to call `.into_iter()` before they can do anything with the returned iterator, which is less ergonomic. Any implementer of the trait can (and should) always call `.into_iter()` if they are about to return a collection. Again, not doing so would just force this burden upon the user of the function which is less than ideal without providing any benefit whatsoever.

Let me know if you disagree and I'll be happy to further discuss this.

netrome:
Got some initial benchmark results in https://github.com/FuelLabs/fuel-core/pull/2433, and it doesn't look good for the multi-get implementation. So far it's either slower or the same for all workloads I've tried. I've done the following:

* Measure end to end query times for getting transactions by owner over GraphQL, database populated by submitting a bunch of transactions (100k or 1M in most tests). with the same owner.
* Measure transaction lookup times using a ReadView directly reading a database pre-populated with 1M transactions (also tried different parameters here).
* Measure coin lookup times using a ReadView directly reading a database pre-populated with 1M coins. Tried with different parameters, and also with concurrent reads. Experimented with enabling/disabling caching in rocksdb.

I'm all ears if anyone has any suggestions for workloads to try, but until we can find evidence that this change actually is an improvement in any way we should put this PR on hold.

rymnc:
> Got some initial benchmark results in #2433, and it doesn't look good for the multi-get implementation. So far it's either slower or the same for all workloads I've tried. I've done the following:
> 
> * Measure end to end query times for getting transactions by owner over GraphQL, database populated by submitting a bunch of transactions (100k or 1M in most tests). with the same owner.
> * Measure transaction lookup times using a ReadView directly reading a database pre-populated with 1M transactions (also tried different parameters here).
> * Measure coin lookup times using a ReadView directly reading a database pre-populated with 1M coins. Tried with different parameters, and also with concurrent reads. Experimented with enabling/disabling caching in rocksdb.
> 
> I'm all ears if anyone has any suggestions for workloads to try, but until we can find evidence that this change actually is an improvement in any way we should put this PR on hold.

this is interesting, do you find any deviation from your impl and the benchmark code in https://github.com/FuelLabs/fuel-core/blob/896e4cfaa6e2235e686cc642ce235d080d8961d9/benches/src/db_lookup_times_utils/utils.rs#L82-L101 ? I would imagine the performance gains would be similar though

crates/client/src/client/schema/primitives.rs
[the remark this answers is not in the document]
rafal-ch: I don't have a strong opinion here, leaving this up to you :-)

crates/client/src/client/schema/primitives.rs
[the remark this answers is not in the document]
rafal-ch: resolving this - if some more discussion is needed, let's continue in the [other thread](https://github.com/FuelLabs/fuel-core/pull/2396/files#r1842356454).

crates/client/src/client/schema/primitives.rs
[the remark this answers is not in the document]
netrome: Alright, then I'll keep them around. It's unclear if this PR will me merged as for now, with the current benchmarks.

netrome:
> this is interesting, do you find any deviation from your impl and the benchmark code in
> 
> https://github.com/FuelLabs/fuel-core/blob/896e4cfaa6e2235e686cc642ce235d080d8961d9/benches/src/db_lookup_times_utils/utils.rs#L82-L101
> ? I would imagine the performance gains would be similar though

Yes, that code only tests the multi-get operation directly. I have no doubt that the multi-get operation gives a performance benefit in isolation, but with the current change it seems to be a net-negative performance effect. I suspect it's the boxed iterators, since each boxed iterator requires a vtable lookup to reference a single element if I'm not mistaken, and I would not be surprised if this overhead is enough to negate any positive effects of the multi-get. But we need to investigate this further before we can conclude anything for certain, and on my side this is de-prioritized in favor of the fraud-proving work.

With some larger refactors in the GraphQL crate, we should be able to avoid the boxed iterators - but we need to decide how to proceed and prioritize that work first.

netrome:
Closing this PR for now, and we can re-open it when we want to revisit this.

Issue #2044, opened 2024-07-23
transactionsByOwner graphql query doesn't work properly when providing `last`


xgreenx:
The next query return an empty response.

```query
query {
  transactionsByOwner(owner: "0x5502f46f90c3c68c85489b1f57ae6146c81e8549daf408574e3a9a0bb80c4ba0", last: 10) {
    nodes {
      id
    }
  }
}
```

We had the same problem in the past https://github.com/FuelLabs/fuel-core/issues/1040. But we should support it now. We need to investigate why it doesn't work.

mmyyrroonn:
Hi @xgreenx, I'd like to take a look at this issue. Before starting my investigation, let me clarify the situation. In #1040, you mentioned that "we don't allow backward queries." Then, in #1492, "backward iteration" was supported. The expected behavior should be that when providing 'last', the query returns a non-empty result, similar to the case with 'first', but with a different ordering. Am I missing anything?

xgreenx:
Hello, yes, you are correct=)

AurelienFT:
After running multiple queries and tests I didn't found any issue. 
One example : 
- The query with first 10 : 
```
query {
  transactionsByOwner(owner: "0x6b63804cfbf9856e68e5b6e7aef238dc8311ec55bec04df774003a2c96e0418e", first: 10) {
    nodes {
      id
    }
  }
}
```
Result : 
```
{
  "data": {
    "transactionsByOwner": {
      "nodes": [
        {
          "id": "0xc1936a67fffecf035816f545af60ea5f3361e032df2c610d9f4a37f9c86839c8"
        },
        {
          "id": "0xd5c0eeb9fcabc4c6349baee6d75e06c3fe607488fb2377307446fcab4c9ddeea"
        },
        {
          "id": "0xe390430d0ba417504a292e0b1fff738159cfb93d454f2051ee1b1fd40836b572"
        },
        {
          "id": "0x34ab45d03adda5fb5ac4a94a54fa027cfb0f3decfb0c411dd9ab8297f800f774"
        },
        {
          "id": "0xab5b6bc8550fd16c41784b23122c591ef9785db05ad5f05a7be4bbdc8187cef2"
        },
        {
          "id": "0x19df7dbb700054f620147c3e27a63545ce8b6c627a6900cae56dfaf27ade7eb9"
        }
      ]
    }
  }
}
- The query with last 10 : 
```
query {
[...]

PR #2398, opened 2024-10-25
Add first working version but not sure of the correctness.

AurelienFT:
## Linked Issues/PRs
Closes https://github.com/FuelLabs/fuel-core/issues/2044

### Problem analysis

Our database columns use the prefix extractor configuration which allow rocksdb to make optimizations on storage based on prefix. However, this breaks the lexicographic order between prefixs and we need to pass a special option to the read iterator to bypass the "prefix sharding" (source: https://github.com/facebook/rocksdb/wiki/Prefix-Seek/3a5e28faf6c2d0ec1bdb99763043e1e3322007e9#how-to-ignore-prefix-bloom-filters-in-read).

### Current solution

I took the same approach as before and started iteration from the next prefix using the bypass read option argument.

However I change the behavior to make only iterator to simplify and optimize the solution. But now, this iterator can include elements that have the next prefix in their key and so I added a combinator `skip_while` to skip them.

### Better solution for the future

I think that breaking the "prefix sharding" from RocksDB to iterate can make the iterator way more costly.

There is a way to avoid it because we can iterate in reverse order inside a prefix (source: https://github.com/facebook/rocksdb/wiki/SeekForPrev) but we need to get the maximum value for a key in a given prefix to use it as a starting point of our iteration which is not possible to get in the current code because we don't have any information about the key.

## Checklist

### Before requesting review

### DIFF
--- a/crates/fuel-core/src/state/rocks_db.rs
+++ b/crates/fuel-core/src/state/rocks_db.rs
@@ -436,12 +436,9 @@ where
     }
 
     /// RocksDB prefix iteration doesn't support reverse order,
-    /// but seeking the start key and iterating in reverse order works.
-    /// So we can create a workaround. We need to find the next available
-    /// element and use it as an anchor for reverse iteration,
-    /// but skip the first element to jump on the previous prefix.
-    /// If we can't find the next element, we are at the end of the list,
-    /// so we can use `IteratorMode::End` to start reverse iteration.
+    /// so we need to to force the RocksDB iterator to order
+    /// all the prefix in the iterator so that we can take the next prefix
+    /// as start of iterator and iterate in reverse.
     fn reverse_prefix_iter<T>(
         &self,
         prefix: &[u8],
@@ -450,28 +447,20 @@ where
     where
         T: ExtractItem,
     {
-        let maybe_next_item = next_prefix(prefix.to_vec())
-            .and_then(|next_prefix| {
-                self.iter_store(
-                    column,
-                    Some(next_prefix.as_slice()),
-                    None,
-                    IterDirection::Forward,
-                )
-                .next()
-            })
-            .and_then(|res| res.ok());
+        let reverse_iterator = next_prefix(prefix.to_vec())
+            .map(|next_prefix| {
+                    let mut opts = self.read_options();
+                    opts.set_total_order_seek(true);
+                    self.iterator::<T>(
+                        column,
+                        opts,
+                        IteratorMode::From(next_prefix.as_slice(), rocksdb::Direction::Reverse)
+                    )
+            });
 
-        if let Some((next_start_key, _)) = maybe_next_item {
-            let iter_mode = IteratorMode::From(
-                next_start_key.as_slice(),
-                rocksdb::Direction::Reverse,
-            );
+        if let Some(iterator) = reverse_iterator {
             let prefix = prefix.to_vec();
-            self
-                .iterator::<T>(column, self.read_options(), iter_mode)
-                // Skip the element under the `next_start_key` key.
-                .skip(1)
+            iterator
                 .take_while(move |item| {
                     if let Ok(item) = item {
                         T::starts_with(item, prefix.as_slice())


crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -450,28 +447,26 @@ where
...
+                column,
+                opts,
+                IteratorMode::From(next_prefix.as_slice(), rocksdb::Direction::Reverse),
+            )
+            .skip_while(move |item| {
+                if let Ok(item) = item {
+                    T::starts_with(item, next_prefix.as_slice())
+                } else {
+                    false
+                }
```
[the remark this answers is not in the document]
AurelienFT: No from what I tested it doesn't work like that as the rocksDB function use by our iterator is `SeekForPrev` we will not go after `2` and so if two doesn't exists it will take the key that is the nearest but lower than `2`. In your case it's 1. I will add tests and you will see it from tests.

AurelienFT:
@xgreenx I have added a test at rocksdb level and integration level.

crates/fuel-core/src/state/rocks_db.rs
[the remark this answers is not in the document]
MitchTurner: Can you create a followup issue for the optimization? And add a TODO in the code?

tests/tests/tx.rs
```diff
@@ -751,6 +751,56 @@ async fn get_transactions_by_owner_returns_correct_number_of_results(
...
+        cursor: None,
+        results: 10,
+        direction,
+    };
+
+    // When
+    let response = client
+        .transactions_by_owner(&bob, all_transactions_with_direction)
+        .await
+        .unwrap();
```
MitchTurner, line 789: We showed that this would fail with the old code, but didn't we do that with `alice`? Do we expect this to fail on `bob` with the old code?

AurelienFT: Yeah it failed in both cases.

crates/fuel-core/src/state/rocks_db.rs
```diff
@@ -1218,4 +1206,33 @@ mod tests {
         let _ = open_with_part_of_columns
             .expect("Should open the database with shorter number of columns");
     }
+
+    #[test]
+    fn iter_store__reverse_iterator() {
+        // Given
+        let (mut db, _tmp) = create_db();
+        let value = Arc::new(vec![1, 2, 3]);
```
netrome, line 1214: Nit: As we don't use `value` I'd suggest leaving it empty.

```suggestion
        let value = Arc::new(Vec::new());
```

netrome (dismissed):
Nice stuff. Thanks for spotting and fixing this!

crates/fuel-core/src/state/rocks_db.rs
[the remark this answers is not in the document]
AurelienFT: https://github.com/FuelLabs/fuel-core/issues/2405

crates/fuel-core/src/state/rocks_db.rs
AurelienFT: Done

netrome (approved):

xgreenx (approved):
It seems that final solution doesn't match the description of the PR=D

xgreenx (commented):
I think the current solution is more elegant. But it's still not clear to me why the old one was not working

xgreenx:
Ohh, I see. We used `iter_store` with the next prefix, not the key, and because of that, if it returns `None`, we will fall back into the `End` branch.

AurelienFT:
@xgreenx Yes the description had one sentence that was outdated, I removed. 
It was failing before, essentially because we wasn't using the read option `total_order_seek` that is needed if you want to iterate correctly from outside of a prefix befined in column creation. We could change to not use this option anymore with my futur proposition.

xgreenx:
All tests are passing even without `total_order_seek`, I think the problem were in the old logic=D

BTW, documentation says that it is better to use `auto_prefix_mode`
image

AurelienFT:
The `auto_prefix_mode` doesn't seems to exists in rust-rocksdb (https://github.com/rust-rocksdb/rust-rocksdb/issues/624).

You are right all the tests are passing but the test on the playground on the mainnet data isn't working without `total_order_seek`. I think the unit test it's because they are not using a column with prefix set and the integration test maybe don't have enough addresses defined.
Will try to change some tests to fail without `total_order_seek`

AurelienFT:
I don't find a way to make one of the integration or unit test fails :/

xgreenx:
> You are right all the tests are passing but the test on the playground on the mainnet data isn't working without total_order_seek

So, did I get right that If you use the current fix but without `total_order_seek`, the query from the issue will fail?

AurelienFT:
@xgreenx yes.

xgreenx (approved):

xgreenx (approved):

xgreenx (approved):

netrome (approved):

PR #2411, opened 2024-10-30
perf: Change `kv_store::Value` to be Arc<[u8]> instead of Arc<Vec<u8>>

### LINKED ISSUE #2379: perf: Remove redundant heap allocation in `kv_store::Value` type definition
### Context
The `Value` type is currently defined as
```
/// The value of the storage. It is wrapped into the `Arc` to provide less cloning of massive objects.
pub type Value = alloc::sync::Arc<Vec<u8>>;
```

Both the `Arc` and `Vec` are pointers to heap-allocated data, which means that we do two heap-lookups every time we read data from this type. We could skip the second heap allocation by replacing this type with `Arc<[u8]>`.

### Definition of done
The `kv_store::Value` type is defined as an `Arc<[u8]>`.

### References
Logan Smith has an interesting video on the topic which elaborates more on this https://www.youtube.com/watch?v=A4cKi7PTJSs

netrome:
## Linked Issues/PRs
Closes #2379 

## Description
This PR simplifies the `kv_store::Value` type to be `Arc<[u8]>` instead of `Arc<Vec<u8>>`. This makes the code slightly cleaner and also provides a performance benefit as we skip an unnecessary extra heap allocation.

## Checklist

### Before requesting review


rymnc:
perhaps we want to benchmark the difference in performance :)

netrome:
> perhaps we want to benchmark the difference in performance :)

Yep, running the `db_lookup_times` benches right now.

netrome:
> I can't see any conclusive difference between the implementations with the existing benchmarks. There's a high degree of variation between the runs and the benchmarks are mostly measuring retrieval times for large objects - so the extra heap allocation is negligible for these benchmarks. I'll experiment with a benchmarks where we do more reads of the returned values to see if I can measure a performance impact.

Though thinking more on this, since this represent a database value and will always be associated with a database operation - I don't see any scenario where we use this where the performance impact of the extra heap allocation isn't completely negligible in comparison to the database lookup.

So given that this change is very low prio I can't motivate taking the time to construct a benchmark exposing this difference, although it would have been fun to do so :sweat_smile:

xgreenx (dismissed):
I think we can merge it without benchmarks=)

netrome:
> I think we can merge it without benchmarks=)

Wonderful, let's get the multi-get merged first and follow up with this one afterwards https://github.com/FuelLabs/fuel-core/pull/2396

netrome:
Retargeted this PR to `master` since the multi-get PR is on hold for now.

xgreenx (approved):

rafal-ch (approved):
LGTM :+1:

xgreenx (approved):

Issue #2414, opened 2024-10-30
Allow DA recorded blocks to come in out-of-order


MitchTurner:
Kinda related to https://github.com/FuelLabs/fuel-core/issues/2397

We don't really care if the blocks come in out of order if we can track the state of each "unrecorded block" independently.

Issue #2397, opened 2024-10-25
Make da_recorded_height in V1 algorithm be an `Option` so it doesn't break on the first run


MitchTurner:
This is a simple fix to simplify the first run of the V1 algorithm to just accept whatever values it gets first.

PR #2415, opened 2024-10-30
Allow DA recorded blocks to come out-of-order

MitchTurner:
## Linked Issues/PRs
Closes: https://github.com/FuelLabs/fuel-core/issues/2414

Eliminates the need for: https://github.com/FuelLabs/fuel-core/issues/2397

## Description
It sounds like it is likely we will occasionally receive recorded block information from the committer out of order. There was no real reason to expect them in order so I've generalized the algorithm to accept whatever order.

## Checklist

### Before requesting review


crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -514,48 +507,35 @@ impl AlgorithmUpdaterV1 {
...
+            .saturating_add(recording_cost);
+        self.latest_known_total_da_cost_excess = new_da_block_cost;
+        self.latest_da_cost_per_byte = new_cost_per_byte;
+        Ok(())
     }
 
     fn drain_l2_block_bytes_for_range(
         &mut self,
-        height_range: Range<u32>,
+        heights: Vec<u32>,
```
rymnc, line 530: perhaps 
```suggestion
        heights: &[u32],
```
?

MitchTurner: Nice catch.

crates/fuel-gas-price-algorithm/gas-price-analysis/src/charts.rs
```diff
@@ -267,16 +267,14 @@ pub fn draw_profit(
     const ACTUAL_PROFIT_COLOR: RGBColor = BLACK;
     const PROJECTED_PROFIT_COLOR: RGBColor = RED;
     const PESSIMISTIC_BLOCK_COST_COLOR: RGBColor = BLUE;
-    let actual_profit_gwei: Vec<_> = actual_profit
-        .into_iter()
-        .map(|x| x / ONE_GWEI as i128)
-        .collect();
+    let actual_profit_gwei: Vec<_> =
+        actual_profit.iter().map(|x| x / ONE_GWEI as i128).collect();
```
netrome, line 271: While unrelated to this PR, can we use `i28::from` here instead of an `as` conversion to show that the conversion is infallible (which I hope it is 😅)

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -514,48 +507,32 @@ impl AlgorithmUpdaterV1 {
...
+        self.latest_known_total_da_cost_excess = new_da_block_cost;
+        self.latest_da_cost_per_byte = new_cost_per_byte;
+        Ok(())
     }
 
-    fn drain_l2_block_bytes_for_range(
-        &mut self,
-        height_range: Range<u32>,
-    ) -> Result<u128, Error> {
+    fn drain_l2_block_bytes_for_range(&mut self, heights: &[u32]) -> Result<u128, Error> {
```
netrome, line 528: Since we no longer take a range here we should rename this function. Perhaps just simplifying the name to
```suggestion
    fn drain_l2_block_bytes(&mut self, heights: &[u32]) -> Result<u128, Error> {
```

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -566,7 +543,7 @@ impl AlgorithmUpdaterV1 {
         let projection_portion: u128 = self
             .unrecorded_blocks
             .iter()
-            .map(|(_, &bytes)| (bytes as u128))
+            .map(|(_, &bytes)| bytes as u128)
```
netrome, line 546: Again unrelated but it would be nice to use and explicit infallible conversion: `u128::from`.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -307,11 +300,11 @@ impl core::ops::Deref for ClampedPercentage {
 impl AlgorithmUpdaterV1 {
     pub fn update_da_record_data(
         &mut self,
-        height_range: Range<u32>,
-        range_cost: u128,
+        heights: &[u32],
```
netrome, line 303: It seems like all we do is iterating over the heights, and we do provide a range still in `simulation.rs`. Perhaps we should accept an `impl IntoIterator<Item = u32>` here to allow both ranges and slices to be provided.

No need to do this if you think it's overkill.

crates/fuel-gas-price-algorithm/src/v1/tests/update_da_record_data_tests.rs
```diff
@@ -270,3 +228,132 @@ fn update_da_record_data__da_block_updates_projected_total_cost_with_known_and_g
...
+    let mut updater = UpdaterBuilder::new()
+        .with_da_cost_per_byte(da_cost_per_byte)
+        .with_unrecorded_blocks(unrecorded_blocks)
+        .with_da_cost_per_byte(old_da_cost_per_byte)
+        .with_known_total_cost(old_known_total_cost)
+        .with_projected_total_cost(old_projected_total_cost)
+        .build();
+    let new_cost_per_byte = 100;
+    let recorded_cost = 2 * (block_bytes * new_cost_per_byte) as u128;
+    let recorded_range: Vec<u32> = vec![2, 3];
```
netrome, line 309: Nit: Not sure if we should call this a range if it's a vec.

```suggestion
    let recorded_heights: Vec<u32> = vec![2, 3];
```

crates/fuel-gas-price-algorithm/src/v1/tests/update_da_record_data_tests.rs
```diff
@@ -270,3 +228,132 @@ fn update_da_record_data__da_block_updates_projected_total_cost_with_known_and_g
...
```
netrome, line 264: Same here
```suggestion
    let recorded_heights: Vec<u32> = vec![2, 3];
```

crates/fuel-gas-price-algorithm/src/v1/tests/update_da_record_data_tests.rs
```diff
@@ -270,3 +228,132 @@ fn update_da_record_data__da_block_updates_projected_total_cost_with_known_and_g
...
+
+    // when
+    updater
+        .update_da_record_data(&recorded_range, recorded_cost)
+        .unwrap();
+
+    // and
```
netrome, line 271:
```suggestion
    // then
```

crates/fuel-gas-price-algorithm/src/v1/tests/update_da_record_data_tests.rs
```diff
@@ -270,3 +228,132 @@ fn update_da_record_data__da_block_updates_projected_total_cost_with_known_and_g
     let expected = new_known_total_cost + guessed_part;
     assert_eq!(actual, expected as u128);
 }
+
+#[test]
+fn update_da_record_data__updates_known_total_cost_if_blocks_are_out_of_order() {
```
netrome, line 233: Maybe it's because it's late, but I don't see where we provide blocks out of order in this test case. Is it because we skip `1` in the `recorded_range`?

netrome: Could we be even more ezplicit and set `recorded_range` to `[3, 2]` instead?.

crates/fuel-gas-price-algorithm/src/v1/tests/update_da_record_data_tests.rs
```diff
@@ -270,3 +228,132 @@ fn update_da_record_data__da_block_updates_projected_total_cost_with_known_and_g
...
+    let old_da_cost_per_byte = 20;
+    let mut updater = UpdaterBuilder::new()
+        .with_da_cost_per_byte(da_cost_per_byte)
+        .with_unrecorded_blocks(unrecorded_blocks)
+        .with_da_cost_per_byte(old_da_cost_per_byte)
+        .with_known_total_cost(old_known_total_cost)
+        .with_projected_total_cost(old_projected_total_cost)
+        .build();
+    let new_cost_per_byte = 100;
+    let recorded_cost = 2 * (block_bytes * new_cost_per_byte) as u128;
```
netrome, line 308: The setup here seem identical between this and the previous test. Perhaps we should consider breaking it out to a separate helper function?

netrome (approved):
Looks good overall to me. A few suggestions, nits and questions.

crates/fuel-gas-price-algorithm/gas-price-analysis/src/charts.rs
MitchTurner: I've made the change. I think in general the analysis code is a little messier, but now that we're considering keeping it around and maybe expanding it's role it's probably best we get more strict with it.

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: I think it's overkill. What we'll get from the committer will be a `Vec`. The ranges are just fossils from the old code.

crates/fuel-gas-price-algorithm/src/v1/tests/update_da_record_data_tests.rs
MitchTurner: That's a good idea. I've switched it. In reality, they'll probably be in order inside the vector, but they don't need to be. The point of this PR is that some heights can be skipped and returned to later. These tests skip `1` and do `2` and `3`.

crates/fuel-gas-price-algorithm/src/v1/tests/update_da_record_data_tests.rs
MitchTurner: Yes. I literally had a single test before with 3 asserts, but I broke it out. I have mixed feelings about helpers since they prevent me from accessing whichever values I need from setup.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -514,48 +507,32 @@ impl AlgorithmUpdaterV1 {
...
-            )?;
-            self.da_recorded_block_height = last;
-            let new_da_block_cost = self
-                .latest_known_total_da_cost_excess
-                .saturating_add(range_cost);
-            self.latest_known_total_da_cost_excess = new_da_block_cost;
-            self.latest_da_cost_per_byte = new_cost_per_byte;
-            Ok(())
-        }
+        let recorded_bytes = self.drain_l2_block_bytes_for_range(heights)?;
```
rafal-ch: Maybe just:

```suggestion
        let recorded_bytes = self.drain_l2_block_bytes(heights)?;
```

It'll save us from renaming the function when we change the datatype again. However, the intention here could be to not relate to the `Range<T>`, but to the "range" as arbitrary "noun".

Either way, just a small detail.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -16,14 +13,12 @@ mod tests;
...
     SkippedL2Block { expected: u32, got: u32 },
-    #[error("Skipped DA block update: expected {expected:?}, got {got:?}")]
-    SkippedDABlock { expected: u32, got: u32 },
     #[error("Could not calculate cost per byte: {bytes:?} bytes, {cost:?} cost")]
     CouldNotCalculateCostPerByte { bytes: u128, cost: u128 },
     #[error("Failed to include L2 block data: {0}")]
     FailedTooIncludeL2BlockData(String),
-    #[error("L2 block expected but not found in unrecorded blocks: {0}")]
-    L2BlockExpectedNotFound(u32),
+    #[error("L2 block expected but not found in unrecorded blocks: {height:?}")]
```
rafal-ch, line 20: Unrelated, but this could be just `{height}`, not `{height:?}`. But also I see that we use debug for other messages, so maybe there's a reason.

rafal-ch (approved):
LGTM :+1: 
with just a few nits and minor comments

crates/fuel-gas-price-algorithm/gas-price-analysis/src/charts.rs
netrome: Ah right, didn't notice this was in the analysis code :sweat_smile: thanks anyway!

crates/fuel-gas-price-algorithm/src/v1.rs
netrome: Fair enough :+1:

crates/fuel-gas-price-algorithm/src/v1/tests/update_da_record_data_tests.rs
netrome: > I have mixed feelings about helpers since they prevent me from accessing whichever values I need from setup.

Yeah I can live with a little duplication in test code for this reason. It's easy to make assumptions on shared logic that has to be broken in subsequent tests, so there's no obvious right or wrong here. I'll leave this to your judgement as well.

netrome (approved):
Thanks for updating! :pray:

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -566,7 +546,7 @@ impl AlgorithmUpdaterV1 {
         let projection_portion: u128 = self
             .unrecorded_blocks
             .iter()
-            .map(|(_, &bytes)| (bytes as u128))
+            .map(|(_, &bytes)| u128::from(bytes))
             .fold(0_u128, |acc, n| acc.saturating_add(n))
```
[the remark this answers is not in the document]
MitchTurner: This gets called semi-infrequently (only when we get a batch back from the committer). It could become more frequent I guess.

Yes. An alternative would be having the "total unrecorded bytes" tracked and when we remove blocks we subtract their bytes from that total. Then this would just be a multiplication of that. I can add that if you prefer.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -514,48 +507,35 @@ impl AlgorithmUpdaterV1 {
...
-                .ok_or(Error::L2BlockExpectedNotFound(expected_height))?;
-            if actual_height != expected_height {
-                return Err(Error::L2BlockExpectedNotFound(expected_height));
-            }
+        for expected_height in heights {
+            let bytes = self.unrecorded_blocks.remove(expected_height).ok_or(
+                Error::L2BlockExpectedNotFound {
+                    height: *expected_height,
+                },
+            )?;
```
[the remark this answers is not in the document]
MitchTurner: It's up to us. If this errors we have a serious problem with our code.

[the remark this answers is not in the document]
xgreenx: If the committer has problems and submits several bundles per same height(they can in theory because they do re-bundle),  it will break the node.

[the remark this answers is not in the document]
MitchTurner: That can't happen if it's finalized before they report to us.

crates/fuel-gas-price-algorithm/src/v1.rs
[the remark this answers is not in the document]
MitchTurner: Done.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -355,6 +350,9 @@ impl AlgorithmUpdaterV1 {
 
             // metadata
             self.unrecorded_blocks.insert(height, block_bytes);
+            self.unrecorded_blocks_bytes = self
+                .unrecorded_blocks_bytes
+                .saturating_add(block_bytes as u128);
```
acerone85, line 355: this leaves me with a question. Can we add a block twice, e.g. in the case of a reorganization? If so it might be necessary to subtract the `unrecorded_block_bytes` for the block that is already present at the same height.

acerone85:
had a look, only one question from my side. otherwise LGTM

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: If the reorg happens on L2, then we will roll back the state of the algorithm too. We regularly store the `GasPriceMetadata` to our database so we can return to previous state. That only happens when we _restart_ the client, it populates the algorithm values with the metadata from that height.

acerone85: Thanks, really clear explanation :D

acerone85 (dismissed):

crates/fuel-gas-price-algorithm/src/v1.rs
rymnc: if we just get finalized data we dont need to worry about reorgs too much

rafal-ch (dismissed):

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -538,16 +543,13 @@ impl AlgorithmUpdaterV1 {
             )?;
             total = total.saturating_add(bytes as u128);
         }
+        self.unrecorded_blocks_bytes = self.unrecorded_blocks_bytes.saturating_sub(total);
```
rafal-ch, line 547: I'm thinking about the following scenario, not sure if this is an issue or not:
1. we have 3 blocks in `self.unrecorded_blocks`
2. we `remove()` 2 of them, but we get an error on 3rd one
3. now, the `unrecorded_blocks_bytes` becomes out-of-sync

crates/fuel-gas-price-algorithm/src/v1.rs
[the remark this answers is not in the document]
MitchTurner: I've been thinking about the (very unlikely) case that our gas price metadata is "ahead" of the chain on restart. This could in theory result in a bunch of the blocks being missing from the `unrecorded_blocks`. At that point we might have the da_committer reporting those blocks again. 

Since the precision of the algorithm isn't that important, I think the simplest solution would be to just ignore re-reported blocks.

In that case we don't want it throwing an error here. So I'm going to change this.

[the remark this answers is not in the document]
MitchTurner: In terms of the algorithm, I think the best approach here is to ignore the batch. The problem is if it ever occurs, we will carry those `unrecorded_blocks` forever. It's fine from the algorithm's perspective because the error will become negligible pretty quickly.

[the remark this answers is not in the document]
MitchTurner: Nevermind. We can both remove the blocks and ignore. I think that's the best.

rymnc (approved):

netrome (dismissed):

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -515,60 +526,69 @@ impl AlgorithmUpdaterV1 {
...
+            // minimum of the two costs
+            Incomplete(bytes) => {
+                let new_guess_recording_cost = bytes
+                    .saturating_mul(self.latest_da_cost_per_byte)
+                    .min(recording_cost);
+                let new_da_block_cost = self
+                    .latest_known_total_da_cost_excess
+                    .saturating_add(new_guess_recording_cost);
+                self.latest_known_total_da_cost_excess = new_da_block_cost;
+            }
```
xgreenx, line 559: It is not clear to me why missed bundles don't allow us to calculate the cost per byte. We know the total number of bytes and the cost of posting them. Why we can't use these values? What are we afraid of?

crates/services/gas_price_service/src/v1/metadata.rs
```diff
@@ -3,7 +3,10 @@ use fuel_gas_price_algorithm::v1::{
...
     L2ActivityTracker,
 };
-use std::num::NonZeroU64;
+use std::{
+    collections::BTreeMap,
+    num::NonZeroU64,
+};
 
 #[derive(serde::Serialize, serde::Deserialize, Debug, Clone, PartialEq)]
 pub struct V1Metadata {
```
xgreenx, line 12: Do we have a task to remove `unrecorded_blocks` field from the `V1Metadata`?(We should use separate table to store them). Taking into account that committer posts bundles each hour, it can be a lot of storage space(each `V1Metadata` will have 3600 entries, and we will insert it into the storage for each block height).

crates/services/gas_price_service/src/v1/metadata.rs
```diff
@@ -148,20 +145,25 @@ pub fn v1_algorithm_from_metadata(
...
     );
-    let unrecorded_blocks = metadata
+    let unrecorded_blocks_bytes: u128 = metadata
         .unrecorded_blocks
-        .into_iter()
-        .chain(config.unrecorded_blocks.clone())
-        .collect();
+        .iter()
+        .map(|(_, size)| u128::from(*size))
+        .sum();
```
xgreenx, line 152: Why `unrecorded_blocks_bytes` can't be a field of the metadata, instead of calculating it each time?

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -515,60 +526,69 @@ impl AlgorithmUpdaterV1 {
...
+            let maybe_bytes = self.unrecorded_blocks.remove(expected_height);
+            if let Some(bytes) = maybe_bytes {
+                total = total.saturating_add(bytes as u128);
+            } else {
+                tracing::error!(
+                    "L2 block expected but not found in unrecorded blocks: {}",
+                    expected_height,
+                );
+                missed_some_heights = true;
             }
```
xgreenx, line 579: If block is not part of the `self.unrecorded_blocks`, it means we already posted it to the DA in the past.

We could either just ignore it or only increase the known DA cost on the amount of fee that we've paid to return this amount.

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: > missed bundles

No. This is about missed blocks.

> We know the total number of bytes

Not really. We have the total bytes after compression. 

We could in theory go through each block in the bundle and query our api to find the original bytes for each block that's not listed in `unrecorded_blocks`, but this is an edge case that we probably will never hit and I think this solution works well enough since gas price is off-chain and a little error is fine.

crates/services/gas_price_service/src/v1/metadata.rs
MitchTurner: That's a good idea. I'll add a story. It doesn't need to be removed from `V1Metadata`, it just needs to be handled in the adapter.

MitchTurner: https://github.com/FuelLabs/fuel-core/issues/2454

crates/services/gas_price_service/src/v1/metadata.rs
MitchTurner: It only gets calculated on startup. I figured that would be better than adding another field to every metadata that is stored.

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: > or only increase the known DA cost on the amount of fee that we've paid to return this amount.

I don't know what this means. I'm not sure what `fee` you're talking about

Ignoring makes the most sense to me but maybe I'm missing something.

crates/services/gas_price_service/src/v1/metadata.rs
xgreenx: I prefer the usage of the `unrecorded_blocks_bytes` field over fetching all unrecorded blocks on startup-up from the storage. Iterating over all elements of the storage only works on the `latest_view`. If you want to initiate the gas price service at a specific block height via `view_at,` then fetching will not work.

We don't have use cases right now where we need to support `view_at`, so it is not critical. But it would be nice if we had this option.

crates/services/gas_price_service/src/v1/metadata.rs
xgreenx: image

In the worst case, it is 481 MB per year. But since we use postcard, it will be much lower. Plus, we don't need to store data for the whole year. So I wouldn't concern about the storage=)

crates/fuel-gas-price-algorithm/src/v1.rs
xgreenx: If the block committer submits two bundles(or more) for the same block height, it means that we pay a fee to DA twice(or more). Maybe we want to include the cost of posting the second bundle into the known DA cost.

crates/fuel-gas-price-algorithm/src/v1.rs
xgreenx: Question to you: What is `new_cost_per_byte`? Is it the cost per 1 compressed byte submitted by the block committer to DA? Or is it the cost per 1 uncompressed byte for the Fuel block with the compression ratio included?

I would expect the first one to be used. But it seems you use the second in the code.

An example:

Block X
The Fuel Block is - 256KB.
The compressed Fuel Block is - 64KB.
Let's assume the bundle has 1 Fuel block(the block above).
The cost of posting this bundle is 1ETH.

I will expect that `new_cost_per_byte` is `1ETH / 64KB` because it is the actual cost to post bytes to the DA.
The current implementation will use `1ETH / 256KB`. So the `new_cost_per_byte` will be lower - cheaper.

Let's see what the cost will be in the next block, assuming we updated `new_cost_per_byte` by the rules above.

Block X + 1
The Fuel Block is - 256KB.
The projected cost with the current implementation is `256KB * 1 ETH / 256KB = 1 ETH`.
The projected cost I would expect - `256 KB * 1ETH / 64KB = 4 ETH`.

The compressed Fuel Block is - 256KB(all data is new, so compression is bad).
The current implementation will have `3 ETH` debt, assuming the same cost on DA. But if it increases, the debt will be more.

MitchTurner: `new_cost_per_byte` is the cost per uncompressed bytes.

Absolutely, the compression ratio will be different from block-to-block. We can't get the fully compressed and zipped block, as that doesn't get produced until we actually commit (at which point projecting the cost isn't needed any more).

We could split the difference and get the pre-zipped, compressed block in theory. But I feel like for now, since the blobs include so many blocks, it averages out.

We could look into adding this improvement after we get DA gas costs running and before things get too congested.

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: Ah ah. Yeah. That is possible in theory. But probably only when finalization fails on L1, right? In that case we don't pay the first fee, but only the second.

This is another "improvement" I've discussed. Where we track "committed but not finalized" blocks, not just `unrecorded_blocks`. I've considered this something we'd like to do after we get the DA gas cost live.

crates/services/gas_price_service/src/v1/metadata.rs
MitchTurner: Is this more about the next comment:

> Why unrecorded_blocks_bytes can't be a field of the metadata, instead of calculating it each time?

We still need the bytes for each unrecorded block, even if we store the total as well.

crates/fuel-gas-price-algorithm/src/v1.rs
xgreenx: I think it would be simpler if `new_cost_per_byte` were just the last DA cost per 1 byte(regardless of whether it is compressed or uncompressed). In this case, in the example above, we will charge the user `4 ETH` based on the uncompressed data, assuming the worst-case without compression.

crates/services/gas_price_service/src/v1/metadata.rs
xgreenx: > We still need the bytes for each unrecorded block, even if we store the total as well.

Yes, we need it, but we can fetch this information during iteration over `heights` requesting `maybe_bytes` from the table in the database.

image

crates/fuel-gas-price-algorithm/src/v1.rs
xgreenx: It is not additional "improvement", it is the case that you handle by this code=) Basically, it is only possible in the case described above. If `unrecorded_blocks` doesn't have some block, it is only possible if you already processed that block height in the past(so you had two or more bundles that covered this block height).

Based on your comments we can just do nothing for now, and cover this case later=)

crates/services/gas_price_service/src/v1/metadata.rs
MitchTurner: We can address it in https://github.com/FuelLabs/fuel-core/issues/2454

xgreenx (approved):

rymnc (approved):

Issue #2140, opened 2024-08-28
Upgrade fuel core to use V1 gas price algo/updater


MitchTurner:
This includes adding the config values once it is fully integrated.

It's not clear to me how we want to smoothly transition the values. We should probably start taking the DA costs into calculation, but configure it to always have DA costs as 0 until we get some good tests up.

MitchTurner:
We have some versioning going on with the algorithm:
```

#[derive(Debug, Clone, PartialEq)]
pub enum Algorithm {
    V0(AlgorithmV0),
    V1(AlgorithmV1),
}
impl GasPriceAlgorithm for Algorithm {
    fn next_gas_price(&self) -> u64 {
        match self {
            Algorithm::V0(v0) => v0.calculate(),
            Algorithm::V1(v1) => v1.calculate(0),
        }
    }
    fn worst_case_gas_price(&self, height: BlockHeight) -> u64 {
        match self {
            Algorithm::V0(v0) => v0.worst_case(height.into()),
            Algorithm::V1(v1) => v1.calculate(0),
        }
    }
}
```
I don't think we need that. I think we can just depend on a single algorithm and change it as we improve it. Shouldn't be a problem.

rymnc:
assigning to myself

MitchTurner:
Completed with https://github.com/FuelLabs/fuel-core/pull/2469

PR #2416, opened 2024-10-31
feat(gas_price_service_v1): define RunnableTask for GasPriceServiceV1

rymnc:
> [!NOTE]
> Some values for the tests need expert opinion from @MitchTurner. A follow up PR will be created to define the `UninitializedTask` that wraps over this task.

## Linked Issues/PRs

part of https://github.com/FuelLabs/fuel-core/issues/2140, but doesn't close it yet.

## Description

- We define a new `RunnableTask`, `GasPriceServiceV1`, which uses the da block cost source in tandem with the l2 block source
- Tests for the same
- Casts to and from the v1 algorithm updater
- we take a direct dependency on v0's metadata so that we can migrate between the two.

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/fuel-gas-price-algorithm/src/v1.rs
+++ b/crates/fuel-gas-price-algorithm/src/v1.rs
@@ -313,6 +313,7 @@ impl AlgorithmUpdaterV1 {
         if !height_range.is_empty() {
             self.da_block_update(height_range, range_cost)?;
             self.recalculate_projected_cost();
+            self.update_da_gas_price();
         }
         Ok(())
     }
--- a/crates/fuel-gas-price-algorithm/src/v1/tests/update_da_record_data_tests.rs
+++ b/crates/fuel-gas-price-algorithm/src/v1/tests/update_da_record_data_tests.rs
@@ -272,7 +272,7 @@ fn update_da_record_data__da_block_updates_projected_total_cost_with_known_and_g
 }
 
 #[test]
-fn update_da_record_data__da_block_updates__da_gas_price() {
+fn update_da_record_data__da_block_lowers_da_gas_price() {
     // given
     let da_cost_per_byte = 40;
     let da_recorded_block_height = 10;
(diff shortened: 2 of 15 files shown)


crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -0,0 +1,473 @@
...
+        l2_block_sender.send(l2_block).await.unwrap();
+        service.shutdown().await.unwrap();
+
+        // then
+        let actual_price = read_algo.next_gas_price();
+        assert_ne!(initial_price, actual_price);
+    }
+
+    #[tokio::test]
+    async fn run__updates_gas_price_with_da_block_cost_source() {
```
rymnc, line 394: cc: @MitchTurner this test seems to be failing with the following error - 
```
called `Result::unwrap()` on an `Err` value: L2 block expected but not found in unrecorded blocks: 1
```

which i narrowed down to https://github.com/FuelLabs/fuel-core/blob/a14922eb17a9ce844132edc240b5157eb860199f/crates/fuel-gas-price-algorithm/src/v1.rs#L546-L562.

This is what happens when the da block cost that is received refers to an L2 block that hasn't been processed by the algorithm yet. that is fine, seems like defined behaviour.


Below, on line 432-434, I define the cost and bytes. Something seems to be wrong in the `drain_l2_block_bytes_for_range` function https://github.com/FuelLabs/fuel-core/blob/a14922eb17a9ce844132edc240b5157eb860199f/crates/fuel-gas-price-algorithm/src/v1.rs#L546 when I call it with just one item in the range. because the `range_bytes` it yields after being called in https://github.com/FuelLabs/fuel-core/blob/a14922eb17a9ce844132edc240b5157eb860199f/crates/fuel-gas-price-algorithm/src/v1.rs#L529 is `0`, which doesn't seem too correct.

rymnc: wondering if you have any thoughts about the params I'm using.

MitchTurner: > wondering if you have any thoughts about the params I'm using.

With these tests, the values don't exactly matter. We use an `assert_ne!` and as long as that's true we're not concerned with the details. The problem with this approach is the test writer/reader still needs to know what values would result in a different gas price. This is part of the reason I originally had an abstraction of gas price updater (which we've (you've) removed)

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -0,0 +1,507 @@
...
+        // here we mimic that behaviour by running the da block cost service.
+        service
+            .da_source_adapter_handle
+            .run(&mut watcher)
+            .await
+            .unwrap();
+
+        // when
+        service.run(&mut watcher).await.unwrap();
+        l2_block_sender.send(l2_block).await.unwrap();
```
MitchTurner, line 500: Is the bug you're seeing the reason we aren't submitting a new DA block here?

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -0,0 +1,507 @@
...
+where
+    Metadata: MetadataStorage,
+    DA: DaBlockCostsSource,
+{
+    pub fn new(
+        l2_block_source: L2,
+        metadata_storage: Metadata,
+        shared_algo: SharedV1Algorithm,
+        algorithm_updater: AlgorithmUpdaterV1,
+        da_source_adapter_handle: DaSourceService<DA>,
```
MitchTurner, line 74: I feel like we should just be holding onto an abstraction of the da block source here, not the concrete `DaSourceService`. That should make tests easier too.

crates/services/gas_price_service/src/v1/service.rs
rymnc: not seeing the bug anymore :)

crates/services/gas_price_service/src/v1/service.rs
rymnc: I think @xgreenx suggested we shouldn't do that because we should tightly couple it and not have more traits. I had it initially implemented like that in a previous PR.

crates/services/gas_price_service/src/v1/service.rs
MitchTurner: I think this test should be "sending" a block to the DA record service, not the `l2_block_sender`.

crates/services/gas_price_service/src/v1/service.rs
MitchTurner: Hmm. Okay. I might have even been part of that convo. It's not the end of the world.

crates/services/gas_price_service/src/v1/service.rs
rymnc: I still need to trigger `update_da_gas_price()`, so I've added it in the `update_da_record_data` fn of the `AlgorithmV1` impl in c516ea9b95

crates/services/gas_price_service/src/v1/service.rs
rymnc: don't understand the action here, do you want me to add the trait or?

MitchTurner: No. I'm saying we can leave it for now.

crates/services/gas_price_service/src/v1/metadata.rs
```diff
@@ -68,19 +67,56 @@ impl V1Metadata {
...
+    pub gas_price_factor: NonZeroU64,
+    pub min_da_gas_price: u64,
+    pub max_da_gas_price_change_percent: u16,
+    pub da_p_component: i64,
+    pub da_d_component: i64,
+    pub normal_range_size: u16,
+    pub capped_range_size: u16,
+    pub decrease_range_size: u16,
+    pub block_activity_threshold: u8,
+    pub unrecorded_blocks: Vec<(u32, u64)>,
```
MitchTurner, line 83: Hmm. I guess it doesn't hurt. I don't know why we'd need this if it wasn't already in the metadata.

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -0,0 +1,510 @@
...
+                crate::common::utils::Error::CouldNotInitUpdater(anyhow::anyhow!(
+                    "Could not convert metadata to V0Metadata"
+                ))
+            })?;
+            let v1_metadata = V1Metadata::construct_from_v0_metadata(v0_metadata, config)
+                .map_err(|err| {
+                    crate::common::utils::Error::CouldNotInitUpdater(anyhow::anyhow!(err))
+                })?;
+            algorithm_updater = v1_algorithm_from_metadata(v1_metadata, config);
+        }
```
MitchTurner, line 273: nit: I think we should just unify this branch and add a method `fn get_v1_metadata(metadata: &Metadata, config: &V1AlgorithmConfig)`. In both branches we end up calling `v1_algorithm_from_metadata`, so something like:
```suggestion
        let v1_metadata = get_v1_metadata(updater_metadata, config)?;
        algorithm_updater = v1_algorithm_from_metadata(v1_metadata, config);
```

crates/services/gas_price_service/src/v1/service.rs
rymnc: sounds good

crates/services/gas_price_service/src/v1/da_source_service/service.rs
```diff
@@ -103,12 +104,14 @@ where
     async fn run(&mut self, state_watcher: &mut StateWatcher) -> Result<bool> {
         let continue_running;
 
+        let tick: BoxFuture<tokio::time::Instant> = Box::pin(self.poll_interval.tick());
```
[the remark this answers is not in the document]
rymnc: you're right, I had it that way when i was testing something. removed in 53291a5c02

crates/services/gas_price_service/src/v1/metadata.rs
rymnc: need this to bootstrap the `AlgorithmV1` with a state that we can test only DA data being sent :)

crates/services/gas_price_service/src/v1/service.rs
rymnc: addressed in 2291207dd3

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -313,6 +313,7 @@ impl AlgorithmUpdaterV1 {
         if !height_range.is_empty() {
             self.da_block_update(height_range, range_cost)?;
             self.recalculate_projected_cost();
+            self.update_da_gas_price();
```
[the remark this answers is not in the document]
rymnc: addressed in 8ae983f95a

crates/services/gas_price_service/src/v1/metadata.rs
MitchTurner: Hmm. K. Technically you could provide a metadata with it, but that's expensive.

crates/fuel-gas-price-algorithm/src/v1/tests/update_da_record_data_tests.rs
```diff
@@ -270,3 +270,56 @@ fn update_da_record_data__da_block_updates_projected_total_cost_with_known_and_g
...
+    let old_da_gas_price = updater.new_scaled_da_gas_price;
+
+    // when
+    updater
+        .update_da_record_data(recorded_range, recorded_cost as u128)
+        .unwrap();
+
+    // then
+    let new_da_gas_price = updater.new_scaled_da_gas_price;
+    assert_ne!(old_da_gas_price, new_da_gas_price);
```
MitchTurner, line 324: We should be more specific at this level (and not just say it is different) You can copy most of the code from
`update_l2_block_data__below_threshold_will_decrease_exec_gas_price`, `update_l2_block_data__above_threshold_will_increase_exec_gas_price`, and `update_l2_block_data__even_threshold_will_not_change_exec_gas_price`

It feels redundant, but they are different endpoints so they change independently.

AurelienFT (approved):
The code looks clean and understandable from an external POV.

MitchTurner (approved):

MitchTurner (approved):

AurelienFT (approved):

AurelienFT (approved):

MitchTurner (approved):

PR #2419, opened 2024-10-31
Proposals to multi get PR

xgreenx:
Proposals to the https://github.com/FuelLabs/fuel-core/pull/2396


crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -395,6 +398,21 @@ where
         self.db.get(key, Column::OriginalColumn(column))
     }
 
+    fn get_batch<'a>(
```
netrome, line 401: Thanks for adding this implementation as well.

netrome (approved):
Thanks for the proposal. It's a great improvement to not have to box the iterators at all levels.

PR #2429, opened 2024-11-12
Replace task `run()` return result with custom enum

MitchTurner:
## Linked Issues/PRs
closes: https://github.com/FuelLabs/fuel-core/issues/2441

## Description
Before any `RunnableTask` would return a `Result<bool>` where, implicitly, the `bool` was telling the service runner if it should re-run the task.

This is confusing from the customer perspective since you need to understand the behavior of `ServiceRunner` in order to choose the right response: i.e. `Ok(true)` meant you want to rerun. Instead this PR adds the `Continue` and `Stop` variants.

It was also confusing with `Error` since, unintuitively, an `Error` would result in the task being run again. This PR adds the `ErrorContinue` variant, so the implementer know explicitly what the behavior is.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -137,8 +138,7 @@ impl RunnableService for SyncTask {
 
 #[async_trait::async_trait]
 impl RunnableTask for SyncTask {
-    #[tracing::instrument(level = "debug", skip_all, err, ret)]
```
MitchTurner, line 140: Had to remove this from a couple locations since the instrument macro assumes the function returns Result

crates/fuel-core/src/graphql_api/api_service.rs
```diff
@@ -196,11 +197,15 @@ impl RunnableService for GraphqlService {
 
 #[async_trait::async_trait]
 impl RunnableTask for Task {
-    async fn run(&mut self, _: &mut StateWatcher) -> anyhow::Result<bool> {
-        self.server.as_mut().await?;
```
MitchTurner, line 200: This change introduces a non-trivial amount of noise with the removal of the `?` operators.

AFAICT, there isn't a way to implement `Try` for arbitrary types in Rust which is a real bummer.

crates/services/p2p/src/service.rs
```diff
@@ -922,7 +923,12 @@ where
                     }
                     Some(TaskRequest::RespondWithGossipsubMessageReport((message, acceptance))) => {
                         // report_message(&mut self.p2p_service, message, acceptance);
-                        self.p2p_service.report_message(message, acceptance)?;
+                        match self.p2p_service.report_message(message, acceptance) {
+                            Ok(_) => {},
+                            Err(e) => {
+                                return e.into()
+                            }
+                        };
```
MitchTurner, line 931:
```suggestion
                        match self.p2p_service.report_message(message, acceptance) {
                            Ok(_) => {},
                            Err(err) => {
                                return err.into()
                            }
                        };
```

crates/fuel-core/src/service/adapters/consensus_parameters_provider.rs
```diff
@@ -135,16 +136,15 @@ impl RunnableTask for Task {
...
+                            return TaskRunResult::Stop
                         }
                     }
                 }
                 should_continue = true;
             }
         }
 
-        Ok(should_continue)
+        TaskRunResult::should_continue(should_continue)
```
xgreenx, line 147: I though the idea was to remove `should_continue` variable and replace with the usage of the enum=D

crates/services/gas_price_service/src/v1/da_source_service/service.rs
```diff
@@ -100,21 +101,23 @@ where
...
             biased;
             _ = state_watcher.while_started() => {
-                continue_running = false;
+                TaskRunResult::Stop
             }
             _ = self.poll_interval.tick() => {
-                let da_block_costs = self.source.request_da_block_cost().await?;
-                self.shared_state.0.send(da_block_costs)?;
-                continue_running = true;
+                match self.source.request_da_block_cost().await.and_then(|da_block_costs| self.shared_state.0.send(da_block_costs).map_err(|err| anyhow::anyhow!(err))) {
```
xgreenx, line 111: It is very hard to read, maybe we could it inside of the function and process the result of the function?=)

crates/services/gas_price_service/src/v0/service.rs
```diff
@@ -121,24 +122,32 @@ where
...
             }
             l2_block_res = self.l2_block_source.get_l2_block() => {
                 tracing::info!("Received L2 block result: {:?}", l2_block_res);
-                let block = l2_block_res?;
+                let block = match l2_block_res {
+                    Ok(block) => block,
+                    Err(err) => {
+                        return anyhow!(err).into()
+                    }
+                };
```
xgreenx, line 139: I see this pattern in many places. Maybe it would be simpler to defined a helper trait with `into_continue_task_result` which either returns `TaskRunResult::Continue` or `TaskRunResult::ErrorContinue`

crates/fuel-core/src/service/adapters/consensus_parameters_provider.rs
MitchTurner: My goal was to change the interface. 

I'm happy to do some more refactoring too.

crates/services/src/service.rs
```diff
@@ -83,6 +83,47 @@ pub trait RunnableService: Send {
...
 }
 
+/// The result of a single iteration of the service task
+pub enum TaskRunResult {
+    /// Request the task to be run again
+    Continue,
+    /// Request the task to be abandoned
+    Stop,
+    /// Request the task to be run again, but report an error
+    ErrorContinue(anyhow::Error),
```
acerone85, line 93: do you envision to have an `ErrorStop` variant in the future?

MitchTurner: We don't have that use case currently. Would make sense wouldn't it? But YAGNI for now.

acerone85: was just curious :) 
Yeah, I don't think it makes sense to add it now if you don't have a use for it

crates/services/src/service.rs
```diff
@@ -361,14 +402,14 @@ async fn run_task<S: RunnableTask>(
...
                 tracing::debug!("run loop");
             }
-            Err(e) => {
+            TaskRunResult::Stop => {
+                tracing::debug!("stopping");
+                break;
+            }
+            TaskRunResult::ErrorContinue(e) => {
                 let e: &dyn std::error::Error = &*e;
                 tracing::error!(e);
```
acerone85, line 414: would it make sense to add a debug! statement to explicitly tell that the loop will continue (similar to the `Continue` case)?

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -194,32 +232,22 @@ where
...
             }
             da_block_costs = self.da_source_channel.recv() => {
-                tracing::info!("Received DA block costs: {:?}", da_block_costs);
-                let da_block_costs = da_block_costs?;
-
-                tracing::debug!("Updating DA block costs");
-                self.apply_da_block_costs_to_gas_algorithm(da_block_costs).await?;
-                should_continue = true;
+                let res = self.process_da_block_costs_res(da_block_costs).await;
+                TaskRunResult::Continue
```
xgreenx, line 248: If we have error here, you will not propagate it to the caller of the `run` function and we will miss log for this error.

crates/services/p2p/src/service.rs
```diff
@@ -916,7 +915,12 @@ where
                     }
                     Some(TaskRequest::RespondWithGossipsubMessageReport((message, acceptance))) => {
                         // report_message(&mut self.p2p_service, message, acceptance);
-                        self.p2p_service.report_message(message, acceptance)?;
+                        match self.p2p_service.report_message(message, acceptance) {
```
xgreenx, line 918: [nit] `if let Err(err) = res` fits better here.

crates/services/src/service.rs
```diff
@@ -83,6 +83,47 @@ pub trait RunnableService: Send {
...
+    Continue,
+    /// Request the task to be abandoned
+    Stop,
+    /// Request the task to be run again, but report an error
+    ErrorContinue(anyhow::Error),
+}
+
+impl TaskRunResult {
+    /// Creates a `TaskRunResult` from a `Result` where `Ok` means `Continue` and any error is reported
+    pub fn continue_if_ok<T, E: Into<anyhow::Error>>(res: Result<T, E>) -> TaskRunResult {
```
xgreenx, line 98: `continue_if_ok` is a bad name. I thought it returns `Continue` only If it is `Ok`.

What do you think about `always_continue`?

MitchTurner: > I thought it returns Continue only If it is Ok.

Yeah. It returns `continue_` only `if_` it is `ok`. I doesn't `always_` return `continue`... I'm confused why you want me to change it.

crates/services/gas_price_service/src/v1/service.rs
MitchTurner: Thank you for catching this.

crates/services/src/service.rs
MitchTurner: I'm not sure I understand. Are you talking in the `ErrorContinue` case? I feel like we should definitely use `error!` there, especially since we don't exit on errors.

crates/services/src/service.rs
xgreenx: Maybe just `continue` then?=) For me, the main point here is that we should still continue the task. What happens with error is not important for me as a reader of the code.

MitchTurner: Yeah. I think you're right. I think that's what the caller is interested in. I'll go with one of those.

MitchTurner: I feel like TaskRunResult is the wrong name then. It's almost a state machine but not quite. But TaskState or TaskStateTransition might be a better name...

xgreenx (dismissed):

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -567,29 +577,27 @@ where
...
             }
             _ = next_block_production => {
                 match self.on_timer().await.context("While processing timer event") {
-                    Ok(()) => should_continue = true,
+                    Ok(()) => TaskNextAction::Continue,
                     Err(err) => {
                         // Wait some time in case of error to avoid spamming retry block production
                         tokio::time::sleep(Duration::from_secs(1)).await;
-                        return Err(err);
+                        err.into()
```
rafal-ch, line 592: Another nit for clarity - maybe we shouldn't hide what happens here in the `From` implementation? Maybe it'll be good to explicitly see here that we'll continue running, despite the error (`TaskNextAction::ErrorContinue(err)`).

crates/services/p2p/src/service.rs
```diff
@@ -995,11 +999,9 @@ where
                     }
                 }
                 self.next_check_time += self.heartbeat_check_interval;
+                TaskNextAction::Continue
             }
         }
-
-        tracing::debug!("P2P task is finished");
```
rafal-ch, line 1001: Maybe we can also get rid of the "is running" [debug print](https://github.com/FuelLabs/fuel-core/pull/2429/files#diff-f4a43f3e5d08c9926062c05e8f8e237ba1373bbdf51be7fda0d91da708e5b959L858)?

rafal-ch (approved):
LGTM :+1: 
with some nits/suggestions

Anyway, good piece of QoL code :)

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -568,16 +568,19 @@ where
                     // de-synchronization between on-chain and off-chain databases.
                     if let Err(e) = result {
                         tracing::error!("Error processing block: {:?}", e);
-                        should_continue = self.continue_on_error;
+                        if self.continue_on_error {
+                            TaskNextAction::Continue
```
acerone85, line 572: why not `ErrorContinue` here?

acerone85 (approved):

crates/services/src/service.rs
```diff
@@ -83,6 +83,49 @@ pub trait RunnableService: Send {
     ) -> anyhow::Result<Self::Task>;
 }
 
+/// The result of a single iteration of the service task
+pub enum TaskNextAction {
+    /// Request the task to be run again
+    Continue,
```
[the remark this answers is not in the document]
MitchTurner: I prefer as explicit as possible.

crates/fuel-core/src/graphql_api/worker_service.rs
MitchTurner: Yeah. Good idea. I was going to say that I want to maintain the old behavior, but that's essentially the same and cleaner.

rafal-ch (approved):

acerone85 (approved):

xgreenx (approved):

xgreenx (approved):

AurelienFT (approved):

Issue #2435, opened 2024-11-14
Flaky test: `heavy_tasks_doesnt_block_graphql`


rafal-ch:
`heavy_tasks_doesnt_block_graphql` seems to occasionally fail in the CI.

Failed build:
https://github.com/FuelLabs/fuel-core/actions/runs/11819543221/job/32929762981?pr=2389

However, the hyper-related error messages seem to be unrelated, because the test intentionally puts a high load on the GraphQL endpoint. What is important is `Health check timed out: Elapsed(())`.

The test itself seem to be constructed correctly, allowing 5 sec. timeout before concluding that node is unresponsive.

Log:
```
failures:

---- dos::heavy_tasks_doesnt_block_graphql stdout ----
thread 'dos::heavy_tasks_doesnt_block_graphql' panicked at tests/tests/dos.rs:720:25:
Health check timed out: Elapsed(())
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'tokio-runtime-worker' panicked at /home/runner/actions-runner/_work/fuel-core/fuel-core/tests/test-helpers/src/lib.rs:31:61:
called `Result::unwrap()` on an `Err` value: reqwest::Error { kind: Request, url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Ipv4(127.0.0.1)), port: Some(36715), path: "/v1/graphql", query: None, fragment: None }, source: hyper::Error(IncompleteMessage) }
thread 'tokio-runtime-worker' panicked at /home/runner/actions-runner/_work/fuel-core/fuel-core/tests/test-helpers/src/lib.rs:31:61:
called `Result::unwrap()` on an `Err` value: reqwest::Error { kind: Request, url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Ipv4(127.0.0.1)), port: Some(36715), path: "/v1/graphql", query: None, fragment: None }, source: hyper::Error(IncompleteMessage) }
thread 'tokio-runtime-worker' panicked at /home/runner/actions-runner/_work/fuel-core/fuel-core/tests/test-helpers/src/lib.rs:31:61:
[...]

PR #2401, opened 2024-10-28
Resolve some falky tests and improve CI times

AurelienFT:
## Linked Issues/PRs
Fix https://github.com/FuelLabs/fuel-core/issues/2408
Fix https://github.com/FuelLabs/fuel-core/issues/2407
Fix https://github.com/FuelLabs/fuel-core/issues/2406
Fix https://github.com/FuelLabs/fuel-core/issues/2351

Fix https://github.com/FuelLabs/fuel-core/issues/2393
Fix https://github.com/FuelLabs/fuel-core/issues/2394
Fix https://github.com/FuelLabs/fuel-core/issues/2395

## Description

This PR fix an issue in P2P heartbeat. The problem was that P2P heartbeat was updated only if new blocks were received or produced. This means that if we start the node from an existing db but doesn't produce blocks and not connect it to anyone it will send block height 0 to the peers that connects to him. We believe that this fix, resolves #2408 #2407 #2406 and #2351.

For #2394 we just increased the timeouts.
For #2393 we removed the panic in the test and just let p2p reconnect
For #2395 we launch this test using multi-threads mode of Tokio to follow the convention of all the others tests that launch a node using `FuelCoreDriver`. Also we added a kill of the driver to try to kill the node in a more graceful way in all of the test, it should fix a lot of flakyness in these tests

This PR also change the CI workflow by removing all docker related jobs and codecov job. These two set of jobs has been moved to separated workflow that are not triggered automatically but can be triggered manually on the "Actions" tab of this repository (after the merge of this PR).

The tests launched by the CI job now use `nextest` that allow us to add timeout for each test and provide more detailed output. The timeout is currently 5 min (and 8 for two really big tests) because we have tests that take a long time but we should lower it in the future.
The steps on the matrix are not cancelled anymore when one failed to allow possible other success and cache their success for a relaunch of the tests.

There is still more improve to do on our tests especially on timeout and rapid execution but this should improve a lot our workflow.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog
- [x] New behavior is reflected in tests
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [x] I have reviewed the code myself
- [x] I have created follow-up issues caused by this PR and linked them here

PR #2437, opened 2024-11-14
Attempt to dodge flakiness in `heavy_tasks_doesnt_block_graphql` test

rafal-ch:
Closes https://github.com/FuelLabs/fuel-core/issues/2435

## Description
This is an attempt to resolve observed flakiness in the `heavy_tasks_doesnt_block_graphql` test.

When I tested the fix locally, with much smaller (250ms) timeouts and debug prints, I could observe the following outcome:
```
running 1 test
ERR
ERR
OK
test dos::heavy_tasks_doesnt_block_graphql ... ok
```

Changes:
1. Allow retrying the "health" check request 3 times
2. Because we now allow 3 tries instead of just 1
   3. Reduce timeout from 5 to 4 seconds
   4. Spam the node with 3 times as much requests in the background (50 -> 150).

This is an effort to make the test more resilient to the performance of the machine it is executed on.

### Before requesting review

### DIFF
--- a/tests/tests/dos.rs
+++ b/tests/tests/dos.rs
@@ -701,7 +701,7 @@ async fn heavy_tasks_doesnt_block_graphql() {
     client.produce_blocks(NUM_OF_BLOCKS, None).await.unwrap();
 
     // Given
-    for _ in 0..50 {
+    for _ in 0..150 {
         let url = url.clone();
         let query = query.clone();
         tokio::spawn(async move {
@@ -713,11 +713,27 @@ async fn heavy_tasks_doesnt_block_graphql() {
     // Wait for all queries to start be processed on the node.
     tokio::time::sleep(Duration::from_secs(1)).await;
 
+    const RETRIES: u32 = 3;
+    const TIMEOUT_SECS: u64 = 4;
+
     // When
-    let result = tokio::time::timeout(Duration::from_secs(5), client.health()).await;
+    let mut retries = RETRIES;
+    let result = loop {
+        if retries == 0 {
+            panic!("Health check timed out after 3 retries");
+        }
+
+        match tokio::time::timeout(Duration::from_secs(TIMEOUT_SECS), client.health())
+            .await
+        {
+            Ok(result) => break result,
+            Err(_) => retries -= 1,
+        }
+
+        // Intentionally no sleep between retries, try ASAP to make sure that the node is still under load.
+    };
 
     // Then
-    let result = result.expect("Health check timed out");
     let health = result.expect("Health check failed");
     assert!(health);
 }


netrome (commented):
I'm a bit hesitant about having retry logic here. Isn't the point of the test to ensure that the service is available despite the load? Adding retry logic seems like it defeats the purpose of the test.

What is the reason the health query can time out other than the heavy tasks actually blocking the queries? I'd take it that if this test still fails occasionally in CI, that we'd need to look into if we can do further improvements in reducing the risk of the heavy tasks blocking the other requests.

rafal-ch:
> I'm a bit hesitant about having retry logic here. Isn't the point of the test to ensure that the service is available despite the load? Adding retry logic seems like it defeats the purpose of the test.

The reason why I did it like that is that we do not give strict guarantees about the service responsiveness under the load. One could ask why we allowed exactly 5 seconds in the original test implementation? Does it mean that we should always respond within 5 sec., regardless the load and the machine we're running at? Probably not. I figured that having a couple of short retries would be better than just increasing the timeout arbitrarily.

> we can do further improvements in reducing the risk of the heavy tasks blocking the other requests

The question here is: reduce to what level? In fact, no matter what improvements you do, you'll always observe some randomness in response time based on the load of the machine, etc.

Anyway, I'm happy to bench and work on potential improvements in this regard if we see the need. Especially if we observe this test is still failing with the retries. For this PR though, the goal is to make CI runs less flaky.

netrome:
> > I'm a bit hesitant about having retry logic here. Isn't the point of the test to ensure that the service is available despite the load? Adding retry logic seems like it defeats the purpose of the test.
> 
> The reason why I did it like that is that we do not give strict guarantees about the service responsiveness under the load. One could ask why we allowed exactly 5 seconds in the original test implementation? Does it mean that we should always respond within 5 sec., regardless the load and the machine we're running at? Probably not. I figured that having a couple of short retries would be better than just increasing the timeout arbitrarily.
> 
> > we can do further improvements in reducing the risk of the heavy tasks blocking the other requests
> 
> The question here is: reduce to what level? In fact, no matter what improvements you do, you'll always observe some randomness in response time based on the load of the machine, etc.
> 
> Anyway, I'm happy to bench and work on potential improvements in this regard if we see the need. Especially if we observe this test is still failing with the retries, which means we have a deeper issue. For this PR though, the goal is to make CI runs less flaky.

Fair enough. I guess part of the problem here is that the original test is inherently flaky, and that the responsiveness behavior we're testing isn't well defined. I'd question if we should even keep the test or just disable it/remove it. But since this actually makes our CI behave better I'll approve.

netrome (approved):

xgreenx (commented):
> One could ask why we allowed exactly 5 seconds in the original test implementation? Does it mean that we should always respond within 5 sec., regardless the load and the machine we're running at? Probably not.

The machine should answer in 5 seconds regardless of its load. Otherwise, the liveness check will fail.

Maybe we should consider running this test without other tests in parallel to avoid thread-blocking tasks or having too many threads.

Also, maybe we need just to increase `number_of_threads` in the config

image

xgreenx (commented):
@rafal-ch Could you check, do you still have a problem after https://github.com/FuelLabs/fuel-core/pull/2401 ?

rafal-ch:
> @rafal-ch Could you check, do you still have a problem after #2401 ?

With the original 5 sec timeout I was never able to reproduce the problem locally (neither with #2401 nor without it). It was happening on CI only.

When I reduce the timeout the test is still flaky on my local machine.

**In short:** no change in the behavior observed.

AurelienFT:
I didn't see this test pop in the last CI runs we had. Can we close the PR but not the branch and instantiate it again if we see the test fails again ? @rafal-ch

rafal-ch:
> I didn't see this test pop in the last CI runs we had. Can we close the PR but not the branch and instantiate it again if we see the test fails again ? @rafal-ch

Yeah, closing for now.

PR #2445, opened 2024-11-20
Create new index for tracking Asset metadata

### LINKED ISSUE #2425: Add indexation `AssetId -> (ContractId, SubId)`

maschad:
**Note:**
Please merge https://github.com/FuelLabs/fuel-core/pull/2463 first

---

## Linked Issues/PRs
- Closes #2425

- Related to https://github.com/FuelLabs/fuels-ts/issues/3373

- It's based on top of https://github.com/FuelLabs/fuel-core/pull/2463 in order to have access to several mechanisms (like `IndexationError`) implemented previously for other indexes.

## Description

This introduces a new GraphQL endpoint that returns info about an asset. Similarly to other indexes, the asset metadata index is only available when the client was started with no database and was able to sync from scratch. If the index is not available, this endpoint will return an error:
```
{
  "data": null,
  "errors": [
    {
      "message": "Asset metadata index is not available",
      "locations": [...],
    }
  ]
}
```

The off-chain worker indexes `Mint` and `Burn` events to store this metadata about an `AssetId`.

## Checklist

### Before requesting review

- [ ] [TS SDK](https://github.com/FuelLabs/fuels-ts/)


crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -356,6 +361,32 @@ where
...
+                    contract_id,
+                    ..
+                }
+                | Receipt::Burn {
+                    sub_id,
+                    contract_id,
+                    ..
+                } => {
+                    let asset_id = AssetId::from(**contract_id);
+                    let current_count = db
```
xgreenx, line 378: You need to increase/decrease based on the minted/burned amount=)

maschad: Good catch 😄 ab09b8df08

tests/tests/assets.rs
```diff
@@ -0,0 +1,117 @@
...
+    };
+    let config = Config::local_node_with_state_config(state);
+
+    FuelService::from_database(Database::default(), config)
+        .await
+        .unwrap()
+}
+
+#[tokio::test]
+#[ignore] // TODO: Need to be able to mint assets for this test to work
```
maschad, line 39: I'm not quite sure how to enable permissions to be able to mint assets for the purposes of this test

tests/tests/assets.rs
```diff
@@ -0,0 +1,117 @@
...
+    let srv = setup_service(vec![coin]).await;
+    let client = FuelClient::from(srv.bound_address);
+
+    // run test
+    let asset_id = AssetId::new([1u8; 32]);
+    let mint_amount = 100;
+    let burn_amount = 50;
+
+    // Mint coins first
+    let mint_tx = TransactionBuilder::mint(
```
AurelienFT, line 60: I think this is not how you mint assets. From the spec : https://github.com/FuelLabs/fuel-specs/blob/master/src/tx-format/transaction.md#transactionmint this transaction is only used by the block producer that's why it's not accepted in the pool.
It's currently used for block rewards I think and it has this much fields for future usage.
However I don't know how to mint asset with the Native assets I never used it but I think it should be that way that you should mint your coins (using script or calls). Maybe someone else in the @FuelLabs/client has more details.

AurelienFT (changes requested):
I think we need the test to be working before merging this.

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -356,6 +361,54 @@ where
...
+                        .storage::<AssetsInfo>()
+                        .get(&asset_id)?
+                        .map(|info| {
+                            info.2
+                                .checked_sub(*val)
+                                .ok_or(anyhow::anyhow!("Asset count overflow"))
```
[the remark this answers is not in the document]
maschad: 6ae205140c

AurelienFT (approved):

tests/tests/assets.rs
AurelienFT: Fixed it https://github.com/FuelLabs/fuel-core/pull/2445/commits/098b0317c87d1ed2712c15d84548e81129bf286e

maschad:
Thanks for all your help @AurelienFT <3

crates/client/assets/schema.sdl
```diff
@@ -2,6 +2,12 @@ scalar Address
 
 scalar AssetId
 
+type AssetInfoDetails {
+	contractId: HexString!
```
rafal-ch, line 6: Not sure here, but maybe we want to use `ContractId!` here, as in [other places](https://github.com/FuelLabs/fuel-core/blob/master/crates/client/assets/schema.sdl#L594).

crates/fuel-core/src/graphql_api/ports.rs
```diff
@@ -128,6 +131,10 @@ pub trait OffChainDatabase: Send + Sync {
     ) -> StorageResult<Option<RelayedTransactionStatus>>;
 
     fn message_is_spent(&self, nonce: &Nonce) -> StorageResult<bool>;
+
+    fn asset_info(&self, asset_id: &AssetId) -> StorageResult<Option<AssetDetails>>;
+
+    fn asset_exists(&self, asset_id: &AssetId) -> StorageResult<bool>;
```
rafal-ch, line 137: Seems like `asset_exists` is not used, can we remove it?

crates/fuel-core/src/graphql_api/storage/assets.rs
```diff
@@ -0,0 +1,48 @@
...
+};
+use fuel_core_types::{
+    fuel_merkle::common::Bytes32,
+    fuel_tx::{
+        AssetId,
+        ContractId,
+    },
+};
+
+/// Contract info
```
rafal-ch, line 18:
```suggestion
/// Asset info
```

crates/fuel-core/src/graphql_api/storage/assets.rs
```diff
@@ -0,0 +1,48 @@
...
+pub struct AssetsInfo;
+
+pub type AssetDetails = (ContractId, Bytes32, u64); // (contract_id, sub_id, total_amount)
```
rafal-ch, line 21: Can we use struct with named fields? Something similar to what we have [here](https://github.com/FuelLabs/fuel-core/pull/2445/files#diff-ac9c0df77a716f8ffa16e831870ec184fd1e5066adc11825d415bcfa776cdd4bR39).

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -381,6 +386,64 @@ where
...
+                        })
+                        .transpose()?
+                        .unwrap_or(*val);
+
+                    db.storage::<AssetsInfo>()
+                        .insert(&asset_id, &(*contract_id, **sub_id, current_count))?;
+                }
+                _ => {}
```
rafal-ch, line 444: Please enumerate all variants explicitly to avoid surprises when new variant is added in the future.

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -381,6 +386,64 @@ where
...
+        for receipt in result.receipts() {
+            match receipt {
+                Receipt::Mint {
+                    sub_id,
+                    contract_id,
+                    val,
+                    ..
+                } => {
+                    let asset_id = contract_id.asset_id(sub_id);
+                    let current_count = match db.storage::<AssetsInfo>().get(&asset_id) {
```
rafal-ch, line 399: What we're getting back here is actually not a count. Can you consider renaming the binding?

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -381,6 +386,64 @@ where
...
+                        Err(_) => {
+                            // If asset doesn't exist yet, create it with 0 count
+                            db.storage::<AssetsInfo>()
+                                .insert(&asset_id, &(*contract_id, **sub_id, 0))?;
+                            Some(Cow::Owned((*contract_id, **sub_id, 0)))
+                        }
+                    }
+                    .map(|info| {
+                        info.2
+                            .checked_add(*val)
```
rafal-ch, line 410:
```suggestion
                        .map(|info| {
                            let (_, _, count) = *info;
                            count
                                .checked_add(*val)
```

May also affect the `Receipt::Burn` branch.

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -381,6 +386,64 @@ where
...
+                    let asset_id = contract_id.asset_id(sub_id);
+                    let current_count = match db.storage::<AssetsInfo>().get(&asset_id) {
+                        Ok(count) => count,
+                        Err(_) => {
+                            // If asset doesn't exist yet, create it with 0 count
+                            db.storage::<AssetsInfo>()
+                                .insert(&asset_id, &(*contract_id, **sub_id, 0))?;
+                            Some(Cow::Owned((*contract_id, **sub_id, 0)))
+                        }
+                    }
```
rafal-ch, line 407: I think we shouldn't treat all storage errors as "value not found". Some of them should just be bubbled up.

Maybe something like this will work:

```
                        let current_count = match db.storage::<AssetsInfo>().get(&asset_id)? {
                            Some(count) => Some(count),
                            None => {
                                // If asset doesn't exist yet, create it with 0 count
                                db.storage::<AssetsInfo>()
                                    .insert(&asset_id, &(*contract_id, **sub_id, 0))?;
                                Some(Cow::Owned((*contract_id, **sub_id, 0)))
                            }
                        }
```

If you agree, please note that now two branches always return `Some`, so maybe we can get rid of the follow-up `map()` as well.

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -381,6 +386,64 @@ where
...
+                    let current_count = db
+                        .storage::<AssetsInfo>()
+                        .get(&asset_id)
+                        .unwrap_or_else(|_| {
+                            tracing::warn!("Asset {} is not currently indexed", asset_id);
+                            None
+                        })
+                        .map(|info| {
+                            info.2
+                                .checked_sub(*val)
```
rafal-ch, line 435: If we go down to `0` on `Burn` should we remove the asset from the index? Unless we need to distinguish between "never minted" and "minted and totally burned" cases.

Anyway, currently the index may grow indefinitely.

Which also makes me think that I should consider this for the balances indexation I added recently :-)

cc: @xgreenx, wdyt?

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -381,6 +386,64 @@ where
...
+                            // If asset doesn't exist yet, create it with 0 count
+                            db.storage::<AssetsInfo>()
+                                .insert(&asset_id, &(*contract_id, **sub_id, 0))?;
+                            Some(Cow::Owned((*contract_id, **sub_id, 0)))
+                        }
+                    }
+                    .map(|info| {
+                        info.2
+                            .checked_add(*val)
+                            .ok_or(anyhow::anyhow!("Asset count overflow"))
```
rafal-ch, line 411: You may consider integrating this with the `IndexationError` available here: https://github.com/FuelLabs/fuel-core/blob/master/crates/fuel-core/src/graphql_api/indexation.rs#L28

It's a common type for all errors related to integration and it'll grow even more variants after the "coins to spend" indexation is merged. See [here](https://github.com/FuelLabs/fuel-core/pull/2463/files#diff-b6049ace8304a0ac864e77b0cc1d40c8501613ff153eb7ac0a66a0bc191134de).

tests/tests/assets.rs
```diff
@@ -0,0 +1,206 @@
...
+    let final_supply = client
+        .asset_info(&contract_id.asset_id(&Bytes32::zeroed()))
+        .await
+        .unwrap()
+        .unwrap()
+        .total_supply
+        .0;
+
+    // Then
+    // We should have the minted amount first
```
rafal-ch, line 204:
```suggestion
    // We should have the minted amount reduced by the burned amount
```

crates/fuel-core/src/graphql_api/worker_service.rs
rafal-ch, line 410: Consider the following scenario:
- client v.0.40 running
- asset A in contract B is minted (count=10)
- we update the client to the one that supports index
- asset A in contract B is minted again (count=10, for a total count of 20)
- we query the asset info
- it notices there is no index, initializes it with 0 and adds 10
- we have inconsistency (10 vs 20)

I may not have full picture, but if you see this as a possible scenario, please consider enabling access to this endpoint conditionally, ie. it should be available only when the client was started with a clear database and have a chance to sync from genesis (and build the complete index). This is what we did with other indexes (see for example around [here](https://github.com/FuelLabs/fuel-core/pull/2383/files#diff-c95a3d57a39feac7c8c2f3b193a24eec39e794413adc741df36450f9a4539898R514-R546)).

crates/fuel-core/src/graphql_api/worker_service.rs
xgreenx: We want to keep it even with 0, because we need relation `AssetId -> (ContractId, SubId)`. it is main feature why we add it=)

crates/fuel-core/src/graphql_api/ports.rs
rafal-ch: Addressed in https://github.com/FuelLabs/fuel-core/pull/2445/commits/319d32d3deeaf32a2ef3ba886dc5decca577532e

crates/fuel-core/src/graphql_api/storage/assets.rs
rafal-ch: Addressed in https://github.com/FuelLabs/fuel-core/pull/2445/commits/319d32d3deeaf32a2ef3ba886dc5decca577532e

crates/fuel-core/src/graphql_api/storage/assets.rs
rafal-ch: Addressed in https://github.com/FuelLabs/fuel-core/pull/2445/commits/319d32d3deeaf32a2ef3ba886dc5decca577532e

crates/fuel-core/src/graphql_api/worker_service.rs
rafal-ch: Addressed in https://github.com/FuelLabs/fuel-core/pull/2445/commits/319d32d3deeaf32a2ef3ba886dc5decca577532e

crates/fuel-core/src/graphql_api/worker_service.rs
rafal-ch: Addressed in https://github.com/FuelLabs/fuel-core/pull/2445/commits/319d32d3deeaf32a2ef3ba886dc5decca577532e

crates/fuel-core/src/graphql_api/worker_service.rs
rafal-ch: Addressed in https://github.com/FuelLabs/fuel-core/pull/2445/commits/319d32d3deeaf32a2ef3ba886dc5decca577532e

tests/tests/assets.rs
rafal-ch: Addressed in https://github.com/FuelLabs/fuel-core/pull/2445/commits/319d32d3deeaf32a2ef3ba886dc5decca577532e

crates/fuel-core/src/graphql_api/worker_service.rs
rafal-ch: Updated in a88e9ef052d99e5e137eaf81720a192d63413b99

crates/fuel-core/src/graphql_api/worker_service.rs
rafal-ch: Addressed in 8a1324cc32555056090b9767ed4fa1d2f86825e3

xgreenx (dismissed):

AurelienFT (approved):
LGTM

AurelienFT:
@rafal-ch I pressed the button myself to allow other PR to update on master now and don't loose time for release. I hope you don't mind :p

rafal-ch:
> @rafal-ch I pressed the button myself to allow other PR to update on master now and don't loose time for release. I hope you don't mind :p

Thanks, I was away for a couple of minutes :)

PR #2447, opened 2024-11-21
Use expiration policy

AurelienFT:
## Linked Issues/PRs
Spec PR : https://github.com/FuelLabs/fuel-specs/pull/616

VM PR : https://github.com/FuelLabs/fuel-vm/pull/871

## Description
This PR use the new policy `expiration` in transaction pool to prune transactions when they are expired to prevent them to be included in a block.

Also verification is added when executing transaction in the executor.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/txpool_v2/src/service.rs
```diff
@@ -323,6 +324,29 @@ where
...
+                .range(..=new_height)
+                .map(|(k, _)| *k)
+                .collect::<Vec<_>>();
+            for height in range_to_remove {
+                let expired_txs = height_expiration_txs.remove(&height);
+                if let Some(expired_txs) = expired_txs {
+                    let mut tx_pool = self.pool.write();
+                    removed_txs
+                        .extend(tx_pool.remove_transaction_and_dependents(expired_txs));
+                }
```
rymnc, line 342: is it possible that when you call `tx_pool.remove_transaction_and_dependents` you end up removing the same tx twice? i.e once as a dependent for another transaction and once on its own?

crates/services/txpool_v2/src/service.rs
```diff
@@ -323,6 +324,29 @@ where
...
+            }
+        }
+        for tx in removed_txs {
```
rymnc, line 345: would be nice to know if `removed_txs` are unique

crates/services/txpool_v2/src/service.rs
```diff
@@ -764,6 +796,7 @@ where
     let pruner = TransactionPruner {
         txs_ttl: config.max_txs_ttl,
         time_txs_submitted: Arc::new(RwLock::new(VecDeque::new())),
+        height_expiration_txs: Arc::new(RwLock::new(BTreeMap::new())),
```
rymnc, line 799: can't we use dashmap here?
will prevent usage of `RwLock`

also may be worth to think about usage of channels to send transactions that need to be pruned and then the `TransactionPruner` can handle them appropriately according to some configuration

crates/services/txpool_v2/src/tests/tests_service.rs
```diff
@@ -236,6 +252,89 @@ async fn test_prune_transactions_the_oldest() {
     service.stop_and_await().await.unwrap();
 }
 
+#[tokio::test]
+async fn prune_expired_transactions() {
```
rymnc, line 256: can you add a test that checks if the `tx_status_sender` does not send duplicate tx statuses for transactions that may have been removed twice acc to my earlier comment?

crates/services/txpool_v2/src/tests/universe.rs
```diff
@@ -93,6 +94,7 @@ use super::mocks::{
 // the byte and gas price fees.
 pub const TEST_COIN_AMOUNT: u64 = 100_000_000u64;
 pub const GAS_LIMIT: Word = 100000;
+pub const DEFAULT_EXPIRATION_HEIGHT: BlockHeight = BlockHeight::new(1000);
```
rymnc, line 97:
```suggestion
pub const DEFAULT_EXPIRATION_HEIGHT: BlockHeight = BlockHeight::new(10);
```
1_000 seems like state would change too much, no?

crates/client/src/client/schema/chain.rs
```diff
@@ -472,6 +473,129 @@ impl TryFrom<GasCosts> for fuel_core_types::fuel_tx::GasCosts {
                 }
                 .into(),
             )),
+            GasCostsVersion::V2 => {
```
rymnc, line 476: curious why we need this change

rymnc (commented):
left some comments

crates/services/txpool_v2/src/service.rs
AurelienFT: Yeah or even if a transaction is the dependency of two others. But in the implementation it will not remove twice because of https://github.com/FuelLabs/fuel-core/blob/ccae2d50c4d694a8c77500d2843a3322f694e80b/crates/services/txpool_v2/src/storage/graph.rs#L144

crates/services/txpool_v2/src/service.rs
AurelienFT: Yeah removed_txs should be unique in any cases because we don't want to push removal notification twice for the same id (it should'nt happens but who knows). I don't know if changing `removed_txs` to `Hashset` is ok because we iterate over it and might be performance consuming

crates/services/txpool_v2/src/tests/universe.rs
AurelienFT: This is a test onyl feature to be sure it doesn't change the existing tests that may be running for long. In the real world the default value is "no expiration"

crates/services/txpool_v2/src/service.rs
AurelienFT: Yeah we can. We didn't use it across the code yet but could be a great first place experiment but is it really useful to it if we want to change to channels etc (which could a great idea that I need to study). Are you ok if I just open an issue about improving the transaction pruner ?

crates/services/txpool_v2/src/service.rs
rymnc: okay :)

crates/services/txpool_v2/src/service.rs
rymnc: sure, feel free to do that :)

crates/services/txpool_v2/src/tests/universe.rs
rymnc: should the default value be "no expiration"? is that for backwards compatibility or some other reason?

rymnc (approved):

crates/services/txpool_v2/src/tests/universe.rs
AurelienFT: For tests I find it easier to write test in the future then and so they all shared a same constant instead of magical numbers in each tests that would endup in a constant

rymnc: oh, I'm curious about the "real world" default for tx expiry, not the tests here :)

rymnc (approved):

crates/services/txpool_v2/src/tests/universe.rs
AurelienFT: In real world it's "no expiration"

Dentosal (approved):

PR #2451, opened 2024-11-23
Shared sequencer integration for `fuel-core 0.40.0`

xgreenx:
## Description

This PR duplicates https://github.com/FuelLabs/fuel-core/pull/1922, but on top of the `fuel-core 0.40.0`.
Since the current `master` is not ready to be released, but we want this feature in the testnet, we will create a minor release with added off-chain logic that posts data to the SS.

All changes are the same as in the original PR, except `reqwest` is bumped to `0.12`, because it has a fix important for SS.

### Before requesting review

### DIFF
--- a/bin/fuel-core/src/cli/run/shared_sequencer.rs
+++ b/bin/fuel-core/src/cli/run/shared_sequencer.rs
@@ -1,0 +1,56 @@
+use fuel_core_types::fuel_types::Bytes32;
+
+#[derive(Debug, Clone, clap::Args)]
+pub struct Args {
+    /// If set to true, new blocks will be posted to the shared sequencer chain
+    #[clap(long = "enable-ss", action)]
+    enable: bool,
+    /// If set to true, new blocks will be posted to the shared sequencer chain
+    #[clap(long = "ss-block-posting-frequency", env, default_value = "12s")]
+    block_posting_frequency: humantime::Duration,
+    /// The RPC address of the sequencer chain tendermint API
+    /// (e.g. "http://127.0.0.1:26657")
+    #[clap(long = "ss-tendermint-api", env)]
+    tendermint_api: Option<String>,
+    /// The REST address of the sequencer chain blockchain/rest API
+    /// (e.g. "http://127.0.0.1:1317")
+    #[clap(long = "ss-blockchain-api", env)]
+    blockchain_api: Option<String>,
+    /// Topic to post blocks to
+    /// (e.g. "1111111111111111111111111111111111111111111111111111111111111111")
+    #[clap(
+        long = "ss-topic",
+        env,
+        default_value = "0000000000000000000000000000000000000000000000000000000000000000"
+    )]
+    topic: Bytes32,
+}
+
+#[cfg(feature = "shared-sequencer")]
+impl TryFrom<Args> for fuel_core_shared_sequencer::Config {
+    type Error = anyhow::Error;
+
+    fn try_from(val: Args) -> anyhow::Result<fuel_core_shared_sequencer::Config> {
+        let endpoints = match (val.tendermint_api, val.blockchain_api) {
+            (Some(tendermint_api), Some(blockchain_api)) => {
+                Some(fuel_core_shared_sequencer::Endpoints {
+                    tendermint_rpc_api: tendermint_api,
+                    blockchain_rest_api: blockchain_api,
+                })
+            }
+            (None, None) => None,
+            _ => {
+                return Err(anyhow::anyhow!(
+                    "Both tendermint and blockchain API must be set or unset"
+                ))
+            }
+        };
+
+        Ok(fuel_core_shared_sequencer::Config {
+            enabled: val.enable,
+            block_posting_frequency: val.block_posting_frequency.into(),
+            endpoints,
+            topic: *val.topic,
+        })
+    }
+}
--- a/bin/fuel-core/src/cli/run/shared_sequencer.rs
+++ b/bin/fuel-core/src/cli/run/shared_sequencer.rs
@@ -2,10 +2,10 @@ use fuel_core_types::fuel_types::Bytes32;
 
 #[derive(Debug, Clone, clap::Args)]
 pub struct Args {
-    /// If set to true, new blocks will be posted to the shared sequencer chain
+    /// If set to true, new blocks will be posted to the shared sequencer chain.
     #[clap(long = "enable-ss", action)]
     enable: bool,
-    /// If set to true, new blocks will be posted to the shared sequencer chain
+    /// The frequency at which to post blocks to the shared sequencer.
     #[clap(long = "ss-block-posting-frequency", env, default_value = "12s")]
     block_posting_frequency: humantime::Duration,
     /// The RPC address of the sequencer chain tendermint API
--- a/crates/services/shared-sequencer/src/config.rs
+++ b/crates/services/shared-sequencer/src/config.rs
@@ -3,10 +3,10 @@ use std::time::Duration;
 /// Endpoints for the shared sequencer client.
 #[derive(Debug, Clone)]
 pub struct Endpoints {
-    /// The RPC address of the sequencer chain tendermint API
+    /// The RPC address of the sequencer chain
     /// (e.g. "http://127.0.0.1:26657")
     pub tendermint_rpc_api: String,
-    /// The REST address of the sequencer chain tendermint API
+    /// The REST address of the sequencer chain
     /// (e.g. "http://127.0.0.1:1317")
     pub blockchain_rest_api: String,
 }
--- a/crates/services/shared-sequencer/src/config.rs
+++ b/crates/services/shared-sequencer/src/config.rs
@@ -1,0 +1,37 @@
+use std::time::Duration;
+
+/// Endpoints for the shared sequencer client.
+#[derive(Debug, Clone)]
+pub struct Endpoints {
+    /// The RPC address of the sequencer chain tendermint API
+    /// (e.g. "http://127.0.0.1:26657")
+    pub tendermint_rpc_api: String,
+    /// The REST address of the sequencer chain tendermint API
+    /// (e.g. "http://127.0.0.1:1317")
+    pub blockchain_rest_api: String,
+}
+
+/// Configuration for the shared sequencer client
+#[derive(Debug, Clone)]
+pub struct Config {
+    /// Whether the sequencer is enabled.
+    pub enabled: bool,
+    /// The frequency at which to post blocks to the shared sequencer.
+    pub block_posting_frequency: Duration,
+    /// Endpoints for the shared sequencer client.
+    pub endpoints: Option<Endpoints>,
+    /// Topic to post blocks to
+    pub topic: [u8; 32],
+}
+
+impl Config {
+    /// Default configuration for locally running shared sequencer node
+    pub fn local_node() -> Self {
+        Self {
+            enabled: false,
+            block_posting_frequency: Duration::from_secs(12),
+            endpoints: None,
+            topic: [0u8; 32],
+        }
+    }
+}
--- a/crates/services/shared-sequencer/src/http_api.rs
+++ b/crates/services/shared-sequencer/src/http_api.rs
@@ -1,0 +1,181 @@
+use base64::prelude::*;
+use cosmrs::AccountId;
+
+mod api_types {
+    use serde::Deserialize;
+
+    #[derive(Debug, Deserialize)]
+    pub struct AccountResponse {
+        pub account: AccountInfo,
+    }
+
+    #[derive(Debug, Deserialize)]
+    pub struct AccountInfo {
+        pub account_number: String,
+        pub sequence: String,
+    }
+
+    #[derive(Debug, Deserialize)]
+    pub struct AccountPrefix {
+        pub bech32_prefix: String,
+    }
+
+    #[derive(Debug, Deserialize)]
+    pub struct NodeInfo {
+        pub default_node_info: DefaultNodeInfo,
+    }
+
+    #[derive(Debug, Deserialize)]
+    pub struct DefaultNodeInfo {
+        pub network: String,
+    }
+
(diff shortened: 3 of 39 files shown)


bin/fuel-core/src/cli/run/shared_sequencer.rs
```diff
@@ -0,0 +1,56 @@
+use fuel_core_types::fuel_types::Bytes32;
+
+#[derive(Debug, Clone, clap::Args)]
+pub struct Args {
+    /// If set to true, new blocks will be posted to the shared sequencer chain
+    #[clap(long = "enable-ss", action)]
+    enable: bool,
+    /// If set to true, new blocks will be posted to the shared sequencer chain
```
AurelienFT, line 8: Comment is not related to this argument

crates/services/shared-sequencer/src/config.rs
```diff
@@ -0,0 +1,37 @@
+use std::time::Duration;
+
+/// Endpoints for the shared sequencer client.
+#[derive(Debug, Clone)]
+pub struct Endpoints {
+    /// The RPC address of the sequencer chain tendermint API
+    /// (e.g. "http://127.0.0.1:26657")
+    pub tendermint_rpc_api: String,
+    /// The REST address of the sequencer chain tendermint API
```
AurelienFT, line 9: I think the comment isn't related to the arg documented. In my mind it the same as the one on clap defined here: https://github.com/FuelLabs/fuel-core/blob/38dfd569eb6f56446b3dedcc8bc4033ce228e333/bin/fuel-core/src/cli/run/shared_sequencer.rs#L15

crates/services/shared-sequencer/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+        &self,
+        signer: &S,
+        account: AccountMetadata,
+        order: u64,
+        blob: Vec<u8>,
+    ) -> anyhow::Result<()> {
+        let latest_height = self.latest_block_height().await?;
+
+        self.send_raw(
+            latest_height.saturating_add(100),
```
AurelienFT, line 147: Do we want to have it in a config variable in the future ? (Follow up issue angle)

crates/services/shared-sequencer/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+    pub async fn send_raw<S: Signer>(
+        &self,
+        timeout_height: u32,
+        signer: &S,
+        account: AccountMetadata,
+        order: u64,
+        topic: [u8; 32],
+        data: Bytes,
+    ) -> anyhow::Result<()> {
+        let dummy_amount = Coin {
```
AurelienFT, line 168: Can we get a comment here about the reason of this dummy computation first before the real one ?

crates/services/shared-sequencer/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+        .to_bytes()?)
+    }
+
+    /// Return all blob messages in the given blob
+    pub async fn read_block_msgs(&self, height: u32) -> anyhow::Result<Vec<MsgPostBlob>> {
+        let mut result = Vec::new();
+
+        let block = self.tendermint()?.block(height).await?;
+
+        for item in block.block.data.iter().skip(1) {
```
AurelienFT, line 269: Why do we skip one ?

crates/services/shared-sequencer/src/service.rs
```diff
@@ -0,0 +1,262 @@
...
+            // If the account is not funded, this code will fail
+            // because we can't sign the transaction without account metadata.
+            let account_metadata = ss.get_account_meta(self.signer.as_ref()).await;
+
+            match account_metadata {
+                Ok(account_metadata) => {
+                    self.account_metadata = Some(account_metadata);
+                }
+                Err(err) => {
+                    tokio::time::sleep(Duration::from_secs(1)).await;
```
AurelienFT, line 158: If I understand correctly this sleep is designed to avoid the `select` in the upper function to not select this branch directly. Can we comment them for future changes because I don't think it's a correct solution to have this correlation and this time-based fix.

crates/services/shared-sequencer/src/service.rs
```diff
@@ -0,0 +1,262 @@
...
+
+        tokio::time::sleep(self.config.block_posting_frequency).await;
+
+        let blobs = {
+            let mut lock = self.blobs.lock().await;
+            core::mem::take(&mut *lock)
+        };
+
+        if blobs.is_empty() {
+            tokio::time::sleep(Duration::from_secs(1)).await;
```
AurelienFT, line 176: Is this sleep really useful ? Because we don't need to wait to avoid instant trigger of this branch because of `tokio::time::sleep(self.config.block_posting_frequency).await;` L.168

crates/services/shared-sequencer/src/service.rs
```diff
@@ -0,0 +1,262 @@
...
+                    return Err(err);
+                }
+            }
+        }
+
+        if self.prev_order.is_none() {
+            self.prev_order = ss.get_topic().await?.map(|f| f.order);
+        }
+
+        tokio::time::sleep(self.config.block_posting_frequency).await;
```
AurelienFT, line 168: We might prefer a interval-based in the `select` solution because if this is cancel and another branch resolves this sleep will reset and we can endup to not publish blocks for hours. This can not happen now because the select doesn't have any other logic (except the service one) and so can't be cancelled. But this coupling in the future might be problematic. (except if we are ok with the sleep reset)

crates/services/shared-sequencer/src/service.rs
```diff
@@ -0,0 +1,262 @@
...
+                    };
+
+                    let ss =  self.shared_sequencer_client
+                        .as_ref().expect("Shared sequencer client is not set; qed");
+                    let blobs_bytes = postcard::to_allocvec(&blobs).expect("Failed to serialize SSBlob");
+                    let result = ss.send(self.signer.as_ref(), account, next_order, blobs_bytes).await;
+
+                    match result {
+                        Ok(_) => {
+                            tracing::info!("Posted block to shared sequencer {blobs:?}");
```
AurelienFT, line 222: Do we want to log the full bytes in the release at an info level ?

xgreenx: Yes, the log is very small and only posted once per 12 seconds.

crates/services/shared-sequencer/src/lib.rs
xgreenx: @Dentosal Added it, so I don't know=) I will remove this function for now, since it is not used

crates/services/shared-sequencer/src/lib.rs
xgreenx: We don't need configure it, added a comment

AurelienFT (approved):
Thank you for the comments :)

crates/services/shared-sequencer/src/http_api.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub account_number: u64,
+    pub sequence: u64,
+}
+
+#[derive(Clone, Debug, serde::Serialize)]
+pub struct SimulateRequest {
+    pub tx_bytes: String,
+}
+
+pub async fn estimate_transaction(
```
rymnc, line 81: do we want any retry logic here? can be a follow up PR

crates/services/shared-sequencer/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+    topic: [u8; 32],
+    chain_id: Id,
+    gas_price: u128,
+    coin_denom: Denom,
+    account_prefix: String,
+}
+
+impl Client {
+    /// Create a new shared sequencer client from config.
+    pub async fn new(endpoints: Endpoints, topic: [u8; 32]) -> anyhow::Result<Self> {
```
rymnc, line 57: i think it should also accept the expected client_id so that we don't misconfigure with the wrong network/chain

crates/services/shared-sequencer/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+            amount: self.gas_price.saturating_mul(used_gas as u128),
+            denom: self.coin_denom.clone(),
+        };
+
+        let fee = Fee::from_amount_and_gas(amount, used_gas);
+        let payload = self
+            .make_payload(timeout_height, fee, signer, account, order, topic, data)
+            .await?;
+
+        let r = self.tendermint()?.broadcast_tx_sync(payload).await?;
```
rymnc, line 207: retry logic here?

crates/types/src/signer.rs
```diff
@@ -71,7 +67,12 @@ impl SignMode {
                 cached_public_key_bytes,
             } => sign_with_kms(client, key_id, cached_public_key_bytes, message).await?,
         };
-        Ok(Consensus::PoA(PoAConsensus::new(poa_signature)))
+        Ok(signature)
+    }
+
+    /// Sign a blob of data
+    pub async fn sign(&self, data: &[u8]) -> anyhow::Result<Signature> {
+        self.sign_message(Message::new(data)).await
```
rymnc, line 75: perhaps we should add a check here for the length of the message?

crates/types/src/signer.rs
```diff
@@ -96,6 +97,31 @@ impl SignMode {
...
+            #[cfg(feature = "aws-kms")]
+            SignMode::Kms {
+                cached_public_key_bytes,
+                ..
+            } => {
+                use k256::pkcs8::DecodePublicKey;
+
+                let k256_public_key =
+                    k256::PublicKey::from_public_key_der(cached_public_key_bytes)?;
+                Ok(Some(k256_public_key.into()))
```
rymnc, line 120: q: why don't we just set the public key into the enum variant instead of its raw form?  we won't have to recalculate it when `verifying_key()` is called every time.

as i understand the code, `make_payload()` is leading to this function being called repeatedly

crates/services/shared-sequencer/src/http_api.rs
xgreenx: The current implementation doesn't care if we are not able to submit the blob. We just *try* to do that, and if not, then it's okay.

In the future(follow-up issue), we need to worry about it and guarantee that all blobs are posted and there are no gaps.

crates/services/shared-sequencer/src/lib.rs
xgreenx: What is the `client_id`? Do you mean `chain_id`? I want to keep the current implementation as simple, as possible. 

Plus, if the RPC is malicious, it can lie about its `chain_id`

crates/types/src/signer.rs
xgreenx: It is more question to @Dentosal =) Performance is not critical in `make_payload` because we call this function twice each 12 seconds

crates/services/shared-sequencer/src/service.rs
```diff
@@ -0,0 +1,266 @@
...
+    shared_sequencer::{
+        SSBlob,
+        SSBlobs,
+    },
+};
+use futures::StreamExt;
+use std::sync::Arc;
+
+/// Non-initialized shared sequencer task.
+pub struct NonInitializedTask<S> {
```
netrome, line 33: Nit: I suggest renaming this to `UninitializedTask` to maintain parity with https://github.com/FuelLabs/fuel-core/pull/2442

crates/services/shared-sequencer/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+        let ss_config = http_api::config(&endpoints.blockchain_rest_api).await?;
+
+        let mut minimum_gas_price = ss_config.minimum_gas_price;
+
+        if let Some(index) = minimum_gas_price.find('.') {
+            minimum_gas_price.truncate(index);
+        }
+        let gas_price: u128 = minimum_gas_price.parse()?;
+        // Ceil the gas price to the next integer.
+        let gas_price = gas_price.saturating_add(1);
```
netrome, line 77: What's the purpose of incrementing the gas price here? The comment says we're ceiling it, but it's already parsed as a u128 so that's not correct. If we get a float number in `minimum_gas_price.parse()`, we'll get an error, not a floored value.

crates/services/shared-sequencer/src/service.rs
```diff
@@ -0,0 +1,266 @@
...
+
+        tokio::select! {
+            biased;
+            _ = watcher.while_started() => {
+                Ok(STOP)
+            },
+
+            // The `blobs` function calls sleep inside and if it not cancel safe.
+            // If someone add new logic into the `tokio::select`, please
+            // rework the `blobs` function to be cancel safe and use interval inside.
```
netrome, line 207:
```suggestion
            // The `blobs` function is not cancel safe, as it calls sleep inside.
            // If someone add new logic into the `tokio::select`, please
            // rework the `blobs` function to be cancel safe and use interval inside.
```

bin/fuel-core/src/cli/run/shared_sequencer.rs
```diff
@@ -0,0 +1,56 @@
...
+            }
+        };
+
+        Ok(fuel_core_shared_sequencer::Config {
+            enabled: val.enable,
+            block_posting_frequency: val.block_posting_frequency.into(),
+            endpoints,
+            topic: *val.topic,
+        })
+    }
```
netrome, line 55: I think we should allow the case where `enable` is false and the `tendermint_api` and `blockchain_api` are unset. That seems to be the purpose of having them be optional afaiu, otherwise we could just make `tendermint_api` and `blockchain_api` be required arguments and skip this match statement.

netrome (approved):
Overall looks good to me. I have a few suggestions and questions, but nothing blocking.

crates/services/shared-sequencer/src/lib.rs
rymnc: sorry, `chain_id` :)

rymnc (approved):

Issue #2391, opened 2024-10-24
Optimize GraphQL "coins to spend" query


rafal-ch:
The source of this issue is https://github.com/FuelLabs/fuel-core/issues/623.

Initially, the idea was to remove all complex queries from `fuel-core` and make it part of the indexer's responsibility. But the SDK is deeply integrated with these basic queries, that we need to support them on `fuel-core`. For that, we need to optimize the following query by aggregating some information in the off-chain worker or adding new indexes to the database:

- The `coins_to_spend` query is very complex and requires `n^2` operations where `n` is the number of coins and messages. The algorithm itself can use additional indexes from RocksDB to solve the issue with dust coins and works fast.

Extracted from the https://github.com/FuelLabs/fuel-core/issues/1965, which initially covered optimizations for 3 different areas: balances, coins to spend, dry run.

rafal-ch:
This issue will be delivered in at least 2 PRs:
1. Add ability to store coins to spend in an sorted way
1. Leverage the fact that coins are sorted in the selection algorithm

How to achieve #1:
1. Build the "reverse lookup index" which will use the RocksDB key sorting capabilities
   1. RocksDB uses lexicographical sort
   2. To maintain the ordering, the integer representing the amount will be converted to big-endian bytes
1. To solve the conflicts, for example, two coins with same value...


**To be continued...**

rafal-ch:
Current flow:
1. `coinsToSpend()` arrives at the GraphQL API
2. Coins are collected
   1. `off_chain` DB is queried for "owned coin ids"
      * these are taken from `OwnedCoins` storage
   1. coin exclusion filter is applied
   1. coins are read from `on_chain` DB (`coins_iter()`)
      * these are taken from `Coins` storage
6. if querying for `base_asset_id`, the above is repeated for "messages":
   1. `off_chain` DB is queried for "owned message ids"
      * these are taken from `OwnedMessageIds` storage
   1. messages are read from `on_chain` DB (`messages_iter()`)
      * these are taken from `Messages` storage
7. `random_improve()` algo is used to select coins
   * if it cannot satisfy the request, fallback to `largest_first()` algorithm

New flow:
1. Introduce `AvailableCoinsManager`
   * optionally, let it support plugable selection algo
1. Notify it about every change to `Messages` and `Coins` storage
   * `AvailableCoinsManager` will update the internal state in the `off_chain` DB indexation cache
   * this should be done on the same transaction for consistency
[...]

PR #2455, opened 2024-11-26
Create indexation cache for "coins to spend" queries

rafal-ch:
This is the 1/2 PR to fix the https://github.com/FuelLabs/fuel-core/issues/2391
It is stacked on top of https://github.com/FuelLabs/fuel-core/pull/2383 (balances cache)

## Description
The scope of this PR is to build the proper index upon processing the coin related events.
The follow-up PR will contain the actual usage of the index, hence there are a couple of TODOs left that mention this follow-up PR.

### Before requesting review


crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -143,7 +150,7 @@ pub struct ReadView {
     pub(crate) genesis_height: BlockHeight,
     pub(crate) on_chain: OnChainView,
     pub(crate) off_chain: OffChainView,
-    pub(crate) balances_enabled: bool,
+    pub(crate) balances_indexation_enabled: bool,
```
xgreenx, line 153: Maybe we also could add coins to spend here with `allow(dead_code)`?

crates/fuel-core/src/graphql_api/indexation/balances.rs
```diff
@@ -0,0 +1,592 @@
+use fuel_core_storage::StorageAsMut;
```
xgreenx, line 1: This file is almost the same as `crates/fuel-core/src/graphql_api/indexation.rs`(compared with the original file), except moving balance related errors to `error` module and small renaming.

Skipping this file from reviewing.

crates/fuel-core/src/graphql_api/indexation/error.rs
```diff
@@ -0,0 +1,204 @@
...
+    StorageError(StorageError),
+}
+
+#[cfg(test)]
+mod tests {
+    use super::IndexationError;
+
+    impl PartialEq for IndexationError {
+        fn eq(&self, other: &Self) -> bool {
+            match (self, other) {
```
xgreenx, line 98: For simplicity(because it is used only in tests) you can just convert `self` and `other` t strings and compare strings. It is how we do for other error in the tests=)

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -280,6 +280,11 @@ pub fn init_sub_services(
 
     let graphql_block_importer =
         GraphQLBlockImporter::new(importer_adapter.clone(), import_result_provider);
+    let base_asset_id = config
+        .snapshot_reader
+        .chain_config()
+        .consensus_parameters
+        .base_asset_id();
```
xgreenx, line 287:
```suggestion
    let base_asset_id = *chain_config.consensus_parameters. base_asset_id();
```

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -288,6 +293,7 @@ pub fn init_sub_services(
         chain_id,
         config.da_compression.clone(),
         config.continue_on_error,
+        *base_asset_id,
```
xgreenx, line 296:
```suggestion
        base_asset_id,
```

crates/fuel-core/src/query/coin.rs
```diff
@@ -63,4 +72,16 @@ impl ReadView {
...
             .try_flatten()
     }
+
+    pub fn coins_to_spend(
+        &self,
+        owner: &Address,
+        asset_id: &AssetId,
+        max: u16,
+    ) -> Result<Vec<CoinType>, CoinsQueryError> {
+        let _coin_ids = self.off_chain.coins_to_spend(owner, asset_id, max);
```
xgreenx, line 82: Do we need to call it, if below we have todo?

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -631,6 +676,7 @@ where
     }
 }
 
+#[allow(clippy::too_many_arguments)]
```
xgreenx, line 679: Maybe it makes sense to pass `&ConsensusParameters` since now we have 2 parameters from it, in this case you can remove `too_many_arguments`

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -23,6 +41,180 @@ pub fn owner_coin_id_key(owner: &Address, coin_id: &UtxoId) -> OwnedCoinKey {
...
+
+        let mut arr = [0; CoinsToSpendIndexKey::LEN];
+        let mut offset = 0;
+        arr[offset..offset + Address::LEN].copy_from_slice(address_bytes);
+        offset += Address::LEN;
+        arr[offset..offset + AssetId::LEN].copy_from_slice(asset_id_bytes);
+        offset += AssetId::LEN;
+        arr[offset..offset + u8::BITS as usize / 8].copy_from_slice(
+            if message.is_retryable_message() {
+                &RETRYABLE_BYTE
```
xgreenx, line 118: We should exclude retryable messages from the coins to spend because we don't want to suggest them to the user to spend.

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -23,6 +41,180 @@ pub fn owner_coin_id_key(owner: &Address, coin_id: &UtxoId) -> OwnedCoinKey {
...
+    }
+}
+
+impl CoinsToSpendIndexKey {
+    const LEN: usize = Address::LEN
+        + AssetId::LEN
+        + u8::BITS as usize / 8
+        + u64::BITS as usize / 8
+        + TxId::LEN
+        + 2;
```
xgreenx, line 80: I think it will be better if the type `u8::BITS as usize / 8` will be represented by `enum` with `repr(u8)`.

Instead of `u64::BITS as usize / 8` you can use `size_of::<Amount>`.

Instead of `TxId::LEN + 2` you could use `size_of::<UtxoId>`.

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -23,6 +41,180 @@ pub fn owner_coin_id_key(owner: &Address, coin_id: &UtxoId) -> OwnedCoinKey {
...
+        let amount_bytes = coin.amount.to_be_bytes();
+        let utxo_id_bytes = utxo_id_to_bytes(&coin.utxo_id);
+
+        let mut arr = [0; CoinsToSpendIndexKey::LEN];
+        let mut offset = 0;
+        arr[offset..offset + Address::LEN].copy_from_slice(address_bytes);
+        offset += Address::LEN;
+        arr[offset..offset + AssetId::LEN].copy_from_slice(asset_id_bytes);
+        offset += AssetId::LEN;
+        arr[offset..offset + u8::BITS as usize / 8].copy_from_slice(&NON_RETRYABLE_BYTE);
```
xgreenx, line 95: I think it will make more sense if the type of the coin `Message/Coin` will be after `amount`, before `Nonce/UtxoId`. In this case we also will sort messages by amount.

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -2,16 +2,34 @@ use fuel_core_storage::{
...
+        TxId,
+        UtxoId,
+    },
+    fuel_types::Nonce,
+};
+
+use crate::graphql_api::indexation;
+
+use self::indexation::coins_to_spend::{
+    NON_RETRYABLE_BYTE,
```
xgreenx, line 31: In another comment I mentioned, that we don't need to include retryable messages into the coins to spend query. So we can remove it.

But you need to know the difference between coin and message, so I think we need to use this 1 byte for the `Message` or `Coin` enum representation.

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -23,6 +41,180 @@ pub fn owner_coin_id_key(owner: &Address, coin_id: &UtxoId) -> OwnedCoinKey {
...
+            } else {
+                &NON_RETRYABLE_BYTE
+            },
+        );
+        offset += u8::BITS as usize / 8;
+        arr[offset..offset + u64::BITS as usize / 8].copy_from_slice(&amount_bytes);
+        offset += u64::BITS as usize / 8;
+        arr[offset..offset + Nonce::LEN].copy_from_slice(nonce_bytes);
+        offset += Nonce::LEN;
+        arr[offset..].copy_from_slice(&indexation::coins_to_spend::MESSAGE_PADDING_BYTES);
```
xgreenx, line 128: Hmm, how well RocksDB works with the dynamic sized keys? Maybe based on the type of the message/coin we could use 32/34 bytes for `Nonce`/`UtxoId` types and during decoding decide what type to return? 

Just a thought, if it is hard to support or implement, I'm okay with the current padding approach =)

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -23,6 +41,180 @@ pub fn owner_coin_id_key(owner: &Address, coin_id: &UtxoId) -> OwnedCoinKey {
...
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
+
+impl Mappable for CoinsToSpendIndex {
+    type Key = Self::OwnedKey;
+    type OwnedKey = CoinsToSpendIndexKey;
+    type Value = Self::OwnedValue;
+    type OwnedValue = u8;
```
xgreenx, line 206: Why do we need 1 byte here? If you don't use it, we can just use `()` type.

xgreenx: Ah, I see, it is `IndexedCoinType`. Why not to use this type here instead of `u8`? Also, in the comment above I said that maybe we could use `retryable` and `non-retryable` byte to track message/coin type.

xgreenx (commented):
I think it will be simpler to review if the second part was actually the first, and vice versa=)

Because the second part only requires you to have sorted backward and forward iterators. You could update the current algorithm to work with this iterator(you could use `iter` and `iter().rev()` from the vector) and, in the next PR, replace the sorted vector with the new indexation. In this case, we don't need to have `todo!` in the code and it is easier to review the final variant of the feature.

rafal-ch:
> I think it will be simpler to review if the second part was actually the first, and vice versa=)

Yes, that's true. Also because after I started implementing the actual usage of the index, I noticed that a couple of things implemented here need to be adjusted.

I'll prepare a single PR with the complete feature which will also include responses to the comments you placed here.

rafal-ch:
Closing now in favor of https://github.com/FuelLabs/fuel-core/pull/2463
All review comments added to this PR will be addressed and eventually incorporated into the new PR.

crates/fuel-core/src/graphql_api/database.rs
rafal-ch: It's now added and used in the [complete PR](https://github.com/FuelLabs/fuel-core/pull/2463).

crates/fuel-core/src/graphql_api/indexation/balances.rs
rafal-ch: Yes, even in the new PR, the logic around the "balances indexation" has not been changed. There are minor changes like variable renames.

crates/fuel-core/src/query/coin.rs
rafal-ch: This piece of code is now removed.

crates/fuel-core/src/graphql_api/storage/coins.rs
rafal-ch: Actually, in the complete PR I decided to remove the padding approach and use variable length keys. I don't know about any performance implications on RocksDB side. Also, we never query for a large amounts of data (255 items at most with the current limits), so we should be good. Taking this into consideration I though it's not worth "wasting" additional two bytes for _every_ indexed message.

crates/fuel-core/src/service/sub_services.rs
rafal-ch: Updated in a915976cc879aa058264fe0fd4a24168eddd8cac

crates/fuel-core/src/service/sub_services.rs
rafal-ch: Updated in https://github.com/FuelLabs/fuel-core/commit/a915976cc879aa058264fe0fd4a24168eddd8cac

crates/fuel-core/src/graphql_api/storage/coins.rs
rafal-ch: This assumes we change the structure of the index key, because with the current approach we cannot switch places since this byte is used as a part of prefix when querying the index. Let's huddle this out.

crates/fuel-core/src/graphql_api/storage/coins.rs
rafal-ch: Updated in d720d1db3569256c8b34e0fa1264701c8ac19290

crates/fuel-core/src/graphql_api/indexation/error.rs
rafal-ch: That's a good hint :+1: 
Updated in 4c55857c5df8d242a7e0bf66bae6bd1f1a631bdc

crates/fuel-core/src/graphql_api/worker_service.rs
rafal-ch: Updated in dd03f921f3aa7b2aa861627fa62dc9f6f5ee8401.
Soon we might need some _POD_ structure to carry these parameters.

Issue #2448, opened 2024-11-22
Make use of the "retryable" message balance


rafal-ch:
This PR added the indexation for balances. In particular, for messages it tracks the "retryable" and "non-retryable" balances distinctively.

However, the "retryable" balance is not exposed in any way. This issue is to design & implement the way we want to allow users to interact with the "retryable" balance.

AurelienFT:
@rafal-ch is it something we want to do in near future ?

rafal-ch:
@AurelienFT we'd need to discuss this with potential consumers of the feature (business, downstream apps?). I'll check this and send an update.

rafal-ch:
Not doing this since there's no demand at the moment.

Issue #2428, opened 2024-11-11
Handle the overflow in "balances" calculation.


rafal-ch:
_Originally posted by @rafal-ch in https://github.com/FuelLabs/fuel-core/pull/2383#discussion_r1823020725_
            

At the moment all "amounts" for coins, messages and contracts is stored as `u64`. When implementing the "[balances cache](https://github.com/FuelLabs/fuel-core/issues/1965)" I noticed that when we calculate the total balances, we use the accumulator which is `u64` and we call `saturating_add()` on it. This may lead to the balances being calculated incorrectly.

In the [original PR](https://github.com/FuelLabs/fuel-core/pull/2383) the accumulator has been changed to `u128`, which reduces, but not removes the problem. Also, now we calculate the balance using `checked_*()` functions, but in case of error, we just log it and continue.

To be considered:
1. Properly handle the overflow error 
  - maybe change the return type of "balances()" from `StorageResult` to something else, since overflow is not a storage error
  - update integration tests - some of them are written with the assumption that balances calculation is infallible
2. Consider changing all "amounts" to be `u128`, so that even a single coin can have a balance greater than `u64`

rafal-ch:
@xgreenx 

Maybe in case we detect an error in the index we should just bail with `?` so that the caller will treat these errors in the same way as it treats errors on `insert()`, `replace()`, etc.?

PR #2463, opened 2024-11-28
Use indexation cache to satisfy "coins to spend" queries

rafal-ch:
Closes https://github.com/FuelLabs/fuel-core/issues/2391

This PR includes all changes from the [Part 1 PR](https://github.com/FuelLabs/fuel-core/pull/2455), making it deprecated.

## Description
Changes in this PR:

#### The new `CoinsToSpend` index
* This is the database that stores all coins to spend sorted by the amounts (i.e. largest-by-value coins first)
* The key consists of several parts
  * _Retryable flag_ - to distinguish between retryable messages and other coins
  * _Address_ (owner)
  * _AssetID_
  * _Amount_ - as "big-endian" bytes to leverage the RocksDB key sorting capabilities
  * _Foreign Key_ - this are bytes of the key from either the `Messages` or `Coins` on-chain databases
    * for messages this is a 32-bytes `Nonce`
    * for coins this is a 34-bytes `UtxoId`
* The value is an instance of `IndexedCoinType` enum, so we know which on-chain database to query when returning the actual coins
* This index is updated when executor events are processed
* When querying for "coins to spend" the following algorithm is applied:
   * First, we get as many "big" coins as required to satisfy _double the amount_ from the query (respecting `max` and `excluded` params)
   * If we have enough coins, but there are still some "slots" in the query left (because we selected less coins than `max`) we fill the remaining slots with a **random** number of "dust" coins
   * If it happens that the value of selected "dust coins" is able to cover the value of some of the already selected "big coins", we remove the latter from the response
   * If at any step we encounter a problem (reading from database, integer conversions, etc.) we bail with an appropriate error

#### Changes to `CoinsQueryError` type
* The `MaxCoinsReached` variant has been removed because in the new algorithm we never query for more coins than the specified `max`, hence, without additional effort, we are not able to tell whether the query could be satisfied if user provided bigger `max`
* The `InsufficientCoins` has been renamed to `InsufficientCoinsForTheMax` and it now contains the additional `max` field

#### Off-chain database metadata
* The metadata for off-chain database now contain the additional `IndexationKind` - `CoinsToSpend`

#### Refactoring
* The `indexation.rs` module was split into separate files, each per indexation type + errors + some utils.

#### Other
* Integration tests have to be updated to not expect the exact number of coins to spend in the response (currently, due to randomness, we're not deterministic in this regard)
* The number of excluded ids in the `coinsToSpend` GraphQL query is now limited to the maximum number of inputs allowed in transaction.

### Before requesting review
- [X] I have created follow-up issues caused by this PR and linked them here

### Follow-up issues
* https://github.com/FuelLabs/fuel-core/issues/2498
* https://github.com/FuelLabs/fuel-core/issues/2448
* https://github.com/FuelLabs/fuel-core/issues/2428
* https://github.com/FuelLabs/fuel-core/issues/2499
* https://github.com/FuelLabs/fuel-core/issues/2496


crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -2,27 +2,198 @@ use fuel_core_storage::{
...
+    }
+}
+
+impl CoinsToSpendIndexKey {
+    pub fn from_coin(coin: &Coin) -> Self {
+        let address_bytes = coin.owner.as_ref();
+        let asset_id_bytes = coin.asset_id.as_ref();
+        let retryable_flag_bytes = NON_RETRYABLE_BYTE;
+        let amount_bytes = coin.amount.to_be_bytes();
+        let utxo_id_bytes = utxo_id_to_bytes(&coin.utxo_id);
```
rafal-ch, line 100: As per the huddle, the `retryable_flag` will be moved to the very beginning of the key to allow queries like:
- give me all "retryable messages", regardless of the owner

crates/fuel-core/src/schema/coins.rs
```diff
@@ -143,6 +157,49 @@ pub struct ExcludeInput {
...
+}
+
+impl CoinOrMessageIdBytes {
+    pub(crate) fn from_utxo_id(utxo_id: &fuel_tx::UtxoId) -> Self {
+        Self::Coin(utxo_id_to_bytes(utxo_id))
+    }
+
+    pub(crate) fn from_nonce(nonce: &fuel_types::Nonce) -> Self {
+        let mut arr = [0; MESSAGE_FOREIGN_KEY_LEN];
+        arr[0..32].copy_from_slice(nonce.as_ref());
```
xgreenx, line 181:
```suggestion
        arr.copy_from_slice(nonce.as_ref());
```

crates/fuel-core/src/schema/coins.rs
```diff
@@ -233,55 +298,126 @@ impl CoinQuery {
...
-                .map(|message| coins::CoinId::Message(message.into()));
-            utxos.chain(messages).collect()
-        });
-
-        let base_asset_id = params.base_asset_id();
-        let spend_query =
-            SpendQuery::new(owner, &query_per_asset, excluded_ids, *base_asset_id)?;
+        let read_view = ctx.read_view()?;
+        let indexation_available = read_view.coins_to_spend_indexation_enabled;
+        if indexation_available {
```
xgreenx, line 303: Maybe we can do the same as for balance where we have a separate function for old way of calculation and new way of calculation?=)

crates/fuel-core/src/schema/coins.rs
```diff
@@ -233,55 +298,126 @@ impl CoinQuery {
...
+                let max = asset
+                    .max
+                    .and_then(|max| u16::try_from(max.0).ok())
+                    .unwrap_or(max_input)
+                    .min(max_input);
+
+                let mut coins_per_asset = vec![];
+                for coin_or_message_id in read_view
+                    .off_chain
+                    .coins_to_spend(&owner, &asset_id, total_amount, max, &excluded)?
```
xgreenx, line 343: We still need to call `yield_each` to allow other tasks to work=)

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
```diff
@@ -282,6 +304,220 @@ impl OffChainDatabase for OffChainIterableKeyValueView {
...
+
+        let selected_iter = select_coins_to_spend(
+            big_first_iter,
+            dust_first_iter,
+            target_amount,
+            max_coins,
+            excluded_ids,
+        )?;
+
+        into_coin_id(selected_iter, max_coins as usize)
```
xgreenx, line 344: Do we want have this logic here on adapter level or in the `graph_ql` module? I think the database just can expose `self.iter_all_filtered::<CoinsToSpendIndex` functionality and the algorithm itself will leave in the `graph_ql` module, allowing us keep the business logic in the module instead of putting it to the `fuel-core`

crates/fuel-core/src/schema/coins.rs
rafal-ch: Updated in 3ee2a1a9e1ab78c3551b3ab9972b90c42b7e6105

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -2,27 +2,198 @@ use fuel_core_storage::{
...
+        let amount_start = offset;
+        let amount_end = amount_start + AMOUNT_SIZE;
+        u64::from_be_bytes(
+            self.0[amount_start..amount_end]
+                .try_into()
+                .expect("should have correct bytes"),
+        )
+    }
+
+    #[allow(clippy::arithmetic_side_effects)]
```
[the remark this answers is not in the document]
rafal-ch: As per the comment above.

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -2,27 +2,198 @@ use fuel_core_storage::{
...
+            .expect("should have correct bytes");
+        AssetId::new(asset_id)
+    }
+
+    pub fn retryable_flag(&self) -> u8 {
+        let offset = 0;
+        self.0[offset]
+    }
+
+    #[allow(clippy::arithmetic_side_effects)]
```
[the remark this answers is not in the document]
rafal-ch: Actually, clippy stopped complaining about this code, but I changed these bindings to consts anyway in this commit: 202dcdf782d400c595c8bbf6851ed0bbeb831f5f

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -2,27 +2,198 @@ use fuel_core_storage::{
...
+// when the key is malformed (incorrect length). This is a bit of a code smell, but it's
+// consistent with how the `double_key!` macro works. We should consider refactoring this
+// in the future.
+pub struct CoinsToSpendIndex;
+
+impl Mappable for CoinsToSpendIndex {
+    type Key = Self::OwnedKey;
+    type OwnedKey = CoinsToSpendIndexKey;
+    type Value = Self::OwnedValue;
+    type OwnedValue = u8;
```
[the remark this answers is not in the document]
rafal-ch: Updated in 5e6e6d32f1153af03d371068089e0d4be7adc6ce

crates/fuel-core/src/schema/coins.rs
rafal-ch: Updated in 8319abd36c40e20093da15c046a79f8cf72c25db

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
rafal-ch: Is this more or less what you had in mind? Commit: 77f8caf0d6e7839f73097ba09a215aedad0da09f

I decided to not expose `iter_all_filtered::<CoinsToSpendIndex, _>` directly, but keep some guardrails around prefix structure and iterator directions so that the caller doesn't have to worry about it.

crates/fuel-core/src/graphql_api/ports.rs
```diff
@@ -108,6 +113,15 @@ pub trait OffChainDatabase: Send + Sync {
         direction: IterDirection,
     ) -> BoxedIter<StorageResult<(TxPointer, TxId)>>;
 
+    fn coins_to_spend(
+        &self,
+        owner: &Address,
+        asset_id: &AssetId,
+        target_amount: u64,
+        max_coins: u16,
+        excluded_ids: &ExcludedKeysAsBytes,
```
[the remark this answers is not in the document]
rafal-ch: Currently we do not have limits on GraphQL side. And both legacy and new implementation iterate over `excluded_ids` so indeed this could potentially be problematic.

I'll added the limit, equal to max inputs in this commit: a5c80e171d62098e942cbee4b7440a0344222359

User will be presented with the following error:
image

crates/fuel-core/src/graphql_api/database.rs
```diff
@@ -88,8 +88,10 @@ pub struct ReadDatabase {
     on_chain: Box<dyn AtomicView<LatestView = OnChainView>>,
     /// The off-chain database view provider.
     off_chain: Box<dyn AtomicView<LatestView = OffChainView>>,
-    /// The flag that indicates whether the Balances cache table is enabled.
-    balances_enabled: bool,
+    /// The flag that indicates whether the Balances indexation is enabled.
+    balances_indexation_enabled: bool,
```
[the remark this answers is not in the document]
rafal-ch: My usual approach is _"generalize if you need more than 2"_. If we ever gonna add one more I will probably try to re-use the approach taken at the DB metadata level ([here](https://github.com/FuelLabs/fuel-core/pull/2463/files#diff-33f0910104b2e2e4219d73505f68c40b804e1bc540c9d8e3d774a8cd73ab82dcR83-R85)).

I'm not super strong at sticking to my approach, though, so happy to hear other opinions :)

Also, for this case, since we don't need to care that much about the storage, I'd probably not go with bitmask, but a regular `HashSet`.

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -8,21 +8,188 @@ use fuel_core_storage::{
...
+        const ADDRESS_START: usize = RETRYABLE_FLAG_SIZE;
+        const ADDRESS_END: usize = ADDRESS_START + Address::LEN;
+        let address: [u8; Address::LEN] = self.0[ADDRESS_START..ADDRESS_END]
+            .try_into()
+            .expect("should have correct bytes");
+        Address::new(address)
+    }
+
+    pub fn asset_id(&self) -> AssetId {
+        const OFFSET: usize = Address::LEN + RETRYABLE_FLAG_SIZE;
```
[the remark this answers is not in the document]
rafal-ch: You're right. The order got messed up when I implemented [this change](https://github.com/FuelLabs/fuel-core/pull/2463#discussion_r1867951068).

Now updated in c1fe4fd85331f7d468a90c6a999a4f149e90146c

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -61,4 +262,212 @@ mod test {
...
+        a.as_ref()
+            .iter()
+            .copied()
+            .chain(b.as_ref().iter().copied())
+            .collect::<Vec<_>>()
+            .try_into()
+            .expect("should have correct length")
+    }
+
+    #[test]
```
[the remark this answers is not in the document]
rafal-ch: Added in 6c3720976e6796dbd17c8bdf1f41ee1716a10c22
Some tests had to be modified, this shows the benefits of G/W/T :-)

I added G/W/T to all tests that I've added in this PR, except:
* the ones that were just moved around as a result of moving/renaming the files.
* `enough_coins_in_the_db_to_reach_target_u64_max_but_limit_is_zero` - which is just a rewrite of one of pre-existing tests

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2154](https://github.com/FuelLabs/fuel-core/pull/2154): Transaction graphql endpoints use `TransactionType` instead of `fuel_tx::Transaction`.
 - [2446](https://github.com/FuelLabs/fuel-core/pull/2446): Use graphiql instead of graphql-playground due to known vulnerability and stale development.
 - [2379](https://github.com/FuelLabs/fuel-core/issues/2379): Change `kv_store::Value` to be `Arc<[u8]>` instead of `Arc<Vec<u8>>`.
+- [2463](https://github.com/FuelLabs/fuel-core/pull/2463): 'CoinsQueryError::MaxCoinsReached` variant has been removed. The `InsufficientCoins` variant has been renamed to `InsufficientCoinsForTheMax` and it now contains the additional `max` field
```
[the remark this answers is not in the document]
rafal-ch: Mentioned in this commit: 9786f3e9fb5c0e02543ba603ffac41fd13863d16

I've also added a note about previously merged "balances" cache in 2cfafc2f0417a79e95a74a223e8f408fadf38e40

rafal-ch:
> * If it happens that the value of selected "dust coins" is able to cover the value of some of the already selected "big coins", we remove the latter from the response
> 
> 
> One potential issue with this approach is that you may be more likely to create even worse dust as change. The big idea with random-improve, which is what we did before, is that you select an amount higher than what's needed if you can, such that the change you end up with after the transaction is likely to be useful and not dust. If you only select just barely enough coins to cover the spending amount, then you are much more likely to create a dust change output.
> 
> In random-improve, it assumes that if transfers are recurrent, then you should target 2x the required amount from the coin selection. That way your change will be enough to cover a similarly sized transaction in the future, (such as gas fees for a dex trade which may not change very often)

Would it be ok then if we just double the amount from the request and use this as an input the new algorithm? Or maybe we need some heuristics? cc: @Voxelot @xgreenx

xgreenx:
I would suggest using a `2 * target_from_user` as our `target`. If the `amount` that we collected ` >= target_from_user` return it; otherwise, an error that insufficient funds.

rafal-ch:
> I would suggest using a `2 * target_from_user` as our `target`. If the `amount` that we collected ` >= target_from_user` return it; otherwise, insufficient funds error.

Implemented in https://github.com/FuelLabs/fuel-core/pull/2463/commits/88471eaa7eb7e7e7fd52d46251db4ccf5cf864c8

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
rafal-ch: Please note that I also updated location of the code in this commit: https://github.com/FuelLabs/fuel-core/pull/2463/commits/7459c3ce744496ca84263e2b01e9f4eacd33604f

crates/fuel-core/src/coins_query.rs
```diff
@@ -221,6 +289,160 @@ pub async fn random_improve(
...
+) -> Result<bool, CoinsQueryError> {
+    match coin_type {
+        IndexedCoinType::Coin => {
+            let foreign_key = CoinOrMessageIdBytes::Coin(
+                key.foreign_key_bytes()
+                    .as_slice()
+                    .try_into()
+                    .map_err(|_| CoinsQueryError::IncorrectCoinKeyInIndex)?,
+            );
+            Ok(excluded_ids.coins().contains(&foreign_key))
```
xgreenx, line 411: It can be expensive operation in the case of `Vec`, maybe we want to consider use `HashSet` instead

rafal-ch: Updated in 3cf743acbcc34a7bffaba0cafae00c54ef66542b.

Also in this commit (2e0b7fd619f8ebcd2a2f41e2c755ea4581a03038) I simplified the code to avoid some allocs/collects.

tests/tests/coins.rs
```diff
@@ -562,14 +604,14 @@ mod all_coins {
...
         assert_eq!(coins_per_asset.len(), 2);
-        assert_eq!(coins_per_asset[0].len(), 1);
+        assert!(coins_per_asset[0].len() >= 1);
         assert!(coins_per_asset[0].amount() >= 1);
-        assert_eq!(coins_per_asset[1].len(), 1);
+        assert!(coins_per_asset[1].len() >= 1);
         assert!(coins_per_asset[1].amount() >= 1);
     }
 
     async fn query_target_300(owner: Address, asset_id_b: AssetId) {
```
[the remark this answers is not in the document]
rafal-ch: Updated in ce637a8ad6f119f830096068c81ef547a5267525

tests/tests/coins.rs
```diff
@@ -549,7 +591,7 @@ mod all_coins {
     }
 
     async fn query_target_1(owner: Address, asset_id_b: AssetId) {
```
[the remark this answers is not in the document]
rafal-ch: Updated in ce637a8ad6f119f830096068c81ef547a5267525

tests/tests/coins.rs
```diff
@@ -158,14 +167,14 @@ mod coin {
...
         assert_eq!(coins_per_asset.len(), 2);
-        assert_eq!(coins_per_asset[0].len(), 1);
+        assert!(coins_per_asset[0].len() >= 1);
         assert!(coins_per_asset[0].amount() >= 1);
-        assert_eq!(coins_per_asset[1].len(), 1);
+        assert!(coins_per_asset[1].len() >= 1);
         assert!(coins_per_asset[1].amount() >= 1);
     }
 
     async fn query_target_300(owner: Address, asset_id_a: AssetId, asset_id_b: AssetId) {
```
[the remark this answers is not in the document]
rafal-ch: Updated in ce637a8ad6f119f830096068c81ef547a5267525

tests/tests/coins.rs
```diff
@@ -145,7 +154,7 @@ mod coin {
     }
 
     async fn query_target_1(owner: Address, asset_id_a: AssetId, asset_id_b: AssetId) {
```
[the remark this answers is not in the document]
rafal-ch: Updated in ce637a8ad6f119f830096068c81ef547a5267525

crates/fuel-core/src/schema/coins.rs
```diff
@@ -233,75 +298,187 @@ impl CoinQuery {
...
+            select_coins_to_spend(
+                db.off_chain.coins_to_spend_index(&owner, &asset_id),
+                total_amount,
+                max,
+                &excluded,
+                db.batch_size,
+            )
+            .await?,
+        )
+        .yield_each(db.batch_size);
```
xgreenx, line 366: The slowest operation is iteration over the storage. Because you already have `Vec<CoinsToSpendIndexEntry>` in the memory, you can just process everything together for one run. It should be very fast.

So we can remove `yield_each` and usage fo the stream here=)

crates/fuel-core/src/schema/coins.rs
```diff
@@ -233,75 +298,187 @@ impl CoinQuery {
...
+                }
+            };
+
+            coins_per_asset.push(coin_type);
+        }
+
+        if coins_per_asset.is_empty() {
+            return Err(CoinsQueryError::InsufficientCoinsForTheMax {
+                asset_id,
+                collected_amount: total_amount,
```
xgreenx, line 387: It looks like `collected_amount` should be zero in this case=)

crates/fuel-core/src/schema/coins.rs
```diff
@@ -233,75 +298,187 @@ impl CoinQuery {
...
+                    let message = db.message(&nonce)?;
+                    let message_coin: message_coin::MessageCoin = message.try_into()?;
+                    CoinType::MessageCoin(message_coin.into())
+                }
+            };
+
+            coins_per_asset.push(coin_type);
+        }
+
+        if coins_per_asset.is_empty() {
```
xgreenx, line 384: Hmm, I see that we rely several time on empty vector in the case of error inside of the `select_coins_to_spend`. But maybe it will be more clear if `select_coins_to_spend` returned an error and we shouldn't have empty `coins_per_asset` case.

crates/fuel-core/src/schema/coins.rs
```diff
@@ -233,75 +298,187 @@ impl CoinQuery {
...
+                max,
+                &excluded,
+                db.batch_size,
+            )
+            .await?,
+        )
+        .yield_each(db.batch_size);
+
+        let mut coins_per_asset = vec![];
+        for coin_or_message_id in into_coin_id(selected_stream, max as usize).await? {
```
xgreenx, line 369: `max` here is used only to do `Vec::with_capacity(max_coins)`. It will be more clear to not do `with_capacity`, or pass the `len` of the `selected_stream`.

Taking into account another my comment to remove usage of the stream from the result of the `select_coins_to_spend`, you can just use `selected_stream.len()` because it will be a vector=)

crates/fuel-core/src/schema/coins.rs
```diff
@@ -233,75 +298,187 @@ impl CoinQuery {
...
+        all_coins.push(coins_per_asset);
     }
+    Ok(all_coins)
 }
 
-impl From<MessageCoinModel> for MessageCoin {
-    fn from(value: MessageCoinModel) -> Self {
-        MessageCoin(value)
-    }
+async fn coins_to_spend_without_cache(
```
xgreenx, line 397: I reviewed this function, almost the same as before, with no changes to the logic, skipping from the review.

crates/fuel-core/src/schema/coins.rs
```diff
@@ -233,75 +298,187 @@ impl CoinQuery {
...
+                    .foreign_key_bytes()
+                    .try_into()
+                    .map_err(|_| CoinsQueryError::IncorrectCoinKeyInIndex)?;
+
+                let (tx_id_bytes, output_index_bytes) = bytes.split_at(TxId::LEN);
+                let tx_id = TxId::try_from(tx_id_bytes).map_err(StorageError::from)?;
+                let output_index = u16::from_be_bytes(
+                    output_index_bytes.try_into().map_err(StorageError::from)?,
+                );
+                CoinId::Utxo(fuel_tx::UtxoId::new(tx_id, output_index))
```
xgreenx, line 470: It would be nice if this conversion logic is a part of the `CoinsToSpendIndexKey`. Either during decoding from the storage you already can decode it and just use the nice typed object(you can check how it was done via `Manual<OwnedTransactionIndexKey>`). Or you can support something like `foreign_key.try_into_utxo_id()?`.

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -8,21 +8,221 @@ use fuel_core_storage::{
...
+
+impl Mappable for CoinsToSpendIndex {
+    type Key = Self::OwnedKey;
+    type OwnedKey = CoinsToSpendIndexKey;
+    type Value = Self::OwnedValue;
+    type OwnedValue = IndexedCoinType;
+}
+
+impl TableWithBlueprint for CoinsToSpendIndex {
+    type Blueprint = Plain<Raw, Raw>;
```
xgreenx, line 64: nit: We can embed all conversations into the codec like we did with `Manual<OwnedTransactionIndexKey>`. In this case, from the usage perspective, you will always work with nice types with fields instead of vectors or slices.

crates/fuel-core/src/graphql_api/ports.rs
```diff
@@ -64,7 +64,20 @@ use fuel_core_types::{
...
+    coins::{
+        CoinsToSpendIndexKey,
+        IndexedCoinType,
+    },
+};
+
+pub struct CoinsToSpendIndexIter<'a> {
+    pub(crate) big_coins_iter:
+        BoxedIter<'a, Result<(CoinsToSpendIndexKey, IndexedCoinType), StorageError>>,
+    pub(crate) dust_coins_iter:
```
xgreenx, line 78: In the future we plan to move GraphQL into its own service, so it will be nice either use `pub` or define a structure that can be used outside of the crate(like add constructor, getter, `into_inner` for example) =)

crates/fuel-core/src/coins_query.rs
```diff
@@ -221,6 +274,168 @@ pub async fn random_improve(
...
+        big_coins_iter,
+        dust_coins_iter,
+    }: CoinsToSpendIndexIter<'_>,
+    total: u64,
+    max: u16,
+    excluded_ids: &ExcludedCoinIds<'_>,
+    batch_size: usize,
+) -> Result<Vec<CoinsToSpendIndexEntry>, CoinsQueryError> {
+    const TOTAL_AMOUNT_ADJUSTMENT_FACTOR: u64 = 2;
+    if total == 0 && max == 0 {
```
xgreenx, line 288: Based on the current logic, it will return a strange error to the user. Maybe we want to return more user-friendly error here=) Or an empty result, but `coins_per_asset.is_empty()` will not allow it in the current implementation

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
```diff
@@ -282,6 +287,33 @@ impl OffChainDatabase for OffChainIterableKeyValueView {
                 .into_boxed()
         }
     }
+
+    fn coins_to_spend_index(
+        &self,
+        owner: &Address,
+        asset_id: &AssetId,
+    ) -> CoinsToSpendIndexIter {
+        let prefix: Vec<_> = NON_RETRYABLE_BYTE
```
xgreenx, line 296: Maybe you want to return an error if indexation is disabled=)

crates/fuel-core/src/coins_query.rs
```diff
@@ -221,6 +274,168 @@ pub async fn random_improve(
...
+            count = count.saturating_add(1);
+            let amount = coin.0.amount();
+            coins_total_value = coins_total_value.saturating_add(amount);
+            coins.push(coin);
+        }
+    }
+    Ok((coins_total_value, coins))
+}
+
+fn is_excluded(
```
xgreenx, line 384: It would be nice if codec handled this kind of conversions and we keep the business logic without noise=)

crates/fuel-core/src/coins_query.rs
```diff
@@ -221,6 +274,168 @@ pub async fn random_improve(
...
+where
+    F: Fn(&CoinsToSpendIndexEntry, u64) -> bool,
+{
+    let mut coins_total_value: u64 = 0;
+    let mut count = 0;
+    let mut coins = Vec::with_capacity(max as usize);
+    while let Some(coin) = coins_stream.next().await {
+        let coin = coin?;
+        if !is_excluded(&coin, excluded_ids)? {
+            if count >= max || predicate(&coin, coins_total_value) {
```
xgreenx, line 372:
```suggestion
            if coins.len() >= max || predicate(&coin, coins_total_value) {
```

I think we can remove `count`=)

crates/fuel-core/src/coins_query.rs
```diff
@@ -221,6 +274,168 @@ pub async fn random_improve(
...
+
+    let big_coins_stream = futures::stream::iter(big_coins_iter).yield_each(batch_size);
+    let dust_coins_stream = futures::stream::iter(dust_coins_iter).yield_each(batch_size);
+
+    let (selected_big_coins_total, selected_big_coins) =
+        big_coins(big_coins_stream, adjusted_total, max, excluded_ids).await?;
+
+    if selected_big_coins_total < total {
+        return Ok(vec![]);
+    }
```
xgreenx, line 302: It would be nice to return an error here=)

crates/fuel-core/src/coins_query.rs
```diff
@@ -221,6 +274,168 @@ pub async fn random_improve(
...
+    let Some(last_selected_big_coin) = selected_big_coins.last() else {
+        // Should never happen.
+        return Ok(vec![]);
+    };
```
xgreenx, line 306: t would be nice to return an error here=)

crates/fuel-core/src/coins_query.rs
```diff
@@ -221,6 +274,168 @@ pub async fn random_improve(
...
+) -> impl Iterator<Item = CoinsToSpendIndexEntry> {
+    big_coins.into_iter().skip_while(move |item| {
+        let amount = item.0.amount();
+        dust_coins_total
+            .checked_sub(amount)
+            .map(|new_value| {
+                dust_coins_total = new_value;
+                true
+            })
+            .unwrap_or_default()
```
xgreenx, line 435:
```suggestion
            .unwrap_or(false)
```

crates/fuel-core/src/coins_query.rs
```diff
@@ -221,6 +274,168 @@ pub async fn random_improve(
...
+fn skip_big_coins_up_to_amount(
+    big_coins: impl IntoIterator<Item = CoinsToSpendIndexEntry>,
+    mut dust_coins_total: u64,
+) -> impl Iterator<Item = CoinsToSpendIndexEntry> {
+    big_coins.into_iter().skip_while(move |item| {
+        let amount = item.0.amount();
+        dust_coins_total
+            .checked_sub(amount)
+            .map(|new_value| {
+                dust_coins_total = new_value;
```
xgreenx, line 432:
```suggestion
fn skip_big_coins_up_to_amount<'a>(
    big_coins: impl IntoIterator<Item = CoinsToSpendIndexEntry> + 'a,
    dust_coins_total: &'a mut u64,
) -> impl Iterator<Item = CoinsToSpendIndexEntry> + 'a {
    big_coins.into_iter().skip_while(move |item| {
        let amount = item.0.amount();
        dust_coins_total
            .checked_sub(amount)
            .map(|new_value| {
                *dust_coins_total = new_value;
```

Can we use reference here?=D I don't trust closures with `mut` and `move` =D Sorry for being paranoic=)

xgreenx (commented):
Looks nice to me=D Left some comments to improve the errors and the code maintainance

crates/fuel-core/src/coins_query.rs
rafal-ch: Actually, our intention here is _not_ to modify the value outside of this function (on the callsite). The `mut` in `mut dust_coins_total: u64,` is just a shortcut allowing to use this binding as a counter meaningful only within the closure.

I modified the code to show the intent more explicitly in this commit: c001fd32613b4258ec119ee2ebe1722a629da925

crates/fuel-core/src/coins_query.rs
rafal-ch: The empty vector will be converted to `CoinsQueryError::InsufficientCoinsForTheMax` on the callsite, as you noticed here: https://github.com/FuelLabs/fuel-core/pull/2463/files#r1881319248

It's not the best solution, but returning this error from within the `fn select_coins_to_spend()` function would require the function to know the `asset_id` and currently it's asset agnostic (I'd like to keep it that way).

I was considering adding a new error variant (for example `CoinSelectionError`) and convert it to proper `CoinsQueryError` on the callsite, but this complicates stuff.

I think I'll probably replace `vec![]` with `None` and see how it fits the design.

crates/fuel-core/src/schema/coins.rs
rafal-ch: That's a good point. See the explanation here: https://github.com/FuelLabs/fuel-core/pull/2463#discussion_r1882006095

crates/fuel-core/src/coins_query.rs
rafal-ch: Updated in 0175e415c2e54fd44c10c0a3df7cc4eacdbf82ce

crates/fuel-core/src/coins_query.rs
rafal-ch: Added a proper error and some more explanation in 87245d406d2aaefba627d9f9345b98ae68d2a696

crates/fuel-core/src/coins_query.rs
rafal-ch: A proper error has been added in 2c4006939925e227771ee2a2bd800bbf3bc1be5c

rafal-ch: Also, a little fix to the `if` together with improved error messages has been added in bef1548de9e7dffbdeac4cd0b19eb420baa6a56f

crates/fuel-core/src/coins_query.rs
rafal-ch: I use the approach when I return `None` and construct the proper error with all necessary data upstream - commit: 2c4006939925e227771ee2a2bd800bbf3bc1be5c

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
rafal-ch: Yeah, that could come in handy, we just don't have this information available here. We would need to query metadata for every request or stick a boolean flag in, hmm... "IterableKeyValueView<ColumnType<OffChain>>"?

Am I maybe missing something?

xgreenx: It is true that we need to request it. If each request will do that, it should be fast because the result will be cached. But maybe we want to keep this information inside of the `GenericDatabase` idk and make this check fast=) It can be follow up PR and there we can decide how better to do it

crates/fuel-core/src/coins_query.rs
xgreenx: You can get `AssetId` from the `last_selected_big_coin` which you have below=)

crates/fuel-core/src/coins_query.rs
xgreenx: I was not sure that it works, because I don't know where `dust_coins_total` lives while iterator is alive. But seems it is "Rust's magic"=) Because it works in playground

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a01bef2288c09f61ac6691bff996144a

image

crates/fuel-core/src/coins_query.rs
rafal-ch: Yes, well spotted. Removed in 0602da1a562278603c47443b35cb5d09101f74fd

crates/fuel-core/src/coins_query.rs
rafal-ch: That won't work in case we don't select any coins in `big_coins()`. That's also why I cannot get it from one of the `selected_big_coins`, at least not consistently.

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
rafal-ch: Hmm, maybe limit access on trait level? Something like `pub trait OffChainDatabaseWithCoinIndex: OffChainDatabase` - not sure how feasible this is. 

But yeah, I'll create a follow-up issue anyway, because it's not a critical flaw as long as we only call it from the context of `coins_to_spend_with_cache()`

tests/tests/coins.rs
```diff
@@ -233,7 +241,7 @@ mod coin {
         asset_id_a: AssetId,
         asset_id_b: AssetId,
     ) {
-        let context = setup(owner, asset_id_a, asset_id_b).await;
+        let (context, max_inputs) = setup(owner, asset_id_a, asset_id_b).await;
```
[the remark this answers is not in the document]
rafal-ch: Yes, that makes sense. I changed the `setup()` to accept the `ConsensusParameters` in this commit: 1645c14e8b580e496a8a7e812a8fd2ca93f58a66

crates/fuel-core/src/schema/coins.rs
```diff
@@ -233,75 +298,187 @@ impl CoinQuery {
...
+            )
+            .await
+        }
     }
 }
 
-impl From<CoinModel> for Coin {
-    fn from(value: CoinModel) -> Self {
-        Coin(value)
+async fn coins_to_spend_with_cache(
```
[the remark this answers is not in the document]
rafal-ch: Swapped in 9d526830358f2f60c2d5167d2252acd61f05e20d

crates/fuel-core/src/schema/coins.rs
rafal-ch: Updated in f1420e385039e1250be0176f245ca890cbc578e7

crates/fuel-core/src/schema/coins.rs
rafal-ch: The solution with `Manual` is nice, but tricky due to the variable length of the `CoinsToSpendIndexKey`. I would need to specify the `Encoder` associated type as `Vec<u8>`. However, I like the idea of `CoinsToSpendIndexKey` being a struct with specific types, instead of just bytes, but that will incur quite a big rewrite, because I didn't think about it from day 1 :)

I can create a follow-up good-first issue to improve this later.

For now I took the approach of `try_into...()` in this commit: bdc8ce233df9de9a611a00efc93bead03d7fcd20

crates/fuel-core/src/graphql_api/storage/coins.rs
rafal-ch: See here: https://github.com/FuelLabs/fuel-core/pull/2463#discussion_r1883811058

crates/fuel-core/src/coins_query.rs
rafal-ch: Done in 8e42b87767d9661cb6746dbd3deb24084c1e4408

crates/fuel-core/src/graphql_api/ports.rs
rafal-ch: I mad the fields `pub` in this commit: 253306085f404e17ae662f15c08a0d26caeaa4c1

crates/fuel-core/src/coins_query.rs
rafal-ch: I decided to just pass asset id to the function, so it can now return proper error (with correct `asset_id` and `collected_amount`). Commit: 6740e7d2707ab976b5595ed9487c9a5fa178b148

crates/fuel-core/src/schema/coins.rs
rafal-ch: This is now reworked and fixed in commit: 6740e7d2707ab976b5595ed9487c9a5fa178b148

crates/fuel-core/src/schema/coins.rs
rafal-ch: This is now reworked and fixed in commit: 6740e7d2707ab976b5595ed9487c9a5fa178b148

crates/fuel-core/src/schema/coins.rs
rafal-ch: Follow-up issue created: https://github.com/FuelLabs/fuel-core/issues/2498

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
rafal-ch: Follow-up issue: https://github.com/FuelLabs/fuel-core/issues/2499

crates/fuel-core/src/coins_query.rs
```diff
@@ -221,6 +276,168 @@ pub async fn random_improve(
...
+                .map_err(|_| CoinsQueryError::IncorrectMessageKeyInIndex)?;
+            Ok(excluded_ids.is_message_excluded(&nonce))
+        }
+    }
+}
+
+fn max_dust_count(max: u16, big_coins_len: u16) -> u16 {
+    let mut rng = rand::thread_rng();
+    rng.gen_range(0..=max.saturating_sub(big_coins_len))
+}
```
netrome, line 422: Nit: I prefer when utility functions like this one are pure. In my view, rngs should be treated as interaction points and isolated as much as possible from business logic.
```suggestion

fn max_dust_count<Rng: rand::RngCore>(max: u16, big_coins_len: u16, rng: &mut Rng) -> u16 {
    rng.gen_range(0..=max.saturating_sub(big_coins_len))
}
```

crates/fuel-core/src/graphql_api.rs
```diff
@@ -80,8 +80,8 @@ impl Default for Costs {
 }
 
 pub const DEFAULT_QUERY_COSTS: Costs = Costs {
-    // TODO: The cost of the `balance` and `balances` query should depend on the
-    //  `OffChainDatabase::balances_enabled` value. If additional indexation is enabled,
+    // TODO: The cost of the `balance`, `balances` and `coins_to_spend` query should depend on the
```
netrome, line 83: While the TODO comment is not introduced in this PR, I'd like to know if we have any issue for this that we can link?

crates/fuel-core/src/graphql_api/indexation/balances.rs
```diff
@@ -0,0 +1,598 @@
...
+    let storage = block_st_transaction.storage::<MessageBalances>();
+    let current_balance = storage.get(key)?.unwrap_or_default().into_owned();
+    let MessageBalance {
+        mut retryable,
+        mut non_retryable,
+    } = current_balance;
+    if message.is_retryable_message() {
+        retryable = retryable.saturating_add(message.amount() as u128);
+    } else {
+        non_retryable = non_retryable.saturating_add(message.amount() as u128);
```
netrome, line 39: Nit: Personally I try to avoid `as` conversions, since they are often fallible and can cause unexpected panics. I believe this one is safe, but I think it both feels and reads better to use `u128::from` instead, to clearly mark that this conversion is infallible

```suggestion
        retryable = retryable.saturating_add(u128::from(message.amount()));
    } else {
        non_retryable = non_retryable.saturating_add(u128::from(message.amount()));
```

crates/fuel-core/src/graphql_api/indexation/balances.rs
```diff
@@ -0,0 +1,598 @@
...
+        non_retryable,
+    } = storage.get(key)?.unwrap_or_default().into_owned();
+    let current_balance = if message.is_retryable_message() {
+        retryable
+    } else {
+        non_retryable
+    };
+
+    let new_amount = current_balance
+        .checked_sub(message.amount() as u128)
```
netrome, line 72: Same here
```suggestion
        .checked_sub(u128::from(message.amount()))
```

crates/fuel-core/src/graphql_api/indexation/balances.rs
```diff
@@ -0,0 +1,598 @@
...
+        .ok_or_else(|| IndexationError::MessageBalanceWouldUnderflow {
+            owner: *message.recipient(),
+            current_amount: current_balance,
+            requested_deduction: message.amount() as u128,
```
netrome, line 76:
```suggestion
            requested_deduction: u128::from(message.amount()),
```

crates/fuel-core/src/graphql_api/indexation/balances.rs
```diff
@@ -0,0 +1,598 @@
...
+    block_st_transaction: &mut T,
+    coin: &Coin,
+) -> Result<(), IndexationError>
+where
+    T: OffChainDatabaseTransaction,
+{
+    let key = CoinBalancesKey::new(&coin.owner, &coin.asset_id);
+    let storage = block_st_transaction.storage::<CoinBalances>();
+    let current_amount = storage.get(&key)?.unwrap_or_default().into_owned();
+    let new_amount = current_amount.saturating_add(coin.amount as u128);
```
netrome, line 107:
```suggestion
    let new_amount = current_amount.saturating_add(u128::from(coin.amount));
```

crates/fuel-core/src/graphql_api/indexation/balances.rs
```diff
@@ -0,0 +1,598 @@
...
+where
+    T: OffChainDatabaseTransaction,
+{
+    let key = CoinBalancesKey::new(&coin.owner, &coin.asset_id);
+    let storage = block_st_transaction.storage::<CoinBalances>();
+    let current_amount = storage.get(&key)?.unwrap_or_default().into_owned();
+
+    let new_amount =
+        current_amount
+            .checked_sub(coin.amount as u128)
```
netrome, line 128:
```suggestion
            .checked_sub(u128::from(coin.amount))
```

crates/fuel-core/src/graphql_api/indexation/balances.rs
```diff
@@ -0,0 +1,598 @@
...
+            .ok_or_else(|| IndexationError::CoinBalanceWouldUnderflow {
+                owner: coin.owner,
+                asset_id: coin.asset_id,
+                current_amount,
+                requested_deduction: coin.amount as u128,
```
netrome, line 133:
```suggestion
                requested_deduction: u128::from(coin.amount),
```

crates/fuel-core/src/graphql_api/indexation/coins_to_spend.rs
```diff
@@ -0,0 +1,739 @@
...
+
+        let asset_id_1 = AssetId::from([11; 32]);
+        let asset_id_2 = AssetId::from([12; 32]);
+
+        let coin_1 = make_coin(&owner_1, &asset_id_1, 100);
+        let coin_2 = make_coin(&owner_1, &asset_id_2, 200);
+        let message_1 = make_retryable_message(&owner_1, 300);
+        let message_2 = make_nonretryable_message(&owner_2, 400);
+
+        // TODO[RC]: No clone() required for coins? Double check the types used,
```
netrome, line 229: What does RC mean?

crates/fuel-core/src/graphql_api/indexation/coins_to_spend.rs
```diff
@@ -0,0 +1,739 @@
...
+        // maybe we want `MessageCoin` (which is Copy) for messages?
+        // 1) Currently we use the same types as embedded in the executor `Event`.
+        // 2) `MessageCoin` will refuse to construct itself from a `Message` if the data is empty
+        //      impl TryFrom<Message> for MessageCoin { ... if !data.is_empty() ... }
+        // Actually it shouldn't matter from the indexation perspective, as we just need
+        // to read data from the type and don't care about which data type we took it from.
```
netrome, line 235: This seems resolved, should we delete it?
```suggestion
```

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -8,21 +8,253 @@ use fuel_core_storage::{
...
+    }
+
+    pub fn amount(&self) -> u64 {
+        const OFFSET: usize = RETRYABLE_FLAG_SIZE + Address::LEN + AssetId::LEN;
+        const AMOUNT_START: usize = OFFSET;
+        const AMOUNT_END: usize = AMOUNT_START + AMOUNT_SIZE;
+        u64::from_be_bytes(
+            self.0[AMOUNT_START..AMOUNT_END]
+                .try_into()
+                .expect("should have correct bytes"),
```
netrome, line 235: Not super happy about the expects, but I see the comment. However, I'm also not a fan of using the slice operator as it may cause unexpected panics. I'd prefer if we use `.get().expect("...")` explicitly instead?

```suggestion
            self.0.get(AMOUNT_START..AMOUNT_END)
                .expect("should span the whole range from AMOUNT_START to AMOUNT_END")
                .try_into()
                .expect("should have correct bytes"),
```

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -8,21 +8,253 @@ use fuel_core_storage::{
...
+        )
+    }
+
+    pub fn foreign_key_bytes(&self) -> &[u8] {
+        const OFFSET: usize =
+            RETRYABLE_FLAG_SIZE + Address::LEN + AssetId::LEN + AMOUNT_SIZE;
+        &self.0[OFFSET..]
```
netrome, line 242: Same here

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -8,21 +8,253 @@ use fuel_core_storage::{
...
+        const ASSET_ID_END: usize = ASSET_ID_START + AssetId::LEN;
+        let asset_id: [u8; AssetId::LEN] = self.0[ASSET_ID_START..ASSET_ID_END]
+            .try_into()
+            .expect("should have correct bytes");
+        AssetId::new(asset_id)
+    }
+
+    pub fn retryable_flag(&self) -> u8 {
+        const OFFSET: usize = 0;
+        self.0[OFFSET]
```
netrome, line 225: And here

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -8,21 +8,253 @@ use fuel_core_storage::{
...
+    }
+
+    fn from_slice(slice: &[u8]) -> Self {
+        Self(slice.into())
+    }
+
+    pub fn owner(&self) -> Address {
+        const ADDRESS_START: usize = RETRYABLE_FLAG_SIZE;
+        const ADDRESS_END: usize = ADDRESS_START + Address::LEN;
+        let address: [u8; Address::LEN] = self.0[ADDRESS_START..ADDRESS_END]
```
netrome, line 207: And here

crates/fuel-core/src/graphql_api/storage/coins.rs
```diff
@@ -8,21 +8,253 @@ use fuel_core_storage::{
...
+            .try_into()
+            .expect("should have correct bytes");
+        Address::new(address)
+    }
+
+    pub fn asset_id(&self) -> AssetId {
+        const OFFSET: usize = RETRYABLE_FLAG_SIZE + Address::LEN;
+        const ASSET_ID_START: usize = OFFSET;
+        const ASSET_ID_END: usize = ASSET_ID_START + AssetId::LEN;
+        let asset_id: [u8; AssetId::LEN] = self.0[ASSET_ID_START..ASSET_ID_END]
```
netrome, line 217: And here

netrome (commented):
Nice stuff so far. I have only done a relatively shallow sweep and will get back again during the week for a deeper look. I have a few nits and suggestions so far, but nothing major.

I'm not a fan of the lack of explicit error handling in `crates/fuel-core/src/graphql_api/storage/coins.rs` but other than that I think the code looks generally really well-written.

crates/fuel-core/src/graphql_api/indexation/coins_to_spend.rs
rafal-ch: No, actually it's not solved. I mean, the code is most likely correct, but what caught my attention is that in the code which handles coins to spend:
1. coins are represented by [entities::coins::Coin](https://github.com/FuelLabs/fuel-core/blob/master/crates/types/src/entities/coins/coin.rs#L23) , which is `Copy`
2. messages are represented by [entities::relayer::Message](https://github.com/FuelLabs/fuel-core/blob/master/crates/types/src/entities/relayer/message.rs#L34) , which is not `Copy`

For a moment I thought that we should use [`MessageCoin`](https://github.com/FuelLabs/fuel-core/blob/master/crates/types/src/entities/coins/message_coin.rs#L15) for messages, but we cannot construct retryable message coin due to [this](https://github.com/FuelLabs/fuel-core/blob/master/crates/types/src/entities.rs#L27-L31).

So, it's not an issue with `Copy` vs `non-Copy`, but potentially with incorrect type being used (still unlikely).

cc: @xgreenx - I'd like to double check this with you. If the current code is ok, I'll just remove `TODO`.

crates/fuel-core/src/graphql_api.rs
rafal-ch: Yes, [issue](https://github.com/FuelLabs/fuel-core/issues/2496) has been created, but not linked. Updated in 65f701d7c05e0ddc26c8dc7adc97211bdfeff152

crates/fuel-core/src/graphql_api/indexation/balances.rs
rafal-ch: I'm generally ok with `as` conversions for infallible cases, but `u128::from()` is also good. I updated a couple of places in this commit: 5eb9bb06309ef064a281043fc32a8aaffc916989

crates/fuel-core/src/graphql_api/indexation/balances.rs
rafal-ch: Updated in 5eb9bb06309ef064a281043fc32a8aaffc916989

crates/fuel-core/src/graphql_api/indexation/balances.rs
rafal-ch: Updated in 5eb9bb06309ef064a281043fc32a8aaffc916989

crates/fuel-core/src/graphql_api/indexation/balances.rs
rafal-ch: Updated in 5eb9bb06309ef064a281043fc32a8aaffc916989

crates/fuel-core/src/graphql_api/indexation/balances.rs
rafal-ch: Updated in 5eb9bb06309ef064a281043fc32a8aaffc916989

crates/fuel-core/src/graphql_api/indexation/balances.rs
rafal-ch: Updated in 5eb9bb06309ef064a281043fc32a8aaffc916989

crates/fuel-core/src/graphql_api/storage/coins.rs
rafal-ch: Yeah, I wasn't super happy to so I converted the functions. Instead of panicking they will return `Option<T>`. Rest of the code is adjusted to handle the `None` case. Changes in https://github.com/FuelLabs/fuel-core/pull/2463/commits/792dc332f334c3e49b99f48b5c42633e185e4eb1

crates/fuel-core/src/graphql_api/storage/coins.rs
rafal-ch: See here: https://github.com/FuelLabs/fuel-core/pull/2463#discussion_r1889854801

crates/fuel-core/src/graphql_api/storage/coins.rs
rafal-ch: See here: https://github.com/FuelLabs/fuel-core/pull/2463#discussion_r1889854801

crates/fuel-core/src/graphql_api/storage/coins.rs
rafal-ch: See here: https://github.com/FuelLabs/fuel-core/pull/2463#discussion_r1889854801

crates/fuel-core/src/coins_query.rs
rafal-ch: In this case, threading the randomizer through the entire call-path was cumbersome. Also, it would need to live somewhere, probably in `FuelClient`.

As an alternative I was considering using some `OnceCell` or `OnceLock` approach, but this brought in boilerplate connected to synchronization (mutexes, thread_local!, et al.).

Since we don't really need any crypotgraphic-grade randomness here I decided to go with the simplest approach. Any suggestions are still appreciated, though.

netrome: Fair enough :+1:

crates/fuel-core/src/graphql_api.rs
netrome: Nice thanks!

crates/fuel-core/src/graphql_api/indexation/balances.rs
netrome: Thanks :pray:

crates/fuel-core/src/graphql_api/indexation/coins_to_spend.rs
netrome: Right, nice! I typically reference the issue ID I'm working on: `//TODO(<issue-number>)`. It also makes it fast to find todos relating to follow-ups :)

crates/fuel-core/src/graphql_api/storage/coins.rs
netrome: Nice, looks great! Thanks for updating 🙏

crates/fuel-core/src/schema/coins.rs
xgreenx: It affects encoding, so I think it is better to implement it now=) https://github.com/FuelLabs/fuel-core/pull/2529

crates/fuel-core/src/coins_query.rs
```diff
@@ -221,6 +279,222 @@ pub async fn random_improve(
...
+        big_coins_iter,
+        dust_coins_iter,
+    }: CoinsToSpendIndexIter<'_>,
+    total: u64,
+    max: u16,
+    asset_id: &AssetId,
+    excluded_ids: &ExcludedCoinIds<'_>,
+    batch_size: usize,
+) -> Result<Vec<CoinsToSpendIndexEntry>, CoinsQueryError> {
+    const TOTAL_AMOUNT_ADJUSTMENT_FACTOR: u64 = 2;
```
xgreenx, line 293: It would be nice to put a comment why we decided to do it like this=)

crates/fuel-core/src/graphql_api/indexation/coins_to_spend.rs
xgreenx: It is okay to use clone in tests=) I don't think that we need to add more specific type

tests/tests/coins.rs
netrome, line 1: Could we add an integration test that demonstrates the expected behavior of the algorithm? I.e. a test that sets up a wallet with a few big and small coins, makes a query that should be satisfied by adding X big coins and Y small ones for a few sensible scenarios.

netrome (changes requested):
Apologies for taking longer than expected. While most of this looks good to me, I have a few questions of some things that puzzle me. Moreover, there are two concrete things I'd like to see updated before approving:

First, we should update the description and name of the `random_improve` function. Second, I'd like to see an integration test that shows how this algorithm behaves end to end.

crates/fuel-core/src/coins_query.rs
```diff
@@ -221,6 +279,222 @@ pub async fn random_improve(
...
+        return Err(CoinsQueryError::IncorrectQueryParameters {
+            provided_total: total,
+            provided_max: max,
+        });
+    }
+
+    let adjusted_total = total.saturating_mul(TOTAL_AMOUNT_ADJUSTMENT_FACTOR);
+
+    let big_coins_stream = futures::stream::iter(big_coins_iter).yield_each(batch_size);
+    let dust_coins_stream = futures::stream::iter(dust_coins_iter).yield_each(batch_size);
```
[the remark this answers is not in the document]
rafal-ch: IIRC, max number of inputs value lives in the consensus parameters, so there could be other chains with much higher limits. Also Ignition could get higher limits some day. I'm not concerned with this code causing too big of a performance hit.

crates/fuel-core/src/graphql_api/indexation/coins_to_spend.rs
```diff
@@ -0,0 +1,739 @@
...
+        CoinsToSpendIndex,
+        CoinsToSpendIndexKey,
+        IndexedCoinType,
+    },
+};
+
+use super::error::IndexationError;
+
+// Indicates that a message is retryable.
+pub(crate) const RETRYABLE_BYTE: [u8; 1] = [0x00];
```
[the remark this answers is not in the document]
rafal-ch: Here's some more context about this feature: https://github.com/FuelLabs/fuel-core/pull/2383#discussion_r1871333982

Information about this is scarce, we might want to tag someone form the docs team to document this more thoroughly.

tests/tests/coins.rs
rafal-ch: The tests in `tests/tests/coins.rs` do exactly that, but on the `FuelClient` level. Also, we have a bunch of UTs to check if index behaves properly in `crates/fuel-core/src/graphql_api/indexation/coins_to_spend.rs`. And there are tests dedicated to querying the index in `crates/fuel-core/src/coins_query.rs` (see: `mod indexed_coins_to_spend`)

Plus, a couple of pre-existing integration tests call `coins_to_spend()` to select test coins (like `tests/tests/relayer.rs` or `bin/e2e-test-client/src/tests/script.rs`)

In short, since this PR doesn't change the top-level feature but it's more like an implementation detail (although quite a big detail) I thought it's ok to rely on the pre-existing tests that were used to verify the coins to spend algorithm in it's previous form.

The added value from the e2e test would be
1. test if the GraphQL query really reaches the client
2. test if offchain worker actually calls the code to update the index
3. data returned from the query are included in GraphQL response

Unless I'm missing something obvious, my conjecture is that implementing the selection scenarios in full e2e tests is a bit of an overkill.

Happy to hear some more opinions. Also, if you feel strongly about this, I'm happy to work on some e2e tests for that but preferably in a follow-up PR.

crates/fuel-core/src/coins_query.rs
rafal-ch: Added in d9d024bd16777cb2a9468fb616b273b15b3e1fbe

crates/fuel-core/src/coins_query.rs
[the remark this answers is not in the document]
rafal-ch: Comment updated in 2b3c4de5411938ab67c3bed5b50353d714d12e51

crates/fuel-core/src/coins_query.rs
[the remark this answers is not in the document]
netrome: Fair enough :+1:

crates/fuel-core/src/graphql_api/indexation/coins_to_spend.rs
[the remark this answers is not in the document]
netrome: Thanks for the pointer. It seems to be documented already. It's just that "retryable" is not an externally facing term, which I think is fine since it just feels like it causes confusion. https://github.com/FuelLabs/fuel-specs/pull/443/files

tests/tests/coins.rs
netrome: Fair enough. I would like to see a more illustrative test case for the expected behavior for the dust coins in the `coins_query` test suite but I understand that the non-determinism makes that hard, so we can leave that for now.

crates/fuel-core/src/coins_query.rs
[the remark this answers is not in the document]
netrome: Thanks for removing the comment. I think we should also rename the function, since I don't think we intend to improve it anymore right?

netrome (dismissed):
Thanks for updating. I would like to see the "random_improve" function renamed, perhaps just to "random" but I can live with the old name as well.

rafal-ch:
> Thanks for updating. I would like to see the "random_improve" function renamed, perhaps just to "random" but I can live with the old name as well.

My understanding was that the "random improve" is actually the name of the algorithm. The [linked article](https://github.com/FuelLabs/fuel-core/blob/master/crates/fuel-core/src/coins_query.rs#L160) literally mentions this name. Cardano seems to [be using](https://hackage.haskell.org/package/cardano-coin-selection-1.0.1/docs/Cardano-CoinSelection-Algorithm-RandomImprove.html) the exact same name as well.

crates/fuel-core/src/coins_query.rs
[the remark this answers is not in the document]
rafal-ch: Please see [here](https://github.com/FuelLabs/fuel-core/pull/2463#issuecomment-2579887684). In general, I'm ok with renaming this function, but I'm not sure if this is the way to go.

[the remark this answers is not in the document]
netrome: Oh I see. Makes sense.

crates/fuel-core/src/coins_query.rs
```diff
@@ -221,6 +272,227 @@ pub async fn random_improve(
...
+    total: u64,
+    max: u16,
+    asset_id: &AssetId,
+    excluded_ids: &ExcludedCoinIds<'_>,
+    batch_size: usize,
+) -> Result<Vec<CoinsToSpendIndexEntry>, CoinsQueryError> {
+    // We aim to reduce dust creation by targeting twice the required amount for selection,
+    // inspired by the random-improve approach. This increases the likelihood of generating
+    // useful change outputs for future transactions, minimizing unusable dust outputs.
+    // See also "let upper_target = target.saturating_mul(2);" in "fn random_improve()".
```
AurelienFT, line 289: Very interesting !

crates/fuel-core/src/graphql_api/indexation/coins_to_spend.rs
```diff
@@ -0,0 +1,767 @@
...
+
+fn add_coin<T>(block_st_transaction: &mut T, coin: &Coin) -> Result<(), IndexationError>
+where
+    T: OffChainDatabaseTransaction,
+{
+    let key = CoinsToSpendIndexKey::from_coin(coin);
+    let storage = block_st_transaction.storage::<CoinsToSpendIndex>();
+    let maybe_old_value = storage.replace(&key, &IndexedCoinType::Coin)?;
+    if maybe_old_value.is_some() {
+        return Err(IndexationError::CoinToSpendAlreadyIndexed {
```
AurelienFT, line 37: If we return an error here the coin will still be inserted in the transaction and not revert. If we drop the transaction in an outer function it might not be an issue but maybe the dev of the outer function isn't aware of this. Same for `add_messages`

rafal-ch: Yes, there is an entire issue that tracks the missing error handling for indexation: https://github.com/FuelLabs/fuel-core/issues/2428 (referenced [here](https://github.com/FuelLabs/fuel-core/pull/2463/files#diff-1ac4adf5631afae925791c3897c52c2fa6282bd70901fddf552e7e1f0492c447R225)).

It's a little bit tricky, since these errors should never happen in sunny-day flow. If they happen, it means that the index got corrupted and we should probably disable them and fallback to legacy handling of the queries. Such corruption is fatal and cannot be recovered from, so we'd need to investigate how it happened. We might also consider advising the admin (via log) to re-sync the node from scratch to re-enable indexing.

But yeah, this is to be discussed/designed in the follow-up issue.

AurelienFT (approved):

xgreenx (approved):

Issue #2454, opened 2024-11-24
Modify `MetadataStorage` impl to store `unrecorded_blocks` efficiently


MitchTurner:
The list can grow very large, so we need to find a data structure that doesn't store every value redundantly for each new block.

MitchTurner:
Another solution than keeping the `unrecorded_blocks` on the metadata, is moving it to an adapter for `AlgorithmUpdater` So it is generic for some `T: UnrecordedBlosksHistory`, then we don't need to support iteratng over all heights, and can just use KV lookups.

PR #2468, opened 2024-11-29
Add unrecorded blocks abstraction to gas price algo

MitchTurner:
## Linked Issues/PRs
Closes https://github.com/FuelLabs/fuel-core/issues/2454

## Description
This PR:

- Adds a trait to the V1 algorithm for storing unrecorded blocks
- Adds `TransactionableStorage` trait to gas price service to allow abstraction for atomic commits to storage (once per L2 block
- Separates `Get` and `Set` traits for service to specify which need a transaction to execute
- Add a `GetSequenceNumber` and `SetSequence` number concept for tracking the DA sequence number the DA service should start on at startup (I can't remember why this got included in this PR, but it felt logical at the time...)

We only need a key-value lookup for the unrecorded blocks. Instead of storing all the data in the metadata, we just moved it to its own abstraction that can be accessed directly from the algorithm. This will also help with historical view, since historical view doesn't have a way to iterate over all values for that height, just do KV lookups. We were iterating before this change.

Also added the `unrecorded_block_bytes` field to metadata so we can still track that in the case of historical view.

This exposed a problem with the design, which is that we need changes to the algorithm to be atomic per l2 block, which is how our Fuel Storage is designed. This led us to introducing a `TransactionableStorage` trait for the service, which can be handed to the algorithm updater before being "commit"ed.  The same applies to the `MetadataStorage`, so the service was redesigned slightly to accomodate this change.

## Checklist

### Before requesting review


crates/services/gas_price_service/src/v1/uninitialized_task/fuel_storage_unrecorded_blocks/storage.rs
```diff
@@ -0,0 +1,71 @@
...
+    PartialEq,
+    Eq,
+    enum_iterator::Sequence,
+    Hash,
+    num_enum::TryFromPrimitive,
+)]
+pub enum UnrecordedBlocksColumn {
+    Metadata = 0,
+    State = 1,
+}
```
MitchTurner, line 31: I remove this in my next PR and combine with the GasPriceColumn

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -46,12 +52,11 @@ use tokio::sync::broadcast::{
...
 };
 
-use crate::common::utils::Result as GasPriceResult;
-
 /// The service that updates the gas price algorithm.
-pub struct GasPriceServiceV1<L2, Metadata, DA>
+pub struct GasPriceServiceV1<L2, Metadata, DA, U>
 where
     DA: DaBlockCostsSource,
+    U: Clone,
```
AurelienFT, line 59: I don't understand why `U` is bound on `Clone` here and also why it's not bound on `UnrecordedBlocks`

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -59,6 +63,25 @@ impl AlgorithmV1 {
     }
 }
 
+pub type Height = u32;
```
acerone85, line 66:
```suggestion
pub struct Height(u32);
```

nit: I usually try do advocate for the `Newtype` pattern (see https://www.lurklurk.org/effective-rust/newtype.html).
In this case I think that we have this type defined somewhere else already https://github.com/FuelLabs/fuel-core/blob/773226cb326c2607d274717348ff50aab30b0828/crates/types/src/blockchain/primitives.rs#L111

crates/services/gas_price_service/src/common/fuel_core_storage_adapter.rs
```diff
@@ -43,17 +52,41 @@ use fuel_core_types::{
...
-impl<Storage> MetadataStorage for StructuredStorage<Storage>
+impl<Storage> SetMetadataStorage for StructuredStorage<Storage>
 where
     Storage: KeyValueInspect<Column = GasPriceColumn> + Modifiable,
     Storage: Send + Sync,
+{
+    fn set_metadata(&mut self, metadata: &UpdaterMetadata) -> GasPriceResult<()> {
+        let block_height = metadata.l2_block_height();
+        let mut tx = self.write_transaction();
+        tx.storage_as_mut::<GasPriceMetadata>()
```
AurelienFT, line 76: Why are we creating a tx for 1 insert ?

crates/services/gas_price_service/src/common/fuel_core_storage_adapter/storage.rs
```diff
@@ -68,3 +70,41 @@ impl TableWithBlueprint for GasPriceMetadata {
...
+    fn column() -> Self::Column {
+        GasPriceColumn::UnrecordedBlocks
+    }
+}
+
+pub struct SequenceNumberTable;
+
+impl Mappable for SequenceNumberTable {
+    type Key = Self::OwnedKey;
+    type OwnedKey = u32;
```
AurelienFT, line 98: Can we use BlockHeight type here to be more explicit ?

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -136,63 +139,79 @@ where
...
-        da_block_costs: DaBlockCosts,
-    ) -> anyhow::Result<()> {
-        self.algorithm_updater.update_da_record_data(
-            &da_block_costs.l2_blocks,
-            da_block_costs.blob_size_bytes,
-            da_block_costs.blob_cost_wei,
-        )?;
-
-        self.set_metadata().await?;
+        let metadata = self.algorithm_updater.clone().into();
```
AurelienFT, line 188: AlgorithmUpdater have `algorithm()` method to get an algorithm. I find it strange to not have a `metadata()` but a `into()` to get metadata

crates/services/gas_price_service/src/v1/da_source_service/dummy_costs.rs
```diff
@@ -34,4 +34,8 @@ impl DaBlockCostsSource for DummyDaBlockCosts {
             }
         }
     }
+
+    async fn set_last_value(&mut self, sequence_number: u32) -> DaBlockCostsResult<()> {
+        todo!()
```
rafal-ch, line 39: Seems like the intent here is not to add implementation but to show that this function should never be called in the test. If I got it right, maybe it'll be better to do `panic!("should not be called");`?

crates/services/gas_price_service/src/v1/da_source_service/dummy_costs.rs
```diff
@@ -34,4 +34,8 @@ impl DaBlockCostsSource for DummyDaBlockCosts {
             }
         }
     }
+
+    async fn set_last_value(&mut self, sequence_number: u32) -> DaBlockCostsResult<()> {
```
rafal-ch, line 38: Can `DummyDaBlockCosts` be behind the `test` flag?

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -32,94 +44,85 @@ use fuel_core_services::{
...
     /// The algorithm updater
     algorithm_updater: AlgorithmUpdaterV1,
     /// the da source adapter handle
     da_source_adapter_handle: DaSourceService<DA>,
     /// The da source channel
     da_source_channel: Receiver<DaBlockCosts>,
+    /// Buffer of block costs from the DA chain
+    da_block_costs_buffer: Vec<DaBlockCosts>,
+    /// Storage transaction provider for metadata and unrecorded blocks
+    pub(crate) storage_tx_provider: StorageTxProvider,
```
rafal-ch, line 78: Seems like this is `pub(crate)` only because of tests. Maybe we can add a test-only getter?

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -32,94 +44,85 @@ use fuel_core_services::{
...
         let da_source_channel =
             da_source_adapter_handle.shared_data().clone().subscribe();
         Self {
             shared_algo,
             l2_block_source,
-            metadata_storage,
             algorithm_updater,
             da_source_adapter_handle,
             da_source_channel,
+            da_block_costs_buffer: Vec::new(),
```
rafal-ch, line 124: We only drain this buffer upon receiving L2 block (and on shutdown). I think there's no risk of unbounded growth then, but leaving the comment so you can double check :)

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -136,63 +139,79 @@ where
...
-    async fn set_metadata(&mut self) -> anyhow::Result<()> {
-        let metadata: UpdaterMetadata = self.algorithm_updater.clone().into();
-        self.metadata_storage
-            .set_metadata(&metadata)
-            .map_err(|err| anyhow!(err))
-    }
-
-    async fn handle_normal_block(
-        &mut self,
+    async fn handle_normal_block<'a>(
```
rafal-ch, line 148: What does "normal" mean in this context? Maybe "non_genesis" would be more clear?

crates/services/gas_price_service/src/v1/uninitialized_task.rs
```diff
@@ -307,10 +374,12 @@ where
             block_gas_capacity,
             block_bytes,
             fee_wei.into(),
+            &mut tx,
         )?;
         let metadata: UpdaterMetadata = updater.clone().into();
```
rafal-ch, line 379: Not related to this PR, but I noticed that we clone the entire updater and then just pick up a couple of fields from the cloned one to construct metadata. I think we can consider cloning only the relevant fields, especially because we do it in a loop.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -269,6 +288,22 @@ impl L2ActivityTracker {
...
+    pub fn capped_activity_threshold(&self) -> u16 {
+        self.capped_activity_threshold
+    }
+
+    pub fn decrease_activity_threshold(&self) -> u16 {
+        self.decrease_activity_threshold
+    }
+
+    pub fn block_activity_threshold(&self) -> ClampedPercentage {
+        self.block_activity_threshold.clone()
```
rafal-ch, line 305: Consider enabling `Copy` for the `ClampedPercentage`

crates/services/gas_price_service/src/v1/da_source_service/block_committer_costs.rs
```diff
@@ -15,7 +15,7 @@ use serde::{
 };
 
 #[async_trait::async_trait]
-trait BlockCommitterApi: Send + Sync {
+pub trait BlockCommitterApi: Send + Sync {
```
rafal-ch, line 18: It looks like this change is not necessary.

crates/services/gas_price_service/src/common/fuel_core_storage_adapter.rs
```diff
@@ -67,20 +100,150 @@ where
...
+        let sequence_number = self
+            .storage::<SequenceNumberTable>()
+            .get(block_height)
+            .map_err(|err| GasPriceError::CouldNotFetchDARecord(err.into()))?
+            .map(|no| *no);
+        Ok(sequence_number)
+    }
+}
+
+pub struct NewStorageTransaction<'a, Storage> {
```
rafal-ch, line 122: We need this wrapper purely to dodge the orphan-rule, right?

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -136,63 +139,79 @@ where
...
         block_bytes: u64,
         block_fees: u64,
-    ) -> anyhow::Result<()> {
-        let capacity = self.validate_block_gas_capacity(block_gas_capacity)?;
+    ) -> anyhow::Result<()>
+    where
+        StorageTxProvider::Transaction<'a>:
+            UnrecordedBlocks + SetMetadataStorage + SetDaSequenceNumber,
+    {
+        let mut storage_tx = self.storage_tx_provider.begin_transaction()?;
```
rafal-ch, line 160: We might want to start the transaction after we validate the block below. The validation itself is super-cheap, but still it may fail and we'll create the transaction unnecessarily.

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -212,26 +231,19 @@ where
...
 
 #[async_trait]
-impl<L2, Metadata, DA> RunnableTask for GasPriceServiceV1<L2, Metadata, DA>
+impl<L2, DA, StorageTxProvider> RunnableTask
+    for GasPriceServiceV1<L2, DA, StorageTxProvider>
 where
     L2: L2BlockSource,
-    Metadata: MetadataStorage,
     DA: DaBlockCostsSource,
+    StorageTxProvider: TransactionableStorage + 'static,
```
rafal-ch, line 244: Probably we don't need `'static`

rafal-ch (commented):
So far looks good, but I didn't finish the review yet, sorry. I'll do another pass ASAP. Currently leaving just a few minor comments.

crates/services/gas_price_service/src/v1/da_source_service/dummy_costs.rs
MitchTurner: I changed it to 
```rs
        unimplemented!("This is a dummy implementation");
```

crates/services/gas_price_service/src/v1/service.rs
MitchTurner: Yeah. Should be good. If we aren't hitting L2 blocks to drain this, then we have a much bigger problem than memory growth.

crates/services/gas_price_service/src/v1/service.rs
MitchTurner: As opposed to genesis block. I know it's not a great name, but it's true that normally it will be this block. Haha. I'd be open to other names. "non-genesis" seems strange.

crates/services/gas_price_service/src/common/fuel_core_storage_adapter.rs
MitchTurner: Exactly. Because the `UnrecordedBlocks` trait is in a separate crate. I think we could consider moving all the gas price algo stuff into the service, but until then we need a newtype.

crates/services/gas_price_service/src/v1/service.rs
MitchTurner: Nice catch. There might be other places too since I kinda spread the love around in an earlier iteration.

crates/services/gas_price_service/src/v1/service.rs
rafal-ch: Maybe just "handle_block()"?

rafal-ch:
Generic comment, probably not related to this PR.

It seems like some functions do not need to be `async`, for example: `SharedGasPriceAlgo::update()`. As a consequence, we might also remove `async` from some of the callers. We could revisit that, unless there is a plan for async processing in follow-up PRs.

crates/services/gas_price_service/src/common/fuel_core_storage_adapter.rs
MitchTurner: I feel like that's what we do most places.

crates/fuel-gas-price-algorithm/src/v1.rs
MitchTurner: Yeah. I don't want to use foreign types here. I want to keep the domain independent.

I'm not completely sold on using newtypes for situations like this. I mostly just added this for readability on the `BTreeMap` generics, otherwise people would just be guessing what the `u32` and `u64` represent. But the newtype would add an additional construct or `From` conversion for the caller, an that's noisy IMO.

crates/services/gas_price_service/src/common/fuel_core_storage_adapter.rs
MitchTurner: If we constrained with `KeyValueMutate` we could use `put` directly, but that's not what we have access to.

crates/services/gas_price_service/src/v1/uninitialized_task.rs
```diff
@@ -151,58 +168,62 @@ where
         // TODO: Add to config
         // https://github.com/FuelLabs/fuel-core/issues/2140
         let poll_interval = None;
+        if let Some(sequence_number) = self
+            .gas_price_db
+            .get_sequence_number(&metadata_height.into())?
+        {
+            self.da_source.set_last_value(sequence_number).await?;
+        }
```
xgreenx, line 176: I don't understand how this part should work. You only set the sequencer number when you receive the bundle from the DA. 

image

Let's say you received the bundle with sequencer number `10` at L2 block height X. After that, you processed the L2 block at height X + 1. After that, the node was restarted.

During start up the `metadata_height` will be `X + 1`, but the `get_sequence_number` will return `None` in this case. So you will not continue from the sequencer number `10`.

crates/services/gas_price_service/src/v1/da_source_service.rs
```diff
@@ -2,11 +2,13 @@ use crate::v1::da_source_service::service::DaBlockCostsSource;
 use std::time::Duration;
 
 pub mod block_committer_costs;
+#[cfg(test)]
 pub mod dummy_costs;
 pub mod service;
 
 #[derive(Debug, Default, Clone, Eq, Hash, PartialEq)]
 pub struct DaBlockCosts {
+    pub sequence_number: u32,
```
xgreenx, line 11: It would be better if it was `bundle_id` or if we somehow connected it to the block committer(comment can help)=)

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -192,7 +197,13 @@ where
     ) -> anyhow::Result<()> {
         match l2_block {
             BlockInfo::GenesisBlock => {
-                self.set_metadata().await?;
+                let metadata: UpdaterMetadata = self.algorithm_updater.clone().into();
+                let mut tx = self.storage_tx_provider.begin_transaction()?;
+                tx.set_metadata(&metadata).map_err(|err| anyhow!(err))?;
+                AtomicStorage::commit_transaction(tx)?;
+                let new_algo = self.algorithm_updater.algorithm();
+                // self.update(new_algo).await;
```
xgreenx, line 205:
```suggestion
```

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -158,31 +153,41 @@ where
         block_bytes: u64,
         block_fees: u64,
     ) -> anyhow::Result<()> {
-        let capacity = self.validate_block_gas_capacity(block_gas_capacity)?;
+        let capacity = Self::validate_block_gas_capacity(block_gas_capacity)?;
+        let mut storage_tx = self.storage_tx_provider.begin_transaction()?;
+
+        for da_block_costs in self.da_block_costs_buffer.drain(..) {
```
xgreenx, line 159: In the case of an error, you will lose `da_block_costs_buffer` data. Maybe it is better just to iterate here and clear it later after you committed changes to the database.

xgreenx (commented):
Almost done=) Some questions and suggestions

crates/services/gas_price_service/src/v1/uninitialized_task.rs
MitchTurner: Ah. I see. So we need to set the most recent value again if it hasn't changed.

xgreenx (dismissed):

crates/fuel-gas-price-algorithm/src/v1/tests/update_da_record_data_tests.rs
```diff
@@ -405,14 +366,15 @@ fn update_da_record_data__da_block_lowers_da_gas_price() {
...
     let l2_block_height = 11;
     let original_known_total_cost = 150;
-    let unrecorded_blocks = vec![BlockBytes {
-        height: 11,
-        block_bytes: 3000,
-    }];
+    // let unrecorded_blocks = vec![BlockBytes {
+    //     height: 11,
+    //     block_bytes: 3000,
+    // }];
```
netrome, line 372:
```suggestion
```

crates/services/gas_price_service/src/common/fuel_core_storage_adapter/storage.rs
```diff
@@ -68,3 +70,46 @@ impl TableWithBlueprint for GasPriceMetadata {
...
+
+    fn column() -> Self::Column {
+        GasPriceColumn::UnrecordedBlocks
+    }
+}
+
+pub struct SequenceNumberTable;
+
+/// The sequence number or bundle id of the posted blocks.
+type BundleId = u32;
```
netrome, line 99: Why do we use both "Bundle ID" and "Sequence number" to refer to the same thing here? I'd prefer it if we can be consistent and just refer to it as "Bundle ID".

crates/services/gas_price_service/src/v1/uninitialized_task/fuel_storage_unrecorded_blocks.rs
```diff
@@ -0,0 +1,123 @@
...
+    v1::UnrecordedBlocks,
+};
+
+pub trait AsUnrecordedBlocks {
+    type Wrapper<'a>: UnrecordedBlocks
+    where
+        Self: 'a;
+
+    fn as_unrecorded_blocks(&mut self) -> Self::Wrapper<'_>;
+}
```
netrome, line 31: Nice wrapper!

netrome (commented):
Overall looks good to me. There's some stale commented out code that needs to be cleaned up. Beyond that I have some questions and considerations, but nothing blocking, so if you apply the latest changes from master and clean up the stale comment I'll be good to approve.

crates/services/gas_price_service/src/common/fuel_core_storage_adapter/storage.rs
MitchTurner: I've changed them all to `BundleId` and `bundle_id`.

crates/services/gas_price_service/src/v1/uninitialized_task/fuel_storage_unrecorded_blocks.rs
MitchTurner: It was @xgreenx 's I had a newtype to solve the orphan rule problem before and Green switch it to this in his suggestions.

crates/fuel-gas-price-algorithm/src/v1.rs
```diff
@@ -59,6 +63,27 @@ impl AlgorithmV1 {
...
+pub type Height = u32;
+pub type Bytes = u64;
+
+pub trait UnrecordedBlocks {
+    fn insert(&mut self, height: Height, bytes: Bytes) -> Result<(), String>;
+
+    fn remove(&mut self, height: &Height) -> Result<Option<Bytes>, String>;
+}
+
+impl UnrecordedBlocks for BTreeMap<Height, Bytes> {
```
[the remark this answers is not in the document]
MitchTurner: It's mainly a holdover from before. It used to be that we'd iterate over them in order, but that requirement was removed. It's still likely that they'll be accessed in order, but `HashMap` probably works just fine here.

crates/services/gas_price_service/src/common/fuel_core_storage_adapter/storage.rs
netrome: Thanks! :pray:

crates/fuel-gas-price-algorithm/src/v1.rs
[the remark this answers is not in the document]
netrome: Fair enough :+1: I also think `BTreeMap` feels better since the data is ordered, but just looking at the access pattern I can't really motivate it :sweat_smile:

netrome (approved):
Thanks for the update :pray:

PR #2472, opened 2024-12-05
Revert balances amount to `U64` and introduce new `amountU128` getter

rafal-ch:
## Description
This PR partially reverts the changes introduced in the [balances indexation PR](https://github.com/FuelLabs/fuel-core/pull/2383).

In particular:
* reverts the type of `amount` in the `Balance` GraphQL type back to `U64`
* introduces a new field `amountU128` in the `Balance` GraphQL type
* reverts the `FuelClient::balance()` function to return `u64`. There is no `balance_u128()` function added to the `FuelClient` for now.

image

This is done to maintain the full backward compatibility on the GraphQL level, internally the total balance is still processed using `u128`.

### Before requesting review


AurelienFT (approved):
We don't want to start adding a deprecation notice to the first one ?

xgreenx (approved):

rafal-ch:
> We don't want to start adding a deprecation notice to the first one ?

Tbh, I'm not sure if we want to deprecate this.

PR #2473, opened 2024-12-05
RPC Consistency Proposal

### LINKED ISSUE #1897: RPC Consistency Proposal
https://www.notion.so/fuellabs/RPC-Consistency-Proposal-c8442b5d8f4c492abab08449cb43b4f2?pvs=4

acerone85:
## Linked Issues/PRs

Related issue:  #1897 
Closes https://github.com/FuelLabs/fuel-core/issues/2605

See https://www.notion.so/fuellabs/RPC-Consistency-Proposal-V2-13b2f2293f31809bbce0d93a4c28d633 for the proposal specs.

This PR implements a variation of the proposal in the link, due to techincal limitations of the HTTP protocol when using graphql subscriptions. 

for context: Headers cannot be used, the reason being that subscriptions will return multiple graphql responses in a single HTTP response (and therefore have a single `HeaderMap`), while we require that each item in the response has its associated `CURRENT_FUEL_BLOCK_HEIGHT` field.  Moreover the Response metadata for subscriptions is returned before the actual response body (as it is standard with HTTP).

in the end graphql has a (very undocumented) way to add an `extensions` field to responses, at the same level of `data/errors` field. This does not require to tweak anything at the HTTP level.

We will implement an improvement where the server node will wait for a configurable number of blocks before returning a `PRECONDITION FAILED` response, in a follow-up PR.

Entrypoint for reviews: 
- https://github.com/FuelLabs/fuel-core/pull/2473/files#diff-9caf873d3167ab9485e7882d621b5147e98b2478cc6462095b88374e2f914f3aR428 for the graphql handler
- https://github.com/FuelLabs/fuel-core/pull/2473/files#diff-9caf873d3167ab9485e7882d621b5147e98b2478cc6462095b88374e2f914f3aR428 for the extension definition

## TODO: 

- [x] Tests 

## Checklist
- [ ] ~Breaking changes are clearly marked as such in the PR description and changelog~

### Before requesting review
- [ ] ~I have created follow-up issues caused by this PR and linked them here~

- [ ] Someone else?


crates/client/src/client.rs
```diff
@@ -240,6 +240,28 @@ impl FuelClient {
         Self::decode_response(response)
     }
 
+    pub async fn query_with_headers<ResponseData, Vars>(
+        &self,
+        q: Operation<ResponseData, Vars>,
+        headers: impl IntoIterator<Item = (String, String)>,
```
acerone85, line 246: TODO: HeadersMap here?

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -100,11 +100,18 @@ use futures::{
...
+        atomic::AtomicU32,
+        OnceLock,
+    },
 };
 
 #[cfg(test)]
 mod tests;
 
+// The last known block height that was processed by the GraphQL service.
+pub static LAST_KNOWN_BLOCK_HEIGHT: OnceLock<AtomicU32> = OnceLock::new();
```
acerone85, line 113: todo: remove Static and store in Service? Should still be accessible from middleware

crates/client/src/client.rs
netrome: I assume this one https://docs.rs/http/1.2.0/http/header/struct.HeaderMap.html - makes sense :+1:

crates/client/src/client.rs
```diff
@@ -1080,6 +1102,32 @@ impl FuelClient {
...
+        let asset_id: schema::AssetId = match asset_id {
+            Some(asset_id) => (*asset_id).into(),
+            None => schema::AssetId::default(),
+        };
+        let query = schema::balance::BalanceQuery::build(BalanceArgs { owner, asset_id });
+        let balance: types::Balance = self
+            .query_with_headers(
+                query,
+                vec![(
+                    "REQUIRED_FUEL_BLOCK_HEIGHT".to_string(),
```
netrome, line 1121: Hmm, I'm not 100% satisfied with this name. Especially since we provide a block height as input to some queries. When squinting your eyes it could be confusing if the "required_bock_height" is part of the query, when it is more a notion of the state of the node to ensure we're in sync.

How would `REQUIRED_CHAIN_TIP_HEIGHT` sound to you? Here it's clear that we're referring to the block height of the chain tip.

crates/client/src/client.rs
```diff
@@ -1080,6 +1102,32 @@ impl FuelClient {
         Ok(balance.amount)
     }
 
+    pub async fn balance_with_required_block_header(
+        &self,
+        owner: &Address,
+        asset_id: Option<&AssetId>,
+        required_block_height: u32,
+    ) -> io::Result<u128> {
```
netrome, line 1110: I'm a bit hesitant to have the `required_block_height` be a parameter here. It feels like we're cluttering the interface and are increasing the cognitive load for our users. This feels more like a state that should be maintained by the client and not even directly exposed to the user. 

I'd imagine the client maintaining a `chain_tip` parameter locally which is updated whenever it receives a response, and it's appended to the header of all requests made.

What are your thoughts on this?

crates/fuel-core/src/graphql_api/worker_service.rs
netrome: Is this so performance critical that we can't do a database lookup for this value? I'd default to querying this from the database, to ensure we don't have any weird consistency issues when this field and the db isn't in sync. I guess this isn't a big problem right now, but it just doesn't feel right to allow for inconsistencies in a field used to synchronize with other nodes.

crates/fuel-core/src/graphql_api/api_service.rs
```diff
@@ -214,6 +226,66 @@ impl RunnableTask for Task {
...
+const REQUIRED_FUEL_BLOCK_HEIGHT_HEADER: &str = "REQUIRED_FUEL_BLOCK_HEIGHT";
+const CURRENT_FUEL_BLOCK_HEIGHT_HEADER: &str = "CURRENT_FUEL_BLOCK_HEIGHT";
+
+async fn required_fuel_block_height<B>(
+    req: AxumRequest<B>,
+    next: Next<B>,
+) -> impl IntoResponse {
+    let last_known_block_height: BlockHeight = LAST_KNOWN_BLOCK_HEIGHT
+    .get()
+    //Maybe too strict?
```
netrome, line 238: I find it weird that this can be unset. We could always default to 0 as well, which would just mean that we're not in sync yet (which I presume we're not if this value hasn't been set).

crates/fuel-core/src/graphql_api/api_service.rs
```diff
@@ -42,6 +44,13 @@ use axum::{
             ACCESS_CONTROL_ALLOW_ORIGIN,
         },
         HeaderValue,
+        Request as AxumRequest,
+        Response as AxumResponse,
```
netrome, line 48: Nit: This is a re-export so I'd alias it to `HttpRequest`, or even better just import `axum::http` and keep the namespacing: `http::Request`.

crates/fuel-core/src/graphql_api/api_service.rs
```diff
@@ -214,6 +226,66 @@ impl RunnableTask for Task {
     }
 }
 
+const REQUIRED_FUEL_BLOCK_HEIGHT_HEADER: &str = "REQUIRED_FUEL_BLOCK_HEIGHT";
+const CURRENT_FUEL_BLOCK_HEIGHT_HEADER: &str = "CURRENT_FUEL_BLOCK_HEIGHT";
+
+async fn required_fuel_block_height<B>(
+    req: AxumRequest<B>,
+    next: Next<B>,
+) -> impl IntoResponse {
```
netrome, line 235: Nit: I prefer more informative names for generics if they don't have any clear trait bounds:
```suggestion
async fn required_fuel_block_height<Body>(
    req: AxumRequest<Body>,
    next: Next<Body>,
) -> impl IntoResponse {
```

netrome (approved):
Nice stuff! I have some thoughts and comments, but no showstoppers.

crates/fuel-core/src/graphql_api/worker_service.rs
xgreenx: We shouldn't use static. Each extension has access to the context and you can get `ReadDatabase` from the context. `ReadDatabase` knows the latest block height.


You can check how `ViewExtension` works

acerone85: I agree that we should fetch the value from the DB (it is probably going to be fetched from the memtable anyway, since it's frequently updated). I don't agree that we should do this with `async_graphql` Extensions, the reason being that support for headers is pretty limited in there. 

1. Headers do not seem to be set when building the query context in `async_graphql` (see https://github.com/async-graphql/async-graphql/blob/7f1791488463d4e9c5adcd543962173e2f6cbd34/src/schema.rs#L927). 
I have confirmed this by applying the patch below to my PR and running the test. The output confirms that the header is not set by the time we execute the query.
```diff
diff --git a/crates/fuel-core/src/schema/balance.rs b/crates/fuel-core/src/schema/balance.rs
index b6b95228e8..956d5961bb 100644
--- a/crates/fuel-core/src/schema/balance.rs
+++ b/crates/fuel-core/src/schema/balance.rs
@@ -60,6 +60,10 @@ impl BalanceQuery {
         #[graphql(desc = "address of the owner")] owner: Address,
         #[graphql(desc = "asset_id of the coin")] asset_id: AssetId,
     ) -> async_graphql::Result<Balance> {
+        println!(
+            "header set: {}",
+            ctx.http_header_contains("REQUIRED_FUEL_BLOCK_HEIGHT")
+        );
         let query = ctx.read_view()?;
         let base_asset_id = *ctx
             .data_unchecked::<ConsensusProvider>()
```
2. Maybe we could get the header somehow from the ExtensionContext, but this is going to be an ad hoc, undocumented solution and likely not readable. 

What I think we should do instead is stick with Axum's middleware: get a handle to the OnchainDatabase from the CombinedDatabase (just because I don't want to wrap the instance to the CombinedDatabase inside an Arc), and use it to fetch the value of the last height in the middleware. 

Let me know what you think.

crates/client/src/client.rs
acerone85: I agree that we could have a better name (maybe `CHAIN_HEAD` instead of `CHAIN_TIP`) , but if we change it we have to notify other teams as well as we are deviating from the spec: see for example https://github.com/FuelLabs/fuels-ts/issues/3443

crates/client/src/client.rs
acerone85: If I understood correctly you suggest to change the client to have the headers it will send in the requests in its state. I think that makes a lot of sense, let me have a go :)

crates/fuel-core/src/graphql_api/api_service.rs
acerone85: It will be unset until the first block is processed. But if we are going to fetch the value directly from the DB then it won't be the case anymore. I am waiting to hear what approach @xgreenx prefers before refactoring this bit

crates/fuel-core/src/graphql_api/api_service.rs
acerone85: agreed and done: 90c411f54c

crates/client/src/client.rs
acerone85: Forgot to mention this is done, see d9091e6465f140db830bd5f626dc4535a8ba93d0

crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
```diff
@@ -0,0 +1,149 @@
...
+                    path: vec![],
+                    extensions: None,
+                });
+            }
+        }
+
+        next.run(ctx, request).await
+    }
+
+    async fn execute(
```
acerone85, line 119: TODO: Might be better to forget about setting the response header here and just do it in the axum handler (we have the value stored in the Arc)

crates/fuel-core/src/graphql_api/api_service.rs
```diff
@@ -346,11 +361,73 @@ async fn health() -> Json<serde_json::Value> {
...
+            let current_fuel_block_height = current_fuel_block_height_data.lock().await;
+
+            header_map.insert(
+                CURRENT_FUEL_BLOCK_HEIGHT_HEADER,
+                (*current_fuel_block_height.expect("Block height is set")).into(),
+            );
+        }
+        Err((
+            StatusCode::PRECONDITION_FAILED,
+            header_map,
```
acerone85, line 425: todo: need to test that this does not get wiped out when processing the response upstream

crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
```diff
@@ -0,0 +1,149 @@
...
+            .expect("Data to store current fuel block height was set when forming the request")
+            .lock()
+            .await;
+
+            *current_block_height = Some(latest_known_block_height);
+        }
+
+        if let Some(required_fuel_block_height) = required_fuel_block_height_header_value
+        {
+            if required_fuel_block_height > latest_known_block_height {
```
acerone85, line 105: todo: implement From<RequiredFuelBlockHeightTooFarInTheFuture> for ServerError

crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
```diff
@@ -0,0 +1,149 @@
...
+            // See https://github.com/async-graphql/async-graphql/blob/7f1791488463d4e9c5adcd543962173e2f6cbd34/src/schema.rs#L521
+            // We need to fetch the mutable location to store the current fuel block height
+            // directly from request.data
+            let mut current_block_height = request
+            .data
+            .get(&TypeId::of::<Arc<Mutex<Option<BlockHeight>>>>())
+            .and_then(|data| { println!("Request was here"); data.downcast_ref::<Arc<Mutex<Option<BlockHeight>>>>()})
+            .expect("Data to store current fuel block height was set when forming the request")
+            .lock()
+            .await;
```
acerone85, line 98: todo: isolate to helper methods

crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
```diff
@@ -0,0 +1,149 @@
...
+                    // Safety: The value was constructed
+                    BlockHeight::new(
+                        number
+                            .as_u64()
+                            .and_then(|n| n.try_into().ok())
+                            .expect("The REQUIRED_FUEL_BLOCK_HEIGHT_HEADER value has been constructed from a u64 value"),
+                    )
+                }
+                _ => panic!("The REQUIRED_FUEL_BLOCK_HEIGHT_HEADER value has been constructed from a u64 value"),
+            });
```
acerone85, line 71: todo: isolate to helper function

Torres-ssf:
Hi @acerone85,

Could you kindly confirm if the response header for subscriptions is also intended to include `current_fuel_block_height`?

acerone85:
> Hi @acerone85,
> 
> Could you kindly confirm if the response header for subscriptions is also intended to include `current_fuel_block_height`?

In the current implementation there is no response header for the current_fuel_block_height. Would it be desirable to have one?

crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
acerone85: edit: I decided not to, as the final code seemed to be less readable than the current one

crates/client/src/client.rs
```diff
@@ -230,9 +270,12 @@ impl FuelClient {
         Vars: serde::Serialize,
         ResponseData: serde::de::DeserializeOwned + 'static,
     {
-        let response = self
-            .client
-            .post(self.url.clone())
+        let mut request_builder = self.client.post(self.url.clone());
+        for (header_name, header_value) in self.headers.iter() {
+            request_builder = request_builder.header(header_name, header_value);
+        }
```
rafal-ch, line 276: Can we use `headers()` to avoid manual loop, like:
```
    request_builder = request_builder.headers(&self.headers);
```

crates/fuel-core/src/graphql_api/api_service.rs
```diff
@@ -274,6 +288,9 @@ where
         ))
         .extension(async_graphql::extensions::Tracing)
         .extension(ViewExtension::new())
+        // `RequiredFuelBlockHeightExtension`` uses the view set by the ViewExtension.
```
rafal-ch, line 291: A nit:

```suggestion
        // `RequiredFuelBlockHeightExtension` uses the view set by the ViewExtension.
```

crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
```diff
@@ -0,0 +1,112 @@
...
+
+impl ExtensionFactory for RequiredFuelBlockHeightExtension {
+    fn create(&self) -> Arc<dyn Extension> {
+        Arc::new(RequiredFuelBlockHeightExtension::new())
+    }
+}
+
+/// Error value returned by the extension when the required fuel block height
+/// precondition is not met.
+pub(crate) struct RequiredFuelBlockHeightTooFarInTheFuture;
```
rafal-ch, line 48: If this is supposed to be used as an error you might consider implementing `Display` and `std::Error` for it.

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -105,6 +105,8 @@ use std::{
 #[cfg(test)]
 mod tests;
 
+// The last known block height that was processed by the GraphQL service.
+
```
rafal-ch, line 109: Is this a stray comment?

Torres-ssf:
> 
> In the current implementation there is no response header for the current_fuel_block_height. Would it be desirable to have one?

@acerone85 I believe it would because I have seen the following scenario happening a few times when submitting TXs sequentially:          

```mermaid
flowchart TD
    %% Transaction n1
    A1[Start TX n1] --> A2[Query coinsToSpend to fetch resources]
    A2 --> A3[Submit TX n1 using submitAndAwaitStatus]
    A3 --> A4[Await status update]
    A4 --> A5[Success]
    A5 --> B1[Start TX n2]

%% Transaction n+1
    B1 --> B2[Query coinsToSpend]
    B2 --> B3[Query hits unsync node, fetching a spent UTXO on TX n1]
    B3 --> B4[Submit TX n2 using submitAndAwaitStatus]
    B4 --> B5[Sync node validates UTXO was spent]
    B5 --> B6[Error]
```

crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
acerone85: not an error, it is used upstream to populate the `CURRENT_FUEL_BLOCK_HEIGHT` header in case the precondition is not met.

rafal-ch: Ok, thanks for explanation. I got confused by the comment above :-)

crates/fuel-core/src/graphql_api/api_service.rs
netrome: Thanks!

crates/fuel-core/src/graphql_api/worker_service.rs
netrome: Sorry completely lost this discussion, I think this makes sense. I guess you've already implemented this. Re-reviewing now...

crates/client/src/client.rs
netrome: Fair enough

crates/client/src/client.rs
netrome: Nice, happy it worked out :)

netrome (dismissed):
Looks good to me. The current solution pushes more responsibility on the caller for constructing and maintaining the headers in the client. While this is fine, I wonder if we should take more responsibility for this particular header to relieve users from having to think about it.

I'd imagine `FuelClient` to just hold a `required_fuel_block_height` parameter which it maintains internally (it could be exposed to users as well), and uses to construct the header when making requests. When receiving responses it would update the `required_fuel_block_height` with the returned block height for subsequent requests.

crates/client/src/client.rs
netrome: Commented in my PR. What I originally imagined was just having a parameter maintained by the client. Now it seems like we require the users to set the headers manually in the client which is a bit more cognitive load on them. I'm fine with that too.

crates/fuel-core/src/graphql_api/worker_service.rs
acerone85: for context: Headers cannot be used anymore, the reason being that subscriptions will return multiple graphql responses in a single HTTP response (and therefore have a single HeaderMap), while we require that each item in the response has its associated `CURRENT_FUEL_BLOCK_HEIGHT` field.  Moreover the Response metadata for subscriptions is returned before the actual response body (as it is standard with HTTP).

in the end graphql has a (very undocumented) way to add an `extensions` field to responses, at the same level of `data/errors` field. This does not require to tweak anything at the HTTP level.

crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
```diff
@@ -0,0 +1,148 @@
...
+                    );
+
+                    return response
+                }
+            }
+        }
+
+        let mut response = next.run(ctx, operation_name).await;
+
+        let current_block_height = view.latest_block_height();
```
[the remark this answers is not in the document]
netrome: Alright, fair enough :+1:

crates/client/src/client.rs
```diff
@@ -223,6 +250,34 @@ impl FuelClient {
         Self::from_str(url.as_ref())
     }
 
+    pub fn set_header(
```
xgreenx, line 253: I think we don't use it and can remove

crates/client/src/client.rs
```diff
@@ -223,6 +250,34 @@ impl FuelClient {
...
+        self
+    }
+
+    pub fn without_required_fuel_block_height(&mut self) -> &mut Self {
+        self.extensions.remove("required_fuel_block_height");
+        self
+    }
+
     /// Send the GraphQL query to the client.
     pub async fn query<ResponseData, Vars>(
```
xgreenx, line 282: What about `fn subscribe`? We also want to support setting heights for the `submit_and_await_commit` like queries

tests/tests/required_fuel_block_height_extension.rs
```diff
@@ -0,0 +1,185 @@
...
+        .get("current_fuel_block_height")
+        .unwrap()
+        .as_u64()
+        .unwrap();
+    assert_eq!(current_fuel_block_height, 0);
+}
+
+#[tokio::test]
+async fn current_fuel_block_height_header_is_present_on_successful_request() {
+    // TODO: Figure out a way to get the current fuel block height from FuelClient queries
```
xgreenx, line 103: I think it is very important to extract the current height in `fuel-core-client`. But yeah, it can be done in a follow up PR. Could you create an issue to track it and put link in todos(below you have the same todos).

xgreenx (commented):
Also in proposal we wanted to return 412. But based on the implementation we just return an error. Can we customize the number?

tests/tests/required_fuel_block_height_extension.rs
acerone85, line 103:
```suggestion
    // TODO: https://github.com/FuelLabs/fuel-core/issues/2605 
    // Figure out a way to get the current fuel block height from FuelClient queries
```

tests/tests/required_fuel_block_height_extension.rs
```diff
@@ -0,0 +1,186 @@
...
+        .get("current_fuel_block_height")
+        .unwrap()
+        .as_u64()
+        .unwrap();
+    assert_eq!(current_fuel_block_height, 0);
+}
+
+#[tokio::test]
+async fn current_fuel_block_height_header_is_present_on_no_required_fuel_block_height() {
+    // TODO: Figure out a way to get the current fuel block height from FuelClient queries
```
acerone85, line 148:
```suggestion
    // TODO: https://github.com/FuelLabs/fuel-core/issues/2605
    // Figure out a way to get the current fuel block height from FuelClient queries
```

acerone85:
> Also in proposal we wanted to return 412. But based on the implementation we just return an error. Can we customize the number?

This can be done for simple requests, but it might be more problematic for subscription requests, the reason is roughly the same as why we do not use headers anymore. In particular, The HTTP response status will be returned before the first response item is streamed into the response body. 

Furthermore, returning 4xx status code for failed requests is discouraged by the official graphql documentation:  see https://graphql.org/learn/serving-over-http/#status-codes

crates/client/src/client.rs
acerone85: done in d4793069b9

crates/client/src/client.rs
acerone85: done in d4793069b9

tests/tests/required_fuel_block_height_extension.rs
acerone85: yep, there is a similar issue for this: #2605. From what I understood you prefer the solution where the client keeps track of the current fuel block height. I can start working on a follow-up

xgreenx (approved):

xgreenx (approved):

crates/client/src/reqwest_ext.rs
```diff
@@ -0,0 +1,210 @@
...
+/// #    schema_module = "schema",
+/// # )]
+/// # struct Film {
+/// #    title: Option<String>,
+/// #    director: Option<String>
+/// # }
+/// #
+/// # #[derive(cynic::QueryFragment)]
+/// # #[cynic(
+/// #     schema_path = "../schemas/starwars.schema.graphql",
```
acerone85, line 78: would be nice to have an example related to how we use the `reqwest_ext` module, but not a big deal at this point :)

crates/client/src/reqwest_ext.rs
```diff
@@ -0,0 +1,214 @@
...
+            },
+        }
+    }
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+type BoxFuture<'a, T> = Pin<Box<dyn Future<Output = T> + Send + 'a>>;
+
+#[cfg(target_arch = "wasm32")]
+type BoxFuture<'a, T> = Pin<Box<dyn Future<Output = T> + 'a>>;
```
acerone85, line 57: curious to know why the `Send` trait bound is not available/needed on the `wasm32` architecture

crates/client/src/reqwest_ext.rs
```diff
@@ -0,0 +1,214 @@
...
+}
+
+impl<ResponseData, Errors> CynicReqwestBuilder<ResponseData, Errors> {
+    pub fn new(builder: reqwest::RequestBuilder) -> Self {
+        Self {
+            builder,
+            _marker: std::marker::PhantomData,
+        }
+    }
+}
```
acerone85, line 134: nit: is `impl<ResponseData, Errors> From<reqwest::RequestBuilder>for CynicReqwestBuilder<ResponseData, Errors>` more idiomatic?

crates/client/src/reqwest_ext.rs
```diff
@@ -0,0 +1,214 @@
...
+
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct ExtensionsRequest {
+    pub required_fuel_block_height: Option<BlockHeight>,
+}
+
+#[derive(Debug, Clone, serde::Deserialize)]
+pub struct ExtensionsResponse {
+    pub required_fuel_block_height: Option<BlockHeight>,
+    pub current_fuel_block_height: Option<BlockHeight>,
```
acerone85, line 21: nit: I think we require that this field is set in all graphql responses

crates/client/src/reqwest_ext.rs
```diff
@@ -0,0 +1,214 @@
...
+};
+
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct ExtensionsRequest {
+    pub required_fuel_block_height: Option<BlockHeight>,
+}
+
+#[derive(Debug, Clone, serde::Deserialize)]
+pub struct ExtensionsResponse {
+    pub required_fuel_block_height: Option<BlockHeight>,
```
acerone85, line 20: we don't set this field in the response in the graphql extension. But happy to add it if needed

crates/client/src/client.rs
```diff
@@ -223,6 +285,36 @@ impl FuelClient {
         Self::from_str(url.as_ref())
     }
 
+    pub fn with_required_fuel_block_height(
+        &mut self,
+        new_height: Option<BlockHeight>,
+    ) -> &mut Self {
+        match &mut self.require_height {
+            ConsistencyPolicy::Auto { height } => {
+                *height.lock().expect("Mutex poisoned") = new_height;
```
acerone85, line 294: should we use `parking_lot` Mutex, or use directly a `AtomicU32` and avoid a Mutex altogether?

netrome: Why do we need locks for this at all? We have a mutable reference to self here already.

crates/client/src/client.rs
```diff
@@ -287,40 +324,59 @@ impl FuelClient {
...
+        };
+
+        Self::decode_response(response, inner_required_height)
     }
 
-    fn decode_response<R>(response: GraphQlResponse<R>) -> io::Result<R>
+    fn decode_response<R, E>(
+        response: FuelGraphQlResponse<R, E>,
+        inner_required_height: Option<Arc<Mutex<Option<BlockHeight>>>>,
+    ) -> io::Result<R>
```
netrome, line 347: While this is okay, I think it would be cleaner to separate the height extraction to a separate `extract_required_height` function.

crates/client/src/client.rs
netrome: So I see that we need some type of interior mutability since the `query` function takes an immutable self receiver. However, I think it would be cleaner to change the signature of the `query` function to take a mutable self receiver.

Since that's a breaking change, how about we leave this as-is for now but create a follow-up to simplify this which we can include in the next breaking release?

crates/client/src/reqwest_ext.rs
netrome: TIL: It seems like it's a general assumption that the `wasm32-unknown-unknown` target is single-threaded https://users.rust-lang.org/t/multithreading-in-wasm-how-did-that-come-out/87518 - although there seem to be work to implement multi-threading in WASM coming up in the future. Interesting.

netrome: (and if the target is single-threaded you should not have to require `Send` anywhere)

netrome (approved):
The code seems correct to me, but I'm pretty annoyed by the `Option<Arc<Mutex>>` and would love to get rid of it as soon as possible. I understand that we might need some sort of interior mutability to preserve backwards compatibility in the client though.

PR #2478, opened 2024-12-06
Ignore receipts from failed transactions in `message_receipts_proof`

### LINKED ISSUE #2421: `message_receipts_proof` creates invalid proof if transaction with `MessageOut` receipt failed
The `message_receipts_proof` function creates invalid proof if transaction with `MessageOut` receipt failed during execution of the block.

The code creates Merkle tree over all `MessageOut` receipts even if transaction failed. While the `message_outbox_root` includes only messages from successful transactions.

image
image

## Definition of done

The user should be able to generate valid proof for blocks containing successful and failed transactions that emit a `MessageOut` receipt.

AurelienFT:
Resolves #2421 

In this PR we changed the  `message_receipts_proof` to ignore the receipts from the failed transactions.

## Checklist

### Before requesting review

### DIFF
--- a/crates/fuel-core/src/query/message.rs
+++ b/crates/fuel-core/src/query/message.rs
@@ -278,7 +278,11 @@ fn message_receipts_proof<T: MessageProofData + ?Sized>(
     // Get the message receipts from the block.
     let leaves: Vec<Vec<Receipt>> = message_block_txs
         .iter()
-        .map(|id| database.receipts(id))
+        .filter_map(|id| match database.transaction_status(id) {
+            Ok(TransactionStatus::Success { receipts, .. }) => Some(Ok(receipts)),
+            Ok(_) => None,
+            Err(err) => Some(Err(err)),
+        })
         .try_collect()?;
     let leaves = leaves.into_iter()
         // Flatten the receipts after filtering on output messages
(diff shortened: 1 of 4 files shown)


crates/fuel-core/src/query/message.rs
```diff
@@ -278,7 +278,11 @@ fn message_receipts_proof<T: MessageProofData + ?Sized>(
     // Get the message receipts from the block.
     let leaves: Vec<Vec<Receipt>> = message_block_txs
         .iter()
-        .map(|id| database.receipts(id))
```
xgreenx, line 281: Maybe we also need to update another place with `database.receipts` function call. And remove `receipts` function at all=)

image

crates/fuel-core/src/query/message.rs
```diff
@@ -337,3 +341,184 @@ pub fn message_status(
...
+    fn test_message_receipts_proof_ignore_failed() {
+        // Create a fake database
+        let mut database = FakeDB::new();
+
+        // Given
+        // Create a block with a valid transaction and receipts
+        let mut block = CompressedBlock::default();
+        let valid_tx_id = Bytes32::new([1; 32]);
+        let mut valid_tx_receipts = vec![];
+        for i in 0..100 {
```
xgreenx, line 448: Why do we need `100` receipts?=D

crates/fuel-core/src/query/message.rs
```diff
@@ -337,3 +341,184 @@ pub fn message_status(
...
+                receipts: valid_tx_receipts.clone(),
+                total_fee: 0,
+                total_gas: 0,
+                result: None,
+            },
+        );
+        database.insert_receipts(valid_tx_id, valid_tx_receipts.clone());
+
+        // Get the message proof with the valid transaction
+        let message_proof_valid_tx = message_receipts_proof(
```
xgreenx, line 475: It would be nice if we tested public function `message_proof` instead of the private

crates/fuel-core/src/query/message.rs
AurelienFT: Its in order to build a bigger tree and have some path for the proof. It could be 10 also.

crates/fuel-core/src/query/message.rs
AurelienFT: Done

crates/fuel-core/src/query/message.rs
```diff
@@ -337,3 +332,173 @@ pub fn message_status(
...
+
+        // When
+        // Get the message proof with the same message id
+        let message_proof_invalid_tx =
+            message_proof(&database, valid_tx_id, Nonce::default(), block_height)
+                .unwrap();
+
+        // Then
+        // The proof should be the same because the invalid transaction was ignored
+        assert_eq!(message_proof_valid_tx, message_proof_invalid_tx);
```
xgreenx, line 502: Have you tested it without the fix, and it was failing here, yeah?

xgreenx (approved):
Looks good to me=) I hope the answer on my question is "yes"=D

crates/fuel-core/src/query/message.rs
AurelienFT: "yes" :)

netrome (approved):
Nice stuff!

Issue #2022, opened 2024-07-05
Sequential opcodes return an error when touching the last storage key


xgreenx:
All sequential opcodes return an error when touching the last storage key because we increase the key after that:

<img width="1667" alt="image" src="https://github.com/FuelLabs/fuel-core/assets/18346821/cc04cfca-9ad0-461f-b108-acb0c5f2bc86">

If the user wants to update only the last storage key, the VM will return an error.

PR #2479, opened 2024-12-06
Fix last iteration in sequential opcode

AurelienFT:
Fixes https://github.com/FuelLabs/fuel-core/issues/2022

Previously the read and write of the last key of the contract state was creating error because of multiple bad checks of limits. Also the key was increase in all case at the end of the loop causing the `increase()` to fail on overflow even if there is values to be inserted left.

## Checklist

### Before requesting review

### DIFF
--- a/crates/storage/src/vm_storage.rs
+++ b/crates/storage/src/vm_storage.rs
@@ -344,11 +344,13 @@ where
         let mut state_key = Bytes32::zeroed();
 
         let mut results = Vec::new();
-        for _ in 0..range {
+        for i in 0..range {
             key.to_big_endian(state_key.as_mut());
             let multikey = ContractsStateKey::new(contract_id, &state_key);
             results.push(self.database.storage::<ContractsState>().get(&multikey)?);
-            key.increase()?;
+            if i.saturating_add(1) != range {
+                key.increase()?;
+            }
         }
         Ok(results)
     }
@@ -367,12 +369,12 @@ where
 
         // verify key is in range
         current_key
-            .checked_add(U256::from(values.len()))
+            .checked_add(U256::from(values.len().saturating_sub(1)))
             .ok_or_else(|| anyhow!("range op exceeded available keyspace"))?;
 
         let mut key_bytes = Bytes32::zeroed();
         let mut found_unset = 0u32;
-        for value in values {
+        for (idx, value) in values.iter().enumerate() {
             current_key.to_big_endian(key_bytes.as_mut());
 
             let option = self
@@ -386,7 +388,9 @@ where
                     .expect("We've checked it above via `values.len()`");
             }
 
-            current_key.increase()?;
+            if idx.saturating_add(1) != values.len() {
+                current_key.increase()?;
+            }
         }
 
         Ok(found_unset as usize)
--- a/crates/storage/src/vm_storage.rs
+++ b/crates/storage/src/vm_storage.rs
@@ -345,12 +345,12 @@ where
 
         let mut results = Vec::new();
         for i in 0..range {
+            if i != 0 {
+                key.increase()?;
+            }
             key.to_big_endian(state_key.as_mut());
             let multikey = ContractsStateKey::new(contract_id, &state_key);
             results.push(self.database.storage::<ContractsState>().get(&multikey)?);
-            if i.saturating_add(1) != range {
-                key.increase()?;
-            }
         }
         Ok(results)
     }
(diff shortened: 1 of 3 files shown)


crates/storage/src/vm_storage.rs
```diff
@@ -344,11 +344,13 @@ where
...
         let mut results = Vec::new();
-        for _ in 0..range {
+        for i in 0..range {
             key.to_big_endian(state_key.as_mut());
             let multikey = ContractsStateKey::new(contract_id, &state_key);
             results.push(self.database.storage::<ContractsState>().get(&multikey)?);
-            key.increase()?;
+            if i.saturating_add(1) != range {
+                key.increase()?;
+            }
```
xgreenx, line 353: We can put it at the begin of the loop and only increase if `i != 0`

tests/tests/vm_storage.rs
```diff
@@ -191,6 +201,7 @@ mod tests {
             )
             .map_err(|_| ())
             .map(|v| v == 0);
+        dbg!(&insert_status);
```
xgreenx, line 204:
```suggestion
```

xgreenx (commented):
We also have `contract_state_remove_range` that has the same problem=)

xgreenx (dismissed):

xgreenx (dismissed):

crates/storage/src/vm_storage.rs
```diff
@@ -344,11 +344,13 @@ where
         let mut state_key = Bytes32::zeroed();
 
         let mut results = Vec::new();
-        for _ in 0..range {
+        for i in 0..range {
```
rymnc, line 347: couldn't we alter the range from 1..range instead of checking every time if the i !=0 ?

crates/storage/src/vm_storage.rs
```diff
@@ -367,12 +369,15 @@ where
 
         // verify key is in range
         current_key
-            .checked_add(U256::from(values.len()))
+            .checked_add(U256::from(values.len().saturating_sub(1)))
```
rymnc, line 372: can we use `checked_sub` instead?

crates/storage/src/vm_storage.rs
```diff
@@ -367,12 +369,15 @@ where
...
             .ok_or_else(|| anyhow!("range op exceeded available keyspace"))?;
 
         let mut key_bytes = Bytes32::zeroed();
         let mut found_unset = 0u32;
-        for value in values {
+        for (idx, value) in values.iter().enumerate() {
+            if idx != 0 {
+                current_key.increase()?;
+            }
             current_key.to_big_endian(key_bytes.as_mut());
```
rymnc, line 381:
```suggestion
if let Some((first_idx, first_value)) = values.iter().enumerate().next() {
    current_key.to_big_endian(key_bytes.as_mut());

    for (idx, value) in values.iter().enumerate().skip(1) {
        current_key.increase()?; 
        current_key.to_big_endian(key_bytes.as_mut());
    }
}

```

rymnc (commented):
lgtm! added some comments in line

crates/storage/src/vm_storage.rs
AurelienFT: The problem is that for i == 0 we still want to do the operations below in the loop. We could try to start at `start_key - 1` but that will cause issue in case start key is the lowest possible. I never see good solution in these cases :/ If you have an idea you can try to apply it

crates/storage/src/vm_storage.rs
AurelienFT: I don't want to because if values is 0 then I don't want to throw and error or anything I just want the checked_add to not do anything

crates/storage/src/vm_storage.rs
AurelienFT: If I try to make it compile and remove the unused part we endup with : 
``` rust
if let Some(value) = values.iter().next() {
            current_key.to_big_endian(key_bytes.as_mut());
            let option = self
                .database
                .storage::<ContractsState>()
                .replace(&(contract_id, &key_bytes).into(), value)?;

            if option.is_none() {
                found_unset = found_unset
                    .checked_add(1)
                    .expect("We've checked it above via `values.len()`");
            }
            for value in values.iter().skip(1) {
                current_key.increase()?; 
                current_key.to_big_endian(key_bytes.as_mut());
                let option = self
                .database
                .storage::<ContractsState>()
                .replace(&(contract_id, &key_bytes).into(), value)?;

                if option.is_none() {
                    found_unset = found_unset
                        .checked_add(1)
                        .expect("We've checked it above via `values.len()`");
                }
            }
        }
```
Which seems a lot of code duplication to me, wdyt ?

rymnc: eh yeah that looks not so nice, approving!

rymnc (approved):

xgreenx (approved):

PR #2483, opened 2024-12-07
Suggestions to simplify the V1 storage requirements and initialization

xgreenx:
Suggestions for the https://github.com/FuelLabs/fuel-core/pull/2468 PR.

### DIFF
--- a/crates/services/gas_price_service/src/ports.rs
+++ b/crates/services/gas_price_service/src/ports.rs
@@ -11,7 +11,10 @@ use crate::{
         utils::Result,
     },
     v0::metadata::V0AlgorithmConfig,
-    v1::metadata::V1AlgorithmConfig,
+    v1::{
+        metadata::V1AlgorithmConfig,
+        uninitialized_task::fuel_storage_unrecorded_blocks::AsUnrecordedBlocks,
+    },
 };
 
 pub trait L2Data: Send + Sync {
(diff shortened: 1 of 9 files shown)


crates/services/gas_price_service/src/ports.rs
```diff
@@ -43,12 +46,22 @@ pub trait GetDaSequenceNumber: Send + Sync {
...
+where
+    Self: 'static,
+    Self: Send + Sync,
+    Self: GetMetadataStorage + GetDaSequenceNumber,
+{
+    type Transaction<'a>: AsUnrecordedBlocks
+        + SetMetadataStorage
+        + GetMetadataStorage
+        + SetDaSequenceNumber
+        + GetDaSequenceNumber
```
MitchTurner, line 59: Generally I think pushing many constraints into a single trait is an anti-pattern. I prefer to keep all the constraints as close to place that needs those constraints met. Namely, the `GasPriceService` needs these things to be transactionable together in order to construct the storage tx.

You can make the argument that the _domain_ requirement is that our storage is transactionable for these things. Which I would agree with, but then we need to include the `UnrecordedBlocks` as a requirement as well.

crates/services/gas_price_service/src/v1/uninitialized_task/fuel_storage_unrecorded_blocks.rs
```diff
@@ -11,15 +11,38 @@ use fuel_core_storage::{
...
+{
+    type Wrapper<'a> = FuelStorageUnrecordedBlocks<&'a mut Self>
+        where
+            Self: 'a;
+
+    fn as_unrecorded_blocks(&mut self) -> Self::Wrapper<'_> {
+        FuelStorageUnrecordedBlocks::new(self)
+    }
+}
+
```
MitchTurner, line 45: What does this trait give us that we didn't get with the newtype wrapper? I feel like the newtype pattern is the idiomatic way to deal with the orphan rule problem in Rust, but maybe this solution freed up some constraints that I can't see.

crates/services/gas_price_service/src/ports.rs
MitchTurner: I feel like I'd prefer these constraints on the RunableTask impl for GasPriceService than on some obfuscated trait, but if we can name this something informative, not just Transactionable then I might be okay with it. e.g. GasPriceServiceAtomicStorage or something like that :)

xgreenx: Naming is fully up to you, I just tried to show you an exampel fo how it can looks like=)

crates/services/gas_price_service/src/v1/uninitialized_task/fuel_storage_unrecorded_blocks.rs
xgreenx: Without this helpers trait, you need to mock `StorageMutate<UnrecordedBlocksTable, Error = StorageError>` in tests, instead of just mocking `AsUnrecordedBlocks`. You can remove it and see the problem=)

PR #2487, opened 2024-12-09
refactor: The current logic for extracting and matching receipts is somewhat verbose. It can be simplified using if let.

RiceChuan:
The current logic for extracting and matching receipts is somewhat verbose. It can be simplified using if let.


xgreenx:
Hey! Thanks for the contribution. We are open to accepting contributions from community members but we don't accept small PRs from first-time contributors. We are spammed by them, unfortunately. Feel free to look through our issues or create an issue for something you'd like to see improved!

Issue #2032, opened 2024-07-10
Add `pagination` support for `balances` endpoint on GraphQL


arboleya:
In short, it should provide the same functionality we have for `coins`, `messages`, etc.

PR #2490, opened 2024-12-10
Pagination queries for `balances` endpoint

### LINKED ISSUE #2023: P2P is doing a lot of database lookups

rafal-ch:
Closes [#2023](https://github.com/FuelLabs/fuel-core/issues/2032)
Closes https://github.com/FuelLabs/fuel-core/issues/2496

## TODO:
* [X] Add integration test due to custom logic related to base asset id
* [X] Make it work properly with base asset id

## Description
This PR enables the paginated queries for balances, but only if the balance indexation is enabled. Otherwise, the attempt to send paginated query will result with `Can not use pagination when balances indexation is not available` error.

Additionally, if the balances indexation is available the default query cost for "balances" will be set to 0, and the complexity will be calculated based on the amount of items requested. When indexation is not available, the default query cost stays intact (40001).

## Checklist

### Before requesting review


crates/fuel-core/src/schema/balance.rs
```diff
@@ -79,8 +79,6 @@ impl BalanceQuery {
         Ok(balance)
     }
 
-    // TODO: This API should be migrated to the indexer for better support and
-    //  discontinued within fuel-core.
     #[graphql(complexity = "query_costs().balance_query")]
```
xgreenx, line 82:
```suggestion
    #[graphql(complexity = "query_costs().balance_query + query_costs().storage_iterator \
        + (query_costs().storage_read + first.unwrap_or_default() as usize) * child_complexity \
        + (query_costs().storage_read + last.unwrap_or_default() as usize) * child_complexity\")]
```

We need to limit the number of requested balances=)

xgreenx: It also makes this change breaking

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
```diff
@@ -233,15 +233,48 @@ impl OffChainDatabase for OffChainIterableKeyValueView {
         }
     }
 
-    fn balances(
-        &self,
+    fn balances<'a>(
```
xgreenx, line 236: Yeah, it is a tricky function because of how we handle balances.

~Hmm, do you remember why we decided to skip balances during iteration and add them at the begin?~

I remember: if you don't have coins but have messages, they will not appear during the iteration.

The problem with the current solution is when `start == base_asset_id`. The current code will return an incorrect iterator.

Either we need to handle it differently, or come up with another solution. It would be nice if we could insert base asset balance into the mid of the iterator. Maybe we could change the indexation and make it always add empy indexation for the base asset(like entry in the balances table with 0 amount just for balances)?

Then during iteration we call always fetch its value by using `self.base_asset_balance`?

rafal-ch:
Converting to draft, some rework is needed here.

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
rafal-ch: [This](https://github.com/FuelLabs/fuel-core/pull/2490/commits/b1bfb20084cfd1797b4ac14e966826269f7320ed) [two](https://github.com/FuelLabs/fuel-core/pull/2490/commits/3ae5222afee478bb2d19ba175f8ed3a755345625) commits should solve the problem

crates/fuel-core/src/schema/balance.rs
rafal-ch: Updated in https://github.com/FuelLabs/fuel-core/pull/2490/commits/fd2b0191c7ede59eec061bc7f5d78c2de60ef0ab

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
```diff
@@ -233,15 +233,50 @@ impl OffChainDatabase for OffChainIterableKeyValueView {
...
+            (IterDirection::Forward, None) => {
+                let base_asset_balance = self.base_asset_balance(base_asset_id, owner);
+                let non_base_asset_balance =
+                    self.non_base_asset_balances(owner, None, direction, base_asset_id);
+                base_asset_balance
+                    .chain(non_base_asset_balance)
+                    .into_boxed()
+            }
+            (IterDirection::Forward, Some(asset_id)) => {
+                let start = (asset_id != *base_asset_id)
```
xgreenx, line 253: If `asset_id == base_asset_id`, you will skip base asset while we should return it, since it is starting entry=)

rafal-ch: Ok, I was wondering why it actually works... :)

It turned out that even if we return the bass asset id here, it'll be skipped by the logic inside `query_pagination` [here](https://github.com/FuelLabs/fuel-core/blob/master/crates/fuel-core/src/schema.rs#L170). Internally, it kinda changes the notion of `start` to actually mean `after`.

So now the [updated](https://github.com/FuelLabs/fuel-core/pull/2490/commits/a943e2be2cee62389a14e4cd5d8e0999ffff0cf2) behavior always includes the `start` entry in the output, even though it'll be filtered by the query. While it won't change the returned balances, it is now consistent with other pagination queries. This consistency might be actually important because the query code makes some further calculations about `has_next_page` and other flags.

crates/fuel-core/src/schema/balance.rs
```diff
@@ -79,9 +79,11 @@ impl BalanceQuery {
         Ok(balance)
     }
 
-    // TODO: This API should be migrated to the indexer for better support and
```
AurelienFT, line 82: This is not something we want to do in the future anymore ? In my mind it's still something we want to do in the future to decouple both use cases of the node.

crates/fuel-core/src/schema/balance.rs
```diff
@@ -79,9 +79,11 @@ impl BalanceQuery {
...
-    //  discontinued within fuel-core.
-    #[graphql(complexity = "query_costs().balance_query")]
+    #[graphql(
+        complexity = "query_costs().balance_query + query_costs().storage_iterator \
+        + (query_costs().storage_read + first.unwrap_or_default() as usize) * child_complexity \
```
AurelienFT, line 84: The cost computation is hard to understand for me, maybe it's normal but it might requires a comment. What I don't understand : 
- Why are we making pay for `first * child_complexity` + `last * child_complexity`. Example : `(4 * 10) + (9 * 10) = 130`. Inner question : Why are we paying if the first is far away and why are we paying again for all the numbers from `0` to `last` some where already included in the `first` computation.

What is logic to me is two solutions : 
- Either we want people to pay for numbers before their first and so it would be : 
```rust
(query_costs().storage_read + last.unwrap_or_default() as usize) * child_complexity
``` 
(because last is higher and include all the precents numbers)
- Either we just want to pay for the interval : 
```rust
(query_costs().storage_read + (last.unwrap_or_default() as usize - first.unwrap_or_default() as usize)) * child_complexity
```

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
```diff
@@ -233,15 +233,49 @@ impl OffChainDatabase for OffChainIterableKeyValueView {
...
+        match (direction, start) {
+            (IterDirection::Forward, None) => {
+                self.base_asset_first(owner, base_asset_id, direction)
+            }
+            (IterDirection::Forward, Some(asset_id)) => {
+                if asset_id == *base_asset_id {
+                    self.base_asset_first(owner, base_asset_id, direction)
+                } else {
+                    let start = (asset_id != *base_asset_id)
+                        .then_some(CoinBalancesKey::new(owner, &asset_id));
```
AurelienFT, line 252: Why are we redoing the comparaison here as we are already in the `else`
```suggestion
                    let start = Some(CoinBalancesKey::new(owner, &asset_id));
```

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
```diff
@@ -233,15 +233,49 @@ impl OffChainDatabase for OffChainIterableKeyValueView {
...
+                    self.non_base_asset_balances(owner, start, direction, base_asset_id)
+                }
+            }
+            (IterDirection::Reverse, None) => {
+                self.non_base_assets_first(start, owner, base_asset_id, direction)
+            }
+            (IterDirection::Reverse, Some(asset_id)) => {
+                if asset_id == *base_asset_id {
```
AurelienFT, line 260: You can leverage pattern matching as you did in `non_base_asset_balances` don't really know if it's better or not. up to you :)

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
```diff
@@ -250,37 +284,68 @@ impl OffChainDatabase for OffChainIterableKeyValueView {
...
+        start: Option<CoinBalancesKey>,
+        direction: IterDirection,
+        base_asset_id: &'a AssetId,
+    ) -> BoxedIter<'_, Result<(AssetId, u128), StorageError>> {
+        self.iter_all_filtered_keys::<CoinBalances, _>(
+            Some(owner),
+            start.as_ref(),
+            Some(direction),
+        )
+        .filter_map(move |result| match result {
```
AurelienFT, line 302: It would be valuable in the future to have a comment about this behavior because reading the comments on the review helped me but it's not really future proof and without it I wouldn't understand why `None` on base asset etc..

AurelienFT (commented):
Very good usage of rust syntax and bricks. Liked it :)

crates/fuel-core/src/schema/balance.rs
rafal-ch: Hard to tell, as far as I know, indexer is no longer a thing. But even if it is, we should track it via a separate issue, not a `TODO`.

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
rafal-ch: I think I'll leave it as is, should be ok.

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
rafal-ch: I think it's quite clear, because the function name clearly says "NON_base_asset_balance", hence we map base assets to `None` to not process them further.

But of course I'm biased due to spending some significant time on this feature.

Can you suggest the message that would explain the exact thing that was confusing to you? I'm happy to include it for other readers.

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
rafal-ch: Nice catch - it's been quite tricky to implement this function correctly, it's probably a leftover from one of the trials.

Fixed in e3e890a5095cb51b5bef0af2316441fea6345006

crates/fuel-core/src/service/adapters/graphql_api/off_chain.rs
AurelienFT: Mhhh yeah with the function name it's more understandable, didn't paid attention correctly when first read. I'm resolving

crates/fuel-core/src/schema/balance.rs
AurelienFT: It makes sense on this comment but on the code it's `(storage_read + 0) * child_complexity` not `*` which doesn't simplify to 0 and change the consumption. If you indeed change to `*` might be interesting to comment that only one of the two should not be 0

rafal-ch: Yes, indeed, I overlooked that.
The complexity calculation is updated to match the description above in this commit: 3522efc84111c93b93ccffda6718b697554371d5

cc: @xgreenx

AurelienFT (dismissed):

crates/fuel-core/src/schema/balance.rs
```diff
@@ -79,9 +79,11 @@ impl BalanceQuery {
...
     }
 
-    // TODO: This API should be migrated to the indexer for better support and
-    //  discontinued within fuel-core.
-    #[graphql(complexity = "query_costs().balance_query")]
+    #[graphql(
+        complexity = "query_costs().balance_query + query_costs().storage_iterator \
+        + (query_costs().storage_read * first.unwrap_or_default() as usize) * child_complexity \
+        + (query_costs().storage_read * last.unwrap_or_default() as usize) * child_complexity"
+    )]
```
xgreenx, line 86: Actually, we want to have something like `count * child_complexity + count * storage_read`, not `storage_read * child_complexity`

rafal-ch: Please see here: https://github.com/FuelLabs/fuel-core/pull/2490/commits/afc133b3200feb2f0750266a0beb86f464b85d0d

This commit contains changes that are needed to support ~11k balances as sent from the Rust SDK and bail when more is requested.

AurelienFT (dismissed):

tests/tests/balances.rs
```diff
@@ -330,3 +330,260 @@ async fn first_5_balances() {
...
+            if !paginated_result.has_next_page {
+                break
+            }
+        }
+
+        // Then
+
+        // Please mind that if present, base asset id is always reported first
+        // (or last in case of backward pagination).
+        let mut expected_balances = match (message_coin, base_asset_coin) {
```
xgreenx, line 508: [nit] It would be nice to have the `expected_balances` part of the input to the test. But I'm not sure how it will affect readability=)

xgreenx (approved):

crates/fuel-core/src/schema/balance.rs
```diff
@@ -79,9 +79,27 @@ impl BalanceQuery {
...
+    // + (child_complexity * first.unwrap_or_default() as usize) \
+    // + (query_costs().storage_read * last.unwrap_or_default() as usize) \
+    // + (child_complexity * last.unwrap_or_default() as usize)"
+    // )]
+
+    // The 0.66 factor is a Goldilocks approach to balancing the complexity of the query for the SDKs.
+    // Rust SDK sends a query with child_complexity ≅ 11 and we want to support slightly more
+    // than 10k items in a single query (so we target 11k). The total complexity would be 11k * 11 = 121k,
+    // but since our default limit is 80k, we need the 0.66 factor.
+    #[graphql(complexity = "
```
xgreenx, line 99: We still need to use `query_costs().balance_query` and make `balance_query = 40001` in the case when no caching and `balance_query = 0` in the case when caching is enabled

xgreenx (changes requested):

crates/fuel-core/src/schema/balance.rs
rafal-ch: Updated in https://github.com/FuelLabs/fuel-core/pull/2490/commits/546b713186a9dc57e74c3fddc663dbd2f1a6beca

AurelienFT (approved):

xgreenx (approved):

AurelienFT (approved):

PR #2491, opened 2024-12-11
Execution tracing: GraphQL query to get storage inputs for past blocks

Dentosal:
## Description

To support local debugging and execution tracing, we need to add an API that returns the state that VM needs to execute transactions. This is done by recording all database accesses during tx validation, called *storage read replay* (bikeshedding in progress). This data is then exposed as-is through GraphQL. The API is non-trivial to consume, but a seperate client is proviced, see https://github.com/FuelLabs/execution-trace

The implementation exposes database table names and representations directly. Maintaining backwards compatibility with this could turn out to be quite hard.

Requires `--historical-execution` flag to enable, as this is otherwise quite expensive.

Follow-ups:
* https://github.com/FuelLabs/fuel-core/issues/2625

Example query:

```graphql
mutation {
  storageReadReplay(height:"4") { column, key, value }
}
```

## Open questions:
- What should the query cost be?

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog: No breaking changes!
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior: This is outside spec scope.

### Before requesting review

See the VM PR https://github.com/FuelLabs/fuel-vm/pull/881.


AurelienFT (commented):
Good integration with database types ! Liked it !

crates/services/executor/src/executor.rs
```diff
@@ -702,6 +702,114 @@ where
         Ok(())
     }
 
+    #[tracing::instrument(skip_all)]
+    fn record_storage_reads_for<D>(
```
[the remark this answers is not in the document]
Dentosal: Hahahaha I already managed to refactor that code out. Nice. I'll remove it.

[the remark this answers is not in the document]
Dentosal: a4a6bbf

crates/fuel-core/src/schema/tx.rs
```diff
@@ -331,6 +336,30 @@ impl TxMutation {
...
     }
 
+    /// Get execution trace for an already-executed transaction.
+    #[graphql(complexity = "query_costs().storage_read_replay + child_complexity")]
+    async fn storage_read_replay(
+        &self,
+        ctx: &Context<'_>,
+        height: U32,
+    ) -> async_graphql::Result<Vec<StorageReadReplayEvent>> {
+        let config = ctx.data_unchecked::<GraphQLConfig>().clone();
```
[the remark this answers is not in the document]
Dentosal: a4a6bbf

AurelienFT (approved):
LGTM. I would vouch for an indepedent flag that can be enabled by default when we use debug but that could be enabled by himself also.

tests/tests/storage_read_replay.rs
```diff
@@ -0,0 +1,178 @@
...
+        vec![StorageSlot::new(Bytes32::zeroed(), Bytes32::zeroed())],
+    )
+    .maturity(maturity)
+    .add_fee_input()
+    .add_contract_created()
+    .finalize();
+
+    let contract_id = CreateMetadata::compute(&tx).unwrap().contract_id;
+
+    let mut status_stream = client.submit_and_await_status(&tx.into()).await.unwrap();
```
rymnc, line 72: can we use `submit_and_await_commit_with_tx` here instead?
https://github.com/FuelLabs/fuel-core/blob/bb0368e262ce3793cff3990dcc92c11354f2dc89/crates/client/src/client.rs#L569-L570

it'll need the `subscriptions` feature to be enabled, but i think it might be already

tests/tests/storage_read_replay.rs
```diff
@@ -0,0 +1,178 @@
...
+    let mut status_stream = client.submit_and_await_status(&tx).await.unwrap();
+    let intermediate_status = status_stream.next().await.unwrap().unwrap();
+    assert!(matches!(
+        intermediate_status,
+        TransactionStatus::Submitted { .. }
+    ));
+    let final_status = status_stream.next().await.unwrap().unwrap();
+    let TransactionStatus::Success { block_height, .. } = final_status else {
+        panic!("Tx wasn't included in a block: {:?}", final_status);
+    };
```
rymnc, line 133: same as prev

bin/fuel-core/src/cli/run/graphql.rs
```diff
@@ -112,6 +112,14 @@ pub struct QueryCosts {
     )]
     pub dry_run: usize,
 
+    /// Query costs for generating execution trace for a block.®
+    #[clap(
+        long = "query-cost-storage-read-replay",
+        default_value = DEFAULT_QUERY_COSTS.storage_read_replay.to_string(),
+        env
+    )]
+    pub storage_read_replay: usize,
```
MitchTurner, line 121: The verbiage around this isn't clear to me. 

This argument is for setting the costs? I think more context is required here and perhaps a more specific name for the field.

bin/fuel-core/src/cli/run/graphql.rs
```diff
@@ -112,6 +112,14 @@ pub struct QueryCosts {
     )]
     pub dry_run: usize,
 
+    /// Query costs for generating execution trace for a block.®
```
MitchTurner, line 115:
```suggestion
    /// Query costs for generating execution trace for a block
```
I feel like I'm ruining some fun here. But I'm not sure it belongs

MitchTurner (commented):
I've read a lot more, but I'm still WIP with my review. Will add more feedback.

tests/tests/storage_read_replay.rs
```diff
@@ -0,0 +1,178 @@
...
+    block_height
+}
+
+/// Create a counter contract.
+/// Increment it multiple times, and make sure the replay gives correct storage state every time.
+#[tokio::test(flavor = "multi_thread")]
+async fn test_storage_read_replay_returns_counter_state() {
```
MitchTurner, line 140:
```suggestion
async fn storage_read_replay__returns_correct_contract_state() {
```

crates/client/src/client.rs
```diff
@@ -508,6 +511,28 @@ impl FuelClient {
...
 
+    /// Get storage read replay for a block
+    pub async fn storage_read_replay(
+        &self,
+        height: &BlockHeight,
+    ) -> io::Result<Vec<StorageReadReplayEvent>> {
+        let query: Operation<
+            schema::storage_read_replay::StorageReadReplay,
+            schema::storage_read_replay::StorageReadReplayArgs,
+        > = schema::storage_read_replay::StorageReadReplay::build(
```
MitchTurner, line 522: nit: Does this need the type annotation? I feel like all the info is in the construction.

tests/tests/storage_read_replay.rs
```diff
@@ -0,0 +1,178 @@
...
+            .iter()
+            .find(|item| item.column == "ContractsState" && item.key == storage_slot_key)
+            .expect("No storage read found")
+            .value
+            .clone();
+
+        let mut expected_value = [0; 32];
+        expected_value[..8].copy_from_slice(&i.to_be_bytes());
+
+        assert!(value == Some(expected_value.to_vec()));
```
MitchTurner, line 176: Could create a small helper:
```rs
fn get_counter_from_storage_bytes(bytes: &[u8]) -> u64 {
    let counter_bytes = bytes[..8];
    u64::from_be_bytes(counter_bytes)
}
```
and then
```rs
let value = (...).unwrap();
let actual = get_counter_from_storage_bytes(&value);
let expected = i;

assert_eq!(expected, actual);
```

crates/types/Cargo.toml
```diff
@@ -27,6 +27,7 @@ derive_more = { version = "0.99" }
 fuel-vm-private = { workspace = true, default-features = false, features = [
   "alloc",
 ] }
+hex = { version = "0.4", default-features = false }
```
MitchTurner, line 30: Can you not use the `hex` from the workspace?

crates/client/src/client/schema/storage_read_replay.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub storage_read_replay: Vec<StorageReadReplayEvent>,
+}
+
+#[cfg(test)]
+pub mod tests {
+    use super::*;
+    use fuel_core_types::fuel_types::BlockHeight;
+
+    #[test]
+    fn execution_trace_block_tx_gql_output() {
```
MitchTurner, line 51: It would be good to specify that this is a compatibility regression/snapshot test. I don't get anything from this test name.

MitchTurner (commented):
Looking good overall. I add a few easy-to-fix comments. 

One thing I'd like to call out is there doesn't seem to be any UTs for the new behavior. I feel like they would be relatively easy to add and would push things left a bit.

crates/client/src/client/schema/storage_read_replay.rs
Dentosal: We have loads of `*_gql_output` tests that do this. Should those all be renamed?

bin/fuel-core/src/cli/run/graphql.rs
Dentosal: Yes. It's a filed in a struct called `QueryCosts`, and all other arguments have similar purpose.

tests/tests/storage_read_replay.rs
Dentosal: Good catch, done in fc097bf56e

tests/tests/storage_read_replay.rs
Dentosal: fc097bf56e

bin/fuel-core/src/cli/run/graphql.rs
Dentosal: Hahahah it was a typo, but keeping it would have been funny. Removed in fc097bf56e.

tests/tests/storage_read_replay.rs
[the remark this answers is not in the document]
Dentosal: fc097bf56e

tests/tests/storage_read_replay.rs
Dentosal: fc097bf56e

tests/tests/storage_read_replay.rs
Dentosal: fc097bf56e

Dentosal:
Taking this back to draft until https://github.com/FuelLabs/fuel-core/pull/2661 is merged, as I want to refactor this on top of it.

bin/fuel-core/src/cli/run/graphql.rs
MitchTurner: Ah ah ah. I remember this now. I suggested on another PR that we should rename this to "Complexity" or something (just because we already have concepts of gas costs etc).  I won't nitpick too much on this since at least it's within the `graphql` crate, but giving some clarity in the comment on what types of costs this is would be valuable (even if the other args aren't that clear).

crates/client/src/client/schema/storage_read_replay.rs
MitchTurner: > We have loads of *_gql_output tests that do this. Should those all be renamed?

Not your responsibility. I don't think upholding the precedence gains us anything here. I'd prefer clear names wherever possible.

crates/client/src/client.rs
Dentosal: No. I cleaned it up.

crates/types/Cargo.toml
Dentosal: Apparently I can. I had some issues with `default-features` but now it's fine.

CHANGELOG.md
```diff
@@ -7,8 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 
 ### Added
-
 - [2648](https://github.com/FuelLabs/fuel-core/pull/2648): Add feature-flagged field to block header `fault_proving_header` that contains a commitment to all transaction ids.
+- [2491](https://github.com/FuelLabs/fuel-core/pull/2491): Storage read replays of historical blocks for execution tracing. Only available behind `--debug` flag.
```
xgreenx, line 11:
```suggestion
- [2491](https://github.com/FuelLabs/fuel-core/pull/2491): Storage read replays of historical blocks for execution tracing. Only available behind `--historical-execution` flag.
```

crates/fuel-core/src/graphql_api.rs
```diff
@@ -91,6 +92,7 @@ pub const DEFAULT_QUERY_COSTS: Costs = Costs {
     get_peers: 40001,
     estimate_predicates: 40001,
     dry_run: 12000,
+    storage_read_replay: 12000,
```
xgreenx, line 95:
```suggestion
    storage_read_replay: 40001,
```

I think we want to allow only one request like this

crates/fuel-core/src/schema/tx/types.rs
```diff
@@ -980,6 +980,39 @@ impl DryRunTransactionExecutionStatus {
     }
 }
 
+pub struct StorageReadReplayEvent {
+    column: String,
```
xgreenx, line 984: I think for `Column` we can use `u32` instead of string

crates/fuel-core/src/service/config.rs
```diff
@@ -52,6 +52,7 @@ pub struct Config {
     /// When `true`:
     /// - Enables manual block production.
     /// - Enables debugger endpoint.
+    /// - Enables storage read replay for historical blocks.
```
xgreenx, line 55: We can move this comment to `historical_execution`

crates/services/producer/src/block_producer.rs
```diff
@@ -382,6 +384,34 @@ where
...
+    ) -> anyhow::Result<Vec<StorageReadReplayEvent>> {
+        let view = self.view_provider.latest_view()?;
+
+        let block = view.get_block(&height)?;
+        let transactions = block
+            .transactions()
+            .iter()
+            .map(|id| view.get_transaction(id).map(|tx| tx.into_owned()))
+            .collect::<Result<Vec<_>, _>>()?;
+        let block = block.into_owned().uncompress(transactions);
```
xgreenx, line 409: I think you can a function `get_full_block` and it will be responsibility of the `fuel-core` to provide you a full `Block` type

crates/services/upgradable-executor/src/storage_access_recorder.rs
```diff
@@ -0,0 +1,49 @@
...
+impl<S> KeyValueInspect for StorageAccessRecorder<S>
+where
+    S: KeyValueInspect,
+{
+    type Column = S::Column;
+
+    fn get(&self, key: &[u8], column: Self::Column) -> StorageResult<Option<Value>> {
+        let value = self.storage.get(key, column)?;
+        self.record.lock().push(StorageReadReplayEvent {
+            column: column.name(),
```
xgreenx, line 43: You can use `column.id()`(`u32`) instead of the string

xgreenx (commented):
Looks good to me=)

crates/services/upgradable-executor/src/storage_access_recorder.rs
Dentosal: I think the column names are better for forwards-compatibility, whatever little of that we can have. Do you think the performance matters so much that it makes sense to use u32 here?

crates/services/producer/src/block_producer.rs
Dentosal: I moved it to the trait in 71a6a5607f, but it doesn't look much better. Did you have something else in mind?

crates/fuel-core/src/service/config.rs
Dentosal: 71a6a5607f

crates/services/producer/src/ports.rs
```diff
@@ -34,6 +39,20 @@ pub trait BlockProducerDatabase: Send + Sync {
...
+    /// Gets the full committed block at the `height`.
+    fn get_full_block(&self, height: &BlockHeight) -> StorageResult<Block> {
+        let block = self.get_block(height)?;
+        let transactions = block
+            .transactions()
+            .iter()
+            .map(|id| self.get_transaction(id).map(|tx| tx.into_owned()))
+            .collect::<Result<Vec<_>, _>>()?;
+        Ok(block.into_owned().uncompress(transactions))
+    }
```
xgreenx, line 51: I meant we can move `get_full_block` implementation to the `fuel-core`. Later, when we optimize the getting of the full block, we will avoid iteration over transactions. Plus it will be clear from the business perspective

crates/services/upgradable-executor/src/storage_access_recorder.rs
xgreenx: For backward compatibility we will continue to use the same column, because STF deepens on them. So I think it is better to rely on `u32` in the same way as WASM executor.

Plus, it is less data to serialize.

crates/services/producer/src/block_producer.rs
```diff
@@ -382,6 +384,27 @@ where
...
+    Producer<ViewProvider, TxPool, Executor, GasPriceProvider, ConsensusProvider>
+where
+    ViewProvider: HistoricalView + 'static,
+    ViewProvider::LatestView: BlockProducerDatabase,
+    Executor: ports::StorageReadReplayRecorder + 'static,
+    GasPriceProvider: GasPriceProviderConstraint,
+    ConsensusProvider: ConsensusParametersProvider,
+{
+    /// Re-executes an old block, getting the storage read events.
+    pub async fn storage_read_replay(
```
xgreenx, line 397: Why `storage_read_replay` is an `async` method? It seems all operations are sync inside.

crates/fuel-core/src/service/adapters/graphql_api.rs
```diff
@@ -130,6 +133,13 @@ impl BlockProducerPort for BlockProducerAdapter {
             .dry_run(transactions, height, time, utxo_validation, gas_price)
             .await
     }
+
+    async fn storage_read_replay(
+        &self,
+        height: BlockHeight,
+    ) -> anyhow::Result<Vec<StorageReadReplayEvent>> {
+        self.block_producer.storage_read_replay(height).await
```
xgreenx, line 141: I think we need to add a back pressure here. Or use `tokio_rayon::spawn_fifo` like we do for `dry_run` inside of the `BlockProducer`

crates/client/assets/schema.sdl
```diff
@@ -762,6 +762,10 @@ type Mutation {
 	"""
 	dryRun(txs: [HexString!]!, utxoValidation: Boolean, gasPrice: U64, blockHeight: U32): [DryRunTransactionExecutionStatus!]!
 	"""
+	Get execution trace for an already-executed transaction.
+	"""
+	storageReadReplay(height: U32!): [StorageReadReplayEvent!]!
```
xgreenx, line 767: Why is it a mutation query?

crates/client/assets/schema.sdl
```diff
@@ -762,6 +762,10 @@ type Mutation {
 	"""
 	dryRun(txs: [HexString!]!, utxoValidation: Boolean, gasPrice: U64, blockHeight: U32): [DryRunTransactionExecutionStatus!]!
 	"""
+	Get execution trace for an already-executed transaction.
```
xgreenx, line 765:
```suggestion
	Get execution trace for an already-executed block.
```

crates/services/producer/src/block_producer.rs
Dentosal: Now needed since I use spawn_fifo like dry_run does

crates/fuel-core/src/service/adapters/graphql_api.rs
Dentosal: Right. Done in aaaa57bba6.

crates/services/producer/src/ports.rs
Dentosal: I still don't understand where it would live to be part of `fuel-core`.

xgreenx: In the implementation of the `BlockProducerDatabase` port in `fuel-core`.

image

crates/services/upgradable-executor/src/storage_access_recorder.rs
```diff
@@ -0,0 +1,49 @@
...
+use std::sync::Arc;
+
+#[derive(Clone, Debug)]
+pub struct StorageAccessRecorder<S>
+where
+    S: KeyValueInspect,
+{
+    pub storage: S,
+    pub record: Arc<Mutex<Vec<StorageReadReplayEvent>>>,
+}
```
xgreenx, line 20: Do we even need to derive `Clone`?

crates/types/src/services/executor.rs
```diff
@@ -256,6 +258,27 @@ impl TransactionExecutionResult {
...
+    pub value: Option<Vec<u8>>,
+}
+impl fmt::Debug for StorageReadReplayEvent {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("StorageReadReplayEvent")
+            .field("column", &self.column)
+            .field("key", &hex::encode(&self.key))
+            .field("value", &self.value.as_ref().map(hex::encode))
+            .finish()
+    }
```
xgreenx, line 279: Do we want to decode the whole value? In the case of contract it can be a lot of bytes, maybe we want to limit it with first 16 bytes?

tests/tests/storage_read_replay.rs
```diff
@@ -0,0 +1,62 @@
...
+            .await
+            .expect("Failed to replay storage read");
+
+        // then
+        assert_eq!(value_after, i + 1, "Counter value mismatch");
+
+        let storage_bytes = replay
+            .iter()
+            .find(|item| {
+                item.column == Column::ContractsState as u32 && item.key == storage_slot_key
```
xgreenx, line 49:
```suggestion
                item.column == Column::ContractsState.id() && item.key == storage_slot_key
```

crates/types/src/services/executor.rs
Dentosal: Full debug logs were useful sometimes. But yep, I think it's probably better to not do that now that there's an easy way to get the full values after https://github.com/FuelLabs/fuel-core/pull/2682 is implemented

crates/client/assets/schema.sdl
Dentosal: Ok I'm just going to change this to be a query. Created a follow-up to move dryRun over as well. https://github.com/FuelLabs/fuel-core/issues/2717

crates/client/assets/schema.sdl
Dentosal: b4a1d08bbc

crates/services/upgradable-executor/src/storage_access_recorder.rs
Dentosal: Apparently not. Just removed the derives completely in b4a1d08bbc

crates/services/producer/src/ports.rs
Dentosal: There? Ok sure. Moved in 582a30c573.

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -409,6 +413,124 @@ where
...
+            ReturnType::ExecutionV0(result) => {
+                let _ = convert_from_v0_execution_result(result)?;
+            }
+            ReturnType::ExecutionV1(result) => {
+                let _ = convert_from_v1_execution_result(result)?;
+            }
+            ReturnType::Validation(result) => {
+                let _ = result?;
+            }
+        }
```
xgreenx, line 529: I think for simplicity and performance we can avoid decoding the output type

xgreenx: Also, maybe for performance we want to avoid deserialization of the `output` in the `instance.run`, but it is up to you. Because it required a new method like `run_without_result`

crates/services/upgradable-executor/src/storage_access_recorder.rs
```diff
@@ -0,0 +1,48 @@
...
+use fuel_core_types::services::executor::StorageReadReplayEvent;
+use parking_lot::Mutex;
+use std::sync::Arc;
+
+pub struct StorageAccessRecorder<S>
+where
+    S: KeyValueInspect,
+{
+    pub storage: S,
+    pub record: Arc<Mutex<Vec<StorageReadReplayEvent>>>,
```
xgreenx, line 18: Instead of `Arc<Mutex<_>>` I think you could use `RefCell`. I will make `StorageAccessRecorder` non `Send`, but I don't think that we have cases where you need it to be `Send`. But maybe I'm not right

xgreenx (dismissed):

crates/services/upgradable-executor/src/executor.rs
Dentosal: I think we still need to ensure the execution was successful, and my understanding is that this is the way to do that. I expect ther performance impact of decoding the type to be neglible anyway.

crates/services/upgradable-executor/src/storage_access_recorder.rs
Dentosal: We require it to be `Send` here: https://github.com/FuelLabs/fuel-core/blob/686f8ce24ec8e1e7298dd9b3c671ffe6adbdebc8/crates/services/upgradable-executor/src/instance.rs#L304

tests/tests/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(non_snake_case)]
```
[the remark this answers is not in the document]
Dentosal: No. We want to encourage the use of `__` in tests, and migrating to top-level lint allow in tests helps with that.

crates/services/executor/src/executor.rs
```diff
@@ -1998,9 +1999,9 @@ where
     }
 
     /// Log a VM backtrace if configured to do so
-    fn log_backtrace<M, T, Tx>(
+    fn log_backtrace<M, T, Tx, Ecal>(
         &self,
-        vm: &Interpreter<M, VmStorage<T>, Tx>,
+        vm: &Interpreter<M, VmStorage<T>, Tx, Ecal>,
```
acerone85, line 2004: question for my understanding: where does the additional type parameter come from?

crates/fuel-core/src/service/config.rs
```diff
@@ -56,6 +56,7 @@ pub struct Config {
     pub debug: bool,
     /// When `true`:
     /// - Enables dry run in the past.
+    /// - Enables storage read replay for historical blocks.
     pub historical_execution: bool,
```
acerone85, line 60: is there any case when we would like `historical_execution` without `read_replay`?

crates/fuel-core/src/graphql_api.rs
```diff
@@ -98,6 +99,7 @@ pub const DEFAULT_QUERY_COSTS: Costs = Costs {
     get_peers: 40001,
     estimate_predicates: 40001,
     dry_run: 12000,
+    storage_read_replay: 40001,
```
acerone85, line 102: given the PR description, is this value final or should we there be a TODO with follow up issue to revise it later?

crates/services/producer/src/block_producer.rs
```diff
@@ -382,6 +384,33 @@ where
...
+        let executor = self.executor.clone();
+
+        // use the blocking threadpool to avoid clogging up the main async runtime
+        tokio_rayon::spawn_fifo(move || {
+            let block = view.get_full_block(&height)?;
+            Ok(executor.storage_read_replay(&block)?)
+        })
+        .await
+    }
+}
```
acerone85, line 412: for my own curiosity, I'd like to understand  if I get the motivation behing this code snippet correctly: the operation can be expensive, so we move it to a dedicated thread pool while the runtime is free to keep scheduling other green threads?

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -409,6 +413,124 @@ where
...
+            self.wasm_storage_read_replay(&module, block)
+        }
+    }
+
+    pub fn native_storage_read_replay(
+        &self,
+        block: &Block,
+    ) -> ExecutorResult<Vec<StorageReadReplayEvent>> {
+        let previous_block_height = block.header().height().pred();
+        let relayer = self.relayer_view_provider.latest_view()?;
```
acerone85, line 462: why is the relayer latest view fetched? I would expect the relayer view to be the same  sync with the `block_height`? I.e. it would fetch the same view when it processed the block the first time

crates/services/executor/src/executor.rs
Dentosal: From whoever calls log_backtrace. It's a generic. In some cases you might not want to used the default type. It's unrelated to this PR, but I've added it while doing this, as I like to improve things as I go. It could be removed if you think it's unnecessary.

crates/fuel-core/src/service/config.rs
Dentosal: Not yet, at least. We can decouple them if they ever appear.

crates/fuel-core/src/graphql_api.rs
Dentosal: @xgreenx changed it to this value, maybe he knows. I think this is the final value.

crates/services/executor/src/executor.rs
acerone85: nope, I was just trying to understand. Thanks for explaining

crates/services/producer/src/block_producer.rs
Dentosal: Yes. It's computationally heavy with sparse yield points. We want to run it separately, as otherwise it can pause the main runtime for a long time.

crates/services/upgradable-executor/src/executor.rs
Dentosal: Here, latest relayer view is valid for any height before it. The relayer only adds new information and never removes it, at least for the columns we access here.

acerone85 (approved):
Thanks for explanations and reply to all comments. 
Happy to approve

crates/services/upgradable-executor/src/executor.rs
xgreenx: Usually you re-execute blocks that already were executed and validated in the past, so the result should be correct, unless storage has different state.

Issue #2494, opened 2024-12-11
Shared sequencer: Handle response decode errors


Dentosal:
From [Slack](https://fuellabs.slack.com/archives/C04EWEF3XEZ/p1733912620397979)

> With the new SS URL for testnet () we got these errors on devnet and testnet :
> ```
> fuelcore-authority {"timestamp":"2024-12-11T10:15:18.992963Z","level":"ERROR","fields":{"e":"missing field `gas_info` at line 1 column 222"},"target":"fuel_core_services::service","line_number":373,"span":{"service":"SharedSequencer","name":"initialize_loop"},"spans":[{"name":"fuelcore-authority","name":"new"},{"service":"SharedSequencer","name":"initialize_loop"}]} 
> ```

> ll places where we decode the response can fail and then it would be nice to include the text of the response. In this case we can debug the issue
> ![image](https://github.com/user-attachments/assets/3d9708f4-9dde-4328-a2ef-d4e2d32c30eb)

PR #2495, opened 2024-12-11
Log original text when parsing sequencer response fails

Dentosal:
Hotfix for #2494. Needs to be duplicated for https://github.com/FuelLabs/fuel-core/pull/1922.

## Checklist

### Before requesting review


xgreenx:
Do we need to release it as a `0.40.3`? Looks like it doesn't interrupt our networks, so we can just do this change for `master` and have better logs later, when we migrate to new release based on `master`

xgreenx:
Let's close it for now and release it with the next major release. If it became more critical for the current networks, we can create this release

PR #2501, opened 2024-12-16
Use gas prices from actual blocks to calculate estimate gas prices

MitchTurner:
## Description
Because some nodes won't be running the gas price algorithm, or might not be synced properly, we are going to calculate the price based on the latest block gas price. That is a reliable source of truth.

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -485,7 +485,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
     The change has many minor improvements in different areas related to the state transition bytecode:
     - The state transition bytecode lies in its own file(`state_transition_bytecode.wasm`) along with the chain config file. The `ChainConfig` loads it automatically when `ChainConfig::load` is called and pushes it back when `ChainConfig::write` is called.
     - The `fuel-core` release bundle also contains the `fuel-core-wasm-executor.wasm` file of the corresponding executor version.
-    - The regenesis process now considers the last block produced by the previous network. When we create a (re)genesis block of a new network, it has the `height = last_block_of_old_netowkr + 1`. It continues the old network and doesn't overlap blocks(before, we had `old_block.height == new_genesis_block.height`).
+    - The regenesis process now considers the last block produced by the previous network. When we create a (re)genesis block of a new network, it has the `height = last_block_of_old_network + 1`. It continues the old network and doesn't overlap blocks(before, we had `old_block.height == new_genesis_block.height`).
     - Along with the new block height, the regenesis process also increases the state transition bytecode and consensus parameters versions. It guarantees that a new network doesn't use values from the previous network and allows us not to migrate `StateTransitionBytecodeVersions` and `ConsensusParametersVersions` tables.
     - Added a new CLI argument, `native-executor-version,` that allows overriding of the default version of the native executor. It can be useful for side rollups that have their own history of executor upgrades.
     - Replaced:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2033](https://github.com/FuelLabs/fuel-core/pull/2033): Remove `Option<BlockHeight>` in favor of `BlockHeightQuery` where applicable.
 - [2472](https://github.com/FuelLabs/fuel-core/pull/2472): Added the `amountU128` field to the `Balance` GraphQL schema, providing the total balance as a `U128`. The existing `amount` field clamps any balance exceeding `U64` to `u64::MAX`.
 - [2526](https://github.com/FuelLabs/fuel-core/pull/2526): Add possibility to not have any cache set for RocksDB. Add an option to either load the RocksDB columns families on creation of the database or when the column is used.
+- [2532](https://github.com/FuelLabs/fuel-core/pull/2532): Getters for inner rocksdb database handles.
 
 ### Fixed
 - [2365](https://github.com/FuelLabs/fuel-core/pull/2365): Fixed the error during dry run in the case of race condition.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -473,7 +473,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
     The change has many minor improvements in different areas related to the state transition bytecode:
     - The state transition bytecode lies in its own file(`state_transition_bytecode.wasm`) along with the chain config file. The `ChainConfig` loads it automatically when `ChainConfig::load` is called and pushes it back when `ChainConfig::write` is called.
     - The `fuel-core` release bundle also contains the `fuel-core-wasm-executor.wasm` file of the corresponding executor version.
-    - The regenesis process now considers the last block produced by the previous network. When we create a (re)genesis block of a new network, it has the `height = last_block_of_old_netowkr + 1`. It continues the old network and doesn't overlap blocks(before, we had `old_block.height == new_genesis_block.hegiht`).
+    - The regenesis process now considers the last block produced by the previous network. When we create a (re)genesis block of a new network, it has the `height = last_block_of_old_netowkr + 1`. It continues the old network and doesn't overlap blocks(before, we had `old_block.height == new_genesis_block.height`).
     - Along with the new block height, the regenesis process also increases the state transition bytecode and consensus parameters versions. It guarantees that a new network doesn't use values from the previous network and allows us not to migrate `StateTransitionBytecodeVersions` and `ConsensusParametersVersions` tables.
     - Added a new CLI argument, `native-executor-version,` that allows overriding of the default version of the native executor. It can be useful for side rollups that have their own history of executor upgrades.
     - Replaced:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2446](https://github.com/FuelLabs/fuel-core/pull/2446): Use graphiql instead of graphql-playground due to known vulnerability and stale development.
 - [2379](https://github.com/FuelLabs/fuel-core/issues/2379): Change `kv_store::Value` to be `Arc<[u8]>` instead of `Arc<Vec<u8>>`.
 
+## [Version 0.40.2]
+
+### Fixed
+
+- [2476](https://github.com/FuelLabs/fuel-core/pull/2476): Hardcode the timestamp of the genesis block.
+
+## [Version 0.40.1]
+
+### Added
+
+- [2450](https://github.com/FuelLabs/fuel-core/pull/2450): Added support for posting blocks to the shared sequencer.
+
 ## [Version 0.40.0]
 
 ### Added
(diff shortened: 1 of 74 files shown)


crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -481,10 +506,12 @@ mod tests {
             ),
             None,
         );
+        let latest_gas_price = Arc::new(parking_lot::RwLock::new((0, 0)));
```
rymnc, line 509: curious why we need this when perhaps the algo could contain this internally?

rymnc (commented):
left a few questions :)

crates/services/gas_price_service/src/v1/service.rs
MitchTurner: Yeah... That works too. It's kinda a separate value so I thought to keep it separate but that would work.

crates/fuel-core/src/service/adapters.rs
```diff
@@ -85,6 +86,155 @@ impl StaticGasPrice {
...
 }
 
+#[cfg(test)]
+mod arc_gas_price_estimate_tests {
+    #![allow(non_snake_case)]
+
+    use super::*;
+    use proptest::proptest;
+
+    async fn _worst_case__correctly_calculates_value(
```
acerone85, line 96: nit: this function is called later on
```suggestion
    async fn worst_case__correctly_calculates_value(
```

crates/fuel-core/src/service/adapters.rs
```diff
@@ -85,6 +86,155 @@ impl StaticGasPrice {
...
+            let _ = rt.block_on(subject.worst_case_gas_price(target_height.into()));
+
+            // then
+            // doesn't panic with an overflow
+        }
+    }
+}
+
+#[allow(dead_code)]
+pub struct ArcGasPriceEstimate<Height, GasPrice> {
```
acerone85, line 168: would it be possible to have a bit of documentation here? 
In particular, I'd like to understand what the "best_height" is here

crates/fuel-core/src/service/adapters.rs
```diff
@@ -85,6 +86,155 @@ impl StaticGasPrice {
...
+}
+impl<Height, GasPrice: Copy> ArcGasPriceEstimate<Height, GasPrice> {
+    pub fn get_gas_price(&self) -> GasPrice {
+        self.inner.read().1
+    }
+}
+
+#[async_trait::async_trait]
+impl GasPriceEstimate for ArcGasPriceEstimate<u32, u64> {
+    async fn worst_case_gas_price(&self, height: BlockHeight) -> Option<u64> {
```
acerone85, line 202: so what I get here is that we assume that the price is going to raise by the specified percentage every blocknumber until `height` is reached, is that correct?

crates/fuel-core/src/service/adapters.rs
```diff
@@ -85,6 +86,155 @@ impl StaticGasPrice {
...
+            percentage as u64,
+            height.into(),
+        );
+        Some(worst)
+    }
+}
+
+#[allow(clippy::cast_possible_truncation)]
+pub(crate) fn cumulative_percentage_change(
+    best_gas_price: u64,
```
acerone85, line 219: nit: maybe `start_gas_price` is better?

crates/fuel-core/src/service/adapters.rs
```diff
@@ -85,6 +86,155 @@ impl StaticGasPrice {
...
+    best_height: u32,
+    percentage: u64,
+    target_height: u32,
+) -> u64 {
+    let blocks = target_height.saturating_sub(best_height) as f64;
+    let percentage_as_decimal = percentage as f64 / 100.0;
+    let multiple = (1.0f64 + percentage_as_decimal).powf(blocks);
+    let mut approx = best_gas_price as f64 * multiple;
+    // account for rounding errors and take a slightly higher value
+    const ROUNDING_ERROR_CUTOFF: f64 = 16948547188989277.0;
```
acerone85, line 229: can you put a comment about how this value has been obtained?

crates/fuel-core/src/service/adapters.rs
```diff
@@ -85,6 +86,155 @@ impl StaticGasPrice {
...
+    if approx > ROUNDING_ERROR_CUTOFF {
+        const ROUNDING_ERROR_COMPENSATION: f64 = 2000.0;
```
acerone85, line 231: same as above https://github.com/FuelLabs/fuel-core/pull/2501/files#r1886892073

crates/fuel-core/src/service/adapters.rs
```diff
@@ -85,6 +86,155 @@ impl StaticGasPrice {
...
+        let mut expected = gas_price;
+
+        for _ in 0..block_horizon {
+            let change_amount = expected
+                .saturating_mul(percentage as u64)
+                .saturating_div(100);
+            expected = expected.saturating_add(change_amount);
+        }
+
+        assert!(actual >= expected);
```
acerone85, line 122: I think we can be more precise here, and place an upper bound on actual, e.g.
```suggestion
        assert!(actual >= expected);
        assert!(actual <= expected.saturating_add(2000.0))
```

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -58,12 +56,18 @@ use fuel_gas_price_algorithm::{
...
+    sync::Arc,
+};
 use tokio::sync::broadcast::Receiver;
 
 /// The service that updates the gas price algorithm.
 pub struct GasPriceServiceV1<L2, DA, StorageTxProvider> {
     /// The algorithm that can be used in the next block
     shared_algo: SharedV1Algorithm,
+    /// The latest gas price
+    latest_gas_price: Arc<parking_lot::RwLock<(u32, u64)>>,
```
acerone85, line 70: I wonder if it's best to move the `ArcGasPriceEstimate` type to `fuel-types` or some other places that is accessible from the gas_price_service? 
The implementation of the trait `GasPriceEstimate` can remain in `fuel_core`

acerone85: edit: I guess I missed the point of `ArcGasPriceEstimate`, closing

acerone85:
I'm not sure I understand what the role  of `ArcGasPriceEstimate` is in this PR, since it seems to be used only in tests

crates/fuel-core/src/service/adapters.rs
```diff
@@ -85,6 +86,155 @@ impl StaticGasPrice {
...
+    pub fn get_gas_price(&self) -> GasPrice {
+        self.inner.read().1
+    }
+}
+
+#[async_trait::async_trait]
+impl GasPriceEstimate for ArcGasPriceEstimate<u32, u64> {
+    async fn worst_case_gas_price(&self, height: BlockHeight) -> Option<u64> {
+        let best_gas_price = self.get_gas_price();
+        let best_height = self.get_height();
```
xgreenx, line 204: You can have a race condition here, because between `self.get_gas_price` and `self.get_height` the value can change. You can lock it, read both values, and unlock it again to avoid this issue.

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -105,6 +126,7 @@ where
     pub fn new(
         l2_block_source: L2,
         shared_algo: SharedV1Algorithm,
+        latest_gas_price: Arc<parking_lot::RwLock<(u32, u64)>>,
```
xgreenx, line 129: It would be nice to have a type that hides this lock(`Arc<parking_lot::RwLock<(u32, u64)>>`) inside. Later we can easier refactor it, plus you will not expose direct access to the lock.

crates/services/gas_price_service/src/v1/service.rs
xgreenx: Yeah, maybe we oversplitting here. Algorithm could answer this, if we store the last used gas price there. But I think Mitch wants to reuse the code for v0. But maybe it can live in common and used in both algorithm.

crates/fuel-core/src/service/adapters.rs
MitchTurner: Yes. It's the actual test code, but I don't like writing it inside the `prop_test` macro because the IDE has a hard time parsing it:
```
proptest! {
    #[test]
    fn worst_case__correctly_calculates_value(
        price: u64,
        starting_height: u32,
        block_horizon in 0..10_000u32,
        percentage: u64
    ) {
        _worst_case__correctly_calculates_value(price, starting_height, block_horizon, percentage);
    }
}
```
So this is my naming convention for that case.

crates/fuel-core/src/service/adapters.rs
MitchTurner: Correct.

crates/fuel-core/src/service/adapters.rs
```diff
@@ -85,6 +86,155 @@ impl StaticGasPrice {
...
+        Some(worst)
+    }
+}
+
+#[allow(clippy::cast_possible_truncation)]
+pub(crate) fn cumulative_percentage_change(
+    best_gas_price: u64,
+    best_height: u32,
+    percentage: u64,
+    target_height: u32,
```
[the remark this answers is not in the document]
MitchTurner: I'm fine with this for now.

crates/fuel-core/src/service/adapters.rs
MitchTurner: Sure. But we don't actually care about the lower bound of this. In reality, the gas price given by the block producer will be much lower than what we estimate 99% of the time. So we only really care that the block producer never gives a value higher than what we estimate.

MitchTurner:
> I'm not sure I understand what the role of ArcGasPriceEstimate is in this PR, since it seems to be used only in tests
Yeah. It will be used by the GraphQL service. Just showing that it works here.

xgreenx (dismissed):

xgreenx (dismissed):

CHANGELOG.md
```diff
@@ -484,7 +485,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
     The change has many minor improvements in different areas related to the state transition bytecode:
     - The state transition bytecode lies in its own file(`state_transition_bytecode.wasm`) along with the chain config file. The `ChainConfig` loads it automatically when `ChainConfig::load` is called and pushes it back when `ChainConfig::write` is called.
     - The `fuel-core` release bundle also contains the `fuel-core-wasm-executor.wasm` file of the corresponding executor version.
-    - The regenesis process now considers the last block produced by the previous network. When we create a (re)genesis block of a new network, it has the `height = last_block_of_old_network + 1`. It continues the old network and doesn't overlap blocks(before, we had `old_block.height == new_genesis_block.height`).
+    - The regenesis process now considers the last block produced by the previous network. When we create a (re)genesis block of a new network, it has the `height = last_block_of_old_netowkr + 1`. It continues the old network and doesn't overlap blocks(before, we had `old_block.height == new_genesis_block.height`).
```
rymnc, line 488: merge issue here

rymnc (changes requested):

xgreenx (approved):

rymnc (approved):

Issue #2474, opened 2024-12-05
Add integration tests for balances with (non)retryable messages


rafal-ch:
Currently our integration tests for "balance(s)" endpoint doesn't check if non/retryable messages are correctly handled. We need to improve the coverage.

PR #2505, opened 2024-12-17
Integration test for balances and (non)retryable messages

rafal-ch:
Closes https://github.com/FuelLabs/fuel-core/issues/2474

## Description
This PR changes the following:
1. Adds new test in `tests/tests/relayer.rs` - to ensure that only **non**-retryable messages contribute to the balance returned from `balance()`, `balances()` and `coins_to_spend()`[^1] endpoints
2. Modifies the `balance()` test in `tests/tests/balances.rs` to also include some retryable message which is expected **not** to contribute to the total balance.
   * for `balances()` and `coins_to_spend()` it was already the case

1<sup>st</sup> test checks the behavior of the actual blockchain operations while the 2<sup>nd</sup> tests the proper initialization of the balances index at genesis.

## Checklist
- [X] Breaking changes are clearly marked as such in the PR description and changelog

### Before requesting review

[^1]: currently, `coins_to_spend()` is not using indexation, this will change when [this PR](https://github.com/FuelLabs/fuel-core/pull/2463) is merged.

### DIFF
--- a/tests/tests/relayer.rs
+++ b/tests/tests/relayer.rs
@@ -486,7 +486,7 @@ async fn handle(
 }
 
 #[tokio::test(flavor = "multi_thread")]
-async fn balances_do_not_return_retryable_messages() {
+async fn balances_and_coins_to_spend_never_return_retryable_messages() {
     let mut rng = StdRng::seed_from_u64(1234);
     let mut config = Config::local_node();
     config.relayer = Some(relayer::Config::default());
@@ -496,7 +496,7 @@ async fn balances_do_not_return_retryable_messages() {
     const TIMEOUT: Duration = Duration::from_secs(1);
 
     // Large enough to get all messages, but not to trigger the "query is too complex" error.
-    const UNLIMITED_QUERY: i32 = 100;
+    const UNLIMITED_QUERY_RESULTS: i32 = 100;
 
     // Given
 
@@ -576,7 +576,7 @@ async fn balances_do_not_return_retryable_messages() {
                     &recipient,
                     PaginationRequest {
                         cursor: None,
-                        results: UNLIMITED_QUERY,
+                        results: UNLIMITED_QUERY_RESULTS,
                         direction: PageDirection::Forward,
                     },
                 )
@@ -600,7 +600,7 @@ async fn balances_do_not_return_retryable_messages() {
             None,
             PaginationRequest {
                 cursor: None,
-                results: UNLIMITED_QUERY,
+                results: UNLIMITED_QUERY_RESULTS,
                 direction: PageDirection::Forward,
             },
         )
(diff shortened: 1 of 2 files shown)


tests/tests/relayer.rs
```diff
@@ -477,3 +484,196 @@ async fn handle(
...
+                number_of_blocks: 1,
+            },
+        )
+        .await
+        .unwrap();
+
+    // Balances are processed in the off-chain worker, so we need to wait for it
+    // to process the messages before we can assert the balances.
+    let result = tokio::time::timeout(TIMEOUT, async {
+        loop {
```
rafal-ch, line 573: Is there a better way to ensure that off chain worker has processed the events?

acerone85: I don't think so. Ideally we should be able to register and listen for events from the tracing module.
To me it looks like waiting for the balances to be available is a viable alternative

acerone85 (approved):
Thanks. I could follow the code without problems (comments helped), and checked that tests pass.

tests/tests/relayer.rs
```diff
@@ -477,3 +484,233 @@ async fn handle(
...
+    let messages = vec![
+        MessageKind::Retryable {
+            nonce: RETRYABLE_NONCE,
+            amount: RETRYABLE_AMOUNT,
+        },
+        MessageKind::NonRetryable {
+            nonce: NON_RETRYABLE_NONCE,
+            amount: NON_RETRYABLE_AMOUNT,
+        },
+    ];
```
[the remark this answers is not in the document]
rafal-ch: If recipient is also the same one message will overwritten by the other in DB. See [here](https://github.com/FuelLabs/fuel-core/blob/master/crates/fuel-core/src/graphql_api/worker_service.rs#L225).

A lack of explicit checks suggest that we assume that such message will never come from the relayer.

IIUC, we don't have any special handling if two or more messages have the same nonce but different recipients.

tests/tests/relayer.rs
```diff
@@ -477,3 +484,233 @@ async fn handle(
...
+    let logs: Vec<_> = setup_messages(&messages, &recipient, &contract_address);
+
+    eth_node.update_data(|data| data.logs_batch = vec![logs.clone()]);
+    // Setup the eth node with a block high enough that there
+    // will be some finalized blocks.
+    eth_node.update_data(|data| data.best_block.number = Some(200.into()));
```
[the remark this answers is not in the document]
rafal-ch: None that I know of.

tests/tests/relayer.rs
xgreenx: When you submit transaction and subscribe for its status, you only receive notification when block is processed by the off-chain worker

xgreenx (approved):

PR #2507, opened 2024-12-17
Chore/non breaking cli arguments with sensible defaults

netrome:
### Context
Based on the description in the [DA gas price push doc](https://www.notion.so/fuellabs/DA-Gas-Price-Final-Final-Push-15d2f2293f31805ab4d1fac196b83b38).

### PR description
This PR reverts the `starting_gas_price` => `starting_exec_gas_price` CLI parameter name change to maintain backwards compatibility for existing configurations.

In addition, a few sensible default values are introduced for parameters to ensure the gas price algo is operational for a new node even if these parameters are unset.

### DIFF
--- a/bin/fuel-core/src/cli/run.rs
+++ b/bin/fuel-core/src/cli/run.rs
@@ -206,19 +206,19 @@ pub struct Command {
     pub gas_price_threshold_percent: u8,
 
     /// Minimum DA gas price
-    #[arg(long = "min-da-gas-price", default_value = "10_000_000", env)]
+    #[arg(long = "min-da-gas-price", default_value = "10000000", env)]
     pub min_da_gas_price: u64,
 
     /// P component of DA gas price calculation
     /// **NOTE**: This is the **inverse** gain of a typical P controller.
     /// Increasing this value will reduce gas price fluctuations.
-    #[arg(long = "da-p-component", default_value = "620_090", env)]
+    #[arg(long = "da-p-component", default_value = "620090", env)]
     pub da_p_component: i64,
 
     /// D component of DA gas price calculation
     /// **NOTE**: This is the **inverse** anticipatory control factor of a typical PD controller.
     /// Increasing this value will reduce the dampening effect of quick algorithm changes.
-    #[arg(long = "da-d-component", default_value = "3_528_576", env)]
+    #[arg(long = "da-d-component", default_value = "3528576", env)]
     pub da_d_component: i64,
 
     /// Maximum DA gas price change percent
--- a/bin/fuel-core/src/cli/run.rs
+++ b/bin/fuel-core/src/cli/run.rs
@@ -194,7 +194,7 @@ pub struct Command {
     pub starting_gas_price: u64,
 
     /// The percentage change in gas price per block
-    #[arg(long = "gas-price-change-percent", default_value = "2", env)]
+    #[arg(long = "gas-price-change-percent", default_value = "10", env)]
     pub gas_price_change_percent: u16,
 
     /// The minimum allowed gas price
--- a/bin/fuel-core/src/cli/run.rs
+++ b/bin/fuel-core/src/cli/run.rs
@@ -190,7 +190,7 @@ pub struct Command {
     pub native_executor_version: Option<StateTransitionBytecodeVersion>,
 
     /// The starting execution gas price for the network
-    #[arg(long = "starting-gas-price", default_value = "0", env)]
+    #[arg(long = "starting-gas-price", default_value = "1000", env)]
     pub starting_gas_price: u64,
 
     /// The percentage change in gas price per block
--- a/bin/fuel-core/src/cli/run.rs
+++ b/bin/fuel-core/src/cli/run.rs
@@ -194,7 +194,7 @@ pub struct Command {
     pub starting_gas_price: u64,
 
     /// The percentage change in gas price per block
-    #[arg(long = "gas-price-change-percent", default_value = "0", env)]
+    #[arg(long = "gas-price-change-percent", default_value = "2", env)]
     // https://docs.rs/tokio/latest/tokio/time/fn.pause.html
     pub gas_price_change_percent: u16,
 
@@ -211,10 +211,14 @@ pub struct Command {
     pub min_da_gas_price: u64,
 
     /// P component of DA gas price calculation
+    /// **NOTE**: This is the **inverse** gain of a typical P controller.
+    /// Increasing this value will reduce gas price fluctuations.
     #[arg(long = "da-p-component", default_value = "1", env)]
     pub da_p_component: i64,
 
     /// D component of DA gas price calculation
+    /// **NOTE**: This is the **inverse** anticipatory control factor of a typical PD controller.
+    /// Increasing this value will reduce the dampening effect of quick algorithm changes.
     #[arg(long = "da-d-component", default_value = "0", env)]
     pub da_d_component: i64,
 
--- a/bin/fuel-core/src/cli/run.rs
+++ b/bin/fuel-core/src/cli/run.rs
@@ -195,7 +195,6 @@ pub struct Command {
 
     /// The percentage change in gas price per block
     #[arg(long = "gas-price-change-percent", default_value = "2", env)]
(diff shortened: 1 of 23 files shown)


bin/fuel-core/src/cli/run.rs
```diff
@@ -210,10 +211,14 @@ pub struct Command {
     pub min_da_gas_price: u64,
 
     /// P component of DA gas price calculation
-    #[arg(long = "da-p-component", default_value = "0", env)]
+    /// **NOTE**: This is the **inverse** gain of a typical P controller.
+    /// Increasing this value will reduce gas price fluctuations.
+    #[arg(long = "da-p-component", default_value = "1", env)]
```
netrome, line 216: Note that we're defaulting to maxing out the gain. This would result in spiky algorithm behavior for unconfigured nodes, which may be acceptable. I would ideally like this to be the smallest possible gain, but I fear a `i64::MAX` would likely result in no algo changes after discretization.

Note that this parameter only matters for the authority node, since any other node will use previous blocks to estimate gas prices as per https://github.com/FuelLabs/fuel-core/pull/2501

So the use case as I understand it is someone spinning up their own chain with default configuration, potentially tweaking other parameters.

bin/fuel-core/src/cli/run.rs
```diff
@@ -189,12 +189,13 @@ pub struct Command {
...
-    #[arg(long = "starting-exec-gas-price", default_value = "0", env)]
-    pub starting_exec_gas_price: u64,
+    /// The starting execution gas price for the network
+    #[arg(long = "starting-gas-price", default_value = "0", env)]
+    pub starting_gas_price: u64,
 
     /// The percentage change in gas price per block
-    #[arg(long = "gas-price-change-percent", default_value = "0", env)]
+    #[arg(long = "gas-price-change-percent", default_value = "2", env)]
+    // https://docs.rs/tokio/latest/tokio/time/fn.pause.html
```
acerone85, line 198: a wild comment appeared?

acerone85 (approved):

bin/fuel-core/src/cli/run.rs
rymnc: may I suggest that this PR should be tested with the SDK's which have some e2e tests which run fuel-core under the hood?
I don't think they change any of these values so it will be nice to see the new defaults in action and if it affects their tests

bin/fuel-core/src/cli/run.rs
netrome: Oh don't know how that got there :sweat_smile:

bin/fuel-core/src/cli/run.rs
```diff
@@ -189,12 +189,13 @@ pub struct Command {
...
-    /// The starting gas price for the network
-    #[arg(long = "starting-exec-gas-price", default_value = "0", env)]
-    pub starting_exec_gas_price: u64,
+    /// The starting execution gas price for the network
+    #[arg(long = "starting-gas-price", default_value = "0", env)]
+    pub starting_gas_price: u64,
 
     /// The percentage change in gas price per block
-    #[arg(long = "gas-price-change-percent", default_value = "0", env)]
+    #[arg(long = "gas-price-change-percent", default_value = "2", env)]
```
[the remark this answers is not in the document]
netrome: Sure, 10% just feels like such a huge number. This is already enough to triple the gas price within a minute (assuming 1s block times), and by 59k times within 10 minutes. I don't know how fast we want to be able to respond to changes, but it just feels like a lot.

But I'll set it to 10 to be consistent with our conception.

bin/fuel-core/src/cli/run.rs
```diff
@@ -210,10 +211,14 @@ pub struct Command {
     pub min_da_gas_price: u64,
 
     /// P component of DA gas price calculation
```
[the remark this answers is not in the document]
netrome: All I know is, never ask a guy about his D component... 😄
For reference: https://en.wikipedia.org/wiki/Proportional%E2%80%93integral%E2%80%93derivative_controller

bin/fuel-core/src/cli/run.rs
netrome: Yeah I would love to get the link if you have it. I'll look into the tests tomorrow.

bin/fuel-core/src/cli/run.rs
netrome: Addressed in https://github.com/FuelLabs/fuel-core/pull/2507/commits/3571bba55e3d260981b27063b4f36ad5168b02b1

bin/fuel-core/src/cli/run.rs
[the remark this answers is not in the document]
MitchTurner: Yeah. @xgreenx said 10% as a default. I'm of the opinion that that is high as well, since we have 1s blocks. But as long as we can give reasonable estimates, I don't think quickly changing prices is a bad thing.

bin/fuel-core/src/cli/run.rs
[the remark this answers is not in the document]
MitchTurner: P (proportional) effects how fast it corrects profit "error".  In this case, a larger P value results in a slower change. Essentially it's `p_component = profit / P`

D (Derivative) effects how it responds to the slope (derivative) of the profit. So if it's quickly changing, but the error isn't big, then it should cancel out and potentially lower the gas price. This is so the profit doesn't oscillate, but instead dampens out.

(we don't have the "I" (integral) value, which I think of as the Inertia value, where if it has been a value for a long time, it resists change)

bin/fuel-core/src/cli/run.rs
MitchTurner: > Note that we're defaulting to maxing out the gain. This would result in spiky algorithm behavior for unconfigured nodes, which may be acceptable. I would ideally like this to be the smallest possible gain, but I fear a i64::MAX would likely result in no algo changes after discretization.

I really appreciate this sort of feedback. I've thought about this a bunch, but I'd prefer to get different perspectives, rather than just guessing what would be the "right" behavior.

bin/fuel-core/src/cli/run.rs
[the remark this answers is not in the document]
netrome: "I" gives more than just inertia though. If you just have P and D you may get systematic deviations from the target for some systems. "I" will help break that down, since it is sensitive to longer deviations whereas P and D are "memoryless". Don't think that matters for our system though.

bin/fuel-core/src/cli/run.rs
netrome: Yeah, I don't see any obvious "smart" default values here since this would have to be tuned against other system parameters (max gas, block time etc.) in combination with the expected system load. So defaulting to the max value here would at least make sure we jump in the right direction - and rely on the change percent clamping to not go too far.

bin/fuel-core/src/cli/run.rs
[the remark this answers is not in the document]
netrome: Done in https://github.com/FuelLabs/fuel-core/pull/2507/commits/02724a7ebf690cf9805a2cbb0e015ac0adf29bd2

bin/fuel-core/src/cli/run.rs
netrome: Writing a test now, I think the `FuelCoreDriver` from the integration test helpers will be enough for this.

netrome: After a lot of debugging I now have a working test in https://github.com/FuelLabs/fuel-core/pull/2507/commits/4d0108210995c15fd19719cf761ce9189688ec99. Because the default gas limit is pretty high, this test is a bit slow, and I am hesitant to whether the test adds any value. Otherwise I'm happy to remove it.

rymnc (approved):

rymnc:
why was this pr merged without checks passing?

netrome:
> why was this pr merged without checks passing?

That's on me. It was based on a branch with failing checks, but I could probably have looked harder if this introduced new failures. I see the base branch is green now, so I hope this didn't cause too big inconveniences for you.

PR #2469, opened 2024-11-30
Integrate with V1 algo for tests

MitchTurner:
## Description

In this PR we:
- upgrade the node to use the V1 algorithm for block production and TxPool
- modify the graphql service to use the latest block for gas price estimation
- add some CLI args for the V1 configuration with defaults
- add support for querying block committer API
- improve syncing logic for startup

## Checklist
- [ ] Breaking changes are clearly marked as such in the PR description and changelog
- [ ] New behavior is reflected in tests
- [ ] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [ ] I have reviewed the code myself
- [ ] I have created follow-up issues caused by this PR and linked them here

### After merging, notify other teams

[Add or remove entries as needed]

- [ ] [Rust SDK](https://github.com/FuelLabs/fuels-rs/)
- [ ] [Sway compiler](https://github.com/FuelLabs/sway/)
- [ ] [Platform documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+) (for out-of-organization contributors, the person merging the PR will do this)
- [ ] Someone else?

PR #2523, opened 2024-12-30
chore(gas_price_service_v1): strictly ensure last_recorded_height is set, to avoid initial poll of da source

rymnc:
> [!WARNING]
> Needs to be merged into master after #2469 is merged

## Linked Issues/PRs

- #2469 

## Description

When starting a node, we request  the da source service for the latest costs. This results in large negative profit / loss influencing the da gas price. We want to avoid this behaviour, and therefore, we ensure that the last_recorded_height is set in the block committer.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -182,6 +182,13 @@ where
             storage_tx
                 .set_recorded_height(recorded_height)
                 .map_err(|err| anyhow!(err))?;
+        } else {
+            // we default to the l2 block height
+            // this is done so that we poll the da with a specific height
+            // to avoid initial loss
+            storage_tx
+                .set_recorded_height(height.into())
+                .map_err(|err| anyhow!(err))?;
```
MitchTurner, line 191: Do we want to add some tests around this?

crates/services/gas_price_service/src/v1/da_source_service/block_committer_costs.rs
```diff
@@ -103,7 +101,10 @@ where
                         .get_costs_by_l2_block_number(*next_height.deref())
                         .await?
                 }
-                None => self.client.get_latest_costs().await?.into_iter().collect(),
+                None => {
+                    // prevent calling the latest endpoint
+                    return Err(anyhow!("last_recorded_height is None"));
```
MitchTurner, line 106: Won't this always throw an error on startup until we get an L2 block and call `handle_normal_block`

MitchTurner (commented):
> When starting a node, we request the da source service for the latest costs. This results in large negative profit / loss influencing the da gas price. We want to avoid this behaviour, and therefore, we ensure that the last_recorded_height is set in the block committer.

Yeah. Nice catch. This isn't how it worked before, as we only added costs for blocks that were included in our `unrecorded_blocks`, but looks like we just add it all blindly now:
```
let new_da_block_cost = self
            .latest_known_total_da_cost_excess
            .saturating_add(recording_cost);
        self.latest_known_total_da_cost_excess = new_da_block_cost;
```

I've also considered adding a config value to specify which block we want to start from. That would avoid the Error entirely.

crates/services/gas_price_service/src/v1/da_source_service/block_committer_costs.rs
rymnc: nop, see crates/services/gas_price_service/src/v1/uninitialized_task.rs, we set it before we start the service.

crates/services/gas_price_service/src/v1/service.rs
rymnc: yeah sure, just wanted to make sure we don't break any other tests before writing it

rymnc: addressed in [50e0299](https://github.com/FuelLabs/fuel-core/pull/2523/commits/50e0299bc3ad4906e89c9b3fc6b00ecd1014a0e1)

rymnc: removed that, the test should be for uninitialized task: [b151207](https://github.com/FuelLabs/fuel-core/pull/2523/commits/b151207e5ba165a56e8d660b7ae0d591d9ab3980)

MitchTurner:
If we set the latest_recorded_height to be the same as the l2 block height on startup, then we still have the same problem of paying for old blocks. i.e. if the committer is working on a bundle of 200..300, and we start on 250, then we will still add the costs of 200..249, although they weren't the V1 service's responsibility.
Unfortunately, I don't know how to avoid this. The problem is that we don't know what proportion of the DA bundle any specific "unrecorded block" would represent. That's why we just take the entire bundle's cost blindly.

rymnc:
> If we set the latest_recorded_height to be the same as the l2 block height on startup, then we still have the same problem (if less extreme) of paying for old blocks. i.e. if the committer is working on a bundle of 200..300, and we start on 250, then we will still add the costs of 200..249, although they weren't the V1 service's responsibility.
> Unfortunately, I don't know how to avoid this. The problem is that we don't know what proportion of the DA bundle any specific "unrecorded block" would represent. That's why we just take the entire bundle's cost blindly.

there's no way for us to know the next bundle height deterministically so yeah, no way to avoid fully

MitchTurner (approved):

rymnc:
moving back to draft as there are some things needing discussion

crates/services/gas_price_service/src/v1/da_source_service.rs
```diff
@@ -191,11 +197,12 @@ mod tests {
         let latest_l2_height = latest_l2_height(l2_height);
         let (sender, mut receiver) =
             tokio::sync::broadcast::channel(DA_BLOCK_COSTS_CHANNEL_SIZE);
+        let expected_height = BlockHeight::new(recorded_height);
         let mut service = DaSourceService::new_with_sender(
             da_block_costs_source,
             Some(Duration::from_millis(1)),
             latest_l2_height,
-            None,
+            expected_height,
```
acerone85, line 205: I'm not sure I understand this test. 
You start and run `DaSourceService` with a block_height of`expected_height`, but then after running the service the recorded_height is still the same? This seems to be the opposite of what the test name suggests.

Should it be 
```suggestion
            BlockHeight::new(2),
```
instead?

rymnc: perhaps.. theres a bunch of things that are wrong in this pr now after #2512 was merged.

crates/services/gas_price_service/src/v1/uninitialized_task.rs
```diff
@@ -189,7 +189,11 @@ where
             self.block_stream,
         );
 
-        let recorded_height = self.gas_price_db.get_recorded_height()?;
+        let recorded_height = match self.gas_price_db.get_recorded_height()? {
+            Some(height) => height,
+            None => BlockHeight::from(latest_block_height),
+        };
```
acerone85, line 195: nit (see https://www.lurklurk.org/effective-rust/transform.html)

```suggestion
        let recorded_height = self
            .gas_price_db
            .get_recorded_height()?
            .unwrap_or(BlockHeight::from(latest_block_height));
```

crates/services/gas_price_service/src/v1/da_source_service.rs
acerone85: cool, I didn't see it was back in draft. Will stop reviewing until it's ready for review again

rymnc: addressed in 73cf35beef

crates/services/gas_price_service/src/v1/uninitialized_task.rs
rymnc: addressed in 4ca7a9455f

crates/services/gas_price_service/src/v1/da_source_service.rs
```diff
@@ -178,10 +184,11 @@ mod tests {
 
         // given
         let l2_height = 10;
-        let recorded_height = 9;
+        let starting_recorded_height = 2;
```
MitchTurner, line 187: Shouldn't the `starting_recorded_height` be 2 - 1? We call `succ` on the `recorded_height` to query the DA heights.

crates/services/gas_price_service/src/v1/da_source_service.rs
```diff
@@ -178,10 +184,11 @@ mod tests {
...
+        let expected_recorded_height = 9;
         let unexpected_costs = DaBlockCosts {
```
MitchTurner, line 189: nit: I think this is probably misnamed but that would be my fault.
```suggestion
        let costs = DaBlockCosts {
```

rymnc: addressed in 53a96cfc31

crates/services/gas_price_service/src/v1/da_source_service.rs
rymnc: addressed in 53a96cfc31

MitchTurner (approved):

crates/services/gas_price_service/src/v1/da_source_service/service.rs
```diff
@@ -44,7 +44,7 @@ pub struct DaSourceService<Source> {
     source: Source,
     shared_state: SharedState,
     latest_l2_height: Arc<Mutex<BlockHeight>>,
-    recorded_height: Option<BlockHeight>,
+    recorded_height: BlockHeight,
```
rafal-ch, line 47: Could you please add comments? It not clear at first glance what's the difference between `latest_l2_height` and `recorded_height`. Especially that in other places we also use `latest_recorded_height` which sounds like a mix of both :)

crates/services/gas_price_service/src/v1/uninitialized_task.rs
```diff
@@ -189,7 +189,11 @@ where
             self.block_stream,
         );
 
-        let recorded_height = self.gas_price_db.get_recorded_height()?;
+        let starting_recorded_height = self
```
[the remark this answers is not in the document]
rymnc: addressed in 2df50bf3ca

crates/services/gas_price_service/src/v1/da_source_service/service.rs
rymnc: let me know if this works: 2fe4efe8e9

xgreenx (approved):

xgreenx (dismissed):

rymnc:
github is being weird with the base branch being different. i'm going to merge this into the previous base and reopen the PR with that.

PR #2524, opened 2025-01-02
feat(txpool|p2p): use seqlock instead of small copy-able RwLocks

rymnc:
## Linked Issues/PRs

- none

## Description

We utilize [SeqLock](https://docs.kernel.org/locking/seqlock.html) instead of an RwLock for data that satisfies the following conditions -

1. Occasional writes
2. Frequent reads

The `write` method no longer has a hot loop, and therefore writes aren't blocked by reads. we don't use an internal locking mechanism, but rather a sequence counter to check if data was changed before/during/after a read. This may result in a few iterations while reading, but it is minimal.

We should have a benchmark for txpool to see the performance difference.

It is important to note that this implementation *assumes* that there will only be 1 writer thread.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,126 @@
...
+
+            let data = unsafe { *self.data.get() };
+
+            // check starting/ending guard
+            let end = self.sequence.load(Ordering::Acquire);
+
+            // if value changed, retry
+            if start == end && start % 2 == 0 {
+                return data;
+            }
```
acerone85, line 71: nit: synchronization is required only when the start and end guard coincide. See https://doc.rust-lang.org/std/sync/atomic/fn.fence.html#examples for an example.

(But I suspect we don't need to optimize further re your implementation, so happy to leave it as is)

```suggestion
            let end = self.sequence.load(Ordering::Relaxed);

            // if value changed, retry
            if start == end && start % 2 == 0 {
                std::sync::atomic::fence(Ordering::Acquire)
                return data;
            }
```

acerone85: yep, it's probably an overkill :)

crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,126 @@
...
+        // starting guard
+        self.sequence.fetch_add(1, Ordering::Release);
+
+        // Modify the data
+        unsafe {
+            f(&mut *self.data.get());
+        }
+
+        // ending guard
+        self.sequence.fetch_add(1, Ordering::Release);
```
acerone85, line 46: What happens if two writers try to update the value concurrently?

rymnc: well they shouldn't
this is designed for single writer *only*
there's a lot of foot guns with this implementation, and makes it highly specific for one use case -
1. single writer
2. small data structure
3. heavy reads

I can add as a warning comment if you'd like 

I want to avoid write locks because this assumes only a single writer exists

rymnc: lmk if this comment works: 118c37acba

acerone85: thanks, the comment looks good :)

MitchTurner: Can't we guarantee that only a single writer exists? Like, on construction we just generate a `SeqLockWriter` and from that we can generate any number of `SeqLockReader`s, but it's `SeqLockWriter` isn't cloneable?

Essentially like a single producer, multiple consumer channel.

crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,132 @@
...
+        Ordering,
+    },
+};
+
+/// A simple implementation of a sequential lock.
+/// some usage of unsafe, T must be Copy
+#[derive(Debug)]
+pub struct SeqLock<T> {
+    sequence: AtomicU64,
+    data: UnsafeCell<T>,
```
MitchTurner, line 24: We could create a trait for this `T` to guarantee that it's below 64 bytes and provide some default impls for the types we expect to see.

```rs
/// Type is equal to or smaller than 64 bytes
pub trait Small;
```

netrome: (chiming in) perhaps `Small` should imply `Copy` in this case: `pub trait Small: Copy;`

rymnc: i don't quite like the idea of creating default impls here ~ we won't be able to implement the trait for types outside the crate either due to the orphan rule.

I can enforce the type size at runtime if you'd like.

crates/services/src/seqlock.rs
rymnc: addressed in [720d486](https://github.com/FuelLabs/fuel-core/pull/2524/commits/720d4861e485c4fd5fdfeea93bab540d9f6ec1f0)

crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,158 @@
...
+            if start == end && start % 2 == 0 {
+                return data;
+            }
+        }
+    }
+}
+
+impl<T> SeqLock<T> {
+    /// Creates a new `SeqLock` and returns a writer and a reader handle.
+    #[allow(clippy::new_ret_no_self)]
```
AurelienFT, line 96: I think it's valuable to add the criteria to use this structure here so that the IDE display to the dev that can be more aware

AurelienFT (dismissed):

crates/services/src/seqlock.rs
rymnc: addressed in a51c60c998

AurelienFT (dismissed):

crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,158 @@
...
+    sync::atomic::{
+        AtomicU64,
+        Ordering,
+    },
+};
+
+/// A simple implementation of a sequential lock.
+/// some usage of unsafe, T must be Copy
+#[derive(Debug)]
+pub struct SeqLock<T> {
```
netrome, line 15: I don't think this struct needs to be public. Instead I'd propose just exposing a public function to create the reader and writer, similar to how channels are created.

```suggestion
struct SeqLock<T> {
```

crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+impl<T> SeqLock<T> {
+    /// Creates a new `SeqLock` and returns a writer and a reader handle.
+    /// Optimized for occasional writes and frequent reads
+    ///  !!WARNING!!
+    /// ONLY USE IF ALL THE BELOW CRITERIA ARE MET
+    ///  1. Internal data < 64 bytes
+    ///  2. Data is Copy
+    ///  3. ONLY 1 writer
```
netrome, line 94: I like how we enforce this by having the writer struct not be cloneable.

crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+}
+
+impl<T> SeqLock<T> {
+    /// Creates a new `SeqLock` and returns a writer and a reader handle.
+    /// Optimized for occasional writes and frequent reads
+    ///  !!WARNING!!
+    /// ONLY USE IF ALL THE BELOW CRITERIA ARE MET
+    ///  1. Internal data < 64 bytes
+    ///  2. Data is Copy
```
netrome, line 93: Is it unsafe or just inefficient if the data isn't `Copy`? My hunch is that it's the latter.

I'm also thinking if we should actually add the trait bound on the struct itself. In general we should avoid trait bounds on structs, but the `SeqLock` seems to be meaningless if the data isn't `Copy`. Would there ever be a use case for it with non-Copy data?

netrome (approved):
Overall looks good to me. I'd propose making the `SeqLock` struct private, potentially including the `T: Clone` bound directly on the struct (though I'm not sure of this). I'm okay with the current implementation as well.

crates/services/src/seqlock.rs
rymnc: it is unsafe if the data is not copy because of torn reads

crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,158 @@
...
+
+            // if value changed, retry
+            if start == end && start % 2 == 0 {
+                return data;
+            }
+        }
+    }
+}
+
+impl<T> SeqLock<T> {
```
Dentosal, line 87: If we require that `T: Copy` here, compiler will enforce that for us.

crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// Creates a new `SeqLock` and returns a writer and a reader handle.
+    /// Optimized for occasional writes and frequent reads
+    ///  !!WARNING!!
+    /// ONLY USE IF ALL THE BELOW CRITERIA ARE MET
+    ///  1. Internal data < 64 bytes
```
Dentosal, line 92: If this is required for safety, then the constructor must be marked as `unsafe fn` since it's up to the caller to guarantee it.

crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,158 @@
...
+pub struct SeqLock<T> {
+    sequence: AtomicU64,
+    data: UnsafeCell<T>,
+}
+
+unsafe impl<T: Send> Sync for SeqLock<T> {}
+
+/// The writer handle for the `SeqLock`.
+/// Only one writer exists for a `SeqLock`.
+#[derive(Debug)]
```
Dentosal, line 24: The docs should mention that `Clone` bound is missing to prevent creating multiple writers.

crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,158 @@
...
+    {
+        let lock = &self.lock;
+
+        // Indicate that a write operation is starting.
+        lock.sequence.fetch_add(1, Ordering::Release);
+
+        // Modify the data.
+        unsafe {
+            f(&mut *lock.data.get());
+        }
```
Dentosal, line 43: If F panics, the lock is in invalid state. Since locks are shared between threads, that seems to be an actual issue.

crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,158 @@
...
+    where
+        F: FnOnce(&mut T),
+    {
+        let lock = &self.lock;
+
+        // Indicate that a write operation is starting.
+        lock.sequence.fetch_add(1, Ordering::Release);
+
+        // Modify the data.
+        unsafe {
```
Dentosal, line 41: Needs `// Safety` comment to explain why the `unsafe´ is acceptable.

Dentosal (changes requested):

crates/services/src/seqlock.rs
rymnc: addressed in [6d626b5](https://github.com/FuelLabs/fuel-core/pull/2524/commits/6d626b582438a4923cbfa4a399a03ee22ac6ffee)

crates/services/src/seqlock.rs
rymnc: addressed in [6d626b5](https://github.com/FuelLabs/fuel-core/pull/2524/commits/6d626b582438a4923cbfa4a399a03ee22ac6ffee)

crates/services/src/seqlock.rs
rymnc: addressed in [6d626b5](https://github.com/FuelLabs/fuel-core/pull/2524/commits/6d626b582438a4923cbfa4a399a03ee22ac6ffee)

crates/services/src/seqlock.rs
rymnc: addressed in [6d626b5](https://github.com/FuelLabs/fuel-core/pull/2524/commits/6d626b582438a4923cbfa4a399a03ee22ac6ffee)

crates/services/src/seqlock.rs
rymnc: addressed in 72e09f9beb

AurelienFT (approved):

crates/services/src/seqlock.rs
rymnc: needs to be public because the constructor returns the writer and reader

crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,165 @@
...
+            f(data);
+        }));
+
+        // Indicate that the write operation has finished.
+        lock.sequence.fetch_add(1, Ordering::Release);
+
+        // resume unwinding if there was an error
+        if let Err(e) = result {
+            std::panic::resume_unwind(e);
+        }
```
Dentosal, line 55: After some further reading, I believe this construct is unsound. The compiler is allowed to reorder the non-atomic write to the `UnsafeCell` to occur before the first `fetch_add`, or after the second one. This in turn leads to aliased mut and/or a data race, which is UB.

My first idea on how to fix this would be to use another `AtomicU64` to hold the data, and convert the values using [`std::mem::transmute_copy`](https://doc.rust-lang.org/std/mem/fn.transmute_copy.html) or similar perform conversion. Alternatively [volatile ptr operations](https://doc.rust-lang.org/std/ptr/fn.read_volatile.html) could be used, but [that might have soundness holes as well](https://old.reddit.com/r/rust/comments/eez9s2/writing_a_seqlock_in_rust/fc8jyk0/).

I've reviewed the [seqlock crate](https://docs.rs/seqlock/latest/seqlock/) which looks decent for this, but it doesn't require single writer, which means it's lacking some optimizations that we could have here.

crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub fn read(&self) -> T {
+        let lock = &self.lock;
+
+        loop {
+            // check starting guard
+            let start = lock.sequence.load(Ordering::Acquire);
+
+            // if odd, write in progress
+            if start % 2 != 0 {
+                continue;
```
Dentosal, line 77: We should probably call [`std::thread::yield_now`](https://doc.rust-lang.org/std/thread/fn.yield_now.html) before `continue`. See https://github.com/Amanieu/seqlock/blob/b72653fd9c38141da7e588a97a1b654239f5d4df/src/lib.rs#L126

crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,165 @@
...
+            }
+
+            let data = unsafe { *lock.data.get() };
```
Dentosal, line 80: Missing `// Safety` comment. The unsoundness issue from the write section also applies here.

Dentosal (changes requested):

crates/services/src/seqlock.rs
rymnc: shouldn't the `fetch_add` prevent reordering of `UnsafeCell` writes?

rymnc: curious if using a `fence` would fix this

Dentosal: I believe that the following access pattern is valid (note the first `fetch_add` having `AcqRel` ordering):

```rust
    lock.sequence.fetch_add(1, Ordering::AcqRel);
    std::sync::atomic::fence(Ordering::Acquire);
    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| unsafe {
        let data = &mut *lock.data.get();
        f(data);
    }));
    std::sync::atomic::fence(Ordering::Release);
    lock.sequence.fetch_add(1, Ordering::Release);
    if let Err(e) = result {
        std::panic::resume_unwind(e);
    }
```

crates/services/src/seqlock.rs
Dentosal: Is there a safety reason for the 64 bytes limit? Is it `<` or `<=`?

crates/services/src/seqlock.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+}
+
+impl<T: Copy> SeqLock<T> {
+    /// Creates a new `SeqLock` and returns a writer and a reader handle.
+    /// Optimized for occasional writes and frequent reads
+    ///  !!WARNING!!
+    /// ONLY USE IF ALL THE BELOW CRITERIA ARE MET
+    ///  1. Internal data < 64 bytes
+    ///  2. ONLY 1 writer
```
Dentosal, line 99: This is already ensured by the type, so having it inside `!!WARNING!!` section seems excessive

crates/services/src/seqlock.rs
rymnc: addressed in 51d079ca61

if your main concern is with the compiler reordering then we can use `compiler_fence` instead but that doesn't prevent the hardware from doing so, so I don't mind using `fence` here

crates/services/src/seqlock.rs
rymnc: addressed in 51d079ca61

crates/services/src/seqlock.rs
rymnc: addressed in 51d079ca61

crates/services/src/seqlock.rs
rymnc: addressed in 51d079ca61

crates/services/src/seqlock.rs
rymnc: `<=` fixed in 51d079ca61

yes there is a safety reason, i have added the comments necessary to describe it

AurelienFT (approved):

Dentosal (approved):

PR #2551, opened 2025-01-13
fault_proving(compression): include block_id in da compressed block headers

rymnc:
## Linked Issues/PRs

- closes https://github.com/FuelLabs/fuel-core/issues/2567

## Description

Defines a new version of the `CompressedBlockPayload` which has a header that contains the block_id, relevant for fault proving.

Also refactored the proptest strategy generation and moved the macro internally so that tests can be ran from IDEs easily.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/compression/src/lib.rs
```diff
@@ -48,12 +128,62 @@ impl VersionedCompressedBlock {
     pub fn height(&self) -> &BlockHeight {
         match self {
             VersionedCompressedBlock::V0(block) => block.header.height(),
+            #[cfg(feature = "fault-proving")]
+            VersionedCompressedBlock::V1(block) => &block.header.consensus.height,
+        }
+    }
```
acerone85, line 134: if you are willing to take a performance hit due to dynamic dispatch, the code here can be simplified to have less usages of #[cfg(feature)] and also support new versions with minimal overhead.
 
* add a new `VersionedBlockPayload` trait that exposes the functions required here.
```rust 
trait VersionedBlockPayload {
    fn height(&self) -> &BlockHeight;
    fn consensus_header(&self) -> &ConsensusHeader<Empty>;
    fn application_header(&self) -> &ApplicationHeader<Empty>;
    fn registrations(&self) -> &RegistrationsPerTable;
    fn transactions(&self) -> Vec<CompressedTransaction>;
    fn partial_block_header(&self) -> PartialBlockHeader;
}
```

* Implement the new trait for `CompressedBlockHeaderV0` and `CompressedBlockHeaderV1`

* define a new function as_trait` in `VersionedCompressedBlock` that retrieves the vtable for the variant used, and then refactor the code with 

```rust
impl VersionedCompressedBlock {
   fn as_trait(&self) -> &dyn VersionedBlockPayload {
     match self {
            VersionedCompressedBlock::V0(block) => &block
            #[cfg(feature = "fault-proving")]
            VersionedCompressedBlock::V1(block) => &block
        }
   }
 
   fn height(&self) -> &BlockHeight {
      self.as_trait().height()
   }
   //...
}
```

An example of how to do this is in the branch `andrea-2551-review`.

acerone85:
Added a couple of comments, I don't have any further remarks. 
All my comments are suggestions rather than requests, feel free to ignore if you don't agree

crates/compression/src/lib.rs
netrome: I'm not a fan of dynamic dispatch in Rust. Dynamic dispatch often requires trait objects which impose object safety limitations on any traits implemented for those objects. This can be a big pain - often when mixed with async, so in general I prefer to avoid dynamic dispatch. In many cases it can be replaced with enums for both clearer and more flexible code, and be much more performant at the same time.

rymnc: yeah i thought of dyn dispatch but I would lean over to static dispatch here since most of the types within are `Copy` and it would make it cheaper ~ curious @acerone85 if this is a hard requirement on your side if we can use something like https://docs.rs/enum_dispatch/latest/enum_dispatch/

acerone85: I'm also not a fan of dynamic dispatching, but in this case I would not be concerned about object safety, as we use the trait only to define accessor functions. The point of static dispatching being cheaper is very valid though. 
Overall mine was just a suggestion and definitely not a hard requirement, but the solution with `enum_dispatch` looks very cool :) I'll let it to @rymnc to decide

crates/compression/src/lib.rs
```diff
@@ -31,10 +42,79 @@ pub struct CompressedBlockPayloadV0 {
...
+#[cfg(feature = "fault-proving")]
+#[derive(Debug, Default, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct CompressedBlockPayloadV1 {
+    /// Temporal registry insertions
+    pub registrations: RegistrationsPerTable,
+    /// Compressed block header
+    pub header: CompressedBlockHeader,
+    /// Compressed transactions
+    pub transactions: Vec<CompressedTransaction>,
+}
```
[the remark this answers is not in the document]
rymnc: addressed in [d906b44](https://github.com/FuelLabs/fuel-core/pull/2551/commits/d906b44762128e09bd059c0221c894e2b7387143)

crates/compression/src/lib.rs
rymnc: addressed in [d906b44](https://github.com/FuelLabs/fuel-core/pull/2551/commits/d906b44762128e09bd059c0221c894e2b7387143)

crates/compression/src/lib.rs
```diff
@@ -4,56 +4,87 @@
...
 pub mod ports;
 mod registry;
 
 pub use config::Config;
+use enum_dispatch::enum_dispatch;
 pub use registry::RegistryKeyspace;
 
+use crate::compressed_block_payload::v0::CompressedBlockPayloadV0;
+#[cfg(feature = "fault-proving")]
+use crate::compressed_block_payload::v1::CompressedBlockPayloadV1;
```
Dentosal, line 20: Why do we feature-gate this? The crate should probably just support all versions all the time.

rymnc: this would make my work easier, I can remove the feature gate np

rymnc: addressed in [9e19ab5](https://github.com/FuelLabs/fuel-core/pull/2551/commits/9e19ab582a0e851ff838e798ee3e583a75a1ee46)

rymnc: based on the discussion, we will keep the feature flag until all the sub-issues are sorted out

Dentosal (approved):

acerone85 (approved):

PR #2571, opened 2025-01-15
fault_proving(compression): include commitment to registry in compressed block header

rymnc:
## Linked Issues/PRs

- closes https://github.com/FuelLabs/fuel-core/issues/2568

## Description

We implement a function `id` for `RegistrationsPerTable` which yields a `RegistryId` that is included within the `CompressedBlockHeader`

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/compression/src/registry.rs
+++ b/crates/compression/src/registry.rs
@@ -14,6 +14,9 @@ use fuel_core_types::{
     tai64::Tai64,
 };
 
+#[cfg(feature = "fault-proving")]
+pub type RegistryId = fuel_core_types::fuel_tx::Bytes32;
+
 macro_rules! tables {
     ($($ident:ty: $type:ty),*) => { paste::paste! {
         #[doc = "RegistryKey namespaces"]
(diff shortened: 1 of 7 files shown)


crates/compression/src/registry.rs
```diff
@@ -90,10 +120,25 @@ macro_rules! tables {
...
+                    }
+                )*
+
+                RegistryId(hasher.digest())
+            }
         }
     }};
 }
 
+// Don't change the order of these, it will affect the order of hashing
```
rymnc, line 141: created: https://github.com/FuelLabs/fuel-specs/issues/623

rymnc:
> Are you sure RegistrationsPerTable contains the entire registry? I was under the impression that this is only the new registrations introduced by the block, and we need to commit to the entire registry here.
> 
> In either case, it would be great if we could add a test which compresses a few subsequent blocks and verifies the compression registry ID.

why do we have to commit to the whole registry?

Dentosal:
> Are you sure RegistrationsPerTable contains the entire registry? I was under the impression that this is only the new registrations introduced by the block, and we need to commit to the entire registry here.

Your impression is correct

rymnc:
needs to be reworked. closing for now

Issue #2449, opened 2024-11-22
Fix flaky `one_spawn_single_tasks_works__thread_id_is_different_than_main` test


rafal-ch:
Link to failed build: https://github.com/FuelLabs/fuel-core/actions/runs/11970991093/job/33374927120?pr=2383

Output:
```
--- STDOUT:              fuel-core-services async_processor::tests::one_spawn_single_tasks_works__thread_id_is_different_than_main ---

running 1 test
test async_processor::tests::one_spawn_single_tasks_works__thread_id_is_different_than_main ... FAILED

failures:

failures:
    async_processor::tests::one_spawn_single_tasks_works__thread_id_is_different_than_main

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 26 filtered out; finished in 1.04s


--- STDERR:              fuel-core-services async_processor::tests::one_spawn_single_tasks_works__thread_id_is_different_than_main ---
thread 'async_processor::tests::one_spawn_single_tasks_works__thread_id_is_different_than_main' panicked at crates/services/src/async_processor.rs:179:9:
assertion `left == right` failed
  left: 8
 right: 10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

  Cancelling due to test failure: 3 tests still running
        PASS [   0.007s] fuel-core-services service::tests::double_await_stop_works
        PASS [   1.005s] fuel-core-services async_processor::tests::second_spawn_works_when_first_is_finished
        PASS [  11.010s] fuel-core-services async_processor::tests::executes_10_tasks_for_10_seconds_with_one_thread
        SLOW [> 60.000s] fuel-core-e2e-client::integration_tests works_in_multinode_local_env
        PASS [  75.317s] fuel-core-e2e-client::integration_tests works_in_multinode_local_env
------------
     Summary [  85.922s] 692/1155 tests run: 691 passed (1 slow), 1 failed, 4 skipped
        FAIL [   1.044s] fuel-core-services async_processor::tests::one_spawn_single_tasks_works__thread_id_is_different_than_main
warning: 463/1155 tests were not run due to test failure (run with --no-fail-fast to run all tests)
error: test run failed
Error: Process completed with exit code 100.
```

rafal-ch:
Another instance: https://github.com/FuelLabs/fuel-core/actions/runs/12137093721/job/33839618712?pr=2463

PR #2577, opened 2025-01-15
Update async processor tests

rafal-ch:
Closes https://github.com/FuelLabs/fuel-core/issues/2449
Or at least, tries to :)

## Description
Summary of changes, each in separate commit for clarity. The change that attempts to fix the actual flakiness reported in the issue in pt. 2 below:
1. Modify `one_spawn_single_tasks_works()` to not use `sleep`
2. Updates assertion in `one_spawn_single_tasks_works__thread_id_is_different_than_main()`
    * The original test expected the exact number of worker threads to be spawned (`assert_eq!(unique_thread_ids.len(), number_of_threads);`), however, the `AsyncProcessor` can potentially reuse threads if the tasks finish quick enough.
    * One way to reduce the flakiness would be to bump the sleep, but I think the more clear way is to ensure that _some_ (but not too many) worker threads have been spawned
        * Especially because the aim of the test is not to ensure the number of worker threads, but to check that main thread id was not used
3. Reduces the arbitrary number of tasks in `executes_10_tasks_for_10_seconds_with_one_thread`, just to make the test execute faster
4. Add some clarifying comments
5. Slightly update and rename the `executes_10_tasks_for_2_seconds_with_10_thread()` and `executes_10_tasks_for_2_seconds_with_1_thread()` - to use proper values and better reflect the idea behind the test

### Before requesting review

### DIFF
--- a/crates/services/src/async_processor.rs
+++ b/crates/services/src/async_processor.rs
@@ -272,7 +272,8 @@ mod tests {
 
         // Then
         while broadcast_receiver.recv().await.is_ok() {}
-        // 5 tasks running on 1 thread, each task taking 1 second, should complete in approximately 5 seconds overall.
+        // 5 tasks running on 1 thread, each task taking 1 second,
+        // should complete in approximately 5 seconds overall.
         // Allowing some LEEWAY to account for runtime overhead.
         const LEEWAY: Duration = Duration::from_millis(300);
         assert!(instant.elapsed() < Duration::from_secs(5) + LEEWAY);
--- a/crates/services/src/async_processor.rs
+++ b/crates/services/src/async_processor.rs
@@ -129,9 +129,9 @@ mod tests {
     #[tokio::test]
     async fn one_spawn_single_tasks_works() {
         // Given
-        let number_of_pending_tasks = 1;
+        const NUMBER_OF_PENDING_TASKS: usize = 1;
         let heavy_task_processor =
-            AsyncProcessor::new("Test", 1, number_of_pending_tasks).unwrap();
+            AsyncProcessor::new("Test", 1, NUMBER_OF_PENDING_TASKS).unwrap();
 
         // When
         let (sender, receiver) = tokio::sync::oneshot::channel();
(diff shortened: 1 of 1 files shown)


crates/services/src/async_processor.rs
```diff
@@ -269,29 +272,34 @@ mod tests {
 
         // Then
         while broadcast_receiver.recv().await.is_ok() {}
-        assert!(instant.elapsed() >= Duration::from_secs(10));
+        // 5 tasks running on 1 thread, each task taking 1 second,
+        // should complete in approximately 5 seconds overall.
+        // Allowing some LEEWAY to account for runtime overhead.
+        const LEEWAY: Duration = Duration::from_millis(300);
+        assert!(instant.elapsed() < Duration::from_secs(5) + LEEWAY);
```
AurelienFT, line 279: If the test complete them all in 1 seconds because it spawned too many thread it's an issue I think an other assert could be added to ensure it : 
```suggestion
        assert!(instant.elapsed() < Duration::from_secs(5) + LEEWAY);
        assert!(instant.elapsed() > Duration::from_secs(5));
```
I don't think thr `LEEWAY` is necessary here because it should never take less than 5 seconds in my mind but if you think there is an expected case then i'm ok to add the ` - LEEWAY` in second assert.

AurelienFT (changes requested):
Thanks a lot for taking the time to (try to) stabilize this :)

crates/services/src/async_processor.rs
rafal-ch: I agree with the additional assert, added in this commit: 8443c00eccaef13b6e607685ceb102b0d8a4fad4
However, I used `>=` to cover the super unlikely case that the execution took exactly 5 secs.

rafal-ch: > I don't think thr LEEWAY is necessary here because it should never take less than 5 seconds in my mind but if you think there is an expected case then i'm ok to add the  - LEEWAY in second assert.

Yes, it will never take less that 5 secs. on traditional computers :-)
So the `elapsed` could be, let's say 5.11 sec, and I wanted to check if it's not bigger, let's say, 7 sec, because ultimately the correct behavior of this test is to execute for just slightly more than 5 secs.

I could do
```
assert!(instant.elapsed() >= Duration::from_secs(5));
assert!(instant.elapsed() < Duration::from_secs(6));
```
but I don't like the additional magic number (6) and I thought that the concept of leeway would be more clear.

Let me know if this makes any sense in the context of async processor :)

AurelienFT: yeah yeah i like the leeway in the first assert :)

AurelienFT (approved):

xgreenx (approved):

Issue #2564, opened 2025-01-15
The `height_expiration_txs` can growth infinitely in TxPool


xgreenx:
The https://github.com/FuelLabs/fuel-core/pull/2447 added the `height_expiration_txs` field to the TxPool service. If someone submits a lot of transactions with expiration in the future(like in one year), this information will live forever in the TxPool, potentially consuming a lot of memory.

AurelienFT:
Same problem exists with `time_txs_submitted`

xgreenx:
It doesn't exist with `time_txs_submitted` since the value in are `Time::now()`, and will be removed after TTL is passed - so after 5 minutes=)

AurelienFT:
Right :)

PR #2579, opened 2025-01-16
Update `height_expiration_txs` to not grow indefinitely

AurelienFT:
## Linked Issues/PRs
https://github.com/FuelLabs/fuel-core/issues/2564

## Description
At a cost of some performance, because we now iterate over an `HashSet` instead of a `Vec`, we remove transactions from `height_expiration_txs` when we prune them and when they are inserted on a block which is the two cases where a transaction can be removed. Maybe we should be clearer on the code where the tx are removed and regroup the behaviour in the same place

## Checklist

### Before requesting review

### DIFF
--- a/crates/services/txpool_v2/src/pool.rs
+++ b/crates/services/txpool_v2/src/pool.rs
@@ -525,7 +525,7 @@ where
     /// Remove transaction and its dependents.
     pub fn remove_transaction_and_dependents(
         &mut self,
-        tx_ids: Vec<TxId>,
+        tx_ids: impl Iterator<Item = TxId>,
     ) -> Vec<ArcPoolTx> {
         let mut removed_transactions = vec![];
         for tx_id in tx_ids {
--- a/crates/services/txpool_v2/src/service.rs
+++ b/crates/services/txpool_v2/src/service.rs
@@ -345,12 +345,24 @@ where
                 let expired_txs = height_expiration_txs.remove(&height);
                 if let Some(expired_txs) = expired_txs {
                     let mut tx_pool = self.pool.write();
-                    removed_txs
-                        .extend(tx_pool.remove_transaction_and_dependents(expired_txs));
+                    removed_txs.extend(
+                        tx_pool
+                            .remove_transaction_and_dependents(expired_txs.into_iter()),
+                    );
                 }
             }
         }
         for tx in removed_txs {
+            {
+                let mut height_expiration_txs = self.pruner.height_expiration_txs.write();
+                let expiration = tx.expiration();
+                if expiration < u32::MAX.into() {
+                    if let Some(expired_txs) = height_expiration_txs.get_mut(&expiration)
+                    {
+                        expired_txs.remove(&tx.id());
+                    }
+                }
+            }
             self.shared_state
                 .tx_status_sender
                 .send_squeezed_out(tx.id(), Error::Removed(RemovedReason::Ttl));
@@ -486,7 +498,7 @@ where
                     if expiration < u32::MAX.into() {
                         let mut lock = height_expiration_txs.write();
                         let block_height_expiration = lock.entry(expiration).or_default();
-                        block_height_expiration.push(tx_id);
+                        block_height_expiration.insert(tx_id);
                     }
 
                     let duration = submitted_time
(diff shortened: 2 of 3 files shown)


crates/services/txpool_v2/src/service.rs
```diff
@@ -345,12 +345,24 @@ where
...
+                    removed_txs.extend(
+                        tx_pool
+                            .remove_transaction_and_dependents(expired_txs.into_iter()),
+                    );
                 }
             }
         }
         for tx in removed_txs {
+            {
+                let mut height_expiration_txs = self.pruner.height_expiration_txs.write();
```
rafal-ch, line 357: Could it be better to get the write lock once, before the loop? Only in case when `removed_txs` is not empty.

crates/services/txpool_v2/src/service.rs
```diff
@@ -662,10 +674,20 @@ where
...
+                let expiration = tx.expiration();
+                if expiration < u32::MAX.into() {
+                    if let Some(expired_txs) = height_expiration_txs.get_mut(&expiration)
+                    {
+                        expired_txs.remove(&tx.id());
+                    }
+                }
+            }
```
rafal-ch, line 690: Do you think it's worth to have a test for this additional behavior?

crates/services/txpool_v2/src/service.rs
AurelienFT: Changed :)

crates/services/txpool_v2/src/service.rs
AurelienFT: It would be cool that all tests verify this state at the end and so the existing test will enforce this is working but this require to add a lot of ultis in test-helpers etc so I opened a followup issue : https://github.com/FuelLabs/fuel-core/issues/2581

crates/services/txpool_v2/src/service.rs
rafal-ch: Thanks! I think the final form should actually be something like below, because we don't want to lock if we're not processing anything.
```
        if !removed_txs.is_empty() {
            let mut height_expiration_txs = self.pruner.height_expiration_txs.write();
            for tx in removed_txs {
                let expiration = tx.expiration();
                if expiration < u32::MAX.into() {
                    if let Some(expired_txs) = height_expiration_txs.get_mut(&expiration)
                    {
                        expired_txs.remove(&tx.id());
                    }
                }
                self.shared_state
                    .tx_status_sender
                    .send_squeezed_out(tx.id(), Error::Removed(RemovedReason::Ttl));
            }
        }
```

crates/services/txpool_v2/src/service.rs
```diff
@@ -315,16 +315,29 @@ where
...
+        let removed_transactions = {
             let mut tx_pool = self.pool.write();
-            tx_pool.remove_transaction(executed_transaction);
+            let removed_transactions = tx_pool.remove_transactions(executed_transactions);
             if !tx_pool.is_empty() {
                 self.shared_state.new_txs_notifier.send_replace(());
             }
+            removed_transactions
+        };
+        if !removed_transactions.is_empty() {
```
rafal-ch, line 330: Looking at the `remove_transactions()` shows that if `executed_transactions` is not empty, the `removed_transactions` will also be not empty (since it pushes either the transaction of `None` to the return vector).

What I mean is that we might get a `vec![None, None, None]` from the function and still do the locking even though we'll not process anything.

I might be missing some deeper context, though.

rafal-ch (approved):
LGTM :+1:

crates/services/txpool_v2/src/pool.rs
```diff
@@ -354,7 +354,8 @@ where
 
     /// Remove transaction but keep its dependents.
     /// The dependents become executables.
-    pub fn remove_transaction(&mut self, tx_ids: Vec<TxId>) {
+    pub fn remove_transactions(&mut self, tx_ids: Vec<TxId>) -> Vec<ArcPoolTx> {
```
rymnc, line 357: can this accept an iterator for `tx_ids` too please?
```suggestion
    pub fn remove_transactions(&mut self, tx_ids: impl Iterator<Item = TxId>) -> Vec<ArcPoolTx> {
```

perhaps you might want to return an Iterator for the `ArcPoolTx` too :)

AurelienFT: I find it useful to be able to have the len for the parameter to pre-alloc the result. And I don't really see how to make the result iterator (maybe with a lot of combinators) but it will resolve the pre-alloc problem so if you have an idea feel free to commit here

rymnc: think you can use size_hint() right?

AurelienFT: Ok, I added for the paramter but couldn't do it for return because of lifetimes etc and I think given the fact that we have to bond lifetime of self on the iteraztor returned it's less flexible so I will let the vec.

rafal-ch (dismissed):

rafal-ch:
@AurelienFT do we still want this?

AurelienFT:
@rafal-ch  Yes but it's low prio so paused for now

Issue #2576, opened 2025-01-15
Add information to the `NodeInfo` about indexation availability


xgreenx:
The `node_info` endpoint should allow anyone to know what indexation is enabled and what is not, to allow them to decide the way how they want to request the data.

PR #2550, opened 2025-01-13
Add current pool gas to the node info endpoint

AurelienFT:
## Linked Issues/PRs
/

## Description
Add the information of the statistics of the pool on `node_info` endpoint. Useful for trying to predicting tx block inclusion in some cases

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog
- [x] New behavior is reflected in tests
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [x] I have reviewed the code myself
- [x] I have created follow-up issues caused by this PR and linked them here

### After merging, notify other teams

[Add or remove entries as needed]

- [ ] [Rust SDK](https://github.com/FuelLabs/fuels-rs/)
- [ ] [Sway compiler](https://github.com/FuelLabs/sway/)
- [ ] [Platform documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+) (for out-of-organization contributors, the person merging the PR will do this)
- [ ] Someone else?

PR #2595, opened 2025-01-17
Expose indexation status in `NodeInfo` endpoint

rafal-ch:
Closes https://github.com/FuelLabs/fuel-core/issues/2576

## Description
This PR extends the `nodeInfo` GraphQL endpoint by adding the `indexation` field which allows to check whether a given indexation is enabled. Example response:

```json
{
  "data": {
    "nodeInfo": {
      "utxoValidation": false,
      "indexation": {
        "balances": true,
        "coinsToSpend": false,
        "assetMetadata": false
      }
    }
  }
}
```

### Before requesting review


crates/fuel-core/src/schema/node_info.rs
```diff
@@ -168,3 +202,24 @@ impl TxPoolStats {
         self.0.total_gas.into()
     }
 }
+
+#[derive(Clone)]
+struct Indexation(HashSet<IndexationKind>);
```
AurelienFT, line 207: I think you could use something like [enumset](https://docs.rs/enumset/latest/enumset/) which allow you to store all the data you need on few bytes and I think will be more than enough for the foreseeable future.

With this new structure you could also change the way you are creating you `HashSet` in `node_info` above and avoid any iteration I think

AurelienFT (changes requested):
You also need to change the type on the client side etc... You can see everything I changed on a very similar PR that has land on master few days ago : https://github.com/FuelLabs/fuel-core/pull/2550/files

crates/fuel-core/src/schema/node_info.rs
rafal-ch: I'm not a big fan of adding a new dependency and rely on new data types just to save a byte or two, that's why I prepare this as a [separate PR](https://github.com/FuelLabs/fuel-core/pull/2601) to this branch. If we want to merge this, I'm ok.

AurelienFT: I think we don't really need to use the dependency directly but it was an example for an idea. I will try to do it on a branch of top of this one today/tomorrow.

AurelienFT: Here it is : https://github.com/FuelLabs/fuel-core/pull/2604

The creation of the `Indexation` if a bit more algorithm complex than just : 
```rust
        if read_view.coins_to_spend_indexation_enabled {
            indexation.insert(IndexationKind::CoinsToSpend);
        }
        if read_view.balances_indexation_enabled {
            indexation.insert(IndexationKind::Balances);
        }
        if read_view.asset_metadata_indexation_enabled {
            indexation.insert(IndexationKind::AssetMetadata);
       }
```
but it enforce at compile level that all `InstructionKind` are listed and inserted with a condition. I think that it provide more safety for the codebase and future contributions but up to you :)

rafal-ch: Approved! :+1:

AurelienFT (changes requested):
This still applies : https://github.com/FuelLabs/fuel-core/pull/2595#pullrequestreview-2561769407

AurelienFT (dismissed):

MitchTurner (approved):

crates/fuel-core/src/schema/node_info.rs
```diff
@@ -94,6 +104,28 @@ impl NodeQuery {
...
+                    if read_view.coins_to_spend_indexation_enabled {
+                        indexation.insert(kind);
+                    }
+                }
+                IndexationKind::AssetMetadata => {
+                    if read_view.asset_metadata_indexation_enabled {
+                        indexation.insert(kind);
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
AurelienFT: Done in https://github.com/FuelLabs/fuel-core/pull/2595/commits/58e2cb50423f8e23be94a37d7430c1445b42618c

xgreenx (approved):

AurelienFT (dismissed):

AurelienFT (approved):

MitchTurner (approved):

xgreenx (approved):

Issue #2597, opened 2025-01-17
fault_proving(global_roots): Extend `update_merklized_tables` to also take relayed messages as input.


netrome:
The `update_merklized_tables` function in https://github.com/FuelLabs/fuel-core/pull/2553 only processes blocks, but we also need to get messages from the L1 to populate the `Messages` table.

### Definition of done
The `update_merklized_tables` function has been extended to take relayed messages for the given block height as well.

PR #2598, opened 2025-01-17
fault_proving(global_roots): Initial test suite for merklized storage updates

### LINKED ISSUE #2582: fault_proving(global_roots): Initial test suite for storage updates
Following the initial scaffolding https://github.com/FuelLabs/fuel-core/pull/2553 we should add tests to ensure that the block processing behaves as expected.

More specifically we should add tests that check that the following tables modifications happen:
* Coin insertions.
* Coin deletions.
* Contract UTXO updates.

### Definition of done
We have an initial test suite for the global merkle root storage crate covering the above modifications.

netrome:
Closes #2582 

## Description
This PR adds an initial test suite for the merkelized storage update functionality. These tests ensure that the insertion and deletion behavior for coins (except message coins, to be added in #2597) and contract UTXOs is correct.

### DIFF
--- a/crates/fraud_proofs/global_merkle_root/storage/Cargo.toml
+++ b/crates/fraud_proofs/global_merkle_root/storage/Cargo.toml
@@ -23,7 +23,7 @@ strum = { workspace = true }
 strum_macros = { workspace = true }
 
 [dev-dependencies]
-fuel-core-storage = { workspace = true, features = ["alloc", "test-helpers"] }
+fuel-core-storage = { workspace = true, default-features = false, features = ["alloc", "test-helpers"] }
 rand = { workspace = true }
 
 [features]
--- a/crates/fraud_proofs/global_merkle_root/storage/Cargo.toml
+++ b/crates/fraud_proofs/global_merkle_root/storage/Cargo.toml
@@ -23,7 +23,10 @@ strum = { workspace = true }
 strum_macros = { workspace = true }
 
 [dev-dependencies]
-fuel-core-storage = { workspace = true, default-features = false, features = ["alloc", "test-helpers"] }
+fuel-core-storage = { workspace = true, default-features = false, features = [
+  "alloc",
+  "test-helpers",
+] }
 rand = { workspace = true }
 
 [features]
--- a/crates/fraud_proofs/global_merkle_root/storage/Cargo.toml
+++ b/crates/fraud_proofs/global_merkle_root/storage/Cargo.toml
@@ -22,6 +22,10 @@ num_enum = { workspace = true }
 strum = { workspace = true }
 strum_macros = { workspace = true }
 
+[dev-dependencies]
+fuel-core-storage = { workspace = true, features = ["alloc", "test-helpers"] }
+rand = { workspace = true }
+
 [features]
 default = ["std"]
 std = ["fuel-core-storage/std", "fuel-core-types/std"]
--- a/crates/fraud_proofs/global_merkle_root/storage/src/update.rs
+++ b/crates/fraud_proofs/global_merkle_root/storage/src/update.rs
@@ -342,7 +342,7 @@ mod tests {
         // Given
         let mut storage: InMemoryStorage<Column> = InMemoryStorage::default();
         let mut storage_tx = storage.write_transaction();
-        let mut update_tx = storage_tx.update_transaction();
+        let mut update_tx = storage_tx.construct_update_merkleized_tables_transaction();
 
         let tx_pointer = random_tx_pointer(&mut rng);
         let utxo_id = random_utxo_id(&mut rng);
@@ -386,7 +386,8 @@ mod tests {
         // Given
         let mut storage: InMemoryStorage<Column> = InMemoryStorage::default();
         let mut storage_tx = storage.write_transaction();
-        let mut storage_update_tx = storage_tx.update_transaction();
+        let mut storage_update_tx =
+            storage_tx.construct_update_merkleized_tables_transaction();
 
         let tx_pointer = random_tx_pointer(&mut rng);
         let utxo_id = random_utxo_id(&mut rng);
@@ -428,7 +429,8 @@ mod tests {
         // Given
         let mut storage: InMemoryStorage<Column> = InMemoryStorage::default();
         let mut storage_tx = storage.write_transaction();
-        let mut storage_update_tx = storage_tx.update_transaction();
+        let mut storage_update_tx =
+            storage_tx.construct_update_merkleized_tables_transaction();
 
         let tx_pointer = random_tx_pointer(&mut rng);
         let utxo_id = random_utxo_id(&mut rng);
@@ -477,7 +479,8 @@ mod tests {
         // Given
         let mut storage: InMemoryStorage<Column> = InMemoryStorage::default();
         let mut storage_tx = storage.write_transaction();
-        let mut storage_update_tx = storage_tx.update_transaction();
+        let mut storage_update_tx =
+            storage_tx.construct_update_merkleized_tables_transaction();
 
         let output_amount = rng.gen();
         let output_address = random_address(&mut rng);
@@ -543,17 +546,19 @@ mod tests {
         contract_id
     }
 
-    trait UpdateTransaction<'a>: Sized + 'a {
+    trait ConstructUpdateMerkleizedTablesTransactionForTests<'a>: Sized + 'a {
         type Storage;
-        fn update_transaction(
+        fn construct_update_merkleized_tables_transaction(
             self,
         ) -> UpdateMerkleizedTablesTransaction<'a, Self::Storage>;
     }
 
-    impl<'a, Storage> UpdateTransaction<'a> for &'a mut StorageTransaction<Storage> {
+    impl<'a, Storage> ConstructUpdateMerkleizedTablesTransactionForTests<'a>
+        for &'a mut StorageTransaction<Storage>
+    {
         type Storage = Storage;
 
-        fn update_transaction(
+        fn construct_update_merkleized_tables_transaction(
             self,
         ) -> UpdateMerkleizedTablesTransaction<'a, Self::Storage> {
             UpdateMerkleizedTablesTransaction {
--- a/crates/fraud_proofs/global_merkle_root/storage/src/update.rs
+++ b/crates/fraud_proofs/global_merkle_root/storage/src/update.rs
@@ -505,17 +505,27 @@ mod tests {
             .process_output(tx_pointer, utxo_id, &inputs, &output)
             .unwrap();
 
+        let coin_was_inserted_before_process_input = !storage_update_tx
+            .storage
+            .storage_as_ref::<Coins>()
+            .get(&utxo_id)
+            .unwrap()
+            .is_none();
+
         storage_update_tx.process_input(&input).unwrap();
 
         storage_tx.commit().unwrap();
 
-        // Then
-        assert!(storage
+        let coin_doesnt_exist_after_process_input = storage
             .read_transaction()
             .storage_as_ref::<Coins>()
             .get(&utxo_id)
             .unwrap()
-            .is_none());
+            .is_none();
+
+        // Then
+        assert!(coin_was_inserted_before_process_input);
+        assert!(coin_doesnt_exist_after_process_input);
     }
 
     fn random_utxo_id(rng: &mut impl rand::RngCore) -> UtxoId {
--- a/crates/fraud_proofs/global_merkle_root/storage/src/update.rs
+++ b/crates/fraud_proofs/global_merkle_root/storage/src/update.rs
@@ -505,12 +505,12 @@ mod tests {
             .process_output(tx_pointer, utxo_id, &inputs, &output)
             .unwrap();
 
-        let coin_was_inserted_before_process_input = !storage_update_tx
+        let coin_was_inserted_before_process_input = storage_update_tx
             .storage
             .storage_as_ref::<Coins>()
             .get(&utxo_id)
             .unwrap()
-            .is_none();
+            .is_some();
 
         storage_update_tx.process_input(&input).unwrap();
 
--- a/crates/fraud_proofs/global_merkle_root/storage/src/update.rs
+++ b/crates/fraud_proofs/global_merkle_root/storage/src/update.rs
@@ -64,7 +64,7 @@ use fuel_core_types::{
 };
 
 pub trait UpdateMerkleizedTables: Sized {
-    fn update_merklized_tables(
+    fn update_merkleized_tables(
         self,
         chain_id: ChainId,
         block: &Block,
@@ -75,12 +75,12 @@ impl<Storage> UpdateMerkleizedTables for &mut StorageTransaction<Storage>
 where
     Storage: KeyValueInspect<Column = Column>,
 {
-    fn update_merklized_tables(
+    fn update_merkleized_tables(
         self,
         chain_id: ChainId,
         block: &Block,
     ) -> anyhow::Result<Self> {
-        let mut update_transaction = UpdateMerklizedTablesTransaction {
+        let mut update_transaction = UpdateMerkleizedTablesTransaction {
             chain_id,
             storage: self,
         };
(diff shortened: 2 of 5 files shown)


crates/fraud_proofs/global_merkle_root/storage/src/update.rs
```diff
@@ -65,18 +65,18 @@ use fuel_core_types::{
 
 pub trait UpdateMerkleizedTables {
     fn update_merklized_tables(
-        &mut self,
+        self,
```
MitchTurner, line 68: This is unintuitive to me. I'd expect something that `update`s, to be `mut`, and definitely not consume the struct.

What prompted this change?

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
```diff
@@ -305,6 +305,246 @@ impl TransactionOutputs for Transaction {
...
+            .unwrap()
+            .unwrap()
+            .into_owned();
+
+        // Then
+        assert_eq!(inserted_contract_utxo.utxo_id(), &utxo_id);
+    }
+
+    #[test]
+    fn process_input__should_remove_coin() {
```
MitchTurner, line 459: As someone with little context here, it's not obvious to me what the difference between `process_output__should_insert_coin` and `process_input__should_remove_coin` are. It would be nice to explain the conditions for coin being added or removed in the test name so I know what I'm looking for.

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
acerone85: I was wondering the same, then I have noted that the trait is implemented for `&mut StorageTransaction<Storage>` rather than `StorageTransaction<Storage>`. In practice I guess this changes leaves the responsibility of whether to take ownership or a mutable reference to the implementor of the trait. 

On a side note, the trait is now not object safe anymore (I doubt we'll need object safety, but worth mentioning)

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
```diff
@@ -305,6 +305,246 @@ impl TransactionOutputs for Transaction {
...
+        SeedableRng,
+    };
+
+    #[test]
+    fn process_output__should_insert_coin() {
+        let mut rng = StdRng::seed_from_u64(1337);
+
+        // Given
+        let mut storage: InMemoryStorage<Column> = InMemoryStorage::default();
+        let mut storage_tx = storage.write_transaction();
```
acerone85, line 339: we have a terminology problem here. 
We use the term transaction both for 
1. The InMemoryStorageTransaction
2. The Fuel transactions in a block. 

I feel that in the future this could create a lot of confusion in the code. This is particularly problematic for `Transaction::Update` and `Transaction::Upgrade`. 
Maybe we can find some convention to distinguish between the two? (E.g. `storage_tx` and `fuel_tx`)?.

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
```diff
@@ -305,6 +305,246 @@ impl TransactionOutputs for Transaction {
...
+        let mut update_tx = storage_tx.update_transaction();
```
acerone85, line 340: we have a terminology problem here.
We use the term transaction both for

The InMemoryStorageTransaction
The Fuel transactions in a block.
I feel that in the future this could create a lot of confusion in the code. This is particularly problematic for Transaction::Update (and maybe Transaction::Upgrade). 
I like how you used storage_tx here as it makes it clear. But `update_tx` got me a bit confused. Perhaps we can agree to a convension to prefix all "storage" transactions with `storage_`, and all fuel transactions with `fuel_`?
```suggestion
        let mut storage_tx = storage.write_transaction();
        let mut storate_update_tx = storage_tx.update_transaction();
```

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
```diff
@@ -305,6 +305,246 @@ impl TransactionOutputs for Transaction {
...
+        fn update_transaction(
+            self,
+        ) -> UpdateMerklizedTablesTransaction<'a, Self::Storage> {
+            UpdateMerklizedTablesTransaction {
+                chain_id: ChainId::default(),
+                storage: self,
+            }
+        }
+    }
+}
```
acerone85, line 550: I feel like this trait adds a level of indirection that is not needed. 
I played a bit with it, and I ended up with an implementation where `update_transaction` is moved directly into the `UpdateMerkleizedTables` trait definition: 

Something along these lines: 

```rust
pub trait UpdateMerkleizedTables<'a>
where
    Self: Sized + 'a,
{
    type Target;

    fn update_transaction(
        self,
        chain_id: ChainId,
    ) -> UpdateMerklizedTablesTransaction<'a, Self::Target>;

    fn update_merklized_tables(
        self,
        chain_id: ChainId,
        block: &Block,
    ) -> anyhow::Result<()>
    where
        Self::Target: KeyValueInspect<Column = Column>,
    {
        let mut update_transaction = self.update_transaction(chain_id);

        update_transaction.process_block(block)
    }

    #[cfg(test)]
    fn update_transaction_with_default_chain_id(
        self,
    ) -> UpdateMerklizedTablesTransaction<'a, Self::Target>
    where
        Self::Target: KeyValueInspect<Column = Column>,
    {
        self.update_transaction(ChainId::default())
    }
}

impl<'a, Storage> UpdateMerkleizedTables<'a> for &'a mut StorageTransaction<Storage>
where
    Storage: KeyValueInspect<Column = Column>,
{
    type Target = Storage;
    fn update_transaction(
        self,
        chain_id: ChainId,
    ) -> UpdateMerklizedTablesTransaction<'a, Storage> {
        UpdateMerklizedTablesTransaction {
            chain_id,
            storage: self,
        }
    }
}

pub struct UpdateMerklizedTablesTransaction<'a, Storage> {
    chain_id: ChainId,
    storage: &'a mut StorageTransaction<Storage>,
}
```

all tests will need to use the test-specific method instead of update_transaction.
What do you think?

acerone85 (approved):
My comments are minor, and proposed changes can be tackled in a follow up if accepted. 

On a side note, we should start thinking about testing the whole `process_block` function, e.g. how to generate a list of blocks contianing transactions of all types.

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
netrome: Yeah I am not too happy about this terminology collision either. I like the suggestion, let's use it.

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
netrome: Sure, I'll add some comments. Just restating here, the difference is when a transaction produces a coin as an output, it should be added to the Coins table, since we have created a coin. When a coin is spent as an input, we remove it because it's no longer in the active UTXO set. I'll figure out a nice way to explain this in the test. Thanks for pointing this out. I'd love for the code to be as understandable as possible with as little context as possible.

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
netrome: Yep, I danced a bit around the traits to arrive at something that was ergonomic for the tests and caller. Not sure if this is an artifact of the dance or something we strictly need, but I like the ability to be agnostic towards whether something is implemented for mutable references, immutable references or owned values that we get from taking self directly. This is inspired by how `sqlx` handles their executor trait, which works for immutable connections and mutable transaction objects.

I see @acerone85 has a big proposal restructuring this so I'll have a look at that.

netrome: > On a side note, the trait is now not object safe anymore (I doubt we'll need object safety, but worth mentioning)

This is more of a feature to me as I strongly dislike trait objects 😅 Though I have to admit that it is possible that trait objects are the right choice in certain situations, but in almost all cases I've encountered trait objects they have been misused and could be replaced by a simple enum or just pure generics.

netrome: > Though I have to admit that it is possible that trait objects are the right choice in certain situations

This should only be in situations where you have to interact with user-defined types that you cannot know beforehand, and even then you might likely be better off with just using generics.

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
netrome: Oh interesting. I haven't fully comprehended your proposal yet, but I'll first add the extra documentation Mitch asked for and then look deeper into this and see what I think.

netrome:
> My comments are minor, and proposed changes can be tackled in a follow up if accepted.
> 
> On a side note, we should start thinking about testing the whole `process_block` function, e.g. how to generate a list of blocks contianing transactions of all types.

Thanks, and yes - though I think that's going to be more relevant when we integration test the binary so we might not need to cross that bridge just yet.

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
netrome: Done in https://github.com/FuelLabs/fuel-core/pull/2598/commits/92b1d22811b720d3d25be1a24ef2be847ae27ec0

netrome: Let me know what you think.

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
netrome: I like it, will update. I'll also remind myself if we need the `self` receiver or if we can use the `&mut self` as @MitchTurner commented above. I want to remember some problem with lifetime annotations in the latter case.

netrome: Oh wait, now I remember why I didn't do this. Notice that the `UpdateMerkelizedTablesTransaction` is a private struct. I don't want this one to be exposed outside the module, because from the outside we should just have to do:

```
storage.write_transaction().update_merklized_tables().commit()
```

The purpose of the `UpdateTransaction` trait is just to provide some ergonomics to the test suite. I'm happy to restructure this if you have any ideas, but I want to keep the public interface of the module as minimal as possible.

Does this make sense to you?

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
netrome: Changed this trait to take a mutable receiver in (ref). I agree this is more intuitive. It's less flexible, but I think that's a good thing for this type of interface since we can be pretty specific in how we intend this to be called.

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
acerone85: yeah, that's why I put the comment "Note the change of visibility here", I had the feeling that it might not be wanted. 

In general as it is this PR only adds an additional trait for tests, so I don't see a big issue here. 
Happy to close this thread

acerone85 (dismissed):

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
```diff
@@ -305,6 +307,265 @@ impl TransactionOutputs for Transaction {
...
+    }
+
+    fn random_contract_id(rng: &mut impl rand::RngCore) -> ContractId {
+        let mut contract_id = ContractId::default();
+        rng.fill_bytes(contract_id.as_mut());
+
+        contract_id
+    }
+
+    trait ConstructUpdateMerkleizedTablesTransactionForTests<'a>: Sized + 'a {
```
acerone85, line 550: would it not be better to leverage the BuilderPattern in this case? 

E.g. something like 
```
UpdateMerkleizedTablesTransactionBuilder::new().with_storage(storage).with_chain_id(&chain_id)
```

But probably best to do it in a follow-up and merge this one already :)

netrome: Yes and yes :) But again this is only for tests, so I'm not sure if it's worth it. I'd happily approve a PR if you implement it as a follow-up.

crates/fraud_proofs/global_merkle_root/storage/Cargo.toml
```diff
@@ -22,6 +22,10 @@ num_enum = { workspace = true }
 strum = { workspace = true }
 strum_macros = { workspace = true }
 
+[dev-dependencies]
+fuel-core-storage = { workspace = true, features = ["alloc", "test-helpers"] }
```
rymnc, line 26: i think std is enabled by default for `fuel-core-storage`, might want to use no-default-features here?

rymnc (approved):
lgtm, non blocking comment

acerone85 (approved):

rymnc (approved):

acerone85 (approved):

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
netrome: Gonna get this merged now, but still happy to hear your opinions when you're back. Can make a follow-up if you have any further improvement suggestions.

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
```diff
@@ -305,6 +307,265 @@ impl TransactionOutputs for Transaction {
...
+        let utxo_id = random_utxo_id(&mut rng);
+        let inputs = vec![];
+
+        let output = Output::Coin {
+            to: output_address,
+            amount: output_amount,
+            asset_id: AssetId::zeroed(),
+        };
+
+        let input = Input::CoinSigned(CoinSigned {
```
xgreenx, line 498: I don't see where you inserted this coin. If you do it somewhere already, then you would be nice if you checked here that coin exists in the `Coins` table=)

netrome: It's being inserted in `storage_update_tx.process_output`. I can add an assertion to make this more clear, thanks for highlighting.

netrome: Done in https://github.com/FuelLabs/fuel-core/pull/2598/commits/2457f06e8390914ee9a2530d3e71441f143b3935

xgreenx (dismissed):

acerone85 (approved):

rymnc (approved):

MitchTurner (approved):

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
```diff
@@ -305,6 +307,275 @@ impl TransactionOutputs for Transaction {
...
+
+        // Then
+        assert_eq!(*inserted_coin.amount(), output_amount);
+        assert_eq!(*inserted_coin.owner(), output_address);
+    }
+
+    #[test]
+    /// When encountering a transaction with a contract created output,
+    /// `process_output` should ensure an appropriate contract UTxO is
+    /// populated in the `ContractCreated` table.
```
rafal-ch, line 383: On the contrary to what is said in the comment, the test checks the `ContractsLatestUtxo` table - is this intended?

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
```diff
@@ -305,6 +307,275 @@ impl TransactionOutputs for Transaction {
...
+            .into_owned();
+
+        // Then
+        assert_eq!(inserted_contract_utxo.utxo_id(), &utxo_id);
+    }
+
+    #[test]
+    /// When encountering a transaction with a contract output,
+    /// `process_output` should ensure an appropriate contract UTxO is
+    /// populated in the `ContractCreated` table.
```
rafal-ch, line 425: On the contrary to what is said in the comment, the test checks the `ContractsLatestUtxo` table - is this intended?

rafal-ch (commented):
Some "late to the party" comments :)

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
netrome: Nice observation. Yeah this is just my typo. I was probably thinking about the output type while writing the wrong table name 😅. I'll create a tiny follow-up PR to fix it.

netrome: Addressed in this PR https://github.com/FuelLabs/fuel-core/pull/2703

crates/fraud_proofs/global_merkle_root/storage/src/update.rs
netrome: Addressed in this PR https://github.com/FuelLabs/fuel-core/pull/2703

PR #2599, opened 2025-01-18
bugfix: Improve the `BlockCommitterHttpApi` client to use `url` apis better

MitchTurner:
## Description

Before we were just building a String from the `Url` type, rather than using its apis. That was bad and this is more robust. I feel like it could be more precise than this, but I'm not finding a better way looking at the `url` docs.

### DIFF
--- a/crates/services/gas_price_service/src/v1/da_source_service/block_committer_costs.rs
+++ b/crates/services/gas_price_service/src/v1/da_source_service/block_committer_costs.rs
@@ -131,8 +131,9 @@ impl BlockCommitterApi for BlockCommitterHttpApi {
         // Specific: http://localhost:8080/v1/costs?variant=specific&value=19098935&limit=5
         if let Some(url) = &self.url {
             tracing::debug!("getting da costs by l2 block number: {l2_block_number}");
-            let formatted_url = format!("{url}/v1/costs?variant=specific&value={l2_block_number}&limit={NUMBER_OF_BUNDLES}");
-            let response = self.client.get(formatted_url).send().await?;
+            let path = format!("/v1/costs?variant=specific&value={l2_block_number}&limit={NUMBER_OF_BUNDLES}");
+            let full_path = url.join(&path)?;
+            let response = self.client.get(full_path).send().await?;
             let parsed = response.json::<Vec<RawDaBlockCosts>>().await?;
             Ok(parsed)
         } else {
@@ -147,6 +148,7 @@ mod test_block_committer_http_api {
 
     use super::*;
     use fake_server::FakeServer;
+    use url::Url;
 
     #[test]
     fn get_costs_by_l2_block_number__when_url_is_none__then_returns_empty_vec() {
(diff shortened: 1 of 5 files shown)


crates/services/gas_price_service/src/v1/da_source_service/block_committer_costs.rs
```diff
@@ -128,11 +128,12 @@ impl BlockCommitterApi for BlockCommitterHttpApi {
...
         l2_block_number: u32,
     ) -> DaBlockCostsResult<Vec<RawDaBlockCosts>> {
-        // Specific: http://localhost:8080/v1/costs?variant=specific&value=19098935&limit=5
+        // Specific: http://committer.url/v1/costs?variant=specific&value=19098935&limit=5
         if let Some(url) = &self.url {
             tracing::debug!("getting da costs by l2 block number: {l2_block_number}");
-            let formatted_url = format!("{url}/v1/costs?variant=specific&value={l2_block_number}&limit={NUMBER_OF_BUNDLES}");
-            let response = self.client.get(formatted_url).send().await?;
+            let path = format!("/v1/costs?variant=specific&value={l2_block_number}&limit={NUMBER_OF_BUNDLES}");
+            let full_path = url.join(&path)?;
```
rymnc, line 135: what path was it constructing previously?

MitchTurner: I was getting "//v1".

AurelienFT (approved):

rymnc (dismissed):

crates/services/gas_price_service/src/v1/da_source_service/block_committer_costs.rs
```diff
@@ -128,11 +128,12 @@ impl BlockCommitterApi for BlockCommitterHttpApi {
         &self,
         l2_block_number: u32,
     ) -> DaBlockCostsResult<Vec<RawDaBlockCosts>> {
-        // Specific: http://localhost:8080/v1/costs?variant=specific&value=19098935&limit=5
+        // Specific: http://committer.url/v1/costs?variant=specific&value=19098935&limit=5
         if let Some(url) = &self.url {
             tracing::debug!("getting da costs by l2 block number: {l2_block_number}");
-            let formatted_url = format!("{url}/v1/costs?variant=specific&value={l2_block_number}&limit={NUMBER_OF_BUNDLES}");
-            let response = self.client.get(formatted_url).send().await?;
+            let path = format!("/v1/costs?variant=specific&value={l2_block_number}&limit={NUMBER_OF_BUNDLES}");
```
xgreenx, line 134: I think we have the same issue with SS service, could you also apply it there please?=D

AurelienFT (dismissed):

crates/services/shared-sequencer/src/http_api.rs
```diff
@@ -87,8 +88,10 @@ pub async fn estimate_transaction(
     let request = SimulateRequest {
         tx_bytes: tx_bytes.to_string(),
     };
+    let path = "/cosmos/tx/v1beta1/simulate";
```
xgreenx, line 91: We have this problem in all requests=D 
image

MitchTurner: Ah.. I just found the first one. Didn't look farther.

xgreenx (approved):

AurelienFT (approved):

PR #2603, opened 2025-01-20
Set Latest Recorded Height on startup

MitchTurner:
## Linked Issues/PRs

closes https://github.com/FuelLabs/fuel-core/issues/2613

## Description

This improves the behavior for debugging the code--for long-running sessions it won't change anything. If this isn't set on. startup, it will be delayed each time you restart the node unless it runs long enough to receive DA costs. These changes persists the expectations from the _first_ run.

## Checklist


crates/services/gas_price_service/src/v1/uninitialized_task.rs
```diff
@@ -198,10 +198,15 @@ where
...
         );
 
-        let starting_recorded_height = self
-            .gas_price_db
-            .get_recorded_height()?
-            .unwrap_or(BlockHeight::from(latest_block_height));
+        let starting_recorded_height = match self.gas_price_db.get_recorded_height()? {
+            Some(height) => height,
+            None => {
+                let mut storage_tx = self.gas_price_db.begin_transaction()?;
```
xgreenx, line 204: Where do you commit this transaction? Also, maybe it makes sense to allow customization of the gas price height from CLI? For example: you want to sync with the network from genesis block, but you only want to calculate the gas price starting the latest block height, to avoid huge debt because data was not posted during previous blocks.

MitchTurner: > Where do you commit this transaction?

Hmm. I'll double check that it makes sense. I got the test to pass so I assumed it was okay but you're right. Might be able to move this to the sync section.

>Also, maybe it makes sense to allow customization of the gas price height from CLI? 

Yeah. I think that's a good idea.

xgreenx (commented):
I meant: Maybe we want to have a new CLI argument to allow control of it from outside=D But is can be a separate PR, if you think you don't need this feature for testing

MitchTurner:
> I meant: Maybe we want to have a new CLI argument to allow control of it from outside=D But is can be a separate PR, if you think you don't need this feature for testing

We can add this too. I like the idea. It just wasn't blocking.

MitchTurner:
> I meant: Maybe we want to have a new CLI argument to allow control of it from outside=D But is can be a separate PR, if you think you don't need this feature for testing

https://github.com/FuelLabs/fuel-core/issues/2613

MitchTurner:
Just went ahead and did https://github.com/FuelLabs/fuel-core/issues/2613 too while waiting.

bin/fuel-core/src/cli/run.rs
```diff
@@ -269,6 +269,11 @@ pub struct Command {
     #[arg(long = "da-poll-interval", env)]
     pub da_poll_interval: Option<humantime::Duration>,
 
+    /// The L2 height the Gas Price Service will assume is already recorded on DA
+    /// i.e. If you want the Gas Price Service to look for the costs of block 1000, set to 999
+    #[arg(long = "starting-recorded-height", env)]
```
xgreenx, line 274:
```suggestion
    #[arg(long = "gas-price-starting-height", env)]
```

tests/tests/gas_price.rs
```diff
@@ -995,3 +995,35 @@ async fn sentry__gas_price_estimate__uses_gas_price_from_produced_block() {
...
+#[tokio::test]
+async fn cli__starting_recorded_height_is_set_in_db() {
+    // given
+    let starting_recorded_height: u32 = 1000;
+    let as_str = &starting_recorded_height.to_string();
+    let args = vec![
+        "--debug",
+        "--poa-instant",
+        "true",
+        "--starting-recorded-height",
```
xgreenx, line 1008:
```suggestion
        "--gas-price-starting-height",
```
```suggestion
        "--starting-recorded-height",
```

crates/fuel-core/src/service/config.rs
```diff
@@ -85,6 +85,7 @@ pub struct Config {
     /// The size of the memory pool in number of `MemoryInstance`s.
     pub memory_pool_size: usize,
     pub da_gas_price_factor: NonZeroU64,
+    pub starting_recorded_height: Option<u32>,
```
xgreenx, line 88: You can work directly with `BlockHeight`

xgreenx (approved):

crates/services/gas_price_service/src/v1/uninitialized_task/fuel_storage_unrecorded_blocks.rs
```diff
@@ -37,9 +37,10 @@ impl<S> AsUnrecordedBlocks for S
 where
     S: StorageMutate<UnrecordedBlocksTable, Error = StorageError>,
 {
-    type Wrapper<'a> = FuelStorageUnrecordedBlocks<&'a mut Self>
-        where
-            Self: 'a;
+    type Wrapper<'a>
+        = FuelStorageUnrecordedBlocks<&'a mut Self>
+    where
+        Self: 'a;
```
rafal-ch, line 43: Just curious about this. Caused by an update to `fmt`?

rafal-ch (approved):
LGTM :+1:

PR #2604, opened 2025-01-20
Propose new idea to store the indexation info

AurelienFT:


crates/fuel-core/src/schema/node_info.rs
```diff
@@ -108,25 +106,26 @@ impl NodeQuery {
...
-                    read_view.asset_metadata_indexation_enabled,
-                    IndexationKind::AssetMetadata,
-                ),
-            ]
-            .into_iter()
-            .filter_map(|(enabled, kind)| enabled.then_some(kind)),
-        );
-
+        let mut indexation = Indexation::new();
+        for kind in IndexationKind::iter() {
```
xgreenx, line 110: Very good way to enforce compiler error=D

AurelienFT: :p https://github.com/FuelLabs/fuel-core/pull/2595#discussion_r1922922496

AurelienFT: a bit less optimised but worth the cost imo

xgreenx (approved):

crates/fuel-core/src/schema/node_info.rs
```diff
@@ -204,22 +203,59 @@ impl TxPoolStats {
...
-struct Indexation(HashSet<IndexationKind>);
+struct Indexation(u8);
+
+impl Indexation {
+    pub fn new() -> Self {
+        Self(0)
+    }
+
+    pub fn contains(&self, kind: &IndexationKind) -> bool {
+        self.0 & (1 << *kind as u8) != 0
```
rafal-ch, line 214: Do you think we can do this is a more readable way?

AurelienFT: Mhhh we can highly comment it on the top of the function but this is the only one I'm used to do it and always do it like that so I don't know any other way.

rafal-ch: Maybe you can remind me why we prefer `u8` over `HashSet` [here](https://github.com/FuelLabs/fuel-core/pull/2604/files#diff-8d14f23f382a31131da03da9061317cef7d129f65322d79dd750bc20445506cbR206)?

AurelienFT: Less memory, less algorithmic complexity

rafal-ch: I would rather not trade code readability over minor performance gains, but it's just me :-)

AurelienFT: Personally, I don't see any code readibility issue here, it's a pretty standard pattern but I can understand. This is just a proposition, you can refuse it I'm fine with it.
However, I think the initialization compile-time safe should be ported to your solution because it provide safety measures for the future.

Personal note, I don't think it's "minor performance gains" has we are working on an API that could be triggered thousands of times per seconds but it's a non-sense discussion without real results so I'm fine not accepting this.

AurelienFT: Feel free to add what ever you want from this on your PR and I will close this one.

rafal-ch: No, I think I'll get this as a whole, seeing @xgreenx being ok with that. Not a big deal ;)
Thanks for implementing this :+1:

PR #2610, opened 2025-01-21
The amount of returned dust coins is limited by factor relative to the amount of selected big coins

rafal-ch:
## Description
This PR implements a slight change to the coins to spend algorithm w.r.t how many dust coins can be selected.

Each coins query can result in "free slots." For example, if max is set to 10 and the query is satisfied with 3 coins, 7 slots remain unused.

The previous algorithm randomly added up to 7 dust coins to fill these slots.
The new algorithm limits dust coins to at most 5 times the number of large coins, while still respecting the max limit on total coins.

This will prevent the "dust coin flood", ie. responses that contain a single big coin and 200 dust coins.

## Checklist

### Before requesting review

### DIFF
--- a/crates/fuel-core/src/coins_query.rs
+++ b/crates/fuel-core/src/coins_query.rs
@@ -285,6 +285,20 @@ pub async fn select_coins_to_spend(
     // See also "let upper_target = target.saturating_mul(2);" in "fn random_improve()".
     const TOTAL_AMOUNT_ADJUSTMENT_FACTOR: u64 = 2;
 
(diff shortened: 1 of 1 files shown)


xgreenx (approved):
Would be nice to get approve from @hal3e

rafal-ch:
> Would be nice to get approve from @hal3e

Sure. I've been looking for his GH handle :)
Now added as a reviewer.

hal3e (approved):

PR #2553, opened 2025-01-13
fault_proving(global_roots): scaffold global merkle root storage crate

netrome:
Closes #2552

This PR adds a new crate with column definitions for the global state root service, which we intend to deploy as a standalone binary.

Moreover, this crate also contains a top-level function to populate these columns with data given a new block.

PR #2611, opened 2025-01-21
Use only one metadata table

acerone85:
## Linked Issues/PRs

Each KeyConverter maps a key from table with column id `n` into a 4-byte representation of `n`, hence the ranges of the key-converter functions are disjoint and we could use only one Metadata table. 

If accepted, this PR also corrects two bugs that affected the storage implementation: 
1. The number of total columns is not correct.
2. The Blueprint for MerkleMetadata writes to the MerkleData column. 

## Checklist

### Before requesting review

- [ ] Someone else?


xgreenx (approved):

crates/fraud_proofs/global_merkle_root/storage/src/column.rs
```diff
@@ -61,31 +61,25 @@ impl TableColumn {
...
+    MerkleMetadataColumn,
 }
 
 impl Column {
     /// The total count of variants in the enum.
-    pub const COUNT: usize = TableColumn::COUNT + TableColumn::COUNT;
-
-    /// The start of the merkle metadata columns.
-    pub const MERKLE_METADATA_COLUMNS_START: u32 = u16::MAX as u32;
+    pub const COUNT: usize = TableColumn::COUNT + TableColumn::COUNT + 1;
```
rymnc, line 70: a comment about why the count has been incremented here would be nice :)

rymnc (approved):

rymnc:
(added the no changelog label because this is not going into master)

netrome (approved):
Thanks for fixing!

crates/fraud_proofs/global_merkle_root/storage/src/column.rs
netrome: Yeah I get why this looks confusing. This is actually two things:

1. The previous value was incorrect and should be `TableColumn::COUNT * 3` since we tripled the columns because each column had a data column and a metadata column. Now this PR changes so we reuse the metadata column for all columns, hence the last `TableColumn::COUNT` is replaced with `1`. We should add a comment about this, I'll do it in the base PR.

netrome:
@acerone85 Thanks for fixing this. Since this is so small I'll just cherry-pick this change to the base PR https://github.com/FuelLabs/fuel-core/pull/2553

acerone85:
Closing as changes have been cherry-picked.

PR #2612, opened 2025-01-21
Use latest gas price to estimate next price for tx pool checks

MitchTurner:
## Description

There are 3 services that rely on the gas price: producer, tx pool, and graphql. In 0.41.0, graphql used the latest block, but the producer and tx pool use the actual next gas price.

This works great on the producer node, but for all the sentries that also have tx pools they need to be perfectly in sync with the block producer or else the expected gas prices will be different.

The new change has the tx pool behave the same as graphql, where it has access to the latest block and can calculate the (worst case) gas price for the next block.

This is already what the end users are doing to set a max fee on their tx, so it shouldn't exclude any txs that should be in there.

Sentries will reject txs if they don't have the correct parameters, this solves this problem by using the "worst case" for next gas price based on the the latest actual block. 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2551](https://github.com/FuelLabs/fuel-core/pull/2551): Enhanced the DA compressed block header to include block id.
 
 ### Fixed
+- [2612](https://github.com/FuelLabs/fuel-core/pull/2612): Use latest gas price to estimate next block gas price in tx pool instead of using algorithm directly
 - [2599](https://github.com/FuelLabs/fuel-core/pull/2599): Use the proper `url` apis to construct full url path in `BlockCommitterHttpApi` client
 
 ## [Version 0.41.0]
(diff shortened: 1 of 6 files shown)


AurelienFT (dismissed):

xgreenx (dismissed):
What gas price do we use during dry run as a default gas price?

MitchTurner:
> What gas price do we use during dry run as a default gas price?

It uses the same value as the producer, unfortunately:
```rs
        let gas_price = if let Some(inner) = gas_price {
            inner
        } else {
            self.calculate_gas_price().await?
        };
```
Does anyone use the default value?

xgreenx:
We provide this functionality, so we need to guarantee it to work=D Looks like it uses gas price service. I think we need to change it to use value based on the previous block

MitchTurner:
> We provide this functionality, so we need to guarantee it to work=D Looks like it uses gas price service. I think we need to change it to use value based on the previous block

K. I'll create a PR.

xgreenx (approved):

rafal-ch (approved):
LGTM :+1:

PR #2618, opened 2025-01-22
Parallelise block/transaction changes creation in Importer

AurelienFT:
## Description

Parallelise block/transaction changes creation in Importer. This can allow us to execute the transactions in the same time as the rest of the data needed to commit the block to the data is encoded in a `Changes` form.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog
- [x] New behavior is reflected in tests
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [x] I have reviewed the code myself
- [x] I have created follow-up issues caused by this PR and linked them here

PR #2619, opened 2025-01-22
Add possibility to submit a list of changes to rocksdb

AurelienFT:
## Description

The goal of this PR is to allow to insert a list of changes instead of one `HashMap` of changes. This avoid multiple update produced by different threads for example, to be merged in one transaction.

By removing this step we remove a whole iteration and insert/deletion in a big `HashMap` of changes.

It's currently not used anywhere in the code but you can see a usage example here : https://github.com/FuelLabs/fuel-core/pull/2618

In this PR when we have to write history modifications we are still merging the changes in one big transaction before calling `store_modifications_history` because I don't think it's possible to use this algorithm with changes splitted in different parts. @acerone85 if you have an idea I would be happy to here it but in any case I think this should go to follow-up PR to not overcharge this one. (I made an attempt in branch [try_adapt_history_vec branch](https://github.com/FuelLabs/fuel-core/tree/try_adapt_history_vec) but it's mehhh)

## Checklist

### Before requesting review


AurelienFT:
Works but draft until mature enough to fit requirements of : https://github.com/FuelLabs/fuel-core/pull/2618

crates/storage/src/iter/changes_iterator.rs
```diff
@@ -88,16 +141,62 @@ where
...
+                        get_insert_keys_from_changes(
+                            changes,
+                            column,
+                            prefix.as_deref(),
+                            start.as_deref(),
+                            direction,
+                        )
+                    })
+                    .into_boxed()
+            }
```
MitchTurner, line 170: IDK if this would be better or worse, but it avoids the `to_vec`s :P
```suggestion
            StorageChanges::ChangesList(changes_list) => {
                let column = column.id();
                changes_list
                    .iter()
                    .flat_map(move |changes| {
                        get_insert_keys_from_changes(
                            changes, column, prefix, start, direction,
                        )
                    })
                    .collect::<Vec<_>>()
                    .into_iter()
                    .into_boxed()
            }
```

AurelienFT: mehhh, i don't really like that because start and prefix are pretty small in most of the cases where the number of elements iterated on could be pretty big so iterating twice and creating a heap allocation with all elements will be more costly than 2 little allocation

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2678](https://github.com/FuelLabs/fuel-core/pull/2678): Removed public accessors for `BlockHeader` fields and replaced with methods instead, moved `tx_id_commitment` to the application header of `BlockHeaderV2`.
 
 ### Added
+- [2619](https://github.com/FuelLabs/fuel-core/pull/2691): Add possibility to submit list of changes to rocksdb.
+- [2635](https://github.com/FuelLabs/fuel-core/pull/2635): Add metrics to gas price service
```
AurelienFT, line 15:
```suggestion
```

xgreenx (approved):
Thank you, very clean change!=)

AurelienFT:
> Thank you, very clean change!=)

Thanks to you for the improvments <3

rymnc (approved):

PR #2310, opened 2024-10-07
Add more metrics

rafal-ch:
Closes https://github.com/FuelLabs/fuel-core/issues/807

## Description
This PR adds a couple of additional metrics (block importer and p2p) and also contains slight refactor of how we initialize bucket sized for histogram based metrics.

The `metrics` command-line parameter has been replaced with `disable-metrics`. Metrics are now enabled by default, with the option to disable them entirely or on a per-module basis. This change is _breaking_ for all dependencies that use CLI to setup the `fuel-core-client`

```
      --disable-metrics <METRICS>
          Disables all metrics, or specify a comma-separated list of modules to disable metrics for specific ones. Available options: importer, p2p, producer, txpool, graphql
          
          [env: METRICS=]
          [default: ]
```

Startup logs also show the metrics config:
```
2024-10-14T20:17:37.536840Z  INFO fuel_core_bin::cli::run: 308: Metrics config: Disable modules: txpool
```

## Checklist
- [X] Breaking changes are clearly marked as such in the PR description and changelog
- [X] New behavior is reflected in tests

### Before requesting review
- [X] I have reviewed the code myself

PR #2635, opened 2025-01-26
Add metrics to gas price service

MitchTurner:
## Description

A lot of the data that effects the gas price is somewhat opaque. Up until now, it's been hard to read without direct access to the Gas Price DB. This will give us visibility into:
```rs
pub struct GasPriceMetrics {
    pub real_gas_price: Gauge,
    pub exec_gas_price: Gauge,
    pub da_gas_price: Gauge,
    pub total_reward: Gauge,
    pub total_known_costs: Gauge,
    pub predicted_profit: Gauge,
    pub unrecorded_bytes: Gauge,
    pub latest_cost_per_byte: Gauge,
    pub recorded_height: Gauge,
}
```
which allows us to make charts like (plz ignore the gaps, this was just testing the visualizations):
image

### Before requesting review

### DIFF
--- a/crates/services/gas_price_service/src/v1/service.rs
+++ b/crates/services/gas_price_service/src/v1/service.rs
@@ -260,14 +260,15 @@ where
     ) -> anyhow::Result<()> {
         let capacity = Self::validate_block_gas_capacity(block_gas_capacity)?;
         let mut storage_tx = self.storage_tx_provider.begin_transaction()?;
-        let mut new_recorded_height = match storage_tx
+        let (old_recorded_height, mut new_recorded_height) = match storage_tx
             .get_recorded_height()
             .map_err(|err| anyhow!(err))?
         {
-            Some(_) => None,
+            Some(old) => (Some(old), None),
             None => {
                 // Sets it on first run
-                self.initial_recorded_height.take()
+                let initial = self.initial_recorded_height.take();
+                (initial, initial)
             }
         };
 
@@ -310,13 +311,18 @@ where
         let new_algo = self.algorithm_updater.algorithm();
         tracing::debug!("Updating gas price: {}", &new_algo.calculate());
         self.shared_algo.update(new_algo).await;
-        Self::record_metrics(&metadata, gas_price);
+        let best_recorded_height = new_recorded_height.or(old_recorded_height);
+        Self::record_metrics(&metadata, gas_price, best_recorded_height);
         // Clear the buffer after committing changes
         self.da_block_costs_buffer.clear();
         Ok(())
     }
 
-    fn record_metrics(metadata: &UpdaterMetadata, gas_price: u64) {
+    fn record_metrics(
+        metadata: &UpdaterMetadata,
+        gas_price: u64,
+        recorded_height: Option<BlockHeight>,
+    ) {
         if let UpdaterMetadata::V1(v1_metadata) = metadata {
             let metrics = gas_price_metrics();
             let real_gas_price_i64 = gas_price.try_into().unwrap_or(i64::MAX);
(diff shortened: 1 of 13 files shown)


crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -303,12 +309,61 @@ where
         AtomicStorage::commit_transaction(storage_tx)?;
         let new_algo = self.algorithm_updater.algorithm();
         tracing::debug!("Updating gas price: {}", &new_algo.calculate());
-        self.shared_algo.update(new_algo);
+        self.shared_algo.update(new_algo).await;
```
rymnc, line 312: why await here?

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -303,12 +309,61 @@ where
...
+        let best_recorded_height = new_recorded_height.or(old_recorded_height);
+        Self::record_metrics(&metadata, gas_price, best_recorded_height);
         // Clear the buffer after committing changes
         self.da_block_costs_buffer.clear();
         Ok(())
     }
 
+    fn record_metrics(
```
rymnc, line 320: we should conditionally export metrics, see description of https://github.com/FuelLabs/fuel-core/pull/2310 for reference

crates/services/gas_price_service/src/v1/service.rs
MitchTurner: nvm need to pull

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -251,17 +255,19 @@ where
...
         {
-            Some(_) => None,
+            Some(old) => (Some(old), None),
             None => {
                 // Sets it on first run
-                self.initial_recorded_height.take()
+                let initial = self.initial_recorded_height.take();
+                (initial, initial)
             }
         };
```
rymnc, line 272: do we need this change?

MitchTurner: There might be a better way of expressing this. It's to give the value here:
```rs
        let best_recorded_height = new_recorded_height.or(old_recorded_height);
```

crates/services/gas_price_service/src/v1/service.rs
```diff
@@ -304,11 +314,60 @@ where
...
+        gas_price: u64,
+        recorded_height: Option<BlockHeight>,
+    ) {
+        if let UpdaterMetadata::V1(v1_metadata) = metadata {
+            let metrics = gas_price_metrics();
+            let real_gas_price_i64 = gas_price.try_into().unwrap_or(i64::MAX);
+            let exec_gas_price_i64 = v1_metadata
+                .new_exec_gas_price()
+                .try_into()
+                .unwrap_or(i64::MAX);
```
netrome, line 335: Nit: I generally prefer explicit conversions for readability. I.e. `i64::try_from()`.

I'm also uncertain if it makes sense to report `i64::MAX` if we're out of range. Wouldn't it be better to not report any metrics and log an error in this case?

netrome (approved):
Looks mostly good to me. I'm a bit concerned about returning `i64::MAX` when we're out of range. I get that this should never happen, but I'd rather see an error or warning log in this scenario rather than reporting an invalid value.

xgreenx (approved):

xgreenx:
Just a question: Are you sure that it is enough information for you to understand the all internal state of the algorithm?

Dentosal (approved):

MitchTurner:
> Just a question: Are you sure that it is enough information for you to understand the all internal state of the algorithm?
I think this is enough. I was missing the `recorded_height`, but added that.

Issue #2488, opened 2024-12-09
Fix RUSTSEC-2024-0421


AurelienFT:
Used by `hickory-proto` used in libp2p waiting for a new libp2p release to fix it.

AurelienFT:
Update : https://github.com/hickory-dns/hickory-dns/issues/2661#issuecomment-2531817129 need an update from libp2p

AurelienFT:
PR merged : https://github.com/libp2p/rust-libp2p/pull/5727 but we need to wait a version

jxs:
Hi @AurelienFT we plan to do it soon'ish, see https://github.com/libp2p/rust-libp2p/discussions/5733

AurelienFT:
Hi @jxs, no pressure at all don't worry. It's very kind to come to give us updates here. Thanks a lot and good luck with the libp2p dev :)

PR #2644, opened 2025-01-28
feat: add pending connection limits + bump libp2p to 0.54.1

MitchTurner:
## Checklist

### Before requesting review

- [ ] Someone else?


AurelienFT:
@MitchTurner  Can we update to libp2p 0.55 directyl it would fix : https://github.com/FuelLabs/fuel-core/issues/2488

MitchTurner:
> @MitchTurner Can we update to libp2p 0.55 directyl it would fix : https://github.com/FuelLabs/fuel-core/issues/2488

@AurelienFT I'd be happy to, but it looks like it requires Rust v1.83. So maybe we wait until we update the Rust version?

rymnc (commented):
will not approve, will allow others to review

crates/services/p2p/src/p2p_service.rs
```diff
@@ -293,14 +301,20 @@ impl FuelP2PService {
             network_codec: codec,
             outbound_requests_table: HashMap::default(),
             inbound_requests_table: HashMap::default(),
+            gossipsub_peer_limiter: SizedHashset::new(
+                config.max_gossipsub_peers_connected as usize,
+            ),
+            request_response_peer_limiter: SizedHashset::new(
+                config.max_request_response_peers_connected as usize,
+            ),
```
netrome, line 309: You probably already remember what I think of `as` conversions in production code 🫠

```suggestion
            gossipsub_peer_limiter: SizedHashset::new(
                usize::try_from(config.max_gossipsub_peers_connected)?,
            ),
            request_response_peer_limiter: SizedHashset::new(
                usize::try_from(config.max_request_response_peers_connected)?,
            ),
```

crates/services/p2p/src/p2p_service.rs
```diff
@@ -293,14 +301,20 @@ impl FuelP2PService {
...
             network_metadata,
             metrics,
             libp2p_metrics_registry,
             peer_manager: PeerManager::new(
                 reserved_peers_updates,
                 reserved_peers,
                 connection_state_writer,
-                config.max_peers_connected as usize,
+                config.max_discovery_peers_connected as usize,
```
netrome, line 317:
```suggestion
                `usize::try_from(config.max_discovery_peers_connected`)?,
```

crates/services/p2p/src/behavior.rs
```diff
@@ -35,6 +39,10 @@ use libp2p::{
     PeerId,
 };
 
+const MAX_PENDING_INCOMING_CONNECTIONS: u32 = 100;
+const MAX_PENDING_OUTGOING_CONNECTIONS: u32 = 100;
+const MAX_ESTABLISHED_CONNECTIONS: u32 = 1000;
```
netrome, line 44: Should we add configuration variables for these constants as well? I'm also not completely following the difference between `MAX_ESTABLISHED_CONNECTIONS` and the existing configuration values. Shouldn't this just be the sum of the `max_(discovery|gossipsub|request_response)_peers_connected` parameters?

netrome (changes requested):
Looks mostly good to me. The only blocking thing I've spotted are some `as` conversions that could potentially panic.

crates/services/p2p/src/behavior.rs
rymnc: nope, my reasoning for keeping them non-configurable is to prevent footguns by the node operator

netrome (dismissed):
Thanks for updating! :pray:

tests/tests/peering.rs
```diff
@@ -0,0 +1,122 @@
...
+        let latest_block_height = client
+            .chain_info()
+            .await
+            .unwrap()
+            .latest_block
+            .header
+            .height;
+        println!("latest_block_height: {}", latest_block_height);
+        assert_eq!(latest_block_height, 10);
+    }
```
MitchTurner, line 121: This is failing with inconsistent results:

Sometimes more of the validators find peers with the correct block:
```
validator
latest_block_height: 10
validator
latest_block_height: 10
validator
latest_block_height: 10
validator
latest_block_height: 10
validator
latest_block_height: 0
```
but usually it fails with a no discovery:
```
validator
latest_block_height: 10
validator
latest_block_height: 10
validator
latest_block_height: 10
validator
latest_block_height: 0
```
even after 60 seconds.

MitchTurner: Actually. I think this is a red herring. I bet this is just a lucky thing with the order they fill up? Maybe not. I need to investigate to make sure we expect this to work.

MitchTurner: It was a problem of saturated peers.

bin/fuel-core/src/cli/run/p2p.rs
```diff
@@ -321,8 +321,9 @@ impl P2PArgs {
             reserved_nodes: self.reserved_nodes,
             reserved_nodes_only_mode: self.reserved_nodes_only_mode,
             enable_mdns: self.enable_mdns,
-            max_peers_connected: self.max_peers_connected,
-            max_connections_per_peer: self.max_connections_per_peer,
+            max_discovery_peers_connected: self.max_peers_connected,
+            max_gossipsub_peers_connected: 100,
+            max_request_response_peers_connected: 100,
```
acerone85, line 326: Some suggestions on these two lines:

1. should these values be configurable, and also have a hard limit? 

e.g. something like 

```suggestion
            max_gossipsub_peers_connected: self.max_gossipsub_peers_connected.min(100),
            max_request_response_peers_connected: self.max_request_response_peers_connected.min(100),
```

2. Maybe best to declare the hardcoded constants as const variables? 

3. You renamed `self.max_peers_connected, but not `self.max_peers_connected`. (Ther argument has `--max-peers-connected` as argument name hardcoded in clap_derive, so there should be no risk in renaming this

crates/services/p2p/src/sized_hashset.rs
```diff
@@ -0,0 +1,36 @@
+use std::{
+    collections::HashSet,
+    hash::Hash,
+};
+
+pub struct SizedHashset<T> {
```
acerone85, line 6: nit: I find `BoundedHashSet` more clear

crates/services/p2p/src/p2p_service.rs
```diff
@@ -596,6 +616,13 @@ impl FuelP2PService {
                 message,
                 message_id,
             } => {
+                if !self
+                    .gossipsub_peer_limiter
+                    .insert_new_if_room(propagation_source)
+                {
+                    return None;
+                }
+
```
acerone85, line 625: I wonder if this makes the node susceptible to a DoS. It requires only 100 peers to effectively block all `GossipSub` messages on a target node.

crates/services/p2p/src/p2p_service.rs
```diff
@@ -596,6 +616,13 @@ impl FuelP2PService {
                 message,
                 message_id,
             } => {
+                if !self
+                    .gossipsub_peer_limiter
+                    .insert_new_if_room(propagation_source)
+                {
```
acerone85, line 622: should there be a warn message that the node has reached the maximum number of peers?

acerone85:
only one comment re the possibility of DoS, otherwise LGTM

crates/services/p2p/src/p2p_service.rs
MitchTurner: Yeah. There are definitely some vulnerabilities still. WE have an issue for tracking it:
https://github.com/FuelLabs/fuel-core/issues/1968

bin/fuel-core/src/cli/run/p2p.rs
MitchTurner: >3. You renamed self.max_peers_connected, but not self.max_peers_connected. (Ther argument has --max-peers-connected` as argument name hardcoded in clap_derive, so there should be no risk in renaming this

I'm not following this point. It looks like you mentioned the same thing twice `self.max_peers_connected`

MitchTurner: I did your first two recs, but my comment I added might be too vague. I don't completely understand why these two were hardcoded but the `discovery` limit wasn't.

crates/services/p2p/src/p2p_service.rs
MitchTurner: I don't think so. It could get very noisy and rejecting requests will happen in a healthy system. I agree that some visibility would be nice maybe in metrics?

bin/fuel-core/src/cli/run/p2p.rs
acerone85: > I'm not following this point. It looks like you mentioned the same thing twice self.max_peers_connected 

sorry, I meant: you renamed `Config::max_peers_connected` into `Config::max_discovery_peers_connected`, while leaving the same name for `P2PArgs::max_peers_connected`.

netrome (dismissed):

netrome (approved):

crates/services/p2p/src/bounded_hashset.rs
```diff
@@ -0,0 +1,36 @@
...
+    // Will idempotently insert if there is room in the set.
+    // Redundant keys don't take up more room
+    pub fn insert_new_if_room(&mut self, k: T) -> bool {
+        if self.inner.len() >= self.capacity && !self.inner.contains(&k) {
+            false
+        } else {
+            self.inner.insert(k);
+            true
+        }
+    }
```
Dentosal, line 31:
```suggestion
    pub fn insert_new_if_room(&mut self, k: T) -> bool {
        if self.inner.len() >= self.capacity {
            !self.inner.contains(&k)
        } else {
            self.inner.insert(k)
        }
    }
```

Dentosal (approved):

crates/services/p2p/src/bounded_hashset.rs
MitchTurner: This doesn't quite work. According to the docs for `HashSet::insert`:
```rs
    /// Returns whether the value was newly inserted. That is:
    ///
    /// - If the set did not previously contain this value, `true` is returned.
    /// - If the set already contained this value, `false` is returned,
    ///   and the set is not modified: original value is not replaced,
    ///   and the value passed as argument is dropped.
    ///
```
That last bullet is the opposite behavior; we want it to return true if it's included in the `HashSet` at all. Perhaps:
```rs
    pub fn insert_new_if_room(&mut self, k: T) -> bool {
        if self.inner.len() >= self.capacity {
            self.inner.contains(&k)
        } else {
            self.inner.insert(k);
            true
        }
    }
```
This doesn't fix the issue with `insert`ing again if it's already contained though. I realized that this wasn't optimized before, but I didn't want to accidentally change the behavior.

PR #2651, opened 2025-01-29
Refactor `iter_store_keys` 

MitchTurner:
## Description

Was looking into:
```rs
                // We have to clone the prefix and start, because we need to pass them to the iterator
                // if someone finds a solution without making it a vec, feel free to contribute :)
                let column = column.id();
                let prefix = prefix.map(|prefix| prefix.to_vec());
                let start = start.map(|start| start.to_vec());
```
and did some refactoring to get to the bottom of the problem.

I don't know if there is a solution unless we decide to change the signature of `iter_store_keys` to something like:

```rs
    fn iter_store_keys<'a>(
        &'a self,
        column: Self::Column,
        prefix: Option<&'a [u8]>,
        start: Option<&'a [u8]>,
        direction: IterDirection,
    ) -> BoxedIter<KeyItem>;
```
But that's not realistic in this PR.

Might as well keep the refactoring because I feel like it's cleaner.

### DIFF
--- a/crates/storage/src/iter/changes_iterator.rs
+++ b/crates/storage/src/iter/changes_iterator.rs
@@ -14,8 +14,13 @@ use crate::{
         Value,
         WriteOperation,
     },
-    transactional::StorageChanges,
+    transactional::{
+        Changes,
+        ReferenceBytesKey,
+        StorageChanges,
+    },
 };
+use alloc::collections::BTreeMap;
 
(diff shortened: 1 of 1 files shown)


AurelienFT (approved):
Thank you for that, yeah I'm sad we can't do better without breaking the trait but it's logic :/

PR #2659, opened 2025-02-03
Change `derivative` to `educe`

AurelienFT:
## Linked Issues/PRs
https://github.com/FuelLabs/fuel-core/issues/2426

## Description
Derivative is unmaintained and `educe` is a close drop-in. A similar PR has been already merged in the VM : https://github.com/FuelLabs/fuel-vm/pull/876

## Checklist

### Before requesting review


rymnc (dismissed):
would be interesting to see change in wasm bytecode size after this

xgreenx (dismissed):

Cargo.lock
```diff
@@ -1472,9 +1472,9 @@ checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
 name = "cc"
-version = "1.2.12"
```
xgreenx, line 1475: Do we need to do `cargo update`? Maybe we can explode it from this pr?

xgreenx (approved):

Cargo.lock
AurelienFT: The goal was to update derive more to the stable "1" version instead of "0.xx" but will remove from this PR then

crates/types/src/blockchain/header/v1.rs
```diff
@@ -29,7 +31,7 @@ pub struct BlockHeaderV1 {
     /// The header metadata calculated during creation.
     /// The field is pub(crate) to enforce the use of the [`PartialBlockHeader::generate`] method.
     #[cfg_attr(feature = "serde", serde(skip))]
-    #[derivative(PartialEq = "ignore")]
+    #[educe(PartialEq(ignore))]
```
rafal-ch, line 34: Not related to this PR, just a side question because I was wondering if it could be better to just write `PartialEq` manually instead of generating it with a macro from a separate crate? Basically this boils down to the rule: use classic `derive`, unless you can't, then write the implementation manually.

I'd be interested to hear opinions.

AurelienFT: Mhh interesting question.
Personally I perfer to use the crate because the ignore is directly on the structure and so you can read everything here compare to when you have to write the implementation yourself it's separated code.

Also with the crate we ensure that all fields that are not explicitly ignored are used which is up to the developers and reviewers to ensure that when writing by hand which lead to more errors IMO.

xgreenx (approved):

xgreenx (approved):

netrome (approved):

PR #2646, opened 2025-01-29
port(storage): use the cached block height from the database at the time of creation of the view

rymnc:
## Linked Issues/PRs

- same as https://github.com/FuelLabs/fuel-core/pull/2639, but for master

## Description

The most important changes include adding metadata support to the `GenericDatabase`, modifying type definitions to include height metadata, and updating various methods to use the new database structure.

### Database Structure Improvements:

[...]

PR #2666, opened 2025-02-04
Rpc consistency proposal: Wait for up to a number of blocks specified in configuration before failing

acerone85:
## I am using a rebase heavy workflow to handle stacked PR. If you want to push updates to this PR and make sure that the whole PR is consistent with the target branch, please do the following instead of `git merge 1897-rpc-consistency-proposal`: 

```
git checkout ac@rpc-consistency-proposal-wait-for-block
git fetch origin
git rebase --onto origin/1897-rpc-consistency-proposal 1897-rpc-consistency-proposal
# Fix conflicts if necessary
git branch -f 1897-rpc-consistency-proposal origin/1897-rpc-consistency-proposal
git push -f origin
```

## Linked Issues/PRs

Follow up on #1897 

This PR implements a mechansim that allows a node operator to configure a tolerance level, in number of blocks, for the required fuel_block_height extension. When checking that the `current_fuel_block_height` is at least the `required_fuel_block_height`, the node will check first whether the `current_fuel_block_height` is at most `required_fuel_block_height_tolerance` "too far away" from the `required_fuel_block_height`. If that is the case, the node will check every second if the precondition has been met before fulfilling the request. Otherwise, the request will fail because the precondition has not been met. 

Because the HTTP server of the node has a timeout layer, requests by a node that has lost synchronisation with the network will eventually fail. 

## Checklist

### Before requesting review

- [ ] Someone else?


crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
```diff
@@ -98,33 +109,51 @@ impl Extension for RequiredFuelBlockHeightInner {
...
                     );
+                }
 
-                    return response
+                // If the node is lagging behind w.r.t. the required fuel block height,
+                // but within an acceptable interval, wait until the current block height
+                // matches the required fuel block height.
+                while current_block_height < *required_block_height {
+                    tokio::time::sleep(SLEEP_DURATION).await;
+                    match view.latest_block_height() {
```
xgreenx, line 132: It would be nice if we could subscribe for the height that we need and notify all dependent tasks.

crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
```diff
@@ -98,33 +109,51 @@ impl Extension for RequiredFuelBlockHeightInner {
...
+                        )]);
+                            return response
+                        }
+                        Ok(block_height) => {
+                            current_block_height = block_height;
+                        }
+                    }
                 }
             }
         }
```
xgreenx, line 159: After you wait for the required height, the view that you have inside of the context is old. We need to update it.

I think it would be better if we combined `ViewExtension` and `RequiredFuelBlockHeightInner` together, where you create a view only if you reach the required height.

crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
xgreenx: BTW, without https://github.com/FuelLabs/fuel-core/pull/2646 it is very expensive operation, and do it for each query can be a lot. Before releasing RPC consistancy, we need merge this optimization

crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
acerone85: I tried a different approach where we subscribe to changes to the block height in the extension. It does not require merging the `ViewExtension` and `RequiredFuelBlockHeightInner`, curious to hear your opinion

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed 
 - [2473](https://github.com/FuelLabs/fuel-core/pull/2473): Graphql requests and responses make use of a new `extensions` object to specify request/response metadata. A request `extensions` object can contain an integer-valued `required_fuel_block_height` field. When specified, the request will return an error unless the node's current fuel block height is at least the value specified in the `required_fuel_block_height` field. All graphql responses now contain an integer-valued `current_fuel_block_height` field in the `extensions` object, which contains the block height of the last block processed by the node.
 
+### Added
+- [2666](https://github.com/FuelLabs/fuel-core/pull/2666) Added `--required-fuel-block-height-tolerance` CLI arg (default: `10`). Requests with `required_fuel_block_height` set will wait if within tolerance of `current_fuel_block_height` instead of failing.
+
 ## [Version 0.41.5]
```
xgreenx, line 15: It seems you need to merge with master, because the latest version is `0.41.6`=)

bin/fuel-core/src/cli/run/graphql.rs
```diff
@@ -66,6 +66,13 @@ pub struct GraphQLArgs {
     #[clap(long = "api-request-timeout", default_value = "30s", env)]
     pub api_request_timeout: humantime::Duration,
 
+    #[clap(
+        long = "required-fuel-block-height-tolerance",
```
xgreenx, line 70:
```suggestion
        long = "graphql_required-fuel-block-height-tolerance",
```

xgreenx: Also it would be nice to have a comment to describe the functionality behind this CLI argument.

crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
xgreenx: You need to update the `ReadView` that was created by the `ViewExtension`. If required height is 9 blocks higher than the current height, `ReadView` will be 9 blocks old.

You need to actualize the view to be at `required_block_height`. It is where you need to create the view second time, and it is where merging `ViewExtension` and `RequiredFuelBlockHeightInner` can help to avoid creation of 2 views.

crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
```diff
@@ -99,50 +130,76 @@ impl Extension for RequiredFuelBlockHeightInner {
...
                 }
+
+                // If the node is lagging behind w.r.t. the required fuel block height,
+                // but within an acceptable interval, wait until the current block height
+                // matches the required fuel block height.
+                // TODO: Add a timeout to prevent the node from waiting indefinitely.
+                if current_block_height < *required_block_height {
+                    self.block_height_subscription_handle
+                        .clone()
+                        .wait_for_block_height(*required_block_height)
```
xgreenx, line 154: You can make `wait_for_block_height` to work with `&self` to avoid cloning here

crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
```diff
@@ -99,50 +130,76 @@ impl Extension for RequiredFuelBlockHeightInner {
...
+                        column,
                     );
-
-                    return response
                 }
+
+                // If the node is lagging behind w.r.t. the required fuel block height,
+                // but within an acceptable interval, wait until the current block height
+                // matches the required fuel block height.
+                // TODO: Add a timeout to prevent the node from waiting indefinitely.
```
xgreenx, line 150: I think you can add a timeout in 1 minute + number of blocks(in seconds) that you need to wait

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -539,6 +544,31 @@ where
...
+    rx_block_height: tokio::sync::watch::Receiver<BlockHeight>,
+}
+
+impl BlockHeightSubscriptionHandle {
+    pub async fn wait_for_block_height(
+        &mut self,
+        height: BlockHeight,
+    ) -> anyhow::Result<()> {
+        loop {
+            let current_height = *self.rx_block_height.borrow_and_update();
```
xgreenx, line 558: You can do just `borrow`, it still should work

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -539,6 +544,31 @@ where
...
+pub struct BlockHeightSubscriptionHandle {
+    rx_block_height: tokio::sync::watch::Receiver<BlockHeight>,
+}
+
+impl BlockHeightSubscriptionHandle {
+    pub async fn wait_for_block_height(
+        &mut self,
+        height: BlockHeight,
+    ) -> anyhow::Result<()> {
+        loop {
```
xgreenx, line 557: While it is very simple to implement it with the `tokio::watch`, it is still expensive from an execution perspective.

Let's imagine the current height is 10. The required height for 10k requests is 20.

10 times, each task related to the request will wake up; it is 100k wakeups that check the state(it is read of the RWLock) of the `watch` and wait again.

If you have the list of subscribers for each block height, we can avoid unnecessary wakeups. And it should be very easy to maintain this list via one mutex. You can have `BTreeMap<BlockHeight, Vec<tokio::sync::oneshot::Sender<()>>>`, where you register all listeners, and on each block you just process most left listeners.

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -539,6 +544,31 @@ where
...
+#[derive(Clone, Debug)]
+pub struct BlockHeightSubscriptionHandle {
+    rx_block_height: tokio::sync::watch::Receiver<BlockHeight>,
+}
+
+impl BlockHeightSubscriptionHandle {
+    pub async fn wait_for_block_height(
+        &mut self,
+        height: BlockHeight,
+    ) -> anyhow::Result<()> {
```
xgreenx, line 556: You can return the new block height from `wait_for_block_height`. In this case you don't need `latest_seen_block_height`

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -745,5 +779,7 @@ where
         da_compression_config,
         continue_on_error,
         base_asset_id: *consensus_parameters.base_asset_id(),
+        // TODO: Should we recover the block height from the DB?
```
xgreenx, line 782: Yes, we need to get the block height from the `off_chain_database`

crates/fuel-core/src/graphql_api/required_fuel_block_height_extension.rs
acerone85: yep, thanks for explaining. 
I had a look, and merging the two views will totally help. Will amend

CHANGELOG.md
acerone85: rebased

crates/fuel-core/src/graphql_api/worker_service.rs
acerone85: happy to do that, but that will introduce busy waiting as we cannot await on the value being changed on the sender side. Is that okay?

bin/fuel-core/src/cli/run/graphql.rs
```diff
@@ -66,6 +66,26 @@ pub struct GraphQLArgs {
...
+
+    /// The minimum time that the node will wait to catch up to the required block height
+    /// of a graphql request. The actual time may be longer and proportional to how
+    /// much the node is behind the required block height.
+    #[clap(
+        long = "graphql-required-block-height-min-timeout-seconds",
+        default_value = "10",
+        env
+    )]
+    pub required_fuel_block_height_min_timeout: u32,
```
xgreenx, line 87: I think this CLI argument is overkill, it is very hard to understand what it is doing, if you don't look into the code. And even then, it is very hard to trigger this timeout

xgreenx, line 87:
```suggestion
    /// The minimum time that the node will wait to catch up to the required block height
    /// of a graphql request. The actual time may be longer and proportional to how
    /// much the node is behind the required block height.
    #[clap(
        long = "graphql-required-block-height-min-timeout-seconds",
        default_value = "30s",
        env
    )]
    pub required_fuel_block_height_min_timeout: humantime::Duration,
```

crates/fuel-core/src/graphql_api.rs
```diff
@@ -46,6 +46,13 @@ pub struct ServiceConfig {
     pub max_queries_directives: usize,
     pub max_concurrent_queries: usize,
     pub request_body_bytes_limit: usize,
+    /// Number of blocks that the node can be lagging behind the required fuel block height
+    /// before it will be considered out of sync.
+    pub required_fuel_block_height_tolerance: u32,
+    /// Minimum time to wait before dropping the request if the node is lagging behind the required
+    /// fuel block height. The actual timeout in seconds is `required_fuel_block_height_min_timeout + lag`,
+    /// where `lag` is the number of blocks the node is lagging behind the required fuel block height.
+    pub required_fuel_block_height_min_timeout: Duration,
```
xgreenx, line 55: I proposed to add `lag` to avoid making the `required_fuel_block_height_min_timeout` configurable via the CLI=D If you plan to have CLI argument, then we need to remove lag for simplicity of the logic. If you remove the CLI argument, then we need to have a lag + some random timeout like 1 minute for example.

xgreenx:
Sorry, clicked wrong button=D Didn't plan to close it=)

xgreenx (approved):
Thank you! The change looks very nice

crates/fuel-core/src/graphql_api.rs
acerone85: Happy to have lag + random timeout, wondering if 1 minute is too high (the default timeout for all RPC requests is 30 seconds IIRC)

CHANGELOG.md
```diff
@@ -12,12 +12,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - [2150](https://github.com/FuelLabs/fuel-core/pull/2150): Upgraded `libp2p` to `0.54.1` and introduced `ConnectionLimiter` to limit pending incoming/outgoing connections.
+- [2666](https://github.com/FuelLabs/fuel-core/pull/2666) Added `--graphql-block-height-tolerance` (default: `10`) and `--graphql-block-height-min-timeout-secs` (default: `30`) CLI args. Requests with `required_fuel_block_height` set will wait if within tolerance of `current_fuel_block_height` instead of failing. The number of seconds that the node will wait is proportional to the number of blocks `lag-behind-blocks` that the node is lagging behind the required fuel block height, and equal to `graphql-block-height-min-timeout-secs + lag-behind-blocks.
```
rafal-ch, line 15: Maybe we can be more concise, because we don't need to treat changelog as a feature documentation :)

```suggestion
Added two new CLI arguments to control the GraphQL queries consistency: `--graphql-block-height-tolerance` (default: `10`) and `--graphql-block-height-min-timeout-secs` (default: `30`). If a request requires a specific block height and the node is slightly behind, it will wait instead of failing. The wait time increases with the block lag, starting at the minimum timeout.
```

crates/fuel-core/src/graphql_api.rs
xgreenx: I think we can stick with CLI argument without lag, it allows us to test timeout logic(it is what you did and I think it makes sense):)

crates/fuel-core/src/graphql_api/block_height_subscription.rs
```diff
@@ -0,0 +1,99 @@
...
+    collections::BTreeMap,
+    sync::Arc,
+};
+
+use fuel_core_types::fuel_types::BlockHeight;
+use parking_lot::RwLock;
+use tokio::sync::oneshot;
+
+#[derive(Default)]
+
```
rafal-ch, line 12:
```suggestion
```

rafal-ch (commented):
Partial review, will continue later.
No issues so far.

crates/fuel-core/src/graphql_api.rs
acerone85: ok, just to confirm we remove the lag altogether and we always wait for a timeout?

xgreenx: Yes

xgreenx (approved):

crates/fuel-core/src/graphql_api/block_height_subscription.rs
```diff
@@ -0,0 +1,98 @@
...
+    pub fn latest_seen_block_height(&self) -> BlockHeight {
+        self.inner.read().latest_seen_block_height
+    }
+}
+
+#[derive(Debug, Default)]
+struct HandlersMapInner {
+    tx_handles: BTreeMap<Reverse<BlockHeight>, Vec<oneshot::Sender<()>>>,
+    latest_seen_block_height: BlockHeight,
+}
```
netrome, line 89: This feels over-complicated to me. I'd consider switching this to a single `tokio::watch` channel over the block height, and have the subscriber await changes to the block height and use the `wait_for` method. https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.wait_for

Any thoughts?

netrome: This is how I imagine we could simplify the implementation https://github.com/FuelLabs/fuel-core/pull/2721

netrome (approved):
Looks generally good to me. I think the block height subscription can be simplified.

crates/fuel-core/src/graphql_api/block_height_subscription.rs
netrome: Oh nvm I see the previous discussion regarding this.

Issue #2672, opened 2025-02-05
CCP and LDC errors if given a buffer size less than the contract size.


netrome:
Discovered by @rymnc, observable in the following commit in the zkvm-primitives repo https://github.com/FuelLabs/zkvm-primitives/commit/ca4ef6e20abdb17a4945bbd21d20c85761777bf2

### How to reproduce
Check out the above commit and run

```
cargo nextest run fuel-zkvm-primitives-test-fixtures opcodes::tests::test_contract_instruction_ccp
```

### Bug description
When calling `CCP` the caller specifies a contract ID, and a number of bytes to read. If this number isn't the same as the number of bytes in the contract we get the following error:
```
VmExecution { error: "Storage error: Buffer size is not equal to the value size", transaction_id: 9fb5fb513aeebfd5ddadcd84e9ff9aae699bdbd52df955d7c6cd79dd0fc946e7 }
```

The expected behavior is to successfully copy the provided number of bytes.

xgreenx:
Do we have more context on the issue?=)

netrome:
We need to sync with @Voxelot and/or @xgreenx to align on how we want this to behave.

xgreenx:
We forgot to add support for `offset` inside of this function

<img width="766" alt="Image" src="https://github.com/user-attachments/assets/ba5912ce-3874-4c1a-a0a1-c274f9d2ab46" />

netrome:
This affects all 0.41.x releases up to 0.41.5. This change has been uploaded in STF version 19 and 20 on devnet. No STF on testnet or mainnet has been affected.

PR #2438, opened 2024-11-14
Refactor fuel-core to use version of StorageRead::read with offset (Full update to 0.59.1)

acerone85:
## Linked Issues/PRs

Related to https://github.com/FuelLabs/fuel-vm/issues/681

## Description
In order to load contracts from a given offset efficiently, we have changed the interface of `StorageRead::read` to accept a n offset in input (see https://github.com/FuelLabs/fuel-vm/pull/863). This PR refactors the fuel-core repo to make use of the new StorageRead trait.

## Checklist
- [ ] Breaking changes are clearly marked as such in the PR description and changelog
- [ ] New behavior is reflected in tests
- [ ] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [ ] I have reviewed the code myself
- [ ] I have created follow-up issues caused by this PR and linked them here

### After merging, notify other teams

[Add or remove entries as needed]

- [ ] [Rust SDK](https://github.com/FuelLabs/fuels-rs/)
- [ ] [Sway compiler](https://github.com/FuelLabs/sway/)
- [ ] [Platform documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+) (for out-of-organization contributors, the person merging the PR will do this)
- [ ] Someone else?

PR #2673, opened 2025-02-05
Change `read` behavior on the `InMemoryTransaction` to use offset and allow not equal buf size

AurelienFT:
## Linked Issues/PRs
closes https://github.com/FuelLabs/fuel-core/issues/2672

## Description
Change code of `read` of `InMemoryTransaction` to match the other `read` functions

Tests had been added and the error in `read` function has been improved to be more clear.

## Checklist

### Before requesting review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2667](https://github.com/FuelLabs/fuel-core/pull/2667): Populate `Blobs` table in global merkle root storage.
 - [2652](https://github.com/FuelLabs/fuel-core/pull/2652): Global Merkle Root storage crate: Add Raw contract bytecode to global merkle root storage when processing Create transactions.
 
+### Fixed
+- [2673](https://github.com/FuelLabs/fuel-core/pull/2673): Change read behavior on the InMemoryTransaction to use offset and allow not equal buf size (fix CCP and LDC broken from https://github.com/FuelLabs/fuel-vm/pull/847)
+
 ## [Version 0.41.5]
 
 ### Changed
--- a/crates/storage/src/transactional.rs
+++ b/crates/storage/src/transactional.rs
@@ -670,8 +670,8 @@ mod test {
             let storage = InMemoryStorage::<Column>::default();
             let mut view = storage.read_transaction();
             let key = vec![0xA, 0xB, 0xC];
-            let expected = Value::from([1, 2, 3]);
-            view.put(&key, Column::Metadata, expected.clone()).unwrap();
+            let value = Value::from([1, 2, 3]);
+            view.put(&key, Column::Metadata, value).unwrap();
             // test
             let mut buf = [0; 3];
             let ret = view.read(&key, Column::Metadata, 0, &mut buf).unwrap();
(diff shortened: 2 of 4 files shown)


rymnc (approved):
can confirm this fixes my bug

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [2667](https://github.com/FuelLabs/fuel-core/pull/2667): Populate `Blobs` table in global merkle root storage.
 - [2652](https://github.com/FuelLabs/fuel-core/pull/2652): Global Merkle Root storage crate: Add Raw contract bytecode to global merkle root storage when processing Create transactions.
 
+### Fixed
+- [2673](https://github.com/FuelLabs/fuel-core/pull/2673): Change read behavior on the InMemoryTransaction to use offset and allow not equal buf size (fix CCP and LDC broken from https://github.com/FuelLabs/fuel-vm/pull/847)
```
acerone85, line 17: technically the bug was introduced in https://github.com/FuelLabs/fuel-core/pull/2438, the change in fuel-vm seems to be working as expected from what I can tell

AurelienFT: For me it's https://github.com/FuelLabs/fuel-vm/pull/847 that broke CCP and LDC opcode if this PR hasn't land on master, the `read` function would have been wrong since #2438 but the opcode would have been fine.

acerone85: yep, but the error is triggered in the `if` branch of an `if else` statement that does not invoke any of the code in fuel-vm.

AurelienFT: yes I talked about the error just before in the sentence I just added this parenthesis with the opcode in case we want to find later to have beteer search results

crates/storage/src/transactional.rs
```diff
@@ -656,6 +664,68 @@ mod test {
         use super::*;
         use crate::column::Column;
 
+        #[test]
+        fn read_returns_from_view_exact_size() {
+            // setup
+            let storage = InMemoryStorage::<Column>::default();
+            let mut view = storage.read_transaction();
+            let key = vec![0xA, 0xB, 0xC];
+            let expected = Value::from([1, 2, 3]);
```
MitchTurner, line 673: nit: we probably shouldn't use `expected` here (**or on the other tests**). Copy pasta.
Perhaps `source_data` makes more sense.

crates/storage/src/transactional.rs
```diff
@@ -656,6 +664,68 @@ mod test {
...
+            view.put(&key, Column::Metadata, expected.clone()).unwrap();
+            // test
+            let mut buf = [0; 2];
+            let ret = view.read(&key, Column::Metadata, 0, &mut buf).unwrap();
+            // verify
+            assert_eq!(ret, Some(2));
+            assert_eq!(buf, [1, 2]);
+        }
```
MitchTurner, line 697: Maybe add another test with offset and passing. Could just be the same as this but 
`assert_eq!(buf, [2, 3]);`

crates/storage/src/transactional.rs
rymnc: should we address in a follow up?

crates/storage/src/transactional.rs
AurelienFT: done

crates/storage/src/transactional.rs
AurelienFT: done

rymnc (approved):
g1

MitchTurner (approved):
LGTM. I'd recommend getting @xgreenx 's eyes on it just in case.

PR #2678, opened 2025-02-07
fix(headers): move tx_id_commitment to application_headers

rymnc:
## Linked Issues/PRs

- https://github.com/FuelLabs/fuel-core/pull/2648

## Description

We get rid of `enum_dispatch` because if you try to derive the impl on a trait with a generic, you also have to apply the generic to the enum ~ which we don't need. rather than having 2 traits, one generic and one not, opted to cut `enum_dispatch` altogether.

idk why we were directly accessing the block header fields in the whole codebase without using accessor methods, even when we had defined an enum for block header. this PR fixes that.

if `fault-proving` is enabled, we now generate `BlockV1` with `BlockHeaderV2`. `tx_id_commitment` has been moved to `GeneratedApplicationFieldsV2`, renamed `GeneratedApplicationFields` to `GeneratedApplicationFieldsV1`. maybe we can remove `pub` from a lot of fields and convert to `pub(crate)` but i can do that in a follow up PR.

I still favour having a separate `FaultProvingHeader` field on the header, since we anyway have to define a new hash function for the `ApplicationHeader<GeneratedApplicationFieldsV2>`, it was less verbose before.

### Reviewers guide

- all of the changes in crates other than `types` is just using getters/setters where necessary
- in `types/header`, refactored, removed `enum_dispatch` due to generic usage that didn't apply on parent enum
- moved `tx_id_commitment` from `fault_proving_header` to `application_header`. New struct `GeneratedApplicationFieldsV2` to accommodate it.

### AI generated desc

This pull request includes several changes to improve the consistency and readability of the code by replacing direct field accesses with method calls for accessing block header fields. The most important changes include updates to the `BlockHeader` field access, test adjustments, and feature additions in the `Cargo.toml` file.

### Block Header Field Access Updates:
* Updated `impl LastBlockConfig` in `crates/chain-config/src/config/state.rs` to use method calls for accessing block header fields instead of direct field accesses.
* Modified `impl From<&BlockHeader> for CompressedBlockHeader` in `crates/compression/src/compressed_block_payload/v1.rs` to use method calls for accessing block header fields.
* Updated `impl Header` in `crates/fuel-core/src/schema/block.rs` to use method calls for accessing block header fields.

### Test Adjustments:
* Adjusted various test cases in `crates/fuel-core/src/executor.rs` to use method calls for accessing block header fields. [[1]](diffhunk://#diff-379b2602e8ed0e8048291c20eff0d6ab59d4cfc1e6c5c9c6d5513b33008ba1c1L366-R367) [[2]](diffhunk://#diff-379b2602e8ed0e8048291c20eff0d6ab59d4cfc1e6c5c9c6d5513b33008ba1c1L2715-R2715) [[3]](diffhunk://#diff-379b2602e8ed0e8048291c20eff0d6ab59d4cfc1e6c5c9c6d5513b33008ba1c1L2754-R2754) [[4]](diffhunk://#diff-379b2602e8ed0e8048291c20eff0d6ab59d4cfc1e6c5c9c6d5513b33008ba1c1L3389-R3389)
* Updated test cases in `crates/fuel-core/src/service/adapters/consensus_parameters_provider.rs` to use method calls for setting consensus parameters version.

### Feature Additions:
* Added new feature `fault-proving` in `Cargo.toml` for `crates/services/consensus_module`.
* Updated imports and function implementations in `crates/services/consensus_module/poa/src/verifier.rs` to support the new feature and improve code structure. [[1]](diffhunk://#diff-bffa43e1ce0d3b4c653637fd70da7d6321dd2528d4bbf496dce57ed2488c552fL8-R11) [[2]](diffhunk://#diff-bffa43e1ce0d3b4c653637fd70da7d6321dd2528d4bbf496dce57ed2488c552fL64-R67) [[3]](diffhunk://#diff-bffa43e1ce0d3b4c653637fd70da7d6321dd2528d4bbf496dce57ed2488c552fR76-R83)

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/types/src/blockchain/header.rs
+++ b/crates/types/src/blockchain/header.rs
@@ -24,24 +24,18 @@ use crate::{
         MessageId,
     },
 };
-use enum_dispatch::enum_dispatch;
 use tai64::Tai64;
 pub use v1::BlockHeaderV1;
+use v1::GeneratedApplicationFieldsV1;
 #[cfg(feature = "fault-proving")]
-pub use v2::{
-    BlockHeaderV2,
-    FaultProvingHeader,
-};
+pub use v2::BlockHeaderV2;
+#[cfg(feature = "fault-proving")]
+use v2::GeneratedApplicationFieldsV2;
 
 /// Version-able block header type
 #[derive(Clone, Debug, derivative::Derivative)]
 #[derivative(PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[enum_dispatch(GetBlockHeaderFields)]
-#[cfg_attr(
-    any(test, feature = "test-helpers"),
-    enum_dispatch(BlockHeaderDataTestHelpers)
-)]
 pub enum BlockHeader {
     /// V1 BlockHeader
     V1(BlockHeaderV1),
@@ -51,12 +45,11 @@ pub enum BlockHeader {
 }
 
 /// Helpful methods for all variants of the block header.
-#[enum_dispatch]
-pub(crate) trait GetBlockHeaderFields {
+pub trait GetBlockHeaderFields<GAF> {
     /// Get the consensus portion of the header.
     fn consensus(&self) -> &ConsensusHeader<GeneratedConsensusFields>;
     /// Get the application portion of the header.
-    fn application(&self) -> &ApplicationHeader<GeneratedApplicationFields>;
+    fn application(&self) -> &ApplicationHeader<GAF>;
     /// Get the metadata of the header.
     fn metadata(&self) -> &Option<BlockHeaderMetadata>;
     /// Re-generate the header metadata.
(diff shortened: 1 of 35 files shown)


crates/types/src/blockchain/header/v2.rs
```diff
@@ -151,4 +155,82 @@ impl crate::blockchain::header::BlockHeaderDataTestHelpers for BlockHeaderV2 {
...
+    pub transactions_count: u16,
+    /// Number of message receipts in this block.
+    pub message_receipt_count: u32,
+    /// Merkle root of transactions.
+    pub transactions_root: Bytes32,
+    /// Merkle root of message receipts in this block.
+    pub message_outbox_root: Bytes32,
+    /// Root hash of all imported events from L1
+    pub event_inbox_root: Bytes32,
+    /// tx id commitment
```
netrome, line 189: Nit
```suggestion
    /// TxID commitment
```

crates/types/src/blockchain/header/v1.rs
```diff
@@ -21,22 +24,22 @@ use crate::{
 #[cfg_attr(any(test, feature = "test-helpers"), derive(Default))]
 pub struct BlockHeaderV1 {
     /// The application header.
-    pub application: ApplicationHeader<GeneratedApplicationFields>,
+    pub(crate) application: ApplicationHeader<GeneratedApplicationFieldsV1>,
     /// The consensus header.
-    pub consensus: ConsensusHeader<GeneratedConsensusFields>,
+    pub(crate) consensus: ConsensusHeader<GeneratedConsensusFields>,
```
netrome, line 29: Why do we change the visibility for the application and consensus headers?
Also, modifying these fields makes this change breaking.

netrome (changes requested):
One nit, and one question. Only blocker from my side is that we should update the changelog:

1. The previous entry from #2648 should be marked as breaking iirc.
2. We should either modify or write a new entry for this PR if we change visibility of public fields in the block header.

Other than that, this looks good to me.

crates/types/src/blockchain/header/v1.rs
rymnc: they should have never been public in the first place :)
it removes the point of having an enum for versioned headers

this is a breaking change, it is intended

crates/types/src/blockchain/header/v2.rs
rymnc: addressed in 44326fdf4f

netrome (dismissed):
Thanks for the update. Looks good now!

crates/types/src/blockchain/header.rs
```diff
@@ -51,12 +45,11 @@ pub enum BlockHeader {
 }
 
 /// Helpful methods for all variants of the block header.
-#[enum_dispatch]
-pub(crate) trait GetBlockHeaderFields {
+pub trait GetBlockHeaderFields<GAF> {
```
xgreenx, line 48: I don't see what is the use case of having a trait for getters(and setters in the case of the tests). 

Because it is generic, it means that it will be implemented only for `HeaderV1` and `HeaderV2` only. It can't be implemented for the `BlockHeader` enum, so it can't be used with it. Then, why do we need it? Looks like just public getters from `impl` section should be enough.

rymnc: leftover code from enum_dispatch land, will fix

rymnc: addressed in [28a816a](https://github.com/FuelLabs/fuel-core/pull/2678/commits/28a816a64b6f5e4bae82f3b1941dd5ed6dad3932)

netrome (approved):

xgreenx (approved):

netrome (approved):

PR #2661, opened 2025-02-03
Dry run in the past

Dentosal:
Closes #2062

## Description

Support dry running in past blocks. It's behind the `--historical-execution` flag.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog
- [x] New behavior is reflected in tests
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [x] I have reviewed the code myself
- [x] I have created follow-up issues caused by this PR and linked them here

### After merging, notify other teams

- [ ] [Rust SDK](https://github.com/FuelLabs/fuels-rs/)
- [ ] [Platform documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+) (for out-of-organization contributors, the person merging the PR will do this)

PR #2682, opened 2025-02-09
GraphQL: historical contract storage and balances

Dentosal:
Requires https://github.com/FuelLabs/fuel-core/pull/2661

## Description

Adds GraphQL APIs for querying contract storage slots and balances at given block height.

### Open questions

- Should this be behind `--historical-exection` flag or similar? It's potentially quite expensive to get past state of the db.

## Checklist

### Before requesting review


tests/tests/storage.rs
```diff
@@ -0,0 +1,203 @@
+use fuel_core::{
```
xgreenx, line 1: @Dentosal I remember you mentioned that you want to add tests for the balances part of the API as well=)

Dentosal: True. And also I want to refactor storage tests. Taking this back to draft.

Dentosal: Added tests in df92693. Refactored existing ones in 49b3f7c.

tests/tests/balances.rs
```diff
@@ -587,3 +593,52 @@ mod pagination {
...
+        .contract_balance_values(
+            &contract_id,
+            Some(minted_height),
+            vec![contract_id.asset_id(&sub_asset_id)],
+        )
+        .await
+        .unwrap();
+
+    dbg!(&balances_at_deployed);
+    dbg!(&balances_at_minted);
```
xgreenx, line 634:
```suggestion
```

xgreenx (approved):

xgreenx (dismissed):

crates/fuel-core/src/graphql_api.rs
```diff
@@ -168,3 +169,15 @@ impl<T> IntoApiResult<T> for Result<T, StorageError> {
...
 }
+
+pub fn require_historical_execution(ctx: &Context<'_>) -> async_graphql::Result<()> {
+    let config = ctx.data_unchecked::<Config>();
+
+    if config.historical_execution {
+        Ok(())
+    } else {
+        Err(async_graphql::Error::new(
+            "`--historical-execution` is required for this operation",
```
AurelienFT, line 180: nit: From what I understand this error is returned to the user. I don't think an API user will be informed about the flags of the CLI and it also cause a potential tech-debt if we change flag name. I think we should provide a more generalized sentence to the user such as : `Historical execution isn't activated on this API` smth like that

AurelienFT (approved):

crates/fuel-core/src/graphql_api.rs
Dentosal: We do the same for `debug`, but I think the suggesting here makes sense. Are you ok if we leave that as a follow-up, and change refrences to debug as well?

AurelienFT: Yep

xgreenx (approved):

AurelienFT (approved):

PR #2706, opened 2025-02-12
Remove reviewers in update README action

AurelienFT:
## Description

Currently, all the action jobs are failing because the group "FuelLabs/client" can't be seen by the Action and so it can't assign it as reviewer. This PR comes back to what we do with cargo update, we just let our repository rules determine the reviewers.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog
- [x] New behavior is reflected in tests
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [x] I have reviewed the code myself
- [x] I have created follow-up issues caused by this PR and linked them here

### After merging, notify other teams

[Add or remove entries as needed]

- [ ] [Rust SDK](https://github.com/FuelLabs/fuels-rs/)
- [ ] [Sway compiler](https://github.com/FuelLabs/sway/)
- [ ] [Platform documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+) (for out-of-organization contributors, the person merging the PR will do this)
- [ ] Someone else?

PR #2708, opened 2025-02-12
Update README.md

MitchTurner:


AurelienFT:
@MitchTurner can you approve this instead : https://github.com/FuelLabs/fuel-core/pull/2706 It should make the bot trigger to make a pr like this and we would be able to test the flow

PR #2715, opened 2025-02-13
Add GraphQL extension to provide current STF and CP versions

### LINKED ISSUE #2596: Include the version of the STF and consensus parameters versions in all GQL responses via `extensions`
Requires https://github.com/FuelLabs/fuel-core/issues/1897

Reuse the same mechanism that we use for `CURRENT_BLOCK_HEIGHT`

rafal-ch:
## Linked Issues/PRs
Closes #2596 

## Description
This PR adds new GraphQL extensions to provide:
- current STF version
- current consensus parameters version

Example response:
```
{
  "data": {
    "nodeInfo": {
      "nodeVersion": "0.41.6"
    }
  },
  "extensions": {
    "current_consensus_parameters_version": 0,
    "current_fuel_block_height": 0,
    "current_stf_version": 21
  }
}
```

### Additional housekeeping changes:
- all graphql extensions have been moved into a subdirectory

## TODO
- [x] Add tests to check that correct versions are returned when STF or CP are bumped
- [x] Rework how we retrieve the current STF version

## Checklist

### Before requesting review


crates/fuel-core/src/graphql_api/extensions/current_stf_version.rs
```diff
@@ -0,0 +1,57 @@
...
+        if let Ok(view) = db.view() {
+            if let Ok(latest_block) = view.latest_block() {
+                let current_stf_version =
+                    latest_block.header().state_transition_bytecode_version();
+                response.extensions.insert(
+                    CURRENT_STF_VERSION.to_string(),
+                    Value::Number(current_stf_version.into()),
+                );
+            }
+        }
```
rafal-ch, line 53: We might prefer reading the current STF version directly from on chain DB table `StateTransitionBytecodeVersions` similarly to how it's done in `BlockProducerDatabase::latest_state_transition_bytecode_version()`, but I couldn't figure out how to access this in GraphQL.

rafal-ch:
> > We're currently attaching some substantial number of bytes to every response. Even if user requests just the node version, he will get a bunch of additional data, which, AFAIK, cannot be opted-out from.
> > ```
> > {
> >   "data": {
> >     "nodeInfo": {
> >       "nodeVersion": "0.41.6"
> >     }
> >   },
> >   "extensions": {
> >     "current_consensus_parameters_version": 0,
> >     "current_fuel_block_height": 0,
> >     "current_stf_version": 21
> >   }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Not sure if this should be a concern, but I thought it's worth pointing out.
> 
> we miss the whole point of graphql here, if we provide data no one asked for :(

Exactly, that's why I was thinking that we should just provide this data via some info endpoint, but IIUC, the idea here is to provide this data proactively instead of requiring users to poll regularly. 

Maybe there's a way to opt out from extensions which I'm just missing.

crates/fuel-core/src/graphql_api/api_service.rs
```diff
@@ -287,6 +291,8 @@ where
         // `RequiredFuelBlockHeightExtension` uses the view set by the ViewExtension.
         // Do not reorder this line before adding the `ViewExtension`.
         .extension(RequiredFuelBlockHeightExtension::new())
+        .extension(CurrentStfVersionExtension::new())
+        .extension(CurrentConsensusParametersVersionExtension::new())
```
xgreenx, line 295: I think spawning a separate extension for the field it is too overkill=D It decreases performance without any reason

crates/fuel-core/src/graphql_api/extensions/current_stf_version.rs
xgreenx: We need to have a place where we cache this kind of information. Maybe, we could extend consensus parameters provider to also provide this kind of information, because fetch it from the database it a lot of waste work.

tests/test-helpers/src/lib.rs
```diff
@@ -20,11 +25,59 @@ use rand::{
...
 
 pub mod builder;
 pub mod counter_contract;
 pub mod fuel_core_driver;
 
+pub fn predicate() -> Vec<u8> {
+    vec![op::ret(1)].into_iter().collect::<Vec<u8>>()
+}
+
+pub fn get_graphql_extension_field_value(result_json: &str, name: &str) -> u64 {
```
xgreenx, line 38: `fuel-core-client` should fetch this information in the same way how it is done for the required block height. We have `ExtensionsResponse` where you can add several new fields.

Implementation details: In the case of the CPV and STFV we don't need any policy, we just can fetch the latest one and provide a getter

crates/fuel-core/src/graphql_api/api_service.rs
xgreenx: I think we can have one extension fully dedicated to return information to the end user. Like STFV, CPV, current block height(we can remove setting of the current height from the `RequiredFuelBlockHeightExtension`).

In this case we will have one extension that we can use to add any metadata

xgreenx:
> we miss the whole point of graphql here, if we provide data no one asked for :(

The `extensions` field is part of the GraphQL specification, and we can put there any data what we want. The user controls only the `data` field.

crates/fuel-core/src/graphql_api/extensions/current_consensus_parameters_version.rs
```diff
@@ -0,0 +1,52 @@
...
+        ctx: &ExtensionContext<'_>,
+        operation_name: Option<&str>,
+        next: NextExecute<'_>,
+    ) -> Response {
+        let consensus_parameters_provider = ctx.data_unchecked::<ConsensusProvider>();
+
+        let current_consensus_parameters_version =
+            consensus_parameters_provider.latest_consensus_parameters_version();
+
+        let mut response = next.run(ctx, operation_name).await;
```
acerone85, line 45: nit: 
```suggestion
        let mut response = next.run(ctx, operation_name).await;
        
        let current_consensus_parameters_version =
            consensus_parameters_provider.latest_consensus_parameters_version();
```

this way by the time you fetch the consensus parameters version you can be sure that the graphql operation has been executed and the value is up-to-date.

crates/fuel-core/src/graphql_api/extensions/current_stf_version.rs
acerone85: If we do that, it might be useful to also add the block height to the information retrieved, as in this case all the extension metadata will come from the same place. 

Btw, If I am not mistaken all the information we want to fetch comes from the block header right now, so maybe a `HeaderProvider` trait with a function that returns the whole block header, and then `stf_version` and `consensus_parameters_version` can be provided methods?

crates/fuel-core/src/graphql_api/api_service.rs
rafal-ch: The following changes have been made:
- Both STF and CP extensions have been merged into "chain state info" extension
- The "chain state info" extension is now also providing data about the current block height

crates/fuel-core/src/graphql_api/extensions/current_stf_version.rs
rafal-ch: I think I'll prefer to extend (and rename) the "Consensus Parameters Provider". However, I find caching the entire block header not optimal and my preference would be to cache exactly what we need. So, in addition to the current data, it'll store:
- current STF version
- current block height

wdyt?

tests/test-helpers/src/lib.rs
rafal-ch: Updated in ef9c44522aa804204730a18ac827b2280d5ec587

rafal-ch:
>We need to have a place where we cache this kind of information. Maybe, we could extend consensus parameters provider to also provide this kind of information, because fetch it from the database it a lot of waste work.

Done in: https://github.com/FuelLabs/fuel-core/pull/2715/commits/b861a51a58f29d9cef044ebc99fb76e731a2c02d

crates/fuel-core/src/graphql_api/extensions/current_stf_version.rs
rafal-ch: Currently, the STF version is cached in the "Consensus Parameters Provider", so we do not read from the DB every time, but handling of the current block height remains unchanged from the original implementation.

crates/fuel-core/src/graphql_api/worker_service.rs
```diff
@@ -195,6 +210,10 @@ where
 
         transaction.commit()?;
 
+        let current_stf_version = block.header().state_transition_bytecode_version();
+        self.stf_parameters_provider
+            .cache_stf_version(current_stf_version);
```
xgreenx, line 215: We can have this login inside of the `ConsensusParametersProvider`. We don't need to teach off-chain worker to provide this information=)

image

crates/fuel-core/src/schema/coins.rs
```diff
@@ -117,7 +117,7 @@ impl MessageCoin {
     async fn asset_id(&self, ctx: &Context<'_>) -> AssetId {
         let params = ctx
             .data_unchecked::<ConsensusProvider>()
-            .latest_consensus_params();
+            .current_consensus_params();
```
xgreenx, line 120: May I also ask you to replace `Mutex` in the `ConsensusProvider` with `RwLock` please?=)

Since we only read in most of the cases and write happens once per month

crates/fuel-core/src/service/adapters/consensus_parameters_provider.rs
```diff
@@ -38,6 +41,7 @@ pub struct SharedState {
         SharedMutex<ConsensusParametersVersion>,
     pub(crate) consensus_parameters:
         SharedMutex<HashMap<ConsensusParametersVersion, Arc<ConsensusParameters>>>,
+    pub(crate) latest_stf_version: SharedMutex<StateTransitionBytecodeVersion>,
```
xgreenx, line 44: May I also ask you to replace `Mutex`  with `RwLock` for `consensus_parameters` and `latest_stf_version` please?=)

Since we only read in most of the cases and write happens once per month

xgreenx (commented):
Looks good to mee, small things to change and we are good to go=)

crates/fuel-core/src/graphql_api/api_service.rs
```diff
@@ -291,11 +294,14 @@ where
             max_queries_resolver_recursive_depth,
         ))
         .extension(async_graphql::extensions::Tracing)
+        // `RequiredFuelBlockHeightExtension` uses the view set by the ViewExtension.
+        // Do not reorder this line before adding the `ViewExtension`.
```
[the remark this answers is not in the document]
rafal-ch: Done in 3f32362701672bc6c2ef2187cfca212f3886e181

crates/fuel-core/src/schema/coins.rs
rafal-ch: Done in d8ea573b14fe0092ebd4c6dc75556e223313973e

crates/fuel-core/src/service/adapters/consensus_parameters_provider.rs
rafal-ch: Done in https://github.com/FuelLabs/fuel-core/commit/d8ea573b14fe0092ebd4c6dc75556e223313973e

crates/fuel-core/src/graphql_api/worker_service.rs
rafal-ch: You're right, that was a bit of a code smell since the beginning.
Updated in 932a1c55506b548fc60ed0527b979559cbdf9124

crates/client/src/client.rs
```diff
@@ -333,6 +388,8 @@ impl FuelClient {
             .await
             .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
 
+        self.update_chain_state_info(&response);
```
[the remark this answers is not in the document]
rafal-ch: Updated in https://github.com/FuelLabs/fuel-core/pull/2715/commits/3aa2d2d07ce7f8dc89a192ad0b390775163896fb

crates/fuel-core/src/service/adapters/consensus_parameters_provider.rs
```diff
@@ -120,23 +170,29 @@ impl RunnableTask for Task {
...
-                        }
-                    }
+                    .header();
+
+                let new_consensus_parameters_version = header.consensus_parameters_version();
+                if !update_cached_value(
+                    &self.shared_state.latest_consensus_parameters_version,
+                    new_consensus_parameters_version,
+                    || self.shared_state.cache_consensus_parameters(new_consensus_parameters_version).map(|_| ())
+                ) {
```
xgreenx, line 183: You have a deadlock here.  `self.shared_state.cache_consensus_parameters` is doing write lock inside and `update_cached_value` is doing upgradable lock

xgreenx (approved):

crates/fuel-core/src/graphql_api/extensions/chain_state_info.rs
```diff
@@ -0,0 +1,77 @@
...
+
+        let consensus_parameters_provider = ctx.data_unchecked::<ConsensusProvider>();
+        let current_consensus_parameters_version =
+            consensus_parameters_provider.current_consensus_parameters_version();
+        response.extensions.insert(
+            CURRENT_CONSENSUS_PARAMETERS_VERSION.to_string(),
+            Value::Number(current_consensus_parameters_version.into()),
+        );
+
+        let current_stf_version = consensus_parameters_provider.current_stf_version();
```
acerone85, line 62: nit: the name `consensus_parameters_provider` should be changed to something that indicates that we also return the stf version, but probably fine to rename it later.

crates/services/src/sync.rs
```diff
@@ -41,3 +41,34 @@ impl<T> From<T> for SharedMutex<T> {
...
+
+/// A RwLock that can safely be in async contexts and avoids deadlocks.
+#[derive(Default, Debug)]
+pub struct SharedRwLock<T>(Shared<parking_lot::RwLock<T>>);
+
+impl<T> Clone for SharedRwLock<T> {
+    fn clone(&self) -> Self {
+        Self(self.0.clone())
+    }
+}
```
acerone85, line 53: this seems to be equivalent to `#[derive(Clone)]`, or am I missing something?

code generated by `#[derive(Clone)]`:
```rust
impl<T: core::clone::Clone> core::clone::Clone for SharedRwLock<T> {
    fn clone(&self) -> Self {
        match self {
            SharedRwLock(f0) => SharedRwLock(f0.clone()),
        }
    }
}
```

crates/fuel-core/src/service/adapters/consensus_parameters_provider.rs
```diff
@@ -56,9 +60,13 @@ impl Debug for Task {
 
 impl SharedState {
     fn new(database: Database) -> Self {
+        // We set versions in the `into_task` function during start of the service.
         let genesis_version = 0;
+        let latest_stf_version = 0;
```
acerone85, line 65: for consistency: 

```suggestion
        let genesis_consensus_parameters_version = 0;
        let genesis_stf_version = 0;
```

acerone85 (approved):
Thanks :) 
All my comments are minor and definitely not show stoppers.

crates/fuel-core/src/graphql_api/extensions/chain_state_info.rs
rafal-ch: Yes, I agree. Here's the follow-up issue: https://github.com/FuelLabs/fuel-core/issues/2754

crates/fuel-core/src/graphql_api/ports.rs
```diff
@@ -459,11 +462,18 @@ pub mod worker {
 }
 
 pub trait ConsensusProvider: Send + Sync {
-    /// Returns latest consensus parameters.
-    fn latest_consensus_params(&self) -> Arc<ConsensusParameters>;
+    /// Returns current consensus parameters.
+    fn current_consensus_params(&self) -> Arc<ConsensusParameters>;
```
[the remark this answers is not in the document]
rafal-ch: Not sure, I just did the rename, didn't investigate further.

crates/services/src/sync.rs
rafal-ch: I noticed that custom clone is required for the `SharedMutex` above, which I used as a base for the `SharedRwLock` and I decided to just follow the pattern. In this case, it seems that the derived `Clone` would do the job, indeed.

crates/fuel-core/src/service/adapters/consensus_parameters_provider.rs
rafal-ch: I'll rename this in a follow-up PR in order no to block this one any longer.

crates/fuel-core/src/graphql_api/extensions/chain_state_info.rs
rafal-ch: Follow-up PR: https://github.com/FuelLabs/fuel-core/pull/2761

PR #2726, opened 2025-02-18
Add preconfirmation gossip messages

MitchTurner:
## Linked Issues/PRs

Closes https://github.com/FuelLabs/fuel-core/issues/2738

## Description

This is part of the greater tx confirmation work. This will enable the block producer to gossip a batch of signed transactions around the network for the senders of those transactions to receive.

~In an attempt to cleanup the P2P code, some of the tests were moved to their own files (this makes reviewing slightly more difficult, so I've marked the new tests with comments in those refactored test suites).~

TODO:

- [x] Finalize the `TxConfirmations` type
    - [x] What is the correct type for `signature`?
    - [x] What are all the statuses for the `txs`?
- [x] Default to having the Preconfirmation topic off to minimize traffic, turn on with CLI arg
- [ ] ~Constrain peers for gossiping to minimize traffic~ https://github.com/FuelLabs/fuel-core/issues/2740

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/p2p/src/service/task_tests.rs
```diff
@@ -0,0 +1,571 @@
...
+    FuelP2PEvent::GossipsubMessage {
+        peer_id,
+        message_id,
+        topic_hash,
+        message,
+    }
+}
+
+#[tokio::test]
+async fn run__gossip_message_from_p2p_service_is_broadcasted__tx_confirmations() {
```
MitchTurner, line 520: This is the only new test. The rest were just copied over during the refactor.

crates/services/p2p/src/p2p_service/tests.rs
```diff
@@ -0,0 +1,1388 @@
...
+            ),
+        )
+        .await
+        .unwrap();
+    }
+}
+
+#[tokio::test]
+#[instrument]
+async fn gossipsub_broadcast_tx_with_reject__tx_confirmations() {
```
MitchTurner, line 654: New test.

crates/services/p2p/src/p2p_service/tests.rs
```diff
@@ -0,0 +1,1388 @@
...
+            ),
+        )
+        .await
+        .unwrap();
+    }
+}
+
+#[tokio::test]
+#[instrument]
+async fn gossipsub_broadcast_tx_with_accept__tx_confirmations() {
```
MitchTurner, line 612: New test.

crates/services/p2p/src/p2p_service/tests.rs
```diff
@@ -0,0 +1,1388 @@
...
+                // never ending loop
+                // break;
+            }
+        }
+    }
+}
+
+// TODO: Move me before tests that use this function
+/// Reusable helper function for Broadcasting Gossipsub requests
+async fn gossipsub_broadcast(
```
MitchTurner, line 792: This method was updated to accept the new `GossipsubBroadcastRequest::Confirmations` variant.

Introduced `check_message_matches_request` as part of that (before it was hardcoded to look for a specific value).

crates/services/p2p/src/p2p_service.rs
```diff
@@ -901,27 +901,33 @@ impl FuelP2PService {
     }
 }
 
-#[allow(clippy::cast_possible_truncation)]
 #[cfg(test)]
 mod tests {
+    #![allow(non_snake_case)]
+    #![allow(clippy::cast_possible_truncation)]
```
netrome, line 907: Perhaps unrelated to this PR but why do we need this one?

crates/types/src/services/p2p.rs
```diff
@@ -72,6 +78,108 @@ pub struct GossipData<T> {
...
+pub type ConfirmationsGossipData = GossipData<PreConfirmationMessage>;
+
+/// A value and an associated signature
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Sealed<Entity> {
+    /// The actual value
+    pub entity: Entity,
+    /// Seal
+    pub signature: Signature,
```
netrome, line 91: Nit: I'd suggest renaming the signature field to `seal` or renaming the type to `Signed` so the naming is more consistent. There's also a "todo" to do the same renaming in the sibling struct in the `blockchain/consensus.rs` module.
```suggestion
    /// Seal
    pub seal: Signature,
```

netrome (approved):
Just a nit and a question, otherwise looks good to me.

crates/services/p2p/src/p2p_service.rs
MitchTurner: You're right, it was already here. In general, I am in favor of allowing panickable arithmetic and dangerous casts in tests, for the same reason I'm okay with `unwrap` in tests.

I'm not sure specifically what this is enabling in the tests, but I'd encourage this pattern.

crates/types/src/services/p2p.rs
MitchTurner: I prefer `Signature`. `Seal` is ambiguous IMO.

crates/services/p2p/src/service.rs
```diff
@@ -371,6 +373,11 @@ pub trait Broadcast: Send {
 
     fn tx_broadcast(&self, transaction: TransactionGossipData) -> anyhow::Result<()>;
 
+    fn confirmations_broadcast(
```
AurelienFT, line 376: nit: maybe call it pre-confirmation?

crates/types/src/services/p2p.rs
```diff
@@ -72,6 +78,108 @@ pub struct GossipData<T> {
...
+}
+
+/// A collection of pre-confirmations that have been signed by a delegate
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Preconfirmations {
+    /// The expiration time of the key used to sign
+    expiration: Tai64,
+    /// The transactions which have been pre-confirmed
+    preconfirmations: Vec<Preconfirmation>,
```
AurelienFT, line 145: What's the idea behind having a vector ? Because we don't have for NewTx and what would be the rule to end a batch of `Preconfirmation` and send it ?

crates/services/p2p/src/p2p_service.rs
netrome: Yeah agreed.

crates/types/src/services/p2p.rs
netrome: Should the type then be called `Signed` instead of `Sealed` perhaps?

crates/services/p2p/src/service.rs
MitchTurner: Thanks. I iterated on the name and this got left behind.

crates/types/src/services/p2p.rs
MitchTurner: We figured it would be more efficient to sign batches of txs, rather than one at a time. That was especially true with KMS signatures, but maybe we can reconsider with the delegate key signing.

MitchTurner: > Because we don't have for NewTx and what would be the rule to end a batch of Preconfirmation and send it ?
I'm not sure I understand the question completely, but we would probably rely on the executor to return its best values or something. TBD.

AurelienFT (dismissed):

netrome (approved):

crates/services/p2p/src/gossipsub/topics.rs
```diff
@@ -10,22 +10,26 @@ use super::messages::{
 };
 
 pub const NEW_TX_GOSSIP_TOPIC: &str = "new_tx";
+pub const TX_PRECONFIRMATIONS_GOSSIP_TOPIC: &str = "tx_confirmations";
```
Dentosal, line 13: question: why it's called both "confirmations" and "preconfirmations"?

MitchTurner: Holdover from earlier naming. Good catch.

Dentosal (approved):

AurelienFT (approved):

netrome (approved):

PR #2731, opened 2025-02-20
fault_proving(compression): temporal registry implementations

rymnc:
## Linked Issues/PRs

part 2/3 of https://github.com/FuelLabs/fuel-core/issues/2568

## Description

Implementation of the `TemporalRegistry` trait for version 2 (v2) tables and the handling of off-chain database transactions. Below is a summary of the most important changes:

### TemporalRegistry Trait Implementations

* Added `TemporalRegistry` trait implementations for v2 tables, including `Address`, `AssetId`, `ContractId`, `ScriptCode`, and `PredicateCode`. This includes methods for reading, writing, and indexing registry values, as well as handling timestamps.

### Off-Chain Database Transactions

* Introduced conditional compilation for off-chain database transactions, separating v1 and v2 implementations. The v2 implementation includes additional traits and methods to support the new `TemporalRegistry` v2 tables. [[1]](diffhunk://#diff-49095c45844c0809b8d244f4555fb1eaac1ee48612c103e553ab396d0dfe93c9L390-R393) [[2]](diffhunk://#diff-49095c45844c0809b8d244f4555fb1eaac1ee48612c103e553ab396d0dfe93c9R442-R526)
* Updated the `OffChainDatabaseTransaction` trait to include new storage mutation traits specific to v2 tables.

### Code Refactoring

* Refactored the `DaCompressionTemporalEvictorCacheV2Encoder` to use `RegistryKey` instead of `MetadataKey`, ensuring consistency with the new v2 table implementations.

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/fuel-core/src/graphql_api/da_compression.rs
+++ b/crates/fuel-core/src/graphql_api/da_compression.rs
@@ -86,192 +86,399 @@ pub struct DecompressDbTx<'a, Tx, Onchain> {
     pub onchain_db: Onchain,
 }
 
-macro_rules! impl_temporal_registry {
-    ($type:ty) => { paste::paste! {
-        impl<'a, Tx> TemporalRegistry<$type> for DbTx<'a, Tx>
-        where
-            Tx: OffChainDatabaseTransaction,
-        {
-            fn read_registry(
-                &self,
-                key: &fuel_core_types::fuel_compression::RegistryKey,
-            ) -> anyhow::Result<$type> {
-                Ok(self
-                    .db_tx
-                    .storage_as_ref::<[< DaCompressionTemporalRegistry $type >]>()
-                    .get(key)?
-                    .ok_or(not_found!([< DaCompressionTemporalRegistry $type>]))?
-                    .into_owned())
-            }
+#[cfg(not(feature = "fault-proving"))]
+mod v1_impl_temporal_registry {
+    use super::*;
 
-            fn read_timestamp(
-                &self,
-                key: &fuel_core_types::fuel_compression::RegistryKey,
-            ) -> anyhow::Result<Tai64> {
-                Ok(self
-                    .db_tx
-                    .storage_as_ref::<[< DaCompressionTemporalRegistryTimestamps >]>()
-                    .get(&TimestampKey {
-                        keyspace: TimestampKeyspace::$type,
-                        key: *key,
-                    })?
-                    .ok_or(not_found!(DaCompressionTemporalRegistryTimestamps))?
-                    .into_owned())
-            }
-
-            fn write_registry(
-                &mut self,
-                key: &fuel_core_types::fuel_compression::RegistryKey,
-                value: &$type,
-                timestamp: Tai64,
-            ) -> anyhow::Result<()> {
-                // Write the actual value
-                let old_value = self.db_tx
-                    .storage_as_mut::<[< DaCompressionTemporalRegistry $type >]>()
-                    .replace(key, value)?;
-
-                // Remove the overwritten value from index, if any
-                if let Some(old_value) = old_value {
-                    let old_reverse_key = (&old_value).into();
-                    self.db_tx
-                        .storage_as_mut::<DaCompressionTemporalRegistryIndex>()
-                        .remove(&old_reverse_key)?;
+    macro_rules! impl_temporal_registry {
+        ($type:ty) => { paste::paste! {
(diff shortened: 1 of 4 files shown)


crates/fuel-core/src/graphql_api/da_compression.rs
```diff
@@ -86,192 +86,395 @@ pub struct DecompressDbTx<'a, Tx, Onchain> {
...
+        asset_id::DaCompressionTemporalRegistryAssetIdV2,
+        contract_id::DaCompressionTemporalRegistryContractIdV2,
+        evictor_cache::DaCompressionTemporalRegistryEvictorCacheV2,
+        predicate_code::DaCompressionTemporalRegistryPredicateCodeV2,
+        registry_index::DaCompressionTemporalRegistryIndexV2,
+        script_code::DaCompressionTemporalRegistryScriptCodeV2,
+        timestamps::DaCompressionTemporalRegistryTimestampsV2,
+    };
+
+    macro_rules! impl_temporal_registry_v2 {
```
netrome, line 292: This macro seems very repeated from the `impl_temporal_registry`. I think we should consider merging them. It seems like it's mostly the "V2" suffix which needs to be added in a bunch of places, and I believe we could just take that suffix as an argument to the macro. But that can be done as a follow-up, it doesn't have to block this PR.

crates/fuel-core/src/graphql_api/ports.rs
```diff
@@ -387,55 +387,144 @@ pub mod worker {
...
+            + StorageMutate<DaCompressionTemporalRegistryScriptCodeV2, Error = StorageError>
+            + StorageMutate<
+                DaCompressionTemporalRegistryPredicateCodeV2,
+                Error = StorageError,
+            > + StorageMutate<DaCompressionTemporalRegistryIndexV2, Error = StorageError>
+            + StorageMutate<DaCompressionTemporalRegistryTimestampsV2, Error = StorageError>
+            + StorageMutate<
+                DaCompressionTemporalRegistryEvictorCacheV2,
+                Error = StorageError,
+            >
```
netrome, line 497: Since these are the only one that has changed, we could gather these in a separate trait and only feature gate on that single trait instead of having all the other requirements repeated.

netrome (approved):
Some suggestions to reduce code repetition but nothing blocking.

crates/fuel-core/src/graphql_api/da_compression.rs
```diff
@@ -86,192 +86,395 @@ pub struct DecompressDbTx<'a, Tx, Onchain> {
...
+
+    impl_temporal_registry!(Address);
+    impl_temporal_registry!(AssetId);
+    impl_temporal_registry!(ContractId);
+    impl_temporal_registry!(ScriptCode);
+    impl_temporal_registry!(PredicateCode);
+};
+
+#[cfg(feature = "fault-proving")]
+const _: () = {
```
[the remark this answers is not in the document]
rymnc: addressed in a5166f3c5a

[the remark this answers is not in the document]
netrome: Thank you!

crates/fuel-core/src/graphql_api/ports.rs
rymnc: addressed in 5befcaa7ff

netrome (approved):

MitchTurner (approved):
I've gone back and forth between the old and new macros. From what I can understand, there isn't really much going on differently (maybe I missed something Copilot is being notably unhelpful), but more that the dependencies have changed in a way that more traits need to be exposed. 

I don't see any reason to hold this back, but I wonder if there is a way to unify the macros still?

netrome:
> I've gone back and forth between the old and new macros. From what I can understand, there isn't really much going on differently (maybe I missed something Copilot is being notably unhelpful), but more that the dependencies have changed in a way that more traits need to be exposed.
> 
> I don't see any reason to hold this back, but I wonder if there is a way to unify the macros still?

Iirc it's basically the same macros but they're in sub modules which results in a large mostly-whitespace diff. The new macro is for the `*V2` types. I also think they potentially could be unified, but apparently there were some extra complications with this as noted in https://github.com/FuelLabs/fuel-core/pull/2731#discussion_r1963446563.

xgreenx (approved):

PR #2733, opened 2025-02-20
Add pending pool in TxPool module

AurelienFT:
## Description

This pull request introduces a pending pool transaction feature that allows transactions to wait for a missing input instead of being deleted immediately. The main changes include updates to the configuration, error handling, and transaction pool logic.

### Pending Pool Feature:

* Added a pending pool transaction mechanism to allow transactions to wait for missing inputs instead of being deleted immediately.

### Configuration Updates:

* Added `pending_pool_tx_ttl` and `max_pending_pool_size_percentage` to the `Config` struct to manage the time-to-live and size limit for the pending pool.
### Error Handling Enhancements:

* Introduced `InsertionErrorType` and `InputValidationErrorType` enums to handle different error scenarios, including missing inputs.

### Transaction Pool Logic:

* Modified the transaction pool to return `InsertionErrorType` instead of `Error` for better error categorization. 
* Updated the `remove_transaction` method to return removed transactions and notify the pending pool. 

### Pool Worker Adjustments:

* Integrated the pending pool into the `PoolWorker` and added logic to handle transactions with missing inputs.

These changes collectively enhance the transaction pool's robustness by allowing it to handle transactions with missing inputs more gracefully.

## Checklist

### Before requesting review


crates/services/txpool_v2/src/tests/tests_pending_pool.rs
```diff
@@ -0,0 +1,128 @@
...
+    // When
+    tokio::time::timeout(
+        tokio::time::Duration::from_secs(1),
+        universe.waiting_txs_insertion(
+            service.shared.new_tx_notification_subscribe(),
+            vec![tx2.id(&Default::default())],
+        ),
+    )
+    .await
+    .unwrap_err();
```
xgreenx, line 42: I think it makes more sense to subscribe first, and insert tx2 after.

crates/services/txpool_v2/src/tests/tests_pending_pool.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    universe
+        .waiting_txs_insertion(
+            service.shared.new_tx_notification_subscribe(),
+            vec![tx1.id(&Default::default()), tx2.id(&Default::default())],
+        )
+        .await;
+
+    // Then
+    service.stop_and_await().await.unwrap();
```
xgreenx, line 53: I expect that you will listen for the `Submitted` notification from TX2 here.

crates/services/txpool_v2/src/tests/tests_pending_pool.rs
```diff
@@ -0,0 +1,128 @@
...
+    tokio::time::timeout(
+        timeout,
+        universe.waiting_txs_insertion(
+            service.shared.new_tx_notification_subscribe(),
+            vec![tx2.id(&Default::default())],
+        ),
+    )
+    .await
+    .unwrap_err();
+    service.shared.try_insert(vec![tx1.clone()]).unwrap();
```
xgreenx, line 87: I think we want to sleep here for 1 second to be sure that it will be not included

xgreenx (commented):
I checked all files except `pending_pool`, will do that now

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,519 @@
...
+                {
+                    let missing_inputs: Vec<MissingInput> = missing_inputs
+                        .into_iter()
+                        .filter(|input| match input {
+                            MissingInput::Contract(contract) => *contract != contract_id,
+                            _ => true,
+                        })
+                        .collect();
+                    if missing_inputs.is_empty() {
+                        resolved_txs.push(tx);
```
xgreenx, line 183: You don't clean `self.current_bytes` and related fields when you resolve a transaction and remove it from the pending pool(the same for Utxos).

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,519 @@
...
+            {
+                self.current_bytes =
+                    self.current_bytes.saturating_sub(tx.metered_bytes_size());
+                self.current_gas = self.current_gas.saturating_sub(tx.max_gas());
+                self.current_txs = self.current_txs.saturating_sub(1);
+                for input in &missing_inputs {
+                    // Remove tx_id from the list of unresolved transactions for this input
+                    // If the list becomes empty, remove the entry
+                    if let Some(tx_ids) = self.unresolved_txs_by_inputs.get_mut(input) {
+                        tx_ids.retain(|id| *id != *tx_id);
```
xgreenx, line 213: We also could use here `HashSet` to simplify the removal

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,519 @@
...
+                        if tx_ids.is_empty() {
+                            self.unresolved_txs_by_inputs.remove(input);
+                        }
```
xgreenx, line 216: If you used `entry`, then you don't need do search in the `self.unresolved_txs_by_inputs` to times

xgreenx (commented):
I don't see code related on resolution of inputs when we select transactions for the execution and when we import the block. Do you plan to add it in a separate PR?

AurelienFT:
> I don't see code related on resolution of inputs when we select transactions for the execution and when we import the block. Do you plan to add it in a separate PR?

It's this function https://github.com/FuelLabs/fuel-core/blob/e2c17e9d7a997a86f40180be9020c626d438e5a4/crates/services/txpool_v2/src/pending_pool.rs#L226

And it's called from : 
https://github.com/FuelLabs/fuel-core/blob/e2c17e9d7a997a86f40180be9020c626d438e5a4/crates/services/txpool_v2/src/pool_worker.rs#L398
and
https://github.com/FuelLabs/fuel-core/blob/e2c17e9d7a997a86f40180be9020c626d438e5a4/crates/services/txpool_v2/src/pool_worker.rs#L476

AurelienFT:
> But `extract_transactions_for_block` also includes transaction into the block=)

Yeah but the block also endup in `import_block` so I placed it here but yeah it has 1 second delay so I will add also.

crates/services/txpool_v2/src/tests/tests_pending_pool.rs
AurelienFT: If you meant before the insert of tx1 it's already done by the timeout. If you mean after, we already sure that it has been squeezed out because we are waiting for this status in the subscriber.

crates/services/txpool_v2/src/tests/tests_pending_pool.rs
AurelienFT: Done

crates/services/txpool_v2/src/tests/tests_pending_pool.rs
AurelienFT: Done

crates/services/txpool_v2/src/pool_worker.rs
```diff
@@ -481,4 +566,59 @@ where
...
+        let txs_used = self.pending_pool.current_txs.saturating_add(1);
+
+        if gas_used
+            > self
+                .pool
+                .config
+                .pool_limits
+                .max_gas
+                .saturating_mul(self.pool.config.max_pending_pool_size_percentage as u64)
+                .saturating_div(100)
```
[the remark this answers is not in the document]
AurelienFT: Done

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,519 @@
...
+    ) {
+        if let Some(tx_ids) = self
+            .unresolved_txs_by_inputs
+            .remove(&MissingInput::Utxo(utxo_id))
+        {
+            for tx_id in tx_ids {
+                if let Some((tx, missing_inputs)) =
+                    self.unresolved_inputs_by_tx.remove(&tx_id)
+                {
+                    let missing_inputs: Vec<MissingInput> = missing_inputs
```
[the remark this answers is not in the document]
AurelienFT: The problem that I have with `HashSet` is that whenever I will return something to the user because we didn't had the missing input in time and there is multiple missing input then we will not return the first one in the `inputs()` order which creates a breaking change and I wanted to avoid breaking changes (otherwise I could change the error type to match the new multiple errors also)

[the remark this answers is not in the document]
AurelienFT: Maybe I could duplicate the error that I want to be returned and save it along? wdyt @xgreenx ?

crates/services/txpool_v2/src/pending_pool.rs
AurelienFT: I don't see how I can do the deletion of the entry if tx_ids is empty directly with the entry API. Can you provide an example ?

crates/services/txpool_v2/src/pending_pool.rs
AurelienFT: Good catch done.

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,519 @@
...
+        self.current_gas = self.current_gas.saturating_add(transaction.max_gas());
+        self.current_txs = self.current_txs.saturating_add(1);
+        for input in &missing_inputs {
+            self.unresolved_txs_by_inputs
+                .entry(*input)
+                .or_default()
+                .push(tx_id);
+        }
+        self.unresolved_inputs_by_tx.insert(
+            transaction.id(),
```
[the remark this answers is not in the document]
AurelienFT: Done

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,523 @@
...
+                    } else {
+                        self.unresolved_inputs_by_tx
+                            .insert(tx_id, (tx_data, missing_inputs));
+                    }
+                }
+            }
+        }
+    }
+
+    fn new_known_contract(
```
rafal-ch, line 162: We can possibly modify the missing inputs in place using an Entry API. Not sure if this would be a perf improvements, so leaving this up to you.

The code could look something like:

```
match self.unresolved_txs_by_inputs.entry(missing_input) {
            Entry::Occupied(entry) => {
                for tx_id in entry.remove() {
                    if let Entry::Occupied(mut entry) =
                        self.unresolved_inputs_by_tx.entry(tx_id)
                    {
                        let missing_inputs = entry.get_mut();
                        missing_inputs.retain(|input| !matches!(input, MissingInput::Contract(c) if *c == contract_id));
                        if missing_inputs.is_empty() {
                            let tx = entry.remove();
                            self.decrease_pool_size(&tx);
                            resolved_txs.push(tx);
                        }
                    }
                }
            }
            Entry::Vacant(_) => (),
        }
```

bin/fuel-core/src/cli/run/tx_pool.rs
```diff
@@ -83,6 +83,16 @@ pub struct TxPoolArgs {
...
+
+    /// The max time to live of the transaction inside of the `PendingPool`.
+    /// The `PendingPool` is a pool of transactions that are waiting for the `TxPool` to accept them
+    /// because one of their inputs is not yet available.
+    #[clap(long = "tx-pending-pool-ttl", default_value = "3s", env)]
+    pub tx_pending_pool_ttl: humantime::Duration,
+
+    /// The max percentage of the `TxPool` that can be used by the `PendingPool`.
+    #[clap(long = "tx-pending-pool-size-percentage", default_value = "50", env)]
+    pub tx_pending_pool_size_percentage: u16,
```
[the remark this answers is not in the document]
AurelienFT: This type is in `gas-price-algorithm` and there is multiple places where we could use it. We need to move the type out of the crate. I created an issue for that : https://github.com/FuelLabs/fuel-core/issues/2773

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,523 @@
...
+    pub(crate) current_bytes: usize,
+    pub(crate) current_txs: usize,
+    pub(crate) current_gas: u64,
+}
+
+#[derive(Debug, Hash, Eq, PartialEq, Copy, Clone)]
+pub(crate) enum MissingInput {
+    Utxo(UtxoId),
+    Contract(ContractId),
+}
```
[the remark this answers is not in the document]
AurelienFT: We will not manage them in the first version because they are not created with output of others transaction but from L1 etc and txpool isn't aware of this.

crates/services/txpool_v2/src/pool_worker.rs
```diff
@@ -481,4 +572,53 @@ where
...
+            .saturating_div(100);
+        let max_txs = self
+            .pool
+            .config
+            .pool_limits
+            .max_txs
+            .saturating_mul(self.pool.config.max_pending_pool_size_percentage as usize)
+            .saturating_div(100);
+
+        if gas_used > max_gas || bytes_used > max_bytes || txs_used > max_txs {
```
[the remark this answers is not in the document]
AurelienFT: When inserting Tx in the pool : 
https://github.com/FuelLabs/fuel-core/blob/3d7dc17cc72b035637ab11548d05a19a6e4ffdb2/crates/services/txpool_v2/src/pool.rs#L462

We are not not managing equal as a problem.

Here also:
https://github.com/FuelLabs/fuel-core/blob/216c2b5e3056084d4a5206d6a07e26c0850c8609/crates/services/txpool_v2/src/pool_worker.rs#L585

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,523 @@
...
+//
+// Insertion rules:
+// - If the transaction has one or more inputs that are not known yet.
+//
+// Deletion rules:
+// - If the transaction missing inputs hasn't been known for a certain amount of time.
+// - If the transaction missing inputs becomes known.
+pub(crate) struct PendingPool {
+    ttl: Duration,
+    unresolved_txs_by_inputs: HashMap<MissingInput, HashSet<TxId>>,
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,523 @@
...
+                    Output::Coin { .. } => {
+                        self.new_known_utxo(utxo_id, &mut updated_txs.resolved_txs);
+                    }
+                    Output::ContractCreated { contract_id, .. } => {
+                        self.new_known_contract(
+                            *contract_id,
+                            &mut updated_txs.resolved_txs,
+                        );
+                    }
+                    _ => {}
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,523 @@
...
+        {
+            for tx_id in tx_ids {
+                if let Some((tx_data, missing_inputs)) =
+                    self.unresolved_inputs_by_tx.remove(&tx_id)
+                {
+                    let missing_inputs: Vec<MissingInput> = missing_inputs
+                        .into_iter()
+                        .filter(|input| match input {
+                            MissingInput::Contract(contract) => *contract != contract_id,
+                            _ => true,
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,523 @@
...
+        {
+            for tx_id in tx_ids {
+                if let Some((tx_data, missing_inputs)) =
+                    self.unresolved_inputs_by_tx.remove(&tx_id)
+                {
+                    let missing_inputs: Vec<MissingInput> = missing_inputs
+                        .into_iter()
+                        .filter(|input| match input {
+                            MissingInput::Utxo(utxo) => *utxo != utxo_id,
+                            _ => true,
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/services/txpool_v2/src/pool_worker.rs
```diff
@@ -377,8 +395,57 @@ where
                         tracing::error!("Failed to send removed notification: {}", e);
                     }
                 }
+                let updated_txs = self.pending_pool.new_known_txs(vec![tx]);
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/services/txpool_v2/src/pending_pool.rs
AurelienFT: Done.

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,523 @@
...
+            .insert(tx_id, ((transaction, insertion_source), missing_inputs));
+        self.ttl_check.push_front((
+            SystemTime::now()
+                .checked_add(self.ttl)
+                .expect("The system time should be valid; qed"),
+            tx_id,
+        ));
+    }
+
+    fn new_known_utxo(
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/services/txpool_v2/src/pending_pool.rs
AurelienFT: Done.

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,523 @@
...
+        let utxo = UtxoId::new(dependency_tx.id(), 0);
+        let dependent_tx = create_pool_tx(vec![utxo], vec![], &mut rng);
+
+        // When
+        pending_pool.insert_transaction(
+            dependent_tx.clone(),
+            InsertionSource::RPC {
+                response_channel: None,
+            },
+            vec![MissingInput::Utxo(utxo)],
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/services/txpool_v2/src/pool_worker.rs
```diff
@@ -377,8 +395,57 @@ where
                         tracing::error!("Failed to send removed notification: {}", e);
                     }
                 }
+                let updated_txs = self.pending_pool.new_known_tx(tx);
+
+                for (tx, source) in updated_txs.resolved_txs {
+                    self.insert(tx, source);
```
xgreenx, line 401: While I like the fact that we will insert transactions without delay, this can be very long recursion leading to stack overflow. So I will suggest to queue transaction via the channel=)

crates/services/txpool_v2/src/pool_worker.rs
```diff
@@ -391,6 +458,12 @@ where
                 }
             }
         }
+        // Allow in tests to use a ttl of 0 for pending pool and directly
+        // trigger the removal of expired transactions
+        #[cfg(any(feature = "test-helpers", test))]
+        {
+            self.remove_executed_transactions(vec![]);
+        }
```
xgreenx, line 466: What does it change, if we don't do that?

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,585 @@
...
+                self.decrease_pool_size(&tx);
+                for input in &missing_inputs {
+                    // Remove tx_id from the list of unresolved transactions for this input
+                    // If the list becomes empty, remove the entry
+                    if let Some(tx_ids) = self.pending_txs_by_inputs.get_mut(input) {
+                        tx_ids.remove(&tx_id);
+                        if tx_ids.is_empty() {
+                            self.pending_txs_by_inputs.remove(input);
+                        }
+                    }
```
xgreenx, line 184:
```suggestion
                    if let Entry::Occupied(mut tx_ids) =
                        self.pending_txs_by_inputs.entry(*input)
                    {
                        tx_ids.get_mut().remove(&tx_id);
                        if tx_ids.get().is_empty() {
                            tx_ids.remove();
                        }
                    }
```

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,585 @@
...
+// - If the transaction has one or more inputs that are not known yet.
+//
+// Deletion rules:
+// - If the transaction missing inputs hasn't been known for a certain amount of time.
+// - If the transaction missing inputs becomes known.
+pub(crate) struct PendingPool {
+    ttl: Duration,
+    pending_txs_by_inputs: HashMap<MissingInput, HashSet<TxId>>,
+    pending_inputs_by_tx:
+        HashMap<TxId, ((ArcPoolTx, InsertionSource), Vec<MissingInput>)>,
```
xgreenx, line 45: It would be nice to use some structure instead of `((ArcPoolTx, InsertionSource), Vec<MissingInput>)`, hard to remember where do we store what=D

xgreenx: Also, why do we use`((ArcPoolTx, InsertionSource), Vec<MissingInput>)` instead of ` (ArcPoolTx, InsertionSource, Vec<MissingInput>)`?

crates/services/txpool_v2/src/pool_worker.rs
AurelienFT: It means that in test like https://github.com/FuelLabs/fuel-core/blob/216c2b5e3056084d4a5206d6a07e26c0850c8609/crates/services/txpool_v2/src/tests/tests_p2p.rs#L382

crates/services/txpool_v2/src/pending_pool.rs
AurelienFT: Done.

Leftover from previous version.

crates/services/txpool_v2/src/pending_pool.rs
AurelienFT: Done

crates/services/txpool_v2/src/pool_worker.rs
AurelienFT: Done.

crates/services/txpool_v2/src/pool_worker.rs
xgreenx: But I think it still should happen after TTL passed in next Tokio task, no?

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,609 @@
...
+            tx_id,
+        ));
+    }
+
+    fn new_known_input_from_output(
+        &mut self,
+        utxo_id: UtxoId,
+        output: &Output,
+        resolved_txs: &mut Vec<(ArcPoolTx, InsertionSource)>,
+    ) {
```
MitchTurner, line 148: Ah. This confused me. I would expect this function to take the outputs and return any txs that are no longer pending. Instead we are pushing them to a `mut Vec`.

My background in FP has biased me against passing mut variables unless it's much simpler.

What do you think about doing a `map().fold()` here:
```rs
        let mut updated_txs = UpdatePendingTxs {
            resolved_txs: Vec::new(),
            expired_txs: Vec::new(),
        };
        // Resolve transactions
        for tx in new_known_txs {
            self.new_known_tx_inner(tx, &mut updated_txs);
        }
```
i.e. something like

```rs
        let (resolved_txs, expired_txs) = new_known_txs
            .iter()
            .map(self.new_known_tx_inner)
            .fold((Vec::new(), Vec::new()), |(acc_resolved, acc_expired), (resolved, expired)|(acc_resolved.append(resolved), acc_expired.append(acc_expired)));
        
        let update_tx = UpdatePendingTxs {
            resolved_txs,
            expired_txs,
        };
```

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,609 @@
...
+        )
+        .expect("Failed to estimate predicates");
+        let tx = tx
+            .into_checked_basic(1u32.into(), &ConsensusParameters::default())
+            .unwrap();
+        Arc::new(PoolTransaction::Script(tx, Metadata::new(1, 1, 0)))
+    }
+
+    #[test]
+    fn test_pending_pool_one_tx_one_dependent_input_utxo() {
```
MitchTurner, line 396: For these tests. I don't think there is any need to specify `test_pending_pool` for all of them--that's what the namespace is for.

The pending pool has 2 pub methods: `insert_transaction` and `new_known_txs`. I'd suggest putting the tests in terms of those methods. 

I'd reframe this specific test as `new_known_tx__resturns_expected_tx_when_dependent_input_provided`

That begs the question: is it possible to test `insert_transaction` without calling `new_known_tx`?

Looks like great coverage in general though.

MitchTurner (commented):
I'm about half-way through the PR. Easy to understand code. I have a few ideas for small changes.

crates/services/txpool_v2/src/pending_pool.rs
AurelienFT: Yeah it's possible !

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,609 @@
...
+//
+// Insertion rules:
+// - If the transaction has one or more inputs that are not known yet.
+//
+// Deletion rules:
+// - If the transaction missing inputs hasn't been known for a certain amount of time.
+// - If the transaction missing inputs becomes known.
+pub(crate) struct PendingPool {
+    ttl: Duration,
+    pending_txs_by_inputs: HashMap<MissingInput, HashSet<TxId>>,
```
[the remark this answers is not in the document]
xgreenx: You can have many transactions that uses the same UtxoId and have different set of inputs. And all of these inputs may be not resolved yet, in this case the first transaction that is fully resolved(all its inputs are known to the TxPool) will enter TxPool, other will be discarded(unless they have higher tip)/

crates/services/txpool_v2/src/pool_worker.rs
xgreenx: Okay, I see that we don't have any trigger for TTL and because of that no-one will execute it. We only clean up pending pool if we have known transaction.

Maybe, we want to try to clean up TxPool each time when we insert something or have a timer that will clean it up. Because if someone only spams us with pending transactions, then the pending pool will grow.

But maybe it is fine, idk, since it is very hard to exploit in the real network. Plus, we have limit on the size of the pending pool.

AurelienFT: i think its fine it's each time we insert a new pending that we are checking and not checking when we insert a normal tx so I think it's ok it cannot grow more than 3 seconds if there is a new block or a new pending coming

crates/services/txpool_v2/src/pending_pool.rs
AurelienFT: I just want to avoid all necessary re-alloc. This code will be triggered everytime we insert a tx and from the benches I made allocations can take a lot of time. If we let the functions return their own vector we will not benefit from the allocation of the big one that might already be 8/16/32/64... and we will have to alloc a new little one for each function.

crates/services/txpool_v2/src/pending_pool.rs
AurelienFT: I have renamed the tests. Tell me if it's better :)

AurelienFT:
@MitchTurner Ty for your kind words :p

xgreenx (approved):

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -0,0 +1,610 @@
...
+            InsertionSource::RPC {
+                response_channel: None,
+            },
+            vec![MissingInput::Utxo(utxo_1), MissingInput::Utxo(utxo_2)],
+        );
+        let new_known_txs = vec![dependency_tx.clone()];
+        let updated_txs = pending_pool.new_known_txs(new_known_txs);
+
+        // Then
+        assert_eq!(updated_txs.resolved_txs.len(), 1);
```
MitchTurner, line 573: It says "expected_two" but returns 1

crates/services/txpool_v2/src/tests/tests_pending_pool.rs
```diff
@@ -0,0 +1,132 @@
...
+    service.shared.try_insert(vec![tx2.clone()]).unwrap();
+
+    // When
+    let mut subscriber_status = service.shared.tx_update_subscribe(tx_id2).unwrap();
+    let status = subscriber_status.next().await.unwrap();
+
+    // Then
+    assert_eq!(status, TxStatusMessage::Status(TransactionStatus::SqueezedOut { reason:
+        "Transaction input validation failed: UTXO (id: cf6532b2371b3efb71ff8ca4ec32cf046fb1be64e620d21a0a98c0298f8196140000) does not exist".to_string(),
+    }));
```
MitchTurner, line 129: I'm not sure what this error has to do with `removed_not_enough_space`. I feel like there is something implicit happening.

crates/services/txpool_v2/src/tests/tests_pending_pool.rs
```diff
@@ -0,0 +1,132 @@
...
+    assert!(matches!(
+        status,
+        TxStatusMessage::Status(TransactionStatus::Submitted { .. })
+    ));
+
+    service.stop_and_await().await.unwrap();
+}
+
+#[tokio::test]
+async fn test_tx_kept_expired() {
```
MitchTurner, line 61: What is the expired thing that is being kept?

It looks like we are asserting an error in the `then` section, but the name says nothing about that.

crates/services/txpool_v2/src/tests/tests_pending_pool.rs
```diff
@@ -0,0 +1,132 @@
...
+};
+use futures::StreamExt;
+
+use crate::{
+    tests::universe::TestPoolUniverse,
+    TxStatusMessage,
+};
+
+#[tokio::test]
+async fn test_tx_keep_missing_input() {
```
MitchTurner, line 17: So, this should pass when the tx containing the dependent input for the pending tx is submitted? Maybe we include that in the test name.

crates/services/txpool_v2/src/tests/tests_pending_pool.rs
AurelienFT: Added a comment

crates/services/txpool_v2/src/tests/tests_pending_pool.rs
AurelienFT: Done.

crates/services/txpool_v2/src/tests/tests_pending_pool.rs
AurelienFT: I changed the name of the test to be more explicit.

MitchTurner (approved):

rymnc (approved):

MitchTurner (approved):

PR #2746, opened 2025-02-21
Fixes for several flaky tests and re-work of CLI arguments

xgreenx:
Remove flakiness from e2e tests. We don't need to check the `UtxoId`, since the transaction was successful. It is possible that another test consumed the coin of Bob already, so that is why it doesn't exist.

To speed up the compatibility test for V36, use native execution(it requires specifying `10` in the chain config of the v36 node). If p2p gets disconnected, resolve flakiness by using interval block production(when we create a new block and push information about it, it restart the synchronization in old version of the `fuel-core`). Speed up compatibility tests by syncing blocks faster(before we propagated information about each block only once per second).

Reworked all CLI arguments related to Duration to use `humantime::Duration` instead of seconds.


.changes/breaking/2746.md
```diff
@@ -0,0 +1,22 @@
+Breaking changes to the CLI arguments:
+- To disable `random-walk` just don't specify it. Before it required to use `--random-walk 0`.
```
AurelienFT, line 2: Looking at. the code before it was : `default_value = 0` which means no need to add it explicitly.

AurelienFT (approved):

crates/fuel-core/src/graphql_api.rs
```diff
@@ -59,7 +59,7 @@ pub struct ServiceConfig {
     pub costs: Costs,
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
```
MitchTurner, line 62: Do we want copy on something this big? This can be over 1280 bits.

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -814,6 +824,9 @@ where
     /// This is a slow operation, and the cached result should be used if possible,
     /// for instancy by calling `get_module` below.
     #[cfg(feature = "wasm-executor")]
+    // TODO: This methods may compile the WASM binary inside. It can take a lot of time,
+    //  blocking the runtime. Instead, we should make this function async and compile
+    //  the WASM in a separate thread.
```
MitchTurner, line 829: Did we create an issue for this? Seems worth tracking as this could cause opaque bugs in the future.

crates/fuel-core/src/graphql_api.rs
xgreenx: 160 bytes sounds like something small and shouldn't affect eprformance=)

crates/services/upgradable-executor/src/executor.rs
xgreenx: https://github.com/FuelLabs/fuel-core/issues/2755

crates/fuel-core/src/graphql_api.rs
MitchTurner: If you say so. I don't have any rules for how big something should be to only implement `Clone`, but this is just bigger than most things I see.

crates/fuel-core/src/graphql_api.rs
```diff
@@ -120,9 +120,23 @@ pub fn query_costs() -> &'static Costs {
...
+    if !balances_indexation_enabled {
+        cost.balance_query = BALANCES_QUERY_COST_WITHOUT_INDEXATION;
+    }
+
+    cost
+}
+
+fn initialize_query_costs(
+    costs: Costs,
+    _balances_indexation_enabled: bool,
```
MitchTurner, line 136: nit: 
```suggestion
    balances_indexation_enabled: bool,
```

crates/services/upgradable-executor/src/executor.rs
MitchTurner: It's probably best to included it inline.

PR #2750, opened 2025-02-22
Dedicated threads for the block importer

xgreenx:
Removed cloning of the block and utilized the feature of rayon runtime with scoped parallel jobs execution. 
Moved code to catch `InvalidDatabaseStateAfterExecution` to the right place. 
Removed requirements from the database to be modifiable. 
Fixed the not working `async` part of the block importer. Before importing of the block were blocking the tokio task and `await` never was reached.

### DIFF
--- a/tests/tests/recovery.rs
+++ b/tests/tests/recovery.rs
@@ -167,6 +167,8 @@ async fn _gas_price_updater__can_recover_on_startup_when_gas_price_db_is_behind(
         Some(BlockHeight::new(height))
     );
 
+    tokio::time::sleep(std::time::Duration::from_millis(20)).await;
+
     let diff = height - lower_height;
     for _ in 0..diff {
         let _ = database.gas_price().rollback_last_block();
--- a/tests/tests/trigger_integration/instant.rs
+++ b/tests/tests/trigger_integration/instant.rs
@@ -50,7 +50,7 @@ async fn poa_instant_trigger_is_produces_instantly() {
         )
         .add_fee_input()
         .finalize_as_transaction();
-        let _tx_id = client.submit(&tx).await.unwrap();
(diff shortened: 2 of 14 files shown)


tests/tests/recovery.rs
```diff
@@ -167,6 +167,8 @@ async fn _gas_price_updater__can_recover_on_startup_when_gas_price_db_is_behind(
         Some(BlockHeight::new(height))
     );
 
+    tokio::time::sleep(std::time::Duration::from_millis(20)).await;
```
AurelienFT, line 170: Do you want to sneak this change or it's leftover ?

tests/tests/trigger_integration/instant.rs
```diff
@@ -50,7 +50,7 @@ async fn poa_instant_trigger_is_produces_instantly() {
         )
         .add_fee_input()
         .finalize_as_transaction();
-        let _tx_id = client.submit(&tx).await.unwrap();
+        let _tx_id = client.submit_and_await_commit(&tx).await.unwrap();
```
AurelienFT, line 53: Do you want to sneak this change or it's leftover ?

AurelienFT (approved):
I like the refactor, I was wondering if the LocalRunner with 1 thread would have been an issue when parallel-executor would have been added but I don't think so. So i'm fine

tests/tests/recovery.rs
xgreenx: No, on my machine, this test was failing after the `run` function became non blocking

tests/tests/trigger_integration/instant.rs
xgreenx: No, on my machine, this test was failing after the `run` function became non blocking

xgreenx:
If you okay with the change, you can merge it into your PR=D

PR #2752, opened 2025-02-24
GraphQL API changes to support pre-confirmations

rafal-ch:
Closes https://github.com/FuelLabs/fuel-core/issues/2771

## Description
This PR contains the changes to the GraphQL API supporting the upcoming pre-confirmations feature.

Summary:
1. `TransactionStatus` has grown three new variants: `SuccessDuringBlockProductionStatus`, `SqueezedOutDuringBlockProductionStatus` and `FailureDuringBlockProductionStatus`
    1. these new variants should be considered as pre-confirmations, i.e. they will be gossiped and provided by the client even before the entire block containing the transaction is produced. Technically, they will be signed and gossiped as soon as the transaction was executed.
    2. the old variants will be created in exactly the same way as they were before this change
1. Introduction of `TransactionStatusStorage` type, which is just an old `TransactionStatus` under the new name. Ultimately, we need to convert it to a new type which only contains `Success` and `Failure` variants, since only these two states are only ever persisted in the DB

### Before requesting review


crates/client/assets/schema.sdl
```diff
@@ -335,6 +335,10 @@ input ExcludeInput {
 	messages: [Nonce!]!
 }
 
+type FailureDuringBlockProductionStatus {
+	blockHeight: U32!
+}
+
```
Torres-ssf, line 341: What can lead to a transaction failing with a `FailureDuringBlockProductionStatus`? 

I’ve noticed this status doesn’t include a `reason` property like `FailureStatus` does.

rafal-ch: > What can lead to a transaction failing with a FailureDuringBlockProductionStatus?

There could be a change in state (for example, different gas price) between these two points in time:
1. executor grabs the transaction from the TxPool (where it sits, validated and ready to be executed)
2. executor actually tries to execute the transaction

> I’ve noticed this status doesn’t include a reason property like FailureStatus does.

The exact shape of the API is not finalized yet. We decided to add the `reason` to `FailureDuringBlockProductionStatus` as well. There will be more commits added to this PR.

rafal-ch: `reason` is now added to the `FailureDuringBlockProductionStatus` variant

crates/client/assets/schema.sdl
```diff
@@ -1206,6 +1218,13 @@ type Subscription {
 	submitAndAwaitStatus(tx: HexString!): TransactionStatus!
 }
 
+type SuccessDuringBlockProductionStatus {
+	txPointer: TxPointer!
+	transactionId: TransactionId
```
AurelienFT, line 1223: In which case tx are not present here ?

crates/client/assets/schema.sdl
```diff
@@ -335,6 +335,14 @@ input ExcludeInput {
 	messages: [Nonce!]!
 }
 
+type FailureDuringBlockProductionStatus {
+	reason: String!
+	txPointer: TxPointer!
+	transactionId: TransactionId
```
AurelienFT, line 341: In which case tx are not present here ?

crates/client/src/client/schema/tx.rs
```diff
@@ -187,8 +190,11 @@ impl TryFrom<ProgramState> for fuel_vm::ProgramState {
 pub enum TransactionStatus {
     SubmittedStatus(SubmittedStatus),
     SuccessStatus(SuccessStatus),
+    SuccessDuringBlockProductionStatus(SuccessDuringBlockProductionStatus),
```
rafal-ch, line 193: Should we use a name which relates more closely to the pre-confirmations? For example `SuccessPreConfirmed`, `FailurePreConfirmed`, etc.

cc: @xgreenx

crates/fuel-core/src/graphql_api/storage/transactions.rs
```diff
@@ -67,7 +67,7 @@ pub struct TransactionStatuses;
 impl Mappable for TransactionStatuses {
     type Key = Bytes32;
     type OwnedKey = Self::Key;
-    type Value = TransactionStatus;
+    type Value = TransactionStatusStorage; // TODO[RC]: TransactionStatusStorage should ultimately only contain `Success` and `Failed` variants. We don't need to persist any other states.
```
xgreenx, line 70: Do you want to solve it in this PR, or in follow-up?

crates/fuel-core/src/query/subscriptions/test.rs
```diff
@@ -100,11 +121,19 @@ enum TxStatus {
...
     /// because it was not valid to include in the block.
     Squeezed,
+    /// Transaction was eligible for execution and inclusion in the block but
+    /// it was squeezed out by block producer.
+    SqueezedOutDuringBlockProduction,
     /// The transaction failed to execute and was included in a block.
     Failed,
+    /// Transaction was eligible for execution and inclusion in the block but
+    /// it failed during the execution.
+    FailureDuringBlockProduction,
```
xgreenx, line 136: IT would be nice to have parity on `Failed` and `Failure`

crates/types/src/services/txpool.rs
```diff
@@ -333,7 +334,7 @@ impl From<&PoolTransaction> for CheckedTransaction {
 /// The status of the transaction during its life from the tx pool until the block.
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub enum TransactionStatus {
+pub enum TransactionStatusStorage {
```
xgreenx, line 337: It would be nice to use some other name, since we also use this type as a result of execution, maybe `TransactionExecutionStatus` , idk

crates/fuel-core/src/schema/tx/types.rs
```diff
@@ -272,6 +311,47 @@ impl FailureStatus {
...
+
+    async fn transaction_id(&self) -> Option<TransactionId> {
+        self.tx_id.map(Into::into)
+    }
+
+    #[graphql(complexity = "query_costs().storage_read + child_complexity")]
+    async fn transaction(
+        &self,
+        ctx: &Context<'_>,
+    ) -> Option<async_graphql::Result<Transaction>> {
```
xgreenx, line 340:
```suggestion
    ) -> async_graphql::Result<Option<Transaction>> {
```

crates/fuel-core/src/schema/tx/types.rs
```diff
@@ -272,6 +311,47 @@ impl FailureStatus {
...
+        self.tx_id.map(|tx_id| {
+            let query = ctx.read_view()?;
+            let transaction = query.transaction(&tx_id)?;
```
xgreenx, line 343: Usually pre-confirmed transaction will live in TxPool, not in the database, so I would suggest to go there=D

crates/fuel-core/src/schema/tx/types.rs
```diff
@@ -272,6 +311,47 @@ impl FailureStatus {
...
+            Ok(Transaction::from_tx(tx_id, transaction))
+        })
+    }
+
+    async fn receipts(&self) -> Option<async_graphql::Result<Vec<Receipt>>> {
```
xgreenx, line 348: Why do we need to return an error? Looks like we can't have error here

crates/fuel-core/src/schema/tx/types.rs
```diff
@@ -212,6 +215,42 @@ impl SuccessStatus {
...
 
+#[derive(Debug)]
+pub struct SuccessDuringBlockProductionStatus {
+    pub tx_pointer: TxPointer,
+    pub tx_id: Option<TxId>,
+    pub receipts: Option<Vec<fuel_tx::Receipt>>,
+}
+
+#[Object]
+impl SuccessDuringBlockProductionStatus {
```
xgreenx, line 226: Same comments as for failure

crates/fuel-core/src/schema/tx/types.rs
```diff
@@ -212,6 +215,42 @@ impl SuccessStatus {
     }
 }
 
+#[derive(Debug)]
+pub struct SuccessDuringBlockProductionStatus {
+    pub tx_pointer: TxPointer,
+    pub tx_id: Option<TxId>,
```
xgreenx, line 221: `tx_id` always will exist in pre-confirmation

crates/fuel-core/src/schema/tx/types.rs
rafal-ch: I did it like that for consistency with already existing types, like `SuccessStatus` [here](https://github.com/FuelLabs/fuel-core/blob/master/crates/fuel-core/src/schema/tx/types.rs#L202-L204).

crates/fuel-core/src/schema/tx/types.rs
rafal-ch: Updated in https://github.com/FuelLabs/fuel-core/pull/2752/commits/87aaf7ca6dc40434d97f12f90bf717edf891417e

crates/fuel-core/src/query/subscriptions/test.rs
rafal-ch: Updated in https://github.com/FuelLabs/fuel-core/pull/2752/commits/d702ca7e8254d75bc8d9825a74cbda64cfeb4bd5

crates/client/src/client/schema/tx.rs
rafal-ch: Renamed in https://github.com/FuelLabs/fuel-core/pull/2752/commits/a473dbb947220b4961ce634ee400d6ecbefe5d44, as agreed with SDK.

The only minor drawback I see is that in case in the future other components than BlockProducer are able to create preconfirmations we'll not carry the information about the preconfirmation source. Most likely a YAGNI case, that's why I didn't add any kind of "source" field to preconfirmations.

crates/types/src/services/txpool.rs
rafal-ch: `TransactionExecutionStatus` already exists in `executor`: https://github.com/FuelLabs/fuel-core/blob/master/crates/types/src/services/executor.rs#L185-L190

We also seem to already have what we need (albeit there are differences in fields) in the form of `TransactionExecutionResult`: https://github.com/FuelLabs/fuel-core/blob/master/crates/types/src/services/executor.rs#L195-L218

For now I used the explicit qualification (`txpool::TransactionExecutionStatus`) to resolve conflicts to avoid introducing yet another name for transaction status.

Follow-up issue for grooming: https://github.com/FuelLabs/fuel-core/issues/2795

crates/fuel-core/src/schema/tx/types.rs
rafal-ch: Corrected

crates/fuel-core/src/schema/tx/types.rs
rafal-ch: Changed in 38d2733adb4a6ac2e8ef692b1e76054de2437f83

Not really sure here, though. I thought that executor "extracts" transactions from TxPool, so they no longer live there, but I may be not up to speed with the most recent changes to the TxPool and its interactions with Executor.

xgreenx (dismissed):

xgreenx (approved):

AurelienFT (approved):

.changes/changed/2752.md
```diff
@@ -0,0 +1 @@
+Extended the `TransactionStatus` to support pre-confirmations.
```
rymnc, line 1: i think we can be a little verbose here because it's a change in the graphql api :) non blocking, feel free to address in follow up

rymnc (approved):

MitchTurner (approved):

Issue #2739, opened 2025-02-20
Create Preconfirmation Signature Task


MitchTurner:
We need to add a signature to Preconfirmations to prove that they were sent by the block producer.

The KMS signatures take longer than 50ms, which isn't acceptable for preconfirmations. We are targetting 30ms from tx submission.

Instead, every ~10 mins the consensus service can generate a new private/public keypair that can be used preconfirmations. The public key will need to be signed by the KMS still, and then gossiped to the network. 

This needs to be added.

PR #2756, opened 2025-02-24
Add new Preconfirmation Task

MitchTurner:
## Linked Issues/PRs

Part of https://github.com/FuelLabs/fuel-core/issues/2739

## Description

Introduces new `PreConfirmationSignatureTask` which will be responsible for receiving Vecs of txs from the producer, signing the txs, and broadcasting them to the network. The signatures will be with delegate keys that will be rotated on a regular frequency.

This does not implement the interfaces for performing the actions, rather defines the tests for the domain behavior of the service.

## Checklist

### Before requesting review


crates/services/consensus_module/poa/src/pre_confirmation_service/broadcast.rs
```diff
@@ -0,0 +1,18 @@
+use super::*;
+
+#[async_trait::async_trait]
```
rymnc, line 3: i don't think we need this :)

crates/services/consensus_module/poa/src/pre_confirmation_service/key_generator.rs
```diff
@@ -0,0 +1,8 @@
+use super::*;
+use crate::pre_confirmation_service::signing_key::SigningKey;
+
+#[async_trait::async_trait]
```
rymnc, line 4: i don't think we need this, same as prev

crates/services/consensus_module/poa/src/pre_confirmation_service/parent_signature.rs
```diff
@@ -0,0 +1,9 @@
+use super::*;
+
+#[async_trait::async_trait]
```
rymnc, line 3: same as prev

crates/services/consensus_module/poa/src/pre_confirmation_service/tests.rs
```diff
@@ -0,0 +1,400 @@
...
+    Success { height: BlockHeight },
+    Fail { height: BlockHeight },
+    SqueezedOut { reason: String },
+}
+
+struct FakeTxReceiver {
+    recv: tokio::sync::mpsc::Receiver<Vec<(Transaction, Status)>>,
+}
+
+#[async_trait::async_trait]
```
rymnc, line 27: same as prev

crates/services/consensus_module/poa/src/pre_confirmation_service/tests.rs
```diff
@@ -0,0 +1,400 @@
...
+        }
+    }
+}
+
+struct FakeBroadcast {
+    delegation_key_sender: tokio::sync::mpsc::Sender<FakeSignedData<FakeSigningKey>>,
+    tx_sender: tokio::sync::mpsc::Sender<FakeSignedData<Vec<(Transaction, Status)>>>,
+}
+
+#[async_trait::async_trait]
```
rymnc, line 45: same as prev

crates/services/consensus_module/poa/src/pre_confirmation_service/tests.rs
```diff
@@ -0,0 +1,400 @@
...
+    dummy_signature: String,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+struct FakeSignedData<T> {
+    data: T,
+    dummy_signature: String,
+}
+
+#[async_trait::async_trait]
```
rymnc, line 86: same as prev

rymnc (commented):
first pass, i think we can remove usages of async trait for new traits

MitchTurner:
@rymnc 

I've always used the `#[async_trait::async_trait]`. I know you have started removing it, but I haven't really heard the argument for why we are moving away from it. Something to do with performance.

rymnc:
> @rymnc 
> 
> I've always used the `#[async_trait::async_trait]`. I know you have started removing it, but I haven't really heard the argument for why we are moving away from it. Something to do with performance. 

there's nearly 40-50% gains if you don't use it, so if you don't need dyn-compatibility it is best to not use it

netrome:
> @rymnc
> 
> I've always used the `#[async_trait::async_trait]`. I know you have started removing it, but I haven't really heard the argument for why we are moving away from it. Something to do with performance.

Another argument is simplicity. The macro hides a lot of `Box<Pins>` that are both inefficient and make it more complicated for downstream users to implement the traits. Imo the `async_trait` crate is a hack that served us well before async traits were stabilized, but nowadays I don't think there is any use for it. It deserves a toast from the community and a calm retirement.

AurelienFT (commented):
Can you comment all the trait/functions you added to understand what is their role in the flow we designed ?

From what I understand there nothing to tell the block the time window is finish or anything like that. It just seems to be a "middle man" that intercept tx, sign them (and manage key management) and re-broadcast them. If that's really the case can we maybe renaming it to "pre-confirmation-signing" or something like that. Because at first I was thinking it will manage the block separation etc.

Side note : I'm not a big fan of the file `tests.rs` in the same folder as the code as it write the changes on GitHub interface in the middle of the application code but that's a personal opinion, not blocking at all.

AurelienFT (approved):

crates/services/consensus_module/poa/src/lib.rs
```diff
@@ -1,6 +1,5 @@
 #![deny(clippy::arithmetic_side_effects)]
 #![deny(clippy::cast_possible_truncation)]
-#![deny(unused_crate_dependencies)]
```
rymnc, line 3: we probably want to keep this rule?

crates/services/consensus_module/poa/src/pre_confirmation_signature_service/tests.rs
```diff
@@ -0,0 +1,395 @@
...
+}
+
+impl Broadcast for FakeBroadcast {
+    type PreConfirmations = Vec<(Transaction, Status)>;
+    type ParentSignature<T> = FakeSignedData<T>;
+    type DelegateKey = FakeSigningKey;
+
+    async fn broadcast_txs(
+        &mut self,
+        txs: <Self::DelegateKey as SigningKey>::Signature<Self::PreConfirmations>,
```
rymnc, line 51: you could make this type a little simpler

```
type Signed<K, T> = <K as SigningKey>::Signature<T>;
```

and then

```suggestion
        txs: Signed<Self::DelegateKey, Self::Preconfirmations>
```

rymnc (dismissed):
non blocking minor nits, but generally lgtm

AurelienFT (approved):

xgreenx (approved):

Issue #2757, opened 2025-02-24
Add `AssembleTx` GraphQL End-Point


Torres-ssf:
# Overview

Quick before/after of what we want to achieve.

![Image](https://github.com/user-attachments/assets/bdf4b079-78d1-4524-ac62-d72a94bd2fad)

# Summary

We need a new GraphQL end-point to **assemble, estimate, and fund** a transaction.

This end-point may receive a transaction in a not-ready state and can be missing elements such as:

- Resources
- Max fee value
- Script gas limit
- Predicate gas usage
- Variable outputs (Sway projects handling token transfers)
- Contract inputs/outputs (scripts that manually call contracts or inter-contract calls)

Also, this method should return the required funds necessary to pay for the transaction.

# Request Parameters

- **tx**: The hex of the transaction to be assembled.
- **blockHorizon**: The number of blocks into the future for estimating gas price.
- **forwards**: The asset amounts being forwarded to contract/script calls.
- **owner**: Owner from which we want from resources to be fetched from
- **excludedIds**: Resources IDs to be excluded when funding the transaction. Same thing that already happens with coinsToSpend

```ts
type AssetForward {
  amount: U64!
  assetId: AssetId!
  owner: Owner
}

type Owner {
  address: Address!
  predicateBytecode?: Bytes
  predicateData?: Bytes
}

type AssembleTxInput {
  tx: HexString!
  blockHorizon: U32!
  forwards?: [AssetForward!]
  owner: Owner!
  excludedIds?: ExcludeInput
}
```

# Response Format

- **`tx`**: The fully assembled, estimated, and funded transaction.
- **`status`**: The status from the dry-run execution.

```ts
type AssembleTxOutput {
  tx: Transaction!
  status: DryRunTransactionStatus!
}
```

# Requirements

> [!IMPORTANT]
> The `assembleTx` must **not** fail if `gasLimit` or fee values are missing or incorrect.

1. **Estimate Predicates**

   - If the owner includes predicateBytecode this means it's predicate and we should estimate them and set the `predicate_gas_used`.
   - If predicate inputs are provided on the TX and the `predicate_gas_used` is **zero**, we should also estimate the `predicate_gas_used` for them.
   - During estimation we should not validate predicate root.

1. **Add Missing `OutputVariables`**

   - Identify and add any missing `OutputVariables` required for the transaction’s execution.
   - Limit attempts to `10` or until sum of `OutputVariables` length and other outputs are greater than 254.

[...]

PR #2769, opened 2025-02-27
feat: `AssembleTx` query

netrome:
Closes https://github.com/FuelLabs/fuel-core/issues/2757

Added a new `assembleTx` GraphQL endpoint.
The endpoint can be used to assemble the transaction based on the provided requirements.

The returned transaction contains:
- Input coins to cover `required_balances`
- Input coins to cover the fee of the transaction based on the gas price from `block_horizon`
- `Change` or `Destroy` outputs for all assets from the inputs
- `Variable` outputs in the case they are required during the execution
- `Contract` inputs and outputs in the case they are required during the execution
- Reserved witness slots for signed coins filled with `64` zeroes
- Set script gas limit(unless `script` is empty)
- Estimated predicates, if `estimate_predicates == true`

Returns an error if:
- The number of required balances exceeds the maximum number of inputs allowed.
- The fee address index is out of bounds.
- The same asset has multiple change policies(either the receiver of
    the change is different, or one of the policies states about the destruction
    of the token while the other does not). The `Change` output from the transaction
    also count as a `ChangePolicy`.
- The number of excluded coin IDs exceeds the maximum number of inputs allowed.
- Required assets have multiple entries.
- If accounts don't have sufficient amounts to cover the transaction requirements in assets.
- If a constructed transaction breaks the rules defined by consensus parameters.

Remaining work do be done:
- [x] Client implementation.
- [ ] Integration tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Fuel client implementation.
 | Network  | Version |
 |----------|---------|
 | Fuel Ignition | 0.40.4 |
-| Testnet | 0.41.7 |
+| Testnet | 0.41.8 |
 | Devnet | 0.41.8 |
 
 ## Contributing


crates/fuel-core/src/coins_query.rs
```diff
@@ -48,7 +42,7 @@ pub enum CoinsQueryError {
     #[error("the target cannot be met due to no coins available or exceeding the {max} coin limit.")]
     InsufficientCoinsForTheMax {
         asset_id: AssetId,
-        collected_amount: Word,
+        collected_amount: u128,
```
netrome, line 45: @xgreenx - why do we use u128 here? It breaks the client currently so I'll revert the change for now and we can re-introduce it later.

crates/fuel-core/src/schema/tx/assemble_tx.rs
```diff
@@ -0,0 +1,879 @@
...
+            script.set_max_fee_limit(0);
+        }
+
+        let gas_costs = self.arguments.consensus_parameters.gas_costs();
+        let fee_params = self.arguments.consensus_parameters.fee_params();
+
+        let mut status;
+        let mut gas_used_by_tx;
+
+        loop {
```
AurelienFT, line 628: Can we document what's the goal of this loop ? it's unclear to me at first look why we can't dry_run once.

crates/fuel-core/src/schema/tx/assemble_tx.rs
```diff
@@ -0,0 +1,879 @@
...
+                        "The total base asset amount \
+                        became too big when tried to cover fee"
+                    ),
+                )?;
+                self.add_input_and_witness_and_change(&fee_payer_account, coin);
+            }
+
+            // In the case when predicates iterates over the inputs,
+            // it increases its used gas. So we need to re-estimate predicates.
+            self = self.estimate_predicates().await?;
```
AurelienFT, line 802: We can't do it only once at the end ? It seems very heavy operation it scares me.

AurelienFT (commented):
Tests are getting crunched !!! Good ! Overall the PR is ok to me. Great work !!! I have few questions but will be able to approve soon.

I'm waiting to see if it's ok for SDK integrations

crates/client/assets/schema.sdl
```diff
@@ -128,6 +138,18 @@ type ChangeOutput {
...
+input ChangePolicy @oneOf {
+	"""
+	Adds `Output::Change` to the transaction if it is not already present.
+	Sending remaining assets to the provided address.
+	"""
+	change: Address
+	"""
+	Destroys the remaining assets by the transaction for provided address.
+	"""
+	destroy: Boolean
```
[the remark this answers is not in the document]
netrome: Agree, updated in https://github.com/FuelLabs/fuel-core/pull/2769/commits/9eddbf95282c49b1f3773424b1e878479d271957

crates/client/assets/schema.sdl
```diff
@@ -1359,6 +1448,7 @@ type VariableOutput {
 }
 
 directive @include(if: Boolean!) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
+directive @oneOf on INPUT_OBJECT
```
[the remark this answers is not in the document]
netrome: My interpretation is that we have to explicitly create this directive because the the `oneOf` directive is not yet officially supported. So client's that doesn't support it will se this declaration and go "ah - there's a noop directive here that doesn't do anything" - instead of returning some error due to an unrecognized directive.

Reference https://async-graphql.github.io/async-graphql/en/define_one_of_object.html

crates/fuel-core/src/schema/tx/assemble_tx.rs
```diff
@@ -0,0 +1,879 @@
...
+            let asset_id = required_balance.asset_id;
+
+            if asset_id == base_asset_id
+                && fee_payer_account.owner() == required_balance.account.owner()
+            {
+                base_asset_reserved = Some(required_balance.amount);
+            }
+        }
+
+        // If the user didn't provide the base asset, we add it to the required balances
```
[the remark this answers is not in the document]
netrome: Good point. Though I suppose getting to `max_inputs` should be quite an unusual case so I'm not sure if it's worth optimizing it right now.

crates/fuel-core/src/schema/tx/assemble_tx.rs
```diff
@@ -0,0 +1,879 @@
...
+            let asset_id = required_balance.asset_id;
+            let amount = required_balance.amount;
+            let owner = required_balance.account.owner();
+
+            let selected_coins = self
+                .arguments
+                .coins(owner, asset_id, amount, remaining_input_slots)
+                .await?;
+
+            for coin in selected_coins {
```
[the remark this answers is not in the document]
netrome: Hmm could be worth it. I'm not sure. Perhaps worth hearing from @xgreenx on this one before adding it.

crates/fuel-core/src/schema/tx/assemble_tx.rs
netrome: We also have a dangling todo to use the continue on error feature, so I'll use that as well.

crates/fuel-core/src/schema/tx/assemble_tx.rs
netrome: Hmm yeah it seems to me that we don't need the predicate estimations within this loop so we might as well do the re-estimation afterwards. Though good to verify with @xgreenx if we're missing something here.

crates/fuel-core/src/schema/tx/assemble_tx.rs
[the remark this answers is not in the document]
AurelienFT: IMO, when we are dealing with user-controled interface each cases can become usual depending on the users input motivated by his goals.

crates/fuel-core/src/schema/tx.rs
```diff
@@ -241,6 +259,161 @@ impl TxQuery {
...
+        }
+
+        let fee_index: u16 = fee_address_index.into();
+        let estimate_predicates: bool = estimate_predicates.unwrap_or(false);
+        let reserve_gas: u64 = reserve_gas.map(Into::into).unwrap_or(0);
+
+        let excluded_id_count = exclude_input.as_ref().map_or(0, |exclude| {
+            exclude.utxos.len().saturating_add(exclude.messages.len())
+        });
+        if excluded_id_count > max_input as usize {
```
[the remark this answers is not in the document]
AurelienFT: I think a limit is useful here to avoid having to skip to much coins in coins_to_spend but I don't see why the current limit used is `max_input`

crates/fuel-core/src/schema/tx/assemble_tx.rs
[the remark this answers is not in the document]
xgreenx: We can add it

crates/fuel-core/src/schema/tx/assemble_tx.rs
xgreenx: The issues that each time when we add a new coin, it is possible that it will change the gas used by predicate, causing us to find more coins, that can change predicates again, etc.

xgreenx: Added protection to not do more than 10 estimation of predicates

AurelienFT (approved):

Torres-ssf:
Can we make a small change in this PR?

In `schema.sdl`, change the `InputCoin.witnessIndex` type from `Int!` to `U16!`:
https://github.com/FuelLabs/fuel-core/blob/71d5b1ea678f4df1b1cc1985284be26758015faf/crates/client/assets/schema.sdl#L637

The idea is to align with `InputMessage.witnessIndex`:
https://github.com/FuelLabs/fuel-core/blob/71d5b1ea678f4df1b1cc1985284be26758015faf/crates/client/assets/schema.sdl#L656

Only asking this because our GraphQL lib is nit picking when querying `assembleTx`, noting that `InputCoin` and `InputMessage` have a property with the same name but different types.

While I could work around this with aliases, it seems more consistent for this property to share the same type across both `InputCoin` and `InputMessage`.

xgreenx (dismissed):

xgreenx (approved):

AurelienFT (approved):

xgreenx (dismissed):

xgreenx (approved):

crates/fuel-core/src/schema/tx.rs
```diff
@@ -242,6 +260,162 @@ impl TxQuery {
...
+                that points to the address who pays fee for the transaction. \
+                If you only want to cover the fee of transaction, you can set the required balance \
+                to 0, set base asset and point to this required address.")]
+        fee_address_index: U16,
+        #[graphql(
+            desc = "The list of resources to exclude from the selection for the inputs"
+        )]
+        exclude_input: Option<ExcludeInput>,
+        #[graphql(
+            desc = "Perform the estimation of the predicates before cover fee of the transaction"
```
rymnc, line 316:
```suggestion
            desc = "Perform the estimation of the predicates before covering fee of the transaction"
```

tests/tests/assemble_tx.rs
```diff
@@ -0,0 +1,292 @@
...
+    let client = FuelClient::from(service.bound_address);
+    let CoinType::Coin(coin) = client
+        .coins_to_spend(&owner, vec![(non_base_asset_id, 100, None)], None)
+        .await
+        .unwrap()[0][0]
+    else {
+        panic!("Expected a coin");
+    };
+
+    // Given
```
rymnc, line 262: nit: multiple `Given`'s here

rymnc (dismissed):
nice, i like how its easier to send tx transactions now :)

AurelienFT (approved):

xgreenx (approved):

Issue #2720, opened 2025-02-14
Make gas price service optional for node operators


xgreenx:
In https://github.com/FuelLabs/fuel-vm/issues/888 we add the max percent on which gas price can change from block to block on consensus level. It allows all nodes in the network have the same rule regarding estimation of the worst gas price.

The regular nodes in the network don't need to run an algorithm to determine the next gas price; they can use the percent to calculate it based on the previous gas price.

We need to make gas price service and all related fields optional for the node operator to use. Mostly, only the block producer should use them.

PR #2770, opened 2025-02-27
 fix: make gas price service optional for node operators

gap-editor:
I have implemented all the necessary changes to resolve issue #2720:

1. Added the `enabled` field to the `GasPriceConfig` structure to allow disabling the gas price service
2. Modified the `estimate_gas_price` method in the GraphQL schema to return an error if the service is disabled
3. Modified the gas price service initialization to take into account the `enabled` option:
   - If `enabled = true`, uses the full service with gas price estimation algorithm
   - If `enabled = false`, uses a static provider that always returns the minimum gas price

Now users can disable the gas price service by setting `enabled = false` in the configuration. As a result:
- The GraphQL `estimate_gas_price` endpoint will return a clear error message
- The system will use the minimum gas price for transactions
- The gas price estimation service will not be created and started


xgreenx:
Closing for now because of now activity, feel free to reopen if you decide to continue work on it=)

PR #2780, opened 2025-02-27
add implementations for pre conf task

MitchTurner:
## Linked Issues/PRs

part of https://github.com/FuelLabs/fuel-core/issues/2739

## Description

Introduced some working and some dummy implementations. 

We still need to implement the P2P broadcast trait and the delegate key generator stuff. We don't know what kind of keys we are going to use, so that needs to be decided on. 

## Checklist

### Before requesting review
https://github.com/FuelLabs/fuel-core/issues/2782
https://github.com/FuelLabs/fuel-core/issues/2783


crates/fuel-core/src/service/adapters/consensus_module/poa/pre_confirmation_signature/trigger.rs
```diff
@@ -0,0 +1,135 @@
...
+    Duration::from_secs(diff)
+}
+
+impl<Time: GetTime + Send> KeyRotationTrigger for TimeBasedTrigger<Time> {
+    async fn next_rotation(&mut self) -> PoAResult<()> {
+        // create a future that will resolve at the next rotation time
+        // if that future resolves, update the next rotation time
+        let next_rotation = self.get_next_rotation()?;
+        let time_to_wait = duration_between(self.time.now(), next_rotation);
+        tokio::time::sleep(time_to_wait).await;
```
AurelienFT, line 50: If we do that, does it means that the service `pre_confirmation_signature` is stuck before we reach the next signing `Tai64`, which prevents us from signing pre-confirmation or it's fine ? Because I believe both are on the same `tokio::task` and so we would be stuck ?

xgreenx: I think we should just use timer and `tick().await` here=)

xgreenx (dismissed):

AurelienFT (dismissed):

AurelienFT (approved):

xgreenx (approved):

xgreenx (approved):

PR #2785, opened 2025-03-01
Rework execution flow to allow preconfirmations

AurelienFT:
### TODO:
- Change the timing of block production inside the PoA to allow direct block production with a timeout
- Pass chanel of preconfirmations to the executor
- Move the executor trait in correct modules to correctly use ports and adapters

### DONE:
- Makes the produce function async and add a trigger as parameter.

## Questions

- Should we add the trigger inside the production mode or within the components ? (within components I tried but because they are serde serializable I had difficulties)
- About the new channel should we also add it as parameter like that or within the components or production mode also ? (within components I tried but because they are serde serializable I had difficulties). Or maybe we can directly place it in the block executor and it will avoid having to pass it from `process_l2_txs` down to `execute_chargeable_transaction`.
- What is the type for the channel for the pre-confirmations in the executors. I don't see any type with the `Receipts` etc...

## Checklist

### Before requesting review


AurelienFT:
Close it in favor of #2799

Issue #2775, opened 2025-02-27
fault_proving(fault_proving_bin): define the compression service worker


rymnc:
We want to remove the da compression from the OCW and instead define it separately. it should -

- hook into the block importer
- perform da compression, store into database
- perform global state root calculation

PR #2788, opened 2025-03-03
fault_proving(compression_service): initial scaffold for dedicated compression service

rymnc:
## Linked Issues/PRs

- part of https://github.com/FuelLabs/fuel-core/issues/2775

## Description

Given that we want to modularize the compression service from the offchain worker, this PR begins the refactor.

Things to keep in mind while reviewing -

- ports/errors/config/storage traits will evolve with this integration
- we will remove temporary compiler/clippy rules once integration is done
- tests will come as well 
- codeowners file has new entry for this service to help expedite reviews

This pull request introduces a new dedicated compression service for fuel blocks. The key changes include scaffolding the service, defining its configuration, errors, and ports, and implementing the service logic. Below are the most important changes:

### New Service Implementation:

* **Scaffold Compression Service**: Added a new compression service responsible for compressing fuel blocks for Data Availability (DA). (`.changes/added/2788.md`)
* **Service Definition**: Implemented the `CompressionService` struct and its associated methods to handle new blocks, perform compression, and manage service state. (`crates/services/compression/src/service.rs`)

### Configuration and Dependencies:

* **Cargo.toml Updates**: Added the new compression service to the workspace and defined its dependencies. (`Cargo.toml`, `crates/services/compression/Cargo.toml`) [[1]](diffhunk://#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R21) [[2]](diffhunk://#diff-3b348e1ded89f349afd45656ad5c57ebbab5e690f2bf64e059dd5d7c32de228dR1-R19)
* **Configuration**: Defined the `CompressionConfig` struct for service configuration. (`crates/services/compression/src/config.rs`)

### Error Handling:

* **Error Definitions**: Created the `CompressionError` enum to handle various errors that may occur during the compression process. (`crates/services/compression/src/errors.rs`)

### Ports and Interfaces:

* **Ports for Service Interaction**: Defined ports for block source, compression storage, and configuration to interact with the service. (`crates/services/compression/src/ports.rs`, `crates/services/compression/src/ports/block_source.rs`, `crates/services/compression/src/ports/compression_storage.rs`, `crates/services/compression/src/ports/configuration.rs`) [[1]](diffhunk://#diff-749e0863f7134e84fbbc12941033fa2eee85105be296fba5853083cded778a6eR1-R8) [[2]](diffhunk://#diff-b35ece6e136e2fe3793da56f4456d50692646e9d371946e26024ada643151182R1-R11) [[3]](diffhunk://#diff-766e163ef7ab6dd00944205be543374cd60cefb2fb5605467d2365ca71938cbbR1-R10) [[4]](diffhunk://#diff-931663b130b61eb67d49e463c0c310c308e6393817ea57317e63ebb543a7a0a5R1-R7)

### Code Ownership:

* **CODEOWNERS Update**: Assigned the new compression service to the appropriate code owners. (`.github/CODEOWNERS`)

## Checklist

### Before requesting review

- [ ] Someone else?


xgreenx (dismissed):

crates/services/compression/src/errors.rs
```diff
@@ -0,0 +1,21 @@
+use thiserror::Error;
```
netrome, line 1: Nit: I believe thiserror doesn't work with no-std. Don't think we need it for now, but if we were to pivot to run compression in a zkvm in the future it's a bit more convenient if we derive `derive_more::Error` instead.

crates/services/compression/src/errors.rs
```diff
@@ -0,0 +1,21 @@
...
+    #[error("failed to write compressed block to storage: `{0}`")]
+    FailedToWriteCompressedBlock(String),
+    // Service errors
+    /// Failed to create service
+    #[error("failed to create service: `{0}`")]
+    FailedToCreateService(String),
+    /// Failed to handle new block
+    #[error("failed to handle new block: `{0}`")]
+    FailedToHandleNewBlock(String),
+}
```
netrome, line 21: I don't see these used anywhere. Are the inner `String`s just placeholders for later dedicated types? Nit: I'd default to `&'static str` instead of `String`.

crates/services/compression/src/ports/block_source.rs
```diff
@@ -0,0 +1,11 @@
...
+pub type Block = fuel_core_types::blockchain::block::Block;
+
+/// Port for L2 blocks source
+/// Each time `.next()` is called, the service expects the same behaviour as an iterator.
+pub trait BlockSource {
+    /// Get the next block from the source
+    fn next(
+        &self,
+    ) -> impl core::future::Future<Output = crate::Result<Option<Block>>> + Send + Sync;
+}
```
netrome, line 11: Nit: I'd rename the method to `next_block` for clarity. But I think `next` probably also works if the variable in the calling code is properly named.
```suggestion
pub type Block = fuel_core_types::blockchain::block::Block;

/// Port for L2 blocks source
/// Each time `.next()` is called, the service expects the same behaviour as an iterator.
pub trait BlockSource {
    /// Get the next block from the source
    fn next_block(
        &self,
    ) -> impl core::future::Future<Output = crate::Result<Option<Block>>> + Send + Sync;
}
```

crates/services/compression/src/service.rs
```diff
@@ -0,0 +1,191 @@
...
+    config: C,
+}
+
+impl<B, S, C> CompressionService<B, S, C>
+where
+    B: BlockSource,
+    S: CompressionStorage,
+    C: CompressionConfigProvider,
+{
+    fn _new(block_source: B, storage: S, config: C) -> Self {
```
netrome, line 34: I assume this should be public, in which case I don't think we'll get the unused warning.
```suggestion
    pub fn new(block_source: B, storage: S, config: C) -> Self {
```

netrome (approved):
Nice stuff. Just some nits and suggestions but nothing major from me.

crates/services/compression/src/errors.rs
rymnc: we will never mount the service inside a zkvm. there's no need to make this no-std compatible yet

crates/services/compression/src/errors.rs
rymnc: placeholder for now.

crates/services/compression/src/ports/block_source.rs
rymnc: block_source.next() makes more sense to me than block_source.next_block(). also given that we allude to an iterator format here I think it is reasonable to assume the stream is of blocks.

crates/services/compression/src/service.rs
rymnc: this is not meant to be public. we're wrapping it in new_service :)

rymnc: fair though

crates/services/compression/src/errors.rs
rymnc: even the `String`'s are just placeholders till we get some concrete types in :)

netrome: Makes sense :+1:

netrome: Also I like the placeholders. It's nice to see what things we anticipate.

crates/services/compression/src/ports/block_source.rs
netrome: Fair enough. I agree that `block_source.next()` sounds best. However I'd rather see `block_source.next_block()` rather than `source.next()`, which feels like it can pop up in client code. But that's just a matter of taste, and I like the conciseness of just calling it `.next()`. Perhaps that will encourage better variable names as well.

crates/services/compression/src/ports/configuration.rs
```diff
@@ -0,0 +1,7 @@
+use crate::config::CompressionConfig;
+
+/// Configuration for the compression service
+pub trait CompressionConfigProvider {
+    /// getter for the compression config
+    fn config(&self) -> CompressionConfig;
```
AurelienFT, line 6: From my perspective of ports and adapter, on each module we define ports for interactions that we need in order to make this module works, we don't define ports for exposing this modules functionnalities, this is done when implementing the adapters through the port traits of modules that actually wants it.

rymnc: sure, but we decouple the configuration here ~ in `fuel-core`, we can just implement
`impl CompressionConfigProvider for Config {}`

tbh, we could just use an `Into<CompressionConfig>` trait bound, but this seems cleaner. we do the same thing for gas price service i thinkso

rymnc: the former :)

AurelienFT: Okey then i'm fine.

AurelienFT (approved):

xgreenx (approved):

PR #2793, opened 2025-03-04
fault_proving(storage_traits): refactor to allow usage of arbitrary `TableColumn`

rymnc:
## Linked Issues/PRs

- part of https://github.com/FuelLabs/fuel-core/issues/2775

## Description

- instead of tying `MerkleizedTableColumn` to the `TableColumn` variant only in use by the global state root service, this refactors the trait to accept an arbitrary `TableColumn` that is returned when `table_column` is called.

we've copied all the common functionality required by both destinations (`fuel-core-global-merkle-root-storage` and `fuel-core-compression-service`), you may see that we haven't deleted the implementations from the `fuel-core-global-merkle-root-storage` crate yet, because that crate doesn't use newtypes and we have an orphan rule error. will be addressed in follow up. For the person implementing the follow up, you can remove -

- `fuel-core-global-merkle-root/src/merkle.rs`
- `fuel-core-global-merkle-root/src/merkle/smt.rs`

and reuse the traits & types from `fuel-core-storage` instead

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/storage/src/merkle/column.rs
+++ b/crates/storage/src/merkle/column.rs
@@ -1,0 +1,73 @@
+//! This module defines a `MerklizedColumn` type which defines data, merkledata and merkle metadata for it
+use crate::kv_store::StorageColumn;
+use alloc::{
+    format,
+    string::{
+        String,
+        ToString,
+    },
+};
+
+/// Almost in the case of all tables we need to prove exclusion of entries,
+/// in the case of malicious block.
+#[repr(u32)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub enum MerkleizedColumn<TC> {
+    /// The column with the specific data.
+    TableColumn(TC),
+    /// The column with the merkle tree nodes.
+    MerkleDataColumn(TC),
+    /// The merkle metadata column.
+    MerkleMetadataColumn,
+}
+
+/// The trait to convert the column to the `u32`.
+pub trait AsU32 {
+    /// Returns the `u32` representation of the `Column`.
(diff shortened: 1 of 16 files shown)


netrome:
Makes a lot of sense. I've skimmed the PR and I think it looks good and I agree with the idea.

One thing though: The generic `Column` type should probably be moved to the `fuel-core-storage` crate if you want to reuse it for the compression - since I don't think the compression crate should be aware of any of the state root service crates.

netrome (approved):
Looks good to me!

crates/storage/src/merkle/column.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    /// The start of the merkle data columns.
+    pub const MERKLE_DATA_COLUMNS_START: u32 = u16::MAX as u32;
+
+    /// Returns the `u32` representation of the `Column`.
+    pub fn as_u32(&self) -> u32 {
+        match self {
+            Self::TableColumn(column) => column.as_u32(),
+            Self::MerkleDataColumn(column) => {
+                Self::MERKLE_DATA_COLUMNS_START.wrapping_add(column.as_u32())
```
rafal-ch, line 47: Not for this PR, but this can cause unexpected behavior if `TC::COUNT` is big enough, which it will probably never be.... Just leaving a comment here.

rymnc: I guess if our column count gets so large that it overflows u32::MAX, we would have many other problems too haha

Dentosal (approved):

PR #2799, opened 2025-03-04
Rework execution to allow preconfirmations

AurelienFT:
### DONE:
- Move the executor trait in correct modules to correctly use ports and adapters
- Makes the produce function async and add a trigger as parameter.
- Pass channel of preconfirmations to the executor
- Added a new production mode for PoA service. The open production mode starts the block production immediately and finishes after the `period` is passed.

## Checklist

### Before requesting review


crates/fuel-core/src/executor.rs
```diff
@@ -467,15 +472,21 @@ mod tests {
...
+                            invalid_duplicate_tx,
+                        ]),
+                        gas_price: price,
+                        coinbase_recipient: recipient,
+                    },
+                    TimeoutOnlyTxWaiter,
+                    TransparentPreconfirmationSender,
+                )
+                .now_or_never()
+                .unwrap()
```
xgreenx, line 489: It is a lot of places where we need to use `now_or_never`. It would be better if we added a new helper sync function that we can reuse in these tests

xgreenx (commented):
Looks good to me=)

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -404,7 +442,15 @@ where
...
-            )?
+                TimeoutOnlyTxWaiter,
+                TransparentPreconfirmationSender,
+            )
+            .now_or_never()
+            .ok_or_else(|| {
+                ExecutorError::Other(
+                    "Impossible to resolve the executor's future immediately".to_string(),
+                )
+            })??
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/services/upgradable-executor/wasm-executor/src/main.rs
```diff
@@ -100,15 +105,21 @@ fn execute_dry_run(
     instance: ExecutionInstance<WasmRelayer, WasmStorage>,
     block: Components<WasmTxSource>,
 ) -> ReturnType {
-    let result = instance.produce_without_commit(block, true);
+    let result = instance
+        .produce_without_commit(block, true, NewTxWaiter, PreconfirmationSender)
+        .now_or_never()
+        .unwrap();
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/types/src/services/executor.rs
```diff
@@ -89,6 +89,14 @@ pub struct ValidationResult {
     pub events: Vec<Event>,
 }
 
+/// The result of waiting for new transactions.
+pub enum WaitNewTransactionsResult {
+    /// We received a new transaction and we can continue the block execution.
+    NewTransaction,
+    /// We didn't receive any new transaction and we can end the block execution.
+    Timeout,
+}
```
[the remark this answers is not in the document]
AurelienFT: Done

crates/services/executor/src/executor.rs
```diff
@@ -644,11 +710,32 @@ where
                     *coinbase_contract_id,
                     memory,
                 ) {
-                    Ok(_) => {}
+                    Ok(_) => {
+                        // SAFETY: Transaction succeed so we know that `tx_status` has been populated
+                        let latest_executed_tx = data.tx_status.last().unwrap();
```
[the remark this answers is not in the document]
AurelienFT: Done.

crates/fuel-core/src/executor.rs
AurelienFT: Done.

crates/services/executor/src/executor.rs
```diff
@@ -644,11 +710,32 @@ where
...
+                                PreconfirmationStatus::SuccessByBlockProducer {
+                                    block_height: *block.header.height(),
+                                }
+                            }
+                            TransactionExecutionResult::Failed { .. } => {
+                                PreconfirmationStatus::FailureByBlockProducer {
+                                    block_height: *block.header.height(),
+                                }
+                            }
+                        };
```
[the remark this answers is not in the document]
AurelienFT: Done in executor because not used elsewhere.

xgreenx (dismissed):

xgreenx (dismissed):

crates/fuel-core/src/service/adapters.rs
```diff
@@ -318,23 +320,49 @@ impl TransactionsSource {
...
 impl ExecutorAdapter {
     pub fn new(
         database: Database,
         relayer_database: Database<Relayer>,
-        // #[cfg(feature = "parallel-executor")]
-        // config: fuel_core_parallel_executor::config::Config,
-        // #[cfg(not(feature = "parallel-executor"))]
         config: fuel_core_upgradable_executor::config::Config,
+        new_txs_watcher: tokio::sync::watch::Receiver<()>,
+        preconfirmation_sender: tokio::sync::mpsc::Sender<Vec<PreconfirmationStatus>>,
```
MitchTurner, line 358: Should this be depending on the ports instead of these concrete types?

I'll keep reading to understand better this flow.

It feels weird to have a "port" for something we construct concretely (`PreconfirmationSender`), maybe it's not a port but just an abstraction?

crates/fuel-core/src/service/adapters/executor.rs
```diff
@@ -65,3 +81,43 @@ impl fuel_core_executor::ports::RelayerPort for RelayerIterableKeyValueView {
...
+    }
+
+    fn try_send(
+        &self,
+        preconfirmations: Vec<PreconfirmationStatus>,
+    ) -> Vec<PreconfirmationStatus> {
+        match self.sender.try_send(preconfirmations) {
+            Ok(()) => vec![],
+            Err(TrySendError::Closed(_)) => vec![],
+            Err(TrySendError::Full(preconfirmations)) => preconfirmations,
```
MitchTurner, line 120: Do we want to notify someone that the channel is closed instead of acting like all of them were sent?

crates/fuel-core/src/service/adapters.rs
MitchTurner:
```rs
pub struct BlockExecutor<R, TxWaiter, PreconfirmationSender> {
    relayer: R,
    consensus_params: ConsensusParameters,
    options: ExecutionOptions,
    new_tx_waiter: TxWaiter,
    preconfirmation_sender: PreconfirmationSender,
}
```

I think we need better names for these things.

But it looks like maybe we are just integrating things here.

crates/services/executor/src/executor.rs
```diff
@@ -440,13 +512,23 @@ where
...
+                &mut block_storage_tx,
+                &mut data,
+                &mut memory,
+            )
+            .await?;
+            match self.new_tx_waiter.wait_for_new_transactions().await {
+                WaitNewTransactionsResult::Timeout => break,
+                WaitNewTransactionsResult::NewTransaction => {
+                    continue;
+                }
```
MitchTurner, line 529: Do we want some kind of safety timeout here that will also break the loop?

crates/services/executor/src/executor.rs
```diff
@@ -644,11 +728,27 @@ where
...
+                        let preconfirmation_status =
+                            convert_tx_execution_result_to_preconfirmation(
+                                &latest_executed_tx.result,
+                                *block.header.height(),
+                                data.tx_count,
+                            );
+                        status.push(preconfirmation_status);
+                    }
                     Err(err) => {
+                        status.push(PreconfirmationStatus::SqueezedOutByBlockProducer {
```
MitchTurner, line 745: Should this be:
```rs
    /// Transaction will not be included in a block, rejected at `block_height`
    FailureByBlockProducer {
        /// The block height at which the transaction will be rejected
        block_height: BlockHeight,
    },
 ```
 instead of squeezed out? I thought squeezing out happens in the txpool.

File: crates/services/upgradable-executor/wasm-executor/src/preconfirmation_sender.rs [hunk 11]
```diff
@@ -0,0 +1,14 @@
+use fuel_core_executor::ports::PreconfirmationSenderPort;
+use fuel_core_types::services::preconfirmation::PreconfirmationStatus;
+
+pub struct PreconfirmationSender;
```
MitchTurner, line 4: What's the difference between this and `TransparentPreconfirmationSender`

crates/services/upgradable-executor/wasm-executor/src/new_tx_waiter.rs
```diff
@@ -0,0 +1,12 @@
+use fuel_core_executor::{
+    executor::WaitNewTransactionsResult,
+    ports::NewTxWaiterPort,
+};
+
+pub struct NewTxWaiter;
+
+impl NewTxWaiterPort for NewTxWaiter {
```
MitchTurner, line 8: What's the difference between  this and `TimeoutOnlyTxWaiter`?

I don't think I mind the redundancy, just figured I'd check that it was intentional.

crates/services/executor/src/executor.rs
MitchTurner, line 529: Can we add some executor tests to show that this works?

crates/services/executor/src/executor.rs
```diff
@@ -665,6 +765,11 @@ where
                 .take(remaining_tx_count as usize)
                 .peekable();
         }
+        if !status.is_empty() {
+            if let Err(e) = self.preconfirmation_sender.send(status).await {
```
MitchTurner, line 769: Can we add some executor tests to show that this indeed sends the expected txs through the channel?

crates/fuel-core/src/service/adapters.rs
xgreenx: It is an adapter, why do we need to have a port here?

crates/fuel-core/src/service/adapters/executor.rs
xgreenx: They were not send, it is why we return them back and it is up to the caller decide what he wants to do with unsent preconfirmations

crates/services/executor/src/executor.rs
xgreenx: No, it is responsibility of the user of the executor, we handle it in the `PoA` service

crates/services/executor/src/executor.rs
xgreenx: No, Failed transaction will be part of the `Ok(result)` branch. If we got an error during execution on executor level, it means transaction was invalidate from the begging, and we shouldn't include it in the block at all(Like uses non existing inputs, or fails basic validation rules).

crates/services/upgradable-executor/wasm-executor/src/preconfirmation_sender.rs
xgreenx: There is not difference, I think @AurelienFT just wanted to define type explicitly in the WASM executor in the case if someone will update logic of `TransparentPreconfirmationSender` in the future

crates/services/upgradable-executor/wasm-executor/src/new_tx_waiter.rs
xgreenx: The same reason: https://github.com/FuelLabs/fuel-core/pull/2799#discussion_r1982152904

crates/services/executor/src/executor.rs
xgreenx: Agree, it would be nice to have it tested=)

crates/services/executor/src/executor.rs
xgreenx: ++

crates/services/upgradable-executor/wasm-executor/src/preconfirmation_sender.rs
AurelienFT: Yeah I tried to follow what we do for relayer and thinks like that have something defined only for wasm executor

crates/services/executor/src/executor.rs
AurelienFT: Added.

crates/services/executor/src/executor.rs
AurelienFT: Added.

xgreenx (approved):

crates/fuel-core/src/service/adapters/executor.rs
MitchTurner: I know they weren't sent. There is just no difference in the success path and the `Closed` path to the caller.

MitchTurner (commented):
Looks good. Just the one small question, then we can merge.

crates/fuel-core/src/service/adapters/executor.rs
AurelienFT: I changed the closed to return the preconfirmations also. In any cases closed shouldn't happens I also added a warn print

MitchTurner (approved):

crates/fuel-core/src/service/adapters/executor.rs
AurelienFT: @MitchTurner Removed behavior but added comments to explain why we are ignoring errors.

MitchTurner (approved):

xgreenx (approved):

PR #2802, opened 2025-03-05
Tx pool output temporary cache

### LINKED ISSUE #2801: Create a cache with extracted transaction outputs in TxPool
This will allow us to have access to outputs of transaction that txpool is not aware of (because extracted to the block producer) but that are in the process of being executed and avoid dependent transaction of them to be placed inside pendin pool and resolved after 1 second.

AurelienFT:
## Linked Issues/PRs
Resolves #2801

## Description
- Add a new cache with outputs extracted from the pool for the duration of the block.
- Change new txs notifier to be notified only on executable transactions

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/services/txpool_v2/src/pool.rs
+++ b/crates/services/txpool_v2/src/pool.rs
@@ -142,7 +142,7 @@ where
         &mut self,
         tx: ArcPoolTx,
         persistent_storage: &impl TxPoolPersistentStorage,
-    ) -> Result<Vec<ArcPoolTx>, InsertionErrorType> {
+    ) -> Result<(Vec<ArcPoolTx>, bool), InsertionErrorType> {
         let insertion_result = self.insert_inner(tx, persistent_storage);
         self.register_transaction_counts();
         insertion_result
@@ -152,7 +152,7 @@ where
         &mut self,
         tx: std::sync::Arc<PoolTransaction>,
         persistent_storage: &impl TxPoolPersistentStorage,
-    ) -> Result<Vec<std::sync::Arc<PoolTransaction>>, InsertionErrorType> {
+    ) -> Result<(Vec<std::sync::Arc<PoolTransaction>>, bool), InsertionErrorType> {
         let CanStoreTransaction {
             checked_transaction,
             transactions_to_remove,
@@ -212,7 +212,7 @@ where
             .map(|data| data.transaction)
             .collect::<Vec<_>>();
         self.update_stats();
-        Ok(removed_transactions)
+        Ok((removed_transactions, !has_dependencies))
     }
 
     fn update_stats(&self) {
--- a/crates/services/txpool_v2/src/pool.rs
+++ b/crates/services/txpool_v2/src/pool.rs
@@ -1,7 +1,10 @@
 mod collisions;
 
 use std::{
-    collections::HashMap,
+    collections::{
+        HashMap,
+        HashSet,
+    },
     iter,
     time::{
         Instant,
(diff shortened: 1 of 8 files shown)


crates/services/txpool_v2/src/pool.rs
```diff
@@ -300,6 +324,43 @@ where
         }
     }
 
+    fn populate_saved_outputs_cache(&mut self, best_txs: &[StorageData]) {
+        self.extracted_outputs.clear();
```
xgreenx, line 328: I don't see why do we need to clear it here. This place is called several times during the block production, and clearing outputs here means we can't actually use them.

It would be nice to add a test that would catch this incorrect behavior

crates/services/txpool_v2/src/storage/graph.rs
```diff
@@ -654,6 +671,16 @@ impl Storage for GraphStorage {
                                 }
                             }
                             Ok(None) => {
+                                if saved_outputs.contains(&SavedOutput::Coin(
```
xgreenx, line 674: Can we put saved output into the variable please?=D It is hard to read=)

crates/services/txpool_v2/src/storage/graph.rs
```diff
@@ -704,6 +731,11 @@ impl Storage for GraphStorage {
                         match persistent_storage.contract_exist(contract_id) {
                             Ok(true) => {}
                             Ok(false) => {
+                                if saved_outputs
```
xgreenx, line 734: The same here, variable will help

crates/services/txpool_v2/src/service.rs
```diff
@@ -411,7 +412,11 @@ where
                         .or_default();
                     block_height_expiration.push(tx_id);
                 }
-                self.shared_state.new_txs_notifier.send_replace(());
+                if executable {
+                    self.shared_state
+                        .new_executable_txs_notifier
+                        .send_replace(());
+                }
```
xgreenx, line 419: It will not handle the case when we promote transaction from dependent to executed. I think we need to move `new_executable_txs_notifier` into the `TxPool` and notify in the case when we add a new executable transaction, and when we promote to the executable

AurelienFT: Done.

crates/services/txpool_v2/src/storage/graph.rs
AurelienFT: Done.

rymnc (dismissed):

xgreenx (commented):
Looks good to me, but it would be nice to have some tests to confirm that it works as expected and we can submit dependent transaction after their parents where already taken for the execution

xgreenx (approved):

crates/services/txpool_v2/src/pool.rs
```diff
@@ -350,7 +417,9 @@ where
 
     /// Remove transaction but keep its dependents.
     /// The dependents become executables.
```
MitchTurner, line 419: I think this comment could be updated with the new stuff as well.

crates/services/txpool_v2/src/tests/tests_pool.rs
```diff
@@ -1259,6 +1259,35 @@ fn insert_tx_with_blob_already_insert_at_lower_tip() {
     universe.assert_pool_integrity(&[result.unwrap().0]);
 }
 
+#[test]
+fn insert__tx_depends_on_extracted_tx() {
```
MitchTurner, line 1263: nit: this name doesn't tell me what I'm expecting. 

perhaps `verify_and_insert__when_dependent_tx_is_extracted_new_tx_still_accepted` or something like that.

MitchTurner (approved):

xgreenx (approved):

PR #2808, opened 2025-03-06
0.41.7 debug image

xgreenx:


xgreenx:
Is not needed, the issue were with the Pyroscope agent image used by the Grafana

PR #2809, opened 2025-03-06
fix(modifications_history): Reworked the migration logic for historical tables

rymnc:
When the state rewind window of 7d was triggered, the `is_migration_in_progress` was repeatedly called, resulting in multiple iterations over the empty `ModificationsHistoryV1` table, causing weird behaviour with rocksdb as noticed in the flamegraph via pyroscope: https://github.com/FuelLabs/fuel-core/pull/2808

Iteration was slow because compaction didn't have a chance to clean up V1 table. This change removes iteration from the migration process.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
-- [2809](https://github.com/FuelLabs/fuel-core/pull/2809): Prevent `is_migration_in_progress` from running multiple times if previously detected.
+- [2809](https://github.com/FuelLabs/fuel-core/pull/2809): When the state rewind window of 7d was triggered, the `is_migration_in_progress` was repeatedly called, resulting in multiple iterations over the empty ModificationsHistoryV1 table. Iteration was slow because compaction didn't have a chance to clean up V1 table. We removed iteration from the migration process.
 
 ## [Version 0.41.7]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [Version 0.41.8]
+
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [Version 0.41.8]
+
+### Fixed
+
+- [2809](https://github.com/FuelLabs/fuel-core/pull/2809): Prevent `is_migration_in_progress` from running multiple times if previously detected.
```
xgreenx, line 13: May I ask you to change description please?=D

xgreenx (commented):
It would be nice if someone migrated this change to the `master`

AurelienFT (approved):
Simplify a lot of things, way more easy to understand the code also. Porting it to master.

CHANGELOG.md
AurelienFT: Done.

crates/fuel-core/src/state/historical_rocksdb.rs
```diff
@@ -270,16 +287,30 @@ where
...
 
+    fn remove_v1_entries(&self) -> StorageResult<()> {
+        if !self.has_v1_history() {
+            return Ok(())
+        }
+
+        self.has_v1_history
+            .store(false, core::sync::atomic::Ordering::Release);
+        self.db
+            .clear_table(ModificationsHistoryV1::<Description>::column())?;
```
rafal-ch, line 298: Shall we flip these two calls to only store `false` when the `clear_table()` actually finishes w/o error?

rymnc: sounds good, feel free to push that fix up :)

rymnc: addressed in d02ec975cd

xgreenx (approved):

PR #2810, opened 2025-03-06
Bump versions 0.41.8

AurelienFT:


rymnc (approved):
nice, let's wait to check if the fix works before creating the release?

AurelienFT:
Yeah, we are not forced to merge this now just it's here. Will place it on draft as in any case CI will not pass thanks to the spellcheck and everything

PR #2828, opened 2025-03-08
Applied part of follow-up changes to the `TxStatusManager`

xgreenx:
Applied part of follow-up changes to the `TxStatusManager`
- Added backpressure for read requests to the manager
- Removed blocking operations for other services by moving logic into `run` method
- Reworked the way how we prune statuses from the cache. It also fixed the bug where we will clean the cache even if event is fresh

Wrapped statuses into `Arc` to avoid heavy cloning in the case of many receipts and outputs. It required moving from `Submitted {}` format of enum to `Submitted(Submitted)`.

Applied mine comments from the original PR.
Added several TODOs with links to issues.


crates/services/tx_status_manager/src/manager.rs
```diff
@@ -508,65 +477,61 @@ mod tests {
...
-                    data.timestamps.contains_key(&timestamp);
-                assert!(!empty_timestamp_cache_exists);
-            }
+            // When
+            // Sleep for TTL, it should expire the first status
+            std::thread::sleep(TTL);
+
+            // Then
+            let status = tx_status_manager.status(&tx1_id);
+            assert!(status.is_some());
```
rafal-ch, line 534: I think this should say `is_none()` (if I got your intention right).
But in such case the test will fail, because we need to call `update_status()` to trigger the pruning, but here we call `sleep` and the directly ask for status.

xgreenx: Added more comments amd thank you for spotting not calling the prune logic, added it part by simulating insertion of status for another transaciton=)

crates/services/tx_status_manager/src/manager.rs
```diff
@@ -745,6 +584,64 @@ mod tests {
...
+        // Trigger the final pruning, making sure we use ID that is not
+        // in the pool
+        let final_tx_id = {
+            let marker: u64 = 0xDEADBEEF;
+            let mut id = [0u8; 32];
+            id[0..8].copy_from_slice(&marker.to_le_bytes());
+            id
+        };
+        assert!(!tx_id_pool.contains(&final_tx_id));
+        tx_status_manager.status_update(final_tx_id.into(), status_1());
```
rafal-ch, line 623: Maybe it's good to also have the last transaction in the `update_times` collection.

```suggestion
        tx_status_manager.status_update(final_tx_id.into(), status_1());
        update_times.insert(final_tx_id, Instant::now());
```

crates/services/tx_status_manager/src/service.rs
```diff
@@ -1,71 +1,121 @@
...
+    },
+    NotifySkipped {
+        tx_ids_and_reason: Vec<(Bytes32, String)>,
+    },
+}
+
+#[derive(Clone)]
+pub struct SharedData {
+    read_requests_sender: mpsc::Sender<ReadRequest>,
+    write_requests_sender: mpsc::UnboundedSender<WriteRequest>,
```
rafal-ch, line 65: Issue similar to the one I reported [here](https://github.com/FuelLabs/fuel-core/pull/2725#discussion_r1963182638).
Basically, I think the less "unbounded" collections we have in the system the better ;-)

xgreenx: We are okay to take that risk, because we rely on the assumption that other services handle back pressure from outside world. We don't want other services to wait when they write to the TxStatusManager

rafal-ch (approved):
LGTM :+1:

AurelienFT (approved):
I'm also ok with these changes

crates/services/tx_status_manager/src/update_sender.rs
```diff
@@ -256,18 +257,18 @@ impl UpdateSender {
         // left for a given tx_id.
         let mut empty = false;
 
-        if let Some(senders) = senders.get_mut(update.tx_id()) {
+        if let Some(senders) = senders.get_mut(&tx_id) {
             // Retain only senders that are able to receive the update.
-            senders
-                .retain_mut(|sender| sender.try_send(update.clone().into_msg()).is_ok());
```
MitchTurner, line 262: nit: this is a SRP violation. In general I try to avoid operations that couple two different ideas together (in this case sending the message and removing dead senders). Code golf should stay out of production code. Could we move this to a `for` loop perhaps? Then explicitly say "if this fails, it's dead and should be removed".

MitchTurner (approved):

crates/services/tx_status_manager/src/update_sender.rs
xgreenx: We need to rework this logic I think, but that code was before me=D

PR #2830, opened 2025-03-09
fix(sub_services): enable block production only after all other services have been started

rymnc:
> [!NOTE]
> deployed to devnet, working as intended, and resumed block production

### DIFF
--- a/crates/fuel-core/src/service.rs
+++ b/crates/fuel-core/src/service.rs
@@ -444,7 +444,6 @@ impl RunnableService for Task {
         }
 
         params.block_production_trigger.send_trigger();
-        tracing::info!("Block Production has been triggered.");
 
         Ok(self)
     }
--- a/crates/fuel-core/src/service.rs
+++ b/crates/fuel-core/src/service.rs
@@ -139,8 +139,14 @@ impl FuelService {
         // initialize sub services
         tracing::info!("Initializing sub services");
         database.sync_aux_db_heights(shutdown_listener)?;
-        let (services, shared, block_production_trigger) =
-            sub_services::init_sub_services(&config, database)?;
+
+        let block_production_trigger = BlockProductionTrigger::new();
+
+        let (services, shared) = sub_services::init_sub_services(
+            &config,
+            database,
+            block_production_trigger.clone(),
+        )?;
 
         let sub_services = Arc::new(services);
         let task = Task::new(sub_services.clone(), shared.clone())?;
--- a/crates/fuel-core/src/service.rs
+++ b/crates/fuel-core/src/service.rs
@@ -3,6 +3,7 @@ use std::{
     sync::Arc,
 };
 
+use adapters::block_production_trigger::BlockProductionTrigger;
 pub use config::{
     Config,
     DbType,
@@ -89,6 +90,8 @@ pub struct SharedState {
     pub executor: ExecutorAdapter,
     /// The config of the service.
     pub config: Config,
+    /// notifier to trigger block production
+    pub block_production_trigger: BlockProductionTrigger,
 }
 
 pub struct FuelService {
@@ -424,6 +427,10 @@ impl RunnableService for Task {
                 }
             }
         }
+
+        self.shared.block_production_trigger.send_trigger();
+        tracing::info!("Block Production has been triggered.");
+
         Ok(self)
     }
 }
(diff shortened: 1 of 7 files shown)


crates/services/consensus_module/poa/src/ports.rs
```diff
@@ -140,3 +140,43 @@ impl PredefinedBlocks for InMemoryPredefinedBlocks {
...
+    TBP: TriggerBlockProduction,
+{
+    pub(crate) fn new(notifier: TBP) -> Self {
+        Self {
+            notifier,
+            notified: std::sync::Arc::new(std::sync::atomic::AtomicBool::new(false)),
+        }
+    }
+
+    pub(crate) fn should_wait_for_trigger(&self) -> bool {
```
rymnc, line 179: maybe i can add a #[must_use] here

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -506,16 +519,22 @@ where
...
     S: BlockSigner,
     PB: PredefinedBlocks,
     C: GetTime,
+    TBP: TriggerBlockProduction,
 {
     async fn run(&mut self, watcher: &mut StateWatcher) -> TaskNextAction {
+        // more performant to do the sync check first
+        if self.block_production_trigger.should_wait_for_trigger() {
+            self.block_production_trigger.wait_for_trigger().await;
+        }
```
xgreenx, line 536:
```suggestion
        tokio::select! {
            biased;
            _ = watcher.while_started() => {
                return TaskNextAction::Stop
            }
            _ = self.block_production_trigger.wait_for_trigger() => {}
        }
```

crates/services/consensus_module/poa/src/ports.rs
xgreenx: I don't see a point of this method, because you already do `self.notified.load(std::sync::atomic::Ordering::Acquire)`

crates/services/consensus_module/poa/src/service_test.rs
```diff
@@ -220,6 +225,13 @@ impl BlockSigner for FakeBlockSigner {
     }
 }
 
+#[derive(Clone)]
+struct FakeBlockProductionTrigger;
+
+impl TriggerBlockProduction for FakeBlockProductionTrigger {
```
xgreenx, line 231: It would be nice if you also added a unit test to check that this feature works=)

crates/fuel-core/src/service.rs
```diff
@@ -89,6 +90,8 @@ pub struct SharedState {
     pub executor: ExecutorAdapter,
     /// The config of the service.
     pub config: Config,
+    /// notifier to trigger block production
+    pub block_production_trigger: BlockProductionTrigger,
```
xgreenx, line 94: We don't need it to be a public field. Even more, we don't need to store it at all, you can pass it via `TaskParams` of the `FuelService` and trigger it there

crates/services/consensus_module/poa/src/ports.rs
rymnc: addressed in fda6b1e2bc

crates/services/consensus_module/poa/src/service.rs
rymnc: addressed in fda6b1e2bc

crates/services/consensus_module/poa/src/service_test.rs
rymnc: addressed in fda6b1e2bc

crates/fuel-core/src/service.rs
rymnc: addressed in fda6b1e2bc

crates/fuel-core/src/service.rs
```diff
@@ -424,6 +442,10 @@ impl RunnableService for Task {
                 }
             }
         }
+
+        params.block_production_trigger.send_trigger();
+        tracing::info!("Block Production has been triggered.");
```
xgreenx, line 447: If block production is disabled, this log will be invalid. Maybe we can move it to the PoA service?

xgreenx (commented):
Looks good to me

crates/fuel-core/src/service.rs
rymnc: i'm moving it to the port impl in the poa crate, that's the only place where we know that it has been triggered, otherwise we will have logs for each time `run` is called. chek d16295bbc9

xgreenx (approved):

rymnc:
I'll do the changelog in the version bumping pr

Issue #2282, opened 2024-10-03
Redesign `RunnableService` and `SharedData`


MitchTurner:
This is addressing two issues:
1. Something that implements `RunnableService` should only be constrained by things related to running the service. "Sharing data" is a separate concern
2. `SharedData` is inherently a violation of [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle) in that we don't need to put all the shared data in one interface. Under the hood, it might be necessary to have disparate objects held together, but we should provide an interface for those things; i.e. instead of `service.shared_data().foo_bar` we should prefer `service.get_foo_bar()`.

We've had many conversations about this and hopefully this can be a single source of truth to track our decision.

It seems that most people agree that `RunnableService` should no longer have a `SharedData` field. How we want to solve the issue of providing data from inside a service to the outside world is up for debate still.

In the end, the trait should look like this:
```rs
#[async_trait::async_trait]
pub trait RunnableService: Send {
    const NAME: &'static str;

    type SharedData: Clone + Send + Sync;

    type Task: RunnableTask;

    type TaskParams: Send;

    async fn run_service(
        self,
        state_watcher: &StateWatcher,
        params: Self::TaskParams,
    ) -> anyhow::Result<Self::Task>;
}
```

With `SharedData` and `shared_data()` removed, and `into_task` renamed to `run_service`.

### Green's idea for separate trait

```rs
pub trait ServiceWithSharedData: RunnableService {
    type SharedData: Clone + Send + Sync;

    fn shared_data(&self) -> Self::SharedData;
}
```
(preliminary trait name)

Where you can optionally describe `SharedData` for some services, but not all of them. This is a good separation of concerns.

Issues:

This solves for #1 but not for #2 of the problems listed above.

### Mitch's idea 

TBD WIP TODO LOL

netrome:
I like the idea of the separate trait as a starting point because it would allow us to migrate implementations of this trait to any new solution one by one, instead of doing a huge refactor in a single PR.

PR #2832, opened 2025-03-10
port(relayer|poa): port fixes from v0.41.9 to master

rymnc:
## Linked Issues/PRs
- #2829
- #2830

## Description
- fixes delayed block production due to relayer not updating internal state after downloading logs
- includes deterministic start of block production.

## Checklist

### Before requesting review

- [ ] [Platform
documentation](https://github.com/FuelLabs/devrel-requests/issues/new?assignees=&labels=new+request&projects=&template=NEW-REQUEST.yml&title=%5BRequest%5D%3A+)
(for out-of-organization contributors, the person merging the PR will do
this)
- [ ] Someone else?

### DIFF
--- a/crates/fuel-core/src/service.rs
+++ b/crates/fuel-core/src/service.rs
@@ -3,6 +3,7 @@ use std::{
     sync::Arc,
 };
 
+use adapters::block_production_trigger::BlockProductionTrigger;
 pub use config::{
     Config,
     DbType,
@@ -75,11 +76,7 @@ pub struct SharedState {
     pub network: Option<fuel_core_p2p::service::SharedState>,
     #[cfg(feature = "relayer")]
     /// The Relayer shared state.
-    pub relayer: Option<
-        fuel_core_relayer::SharedState<
-            Database<crate::database::database_description::relayer::Relayer>,
-        >,
-    >,
+    pub relayer: Option<fuel_core_relayer::SharedState>,
     /// The GraphQL shared state.
     pub graph_ql: crate::fuel_core_graphql_api::api_service::SharedState,
     /// The underlying database.
@@ -139,11 +136,23 @@ impl FuelService {
         // initialize sub services
         tracing::info!("Initializing sub services");
         database.sync_aux_db_heights(shutdown_listener)?;
-        let (services, shared) = sub_services::init_sub_services(&config, database)?;
+
+        let block_production_trigger = BlockProductionTrigger::new();
+
+        let (services, shared) = sub_services::init_sub_services(
+            &config,
+            database,
+            block_production_trigger.clone(),
+        )?;
 
         let sub_services = Arc::new(services);
         let task = Task::new(sub_services.clone(), shared.clone())?;
-        let runner = ServiceRunner::new(task);
+        let runner = ServiceRunner::new_with_params(
+            task,
+            TaskParams {
+                block_production_trigger,
+            },
+        );
         let bound_address = runner.shared.graph_ql.bound_address;
 
         Ok(FuelService {
@@ -396,12 +405,17 @@ impl Task {
     }
 }
 
+#[derive(Default)]
+struct TaskParams {
+    block_production_trigger: BlockProductionTrigger,
+}
+
 #[async_trait::async_trait]
 impl RunnableService for Task {
     const NAME: &'static str = "FuelService";
     type SharedData = SharedState;
     type Task = Task;
-    type TaskParams = ();
+    type TaskParams = TaskParams;
 
     fn shared_data(&self) -> Self::SharedData {
         self.shared.clone()
@@ -410,7 +424,7 @@ impl RunnableService for Task {
     async fn into_task(
         mut self,
         watcher: &StateWatcher,
-        _: Self::TaskParams,
+        params: Self::TaskParams,
     ) -> anyhow::Result<Self::Task> {
         let mut watcher = watcher.clone();
 
@@ -424,6 +438,9 @@ impl RunnableService for Task {
                 }
             }
         }
+
+        params.block_production_trigger.send_trigger();
+
         Ok(self)
     }
 }
--- a/crates/services/consensus_module/poa/src/ports.rs
+++ b/crates/services/consensus_module/poa/src/ports.rs
@@ -142,3 +142,42 @@ impl PredefinedBlocks for InMemoryPredefinedBlocks {
 pub trait GetTime: Send + Sync {
     fn now(&self) -> Tai64;
 }
+
+pub trait TriggerBlockProduction: Send + Sync {
+    fn wait_for_trigger(&self) -> impl core::future::Future<Output = ()> + Send;
+}
+
+pub(crate) struct BlockProductionTrigger<TBP> {
+    notifier: TBP,
+    notified: std::sync::Arc<std::sync::atomic::AtomicBool>,
+}
+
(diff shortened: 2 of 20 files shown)


crates/fuel-core/src/service.rs
```diff
@@ -3,6 +3,7 @@ use std::{
     sync::Arc,
 };
 
+use adapters::block_production_trigger::BlockProductionTrigger;
```
netrome, line 6: I find it a bit confusing to reuse the "trigger" name for this synchronization primitive, as "trigger" already has a meaning for the PoA service.

If I understand this correctly, what we have here is just a type to signal that we're ready to start producing blocks. How about renaming it to something like `ReadySignal`? Or `StartBarrier`?

```suggestion
use adapters::block_production_trigger::ReadySignal;
```

crates/services/consensus_module/poa/src/ports.rs
```diff
@@ -142,3 +142,42 @@ impl PredefinedBlocks for InMemoryPredefinedBlocks {
 pub trait GetTime: Send + Sync {
     fn now(&self) -> Tai64;
 }
+
+pub trait TriggerBlockProduction: Send + Sync {
+    fn wait_for_trigger(&self) -> impl core::future::Future<Output = ()> + Send;
+}
```
netrome, line 148: Again the terminology is a bit confusing. We're not triggering block production here, we're awaiting the dependencies to be ready.

```suggestion
pub trait WaitForReadySignal: Send + Sync {
    fn wait_for_ready_signal(&self) -> impl core::future::Future<Output = ()> + Send;
}
```

crates/services/consensus_module/poa/src/ports.rs
```diff
@@ -142,3 +142,42 @@ impl PredefinedBlocks for InMemoryPredefinedBlocks {
...
+where
+    TBP: TriggerBlockProduction,
+{
+    pub(crate) fn new(notifier: TBP) -> Self {
+        Self {
+            notifier,
+            notified: std::sync::Arc::new(std::sync::atomic::AtomicBool::new(false)),
+        }
+    }
+}
```
netrome, line 183: This looks wrong to me.

1. We should only await the notifier once when the service is started, so in the expected case this is dead code.
2. If we in any place chose to restart the service, wouldn't we like to retain the same control in that scenario that this service only is started when the dependencies has started?

I tried removing this and all tests in the PoA module still pass for me. I think this type should be removed.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -582,16 +595,25 @@ where
...
+    TBP: TriggerBlockProduction,
 {
     async fn run(&mut self, watcher: &mut StateWatcher) -> TaskNextAction {
+        tokio::select! {
+            biased;
+            _ = watcher.while_started() => {
+                return TaskNextAction::Stop
+            }
+            _ = self.block_production_trigger.wait_for_trigger() => {}
+        }
```
netrome, line 615: This solution feels very brittle to me. I take it there's no good way to just wait for the other services to be started before starting this one without doing a major redesign of our service architecture?

Another evidence of why we need to rework our service traits and how we instantiate our services.

netrome (approved):
I understand that this fix was thrown together quite quickly, and the solution is giving me a lot of pause. That being said, it looks correct and like it solves the intended problems. Some things that could improve the PR:

1. I would like to see a change from the "Trigger" term, since it is inappropriate and confusing in this context.

2. The cached `BlockProductionTrigger` seems useless and potentially invalid to me. We should remove it imo.

Ideally I'd also like to see a complete redesign of the solution where we'd simply just not start the service until any precondition for running the service is met, but that would likely require a much larger effort so perhaps it's still best to save that work for https://github.com/FuelLabs/fuel-core/issues/2282 - I know we have higher priorities right now and need to move forward, so I can live with the current hacky solution for the time being.

crates/services/consensus_module/poa/src/ports.rs
rymnc: the notion of a service being "started" doesn't really exist, it's just initialization and the run-loop.

image
from https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html

the tests all probably just complete in a single loop, hence pass. the integ tests would fail. we don't wait to wait every single time in the run loop, which is why we cache the result from the first notify.

crates/services/consensus_module/poa/src/service.rs
rymnc: i wouldn't call it major, we just need to add a `on_startup` trait method that's called before `start_and_await` for the services. we can block on this to ensure deterministic behaviour.

crates/services/consensus_module/poa/src/ports.rs
rymnc: done in [8bbfc81](https://github.com/FuelLabs/fuel-core/pull/2832/commits/8bbfc818b36d286ec04099b5b949520cf0e0b593) & [324d5a0](https://github.com/FuelLabs/fuel-core/pull/2832/commits/324d5a054538567c2889922bdede0beb392ea94a) & [20467cd](https://github.com/FuelLabs/fuel-core/pull/2832/commits/20467cd69c3dbdd3b911979add2329f1e44d66cf) & [e124d36](https://github.com/FuelLabs/fuel-core/pull/2832/commits/e124d364f3fbbcf2f26a41b3373bcde65c5503d5)

AurelienFT (dismissed):

crates/services/consensus_module/poa/src/ports.rs
netrome: > the notion of a service being "started" doesn't really exist, it's just initialization and the run-loop.

I'd consider the act of starting the run loop either explicitly or implicitly through `service_runner.start_and_await()` starting the service. But you're right that the service runner may call this multiple times, so I can see the point of caching the value now.

crates/services/consensus_module/poa/src/ports.rs
netrome: Nice, thank you!

crates/services/consensus_module/poa/src/service.rs
netrome: Sure that could be an approach. It would also be nice if we could express the required preconditions for starting the PoA service explicitly, rather than relying on anyone instantiating the service implicitly knowing what's required before sending the start signal.

netrome (dismissed):

crates/services/relayer/src/service/test.rs
```diff
@@ -173,20 +173,22 @@ async fn deploy_height_does_not_override() {
     assert_eq!(*mock_db.get_finalized_da_height().unwrap(), 50);
 }
 
-#[test_case(6, Some(6), Some(6); "if local is up to date with remote, then update sync")]
-#[test_case(6, Some(100), Some(100); "if local is somehow ahead of remote, then update sync")]
-#[test_case(6, Some(5), None; "if local is behind remote, then nothing sent to sync")]
-#[test_case(6, None, None; "if local is not set, then nothing sent to sync")]
+const STATING_HEIGHT: u64 = 2;
```
xgreenx, line 176:
```suggestion
const STARTING_HEIGHT: u64 = 2;
```

xgreenx (approved):

xgreenx (approved):

xgreenx (approved):

AurelienFT (approved):

PR #2840, opened 2025-03-10
Use a new feature from the VM to get all required inputs in one dry run

xgreenx:
This change starts using the https://github.com/FuelLabs/fuel-vm/pull/918 feature for the `asemble_tx` endpoint.


crates/client/src/client/schema/tx/transparent_receipt.rs
```diff
@@ -72,11 +72,14 @@ impl TryFrom<Receipt> for fuel_tx::Receipt {
     fn try_from(schema: Receipt) -> Result<Self, Self::Error> {
         Ok(match schema.receipt_type {
             ReceiptType::Call => fuel_tx::Receipt::Call {
-                id: schema.id.map(|id| id.into()).unwrap_or_default(),
+                id: schema
+                    .id
+                    .map(Into::into)
+                    .unwrap_or(fuel_tx::ContractId::zeroed()),
```
netrome, line 78: Why do we no longer use `unwrap_or_default()`?

crates/fuel-core/src/schema/tx/assemble_tx.rs
```diff
@@ -783,6 +803,9 @@ where
             *script.script_gas_limit_mut() = max_gas_limit;
 
             let (updated_tx, new_status) = self.arguments.dry_run(script).await?;
+
+            self.dry_run_count = self.dry_run_count.saturating_add(1);
```
netrome, line 807: I think we should do a checked add here. I know we won't hit it in practice, but if we do it feels better to return an error than loop forever.

```suggestion
            self.dry_run_count = self.dry_run_count.checked_add(1).ok_or_else(|| anyhow::anyhow!("dry run count overflow"))?;
```

netrome (changes requested):
Just one saturating add that should be checked imo. Otherwise LGTM.

crates/client/src/client/schema/tx/transparent_receipt.rs
xgreenx: I prefer an explicit way if it is related to business logic. In the case if default implementation is changed. On `fuel-vm` side we use `ContractId::zeroed()`.

crates/services/executor/src/executor.rs
```diff
@@ -324,13 +330,25 @@ impl ExecutionData {
 /// Per-block execution options
 #[derive(serde::Serialize, serde::Deserialize, Clone, Default, Debug)]
 pub struct ExecutionOptions {
-    // TODO: This is a bad name and the real motivation for this field should be specified
-    /// UTXO Validation flag, when disabled the executor skips signature and UTXO existence checks
-    pub extra_tx_checks: bool,
+    /// The flag allows the usage of fake coins in the inputs of the transaction.
+    /// When `false` the executor skips signature and UTXO existence checks.
+    pub forbid_fake_coins: bool,
```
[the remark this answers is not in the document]
rymnc: i think you can still set the serde attribute for what it should serialize _to_, but since we are acknowleding breaking change its ok

crates/fuel-core/src/schema/tx/assemble_tx.rs
```diff
@@ -802,7 +831,7 @@ where
             match &status.result {
                 TransactionExecutionResult::Success { .. } => break,
                 TransactionExecutionResult::Failed { receipts, .. } => {
-                    for receipt in receipts.iter().rev() {
+                    for receipt in receipts.iter() {
```
rymnc, line 834: i think this is also a breaking change if people are using plain indexes to access some specific receipt, why was this change made?

rymnc (commented):
in general lgtm, had a question why we dont reverse receipts in event of transaction failure

AurelienFT (approved):
LGTM

crates/fuel-core/src/schema/tx/assemble_tx.rs
xgreenx: It was added in the parent PR(which is not released yet), and I forgot to remove it there=D I wanted to have an optimization assuming that panic receipts at the end, but we can have multiple panic receipts now, so this optimization will not work because we still need to iterate over all receipts

rymnc: ok sounds good

rymnc (approved):

netrome (approved):

AurelienFT (approved):

PR #2841, opened 2025-03-11
Allow estimation of predicates on submission of the transaction

xgreenx:
Allow estimation of predicates on submission of the transaction.
It affects the following endpoints:
- `submit`
- `submit_and_await`
- `submit_and_await_status`

This behavior is controlled via an optional `estimatePredicates` argument. The change is backward compatible with SDK. The change is not forward-compatible with Rust SDK in the case of the `estiamte_predicates` flag being used.

### DIFF
--- a/crates/fuel-core/src/schema/tx.rs
+++ b/crates/fuel-core/src/schema/tx.rs
@@ -48,6 +48,7 @@ use fuel_core_storage::{
     iter::IterDirection,
     Error as StorageError,
     IsNotFound,
+    PredicateStorageRequirements,
     Result as StorageResult,
 };
 use fuel_core_tx_status_manager::TxStatusMessage;
@@ -75,6 +76,7 @@ use futures::{
 };
 use std::{
     borrow::Cow,
+    future::Future,
     iter,
 };
 use types::{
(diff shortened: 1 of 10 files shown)


netrome (approved):
LGTM

AurelienFT (dismissed):

AurelienFT (approved):

crates/fuel-core/src/schema/tx.rs
```diff
@@ -862,4 +880,40 @@ impl<'a> ContextExt for Context<'a> {
...
+        for input in tx.inputs()? {
+            if input.predicate().is_some() {
+                has_predicates = true;
+                break;
+            }
+        }
+
+        if !has_predicates {
+            return Ok(tx);
+        }
```
Dentosal, line 900: maybe something like:

```suggestion
        if tx.inputs()?.all(|input| input.predicate().is_none()) {
            return Ok(tx);
        }
```

crates/fuel-core/src/schema/tx.rs
```diff
@@ -862,4 +880,40 @@ impl<'a> ContextExt for Context<'a> {
...
+            .data_unchecked::<ChainInfoProvider>()
+            .current_consensus_params();
+
+        let memory_pool = self.data_unchecked::<SharedMemoryPool>();
+        let memory = memory_pool.get_memory().await;
+
+        let parameters = CheckPredicateParams::from(params.as_ref());
+        let tx = tokio_rayon::spawn_fifo(move || {
+            let result = tx.estimate_predicates(&parameters, memory, &query);
+            result.map(|_| tx)
```
Dentosal, line 912: Something like this should be more readable

```suggestion
            let _ = tx.estimate_predicates(&parameters, memory, &query)?;
            Ok(tx)
```

Dentosal (approved):

PR #2850, opened 2025-03-11
chore: Port the block fetcher from the network watchtower

### LINKED ISSUE #2849: Port the block fetcher from `network-watchtower`
The state root service needs a block fetcher, and it would be nice to share the logic with the one in the network watchtower. We tried initially to depend directly on it, however it results in some compatibility problems with different `fuel-core-types` versions.

Instead, we'll incorporate a local version of the block fetcher in fuel core which can be reused in the network watchtower once released.

netrome:
closes #2849

### DIFF
--- a/.changes/added/2850.md
+++ b/.changes/added/2850.md
@@ -1,0 +1,1 @@
+Port the block fetcher from the network watchtower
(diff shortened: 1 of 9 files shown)


netrome:
We decided to move the state root service to the fuel-proof-system repo instead.

PR #2855, opened 2025-03-11
Add expiration check on pending pool and refactor extracted outputs cache

AurelienFT:
## Description

Add an expiration interval check for pending pool and refactor `extracted_ouputs`.

### Changes in extracted outputs
- 2 data structures one for coins one for contracts
- Now it's not fully clear when receiving a block but only clear when received skipped or executed transactions
- Now clears also when the outputs are used in inputs of another extracted transaction

## Checklist

### Before requesting review

### DIFF
--- a/crates/services/txpool_v2/src/extracted_outputs.rs
+++ b/crates/services/txpool_v2/src/extracted_outputs.rs
@@ -21,6 +21,7 @@ use fuel_core_types::{
 
 pub struct ExtractedOutputs {
     contract_created: HashMap<ContractId, TxId>,
+    contract_created_by_tx: HashMap<TxId, Vec<ContractId>>,
     coins_created: HashMap<TxId, HashMap<u16, (Address, u64, AssetId)>>,
 }
 
@@ -28,6 +29,7 @@ impl ExtractedOutputs {
     pub fn new() -> Self {
         Self {
             contract_created: HashMap::new(),
+            contract_created_by_tx: HashMap::new(),
             coins_created: HashMap::new(),
         }
     }
@@ -46,6 +48,10 @@ impl ExtractedOutputs {
             match output {
                 Output::ContractCreated { contract_id, .. } => {
                     self.contract_created.insert(*contract_id, tx_id);
+                    self.contract_created_by_tx
+                        .entry(tx_id)
+                        .or_default()
+                        .push(*contract_id);
                 }
                 Output::Coin {
                     to,
@@ -90,7 +96,12 @@ impl ExtractedOutputs {
     }
 
     pub fn new_executed_transaction(&mut self, tx_id: &TxId) {
-        self.contract_created.retain(|_, v| v != tx_id);
+        let contract_ids = self.contract_created_by_tx.remove(tx_id);
+        if let Some(contract_ids) = contract_ids {
+            for contract_id in contract_ids {
+                self.contract_created.remove(&contract_id);
+            }
+        }
         self.coins_created.remove(tx_id);
     }
 
--- a/crates/services/txpool_v2/src/extracted_outputs.rs
+++ b/crates/services/txpool_v2/src/extracted_outputs.rs
@@ -1,9 +1,6 @@
 //! Temporary module to extract outputs until we know that they have been settled in database or skipped.
 
-use std::collections::{
-    HashMap,
-    HashSet,
-};
+use std::collections::HashMap;
 
 use fuel_core_types::{
     fuel_tx::{
@@ -16,20 +13,21 @@ use fuel_core_types::{
         ContractId,
         Input,
         Output,
+        TxId,
         UtxoId,
     },
     services::txpool::ArcPoolTx,
 };
 
 pub struct ExtractedOutputs {
-    contract_created: HashSet<ContractId>,
-    coins_created: HashMap<UtxoId, (Address, u64, AssetId)>,
+    contract_created: HashMap<ContractId, TxId>,
+    coins_created: HashMap<TxId, HashMap<UtxoId, (Address, u64, AssetId)>>,
 }
 
 impl ExtractedOutputs {
     pub fn new() -> Self {
         Self {
-            contract_created: HashSet::new(),
+            contract_created: HashMap::new(),
             coins_created: HashMap::new(),
         }
     }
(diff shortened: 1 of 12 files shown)


crates/services/txpool_v2/src/pool.rs
```diff
@@ -362,6 +367,40 @@ where
...
+                        ..
+                    })
+                    | Input::CoinPredicate(CoinPredicate {
+                        utxo_id,
+                        owner,
+                        amount,
+                        asset_id,
+                        ..
+                    }) => {
+                        self.extracted_outputs.remove(&SavedOutput::Coin(
```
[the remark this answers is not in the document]
AurelienFT: Will do but it will make the code here : https://github.com/FuelLabs/fuel-core/blob/f5503405d3e438f4d08b960e9428e62f642d9814/crates/services/txpool_v2/src/storage/graph.rs#L680 more complex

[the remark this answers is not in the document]
AurelienFT: Also there is contract_id which for us to have an enum with both for the key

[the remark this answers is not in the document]
AurelienFT: And the value would be different if it's a coin or a contract which wqill force an enum also. Are you sure you want to do this @xgreenx ?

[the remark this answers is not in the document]
xgreenx: It only used in several places, so I don't think that it will affect a lot of code=D 
image


BTW, `self.extracted_outputs.clear();` potentially can clean the outputs added to the `self.extracted_outputs` for the next block. SO maybe we also want to rework it, by removing only outputs from confirmed transactions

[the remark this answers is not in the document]
AurelienFT: Fully refactored this field

crates/services/txpool_v2/src/extracted_outputs.rs
```diff
@@ -0,0 +1,119 @@
...
+        Output,
+        TxId,
+        UtxoId,
+    },
+    services::txpool::ArcPoolTx,
+};
+
+pub struct ExtractedOutputs {
+    contract_created: HashMap<ContractId, TxId>,
+    coins_created: HashMap<TxId, HashMap<UtxoId, (Address, u64, AssetId)>>,
```
xgreenx, line 24:
```suggestion
    coins_created: HashMap<TxId, HashMap<u16, (Address, u64, AssetId)>>,
```

crates/services/txpool_v2/src/extracted_outputs.rs
```diff
@@ -0,0 +1,119 @@
...
+            }
+        }
+        for input in tx.inputs() {
+            match input {
+                Input::CoinSigned(CoinSigned { utxo_id, .. })
+                | Input::CoinPredicate(CoinPredicate { utxo_id, .. }) => {
+                    self.coins_created
+                        .entry(*utxo_id.tx_id())
+                        .or_default()
+                        .remove(utxo_id);
```
xgreenx, line 78: `or_default` will create a trash in the memory, so if it already `Vacant`, you can just do nothing.

crates/services/txpool_v2/src/extracted_outputs.rs
```diff
@@ -0,0 +1,119 @@
...
+                | Input::MessageDataSigned(_) => {
+                    continue;
+                }
+            }
+        }
+    }
+
+    pub fn new_skipped_transaction(&mut self, tx_id: &TxId) {
+        self.new_executed_transaction(tx_id);
+    }
```
xgreenx, line 93: You will add outputs from skipped transaction into saved outputs, that is not right

crates/services/txpool_v2/src/extracted_outputs.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    pub fn new_executed_transaction(&mut self, tx_id: &TxId) {
+        self.contract_created.retain(|_, v| v != tx_id);
```
xgreenx, line 96: It can be very expensive operation, it is better to use double index map

xgreenx: Alternative, you can have a `HashSet` of tx ids to remove, and during one `retain` you could remove all tx ids.

But double index hash map is faster in general use case

crates/services/txpool_v2/src/extracted_outputs.rs
AurelienFT: No, this function do not add the outputs but remove them from the cache

crates/services/txpool_v2/src/extracted_outputs.rs
AurelienFT: Done

crates/services/txpool_v2/src/extracted_outputs.rs
AurelienFT: Done

crates/services/txpool_v2/src/extracted_outputs.rs
AurelienFT: Done

crates/services/txpool_v2/src/extracted_outputs.rs
xgreenx: Oh, you have almost similar names and I thought you call another function here=D `new_extracted_transaction` vs `new_executed_transaction`

Yeah, it should be fine

crates/services/txpool_v2/src/extracted_outputs.rs
```diff
@@ -0,0 +1,129 @@
...
+        TxId,
+        UtxoId,
+    },
+    services::txpool::ArcPoolTx,
+};
+
+pub struct ExtractedOutputs {
+    contract_created: HashMap<ContractId, TxId>,
+    contract_created_by_tx: HashMap<TxId, Vec<ContractId>>,
+    coins_created: HashMap<TxId, HashMap<u16, (Address, u64, AssetId)>>,
```
rafal-ch, line 25: A nit:
For clarity, `HashMap<TxId, HashMap<u16, (Address, u64, AssetId)>>` maybe deserves a dedicated type to be introduced.

It'll also allow to make [code like this](https://github.com/FuelLabs/fuel-core/pull/2855/files#diff-a5b4c3f0c7a4bf653cd9d7e88b8aa121c8b966395ef1c89823ff99638dd5e67aR124-R126) more lean: `a == address && am == amount && asid == asset_id`

crates/services/txpool_v2/src/extracted_outputs.rs
```diff
@@ -0,0 +1,129 @@
...
+                | Input::MessageDataSigned(_) => {
+                    continue;
+                }
+            }
+        }
+    }
+
+    pub fn new_skipped_transaction(&mut self, tx_id: &TxId) {
+        self.new_executed_transaction(tx_id);
+    }
```
rafal-ch, line 96: `new_skipped_transaction()` just calling `new_executed_transaction` is confusing when you don't know the context. Maybe some renaming or at least additional comment could help

At first glance this code is similar to something like this :smile: 
```
    fn insert(&self, val) { self.remove(val) }
```

crates/services/txpool_v2/src/pool_worker.rs
```diff
@@ -330,6 +343,9 @@ where
                     self.insert(tx, source);
                 }
             }
+            _ = pending_pool_expiration_check.tick() => {
+                self.pending_pool.expire_transactions(self.notification_sender.clone());
```
rafal-ch, line 347: It's not introduced in this PR, but I just noticed that internally, the `expire_transactions()` function sends out the `PoolNotification::ErrorInsertion` variant (in case of error).

Not sure if this is correct, but even if it isn't, it's not harmful. Leaving a comment for your consideration.

rafal-ch (approved):
No blockers, just some nits and minor issues.

Issue #2823, opened 2025-03-07
Implement pre-confirmation signature verification


xgreenx:
The transaction status manager should keep track of all public keys emitted by the block producers to generate pre-confirmations and verify that the block producer signed each confirmation.

MitchTurner:
I think there is an edge case where it might receive the signature before receiving the key. We should address that.

PR #2848, opened 2025-03-11
Link and activate preconfirmations and add integrations tests

AurelienFT:
Closes https://github.com/FuelLabs/fuel-core/issues/2590

## Description
Goal of this PR is to link all the pre-confirmation works and to tests all the scenarios possible for preconfirmations in a e2e way.

### Changes
- Authorized to have different values for `utxo_validation` between pool and executor if in debug
- If the `SignMode` is unavailable don't launch block production (instead of launching it but each block production fails)
- Submitted status is sent directly through txpool to avoid inconsistency.

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog
- [x] New behavior is reflected in tests
- [x] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed)

### Before requesting review
- [x] I have reviewed the code myself
- [x] I have created follow-up issues caused by this PR and linked them here

PR #2856, opened 2025-03-11
Add preconfirmation signature verification to tx status manager

MitchTurner:
## Linked Issues/PRs

Closes https://github.com/FuelLabs/fuel-core/issues/2823

## Description

Adds  generic logic to the tx status manager task to accept delegate keys and check signatures with them.

The logic of actually tracking the delegate keys and doing verification is behind the `SignatureVerification` trait, which will be implemented in a follow-up PR.

## Checklist

### Before requesting review


AurelienFT (approved):
Nice I like the implementation !

crates/services/tx_status_manager/src/service.rs
```diff
@@ -94,41 +102,91 @@ impl SharedData {
...
+    early_preconfirmations: Vec<Sealed<Preconfirmations, Bytes64>>,
 }
 
-impl Task {
-    // TODO: Implement signatures verifications logic for preconfirmation logic.
-    //  https://github.com/FuelLabs/fuel-core/issues/2823
-    fn new_preconfirmations_from_p2p(
+#[allow(dead_code)]
+/// Interface for signature verification of preconfirmations
+pub trait SignatureVerification: Send {
```
xgreenx, line 117: It should be in the ports if you want to define its implementation outside of the service crate.

But I think it is an overkill to do it via the port. Because you already expect concrete types from the p2p, so you already know what signature types do you expect.

If you plan to abstract the signature types in the future, then yeah, it makes sense, but for not it looks like an overkill.

crates/services/tx_status_manager/src/service.rs
```diff
@@ -94,41 +102,91 @@ impl SharedData {
...
+                    .signature_verification
+                    .check_preconfirmation_signature(&sealed)
+                    .await
+                {
+                    self.handle_verified_preconfirmation(sealed);
+                } else {
+                    // TODO: Allow to retry later when a new delegate key is added
+                    tracing::warn!("Preconfirmation signature verification failed, will try once more when a new delegate key is added");
+                    self.early_preconfirmations.push(sealed);
+                }
```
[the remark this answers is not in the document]
MitchTurner: We are going to lose some preconfirmations then probably. It's an edge case, but if delegations happen at the same time as the pre-confirmations it's bound to happen.

crates/services/tx_status_manager/src/service.rs
MitchTurner: > you already expect concrete types from the p2p

I know what you mean and I've thought about it. I think it was a mistake to have the signatures concrete. I am implementing the port right now, so we can collapse it after if we think that's better.

crates/services/tx_status_manager/src/service.rs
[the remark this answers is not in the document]
xgreenx: It is fine if you are a new in the network. But if you running for some time, it shouldn't happen

crates/services/tx_status_manager/src/service.rs
xgreenx: To be in parity with pre-confirmation service, I think we need to make signatures and public keys to be generic.

But it is more a tech debt. it is why I'm saying maybe we could go without generic for verification now, and implement it later when we try to move to generics for P2P message.

In this case, it will be easy to maintain a code for now, plus you can implement verification logic right away, plus clean up logic based on expiration.

MitchTurner: The generics make the tests easy :)

xgreenx: It is true, but in that way, you test more the internal details of how the service works=) Instead of testing how it works on actual inputs from p2p(since you know concrete type).

Using concrete types allows you to encapsulate the logic within the crate without requiring the `fuel-core` to implement it and test it on the adaptor level=) Plus, you will not expose internal details via the generic because it will be just a logic within the service itself.

That said, I'm also okay with being generic over the signature and public key.

But from a project management perspective, I still prefer to start with an implementation that uses concrete primitives (like known signature and public key). Later, we can always make them generic if we need customization properties. It just speeds up development and allows us to get feedback about the actual implementation earlier(it is what matters at the end). Plus, tests will give coverage over real service instead of facked one.

With the current approach, I think we moved too much logic to the `fuel-core`. The implementation of the adaptor from https://github.com/FuelLabs/fuel-core/pull/2866 could easily live inside of the service but still be generic over the signature and public key(logic like `signature.verify(public_key, message)` only should be abstracted).

bin/fuel-core/src/cli/run.rs
```diff
@@ -714,6 +714,8 @@ impl Command {
                 max_tx_update_subscriptions: tx_number_active_subscriptions,
                 subscription_ttl,
                 status_cache_ttl: status_cache_ttl.into(),
+                // TODO: Where can I get this value?
+                protocol_public_key: Default::default(),
```
AurelienFT, line 718: I think this is a left over. If not it's not needed check to see how I implemented it without using it : https://github.com/FuelLabs/fuel-core/pull/2848

crates/services/tx_status_manager/src/config.rs
```diff
@@ -8,6 +9,8 @@ pub struct Config {
     pub subscription_ttl: Duration,
     /// Maximum time to keep the status in the cache of the manager.
     pub status_cache_ttl: Duration,
+    /// Protocol Signing Key, i.e. the block signer's public key
+    pub protocol_public_key: PublicKey,
```
AurelienFT, line 13: I think this is a left over. If not it's not needed check to see how I implemented it without using it : https://github.com/FuelLabs/fuel-core/pull/2848

crates/services/tx_status_manager/src/tests/mod.rs
```diff
@@ -9,3 +24,67 @@ mod tests_subscribe;
 mod tests_update_stream_state;
 mod universe;
 mod utils;
+
+use fuel_core_types::services::p2p::DelegatePreConfirmationKey;
+use tracing_subscriber as _;
```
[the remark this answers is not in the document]
AurelienFT: +1

crates/services/tx_status_manager/src/service.rs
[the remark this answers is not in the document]
AurelienFT: +1, even if it can be annoying for tests, maybe keep it under test-helpers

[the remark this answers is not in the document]
MitchTurner: There is no situation where the delegate key signature could come after something signed by the delegate key?

(I've already implemented the `TODO`, I just forgot to remove it)

bin/fuel-core/src/cli/run.rs
MitchTurner: I added this field.

AurelienFT: I don't understand the point of this field. I linked my PR because I did all the linking etc with this field still like that and everything is working so it doesn't seems to be useful to be set. For me the tx stauts manager should have an option and before receiving the first message from p2p it should be None and after use the key received in p2p

crates/services/tx_status_manager/src/config.rs
MitchTurner: Removed.

crates/services/tx_status_manager/src/service.rs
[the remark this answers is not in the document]
xgreenx: It can arrive, but it is fine to ignore, because maybe this key doesn't exist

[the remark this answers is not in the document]
MitchTurner: I feel like ideally we handle that with reputation, but I'll remove it for now and leave a note.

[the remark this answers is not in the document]
MitchTurner: removed

bin/fuel-core/src/cli/run.rs
MitchTurner: The point of the delegate key is that it is delegated by a trusted key, specifically the PoA Consensus Key used for block production.

We can't just trust any signed message, we must ensure that it is the same signer as the blocks.

I'm not sure which part of your code is related to this concern.

I've removed this in favor of a `ProtocolPublicKey` port for which we have an adapter that wraps `ConsensusConfig`.

crates/services/tx_status_manager/src/service.rs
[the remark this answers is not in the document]
xgreenx: Yes, here we need to decrease the reputation of the peer in the case if he tries to spam us with invalid reconfirmations. But it should a follow up=)

crates/services/tx_status_manager/src/service.rs
```diff
@@ -94,41 +108,143 @@ impl SharedData {
...
+
+    async fn add_new_delegate(
+        &mut self,
+        sealed: &Sealed<DelegatePreConfirmationKey<DelegatePublicKey>, ProtocolSignature>,
+    ) -> bool {
+        let Sealed { entity, signature } = sealed;
+        let bytes = postcard::to_allocvec(&entity).unwrap();
+        let message = Message::new(&bytes);
+        let expected_address = self.protocol_pubkey.latest_address();
+        // let verified = signature.verify(&pubkey, &message);
```
xgreenx, line 172:
```suggestion
```

crates/services/tx_status_manager/src/service.rs
```diff
@@ -94,41 +108,143 @@ impl SharedData {
...
+            .map_or(false, |pubkey| Input::owner(&pubkey) == expected_address);
+        self.remove_expired_delegates();
+        if verified {
+            self.delegate_keys
+                .insert(entity.expiration, entity.public_key);
+        };
+        verified
+    }
+
+    async fn check_preconfirmation_signature(
```
xgreenx, line 184:
```suggestion
    fn check_preconfirmation_signature(
```

crates/services/tx_status_manager/src/service.rs
```diff
@@ -94,41 +108,143 @@ impl SharedData {
...
+            }
+        }
+    }
+
+    fn remove_expired_delegates(&mut self) {
+        let now = Tai64::now();
+        self.delegate_keys.retain(|exp, _| exp > &now);
+    }
+
+    async fn add_new_delegate(
```
xgreenx, line 164:
```suggestion
    fn add_new_delegate(
```

crates/services/tx_status_manager/src/service.rs
```diff
@@ -94,41 +108,143 @@ impl SharedData {
...
+        let Sealed { entity, .. } = sealed;
+        entity.preconfirmations.into_iter().for_each(
+            |Preconfirmation { tx_id, status }| {
+                let status: TransactionStatus = status.into();
+                self.manager.status_update(tx_id, status);
+            },
+        );
+    }
+
+    async fn new_preconfirmations_from_p2p(
```
xgreenx, line 216:
```suggestion
    fn new_preconfirmations_from_p2p(
```

crates/services/tx_status_manager/src/service.rs
```diff
@@ -94,41 +108,143 @@ impl SharedData {
...
-                    |Preconfirmation { tx_id, status }| {
-                        let status: TransactionStatus = status.into();
-                        self.manager.status_update(tx_id, status);
-                    },
+            PreConfirmationMessage::Delegate(sealed) => {
+                tracing::debug!(
+                    "Received new delegate signature from peer: {:?}",
+                    sealed.entity.public_key
                 );
+                let _ = self.signature_verification.add_new_delegate(&sealed).await;
```
xgreenx, line 226:
```suggestion
                let _ = self.signature_verification.add_new_delegate(&sealed);
```

crates/services/tx_status_manager/src/service.rs
```diff
@@ -94,41 +108,143 @@ impl SharedData {
...
+            }
+            PreConfirmationMessage::Preconfirmations(sealed) => {
+                tracing::debug!("Received new preconfirmations from peer");
+                if self
+                    .signature_verification
+                    .check_preconfirmation_signature(&sealed)
+                    .await
```
xgreenx, line 233:
```suggestion
                if self
                    .signature_verification
                    .check_preconfirmation_signature(&sealed)
```

xgreenx (approved):
Thank you for the change, very clear tests!=)

bin/fuel-core/src/cli/run.rs
MitchTurner: Ah. But the new delegate keys will all be signed by this key.

AurelienFT: Ah yeahhhh right it didn't connected to my brain. Ty :)

Issue #2860, opened 2025-03-12
Use outputs from the pre-confirmation status to update TxPool sooner


AurelienFT:
Now that we have pre-confirmations created by the executor (optionally) TxPool can take advantage of this to update his state earlier.

PR #2861, opened 2025-03-12
Add function to update txpool on preconfirmation

AurelienFT:
## Linked Issues/PRs
Resolve https://github.com/FuelLabs/fuel-core/issues/2860

## Description
When a pre-confirmation status has arrived to the pool the pool can take action to updates some states

## Checklist

### Before requesting review


crates/services/txpool_v2/src/pool_worker.rs
```diff
@@ -133,16 +134,16 @@ impl PoolWorkerInterface {
...
     }
 
-    pub fn remove_tx_and_coin_dependents(
+    pub fn expired_transactions(
         &self,
         tx_and_dependents_ids: (Vec<TxId>, Error),
     ) -> anyhow::Result<()> {
         self.request_remove_sender
-            .try_send(PoolRemoveRequest::TxAndCoinDependents {
+            .try_send(PoolTxUpdateRequest::ExpiredTransactionsAndDependents {
```
rymnc, line 146: nice, its more clear now why its removed

crates/services/txpool_v2/src/pool_worker.rs
```diff
@@ -291,20 +296,23 @@ where
...
+                        PoolTxUpdateRequest::SkippedTransactions { dependents_ids } => {
                             self.remove_skipped_transactions(dependents_ids);
                         }
-                        PoolRemoveRequest::TxAndCoinDependents { tx_and_dependents_ids } => {
-                            self.remove_and_coin_dependents(tx_and_dependents_ids);
+                        PoolTxUpdateRequest::ExpiredTransactionsAndDependents { tx_and_dependents_ids } => {
+                            self.expired_transactions_and_dependents(tx_and_dependents_ids);
+                        }
+                        PoolTxUpdateRequest::PreconfirmedTransactions { txs } => {
+                            self.preconfirmed_transactions(txs);
```
rymnc, line 315: can we pass an iterator instead?
```suggestion
                            self.preconfirmed_transactions(txs.iter());
```

crates/services/txpool_v2/src/pool_worker.rs
```diff
@@ -533,6 +542,31 @@ where
         }
     }
 
+    fn preconfirmed_transactions(&mut self, txs: Vec<(TxId, PreconfirmationStatus)>) {
```
rymnc, line 545:
```suggestion
    fn preconfirmed_transactions(&mut self, txs: impl Iterator<Item = &(TxId, PreconfirmationStatus)>) {
```
perhaps?

crates/services/txpool_v2/src/pool_worker.rs
```diff
@@ -516,7 +525,7 @@ where
         }
     }
 
-    fn remove_and_coin_dependents(&mut self, tx_ids: (Vec<TxId>, Error)) {
+    fn expired_transactions_and_dependents(&mut self, tx_ids: (Vec<TxId>, Error)) {
```
rymnc, line 528: i think we can take an iterator here too, no?

crates/services/txpool_v2/src/shared_state.rs
```diff
@@ -34,7 +37,7 @@ use crate::{
 
 #[derive(Clone)]
 pub struct SharedState {
-    pub(crate) request_remove_sender: mpsc::Sender<PoolRemoveRequest>,
+    pub(crate) request_remove_sender: mpsc::Sender<PoolTxUpdateRequest>,
```
rymnc, line 40:
```suggestion
    pub(crate) request_update_sender: mpsc::Sender<PoolTxUpdateRequest>,
```

crates/services/txpool_v2/src/shared_state.rs
```diff
@@ -158,7 +161,16 @@ impl SharedState {
...
-            .try_send(PoolRemoveRequest::SkippedTransactions { dependents_ids })
+            .try_send(PoolTxUpdateRequest::SkippedTransactions { dependents_ids })
+        {
+            tracing::error!("Failed to send remove coin dependents request: {}", e);
+        }
+    }
+
+    pub fn notify_preconfirmed_txs(&self, txs: Vec<(TxId, PreconfirmationStatus)>) {
+        if let Err(e) = self
+            .request_remove_sender
```
rymnc, line 172:
```suggestion
            .request_update_sender
```

crates/services/txpool_v2/src/shared_state.rs
```diff
@@ -158,7 +161,16 @@ impl SharedState {
 
         if let Err(e) = self
             .request_remove_sender
```
rymnc, line 163:
```suggestion
            .request_update_sender
```

crates/services/txpool_v2/src/pool_worker.rs
```diff
@@ -58,7 +59,7 @@ const SIZE_THREAD_MANAGEMENT_CHANNEL: usize = 10;
 pub(super) struct PoolWorkerInterface {
     thread_management_sender: Sender<ThreadManagementRequest>,
     pub(super) request_insert_sender: Sender<PoolInsertRequest>,
-    pub(super) request_remove_sender: Sender<PoolRemoveRequest>,
+    pub(super) request_remove_sender: Sender<PoolTxUpdateRequest>,
```
rymnc, line 62:
```suggestion
    pub(super) request_update_sender: Sender<PoolTxUpdateRequest>,
```

crates/services/txpool_v2/src/pending_pool.rs
```diff
@@ -229,16 +229,13 @@ impl PendingPool {
     }
 
     // We expect it to be called a lot (every new tx in the pool).
-    pub fn new_known_tx(
+    pub fn new_known_tx<'a>(
         &mut self,
-        new_known_tx: ArcPoolTx,
+        new_known_tx_outputs: impl Iterator<Item = &'a Output>,
```
rymnc, line 234: perhaps this `Output` should be boxed? 
difficult to say without knowing average size of outputs that we have, but in the case the user has 255 outputs this would be overusing the cache and move other imp code out of the cpu cache

since we have a limit on the total number of outputs, perhaps we can use smallvec?

rymnc: (i only write this because of the comment that says "we expect it to be called a lot")

crates/services/txpool_v2/src/pool_worker.rs
```diff
@@ -197,14 +198,17 @@ pub(super) enum PoolExtractBlockTransactions {
     },
 }
 
-pub(super) enum PoolRemoveRequest {
+pub(super) enum PoolTxUpdateRequest {
+    PreconfirmedTransactions {
```
xgreenx, line 202: I think instead of doing that, the transaction pool should subscribe for all pre-confirmation that comes from the `TxStatusUpdater`. 

In that case you also don't need `SkippedTransactions`, because you will receive `PreConfirmationSqueezedOut`.

crates/services/txpool_v2/src/extracted_outputs.rs
```diff
@@ -88,6 +88,35 @@ impl ExtractedOutputs {
...
+                    asset_id,
+                } => {
+                    self.coins_created
+                        .entry(tx_id)
+                        .or_default()
+                        .insert(UtxoId::new(tx_id, 0), (*to, *amount, *asset_id));
+                }
+                Output::Contract { .. }
+                | Output::Change { .. }
+                | Output::Variable { .. } => {
```
xgreenx, line 113: Resolved outputs only contain `Change` and `Variable` outputs=D

xgreenx: BTW, we need to make a note to add an integration test to check that we can submit transaction that uses `Change` or `Variable` output form its parent after receiving a preconfiramtion

crates/services/txpool_v2/src/pool_worker.rs
AurelienFT: I don't like the type transformation it requires to write to make this works when calling this function because of the tuple, I would pass this one.

AurelienFT:
Closing in favor of https://github.com/FuelLabs/fuel-core/pull/2882 to address @xgreenx comment. Thanks for the review @rymnc I adressed your comments in the new PR.

PR #2859, opened 2025-03-12
fault_proving(compression): glue code for integ into fuel-core

rymnc:
> [!NOTE]
> This PR introduces **no breaking changes** to the graphql api

## Linked Issues/PRs

closes https://github.com/FuelLabs/fuel-core/issues/2776
closes https://github.com/FuelLabs/fuel-core/issues/2778

## Description

ai generated desc:
integrating a new compression service and cleaning up the codebase by removing deprecated compression via OCW. The most important changes include adding the `fuel-core-compression-service` dependency, updating the `CombinedDatabase` struct and its methods to include the new compression database, and removing deprecated DA compression-related code.

### Integration of Compression Service:
* [`Cargo.toml`](diffhunk://#diff-7fc200bcb8e651840f8692da4101cfcdba429acc82847ed81d1d0ed63e0a5bd2L24-R24): Added `fuel-core-compression-service` dependency and removed `fuel-core-compression` dependency. [[1]](diffhunk://#diff-7fc200bcb8e651840f8692da4101cfcdba429acc82847ed81d1d0ed63e0a5bd2L24-R24) [[2]](diffhunk://#diff-7fc200bcb8e651840f8692da4101cfcdba429acc82847ed81d1d0ed63e0a5bd2L52) [[3]](diffhunk://#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R75)
[...]

PR #2875, opened 2025-03-17
fix(storage): custom impl of EnumCount for MerkleizedColumn

rymnc:
## Linked Issues/PRs

- none

## Description

The in memory database fails when trying to access the metadata column ~ the `COUNT`'s were mismatched between the strum::enumCount impl and the actual number of columns. This PR fixes that with a custom impl. tested with https://github.com/FuelLabs/fuel-core/pull/2859.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/storage/src/merkle/column.rs
```diff
@@ -56,7 +63,11 @@ where
             Self::MerkleDataColumn(column) => {
                 Self::MERKLE_DATA_COLUMNS_START.wrapping_add(column.as_u32())
             }
-            Self::MerkleMetadataColumn => u32::MAX,
+            Self::MerkleMetadataColumn => Self::COUNT
+                .checked_sub(1)
+                .expect("qed")
+                .try_into()
+                .expect("qed"),
```
netrome, line 70: Hmm.. I don't like having panics here but I don't see an obvious way around it. First of all, I think "qed" is an unhelpful message so instead we can just unwrap the values.

crates/storage/src/merkle/column.rs
```diff
@@ -56,7 +63,11 @@ where
             Self::MerkleDataColumn(column) => {
                 Self::MERKLE_DATA_COLUMNS_START.wrapping_add(column.as_u32())
             }
```
netrome, line 65: I think we need to fix this one as well. Seems like the in-memory database implementation assumes these values are contiguous.

crates/storage/src/merkle/column.rs
```diff
@@ -31,6 +30,16 @@ pub enum MerkleizedColumn<TC> {
...
+impl<TC> strum::EnumCount for MerkleizedColumn<TC>
+where
+    TC: strum::EnumCount + AsU32,
+{
+    /// The total count of variants in the enum.
+    /// Since we have two columns for each table column and one for the merkle data,
+    /// we have to multiply the count of the table columns by 2 and add one for the merkle metadata.
+    const COUNT: usize = TC::COUNT * 2 + 1;
+}
+
```
netrome, line 42: What's the benefit of breaking this out?

crates/storage/src/merkle/column.rs
netrome:
```suggestion
            Self::MerkleMetadataColumn => Self::COUNT
                .checked_sub(1)
                .unwrap()
                .try_into()
                .unwrap()
```

netrome (dismissed):
I'm a bit torn here. It seems like the purpose of the fix is to ensure the column u32 representations are contiguous. If we want to achieve that, we should also update the representation of the `MerkleDataColumn`s.

However, I'd argue that this is an error on the in-memory database implementation. The best solution here would be to rework the in-memory database so that it can handle non-contiguous column representations.

That being said, by my own argument, nothing prevents us from changing the column representation to be more compatible with the current in-memory implementation. So I have nothing against redefining the column representation as in this PR, hence approved (though I'd prefer to see the merkle data column being updated and the expects changed to unwraps at least).

crates/storage/src/merkle/column.rs
rymnc: it's not a benefit, but the core part of the fix lol

crates/storage/src/merkle/column.rs
rymnc: nop, we don't need to fix these. the in memory db creates the columns based on the total count. anything in between is fine

crates/storage/src/merkle/column.rs
netrome: Oh I see from the description. we need the trait impl. Alright, fair enough 👍

crates/storage/src/merkle/column.rs
netrome: But `u16::MAX` isn't in-between though, or am I missing something here? `Self::MERKLE_DATA_COLUMNS_START` is a constant set to `u16::MAX` from what I see.

netrome (dismissed):

crates/storage/src/merkle/column.rs
```diff
@@ -56,7 +63,9 @@ where
             Self::MerkleDataColumn(column) => {
                 Self::MERKLE_DATA_COLUMNS_START.wrapping_add(column.as_u32())
             }
-            Self::MerkleMetadataColumn => u32::MAX,
+            Self::MerkleMetadataColumn => {
+                Self::COUNT.checked_sub(1).unwrap().try_into().unwrap()
+            }
```
netrome, line 68: Hmm it could be worth to try if we could replace this with a compile-time evaluated expression. Don't know by heart how that would look, but we could get rid of the run-time unwraps that way.

rymnc: i think its overkill, but did it anyway - 50ad25626a

netrome: Niiice! Looks great 🤩 I think this is significantly better. Although the unwraps are technically safe, it still introduces a (very small but non-zero) risk that future bugs accidentally trigger this panic during runtime, thus causing downtime. Now, that risk is instead eliminated.

netrome (dismissed):

Dentosal (dismissed):

rymnc:
needs merge after https://github.com/FuelLabs/fuel-core/pull/2877

crates/storage/src/merkle/column.rs
```diff
@@ -42,21 +51,30 @@ where
...
+    /// The merkle metadata column
+    pub const MERKLE_METADATA_COLUMN: u32 = {
+        assert!(Self::COUNT <= u32::MAX as usize);
+        // this is fine, because we already performed an assertion above
+        // see https://github.com/rust-lang/rust-clippy/issues/9613
+        #[allow(clippy::cast_possible_truncation)]
+        let column_index = (Self::COUNT as u32).saturating_sub(1);
+        assert!(column_index > 0);
+        column_index
+    };
```
xgreenx, line 68: If you do it like this, it means that column id depends on the `Self::COUNT`. So if you add a new column, it will override the metadata column. Metadata column always should be stable regardless of the number of columns in the type.

rymnc: addressed in aa438a2295

Dentosal (approved):

netrome (changes requested):
If we set the merkle metadata column at zero, we need to update the `CompressionColumn` in `crates/services/compression/src/storage/column.rs` to avoid a conflict. We also have to do the same for `TableColumn` in the state root service, but since that's another repo we'll have to do it separately.

PR #2884, opened 2025-03-18
fix(ci): check-changelog should run only if its a PR

rymnc:
`cargo-verifications` depends on `check-changelog`, but we skip it in non-pr's now, so we still need a way to run this because crate publish depends on `cargo-verifications`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,13 +33,16 @@ jobs:
   check-changelog:
     name: Check Changelog
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


AurelienFT (commented):
Maybe just remove the need here : https://github.com/FuelLabs/fuel-core/blob/9befd8879150be00f08dfc369619076d755dc33c/.github/workflows/ci.yml#L121

AurelienFT (approved):

Dentosal (approved):

xgreenx (approved):

PR #2887, opened 2025-03-19
chore: Bump rust version to 1.85.0

### LINKED ISSUE #2607: Bump rust version to 1.85.0
~~In the same time revert : https://github.com/FuelLabs/fuel-core/pull/2608~~

netrome:
Closes #2607


netrome:
Had an intense debugging session but I've managed to identify the issue (with some GPT help and wasm-objdump inspections). The new compiler seems to add multi-byte LEB128-encoded zeros to call_indirect table indices, which is legal in the WASM spec as I understand it but causes problems with older runtimes.

I've been able to patch the binary to make the executor::test::wasm::can_validate_block__wasm_strategy test pass. I'll do the same to verify it's the samme issue in the version_compatibility tests. After that, I'll sync with the team to brainstorm the best way forward.

An intermediate solution would be to add this binary patching to our build, but it's worth thinking about how we want this to be stable long-term.

netrome:
> Had an intense debugging session but I've managed to identify the issue (with some GPT help and wasm-objdump inspections). The new compiler seems to add multi-byte LEB128-encoded zeros to call_indirect table indices, which is legal in the WASM spec as I understand it but causes problems with older runtimes.
> 
> I've been able to patch the binary to make the executor::test::wasm::can_validate_block__wasm_strategy test pass. I'll do the same to verify it's the samme issue in the version_compatibility tests. After that, I'll sync with the team to brainstorm the best way forward.
> 
> An intermediate solution would be to add this binary patching to our build, but it's worth thinking about how we want this to be stable long-term.

We discussed this in the team last week and agreed that adding a post-processing step to our build wasn't a horrible solution. I'll brush this PR up now and hopefully we can push this through.

.github/workflows/ci.yml
```diff
@@ -185,6 +185,8 @@ jobs:
           toolchain: ${{ env.RUST_VERSION }}
           targets: "wasm32-unknown-unknown"
           components: "clippy"
+      - name: Install wasm-tools
+        run: cargo install wasm-tools@1.228.0
```
Voxelot, line 189: Consider using install-action instead for a bug speed up over cargo install

https://github.com/taiki-e/install-action

netrome: I considered that but `wasm-tools` isn't supported https://github.com/taiki-e/install-action/blob/main/TOOLS.md. But we could perhaps contribute upstream to add it. Though I would save that for a follow-up.

crates/client/src/client/schema/message.rs
```diff
@@ -159,8 +159,6 @@ pub struct MessageProofArgs {
     /// The `Nonce` identifier of the output message that requires a proof.
     pub nonce: Nonce,
 
-    /// The query supports either `commit_block_id`, or `commit_block_height` set on, not both.
```
Voxelot, line 162: Why is this comment being removed?

.cargo/config.toml
```diff
@@ -8,3 +8,8 @@ protocol = "sparse"
 # https://github.com/rust-lang/cargo/issues/4133
 [target.aarch64-unknown-linux-gnu]
 linker = "aarch64-linux-gnu-gcc"
+
+[target.wasm32-unknown-unknown]
+rustflags = [
+  "-C", "target-cpu=mvp"
+]
```
xgreenx, line 15: It means everyone who builds our crates should specify that, otherwise WASM executor will not work.

.changes/breaking/2887.md
```diff
@@ -0,0 +1,4 @@
+Bump Rust version to `1.85.0`
+
+Also added a step using `wasm-tools` when building the `wasm-executor`.
+We use it to ensure the WASM binaries have canonical LEB128 encodings.
```
xgreenx, line 4: It forces everyone who uses `fuel-core` with `wasm-executor` feature to install additional dependency. We don't want to force people to install something that potentially could be avoided.

xgreenx (commented):
I think it is fine to break release of `fuel-core 0.44.0` and require people to upgrade their `fuel-core` to get a new WASM executor. Because we still want to support this feature to get full advantage of the LLVM and optimizations for the execution.

I applied changes in the https://github.com/FuelLabs/fuel-core/pull/2929

netrome:
> I think it is fine to break release of `fuel-core 0.44.0` and require people to upgrade their `fuel-core` to get a new WASM executor. Because we still want to support this feature to get full advantage of the LLVM and optimizations for the execution.
> 
> I applied changes in the #2929

Yeah if we can break forward compatibility then that's muuuch better. This was the best solution I could find that didn't break it. Applying your changes.

.cargo/config.toml
netrome: Yep, thanks for changing it. Nice to not have to have this.

.github/workflows/ci.yml
netrome: Alright so now we're skipping this entirely at the cost of breaking forward compatibility for the nodes. So we can skip this extra CI bloat.

AurelienFT (approved):
Looks good to me ! I'm so hyped !!! Thank you for taking the time to do that.

AurelienFT (approved):

xgreenx (approved):

PR #2888, opened 2025-03-19
Improve `TxStatusManager` tests coverage

### LINKED ISSUE #2838: Complete TxStatusManager Service tests
Implement the missing tests:
```rs
// Tests to implement
// 1. `status_update()` sends correct notifications for `tx_status_change` and
//    `new_tx_notification_sender` depending on the `TransactionStatus` variant.
// 2. `status_update()` updates the `statuses` map with the correct `TransactionStatus`.
//     for example `status_update()` does not update the `statuses` map
//     with the `TransactionStatus::SqueezedOut` (if we decide so)
// 3. Pruning of old statuses
//    - upon storage in DB
//    - other pruning scenarios if we decide to implement them
// 4. `status_update()` is called from all components: TxPool, Consensus, WorkerService and P2P
// 5. Signature is verified before the preconfirmations are accepted
```

rafal-ch:
Closes #2838 

## Description
This PR adds some more coverage to the `TxStatusManager`. Details are available in [this comment](https://github.com/FuelLabs/fuel-core/issues/2838#issuecomment-2736390453).

The two outstanding `TODO`s should be re-evaluated after we are ready with the ultimate integration test for preconfirmations, because it's likely that this test will cover such cases.

### Before requesting review


AurelienFT:
Close in favor of https://github.com/FuelLabs/fuel-core/pull/2911 it avoid CLA problems and I addressed some comments

PR #2892, opened 2025-03-20
Add public api tests for tx status manager crate

rafal-ch:
## Description
This PR adds a test that prevents accidental public api changes in the `tx_status_manager` crate. If this makes sense we can generalize it to all crates.

###
It works by generating insta snapshots from the public api documentation and then doing the comparison.

For example, for the initial behavior we see
```
$ cargo test --all-features
   ...
test result: ok. 31 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.31s
```

then we change something in the non-public API:
```
-    fn prune_old_statuses(&mut self) {
+    fn prune_old_statuses(&mut self, foo: u32) {
```

and still the tests are ok:
```
$ cargo test --all-features
   ...
test result: ok. 31 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.31s
```

but if we change anything that is publicly exposed:
```
impl SharedData {
     pub async fn get_status(
         &self,
-        tx_id: TxId,
+        tx_id: Option<TxId>,
     ) -> anyhow::Result<Option<TransactionStatus>> {
```

we're gonna get the test failure and `cargo insta review` will be needed.

image


rymnc:
hm, but sometimes we want to do breaking changes, so how would it work then? i.e, how do we override it?

rafal-ch:
> hm, but sometimes we want to do breaking changes, so how would it work then? i.e, how do we override it?

You do the same as for the GraphQL snapshots, i.e. call `cargo insta review`, review the changes and apply.

rymnc:
> > hm, but sometimes we want to do breaking changes, so how would it work then? i.e, how do we override it?
> 
> You do the same as for the GraphQL snapshots, i.e. call `cargo insta review`, review the changes and "accept" them.

this solution is awesome!! although i think it needs a little more. we should have a ci check that when it detects a snapshot changed, it should add the `breaking` label automatically. this will make things a little more holistic. love the initial approach though, so happy to approve this

rymnc (approved):
would prefer the deps to be moved to workspace root cargo.toml instead if we plan to use elsewhere

rafal-ch:
> would prefer the deps to be moved to workspace root cargo.toml instead if we plan to use elsewhere

Certainly, if we decide this is good to go for other parts I'll refine the implementation. Let's wait for some more comments.

netrome (approved):
Love it!

xgreenx (commented):
What is the compilation time of it?

AurelienFT:
Close in favor of #2913 because of CLA. Fixed CI there.

PR #2900, opened 2025-03-24
refactor: Get rid of `Deref` impl on `ImportResult` by introducing wrapper type.

### LINKED ISSUE #2899: Remove `Deref` implementation for `ImportResult`
Spawned from PR thread: https://github.com/FuelLabs/fuel-core/pull/2895#discussion_r2009846285

### Context
Currently, `ImportResult` dereferences to itself:

```rust
impl Deref for ImportResult {
    type Target = Self;

    fn deref(&self) -> &Self::Target {
        self
    }
}
```

This causes infinite recursion errors whenever any type that dereferences to `ImportResult` (such as an `Arc<ImportResult>`) implements trait methods. See https://github.com/FuelLabs/fuel-core/actions/runs/14033853348/job/39287338546?pr=2895 for example.

I don't see why we'd want to dereference the type to itself. @xgreenx I see you implemented this, so feel free to chime in if there's any reason to keep this around. Otherwise I assume it's an old workaround that would be nice to get rid of.

### Definition of Done
The `Deref` implementation for `ImportResult` is removed.

netrome:
closes #2899 

See `crates/types/src/services/block_importer.rs` for the main change. The rest of the PR is just adapting the code to use explicit dereferencing to access the relevant fields.

### DIFF
--- a/.changes/breaking/2900.md
+++ b/.changes/breaking/2900.md
@@ -1,0 +1,1 @@
+Get rid of `Deref` impl on `ImportResult` by introducing wrapper type.
--- a/crates/fuel-core/src/graphql_api/worker_service/tests.rs
+++ b/crates/fuel-core/src/graphql_api/worker_service/tests.rs
@@ -65,7 +65,7 @@ fn block_importer_for_event(event: Event) -> BoxStream<SharedImportResult> {
         tx_status: vec![],
         events: vec![event],
         source: Default::default(),
-    });
+    }.wrap());
     let blocks: Vec<SharedImportResult> = vec![block];
     tokio_stream::iter(blocks).into_boxed()
 }
--- a/crates/fuel-core/src/graphql_api/worker_service/tests.rs
+++ b/crates/fuel-core/src/graphql_api/worker_service/tests.rs
@@ -60,12 +60,15 @@ async fn run__relayed_transaction_events_are_added_to_storage() {
 }
 
 fn block_importer_for_event(event: Event) -> BoxStream<SharedImportResult> {
-    let block = Arc::new(ImportResult {
-        sealed_block: Default::default(),
-        tx_status: vec![],
-        events: vec![event],
-        source: Default::default(),
-    }.wrap());
+    let block = Arc::new(
+        ImportResult {
+            sealed_block: Default::default(),
+            tx_status: vec![],
+            events: vec![event],
+            source: Default::default(),
+        }
+        .wrap(),
+    );
     let blocks: Vec<SharedImportResult> = vec![block];
     tokio_stream::iter(blocks).into_boxed()
 }
--- a/crates/fuel-core/src/service/adapters/chain_state_info_provider.rs
+++ b/crates/fuel-core/src/service/adapters/chain_state_info_provider.rs
@@ -367,7 +367,7 @@ mod tests {
             sealed_block,
             Default::default(),
             Default::default(),
-        ))
+        ).wrap())
     }
 
     #[tokio::test]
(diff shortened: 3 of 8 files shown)


benches/benches/block_target_gas.rs
```diff
@@ -448,13 +450,27 @@ fn run_with_service_with_extra_inputs(
...
+                    res.deref()
+                        .deref()
+                        .as_ref()
+                        .sealed_block
+                        .entity
+                        .transactions()
+                        .len(),
+                    2
+                );
+                assert_eq!(res.deref().deref().as_ref().tx_status[0].id, tx_id);
```
rafal-ch, line 469:
```suggestion
                assert_eq!(res.as_ref().as_ref().tx_status[0].id, tx_id);
```

Also applies to other, similar places.

crates/fuel-core/src/service/adapters/compression_adapters.rs
```diff
@@ -21,8 +23,8 @@ impl block_source::BlockSource for BlockImporterAdapter {
         use futures::StreamExt;
         self.events_shared_result()
             .map(|result| {
-                let sealed_block = result.sealed_block.clone();
-                let events = result.events.clone();
+                let sealed_block = result.deref().as_ref().sealed_block.clone();
```
rafal-ch, line 26:
```suggestion
                let sealed_block = (*result).as_ref().sealed_block.clone();
```

Explicit `.deref()` makes my eyes hurt :)

.changes/breaking/2900.md
```diff
@@ -0,0 +1 @@
+Use `dyn AsRef<ImportResult>` instead of `dyn Deref<ImportResult>` in `SharedImporterResult`.
```
rafal-ch, line 1:
```suggestion
Use `dyn AsRef<ImportResult>` instead of `dyn Deref<ImportResult>` in `SharedImportResult`.
```

rafal-ch (dismissed):

rymnc (commented):
comment shared on slack

netrome:
> comment shared on slack

Thanks. I'll update with a revamped version of this PR tomorrow morning based on your proposal.

xgreenx (changes requested):
Making `AsRef` definitely not an option. The main idea was to have a nice syntax by keeping the logic used by `Awaiter`.

If you want to solve recursion problem, you can create a new type that will implement `Deref<Target = ImportResult>` and use this type instead of `ImportResult`.

In this case you will not have a recursion

netrome:
> Making `AsRef` definitely not an option. The main idea was to have a nice syntax by keeping the logic used by `Awaiter`.
> 
> If you want to solve recursion problem, you can create a new type that will implement `Deref<Target = ImportResult>` and use this type instead of `ImportResult`.
> 
> In this case you will not have a recursion

Fair enough. After our discussion on Slack I agree that adding a wrapper type is the least intrusive way to get rid of the infinite recursion. I have updated the PR now to use this approach.

.changes/breaking/2900.md
netrome: Outdated now - but thanks anyway!

crates/fuel-core/src/service/adapters/compression_adapters.rs
netrome: Interesting. I prefer the explicit over the *-operator. But I'll give you that the operator is more concise. I'll use it more in the future.

benches/benches/block_target_gas.rs
netrome: Yeah, not relevant anymore but thanks anyway for suggesting. I was just about to push this change when I got the other "use a completely different approach"-suggestions 😅

xgreenx (approved):

rafal-ch (approved):
LGTM :+1:

PR #2901, opened 2025-03-24
Dry run execution tracing

### LINKED ISSUE #2625: Execution tracing: dry run support
Follow-up for #2491.

Storage read replay should be available for dry run as well, and it should cost approximately the same as a normal dry run. It can be available on all nodes without `--debug` flag. At least `forc-call` utility will want this feature.

Dentosal:
Closes #2625.

## Description

Adds a new query `dryRunRecordStorageReads` which works like `dryRun` but also returns storage reads, allowing use of execution tracer or local debugger.

### Alternative approaches

If we're ok with breaking changes, we could modify the existing `dryRun` to do this.

## Checklist

### Before requesting review

- [ ] Someone else?


xgreenx (commented):
Why do we need this feature? Just trying to understand where do we want to use it, and what is the difference with the currently available endpoints

Dentosal:
> Why do we need this feature? Just trying to understand where do we want to use it, and what is the difference with the currently available endpoints

It allows you to receive all storage accesses for a local debugger run in a single request, which improves latency quite a lot. Tools like `for call` can then use this to preview tx execution. It's true that we can achieve this using multiple requests already, and this is not strictly necessary. The execution tracing service can provide the same functionality if we don't want to have it in fuel-core itself, and I won't mind closing this and just doing that instead if you think that the extra complexity here is not worth having.

crates/fuel-core/src/schema/tx/types.rs
```diff
@@ -1059,6 +1060,23 @@ impl DryRunTransactionExecutionStatus {
...
+#[Object]
+impl DryRunStorageReads {
+    async fn tx_statuses(&self) -> Vec<DryRunTransactionExecutionStatus> {
+        self.tx_statuses.clone()
+    }
+
+    async fn storage_reads(&self) -> Vec<StorageReadReplayEvent> {
+        self.storage_reads.clone()
+    }
+}
```
xgreenx, line 1077: Do we really need to clone here? Maybe it is possible to return a reference instead?

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -617,7 +660,9 @@ where
         mode: ProduceBlockMode,
         new_tx_waiter: impl NewTxWaiterPort,
         preconfirmation_sender: impl PreconfirmationSenderPort,
-    ) -> ExecutorResult<Uncommitted<ExecutionResult, Changes>>
+    ) -> ExecutorResult<
+        Uncommitted<(ExecutionResult, Vec<StorageReadReplayEvent>), Changes>,
```
xgreenx, line 664: Would be nice to have some alias, because it is a huge type now=D

xgreenx (dismissed):
Maybe we need to charge user more if he wants to use a new enpdpoint? Just a thought, I don't have any opinion on it

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -858,7 +922,10 @@ where
 
         let _ = preconfirmation_sender.send(preconfirmations).await;
 
-        Ok(result)
+        let mut g = storage_rec.lock();
```
rymnc, line 925: curious, if we `storage_rec` was an Option, and we performed a check like so -
```suggestion
        let storage_reads = match storage_rec {
            Some(storage_rec) => {
                let mut g = storage_rec.lock();
                core::mem::take(&mut *g)
            }
            None => vec![] /* replace with what you actually want / concrete type here */
        }
```

we would avoid the call to `.lock()` altogether. should describe the code better.

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -909,39 +976,76 @@ where
...
+                        mode.is_dry_run(),
+                        new_tx_waiter,
+                        preconfirmation_sender,
+                    )
+                    .await
+            }
+        };
+
+        let mut g = storage_rec.lock();
+        let storage_reads = core::mem::take(&mut *g);
```
rymnc, line 1046: same comment about making `storage_rec` an Option

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -909,39 +976,76 @@ where
...
+            ProduceBlockMode::DryRun { height, .. } => match height {
+                BlockHeightSelection::Latest => None,
+                BlockHeightSelection::Past { height } => height.pred(),
+            },
         };
 
-        if let Some(previous_block_height) = db_height {
+        let mut storage_rec = Default::default();
+
+        let result = if let Some(previous_block_height) = db_height {
```
rymnc, line 997:
```suggestion
        let height = db_height
                .as_ref()
                .map_or_else(
                        || self.storage_view_provider.latest_view(),
                        |h| self.storage_view_provider.view_at(h),
                )?;

        let database = if mode.record_storage_reads() {
                let recorder = StorageAccessRecorder::new(height);
                storage_rec = recorder.record.clone();
                recorder.into()
        } else {
                height
        };

        let result = ExecutionInstance::new(relayer, database, options)
                .produce_without_commit(
                        block,
                        mode.is_dry_run(),
                        new_tx_waiter,
                        preconfirmation_sender,
                )
                .await;
```

would this work instead?

rymnc (commented):
left a few comments :)
since this pr has been open for a while, they can be addressed in a follow up if you'd like. no strong preference from my end, most of the comments are about making some of the code paths a little more clearer

crates/services/upgradable-executor/src/executor.rs
```diff
@@ -783,19 +832,34 @@ where
 
         let db_height = match mode {
             ProduceBlockMode::Produce => block.header_to_produce.height().pred(),
-            ProduceBlockMode::DryRunLatest => None,
-            ProduceBlockMode::DryRunAt { height } => height.pred(),
+            ProduceBlockMode::DryRun { height, .. } => match height {
+                BlockHeightSelection::Latest => None,
+                BlockHeightSelection::Past { height } => height.pred(),
```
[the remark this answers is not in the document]
Dentosal: I considered this as well. It' not many times, and each of them look slightly different, so I don't think this is worth it.

crates/services/upgradable-executor/src/executor.rs
Dentosal: I think the call to the `.lock()` is optimized away anyway. The compiler should see that this is the only reference to the mutex.

crates/services/upgradable-executor/src/executor.rs
Dentosal: No. Database with a recorder has a separate type, and cannot be held in the same variable as a db without recorder.

crates/fuel-core/src/schema/tx/types.rs
Dentosal: Good catch. Some other functions here do have to clone as they wrap the result in a struct that doesn't have a reference, but in this case we can indeed just return a reference.

crates/services/upgradable-executor/src/executor.rs
Dentosal: Sure. Done in 1deb59dfbe.

crates/fuel-core/src/schema/tx/types.rs
Dentosal: 1deb59dfbe

crates/client/src/client.rs
```diff
@@ -733,6 +733,43 @@ impl FuelClient {
...
+            .map(|r| r.dry_run_record_storage_reads)?;
+        let tx_statuses = result
+            .tx_statuses
+            .into_iter()
+            .map(|tx_status| tx_status.try_into().map_err(Into::into))
+            .collect::<io::Result<Vec<_>>>()?;
+        let storage_reads = result
+            .storage_reads
+            .into_iter()
+            .map(|read| read.into())
```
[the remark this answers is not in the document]
Dentosal: 1deb59dfbe

crates/client/src/client/schema/storage_read_replay.rs
```diff
@@ -22,6 +22,17 @@ impl From<StorageReadReplayEvent>
         }
     }
 }
+impl From<fuel_core_types::services::executor::StorageReadReplayEvent>
+    for StorageReadReplayEvent
+{
+    fn from(event: fuel_core_types::services::executor::StorageReadReplayEvent) -> Self {
+        StorageReadReplayEvent {
+            column: event.column.into(),
```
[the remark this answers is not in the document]
rymnc: graphql still supports enums btw, but i feel you would have to duplicate the column enum here. we can ignore for now

xgreenx (approved):

PR #2904, opened 2025-03-25
Rafal/2898 coins to spend updates

### LINKED ISSUE #2898: `CoinsToSpend` query should return selected coins even if the amount cannot be covered
Currently, when the `CoinsToSpend` query cannot find enough coins it returns an error:
```rust
    #[error("the target cannot be met due to no coins available or exceeding the {max} coin limit.")]
    InsufficientCoinsForTheMax {
        asset_id: AssetId,
        collected_amount: u128,
        max: u16,
    },
```

The new behavior should be:
1. add new field to the query, for example: `allow_partial: bool`. If this field is set to `true`, then:
    * do not return an error
    * return coins that we were able to find even if they do not satisfy the query

Definition of Done:
1. `InsufficientCoinsForTheMax` error is never thrown (from both `indexed` and `non-indexed` coins to spend algorithm implementations) if the new flag is set
2. Tests are updated to cover system behavior with this new flag
3. Indexation documentation is updated: https://www.notion.so/fuellabs/Indexation-1932f2293f31806b87c1ef8160d3acd7

rafal-ch:
Closes #2898 

## Description
This PR introduces the new `allow_partial` parameter to the "coins to spend" query. The default value of this parameters is `false` to preserve the old behavior. If set to `true`, the query returns available coins instead of failing when the requested amount is unavailable.

## Checklist
- [X] Breaking changes are clearly marked as such in the PR description and changelog

### Before requesting review

### DIFF
--- a/crates/client/assets/schema.sdl
+++ b/crates/client/assets/schema.sdl
@@ -1099,7 +1099,11 @@ type Query {
 		"""
 		The excluded coins from the selection.
 		"""
-		excludedIds: ExcludeInput
+		excludedIds: ExcludeInput,
+		"""
+		If true, returns available coins instead of failing when the requested amount is unavailable.
+		"""
+		allowPartial: Boolean! = false
 	): [[CoinType!]!]!
 	daCompressedBlock(
 		"""
--- a/crates/fuel-core/src/schema/tx/assemble_tx.rs
+++ b/crates/fuel-core/src/schema/tx/assemble_tx.rs
@@ -105,6 +105,9 @@ impl<'a> AssembleArguments<'a> {
             return Ok(Vec::new());
         }
 
+        // TODO[RC]: Support partial flag in the AssembleTx transaction
+        let allow_partial = false;
+
         let query_per_asset = SpendQueryElementInput {
             asset_id: asset_id.into(),
             amount: (amount as u128).into(),
@@ -117,6 +120,7 @@ impl<'a> AssembleArguments<'a> {
                 owner,
                 &[query_per_asset],
                 &self.exclude,
+                allow_partial,
                 &self.consensus_parameters,
                 remaining_input_slots,
             )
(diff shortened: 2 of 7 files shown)


crates/client/assets/schema.sdl
```diff
@@ -1099,7 +1099,11 @@ type Query {
 		"""
 		The excluded coins from the selection.
 		"""
-		excludedIds: ExcludeInput
+		excludedIds: ExcludeInput,
+		"""
+		If true, returns available coins instead of failing when the requested amount is unavailable.
+		"""
+		allowPartial: Boolean! = false
```
xgreenx, line 1106: I think it should be optional and part of the `SpendQueryElementInput`. In this case user can specify which one can be filled partially, and which one can't.

crates/fuel-core/src/schema/tx/assemble_tx.rs
```diff
@@ -105,6 +105,9 @@ impl<'a> AssembleArguments<'a> {
             return Ok(Vec::new());
         }
 
+        // TODO[RC]: Support partial flag in the AssembleTx transaction
+        let allow_partial = false;
```
xgreenx, line 109: When we fetch required balances, it should be `false`, but when we fetch coins to cover the fee, it can be `true`

AurelienFT:
Close in favor of https://github.com/FuelLabs/fuel-core/pull/2912 because of CLA.

crates/fuel-core/src/schema/tx/assemble_tx.rs
AurelienFT: Addressed in https://github.com/FuelLabs/fuel-core/pull/2912

crates/client/assets/schema.sdl
AurelienFT: Addressed in https://github.com/FuelLabs/fuel-core/pull/2912

PR #2912, opened 2025-03-31
Add allow_partial parameter to the "coins to spend" query

AurelienFT:
## Linked Issues/PRs
Closes https://github.com/FuelLabs/fuel-core/issues/2898

THIS IS A REVIVAL OF https://github.com/FuelLabs/fuel-core/pull/2904 WITH REVIEW COMMENTS ADDRESSED (because of CLA). Inspiration from Rafal. Kudos to @rafal-ch !

## Description
This PR introduces the new allow_partial parameter to the "coins to spend" inputs query. This field is optional to avoid breaking change. If set to true, the query returns available coins instead of failing when the requested amount is unavailable.

I consider this change "non-breaking" since the new parameter is not strictly required when sending the query.

## Checklist

### Before requesting review


xgreenx (approved):

crates/fuel-core/src/coins_query.rs
```diff
@@ -709,6 +719,71 @@ mod tests {
...
+            use crate::{
+                coins_query::tests::{
+                    largest_first::query,
+                    setup_coins,
+                },
+                query::asset_query::AssetSpendTarget,
+            };
+
+            #[tokio::test]
+            async fn largest_first__error_when_not_enough_coins_and_allow_partial_false()
```
MitchTurner, line 733: nit:
```suggestion
            async fn query__error_when_not_enough_coins_and_allow_partial_false()
```
I'm not really sure what the `largest_first` is referring to here`

crates/fuel-core/src/coins_query.rs
```diff
@@ -709,6 +719,71 @@ mod tests {
...
+                    &db.service_database(),
+                )
+                .await;
+
+                // Then
+                assert!(coins.is_err());
+            }
+
+            #[tokio::test]
+            async fn largest_first__ok_when_not_enough_coins_and_allow_partial_true() {
```
MitchTurner, line 760: nit:
```suggestion
            async fn query__ok_when_not_enough_coins_and_allow_partial_true() {
```

MitchTurner (commented):
LGTM. Just a couple small things, then I'll approve.

MitchTurner (dismissed):

xgreenx (changes requested):
Oh, wait. The change only covers the `random_improve` and case of the non indixation. We also need to cover the part when we have indexed balances

AurelienFT:
@xgreenx all the cases should be covered now.

crates/fuel-core/src/coins_query.rs
```diff
@@ -277,12 +287,12 @@ pub async fn select_coins_to_spend(
     let (selected_big_coins_total, selected_big_coins) =
         big_coins(big_coins_stream, adjusted_total, max, exclude).await?;
 
-    if selected_big_coins_total < total {
+    if selected_big_coins_total < total && !allow_partial {
```
xgreenx, line 290: I think if `selected_big_coins_total` is zero we still should return an error

crates/fuel-core/src/coins_query.rs
```diff
@@ -709,6 +720,70 @@ mod tests {
             let (owner, asset_ids, base_asset_id, db) = setup_coins_and_messages();
             multiple_assets_helper(owner, &asset_ids, &base_asset_id, db).await;
         }
+
+        mod allow_partial {
```
xgreenx, line 724: It also would be nice to add tests for the case when indexation is available

crates/fuel-core/src/coins_query.rs
AurelienFT: Done and also modified in the other function

crates/fuel-core/src/coins_query.rs
AurelienFT: I added the tests

xgreenx (approved):

crates/fuel-core/src/coins_query.rs
```diff
@@ -1409,6 +1493,7 @@ mod tests {
                 TOTAL,
                 MAX,
                 &AssetId::default(),
+                false,
```
rymnc, line 1496: i think we've reached a point with this function that it is no longer maintainable / readable due to the number of arguments. this can be a follow up / tech debt. we should investigate a better way to structure it.

rymnc (approved):

crates/fuel-core/src/coins_query.rs
AurelienFT: https://github.com/FuelLabs/fuel-core/issues/2930

PR #2914, opened 2025-03-31
feat: add tests for sparse and merkleized blueprint proof generation

### LINKED ISSUE #2907: Add tests for `Sparse` and `Merkleized` blueprint proof generation
### Context
Currently, the `blueprint::merkleized` and `blueprint::sparse` modules provide macros to test basic operations, including insertions, deletions and root computations. However, as far as I can see we're lacking any tests for proof generation on this level.

While the underlying merkle tree implementations in the `fuel-merkle` crate are well tested, I'd find tests at this level useful for two reasons:

1. Ensure proof computations through the blueprints work as expected.
2. Provide examples of how to get the proofs from the blueprints.

Therefore, I propose we extend these tests with some proof generation and validation scenarios.

### Definition of done
We have tests covering proof generation and validation for `Sparse` and `Mekleized` blueprints.

netrome:
Closes #2907 

This PR does two things:

1. The bulk of this PR is a pure refactor of the `basic_merkleized_storage_tests` and the `root_storage_tests` macros. Instead of having the tests defined within the macros, the tests themselves are broken out to a separate helper trait (one for each module). This provides better type safety and makes it easier to maintain and develop the tests.
2. Secondly, this PR adds tests for proof generation and verification for the merkleized and SMT blueprints. See the `test_can_generate_and_validate_proofs` tests in the `merklized.rs` and `sparse.rs` blueprint impls. These tests provide examples of how proof generation and validation works for the respective merkle tree implementations.


crates/storage/src/blueprint/merklized.rs
```diff
@@ -305,260 +356,469 @@ where
...
-                rand::distributions::Standard: rand::distributions::Distribution<T>,
-                R: rand::Rng,
-            {
-                use rand::Rng;
-                rng.gen()
-            }
+/// A trait that provides basic tests for the merklized storage.
+/// It is used to test the merklized storage with different key and value codecs.
+#[allow(warnings)]
+pub mod basic_tests {
```
rymnc, line 363:
```suggestion
pub mod basic_tests_bmt {
```

rymnc (commented):
looks good

crates/storage/src/blueprint/merklized.rs
netrome: Done in https://github.com/FuelLabs/fuel-core/pull/2914/commits/2769a274cef36534cf628d7da82c476b174ac578

netrome:
> looks good

Nice, I'll proceed with the SMT tests then ☺️

netrome:
After chatting with @xgreenx about this, we agreed that it would be less intrusive to only modify the tests for the refactor and not modify the `TableWithBlueprint` impl blocks to use blanket implementations through wrapper traits.

crates/storage/src/blueprint/sparse.rs
```diff
@@ -484,237 +484,483 @@ where
...
-            use super::*;
-            use $crate::{
-                structured_storage::test::InMemoryStorage,
-                transactional::WriteTransaction,
-                StorageAsMut,
-            };
-            use $crate::rand::{
-                rngs::StdRng,
-                SeedableRng,
+#[allow(warnings)]
```
rymnc, line 488: ?

crates/storage/src/blueprint/sparse.rs
```diff
@@ -484,237 +484,483 @@ where
...
+    };
+
+    /// Provides root storage tests for SMT storage tables.
+    pub trait RootStorageTests: SmtTableWithBlueprint
+    where
+        Self::OwnedValue: PartialEq + core::fmt::Debug,
+        Self::Column: PartialEq,
+        <Self::Metadata as Mappable>::Key: Sized,
+
+        for<'a, 'b> <Self::Metadata as TableWithBlueprint>::Blueprint: BlueprintMutate<
```
rymnc, line 540: helper type would be nice here, as mentioned prev

crates/storage/src/blueprint/sparse.rs
```diff
@@ -484,237 +484,483 @@ where
...
+            Self::Metadata,
+            StructuredStorage<
+                &'a mut StorageTransaction<&'b mut InMemoryStorage<Self::Column>>,
+            >,
+        >,
+        for<'a> <Self::Metadata as TableWithBlueprint>::Blueprint: BlueprintMutate<
```
rymnc, line 546: helper type would be nice here, as mentioned prev

crates/storage/src/blueprint/sparse.rs
```diff
@@ -484,237 +484,483 @@ where
...
+            Self::Metadata,
+            StorageTransaction<&'a mut InMemoryStorage<Self::Column>>,
+        >,
+
+        for<'a, 'b> <Self::Nodes as TableWithBlueprint>::Blueprint: BlueprintMutate<
```
rymnc, line 551: helper type would be nice here, as mentioned prev

crates/storage/src/blueprint/sparse.rs
```diff
@@ -484,237 +484,483 @@ where
...
+            Self::Nodes,
+            StructuredStorage<
+                &'a mut StorageTransaction<&'b mut InMemoryStorage<Self::Column>>,
+            >,
+        >,
+        for<'a> <Self::Nodes as TableWithBlueprint>::Blueprint: BlueprintMutate<
```
rymnc, line 557: helper type would be nice here, as mentioned prev

crates/storage/src/blueprint/sparse.rs
```diff
@@ -484,237 +484,483 @@ where
...
-                structured_storage::test::InMemoryStorage,
-                transactional::WriteTransaction,
-                StorageAsMut,
-            };
-            use $crate::rand::{
-                rngs::StdRng,
-                SeedableRng,
+#[allow(warnings)]
+/// Test module for root storage tests.
+pub mod root_storage_tests {
```
rymnc, line 490: if this is only for smt, then we should include it in the name of the mod

crates/storage/src/blueprint/sparse.rs
```diff
@@ -484,237 +484,483 @@ where
...
+        },
+    };
+
+    use super::{
+        PrimaryKey,
+        Sparse,
+    };
+
+    /// Provides root storage tests for SMT storage tables.
+    pub trait RootStorageTests: SmtTableWithBlueprint
```
xgreenx, line 534: Why do we need a default implementation within the `RootStorageTests` trait when we could have a blanket implementation?

crates/storage/src/blueprint/merklized.rs
```diff
@@ -305,260 +305,511 @@ where
...
+        /// Returns a test key for the table
+        fn key() -> Box<Self::Key>;
 
-            #[test]
-            fn insert() {
-                let mut storage = InMemoryStorage::default();
-                let mut storage_transaction = storage.write_transaction();
-                let key = $key;
+        /// Returns a random key for testing
+        fn random_key(rng: &mut StdRng) -> Box<Self::Key>;
```
xgreenx, line 395: Why do we need to return `Box`? I think you should be able to put a `Sized` constraint on the `Key` and `Value`. Also, you can use `OwnedKey` or `OwnedValue` instead, if it is not possible because of Rust limitation

netrome: Yes this was actually a problem with the old macros. We have tables where the keys aren't sized, so putting the sized constraints would break it. I don't remember if there was any issues using `OwnedKey` and `OwnedValue` - I don't think so - so I can definitely change to use it.

crates/storage/src/lib.rs
```diff
@@ -9,7 +9,7 @@
 #![deny(clippy::cast_possible_truncation)]
 #![deny(unused_crate_dependencies)]
 #![deny(missing_docs)]
-#![deny(warnings)]
+//#![deny(warnings)]
```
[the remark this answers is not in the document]
netrome: Ouch, missed to reenable that. Fixed in https://github.com/FuelLabs/fuel-core/pull/2914/commits/06e806d8c09e16d7e937e3feb8f17272277f8a0d.

crates/storage/src/blueprint/merklized.rs
```diff
@@ -305,260 +305,511 @@ where
...
-            fn random<T, R>(rng: &mut R) -> T
-            where
-                rand::distributions::Standard: rand::distributions::Distribution<T>,
-                R: rand::Rng,
-            {
-                use rand::Rng;
-                rng.gen()
-            }
+/// Test module containing helper trait and macro for implementing merkle storage tests.
+#[allow(warnings)]
```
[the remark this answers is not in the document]
netrome: Yep that was just for development then I forgot about it. Fixed in https://github.com/FuelLabs/fuel-core/pull/2914/commits/06e806d8c09e16d7e937e3feb8f17272277f8a0d

crates/storage/src/blueprint/merklized.rs
```diff
@@ -305,260 +305,511 @@ where
...
-                    storage_transaction
-                        .storage_as_mut::<$table>()
-                        .get(&key)
-                        .expect("Should get without errors")
-                        .expect("Should not be empty")
-                        .into_owned(),
-                    $value_return
-                );
-            }
+        for<'a, 'b> <Self::Nodes as TableWithBlueprint>::Blueprint: BlueprintMutate<
```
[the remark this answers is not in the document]
netrome: Nice, will do.

crates/storage/src/blueprint/sparse.rs
netrome: Sure, though in general I don't think sub modules need to include the name of the parent module since that information is already stored in the namespace `blueprint::sparse::root_storage_tests` vs. `blueprint::sparse::root_storage_tests_smt`. I think the former is cleaner, but I'm fine with the other as well and since that's what we did for the merklized tests I'll do the same here.

crates/storage/src/blueprint/sparse.rs
netrome: Not sure I follow what you mean. Do you mean why we provide default implementations of the tests in the trait definition, as opposed to implementing them in the blanket implementation on line 890?

I think this is cleaner, since redefining the methods in the blanket implementation would cause us to have to repeat them additionally one time. And I don't see any benefit of doing it the other way around - although I have nothing against doing that way either.

Is this what you were referring to or did I misunderstand something?

.changes/added/2914.md
```diff
@@ -0,0 +1 @@
+Add tests for sparse and merkleized blueprint proofs.
```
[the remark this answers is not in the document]
netrome: Sure, I think this warrants a changelog. The tests themselves are debatable so this entry could be skipped - but the refactor is changing the api for downstream crates reusing these tests (which we'd like to do in SRS but couldn't due to the macros making implicit assumptions that didn't work for all tables - such as assuming keys always being sized)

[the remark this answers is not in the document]
netrome: Updated the formulations in https://github.com/FuelLabs/fuel-core/pull/2914/commits/1fad8627286614387b285e45e9d5224c1cbd4154. Let me know what you think.

crates/storage/src/blueprint/merklized.rs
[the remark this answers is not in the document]
netrome: Done in https://github.com/FuelLabs/fuel-core/pull/2914/commits/1a73ef3f5b2a533fa4d7707572f9717da31996a2

crates/storage/src/blueprint/merklized.rs
```diff
@@ -305,260 +305,511 @@ where
...
+
+            assert_eq!(returned_root, root);
+        }
+    }
 
-                let mut rng = rand::rngs::StdRng::seed_from_u64(1234);
+    /// Helper trait enabling referencing generics in
+    /// Merklized `TableWithBlueprint` implementations
+    /// as associated types.
+    pub trait MerklizedTableWithBlueprint:
```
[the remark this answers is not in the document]
netrome: Nope, not remove. Just not have it be the required way of implementing `TableWithBlueprint`. So previously if you wanted to implement `TableWithBlueprint` for a merklized blueprint, you'd have to implement `MerklizedTableWithBlueprint`, and get the `TableWithBlueprint` implementation by a blanket implementation. This could be seen as a bit intrusive for people not caring about the test code.

So what I did was flip the relation between the traits. If you implement the `TableWithBlueprint` trait using a `Merklized` blueprint, you instead get the `MerklizedTableWithBlueprint` implementation through the blanket implementation - thus providing all the tests.

This contains the wrapper trait to the test module and if you don't care about the tests, you don't have to know about it.

crates/storage/src/blueprint/merklized.rs
```diff
@@ -305,260 +305,511 @@ where
...
+
+        type ValueEncoder = ValueEncoder;
+    }
+}
+
+/// Generates test functions for tables using the merklelized structure.
+/// The table must implement `BasicMerkleizedStorageTests`.
+#[cfg(feature = "test-helpers")]
+#[macro_export]
+macro_rules! basic_merklelized_storage_tests {
```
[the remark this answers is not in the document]
netrome: Ah the classic merkleized vs merklized problem... We changed to merkleized in SRS I think because it's more correct, but here we should be consistent with the existing code. I've updated in https://github.com/FuelLabs/fuel-core/pull/2914/commits/1a73ef3f5b2a533fa4d7707572f9717da31996a2

crates/storage/src/blueprint/sparse.rs
netrome: Yep, forgot these... Thanks for spotting! https://github.com/FuelLabs/fuel-core/pull/2914/commits/e1ffdb9556b262eceddad322ebe91cd546001247

crates/storage/src/blueprint/sparse.rs
netrome: Yep, done in https://github.com/FuelLabs/fuel-core/pull/2914/commits/e1ffdb9556b262eceddad322ebe91cd546001247

crates/storage/src/blueprint/sparse.rs
netrome: https://github.com/FuelLabs/fuel-core/pull/2914/commits/e1ffdb9556b262eceddad322ebe91cd546001247

crates/storage/src/blueprint/sparse.rs
netrome: https://github.com/FuelLabs/fuel-core/pull/2914/commits/e1ffdb9556b262eceddad322ebe91cd546001247

crates/storage/src/blueprint/sparse.rs
netrome: Done in https://github.com/FuelLabs/fuel-core/pull/2914/commits/5422fde10d363e5dc344a3c0f999c849185901fc

crates/storage/src/blueprint/merklized.rs
netrome: Okay looked deeper into this and the problem here is that `Mappable` requires that we should be able to construct and `OwnedKey` from a `Key`, but not the other way around.

```
pub trait Mappable {
    /// The key type is used during interaction with the storage. In most cases, it is the
    /// same as `Self::OwnedKey`.
    type Key: ?Sized + ToOwned;
    /// The owned type of the `Key` retrieving from the storage.
    type OwnedKey: From<<Self::Key as ToOwned>::Owned> + Clone;
    /// The value type is used while setting the value to the storage. In most cases, it
    /// is the same as `Self::OwnedValue`, but it is without restriction and can be
    /// used for performance optimizations.
    type Value: ?Sized + ToOwned;
    /// The owned type of the `Value` retrieving from the storage.
    type OwnedValue: From<<Self::Value as ToOwned>::Owned> + Clone;
}
```

And since we need references to `Self::Key` for e.g. `insert`, we'd need something like this constraint to make it work:`Self::OwnedKey: AsRef<Self::Key>`. I think this should hold for most tables, but not necessarily for all of them - so it is more correct to stick with the boxes here. Also while I don't like unnecessary heap allocations, these boxes are only for tests and doesn't affect the rest of the code negatively in any way.

xgreenx (approved):

rymnc (approved):

xgreenx (approved):

rymnc (approved):

PR #2916, opened 2025-04-02
Change the number of write buffers in RocksDB to be more than 3.

AurelienFT:
## Description
When trying high load of transactions on the node we experienced some errors from RocksDB about write stalls. I tried differents things that I detailed just below to end up to this PR : 

- Tried to increase `set_max_total_wal_size` was just making the problem happens afterwards
- Tried to increase `set_max_background_jobs` didn't change much.
- Verifies that `no_slowdown` option on write options is already `false`, so slowdown is already activated by default.
- Final used solution : increase `set_max_write_buffer_number` to be more than `3` to activate this slowndown explained in the doc : `If max_write_buffer_number > 3, writing will be slowed down to options.delayed_write_rate if we are writing to the last write buffer allowed.`. Default value was 2. We only experience a lag of few dozens of milliseconds which is not a problem given our block time.

### Before requesting review

### DIFF
--- a/.changes/fixed/2916.md
+++ b/.changes/fixed/2916.md
@@ -1,0 +1,1 @@
+Change `max_write_buffer_number` to not experience write stalls under high loads.
--- a/crates/fuel-core/src/state/rocks_db.rs
+++ b/crates/fuel-core/src/state/rocks_db.rs
@@ -331,6 +331,7 @@ where
             opts.set_row_cache(&cache);
         }
         opts.set_max_background_jobs(6);
+        opts.set_max_write_buffer_number(4);
         opts.set_bytes_per_sync(1048576);
         opts.set_max_open_files(database_config.max_fds);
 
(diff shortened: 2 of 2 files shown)


xgreenx (commented):
It would be nice to make it configurable

xgreenx:
We found the root cause for stalls writes and https://github.com/FuelLabs/fuel-core/pull/2918 addresses it. If you think that it is better to make number of buffers configurable, feel free to reopen PR=)

rymnc:
> We found the root cause for stalls writes and https://github.com/FuelLabs/fuel-core/pull/2918 addresses it. If you think that it is better to make number of buffers configurable, feel free to reopen PR=)

i think we closed this PR prematurely. @AurelienFT what is in your db LOG files (on_chain/LOG or off_chain/LOG) when this stall happens?

if the log contains "stalling writes because ...." then we will still need this PR

AurelienFT:
Yeah it was something "because memtables 2" but I don't have it anymore now that the fix of green is applied.

PR #2730, opened 2025-02-19
Fix RocksDB closing

xgreenx:
Used the idea from the comment: https://github.com/facebook/rocksdb/issues/11349#issuecomment-2587107605

To address `pthread lock: Invalid argument` error in CI: https://github.com/FuelLabs/fuel-core/actions/runs/13423553399/job/37501700277

PR #2918, opened 2025-04-02
fix: Only cancel background work if primary RocksDB instance is dropped

netrome:
This effectively reverts https://github.com/FuelLabs/fuel-core/pull/2730/files which causes a lot of "Shutdown in progress: stalled writes" errors when multiple rocksdb instances are being dropped in a running node.


Cargo.toml
```diff
@@ -103,7 +103,7 @@ fuel-vm-private = { version = "0.60.0", package = "fuel-vm", default-features =
 
 # Common dependencies
 anyhow = "1.0"
-async-graphql = { version = "7.0.11", features = [
+async-graphql = { version = "=7.0.15", features = [
```
xgreenx, line 106:
```suggestion
async-graphql = { version = "7.0.15", features = [
```

xgreenx (approved):

Cargo.toml
```diff
@@ -104,7 +104,7 @@ fuel-vm-private = { version = "0.60.0", package = "fuel-vm", default-features =
 
 # Common dependencies
 anyhow = "1.0"
-async-graphql = { version = "7.0.11", features = [
+async-graphql = { version = "7.0.15", features = [
```
rymnc, line 107: do we need this change?

rymnc (approved):

Issue #2919, opened 2025-04-03
chore(compression_service): add metrics


rymnc:
Add the following metrics for visibility into our compression service -

1. `compression_ratio`
2. `compression_duration_ms`
3. `compression_block_height`

Issue #2332, opened 2024-10-11
Use `AtomicU128` in the metrics of the futures


xgreenx:
nit: we shoould probably have an issue for this

_Originally posted by @Dentosal in https://github.com/FuelLabs/fuel-core/pull/2324#discussion_r1797036490_

MitchTurner:
`AtomicU128` is not stable yet. We can reevaluate in a few months.

PR #2920, opened 2025-04-03
chore(compression): metrics for compression service

rymnc:
## Linked Issues/PRs

- closes https://github.com/FuelLabs/fuel-core/issues/2919

## Description

Adds 3 new metrics. confirmed working locally.

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/metrics/src/compression_metrics.rs
+++ b/crates/metrics/src/compression_metrics.rs
@@ -1,0 +1,55 @@
+use crate::global_registry;
+use prometheus_client::metrics::gauge::Gauge;
+use std::sync::{
+    atomic::{
+        AtomicU32,
+        AtomicU64,
+    },
+    OnceLock,
+};
+
+#[derive(Debug)]
+pub struct CompressionMetrics {
+    pub compressed_block_size_bytes: Gauge<u32, AtomicU32>,
+    pub compression_duration_ms: Gauge<f64, AtomicU64>,
+    pub compression_block_height: Gauge<u32, AtomicU32>,
+}
+
+impl Default for CompressionMetrics {
+    fn default() -> Self {
+        let compressed_block_size_bytes = Gauge::default();
+        let compression_duration_ms = Gauge::default();
+        let compression_block_height = Gauge::default();
+
+        let metrics = CompressionMetrics {
+            compressed_block_size_bytes,
+            compression_duration_ms,
+            compression_block_height,
+        };
+
+        let mut registry = global_registry().registry.lock();
+        registry.register(
+            "compressed_block_size_bytes",
+            "Compressed block size in bytes",
+            metrics.compressed_block_size_bytes.clone(),
+        );
+        registry.register(
+            "compression_duration_ms",
+            "Compression duration in milliseconds",
+            metrics.compression_duration_ms.clone(),
+        );
+        registry.register(
+            "compression_block_height",
+            "Compression block height",
+            metrics.compression_block_height.clone(),
+        );
+
+        metrics
+    }
+}
+
+static COMPRESSION_METRICS: OnceLock<CompressionMetrics> = OnceLock::new();
+
+pub fn compression_metrics() -> &'static CompressionMetrics {
+    COMPRESSION_METRICS.get_or_init(CompressionMetrics::default)
+}
--- a/crates/services/compression/src/metrics.rs
+++ b/crates/services/compression/src/metrics.rs
@@ -1,0 +1,26 @@
+use fuel_core_metrics::compression_metrics::compression_metrics;
+
+#[derive(Clone, Copy)]
+pub(crate) struct CompressionMetricsManager;
+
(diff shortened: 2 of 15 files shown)


crates/services/compression/src/metrics.rs
```diff
@@ -0,0 +1,32 @@
...
+            .compression_ratio
+            .set(compressed_size as f64 / uncompressed_size as f64);
+    }
+
+    pub(crate) fn record_compression_duration_ms(&self, duration_ms: u128) {
+        // this cast is safe. we would never record a compression longer than 49 days
+        // lets assume that we would be able to detect this and fix it before it happens :)
+        compression_metrics()
+            .compression_duration_ms
+            .set(u32::try_from(duration_ms).unwrap_or(u32::MAX));
```
xgreenx, line 26: Why can't we just use `u64` or `u128`?

crates/services/compression/src/service.rs
```diff
@@ -102,7 +116,7 @@ where
             .commit()
             .map_err(crate::errors::CompressionError::FailedToCommitTransaction)?;
 
-        Ok(())
+        Ok(compressed_block)
```
xgreenx, line 119: Before you commit to the database, you can get the size of the compressed block from the `storage_tx`, because we already serialized it there. And here we can already return the size

crates/services/compression/src/service.rs
```diff
@@ -113,8 +127,32 @@ where
...
+
+            // we calc the ratio after the compression is done so that these allocations dont affect
+            // the metrics for compression duration
+            if let (Ok(uncompressed), Ok(compressed)) = (
+                postcard::to_allocvec(&block_with_metadata.sealed_block.entity),
+                postcard::to_allocvec(&compressed_block),
+            ) {
+                metrics_manager
+                    .record_compression_ratio(uncompressed.len(), compressed.len());
+            }
```
xgreenx, line 148: Instead of recording ration here, you can record just compressed size.

The block importer can be responsible for storing the size of uncompressed block. And later on Grafana level we will just calculate ration by ourself.

Note: On the block importer side you also can avoid `postcard` serialization, and get the information about the size from the storage transaction(I suppose it should be possible, but you also need to sum the size of each transaction. If it is hard to do, then we can just serialize it there for now).

crates/metrics/src/compression_metrics.rs
```diff
@@ -0,0 +1,55 @@
...
+            "compression_duration_ms",
+            "Compression duration in milliseconds",
+            metrics.compression_duration_ms.clone(),
+        );
+        registry.register(
+            "compression_block_height",
+            "Compression block height",
+            metrics.compression_block_height.clone(),
+        );
+
```
netrome, line 46: Nit: Not a fan of side-effects in `Default` implementations. Spontaneously I'd prefer to see a derived `Default` implementation, and a separate ".register` method that takes the registry as an input and registers these metrics.

netrome (approved):
Approved, with one comment

crates/metrics/src/compression_metrics.rs
rymnc: this is how it is for all other metrics modules. while the comment is valuable, let's save that refactor for when we're fixing all of them :)

crates/services/compression/src/metrics.rs
rymnc: u128 not supported by the library i thinkso

rymnc: https://github.com/FuelLabs/fuel-core/issues/2332

crates/services/compression/src/service.rs
rymnc: addressed in 2157152ae3

crates/services/compression/src/service.rs
rymnc: addressed in 2157152ae3

we can deal with block importer metric in another pr

crates/services/compression/src/service.rs
rymnc: extra fixed in [d695a07](https://github.com/FuelLabs/fuel-core/pull/2920/commits/d695a07a83865252ff8100d9cf7fa149df442319)

crates/metrics/src/compression_metrics.rs
netrome: Fair enough :+1:

netrome (approved):

crates/services/compression/src/metrics.rs
xgreenx: We still can use `u64` for milliseconds=)

xgreenx (approved):

netrome (dismissed):

rymnc:
waiting on #2923 to merge this in. lesser conflicts then.

xgreenx (approved):

Issue #2924, opened 2025-04-04
Make pre-confirmation optional on API


AurelienFT:
## Description

To avoid any breaking changes we want to add a boolean to all API endpoints that serves the pre-confirmations to not serves them by default and only on-demand to not break the existing flows.

PR #2925, opened 2025-04-04
Make preconfirmation optional on API endpoints

AurelienFT:
## Linked Issues/PRs
Closes https://github.com/FuelLabs/fuel-core/issues/2924

## Description

| Endpoint affected | Changes realised | Tested where? | 
|-|-|-|
| statusChange  |  Filter the preconfirmations depending on the boolean values after conversion of the transaction status to the API one. | In preconfirmation integration tests |
| submitAndAwaitStatus | Filter the preconfirmations depending on the boolean values after conversion of the transaction status to the API one. | In preconfirmation integration tests |
| `status` field of `Transaction` object | Filter the preconfirmations depending on the boolean values after conversion of the transaction status to the API one. For preconfirmation we still want to send a status that will be Submitted. | Client doesn't support this feature so no explicit tests |

## Checklist

### Before requesting review

### DIFF
--- a/crates/client/assets/schema.sdl
+++ b/crates/client/assets/schema.sdl
@@ -1342,7 +1342,7 @@ type Subscription {
 		"""
 		If true, accept to receive the preconfirmation status
 		"""
-		allowPreconfirmation: Boolean
+		includePreconfirmation: Boolean
 	): TransactionStatus!
 	"""
 	Submits transaction to the `TxPool` and await either success or failure.
@@ -1353,7 +1353,7 @@ type Subscription {
 	Compared to the `submitAndAwait`, the stream also contains
 	`SubmittedStatus` and potentially preconfirmation as an intermediate state.
 	"""
-	submitAndAwaitStatus(tx: HexString!, estimatePredicates: Boolean, allowPreconfirmation: Boolean): TransactionStatus!
+	submitAndAwaitStatus(tx: HexString!, estimatePredicates: Boolean, includePreconfirmation: Boolean): TransactionStatus!
 	contractStorageSlots(contractId: ContractId!): StorageSlot!
 	contractStorageBalances(contractId: ContractId!): ContractBalance!
 }
@@ -1395,7 +1395,7 @@ type Transaction {
 	outputContract: ContractOutput
 	witnesses: [HexString!]
 	receiptsRoot: Bytes32
-	status(allowPreconfirmation: Boolean): TransactionStatus
+	status(includePreconfirmation: Boolean): TransactionStatus
 	script: HexString
 	scriptData: HexString
 	bytecodeWitnessIndex: U16
--- a/crates/client/assets/schema.sdl
+++ b/crates/client/assets/schema.sdl
@@ -1334,7 +1334,11 @@ type Subscription {
 		"""
 		The ID of the transaction
 		"""
-		id: TransactionId!
+		id: TransactionId!,
+		"""
+		If true, accept to receive the preconfirmation status
+		"""
+		allowPreconfirmation: Boolean
 	): TransactionStatus!
 	"""
 	Submits transaction to the `TxPool` and await either success or failure.
@@ -1343,9 +1347,9 @@ type Subscription {
 	"""
 	Submits the transaction to the `TxPool` and returns a stream of events.
 	Compared to the `submitAndAwait`, the stream also contains
-	`SubmittedStatus` as an intermediate state.
+	`SubmittedStatus` and potentially preconfirmation as an intermediate state.
 	"""
-	submitAndAwaitStatus(tx: HexString!, estimatePredicates: Boolean): TransactionStatus!
+	submitAndAwaitStatus(tx: HexString!, estimatePredicates: Boolean, allowPreconfirmation: Boolean): TransactionStatus!
 	contractStorageSlots(contractId: ContractId!): StorageSlot!
 	contractStorageBalances(contractId: ContractId!): ContractBalance!
 }
@@ -1387,7 +1391,7 @@ type Transaction {
 	outputContract: ContractOutput
 	witnesses: [HexString!]
 	receiptsRoot: Bytes32
-	status: TransactionStatus
+	status(allowPreconfirmation: Boolean): TransactionStatus
 	script: HexString
 	scriptData: HexString
 	bytecodeWitnessIndex: U16
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -938,8 +938,10 @@ impl FuelClient {
     pub async fn submit_and_await_status<'a>(
         &'a self,
         tx: &'a Transaction,
+        allow_preconfirmation: bool,
     ) -> io::Result<impl Stream<Item = io::Result<TransactionStatus>> + 'a> {
-        self.submit_and_await_status_opt(tx, None).await
+        self.submit_and_await_status_opt(tx, None, Some(allow_preconfirmation))
+            .await
     }
 
     /// Similar to [`Self::submit_and_await_commit_opt`], but includes all intermediate states.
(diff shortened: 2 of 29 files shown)


crates/client/src/client.rs
```diff
@@ -938,8 +938,10 @@ impl FuelClient {
     pub async fn submit_and_await_status<'a>(
         &'a self,
         tx: &'a Transaction,
+        allow_preconfirmation: bool,
```
rymnc, line 941: can we add a new function which has this preconf argument? submit_and_await_preconfirmation perhaps?

it makes the function call more clear

crates/fuel-core/src/schema/tx/types.rs
```diff
@@ -769,15 +769,22 @@ impl Transaction {
     async fn status(
         &self,
         ctx: &Context<'_>,
+        allow_preconfirmation: Option<bool>,
     ) -> async_graphql::Result<Option<TransactionStatus>> {
+        dbg!(allow_preconfirmation);
```
rymnc, line 774:
```suggestion
```

crates/fuel-core/src/query/subscriptions.rs
```diff
@@ -60,11 +61,15 @@ where
             match status {
                 TxStatusMessage::Status(status) => {
                     let status = ApiTxStatus::new(transaction_id, status);
-                    Ok(status)
+                    if !allow_preconfirmation && (matches!(status, ApiTxStatus::PreconfirmationFailure(_)) || matches!(status, ApiTxStatus::PreconfirmationSuccess(_))) {
```
xgreenx, line 64: We could add a method for the `status` like `is_preconfirmation` and call it here, will be much easier to read=)

tests/tests/preconfirmations.rs
```diff
@@ -118,6 +119,72 @@ async fn preconfirmation__received_after_successful_execution() {
     ));
 }
 
+#[tokio::test]
+async fn preconfirmation__not_received_when_not_asked() {
```
xgreenx, line 123: Can we also add `preconfirmation__received_when_asked` please? Just to be sure that this way of testing covers the right behavior=)

crates/fuel-core/src/query/subscriptions.rs
xgreenx: The same we can apply to all places where we use `matches!`=)

crates/client/assets/schema.sdl
```diff
@@ -1387,7 +1391,7 @@ type Transaction {
 	outputContract: ContractOutput
 	witnesses: [HexString!]
 	receiptsRoot: Bytes32
-	status: TransactionStatus
+	status(allowPreconfirmation: Boolean): TransactionStatus
```
xgreenx, line 1394: I didn't know that it is possible, cool=D

crates/client/src/client.rs
```diff
@@ -938,8 +938,10 @@ impl FuelClient {
     pub async fn submit_and_await_status<'a>(
```
xgreenx, line 938: I think we can leave the signature of this function and just accept new field in the `submit_and_await_status_opt`=)

crates/client/src/client.rs
```diff
@@ -938,8 +938,10 @@ impl FuelClient {
...
         &'a self,
         tx: &'a Transaction,
+        allow_preconfirmation: bool,
     ) -> io::Result<impl Stream<Item = io::Result<TransactionStatus>> + 'a> {
-        self.submit_and_await_status_opt(tx, None).await
+        self.submit_and_await_status_opt(tx, None, Some(allow_preconfirmation))
```
xgreenx, line 943: You also forgot to update `subscribe_transaction_status` function. Maybe instead of updating it, we could add a new `subscribe_transaction_status_opt`

crates/fuel-core/src/service/query.rs
```diff
@@ -89,7 +89,7 @@ impl FuelService {
             db,
             tx_status_manager,
         };
-        Ok(transaction_status_change(state, rx, id).await)
+        Ok(transaction_status_change(state, rx, id, false).await)
```
xgreenx, line 92: I think this function can also benefit from making it configurable.

But I'm fine to not propagate it=) But then I think it is better to use `true`, because we want to allow pre confirmations

crates/fuel-core/src/schema/tx/types.rs
```diff
@@ -1109,7 +1116,28 @@ pub(crate) async fn get_tx_status(
         None => {
             let status = tx_status_manager.status(id).await?;
             match status {
-                Some(status) => Ok(Some(TransactionStatus::new(id, status))),
+                Some(status) => {
```
xgreenx, line 1119: image


`StatusChangeState` also will benefit from this change, because we will not lose `Submitted` status. It owed be nice to use this function there, because it works with exact same types=)

crates/client/src/client.rs
AurelienFT: I don't like the idea of having a new endpoint for each parameter we have. Because what if we had the same thing for `estimate_predicates` and you want to make a `submit_and_await_status` with both options, you have to make two calls ? but you can't because it's a submit.

crates/client/src/client.rs
AurelienFT: Done

tests/tests/preconfirmations.rs
AurelienFT: Done.

crates/client/src/client.rs
AurelienFT: Done.

crates/fuel-core/src/schema/tx/types.rs
AurelienFT: The database types doesn't seems to be the same and the tx_status_manager neither. I tried different things but couldn't get a compiling version. I don't want to spend too much time on that if you have the idea to do it it will be more efficient.

xgreenx (dismissed):

crates/client/assets/schema.sdl
```diff
@@ -1338,7 +1338,11 @@ type Subscription {
 		"""
 		The ID of the transaction
 		"""
-		id: TransactionId!
+		id: TransactionId!,
+		"""
+		If true, accept to receive the preconfirmation status
+		"""
+		allowPreconfirmation: Boolean
```
Dentosal, line 1345: nit: I'd prefer either `includePrecofirmations` or just `preconfirmations`

crates/fuel-core/src/combined_database.rs
```diff
@@ -523,6 +523,14 @@ impl CombinedDatabase {
 
         Ok(())
     }
+
+    pub fn shutdown(self) {
```
Dentosal, line 527: This seems unrelated to the main PR. Does this have a specific purpose compared to just dropping the handles? I'm fine with doing this here, but maybe the changelog should mention it?

crates/fuel-core/src/service/query.rs
```diff
@@ -99,11 +100,28 @@ struct StatusChangeState<'a> {
...
+        allow_preconfirmation: bool,
+    ) -> StorageResult<Option<TxPoolTxStatus>> {
         match self.db.get_tx_status(&id)? {
             Some(status) => Ok(Some(status.into())),
             None => {
                 let status = self.tx_status_manager.status(id).await?;
+                let status = status.map(|status| {
+                    // Filter out preconfirmation statuses if not allowed. Converting to submitted status
+                    // because it's the closest to the preconfirmation status.
+                    // Having `now()` as timestamp isn't ideal but shouldn't cause much inconsistency.
```
Dentosal, line 115: Should we keep submitted timestamp when a tx is preconfirmed?

crates/fuel-core/src/combined_database.rs
AurelienFT: cc @xgreenx that added this change

crates/fuel-core/src/service/query.rs
AurelienFT: Maybe we can open a followup discussion because it would require the emitter of preconfirmation (executor) to get something else than just a transaction but a transaction with some metadata

crates/client/assets/schema.sdl
AurelienFT: Changed to `includePrecomfirmation` it's still singular as you can only receive 1 preconfirmation for a TX.

rymnc (approved):
i would still prefer if preconfirmation related tx submission functions should be separate but its okay, not a blocker

Issue #2950, opened 2025-04-15
Rework executable transactions selection


AurelienFT:
Currently the selection for transaction is pretty inefficient (iterating over all tx, not much indexing, can't take coin dependent transaction directly,  ...).

We need to change the internal representation of executable transactions in order to improve performance of the current behavior but also of the future ones such as #2949

PR #2956, opened 2025-04-16
Add excluded_contract filter to extract of transaction from TxPool

### LINKED ISSUE #2949: Add contract id filtering option on TxPool txs extractions
We want to have a new parameter to the `extract_transactions_for_block` function which allow us to pass a set of contract id that shouldn't be used as inputs of the transactions given by the transaction pool.

This is useful for avoiding to collide too much on execute of transactions if we want to execute them in parallel.

AurelienFT:
## Linked Issues/PRs
Close https://github.com/FuelLabs/fuel-core/issues/2949

## Description
This is an unoptimized version that will be changed with the rest of the already inefficient version of selection when we will do https://github.com/FuelLabs/fuel-core/issues/2950.

## Checklist

### Before requesting review


rymnc (commented):
in general lgtm ~ but im curious about the behaviour of the txpool. how often will we have insertions and deletions?

AurelienFT:
> in general lgtm ~ but im curious about the behaviour of the txpool. how often will we have insertions and deletions?

We can have thousands of new transactions every seconds and thousands extracted every seconds. That's why this algorithm is not in the best shape and https://github.com/FuelLabs/fuel-core/issues/2950 exists

crates/services/txpool_v2/src/storage/graph.rs
```diff
@@ -735,7 +735,9 @@ impl Storage for GraphStorage {
                     }
                 }
                 Input::Contract(Contract { contract_id, .. }) => {
-                    if !self.contracts_creators.contains_key(contract_id) {
+                    if utxo_validation
+                        && !self.contracts_creators.contains_key(contract_id)
+                    {
```
xgreenx, line 740: We always want to require contract to exist=D Regardless of `utxo_validation`.

Otherwise executor will fail

xgreenx (dismissed):

xgreenx (approved):

PR #2959, opened 2025-04-16
chore(compression): merklize compressed blocks table

rymnc:
> [!WARNING]
> wip


crates/compression/src/ports.rs
```diff
@@ -141,6 +141,28 @@ pub mod fault_proving {
             D::registry_root(self)
         }
     }
+
+    pub trait CompressedBlockRootProvider {
+        type Error: core::fmt::Display;
```
netrome, line 146: Nit: I think it's more idiomatic to require `Error` here, which implies `Display`.
```suggestion
        type Error: core::error::Error;
```

crates/services/compression/src/storage.rs
```diff
@@ -46,5 +47,8 @@ pub type RegistryIndex = Merkleized<registry_index::RegistryIndex>;
 /// Merkleized Timestamps table type alias
 pub type Timestamps = Merkleized<timestamps::Timestamps>;
 
-/// Re-export to match api
-pub use compressed_blocks::CompressedBlocks;
+/// Merkleized compressed blocks table type alias
+pub type CompressedBlocks = Merkleized<compressed_blocks::CompressedBlocks>;
```
netrome, line 51: Using `Merklized` means that we're using the SMT blueprint. However, do we need SMTs for this table? The normal block roots are using BMTs instead, which I'd assume is sufficient in this case as well. Or am I forgetting something?

crates/services/compression/src/storage/compressed_blocks_roots.rs
```diff
@@ -0,0 +1,57 @@
...
+    merkle::sparse::MerkleizedTableColumn,
+    structured_storage::TableWithBlueprint,
+    Mappable,
+    MerkleRoot,
+};
+
+use fuel_core_types::fuel_types::BlockHeight;
+
+/// The table for the compressed blocks sent to DA.
+pub struct CompressedBlocksRoots;
```
netrome, line 22: Is this table used? Since we use the `Merkleized` table wrapper we should not need a dedicated table for the roots, as they'll be in the `MerkleMetadata<...>` table for the associated table column.

netrome: (This may be different if we chose to go with BMTs instead of SMTs)

netrome (commented):
A few questions so far. Primarily, I'm curious why we're using SMTs and not BMTs here. Secondly, if using SMTs I believe the `CompressedBlocksRoots` table is redundant, and if not I'd be curious to learn why we need it.

crates/services/compression/src/ports/compression_storage.rs
```diff
@@ -61,10 +63,27 @@ where
...
+                CompressedBlocksRoots,
+            };
+
+            let root =
+                <Self as ComputeRegistryRoot>::root_of_table::<CompressedBlocks>(self)
+                    .map_err(CompressionError::FailedToComputeCompressedBlockRoot)?;
+            self.storage_as_mut::<CompressedBlocksRoots>()
+                .insert(&height, &root)
+                .map_err(CompressionError::FailedToWriteCompressedBlockRoot)?;
+        };
```
xgreenx, line 85: I don't think that we need an additional `CompressedBlocksRoots` table. If  `CompressedBlocks` uses `Sparse` blueprint, the nit shoudl be enought for us to get the root.

crates/compression/src/compress.rs
```diff
@@ -104,6 +121,8 @@ where
         transactions,
         #[cfg(feature = "fault-proving")]
         registry_root,
+        #[cfg(feature = "fault-proving")]
+        prev_compressed_block_root,
```
xgreenx, line 125: We can include the Mekrle root of the `CompressedBlocks` into  `registry_root`, as we do with other roots.

In the proving game we will verify that all entries provided for all tables are correct and connected ot the one `registry_root `.

crates/services/compression/src/storage.rs
rymnc: resolved on slack

crates/services/compression/src/service.rs
```diff
@@ -649,4 +649,39 @@ mod tests {
             .unwrap();
         assert!(maybe_block.is_some());
     }
+
+    #[cfg(feature = "fault-proving")]
+    #[tokio::test]
+    async fn compression_service__merkleizes_registrations_per_table() {
```
netrome, line 655: We're not testing any merkleization here, I suppose the name should be updated.

netrome (dismissed):

crates/services/compression/src/service.rs
rymnc: addressed in e5aeb4b9c3

netrome (approved):

netrome (approved):

xgreenx (dismissed):
I think we should insert registry in the `compress` function before we calculate the root.

Because we want root to be updated.

image


Also, in the follow up PR we need to remove old registrations, plus make registrations merkle root part of the `registry_root`.

But all of that can be follow up PR

xgreenx (approved):

AurelienFT (approved):

PR #2967, opened 2025-04-22
(parallel-executor): tests

### LINKED ISSUE #2951: Create tests for the parallel-executor
This new module could benefit from TDD (test driven development) and so we should create the tests first. The skeleton of this module already exists but it needs to have a bit more data inside in order to allow better tests scenarios.

Design of parallel-executor : https://www.notion.so/fuellabs/Parallel-block-execution-1502f2293f3180d7b861cc02c510e558?pvs=25#1d02f2293f3180ff8285f0bcb87e26c2

Items that need to be tested : 
- `ParallelExecutor` asks correctly the transactions to `TxPool`.
  - Given the state of workers thread some contract id must be filtered, the number of gas remaining must be lower, ...
  - Given the answers from transaction pool about the remaining tx or not, the `ParallelExecutor` must be able to   adapt his next requests to the `TxPool`
- Ordering of transactions must produce a valid block and follow the priority ordering from the pool
- Transactions that require state from other workers must influence the total consumption and taken in the gas left. (time left)
- Coins mustn't create dependencies but a check at the end must ensure that the dependency tree in the block is correct and all coins exists when created
- The worstest case we can find with contract dependency mustn't break our limits rule.

AurelienFT:
## Linked Issues/PRs
Resolve #2951 

## Description
All tests needed to ensure the correct behavior of the parallel-executor

## Checklist

### Before requesting review


AurelienFT:
@xgreenx I would like to have your feedback on the tests created here that should outline the behavior of parallel executor.

I'm still questioning myself about asking to the pool one time for each thread. For example in the first test scenario it means that only worker thread 1 will work and second never. Maybe if transaction source is informing when asking for example for thread 2 that there is nothing we can split the queue of thread 1 in the others.

I will start implementation on top of this PR.

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -0,0 +1,637 @@
...
+                    TransactionFiltered::NotFiltered,
+                ))
+                .unwrap();
+        }
+    });
+}
+
+#[test]
+#[ignore]
+fn execute__utxo_ordering_kept() {
```
netrome, line 522: Interesting. Is it an important property we rely on right now that dependent transactions occur in-order in blocks?

netrome (approved):
Nice with the TDD approach! Some nits and questions from my side. I'm primarily curious about why we use threads and synchronous channels as opposed to tokio tasks and async channels.

Beyond that, I don't spot any obvious errors.

crates/services/parallel-executor/src/tests/mocks.rs
```diff
@@ -0,0 +1,70 @@
...
+                    tx_count_limit,
+                    block_transaction_size_limit,
+                    filter,
+                },
+                tx,
+            ))
+            .expect("Failed to send request");
+        rx.recv().expect("Failed to receive response")
+    }
+}
```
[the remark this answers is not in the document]
AurelienFT: The goal is to match as much as possible the interface that currently exist on the ports of the current executor (with some additional informations.

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -0,0 +1,637 @@
...
+        .produce_without_commit_with_source(Components {
+            header_to_produce: Default::default(),
+            transactions_source,
+            coinbase_recipient: Default::default(),
+            gas_price: 0,
+        })
+        .unwrap()
+        .into_result();
+
+    // Then
```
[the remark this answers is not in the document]
AurelienFT: In my mind it's more "when executing the block, then this calls are made to the transactions sources"

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -0,0 +1,637 @@
...
+    std::thread::spawn({
```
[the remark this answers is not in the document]
AurelienFT: It's not really important but everything is using sync. I don't mind changing but I don't see any reasons to change it to async code

crates/services/parallel-executor/src/tests/mocks.rs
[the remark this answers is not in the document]
netrome: Right and that one has a synchronous interface. I see. It's a bit ugly though since we're having to use this ugly `try_recv` loop to work around the missing `await` mechanism in the current code:

```
    pub fn extract_transactions_for_block(
        &self,
        constraints: Constraints,
    ) -> Result<Vec<ArcPoolTx>, Error> {
        let (select_transactions_sender, mut select_transactions_receiver) =
            oneshot::channel();
        self.select_transactions_requests_sender
            .try_send(
                pool_worker::PoolExtractBlockTransactions::ExtractBlockTransactions {
                    constraints,
                    transactions: select_transactions_sender,
                },
            )
            .map_err(|_| Error::ServiceCommunicationFailed)?;

        loop {
            let result = select_transactions_receiver.try_recv();
            match result {
                Ok(txs) => {
                    return Ok(txs);
                }
                Err(TryRecvError::Empty) => continue,
                Err(TryRecvError::Closed) => {
                    return Err(Error::ServiceCommunicationFailed);
                }
            }
        }
    }
```

But I get the argument for consistency, it makes sense as well. Perhaps we could tackle this separately. There are probably good reasons for why the original trait isn't async as well.

crates/services/parallel-executor/src/tests/tests_executor.rs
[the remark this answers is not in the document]
netrome: Yeah I think the interfaces should be async as well, given that this touches IO. But your argument for consistency with existing patterns also makes sense.

crates/services/parallel-executor/src/tests/mocks.rs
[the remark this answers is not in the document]
AurelienFT: Yeah the original one isn't async because it imply a lot of async in the whole executor but now that executor is a bit async for open block production maybe we can do it. But maybe the interface will change in the next PRs it's a base for the tests

AurelienFT:
@xgreenx in addition of my previous questions here : https://github.com/FuelLabs/fuel-core/pull/2967#issuecomment-2824029714

I would like to ask you if you have an idea on how I can re-use the same DB as for the executor tests and avoid having a circular-dependency problem like this. The existing executor tests lives in fuel-core directly and so doesn't have the problem. Is it my only way to go also ?

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -0,0 +1,637 @@
...
+            let (pool_request_params, response_sender) =
+                tx_pool_requests_receiver.recv().unwrap();
+            assert_eq!(pool_request_params.filter.excluded_contract_ids.len(), 1);
+            assert_eq!(
+                pool_request_params
+                    .filter
+                    .excluded_contract_ids
+                    .get(&contract_id_1),
+                Some(&contract_id_1)
+            );
```
segfault-magnet, line 460: There seems to be a lot of this repeating, maybe something like this:
```rust
            Consumer::receive(&tx_pool_requests_receiver)
                .assert_excluded_eq_to([contract_id_1])
                .respond_with(&[&tx_contract_2], TransactionFiltered::NotFiltered);
```
can replace
```rust
            let (pool_request_params, response_sender) =
                tx_pool_requests_receiver.recv().unwrap();
            assert_eq!(pool_request_params.filter.excluded_contract_ids.len(), 1);
            assert_eq!(
                pool_request_params
                    .filter
                    .excluded_contract_ids
                    .get(&contract_id_1),
                Some(&contract_id_1)
            );
            response_sender
                .send((
                    vec![MaybeCheckedTransaction::CheckedTransaction(
                        tx_contract_2
                            .clone()
                            .into_checked_basic(
                                0u32.into(),
                                &ConsensusParameters::default(),
                            )
                            .unwrap()
                            .into(),
                        0,
                    )],
                    TransactionFiltered::NotFiltered,
                ))
                .unwrap();
```

definition:
```rust
    struct Consumer {
        pool_request_params: PoolRequestParams,
        response_sender:
            std::sync::mpsc::Sender<(Vec<MaybeCheckedTransaction>, TransactionFiltered)>,
    }

    impl Consumer {
        fn receive(receiver: &GetExecutableTransactionsReceiver) -> Self {
            let (pool_request_params, response_sender) = receiver.recv().unwrap();

            Self {
                pool_request_params,
                response_sender,
            }
        }
        fn assert_excluded_eq_to(
            &self,
            expected_excluded_contract_ids: impl IntoIterator<Item = ContractId>,
        ) -> &Self {
            assert_eq!(
                self.pool_request_params.filter.excluded_contract_ids,
                expected_excluded_contract_ids
                    .into_iter()
                    .collect::<std::collections::HashSet<_>>()
            );
            self
        }

        fn respond_with(
            &self,
            txs: &[&Transaction],
            filtered: TransactionFiltered,
        ) -> &Self {
            let txs = into_checked_txs(&txs);

            self.response_sender.send((txs, filtered)).unwrap();
            &self
        }
    };
```

segfault-magnet (commented):
Unfortunately can only suggest readability nits at this point :)

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -0,0 +1,637 @@
+#![allow(non_snake_case)]
+
+use fuel_core::database::Database;
```
AurelienFT, line 3: Use InMemoryStorage

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -0,0 +1,637 @@
...
+                                    &ConsensusParameters::default(),
+                                )
+                                .unwrap()
+                                .into(),
+                            0,
+                        ),
+                    ],
+                    TransactionFiltered::NotFiltered,
+                ))
+                .unwrap();
```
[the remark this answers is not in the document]
AurelienFT: Done along with the last suggestion

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -0,0 +1,637 @@
...
+        tx1.id(&ChainId::default())
+    );
+    assert_eq!(
+        transactions[2].id(&ChainId::default()),
+        tx4.id(&ChainId::default())
+    );
+    assert_eq!(
+        transactions[3].id(&ChainId::default()),
+        tx3.id(&ChainId::default())
+    );
```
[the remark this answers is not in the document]
AurelienFT: Done

crates/services/parallel-executor/src/tests/tests_executor.rs
AurelienFT: Done, Thanks a lot this is a great idea to keep the test very readable !

crates/services/parallel-executor/src/tests/tests_executor.rs
AurelienFT: Done

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -0,0 +1,637 @@
...
+        .add_input(Input::coin_predicate(
+            rng.r#gen(),
+            owner,
+            1000,
+            Default::default(),
+            Default::default(),
+            Default::default(),
+            predicate.clone(),
+            vec![],
+        ))
```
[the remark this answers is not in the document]
AurelienFT: <Done

rymnc (dismissed):

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -0,0 +1,603 @@
...
+            coinbase_recipient: Default::default(),
+            gas_price: 0,
+        })
+        .await
+        .unwrap()
+        .into_changes();
+    (contract_id, StorageChanges::Changes(res))
+}
+
+#[ignore]
```
MitchTurner: nit: if we want to go hard-core TDD with this, you could do `#[should_panic]` instead of `#[ignore]`. Doesn't really matter at the end of the day :)

crates/services/parallel-executor/src/tests/mocks.rs
```diff
@@ -0,0 +1,187 @@
...
+
+pub type GetExecutableTransactionsSender = std::sync::mpsc::Sender<(
+    PoolRequestParams,
+    std::sync::mpsc::Sender<(Vec<CheckedTransaction>, TransactionFiltered, Filter)>,
+)>;
+
+pub type GetExecutableTransactionsReceiver = std::sync::mpsc::Receiver<(
+    PoolRequestParams,
+    std::sync::mpsc::Sender<(Vec<CheckedTransaction>, TransactionFiltered, Filter)>,
+)>;
```
MitchTurner: Why are these named `get__` and `Get___`? It's weird to me for a value or a type be a verb. Seems like methods should be verbs.

I see that it's kinda complicated in that it receives a sender. Maybe we can just have a struct that abstracts the internals of channels sending channels and just have methods with domain names?

crates/services/parallel-executor/src/tests/mocks.rs
```diff
@@ -0,0 +1,187 @@
...
+        // This is a mock implementation, so we return a dummy Notify instance
+        tokio::sync::Notify::new()
+    }
+}
+
+pub struct Consumer {
+    pool_request_params: PoolRequestParams,
+    response_sender:
+        std::sync::mpsc::Sender<(Vec<CheckedTransaction>, TransactionFiltered, Filter)>,
+}
```
MitchTurner: I'm having a hard time understanding what the `Consumer` is.

For example, it doesn't make a lot of sense to have something you call `receive` on... like shouldn't we `send` something and it will by receive on its own? Could that method be just `new` and just pass the `sender` and  params?

After reading through it a bit, it seems like the Consumer is like a container for `asserts` on a specific set of params and sender. Pairing the `respond_with` and `assert`s for the params feels like a violation of SRP and I think it would be easier to read if you just had the checks as individual functions with references to the params.

i.e.
```
     pub fn assert_filter_matches(params: &PoolRequestParams, filter: &Filter) -> &Self {
        assert_eq!(&self.pool_request_params.filter, filter);
        self
    }
```
or even just write it in-line.

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -0,0 +1,603 @@
...
+            Consumer::receive(&tx_pool_requests_receiver)
+                .assert_filter(&empty_filter())
+                .respond_with(&[&tx_contract_1], TransactionFiltered::NotFiltered);
+
+            // Request for thread 2
+            Consumer::receive(&tx_pool_requests_receiver)
+                .assert_filter(&Filter {
+                    excluded_contract_ids: vec![contract_id_1].into_iter().collect(),
+                })
+                .respond_with(&[&tx_contract_2], TransactionFiltered::NotFiltered);
```
MitchTurner: How do we know that these go with `thread 1` and `thread 2`? Just trying to understand.

AurelienFT:
@MitchTurner I updated the mocktxpool and comments on threads to be more clear, tell me if it address your concerns :)

crates/services/parallel-executor/src/ports.rs
```diff
@@ -0,0 +1,59 @@
...
+
+pub trait TransactionsSource {
+    /// Returns the a batch of transactions to satisfy the given parameters
+    fn get_executable_transactions(
+        &mut self,
+        gas_limit: u64,
+        tx_count_limit: u16,
+        block_transaction_size_limit: u32,
+        filter: Filter,
+    ) -> (Vec<CheckedTransaction>, TransactionFiltered, Filter);
```
xgreenx: It would be nice if we returned a structure instead of tuple. Also comments to fields of that structure will be useful for context

crates/services/parallel-executor/src/ports.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    /// Returns a notification receiver for new transactions
+    fn get_new_transactions_notifier(&mut self) -> tokio::sync::Notify;
+}
+
+pub trait Storage {
```
xgreenx, line 61: I don't where do we use methods from this trait.

Or is that a preparation for the next PR?

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -0,0 +1,614 @@
...
+        .block
+        .transactions()
+        .iter()
+        .map(|tx| tx.id(&ChainId::default()))
+        .rev()
+        .skip(1)
+        .rev()
+        .collect::<Vec<_>>();
+
+    assert_eq!(expected_ids, actual_ids);
```
xgreenx, line 323: I wouldn't sure about the order of transactions=D Potentially executor can run them in any order if they are independent

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -0,0 +1,614 @@
...
+    // When
+    let future = executor.produce_without_commit_with_source(Components {
+        header_to_produce: Default::default(),
+        transactions_source,
+        coinbase_recipient: Default::default(),
+        gas_price: 0,
+    });
+
+    // Then
+    let txpool = std::thread::spawn({
```
xgreenx: It would be nice if we didn't use `std::thread::spawn` in "then` section. It looks like it is part of the test setup.

And block execution should fail if it doesn't match the test setup

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -0,0 +1,614 @@
...
+                .waiting_for_request_to_tx_pool()
+                .assert_filter(&empty_filter())
+                .respond_with(&[&tx_contract_1], TransactionFiltered::NotFiltered);
+
+            // Request for the other thread
+            mock_tx_pool
+                .waiting_for_request_to_tx_pool()
+                .assert_filter(&Filter {
+                    excluded_contract_ids: vec![contract_id_1].into_iter().collect(),
+                })
```
xgreenx: I think you can create a `new` constructor for `Filter` and then it will take only one line=)

crates/services/parallel-executor/src/ports.rs
AurelienFT: Yes prep.

crates/services/parallel-executor/src/ports.rs
AurelienFT: Done.

crates/services/parallel-executor/src/tests/tests_executor.rs
AurelienFT: Done.

crates/services/parallel-executor/src/tests/tests_executor.rs
AurelienFT: It shouldn't because they all have been sent in the same batch. And so the goal of this test is to match that if we sent txs in a certain order in a batch (in real world it would be based on gas/tip ratio) they will be in this order in the block.

crates/services/parallel-executor/src/tests/tests_executor.rs
AurelienFT: Done

xgreenx (approved):

Issue #2952, opened 2025-04-15
Implement the logic for the block production in parallel-executor


AurelienFT:
Detailed design : https://www.notion.so/fuellabs/Parallel-block-execution-1502f2293f3180d7b861cc02c510e558?pvs=25#1d02f2293f3180ff8285f0bcb87e26c2

PR #2976, opened 2025-04-24
Start skeleton for the parallel executor scheduler

AurelienFT:
## Linked Issues/PRs
Part of https://github.com/FuelLabs/fuel-core/issues/2952

## Description
Describe in https://github.com/FuelLabs/fuel-core/issues/2952
Will add more implementation details when finished

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/services/parallel-executor/src/scheduler.rs
+++ b/crates/services/parallel-executor/src/scheduler.rs
@@ -1,0 +1,127 @@
+use std::collections::{
+    HashMap,
+    HashSet,
+};
+
+use fuel_core_storage::transactional::StorageChanges;
+use fuel_core_types::fuel_tx::ContractId;
+use fuel_core_upgradable_executor::executor::Executor as UpgradableExecutor;
+use tokio::runtime::Runtime;
+
+use crate::ports::{
+    Filter,
+    TransactionsSource,
+};
+
+/// The scheduler is responsible for managing the state of all the execution workers.
+/// His goal is to gather transactions for the transaction source and organize their execution
+/// through the different workers.
+///
+/// There is few rules that need to be followed in order to produce a valid execution result:
+/// - The dependency chain of the input and output must be maintained across the block.
+/// - The constraints of the block (maximum number of transactions, maximum size, maximum gas, etc.) must be respected.
+///
+/// Current design:
+///
+/// The scheduler creates multiple workers. For each of this workers, the scheduler will ask to the transaction source
+/// to provide a batch of transactions that can be executed.
+///
+/// When a thread has finished his execution then it will notify the scheduler that will re-ask for a new batch to the transaction source.
+/// This new batch mustn't contain any transaction that use a contract used in a batch of any other worker.
+///
+/// For transactions without contracts, they are treat them as independent transactions. A verification is done at the end for the coin dependency chain.
+/// This can be done because we assume that the transaction pool is sending us transactions that are already correctly verified.
+/// If we have a transaction that end up being skipped (only possible cause if consensus parameters changes) then we will have to
+/// fallback a sequential execution of the transaction that used the skipped one as a dependency.
+
+type WorkerID = usize;
+
+pub struct Scheduler<TxSource> {
+    /// The state of each worker
+    workers_state: Vec<WorkerState>,
+    /// Latest worker that executed a transaction with a specific contract
+    contracts_info: HashMap<ContractId, WorkerID>,
+    /// The list of contracts that are currently being executed for each worker (useful to filter them when asking to transaction source)
+    executing_contracts: Vec<Vec<ContractId>>,
+    /// Transaction source to ask for new transactions
+    transaction_source: TxSource,
+    /// Runtime to run the workers
+    runtime: Option<Runtime>,
+}
+
+pub struct WorkerState {
+    pub status: WorkerStatus,
+    pub state: StorageChanges,
+    pub gas_left: u64,
+    pub tx_left: u16,
+    pub tx_size_left: u32,
+    // Maybe something that could wake the scheduler
+}
+
+pub enum WorkerStatus {
+    /// The worker is waiting for a new batch of transactions
+    Idle,
+    /// The worker is currently executing a batch of transactions
+    Executing,
+    /// The worker is merging his state with another
+    Merging,
+}
+
+// Shutdown the tokio runtime to avoid panic if executor is already
+//   used from another tokio runtime
+impl<TxSource> Drop for Scheduler<TxSource> {
+    fn drop(&mut self) {
+        if let Some(runtime) = self.runtime.take() {
+            runtime.shutdown_background();
+        }
+    }
+}
+
+impl<TxSource> Scheduler<TxSource>
+where
+    TxSource: TransactionsSource,
+{
+    pub fn new(number_of_worker: usize, transaction_source: TxSource) -> Self {
+        let mut workers_state = Vec::with_capacity(number_of_worker);
+        for _ in 0..number_of_worker {
+            workers_state.push(WorkerState {
+                status: WorkerStatus::Idle,
+                state: StorageChanges::default(),
+                // TODO
+                gas_left: 0,
+                tx_left: 0,
+                tx_size_left: 0,
+            });
+        }
+        let runtime = tokio::runtime::Builder::new_multi_thread()
+            .worker_threads(number_of_worker)
+            .enable_all()
+            .build()
+            .unwrap();
+
+        Self {
+            workers_state,
+            contracts_info: HashMap::new(),
+            executing_contracts: vec![Vec::new(); number_of_worker],
+            transaction_source,
+            runtime: Some(runtime),
+        }
+    }
+
+    pub async fn run(&mut self) {
+        let runtime = self.runtime.as_ref().unwrap();
+        // All workers starts empty, so we fetch the first batch
+        for state in self.workers_state.iter_mut() {
+            let (batch, _) = self.transaction_source.get_executable_transactions(
+                state.gas_left,
+                state.tx_left,
+                state.tx_size_left,
+                Filter {
+                    excluded_contract_ids: HashSet::new(),
+                },
+            );
+            runtime.spawn(async move {})
+        }
+        // Waiting for the workers to notify us
+    }
+}
--- a/crates/services/parallel-executor/src/lib.rs
+++ b/crates/services/parallel-executor/src/lib.rs
@@ -4,3 +4,5 @@ pub mod ports;
 
 #[cfg(test)]
 mod tests;
+
+pub mod scheduler;


AurelienFT:
@xgreenx I would like to know if you had something in mind for managing all the worker threads. I usually like to go with creating OS threads but I know that you often prefer the tokio runtime with multiple tasks. But compare to the first version this one is fetching batch of txs multiple times and we admitted during the design that the states belong to one worker which is not possible to isolate like that with tokio

crates/services/parallel-executor/Cargo.toml
```diff
@@ -11,15 +11,24 @@ rust-version = { workspace = true }
 description = "Fuel Block Parallel Executor"
 
 [features]
+fault-proving = ["fuel-core-types/fault-proving"]
```
MitchTurner, line 14: Seems orthogonal to a parallel executor haha.

crates/services/parallel-executor/src/column_adapter.rs
```diff
@@ -0,0 +1,41 @@
+use fuel_core_storage::column::Column;
+
+pub(crate) struct ContractColumnsIterator {
+    index: usize,
+}
```
MitchTurner: I'm not sure what a "Contract Column" is. Leaving not to come back.

MitchTurner (commented):
wip

crates/services/parallel-executor/src/executor.rs
```diff
@@ -17,98 +50,349 @@ use fuel_core_types::{
...
+            .run(
+                components,
+                da_changes,
+                block_constraints,
+                execution_data.into(),
+            )
+            .await
+    }
+
+    /// Calculate remaining block constraints after L1 execution
```
MitchTurner: nit: 
```suggestion
    /// Calculate block constraints remaining after executing a partial block execution captured in `ExecutionData`
```

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1139 @@
...
+    CoinInBatch,
+};
+
+pub struct Scheduler<R, S, PreconfirmationSender> {
+    /// Config
+    config: Config,
+    /// Storage
+    pub(crate) storage: S,
+    /// Runtime to run the workers
+    runtime: Option<Runtime>,
```
MitchTurner: I'm trying to find some articles about this, but I can't find anything specific talking about it. From what I understand, there are issues that can occur if you are executing a runtime within a different runtime. i.e. I see we are calling `spawn` on this `runtime` inside of `self.execute_batch` which is called by `async fn run`.

There is no reason to believe these are the same/are compatible runtimes. Perhaps we should pass a `Handle` here? That still doesn't guarantee that it is the same runtime, but it at least requires an outer runtime to take responsibility for this.

I'm gonna dig into this more, but it doesn't seem to best practice in general to have two runtimes at the same time (And we should especially not have it unknown to the caller).

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1139 @@
...
+    used_gas: u64,
+    /// used tx size
+    used_size: u32,
+    /// coinbase
+    coinbase: u64,
+}
+
+/// Error type for the scheduler
+#[derive(Debug, derive_more::Display)]
+
```
MitchTurner:
```suggestion
```

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1139 @@
...
+    pub transactions: Vec<CheckedTransaction>,
+    pub gas: u64,
+    pub blob_transactions: Vec<CheckedTransaction>,
+    // Separated from the other gas because this need to be deduced to the global one and not a core one
+    pub blob_gas: u64,
+    pub total_size: u32,
+    pub contracts_used: Vec<ContractId>,
+    pub coins_used: Vec<CoinInBatch>,
+    pub number_of_transactions: u16,
+}
```
MitchTurner, line 295: Might it make sense to make this into an `enum` if `transactions` and `blob_transactions` won't be used at the same time? Maybe they will?

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1139 @@
...
+            ExecutionOptions {
+                forbid_fake_coins: false,
+                backtrace: false,
+            },
+            ConsensusParameters::default(),
+            NoWaitTxs,
+            preconfirmation_sender,
+            true, // dry run
+        )
+        .unwrap();
```
MitchTurner: We shouldn't have any `unwrap`s in our prod code :P

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1139 @@
...
+                    initial_gas_per_worker,
+                    block_constraints.total_execution_time,
+                )?;
+                let batch_len = batch.number_of_transactions;
+
+                if batch.transactions.is_empty() {
+                    self.blob_transactions
+                        .extend(batch.blob_transactions.into_iter());
+                    continue 'outer;
+                }
```
MitchTurner, line 409: Do we want to `continue` immediately or should we add a small `sleep` so it doesn't spin?

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1139 @@
...
+            .ok_or(SchedulerError::InternalError(
+                "L1 transactions consumed all the gas".to_string(),
+            ))?
+            .checked_div(self.config.number_of_cores.get() as u64)
+            .ok_or(SchedulerError::InternalError(
+                "Invalid block gas limit".to_string(),
+            ))?;
+
+        'outer: loop {
+            if self.is_worker_idling() {
```
MitchTurner, line 392: Maybe move each of these paths to helper functions. `run` is getting too long.

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1139 @@
...
+            .checked_sub(l1_execution_data.tx_count)
+            .ok_or(SchedulerError::InternalError(
+                "Cannot insert more transactions: tx_count full".to_string(),
+            ))?;
+        self.tx_size_left = block_constraints
+            .block_transaction_size_limit
+            .checked_sub(l1_execution_data.used_size)
+            .ok_or(SchedulerError::InternalError(
+                "Cannot insert more transactions: tx_size full".to_string(),
+            ))?;
```
MitchTurner: Just trying to find ways to shorten this function. 

Maybe move these into single `update_constraints` helper.

Probably other similar helpers to add here.

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1139 @@
...
+            } else {
+                tokio::select! {
+                    _ = new_tx_notifier.notified() => {
+                        self.new_executable_transactions();
+                    }
+                    result = self.current_execution_tasks.select_next_some() => {
+                        match result {
+                            Ok(res) => {
+                                if !res.skipped_tx.is_empty() {
+                                    self.sequential_fallback(block_height, &consensus_parameters, res.batch_id, res.txs, res.coins_used, res.coins_created).await?;
```
MitchTurner: I want to understand this better, why having skipped txs requires us to run sequentially. Seems like a DoS vector.

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1139 @@
...
+            self.gas_left
+                .saturating_div(self.config.number_of_cores.get() as u64),
+        );
+
+        let (batch, filtered, filter) = tx_source.get_executable_transactions(
+            current_gas,
+            self.tx_left,
+            self.tx_size_left,
+            Filter {
+                excluded_contract_ids: std::mem::take(
```
MitchTurner: is `take` right here? What if we call `ask_new_transactions_batch` again and we are still executing those contracts?

MitchTurner:
```rs
self.current_executing_contracts = filter.excluded_contract_ids;
```

I see this directly after, so we repopulate it, so maybe it's fine. Hard for me to know how `TransactionsSource` behaves though :)

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1139 @@
...
+            .gas_left
+            .saturating_sub(
+                prepared_batch
+                    .gas
+                    .saturating_div(self.config.number_of_cores.get() as u64),
+            )
+            .saturating_sub(prepared_batch.blob_gas);
+        self.tx_left = self
+            .tx_left
+            .saturating_sub(prepared_batch.number_of_transactions);
```
MitchTurner: Could be `update_constraints` helper. DRY with the earlier instance?

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1139 @@
...
+                                consensus_parameters_version,
+                            ),
+                            coinbase_recipient,
+                            gas_price,
+                        },
+                        storage_tx,
+                        &mut execution_data,
+                    )
+                    .await
+                    .unwrap();
```
MitchTurner: `unwrap`

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1139 @@
...
+        // Need future proof
+        let mut tmp_contracts_changes = HashMap::default();
+        for column in ContractColumnsIterator::new() {
+            let column = column.as_u32();
+            if let Some(changes) = res.changes.remove(&column) {
+                tmp_contracts_changes.insert(column, changes);
+            }
+        }
+        self.contracts_changes
+            .add_changes(res.contracts_used.as_ref(), tmp_contracts_changes);
```
MitchTurner: `self.store_any_contract_changes` would be a nice helper here to help reader understand this declaratively.

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1139 @@
...
+        if now.elapsed() > tolerance_execution_time_overflow {
+            tracing::warn!(
+                "Execution time exceeded the limit by: {}ms",
+                now.elapsed().as_millis()
+            );
+        }
+        Ok(())
+    }
+
+    fn verify_coherency_and_merge_results(
```
MitchTurner, line 800: It would be nice to have some robust (prop?) tests to make sure everything is getting included here.

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1139 @@
...
+        consensus_params: &ConsensusParameters,
+        batch_id: usize,
+        txs: Vec<Transaction>,
+        coins_used: Vec<CoinInBatch>,
+        coins_created: Vec<CoinInBatch>,
+    ) -> Result<(), SchedulerError> {
+        let current_execution_tasks = std::mem::take(&mut self.current_execution_tasks);
+        let mut lower_batch_id = batch_id;
+        let mut higher_batch_id = batch_id;
+        let mut all_txs_by_batch_id = FxHashMap::default();
```
MitchTurner, line 942: Why are we using `FxHashMap`? Not familiar with that one.

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1139 @@
...
+                                SchedulerError::InternalError(format!(
+                                    "Failed to convert transaction to checked: {e:?}"
+                                ))
+                            })?
+                            .into(),
+                    );
+                }
+                all_coins_created.extend(coins_created);
+                all_coins_used.extend(coins_used);
+            } else if let Some(res) = self.execution_results.remove(&id) {
```
MitchTurner, line 1001: It's not obvious to me that this is mutually exclusive to 
```            
if let Some((txs, coins_created, coins_used)) =
                all_txs_by_batch_id.remove(&id)
```
Should we remove it from here regardless?

crates/services/parallel-executor/src/scheduler/contracts_changes.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub fn new() -> Self {
+        Self {
+            contracts_changes: FxHashMap::default(),
+            changes_storage: FxHashMap::default(),
+            latest_id: 0,
+        }
+    }
+
+    pub fn add_changes(&mut self, contract_ids: &[ContractId], changes: Changes) {
+        let id = self.latest_id;
```
MitchTurner: Should this and `latest_id` be `index` instead of `id`? We are incrementing them like indices. It's just confusing having this `id` and the `contract_id` together.

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -601,3 +611,67 @@ async fn execute__utxo_ordering_kept() {
         tx2.id(&ChainId::default())
     );
 }
+
+// We use the overflow of gas to skip the transactions.
+#[tokio::test]
+async fn test_skipped_txs_fallback_mechanism() {
```
MitchTurner: nit: is this still an `execute__` test? technically it is `produce_without_commit_with_source`, but those are the same thing in a way.

crates/services/parallel-executor/Cargo.toml
AurelienFT: Yeah it's annoying but needed for the signature of the `generate` function

crates/services/parallel-executor/src/column_adapter.rs
AurelienFT: We discussed it I hope it's more clear

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: The goal of having a runtime just for this is to be able to give him constraints on the number of cores using etc...

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: They are provided by the transaction pool in the same time and we split them ourself. It could be two separate batch with two different properties but the blob_transaction only have a vec and the gas

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: Removed.

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: It shouldn't spin because `is_worker_idling` should return false because we didn't had any transaction in the privous loop turn

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: It can't really be a DoS because it can't be generated by the user. We assume that everything has be checked before in txpool.

We need to run the skipped transactions sequentially again because their position in the block determined by the amount of transactions in previous batches is now wrong because one of the tx will not be included. This would creates invalid data on elements filled during execution (such as tx_pointer on coin) and so we need to invalidate all current execution and redo without this skipped transaction

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: Yeah Green wanted to avoid the copy of data and we couldn't pass a reference so we used a move

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: @rymnc addition. I think it's for optimization. @rymnc if you can explain more.

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: Because the batches that are in `all_txs_by_batch_id` are the one from currently executing batches and in `self.execution_results.remove(&id)` it's the batches that have already finished their execution (but could have index after the one that contain skipped tx)

crates/services/parallel-executor/src/scheduler/contracts_changes.rs
AurelienFT: Done

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: Done.

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: Done

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: Done

crates/services/parallel-executor/src/scheduler.rs
MitchTurner: > The goal of having a runtime just for this is to be able to give him constraints on the number of cores using etc...

We call `runtime.spawn` on `self.runtime` inside of `execute_batch`.  If we want info on the runtime, we should create a limited interface with just that info (see [ISP](https://en.wikipedia.org/wiki/Interface_segregation_principle)), but it looks like we are doing more than just checking the cores with the runtime.

Here is an explanation of what I'm talking about:
https://morestina.net/blog/1686/rust-async-is-colored

tl;dr It has less to do with different runtimes and more to do with how async and sync code run and how it's not clear from the signature what blocking risks the code has.

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1179 @@
...
+        components: &Components<TxSource>,
+        mut batch: PreparedBatch,
+        batch_id: usize,
+        start_idx_txs: u16,
+        storage_with_da: Arc<StorageTransaction<View>>,
+    ) -> Result<(), SchedulerError> {
+        let worker_id = self.current_available_workers.pop_front().ok_or(
+            SchedulerError::InternalError("No available workers".to_string()),
+        )?;
+        let runtime = self.runtime.as_ref().unwrap();
```
MitchTurner: unwrap

crates/services/parallel-executor/src/scheduler.rs
MitchTurner: Hmm. I'm sure you've thought about this a lot and understand the problem well. Don't we only need the tx_pointer for spending the coin, not for the tx. Txs are atomic and don't depend on the block as a whole--that's why predicates work deterministically. But I'm sure I'm missing something.

AurelienFT: Yes but even if they are not used directly they need to be set correctly + a transaction in the same block can use the coin.

crates/services/executor/src/executor.rs
```diff
@@ -547,7 +547,11 @@ where
     N: NewTxWaiterPort,
     P: PreconfirmationSenderPort,
 {
-    async fn execute<TxSource, D>(
+    pub fn set_consensus_params(&mut self, consensus_params: ConsensusParameters) {
+        self.consensus_params = consensus_params;
+    }
```
xgreenx: Why can't we create a new executor for each thread? It looks like it still will be the same in term on performance, because you will clone all fields regardless

crates/services/executor/src/executor.rs
```diff
@@ -700,7 +705,35 @@ where
...
+        mut block_storage_tx: BlockStorageTransaction<D>,
+        execution_data: &mut ExecutionData,
+    ) -> ExecutorResult<PartialFuelBlock>
+    where
+        TxSource: TransactionsSource,
+        D: KeyValueInspect<Column = Column>,
+    {
+        let mut partial_block =
+            PartialFuelBlock::new(transactions.header_to_produce, vec![]);
+        let mut memory = MemoryInstance::new();
```
xgreenx: I would suggest to pass `MemoryInstance` from the scheduler, because it is possible that each transaction will require huge allocations inside of VM.

crates/services/parallel-executor/src/scheduler/contracts_changes.rs
```diff
@@ -0,0 +1,59 @@
...
+    ) -> Option<(Vec<ContractId>, Changes)> {
+        let id = self.contracts_changes.remove(contract_id)?;
+        let (contract_ids, changes) = self.changes_storage.remove(&id)?;
+        for contract_id in contract_ids.iter() {
+            self.contracts_changes.remove(contract_id);
+        }
+        Some((contract_ids, changes))
+    }
+
+    pub fn extract_all_contracts_changes(&mut self) -> Vec<Changes> {
```
xgreenx: Maybe you want to use `self` here? Just a thought

crates/services/parallel-executor/src/scheduler/contracts_changes.rs
```diff
@@ -0,0 +1,59 @@
...
+        let mut changes = vec![];
+        for id in 0..self.latest_index {
+            if let Some((_, change)) = self.changes_storage.remove(&id) {
+                changes.push(change);
+            }
+        }
+        self.contracts_changes.clear();
```
xgreenx:
```suggestion
        self.clear();
```

crates/services/parallel-executor/src/scheduler/coin.rs
```diff
@@ -0,0 +1,230 @@
...
+        &self.owner
+    }
+
+    pub(crate) fn amount(&self) -> &Word {
+        &self.amount
+    }
+
+    pub(crate) fn asset_id(&self) -> &AssetId {
+        &self.asset_id
+    }
```
xgreenx, line 81: Why do you need getters if fields are public?

crates/services/parallel-executor/src/scheduler/coin.rs
```diff
@@ -0,0 +1,230 @@
...
+            owner: *owner,
+            amount: *amount,
+            asset_id: *asset_id,
+        }
+    }
+
+    pub(crate) fn equal_compressed_coin(&self, compressed_coin: &CompressedCoin) -> bool {
+        match compressed_coin {
+            CompressedCoin::V1(coin) => {
+                self.owner() == &coin.owner
```
xgreenx, line 132: Do we need to compare Utxo here?

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1179 @@
...
+            skipped_txs,
+            transactions_status,
+            changes: StorageChanges::default(),
+            used_gas,
+            used_size,
+            coinbase,
+        };
+
+        let mut storage_changes = vec![];
+        let latest_view = self
```
xgreenx: You need to use storage with DA changes here. Because FTI may create some coins

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1179 @@
...
+                                consensus_parameters_version,
+                            ),
+                            coinbase_recipient,
+                            gas_price,
+                        },
+                        storage_tx,
+                        &mut execution_data,
+                    )
+                    .await?;
+                // TODO: Outputs seems to not be resolved here, why? It should be the case need to investigate
```
xgreenx: Known and dynamic coins must be resolved at this point. We insert into the block tx returned by VM. In the case of successful execution, `Variable` and `Change` outputs should be filled

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1179 @@
...
+        self.current_available_workers.push_back(res.worker_id);
+    }
+
+    fn store_any_contract_changes(
+        &mut self,
+        mut changes: Changes,
+        contracts_used: &[ContractId],
+    ) -> Changes {
+        // Is it useful ?
+        // Did I listed all column ?
```
xgreenx: Yes, you listed all columns related to contracts

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1179 @@
...
+            l1_execution_data.tx_count,
+            l1_execution_data.used_size,
+            l1_execution_data.used_gas,
+        )?;
+
+        let consensus_parameters_version =
+            components.header_to_produce.consensus_parameters_version;
+        let block_height = *components.header_to_produce.height();
+
+        let consensus_parameters = {
```
xgreenx: You can use `storage_with_da` to get this info=)

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1179 @@
...
+        let mut res = self.verify_coherency_and_merge_results(
+            nb_batch_created,
+            components.header_to_produce,
+            l1_execution_data,
+        )?;
+
+        if self.blob_transactions.is_empty() {
+            let mut merged: Changes = Changes::default();
+            for changes in res.changes.extract_list_of_changes() {
+                merged.extend(changes);
```
xgreenx: Don't use extend for `Changes`. It overrides internal `BTreeMap`. Instead I would suggest to use `InMemoryTransaction` and `commit_changes` function to merge results together.


image

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1179 @@
...
+        )
+        .await?;
+
+        let mut res = self.verify_coherency_and_merge_results(
+            nb_batch_created,
+            components.header_to_produce,
+            l1_execution_data,
+        )?;
+
+        if self.blob_transactions.is_empty() {
```
xgreenx:
```suggestion
        if !self.blob_transactions.is_empty() {
```

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1179 @@
...
+                    components,
+                    storage_with_res,
+                    nb_transactions,
+                    consensus_parameters_version,
+                )
+                .await?;
+            res.add_blob_execution_data(blob_execution_data, blob_txs);
+        }
+
+        self.reset();
```
xgreenx: Maybe it will be better if `Scheduler` were created per block. So in this case you don't need rest functionality and you can just consume `self` instead of `&mut self`.

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1179 @@
...
+
+        let mut res = self.verify_coherency_and_merge_results(
+            nb_batch_created,
+            components.header_to_produce,
+            l1_execution_data,
+        )?;
+
+        if self.blob_transactions.is_empty() {
+            let mut merged: Changes = Changes::default();
+            for changes in res.changes.extract_list_of_changes() {
```
xgreenx, line 482: Hmm, merging everything together will increase execution time in the case of blobs. Do we really need to merge everything? Maybe we can revisit the implementation for `InMemoryStorage` with the list of `Changes`? Maybe it is already good enough and will not fail, or do double spend?

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1179 @@
...
+            }
+        }
+
+        let prepared_batch = prepare_transactions_batch(batch)?;
+        self.update_constraints(
+            prepared_batch.number_of_transactions,
+            prepared_batch.total_size,
+            prepared_batch
+                .gas
+                .saturating_div(self.config.number_of_cores.get() as u64)
```
xgreenx: Why do we need to divide here? I thought we want to track the total gas left

xgreenx (commented):
The general logic looks good, I will do the second round of review tomorrow with a fresh head=)

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: I'm having trouble using the functions of the `Storage` port trait on the InMemoryTransaction that are on top of the struct that implement `Storage` that's why I did that but I can change it if you can help me on this.

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: I'm having trouble using the functions of the `Storage` port trait on the InMemoryTransaction that are on top of the struct that implement `Storage` that's why I did that but I can change it if you can help me on this.

crates/services/parallel-executor/src/scheduler/contracts_changes.rs
AurelienFT: It lives for the duration of the executor and so drop it before would require an option in the scheduler etc.. I will leave it with a reference and the scheduler will drop it with himself.

crates/services/executor/src/executor.rs
AurelienFT: Done.

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: Modfied the whole flow of executor and scheduler lifetime and not created here anymore.

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: Done.

crates/services/executor/src/executor.rs
AurelienFT: I added it but I clone for each batch of transaction so might make a big difference, should I make a memory pool such as in txpool ?

crates/services/parallel-executor/src/scheduler/coin.rs
AurelienFT: I wanted to but compressed coin in db doesn't have the utxoid (especially the txid) it just have a block height and a tx index.

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: I need to make new tests to be sure but I tested and it was not working. Will make need tests and investigate if still error

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: Yeah right we are already dividing above to get the gas. The problem is that we are also dividing the gas for blobs which should consume gas for all cores so I added a multiply for the blobs. If it's not clear we can discuss it.

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: I did that (and [here](https://github.com/FuelLabs/fuel-core/blob/8364e0ad124841414ac4eb1ac184313077fbf2e5/crates/services/parallel-executor/src/scheduler.rs#L628) also) because the changes contained in this structure shouldn't override each other because they modify different contracts and they are changes that contains only contract column related

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: "Maybe we can revisit the implementation for InMemoryStorage with the list of Changes" I'm ok to do that but I think it's valuable to peer on that because I have no clue on how to do it.

crates/services/parallel-executor/src/scheduler.rs
rymnc: it's faster, it uses a non-cryptographic hash function, which is OK in this case.

crates/services/parallel-executor/src/scheduler/coin.rs
AurelienFT: Utxo id is verified on usage of the function because the compressed coin passed as parameter is retrieve with utxo

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: Changed to use tables directly

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: Understood now. Modified

crates/services/parallel-executor/src/scheduler.rs
MitchTurner: > a transaction in the same block can use the coin

Ah. I hadn't considered this case. 

Can we not assign their place in the block after execution? A branch shouldn't have dependent txs, right?

MitchTurner:
Do we want the parallel code to be behind a feature flag? Or we just moving over wholesale?

crates/services/executor/src/executor.rs
AurelienFT: Done.

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: I added a test for this. Its working, comment removed.

crates/services/executor/src/executor.rs
```diff
@@ -336,8 +336,11 @@ impl ExecutionData {
 #[derive(serde::Serialize, serde::Deserialize, Clone, Default, Debug)]
 pub struct ExecutionOptions {
     /// The flag allows the usage of fake coins in the inputs of the transaction.
-    /// When `false` the executor skips signature and UTXO existence checks.
-    pub forbid_fake_coins: bool,
+    /// When `false` the executor skips signature.
```
xgreenx: Comment must be updated=)

crates/services/executor/src/executor.rs
```diff
@@ -349,8 +352,11 @@ pub struct ExecutionOptions {
 #[derive(Clone, Default, Debug)]
 struct ExecutionOptionsInner {
     /// The flag allows the usage of fake coins in the inputs of the transaction.
-    /// When `false` the executor skips signature and UTXO existence checks.
-    pub forbid_fake_coins: bool,
+    /// When `false` the executor skips signature.
```
xgreenx: Comment must be udpated=)

crates/services/executor/src/executor.rs
```diff
@@ -700,7 +708,35 @@ where
...
     }
 
-    fn process_l1_txs<T>(
+    pub async fn execute_l2_transactions<TxSource, D>(
+        mut self,
+        transactions: Components<TxSource>,
+        mut block_storage_tx: BlockStorageTransaction<D>,
+        execution_data: &mut ExecutionData,
+        memory: &mut MemoryInstance,
+    ) -> ExecutorResult<PartialFuelBlock>
```
xgreenx: I think it is better to return `(Vec<Transaction>, ExecutionData)`, instead of passing empty `ExecutionData` and returning not used `PartialFuelBlock`=)

crates/services/executor/src/executor.rs
```diff
@@ -1793,15 +1829,21 @@ where
...
-        debug_assert!(checked_tx.checks().contains(Checks::Signatures));
+        if self.options.forbid_fake_utxo {
+            self.verify_inputs_exist_and_values_match(
+                storage_tx,
+                checked_tx.transaction().inputs(),
+                header.da_height,
+            )?;
+        }
+
+        if self.options.forbid_fake_signature {
```
xgreenx: I think we should rename `forbid_fake_signature` into `forbid_unauthorized_inputs`, and also check the `check_predicates` part.

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -143,9 +107,10 @@ where
         storage: &mut Storage,
         amount: u64,
     ) -> &mut Self {
-        let predicate = op::ret(RegId::ONE).to_bytes().to_vec();
-        let owner = Input::predicate_owner(&predicate);
         let utxo_id: UtxoId = rng.r#gen();
+        let secret_key = SecretKey::default();
```
xgreenx: I think it will be better to generate it randomly. Plus, zero key private key is not allowed, form what I know=)

crates/services/parallel-executor/src/tests/tests_executor.rs
```diff
@@ -601,3 +570,126 @@ async fn execute__utxo_ordering_kept() {
...
+    let result = future.await.unwrap().into_result();
+    let transactions = result.block.transactions();
+    assert_eq!(transactions.len(), 2);
+    let output = transactions[0].outputs().into_owned()[0];
+    assert_eq!(output.amount(), Some(1000));
+}
+
+// We use the overflow of gas to skip the transactions.
+// TODO: This test can't be performed anymore now that we lower the gas ourself in the scheduler and so
+// scheduler fails before the executor can skip the transaction.
```
xgreenx: I thought we agreed to use an incorrectly estimated predicate to make it fail, instead of setting gas more than block

crates/services/parallel-executor/src/executor.rs
```diff
@@ -17,98 +56,316 @@ use fuel_core_types::{
...
+            used_gas,
+            used_size,
+        };
+
+        executor.produce_mint_tx(
+            &mut partial_block,
+            components,
+            &mut tx_changes,
+            &mut execution_data,
+            memory,
```
xgreenx:
```suggestion
            MemoryInstance::new(),
```

crates/services/parallel-executor/src/executor.rs
```diff
@@ -17,98 +56,316 @@ use fuel_core_types::{
...
+        let mut execution_data = ExecutionData {
+            coinbase,
+            skipped_transactions: skipped_txs,
+            events,
+            changes: Default::default(),
+            message_ids,
+            tx_count,
+            tx_status: transactions_status,
+            found_mint: false,
+            event_inbox_root: Default::default(),
```
xgreenx: It must be set to be from the L1 part.

crates/services/parallel-executor/src/executor.rs
```diff
@@ -17,98 +56,316 @@ use fuel_core_types::{
...
+
+        // Produce mint transaction (pass the entire scheduler_result)
+        let (execution_data, storage_changes, partial_block) =
+            self.produce_mint_tx(components, scheduler_result, memory, view, executor)?;
+
+        // Generate final block
+        let block = partial_block
+            .generate(
+                &execution_data.message_ids,
+                Default::default(),
```
xgreenx: It must be set to be from the L1 part.

crates/services/parallel-executor/src/executor.rs
```diff
@@ -17,98 +56,316 @@ use fuel_core_types::{
...
+                })?
+                .into_owned()
+        };
+        let scheduler = Scheduler::new(
+            self.config.clone(),
+            self.relayer.clone(),
+            self.storage.clone(),
+            self.preconfirmation_sender.clone(),
+            consensus_parameters,
+            Duration::from_millis(300),
```
xgreenx: That one should be configurable from outside

crates/services/parallel-executor/src/executor.rs
```diff
@@ -17,98 +56,316 @@ use fuel_core_types::{
...
+        )?;
+
+        let mut executor = scheduler.create_executor()?;
```
xgreenx: It would be nice if we didn't create executor from the scheduler, and instead we executed L1 transaction either inside of the scheduler, or before scheduler is created=)

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1211 @@
...
+        blob_txs: Vec<Transaction>,
+    ) {
+        self.transactions.extend(blob_txs);
+        self.events.extend(blob_execution_data.events);
+        self.message_ids.extend(blob_execution_data.message_ids);
+        self.skipped_txs
+            .extend(blob_execution_data.skipped_transactions);
+        self.transactions_status
+            .extend(blob_execution_data.tx_status);
+        // Should contains all the changes from all executions
```
xgreenx, line 271: Can we add `debug_assert` please to verify that `self.changes` is empty?=D

crates/services/parallel-executor/src/scheduler.rs
xgreenx: I just want to highlight that `runtime.block_on` and `block_on` on the current runtime are two different things. The article describes that calling `spawn_blocking` and `block_on` from the same runtime may cause a deadlock, because you do manipulations over the same scheduler. But in our case we do `block_on` on another runtime and it affect only another scheduler, so it should be safe for the current runtime from which we call `run_sheduler`.

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1211 @@
...
+        Ok(Self {
+            runtime: Some(runtime),
+            relayer,
+            preconfirmation_sender,
+            storage,
+            // TODO: Use consensus parameters after https://github.com/FuelLabs/fuel-vm/pull/905 is merged
+            tx_left: u16::MAX,
+            tx_size_left: consensus_parameters.block_transaction_size_limit(),
+            gas_left: consensus_parameters.block_gas_limit(),
+            current_available_workers: ((0..config.number_of_cores.get())
```
xgreenx: It would be nice if we had a memory pool on the parallel executor level, and sheduler will borrow memory instances until the execution, and later return them back.

In this case, we can avoid reallocations between blocks.

crates/services/executor/src/executor.rs
AurelienFT: I don't understand what you means by "passing empty ExecutionData" I pass the `ExecutionData` with a value in the field `tx_count` to have the correct offset in transaction index. I can change the `PartialFuelBlock` to `Vec<Transaction>` tho.

crates/services/executor/src/executor.rs
AurelienFT: Done.

crates/services/executor/src/executor.rs
AurelienFT: Done

crates/services/executor/src/executor.rs
AurelienFT: Done

crates/services/parallel-executor/src/tests/tests_executor.rs
AurelienFT: Done.

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: Done.

crates/services/parallel-executor/src/tests/tests_executor.rs
AurelienFT: Yep that's why it's not passing now. I didn't made the change yet. Now it's done.

crates/services/parallel-executor/src/executor.rs
AurelienFT: Done

crates/services/parallel-executor/src/executor.rs
AurelienFT: Done.

crates/services/parallel-executor/src/executor.rs
AurelienFT: Done.

crates/services/parallel-executor/src/executor.rs
AurelienFT: Done.

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: Done.

crates/services/executor/src/executor.rs
AurelienFT: I changed the argument to pass directly the idx.

crates/services/parallel-executor/src/executor.rs
```diff
@@ -17,98 +59,362 @@ use fuel_core_types::{
...
+            self.relayer.clone(),
+            ExecutionOptions {
+                forbid_unauthorized_inputs: true,
+                forbid_fake_utxo: false,
+                backtrace: false,
+            },
+            consensus_parameters.clone(),
+            NoWaitTxs,
+            self.preconfirmation_sender.clone(),
+            true, // dry run
```
xgreenx: `dry_run` flag must be `false`, because we don't do some checks during dry run.

crates/services/parallel-executor/src/executor.rs
```diff
@@ -17,98 +59,362 @@ use fuel_core_types::{
...
-        if let Some(runtime) = self.runtime.take() {
-            runtime.shutdown_background();
-        }
-    }
+pub struct Executor<S, R, P> {
+    config: Config,
+    relayer: R,
+    storage: S,
+    preconfirmation_sender: P,
+    memory_pool: Option<Vec<MemoryInstance>>,
```
xgreenx: Why is it `Option`?=)

crates/services/parallel-executor/src/executor.rs
```diff
@@ -17,98 +59,362 @@ use fuel_core_types::{
...
+            consensus_parameters,
+            maximum_execution_time,
+        )?;
+
+        let (res, memory_instance) = scheduler
+            .run(components, da_changes, execution_data.into())
+            .await?;
+
+        // Restore memory pool to be re-used
+        self.memory_pool = Some(memory_instance);
```
xgreenx: In the case of error, you will not restore it=) I think it is better to reuse the memory pool with auto return `MemoryInstance` on `drop` as we do in other places.

In this cases we don't need to manage them manually and worry that we will lose them somewhere

crates/services/executor/src/executor.rs
```diff
@@ -335,9 +335,12 @@ impl ExecutionData {
...
+    /// When `false` the executor skips signature and predicate checks.
+    pub forbid_unauthorized_inputs: bool,
     /// The flag allows the usage of fake coins in the inputs of the transaction.
-    /// When `false` the executor skips signature and UTXO existence checks.
-    pub forbid_fake_coins: bool,
+    /// When `false` the executor skips UTXO existence checks.
+    pub forbid_fake_utxo: bool,
     /// Print execution backtraces if transaction execution reverts.
     ///
     /// Deprecated field. Do nothing. This fields exists for serialization and
```
xgreenx: This type is also used in the WASM during deserialziation of the input execution. 

We should update the deserialization type to have only one `utxo_valdiaiton: bool` as before. Otherwise, we need to create a new `V2` type, and then this change will be breaking.


image

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1201 @@
...
+                        *asset_id,
+                    ));
+                }
+                Output::Variable {
+                    to,
+                    amount,
+                    asset_id,
+                } => {
+                    coins.push(CoinInBatch::from_output(
+                        UtxoId::new(tx_id, output_idx as u16),
```
xgreenx: I think we need to create a task to add before we switch to a parallel executor in production. Because code here should return an error in the case if it is not possible to convert it to `u16`

```
```

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1201 @@
...
+
+    for (idx, tx) in batch.into_iter().enumerate() {
+        let tx_id = tx.id();
+        let inputs = tx.inputs();
+        for input in inputs.iter() {
+            match input {
+                fuel_core_types::fuel_tx::Input::Contract(contract) => {
+                    prepared_batch.contracts_used.push(contract.contract_id);
+                }
+                fuel_core_types::fuel_tx::Input::CoinSigned(coin) => {
```
xgreenx, line 1088: I think we also need to take into account messages, because we need to ensure that we don't have double spend

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1201 @@
...
+        let is_blob = matches!(&tx, CheckedTransaction::Blob(_));
+        prepared_batch.total_size += tx.size() as u64;
+        prepared_batch.number_of_transactions += 1;
+        if is_blob {
+            prepared_batch.blob_gas += CheckedTransactionExt::max_gas(&tx)?;
+            prepared_batch.blob_transactions.push(tx);
+        } else {
+            prepared_batch.gas += CheckedTransactionExt::max_gas(&tx)?;
+            prepared_batch.transactions.push(tx);
+        }
```
xgreenx, line 1131:
```suggestion
        let max_gas = CheckedTransactionExt::max_gas(&tx)?;
        if is_blob {
            prepared_batch.blob_gas += max_gas;
            prepared_batch.blob_transactions.push(tx);
        } else {
            prepared_batch.gas += max_gas;
            prepared_batch.transactions.push(tx);
        }
```

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1201 @@
...
+            self.current_available_workers.pop_front().ok_or(
+                SchedulerError::InternalError("No available workers".to_string()),
+            )?;
+        let (transactions, execution_data) = executor
+            .execute_l2_transactions(
+                Components {
+                    header_to_produce: PartialBlockHeader::default(),
+                    transactions_source: OnceTransactionsSource::new(all_txs, 0),
+                    coinbase_recipient: Default::default(),
+                    gas_price: Default::default(),
```
xgreenx: I think we should use the actual data here, like header, gas price, coinbase account, vecrions of consensus parameters

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1201 @@
...
+                skipped_tx: vec![],
+                used_gas: execution_data.used_gas,
+                used_size: execution_data.used_size,
+                coinbase: execution_data.coinbase,
+            },
+        );
+
+        // Register the worker back to the available workers
+        self.current_available_workers
+            .push_back((worker_id, memory_instance));
```
xgreenx: I know that scheduler is a one-time live object currently, but I think we still should guarantee a valid state of it even in the case of error.

If an error occurs above, you will not insert the worker and memory instance back. It would be nice if we can have a pool, where each object will be returned automatically on `drop`.

crates/services/parallel-executor/src/scheduler.rs
```diff
@@ -0,0 +1,1201 @@
...
+                            "coinbase has overflowed u64".to_string(),
+                        )
+                    })?;
+            } else {
+                return Err(SchedulerError::InternalError(format!(
+                    "Batch {batch_id} not found in the execution results"
+                )));
+            }
+        }
+        exec_result.header = partial_block_header;
```
xgreenx: It is done during the construction of `exec_result`

```suggestion
```

crates/services/parallel-executor/src/scheduler.rs
xgreenx: Have you looked into that one after our disucssion?=)

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: Done.

crates/services/parallel-executor/src/executor.rs
AurelienFT: Done.

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: I added the 3 lines and fixed all errors related to it.

rymnc (approved):

crates/services/executor/src/executor.rs
AurelienFT: Done

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: It seems that there is no utxo id in this so I don't really understand how we can catch double spend on that and how it can fit in our data structure. But your comment made me realize I didn't added a double spend check on coins so I added it.

crates/services/parallel-executor/src/executor.rs
AurelienFT: Removed.

crates/services/parallel-executor/src/executor.rs
AurelienFT: Done.

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: I have changed the memory management. I tried to make it generic to use the same for workers but couldn't because memory instance needs a call to `.reset()` which is specific. So I made a simple version for workers also. Done.

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: I didn't found any simple solutions to do it so I dropped it.

crates/services/parallel-executor/src/scheduler.rs
AurelienFT: I check nonce duplication. Done

PR #2982, opened 2025-04-29
fix(tests): flakiness across a few different tests

rymnc:
> [!NOTE]
> discovered the issue, bob doesn't have the delegate keys, and therefore when he receives preconfirmations he can't verify the signature, and then rejects the messages. the rejections reduce the peer score of the authority node and then the two are disconnected. maybe the preconfirmation service should just `GossipsubMessageAcceptance::Ignore` if it has no delegate keys instead of `GossipsubMessageAcceptance::Reject`. 

focusing on improving the configuration of nodes, enhancing consistency checks, and cleaning up unused imports.

### Configuration Enhancements:
* Added a new configuration option `pre_confirmation_signature_service.echo_delegation_interval` to set an interval of 100 milliseconds for echo delegation in `make_nodes`. This improves configurability for testing scenarios.
* Introduced conditional database initialization in `make_config`, using `CombinedDatabase` when the `default` feature is enabled, and falling back to an in-memory database otherwise. This provides flexibility for different environments.

### Consistency Check Improvements:
* Refactored the `consistency` method to use a `tokio::time::interval` for periodic checks, replacing the previous loop and event-based logic. This simplifies the code and ensures consistent behavior.
* Consolidated consistency methods by introducing `consistency_with_duration`, which allows specifying a custom timeout duration. The existing `consistency_10s` and `consistency_20s` methods now delegate to this new method for better code reuse.

### Cleanup:
* Removed an unused import `schema::tx::types::TransactionStatus` to tidy up the codebase.

### DIFF
--- a/crates/fuel-core/src/p2p_test_helpers.rs
+++ b/crates/fuel-core/src/p2p_test_helpers.rs
@@ -630,6 +630,12 @@ impl Node {
             .await;
     }
 
+    /// Wait for the node to reach consistency with the given transactions within 30 seconds.
+    pub async fn consistency_30s(&mut self, txs: &HashMap<Bytes32, Transaction>) {
+        self.consistency_with_duration(txs, Duration::from_secs(30))
+            .await;
+    }
+
     /// Insert the test transactions into the node's transaction pool.
     pub async fn insert_txs(&self) -> HashMap<Bytes32, Transaction> {
         let mut expected = HashMap::new();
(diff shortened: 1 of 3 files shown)


rymnc:
I've re-ran the checks 6 times, cannot see the flakiness anymore.

AurelienFT (approved):
Thanks for taking the time to do that

Dentosal (approved):

PR #2987, opened 2025-05-02
Make txpool pre-conf broadcast conditional on there being some txs in the list

MitchTurner:
## Related Issues
closes https://github.com/FuelLabs/fuel-core/issues/2988

## Description

We noticed many messages while running in production that said:
```
│ stamp":"2025-05-01T20:41:55.374539Z","level":"ERROR","fields":{"message":"Got an error during pre-confirmation message broadcasting Duplicate"},"target":"fu │
```
After some finagling I was able to reproduce this in the integ tests and I added additional tracing to find that those duplicate messages were empty (no pre-confirmations were included).

I found that the txpool was sending these pre-confirmation batches because it would just send on every iteration regardless if any were squeezed out or not.

After preventing this from happening, by checking that the batch wasn't empty first, the error log disappeared. 

### Does this solve the problem in production?

In prod, we would see the log message every time a new tx was submitted. This probably was happening due to the `insert_inner` method that would always call:
```rs
            self.tx_status_manager
                .squeezed_out_txs(removed_transactions);
```
and generate another  empty pre-confirmation which would be considered a duplicate.

Just in case, I have left the extra log information in that tracing log which will give us details on the actual duplicate message:
```rs
                        let result = self.p2p_service.publish_message(broadcast.clone());
                        if let Err(e) = result {
                            tracing::error!("Got an error during pre-confirmation message broadcasting {:?}: {}", broadcast, e);
                        }
```

### DIFF
--- a/.changes/fixed/2987.md
+++ b/.changes/fixed/2987.md
@@ -1,0 +1,1 @@
+Make txpool pre-conf broadcast conditional on there being some txs in the list
--- a/crates/services/consensus_module/poa/src/pre_confirmation_signature_service.rs
+++ b/crates/services/consensus_module/poa/src/pre_confirmation_signature_service.rs
@@ -239,16 +239,20 @@ where
             res = self.tx_receiver.receive() => {
                 tracing::debug!("Received transactions");
                 let preconfirmations = try_or_stop!(res);
-                let expiration = self.current_delegate_key.expiration();
-                let pre_confirmations = Preconfirmations {
-                    expiration,
-                    preconfirmations,
-                };
-                let signature = try_or_stop!(self.current_delegate_key.sign(&pre_confirmations));
-                try_or_continue!(
-                    self.broadcast.broadcast_preconfirmations(pre_confirmations, signature).await,
-                    |err| tracing::error!("Failed to broadcast pre-confirmations: {:?}", err)
-                );
+                if preconfirmations.is_empty() {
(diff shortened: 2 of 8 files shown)


AurelienFT (dismissed):
Nice finding I'm ok with the fix

xgreenx (commented):
I think we should fix PoA signature service to not emit a signature in the case of empty pre confirmations array=) In this case we don't need to worry about the caller of the method

rymnc:
we may want to apply this fix on release/v0.43.2 so that we can test it on devnet/testnet in case there are breaking changes between 0.43.2 and master

MitchTurner:
>I think we should fix PoA signature service to not emit a signature in the case of empty pre confirmations array=) In this case we don't need to worry about the caller of the method

In general, I prefer to have the caller be responsible. Maybe we do both. it seems inappropriate to have the service handle a bunch of empty messages, even if it's not  passing it along to P2P.

xgreenx:
> In general, I prefer to have the caller be responsible. Maybe we do both. it seems inappropriate to have the service handle a bunch of empty messages, even if it's not passing it along to P2P.

We can have both the sender and the receiver

xgreenx (approved):
Yep, can confirm that it is empty pre confirmation arrays when we produce a block


image

AurelienFT (approved):

PR #2993, opened 2025-05-06
fix(graphql): pin the version of assets in the graphql playground

rymnc:
## Linked Issues/PRs

playground broken across various networks due to https://github.com/async-graphql/async-graphql/issues/1703

## Description

2 main things - 
- [x] we generate the html body once, and reuse it across calls to `/v1/playground`. previously we were re-rendering the template every time, which might not be a good solution
- [x] in the raw html generated, we pin the versions of the js and css assets distributed by unpkg 

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/fuel-core/src/graphql_api/api_service.rs
+++ b/crates/fuel-core/src/graphql_api/api_service.rs
@@ -83,7 +83,10 @@ use std::{
         TcpListener,
     },
     pin::Pin,
-    sync::Arc,
+    sync::{
+        Arc,
+        OnceLock,
+    },
 };
 use tokio_stream::StreamExt;
 use tower::limit::ConcurrencyLimitLayer;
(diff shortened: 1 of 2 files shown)


crates/fuel-core/src/graphql_api/api_service.rs
```diff
@@ -373,17 +376,43 @@ where
...
+        // this may not be necessary in the future,
+        // but we need it to patch: https://github.com/async-graphql/async-graphql/issues/1703
+        let raw_html = raw_html.replace(
+            "https://unpkg.com/graphiql/graphiql.min.js",
+            "https://unpkg.com/graphiql@3/graphiql.min.js",
+        );
+        let raw_html = raw_html.replace(
+            "https://unpkg.com/graphiql/graphiql.min.css",
+            "https://unpkg.com/graphiql@3/graphiql.min.css",
+        );
```
rymnc, line 403: while we could get away with a singular replace, i.e 
```suggestion
        let raw_html = raw_html.replace(
            "https://unpkg.com/graphiql/graphiql.min.",
            "https://unpkg.com/graphiql@3/graphiql.min.",
        );
```
i would rather have it be explicit about which assets we are pinning, and since this part of the code only runs once, it seems ok imo

netrome: Agreed.

netrome (approved):

PR #3004, opened 2025-05-09
hotfix: More info in Ethereum provider error logs

netrome:
This PR adds additional error logs when we have Ethereum provider failures. We need this specifically because the upstream `QuorumError` doesn't include any provider errors in the `NoQuorumReached` variant:

```
#[derive(Error, Debug)]
/// Error thrown when sending an HTTP request
pub enum QuorumError {
    #[error("No Quorum reached.")]
    /// NoQuorumReached
    NoQuorumReached {
        /// Returned responses
        values: Vec<Value>,
        /// Returned errors
        errors: Vec<ProviderError>,
    },
}
```

However, by adding logs of the debug representation of this error we'll expose this information which can be helpful when debugging quorum failures.


crates/services/relayer/src/service/syncing.rs
```diff
@@ -46,10 +46,13 @@ where
...
-        Err(err) => Err(anyhow::anyhow!(
-            "Failed to check if DA layer is syncing {}",
-            err
-        )),
+        Err(err) => {
+            tracing::error!(?err, "failed to check if DA layer is syncing");
+            Err(anyhow::anyhow!(
+                "Failed to check if DA layer is syncing {}",
+                err
+            ))
```
MitchTurner, line 54: We can probably throw this all into the error, since the whole error will get logged by the service runner.

This works too though.

crates/services/relayer/src/service/get_logs.rs
```diff
@@ -61,16 +61,22 @@ where
...
-                                    logs,
-                                },
-                                page,
-                            ))
-                        })
+                        eth_node
+                            .get_logs(&filter)
+                            .await
+                            .inspect_err(|err| {
+                                tracing::error!(?err, "eth provider failed to get logs")
```
MitchTurner, line 68: This might be easier than trying to pipe the correct error through the stream unlike below.

crates/services/relayer/src/service.rs
```diff
@@ -345,7 +345,7 @@ where
                 Err(anyhow::anyhow!("The relayer got a stop signal"))
             },
             block = self.eth_node.get_block(ethers_core::types::BlockNumber::Finalized) => {
-                let block_number = block.map_err(anyhow::Error::msg)?
+                let block_number = block.inspect_err(|err| tracing::error!(?err, "eth provider failed to get block")).map_err(anyhow::Error::msg)?
```
MitchTurner, line 348: Either way here.

crates/services/relayer/src/service/syncing.rs
netrome: Yeah, I was a bit torn about that. It feels so wrong to put debug information within the actual error, since the canonical display representation is already provided. But, on the other hand it is practical to get the information there because it's currently missing.

However, the deciding factor for me is that the log message is structured, so it's much easier to parse and hide bloated fields in log explorers (although we need to improve ours). But if we bloat the actual message, it ends up polluting the logs a bit more and is more tedious to clean out because typically you'll want to render the message.

I can go either way here though, and need to sleep soon so feel free to update the PR and add this to the log as well.

crates/services/relayer/src/service/get_logs.rs
netrome: Not really following what you refer to here but feel free to update as I'm just about to log off for the night.

crates/services/relayer/src/service/syncing.rs
xgreenx: I also prefer to just include this information into the error, instead of adding additional log for the same event

crates/services/relayer/src/service.rs
netrome: Changed in https://github.com/FuelLabs/fuel-core/pull/3004/commits/7a52b5039f6ad02b18db4ca30bec4f5d61726424

crates/services/relayer/src/service/syncing.rs
netrome: Fair enough, I've updated to include the debug information in the error instead here https://github.com/FuelLabs/fuel-core/pull/3004/commits/7a52b5039f6ad02b18db4ca30bec4f5d61726424

AurelienFT (approved):

PR #3022, opened 2025-05-22
feat: Support rolling back relayer DB

### LINKED ISSUE #3019: Support rollback of the relayer database
Rollback functionality right now only works with `BlockHeight`. We need to teach ti to work wit h`DaBlockHeight` as well and implement rollback of the relayer database.

Relayer database doesn't have any state rewind, we just need to cleanup the events and set the new height. Or, as an alternative, we can start to use state rewind

netrome:
Closes #3019 

This PR extends the existing rollback functionality to also roll back the relayer database. The DA height to roll back to is inferred based on the DA height of the target block during the rollback.


rymnc (commented):
while you do this, do we want to include the compression db rollback too here? if not, can be addressed in a follow up :)

netrome:
> while you do this, do we want to include the compression db rollback too here? if not, can be addressed in a follow up :)

In the interest of keeping PRs small and focused I'd prefer to tackle that separately. I can start working on it tomorrow. If it ends up being very small and simple we could consider merging the two changes into a single PR, but I see no reason to hold this PR in the meantime.

netrome:
> while you do this, do we want to include the compression db rollback too here? if not, can be addressed in a follow up :)

Created an issue for the compression DB rollback: https://github.com/FuelLabs/fuel-core/issues/3024

tests/tests/relayer.rs
```diff
@@ -705,6 +718,100 @@ async fn balances_and_coins_to_spend_never_return_retryable_messages() {
...
+                cache_capacity: Some(16 * 1024 * 1024 * 1024),
+                max_fds: -1,
+                columns_policy: ColumnsPolicy::Lazy,
+            },
+        )
+        .expect("Failed to create database")
+    };
+    let db = open_db(&tmp_dir);
+
+    let srv = FuelService::from_combined_database(db.clone(), config)
```
rymnc, line 774: can you use the `FuelCoreDriver` to create the node instead?

rymnc (changes requested):

crates/fuel-core/src/combined_database.rs
```diff
@@ -215,7 +219,7 @@ impl CombinedDatabase {
         )?;
         let relayer = Database::open_rocksdb(
             path,
-            StateRewindPolicy::NoRewind,
+            state_rewind_policy,
```
rymnc, line 222: is this change intended? curious

rymnc (dismissed):

AurelienFT (dismissed):

crates/fuel-core/src/combined_database.rs
```diff
@@ -447,6 +461,15 @@ impl CombinedDatabase {
                 ));
             }
 
+            if let Some(relayer_db_height) = relayer_db_height {
+                if relayer_db_height < target_da_height {
+                    return Err(anyhow::anyhow!(
+                        "gas-price-chain database height({relayer_db_height}) \
```
AurelienFT, line 467:
```suggestion
                        "relayer database height({relayer_db_height}) \
```

netrome: Oh nice, thanks for spotting.

crates/fuel-core/src/combined_database.rs
netrome: Yes it is. Having a hard-coded `NoRewind` here kind of defeats the purpose of supporting state rewinds. Though an improvement would be to support a separate argument for the relayer state rewind policy since DA blocks are produced less frequently, but I don't want to prematurely bloat the method interface for that since that could be a subsequent optimization.

crates/fuel-core/src/combined_database.rs
```diff
@@ -419,16 +430,19 @@ impl CombinedDatabase {
...
-            let gas_price_rolled_back = gas_price_chain_height.is_none()
-                || gas_price_chain_height.expect("We checked height before")
-                    == target_block_height;
+            let gas_price_rolled_back =
+                is_equal_or_none(gas_price_chain_height, target_block_height);
 
             if on_chain_height == target_block_height
                 && off_chain_height == target_block_height
                 && gas_price_rolled_back
+                && relayer_db_rolled_back
```
xgreenx, line 445: Relayer database is not related to L2 blocks. So we should revert it independently.

We need a separate function and `fuel-core rollback` should accept another `--da-block-height` argument in order to initiate relayer rollback.

netrome: Yeah makes sense. Done in https://github.com/FuelLabs/fuel-core/pull/3022/commits/e8c2eeb4b4a388a9815cf8da38362859f965756e

rymnc (dismissed):

xgreenx (approved):

Issue #3017, opened 2025-05-19
Prevent dependencies on blob transactions at txpool level


AurelienFT:
When we will implement the parallel-executor, the blob transactions are executed at the end to avoid changing the execution results of transactions that are executed in the same time and use the modified blobs. Executing them at the end breaks the dependency order in the block and will invalidate transaction that depend on this.

For now, we can just disable having dependencies on blob transactions as it's not a common use-case and not used for now.

PR #3030, opened 2025-05-27
Prevent dependency on blob transactions

AurelienFT:
## Linked Issues/PRs
Resolves https://github.com/FuelLabs/fuel-core/issues/3017

## Checklist

### Before requesting review


rymnc (commented):
wondering if this will affect any current users and if we should document a work around in the changelog if so

PR #3031, opened 2025-05-27
Create validation in parallel-executor

AurelienFT:
## Description
TBD


crates/services/parallel-executor/src/validator/dependency_graph.rs
```diff
@@ -0,0 +1,621 @@
...
+        // when: Add transaction with contract (ready - first to use contract)
+        let dependent_tx = create_transaction_with_contracts(vec![contract_id]);
+        graph.add_transaction(1, dependent_tx);
+
+        // then: Only independent transactions should be ready
+        let ready = graph.get_ready_transactions();
+        assert_eq!(ready.len(), 2);
+        assert!(ready.contains(&0));
+        assert!(ready.contains(&1)); // Both are ready now
+    }
```
MitchTurner, line 582: These tests are all great. Awesome coverage. One little nit I have is there is an inconsistent use of `// when` and some of these tests the SUT is unclear. 

`// when` should mark the SUT's usage. So in this test I see the SUT is `get_independent_transactions` but it looks like the SUT might just be `get_ready_transactions()`?

So I'd expect
```rs
        let dependent_tx = create_transaction_with_contracts(vec![contract_id]);
        graph.add_transaction(1, dependent_tx);

        // when
        let ready = graph.get_ready_transactions();
        
        // then
        assert_eq!(ready.len(), 2);
```

I'm not as worried about the larger sanity tests like `dependency_resolution_realistic_scenario` below. That seems fine. But there are a few tests that could match the `// when` and SUT better.

Won't hold up the PR for that though. These tests are already a win.

MitchTurner (commented):
wip

crates/services/parallel-executor/src/validator/dependency_graph.rs
AurelienFT: cc @rymnc

rymnc: i think the function was renamed and the test wasn't

PR #3032, opened 2025-05-28
feat(parallel-executor): dependency graph for validation

rymnc:
This pull request introduces a new `DependencyGraph` module to the `parallel-executor` service and integrates it into the `Validator` implementation. The changes primarily focus on preparing the system for dependency graph management without yet implementing its full functionality.

### Dependency Graph Integration:

* **Added `DependencyGraph` to imports**: The `DependencyGraph` type from the `dependency_graph` module was added to the imports in `validator.rs` to enable its usage.
* **Introduced `dependency_graph` module**: A new `pub(crate)` module named `dependency_graph` was introduced in the `parallel-executor` service, laying the groundwork for dependency management.
* **Initialized `DependencyGraph` in `Validator`**: A `DependencyGraph` instance is now initialized in the `Validator` implementation, using the size hint from the transaction source to determine its initial capacity. However, the graph is not yet actively used in the logic.

### DIFF
--- a/crates/services/parallel-executor/src/validator/dependency_graph.rs
+++ b/crates/services/parallel-executor/src/validator/dependency_graph.rs
@@ -141,9 +141,29 @@ impl DependencyGraph {
                         if other_contracts.contains(&contract_id) {
                             *remaining_deps -= 1;
                             if *remaining_deps == 0 {
+                                // mark as independent
                                 self.independent_transactions.insert(other_tx_idx);
                                 self.newly_independent_queue.push_back(other_tx_idx);
                                 newly_independent.push(other_tx_idx);
+
+                                // clean up dependency tracking for this now-independent transaction
+                                if let Some(contracts) =
+                                    self.transaction_to_contracts.remove(&other_tx_idx)
+                                {
+                                    for contract in contracts {
+                                        if let Some(tx_set) = self
+                                            .contract_to_transactions
+                                            .get_mut(&contract)
+                                        {
+                                            tx_set.remove(&other_tx_idx);
+                                            if tx_set.is_empty() {
+                                                self.contract_to_transactions
+                                                    .remove(&contract);
+                                            }
+                                        }
+                                    }
+                                }
+                                self.remaining_dependencies.remove(&other_tx_idx);
                             }
                         }
                     }
--- a/crates/services/parallel-executor/src/validator/dependency_graph.rs
+++ b/crates/services/parallel-executor/src/validator/dependency_graph.rs
@@ -1,0 +1,522 @@
+use std::collections::{
+    HashMap,
+    HashSet,
+    VecDeque,
+};
+
+use fuel_core_types::{
+    blockchain::transaction::TransactionExt,
+    fuel_tx::{
+        ContractId,
+        Input,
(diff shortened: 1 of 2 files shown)


crates/services/parallel-executor/src/validator/dependency_graph.rs
```diff
@@ -0,0 +1,522 @@
...
+
+                if let Some(remaining_deps) =
+                    self.remaining_dependencies.get_mut(&other_tx_idx)
+                {
+                    if let Some(other_contracts) =
+                        self.transaction_to_contracts.get(&other_tx_idx)
+                    {
+                        if other_contracts.contains(&contract_id) {
+                            *remaining_deps -= 1;
+                            if *remaining_deps == 0 {
```
AurelienFT, line 143: delete from `remaning_deps`, `transactions_to_contracts` etc...

crates/services/parallel-executor/src/validator/dependency_graph.rs
```diff
@@ -0,0 +1,522 @@
...
+    /// Transaction index -> Set of contract IDs it depends on
+    transaction_to_contracts: HashMap<usize, HashSet<ContractId>>,
+    /// Transaction index -> Transaction
+    transactions: HashMap<usize, Transaction>,
+    /// Tracks remaining dependencies for each transaction
+    remaining_dependencies: HashMap<usize, usize>,
+    /// Cache of currently independent transactions
+    independent_transactions: HashSet<usize>,
+    /// Queue for newly independent transactions
+    newly_independent_queue: VecDeque<usize>,
```
AurelienFT, line 30: I don't understand why we need this in addition of `independent_transactions` is it not possible to be only vecdeque ?

rymnc: for O(1) lookup and popping :)

crates/services/parallel-executor/src/validator/dependency_graph.rs
```diff
@@ -0,0 +1,522 @@
...
+            create_transaction_with_contracts(vec![contract_a, contract_b]),
+        );
+        graph.add_transaction(
+            3,
+            create_transaction_with_contracts(vec![contract_b, contract_c]),
+        );
+        graph.add_transaction(4, create_transaction_with_contracts(vec![contract_c]));
+
+        // Initially only TX0 should be independent
+        assert_eq!(graph.get_independent_transactions(), vec![0]);
```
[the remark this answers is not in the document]
rymnc: how is tx1 independent? it has dependency on contract A

crates/services/parallel-executor/src/validator/dependency_graph.rs
rymnc: addressed in 66cdc57f52

crates/services/parallel-executor/src/validator/dependency_graph.rs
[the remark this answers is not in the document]
rymnc: but it isn't independent :) there is another transaction which depends on that contract. but i do understand your point, we want the first transaction which uses a contract to be independent, can make that change in logic

PR #3037, opened 2025-06-02
docs(deepwiki): add deepwiki badge so it gets updated weekly

rymnc:
Adds the deepwiki badge so that the documentation there gets updated weekly


MitchTurner (approved):
Sick

PR #3050, opened 2025-07-02
chore(benchmarks): add benchmarks for niop opcode in all operation modes

rymnc:
wip

### DIFF
--- a/benches/benches/vm_set/alu.rs
+++ b/benches/benches/vm_set/alu.rs
@@ -201,6 +201,59 @@ pub fn run(c: &mut Criterion) {
             .with_prepare_script(vec![op::movi(0x11, 100000)]),
     );
 
+    {
+        let mut niop = c.benchmark_group("niop");
+        let value_count = 10;
+
+        macro_rules! niop_bench {
+            ($niop:expr, $operation:ident, $opwidth:ident, $lhs:expr, $rhs:expr) => {
+                run_group_ref(
+                    $niop,
+                    &format!("niop_{}_{}",
+                        stringify!($operation).to_lowercase(),
+                        stringify!($opwidth).to_lowercase()
+                    ),
+                    VmBench::new(op::niop_args(0x10, 0x11, 0x12, narrowint::MathArgs {
+                        op: narrowint::MathOp::$operation,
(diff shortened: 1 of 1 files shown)


benches/benches/vm_set/alu.rs
```diff
@@ -201,6 +201,77 @@ pub fn run(c: &mut Criterion) {
             .with_prepare_script(vec![op::movi(0x11, 100000)]),
     );
 
+    {
+        let mut niop = c.benchmark_group("niop");
```
MitchTurner: nit: I think this would be more readable as `niop_benchmark_group`. It's confusing passing `niop` to something named `niop_bench`

MitchTurner (commented):
s there somewhere I can read up on what `niop` is? I've been out of the loop on it.

xgreenx (commented):
Hmm, how we convert results of these benchmark into the actual cost? Because we have only one gas cost which we need to fill, but we have 18 benchmarks for it. I see how we can use it in sanity checks, but I don't understand how does it work for the `collect` binary=D

rymnc:
> Hmm, how we convert results of these benchmark into the actual cost? Because we have only one gas cost which we need to fill, but we have 18 benchmarks for it. I see how we can use it in sanity checks, but I don't understand how does it work for the `collect` binary=D

checking

benches/benches/vm_set/alu.rs
rymnc: addressed in ee93d3847f

rymnc: image

https://docs.fuel.network/docs/specs/fuel-vm/instruction-set/#niop-narrow-integer-operation

rymnc:
> Hmm, how we convert results of these benchmark into the actual cost? Because we have only one gas cost which we need to fill, but we have 18 benchmarks for it. I see how we can use it in sanity checks, but I don't understand how does it work for the `collect` binary=D

the collect binary groups benchmark groups and takes the average from them i believe. https://github.com/FuelLabs/fuel-core/blob/ee93d3847fbe3dd675d4173883ae75c05ed62675/benches/src/bin/collect.rs#L588-L605

xgreenx (dismissed):

PR #3053, opened 2025-07-11
chore(compression): disable merkleized compression for now

rymnc:
> [!WARNING]
> this introduces some tech debt on the fault proving feature flag
> 1. introduced `#[allow(unexpected_cfgs)]` in compression and compression service crate
> 2. disabled instances of usage of fault proving feature for the above 2 crates

This PR is meant to be reverted later when we continue work on FP.

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/services/compression/src/storage.rs
+++ b/crates/services/compression/src/storage.rs
@@ -1,5 +1,3 @@
-use fuel_core_storage::merkle::sparse::Merkleized;
-
 /// Generates a random RegistryKey
 #[cfg(feature = "test-helpers")]
 #[allow(clippy::arithmetic_side_effects)]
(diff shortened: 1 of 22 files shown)


crates/services/compression/src/storage.rs
```diff
@@ -25,32 +23,32 @@ pub mod registrations;
 pub mod timestamps;
 
 /// Merkleized Address table type alias
-pub type Address = Merkleized<address::Address>;
+pub type Address = address::Address;
```
xgreenx, line 26: Hmm, curious how it will affect the storage. It looks like we still will have trash in the Merkle-related tables. Maybe we need to come up with the solution how to cleanup them

crates/services/compression/src/storage/column.rs
```diff
@@ -22,27 +19,29 @@ use fuel_core_storage::merkle::{
     Hash,
 )]
 pub enum CompressionColumn {
+    /// Metadata table
+    Metadata = 0,
```
xgreenx, line 23: Is it a good idea to add metadata and change the table numbers now? Doing that over an existing database means database will be corrupted

rymnc: fairly certain it will reuse metadata table from before, we will have to prune the compression database first

crates/services/compression/src/storage.rs
rymnc: yeah we can wipe compression database manually I think?

crates/services/compression/src/storage/column.rs
Voxelot: We should get confirmation on this. @rymnc are you saying that since we didn't put the SMT's into the mainnet compression yet, this is essentially reverting a breaking change that hasn't been deployed yet?

rymnc: it's already in mainnet, I can clone a devnet database and see how it works

rymnc: i checked locally running a testnet node with 0.43.2 with compression enabled and then after with this branch. queries for compressed blocks that were processed with 0.43.2 yield `null`, but the rest come through as normal, no errors in syncing or on graphql. this would be the expected behaviour, so it's fine in my opinion.

1. scale down compression sentry
2. delete compression database
3. configure --da-compression-starting-height and new image
4. scale up compression sentry 

it will work without errors.

PR #3054, opened 2025-07-13
fix(p2p): reduce cached view on req/res to one quarter to prevent mem overuse

rymnc:


crates/services/p2p/src/service.rs
```diff
@@ -918,7 +918,7 @@ where
             heartbeat_max_time_since_last,
             next_check_time,
             heartbeat_peer_reputation_config,
-            cached_view: Arc::new(CachedView::new(614 * 10, metrics)),
+            cached_view: Arc::new(CachedView::new(1_535, metrics)),
```
xgreenx: I think we need to make it configurable

rymnc: oki

rymnc: addressed in [963b302](https://github.com/FuelLabs/fuel-core/pull/3054/commits/963b302376bfe91ab9d2661d5993aa52ac39994a)

Voxelot (dismissed):

xgreenx (approved):

PR #3055, opened 2025-07-13
chore: avoid usage of the BoxedStream

### LINKED ISSUE #2836: Avoid usage of the `BoxedStream` from the ports in services
We have many places in the codebase where we require the port to return a `BoxedStream`, while instead, it could have a generic trait with `async` method

Image

Image

Source: https://github.com/FuelLabs/fuel-core/pull/2791#discussion_r1985721863

grandima:
## Description

Close #2836
Introduce `AsyncIterTrait` that just returns next element.
Refactor returns of `BoxStream` with the trait above.
Note: in stable rust it's not possible to create type alias like 
`type AsyncIterAlias<T> = impl AsyncIterTrait<Item=T> + Send + Sync + Unpin + 'static;`, so expect to see in code a lot of long returns like `-> impl AsyncIterTrait<Item = Self::GossipedStatuses> + Send + Unpin + 'static`.

Screenshot 2025-07-13 at 16 22 15

## Checklist

### Before requesting review

### DIFF
--- a/crates/services/tx_status_manager/src/service.rs
+++ b/crates/services/tx_status_manager/src/service.rs
@@ -7,7 +7,7 @@ use crate::{
         P2PPreConfirmationMessage,
         P2PSubscriptions,
     },
-    subscriptions::Subscriptions,
+    subscriptions::{Subscriptions, TxStatusSubscription},
     update_sender::TxStatusChange,
 };
 use fuel_core_services::{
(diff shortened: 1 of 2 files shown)


grandima:
@MitchTurner Thanks for such useful tips. 

I've updated the implementation. Please check it out.

PR #2973, opened 2025-04-24
Allow syscall during the execution of predicates and scripts

xgreenx:
Allow syscall during the execution of predicates and scripts.

To enable syscalls, `--allow-syscall` and `--debug` flags must be set.

Supported only one syscall for now, to print logs. It follows the syntax supported by Sway's `__dbg!` macro.

Uses feature from https://github.com/FuelLabs/fuel-vm/pull/945

PR #3059, opened 2025-07-21
Enable debug logging in predicates and scripts using ECAL

Dentosal:
## Linked Issues/PRs

VM changes done in https://github.com/FuelLabs/fuel-vm/pull/963 , waiting for the next `fuel-vm` release.

See https://github.com/FuelLabs/fuel-vm/pull/945#issuecomment-2829271997. Replaces https://github.com/FuelLabs/fuel-core/pull/2973.

## Description

Accomodates https://github.com/FuelLabs/fuel-vm/pull/963. Allows setting `--allow-syscall` flag, that in turn allows logging from scripts and predicates using `__dbg!` macro in Sway, which uses the new "ecal syscall api" defined here. 

Currently the API looks like this: `ECAL a b c d` where:
* `a`: Syscall number, set to `1000` for debug log output
* `b`: File descriptor, set to `1` for stdout
* `c`: Address of (pointer to) the message to log
* `d`: Length of the message to log.

### Open questions

* Should we print receipts or at least indicate if the tx reverts in the console output?

### Resolved 

* Should we test the log output somehow? I attempted to use the [tracing-test crate](https://crates.io/crates/tracing-test), but it cannot read logs from the fuel-core. [See this issue](https://github.com/dbrgn/tracing-test/issues/22). I tried to implement my own, but it turns out it's complicated. Code below.
  * Done for now using a separate process and reading stderr.

<details>

<summary>How to capture logs</summary>

Doesn't work since global subscriber cannot be added for multiple simultaneous tests. Would need to force single thread and filter based on thread id, or somehow inject per-test unique span to fuel-core.

```rust
use std::{
    array, io, sync::{
        Arc,
        Mutex,
    }
};
use tracing::Dispatch;
use tracing_subscriber::{
    FmtSubscriber,
    fmt::MakeWriter,
};


/// A struct to capture tracing logs for testing purposes.
/// Inspired by the `tracing-test` crate, which turned out to
/// be insufficient for this.
#[derive(Clone)]
pub struct CaptureLogs(Arc<Mutex<String>>);

impl CaptureLogs {
    pub fn init() -> Self {
        let buffer = Arc::new(Mutex::new(String::new()));
        let this = Self(buffer.clone());
        let subscriber: Dispatch = FmtSubscriber::builder()
            .with_writer(this.clone())
            .with_level(true)
            .with_ansi(false)
            .into();
        tracing::dispatcher::set_global_default(subscriber)
            .expect("Could not set global tracing subscriber");
        this
    }

    /// Finds expected part of logs and discards everything before it. Panics if the expected part is not found.
    pub fn expect(&self, expected: &str) {
        // Just reuse the regex logic
        self.expect_regex(&regex::escape(expected))
    }

    /// Like `expect`, but the next part must follow immediately after the previous one.
    pub fn expect_follows(&self, expected: &str) {
        // Again, reuse the regex logic
        self.expect_regex(&format!("^{}", regex::escape(expected)))
    }

    /// Finds expected part of logs and discards everything before it. Panics if the expected part is not found.
    pub fn expect_regex(&self, regex: &str) {
        let mut data = self.0.lock().expect("Mutex poisoned");
        let Some(m) = regex::Regex::new(regex)
            .expect("Failed to compile regex")
            .find(&data) else {
            println!("Captured logs: {}", data);
            panic!(
                "Expected pattern `{}` not found in logs.",
                regex,
            );
        };
        *data = data.split_at(m.end()).1.to_owned();
    }
}

impl io::Write for CaptureLogs {
    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        let mut target = self.0.lock().expect("Mutex poisoned");
        target.push_str(&String::from_utf8_lossy(buf));
        Ok(buf.len())
    }

    fn flush(&mut self) -> io::Result<()> {
        Ok(())
    }
}

impl MakeWriter<'_> for CaptureLogs {
    type Writer = Self;

    fn make_writer(&self) -> Self::Writer {
        self.clone()
    }
}
```
</details>

## Checklist
- [x] New behavior is reflected in tests - only partially, see discussion above

### Before requesting review

- [ ] Someone else?


crates/syscall/src/handlers/ignore.rs
```diff
@@ -0,0 +1,40 @@
...
+    },
+};
+
+/// If enabled, ignores all ECAL invocations.
+#[derive(Debug, Clone, Default)]
+pub struct IgnoreEcal {
+    /// If this is true, the ECAL invocations will be ignored.
+    /// If false, the ECAL invocations will error.
+    pub enabled: bool,
+}
```
MitchTurner: I think this should be an enum:

```rs
pub enum EcalShouldBeEnabled {
    ErrorBecauseEcalDisabled,
    IgnoreBecauseEcalEnabled
}

impl EcalShouldBeEnabled {
    fn new(is_enabled: bool) -> Self {
        if is_enabled {
            Self::IgnoreBecauseEcalEnabled
        } else {
            Self::ErrorBecauseEcalDisabled
        }
    }
}

impl EcalHandler for EcalShouldBeEnabled {
    fn ecal<M, S, Tx, V>(
        vm: &mut Interpreter<M, S, Tx, Self, V>,
        _a: RegId,
        _b: RegId,
        _c: RegId,
        _d: RegId,
    ) -> SimpleResult<()>
    where
        M: Memory,
    {
        match self {
            Self::IgnoreBecauseEcalEnabled => Ok(()),
            Self::ErrorBecauseEcalDisabled => Err(PanicReason::EcalError.into()),
        }
    }
}
```
or something like that. 

I was having a hard time keeping the `bool` straight, what it signified. Documentation with code is always cleaner than comments imo.

crates/syscall/src/handlers/log_collector.rs
```diff
@@ -0,0 +1,126 @@
...
+                return Err(PanicReason::EcalError.into());
+            }
+        };
+
+        Ok(())
+    }
+}
+
+/// Print ECAL/syscall logs if any produced.
+pub fn maybe_print_logs(logs: &[LogEntry], span: tracing::Span) {
```
MitchTurner: nit: I feel like this is a function that should exist in the sdks, not here, but it's nice to have regardless.

MitchTurner: Maybe we provide a getter to the `&[LogEntry]` and let the caller handle as they see fit?

tests/tests/tx/log_syscall.rs
```diff
@@ -0,0 +1,145 @@
...
+};
+use rand::{
+    Rng,
+    SeedableRng,
+    rngs::StdRng,
+};
+use std::array;
+
+#[tokio::test]
+async fn syscall_logs_produced_by_both_predicates_and_script() {
```
MitchTurner: What is the SUT of these tests? is it `estimate_predicates` or `client.submit_and_await_commit` or both?

MitchTurner: Can we
- somehow ensure that the expected logs are present?
- add some failing tests here as well. Idk if this the right place for all of them, especially the `IgnoreEcal` tests, but don't think I saw those anywhere

crates/syscall/src/handlers/ignore.rs
rymnc: agree

tests/tests/tx/log_syscall.rs
```diff
@@ -0,0 +1,145 @@
...
+    fuel_asm::*,
+    fuel_tx::*,
+};
+use rand::{
+    Rng,
+    SeedableRng,
+    rngs::StdRng,
+};
+use std::array;
+
```
rymnc, line 28: i would suggest one more test, when `debug` is set to false and the `allow-syscall` flag is enabled

tests/Cargo.toml
```diff
@@ -78,6 +78,7 @@ k256 = { version = "0.13.3", features = ["ecdsa-core"] }
 postcard = { workspace = true }
 primitive-types = { workspace = true, default-features = false }
 rand = { workspace = true }
+regex = "1"
```
rymnc: we need this?

crates/services/txpool_v2/src/service/verifications.rs
```diff
@@ -193,7 +200,16 @@ impl InputDependenciesVerifiedTx {
             tx = tx.check_signatures(&consensus_params.chain_id())?;
 
             let parameters = CheckPredicateParams::from(consensus_params);
-            tx = tx.check_predicates(&parameters, memory, view)?;
+
+            let mut ecal_state = EcalLogCollector::default();
+            ecal_state.enabled = allow_syscall;
+
+            tx = tx.check_predicates(&parameters, memory, view, ecal_state.clone())?;
```
rymnc: curious why we dont take a mutable reference here in `check_predicates` instead

crates/services/txpool_v2/src/service/verifications.rs
```diff
@@ -193,7 +200,16 @@ impl InputDependenciesVerifiedTx {
...
+
+            maybe_print_logs(
+                &ecal_state.logs.lock().clone(),
```
rymnc: possible to remove the clone here?

rymnc: afaiu the `ecal_state` isn't used again

crates/syscall/src/handlers/log_collector.rs
rymnc: lets just add this in the impl for `EcalLogCollector`, there's no reason for this function to be free :)

Dentosal: This is what the callers do. It's only included here as a helper because callers in multiple crates want to do this exact same thing.

tests/tests/tx/log_syscall.rs
Dentosal: SUT is both. The PR description contains explanation on why we don't test for expected logs.

tests/Cargo.toml
Dentosal: Not anymore.

crates/services/txpool_v2/src/service/verifications.rs
Dentosal: Because `check_predicates` needs to initialize the VM with the ecal handler, and the VM requires an owned type for that.

crates/syscall/src/handlers/log_collector.rs
Dentosal: This doesn't feel like a part of the log collector. This is something that a caller does, and only exists here because two different crates want to have the same functionality.

crates/services/txpool_v2/src/service/verifications.rs
Dentosal: Yep. Done in 905791781a75738aa7d3fb37dc9f06331fa7a557

tests/tests/tx/log_syscall.rs
MitchTurner: > What is the SUT of these tests? is it estimate_predicates or client.submit_and_await_commit or both?

This was a leading question. Sorry. I meant to say, let's only test one thing in each test and be clear in the name of the test what the SUT is:
https://www.notion.so/fuellabs/Engineering-Style-Guide-83ba6fd81f12443796c814ecfc746b39?source=copy_link#f3232125ea964e959a66b9f41267d96e

i.e. these are integration tests, so we should put this in terms of which apis the client is calling here. Seems to be `estimate_predicates__`?

Same with the next test.

crates/syscall/src/handlers/log_collector.rs
```diff
@@ -0,0 +1,126 @@
...
+use parking_lot::Mutex;
+
+/// Collects logs from ECAL invocations.
+#[derive(Debug, Clone, Default)]
+pub struct EcalLogCollector {
+    /// If this is true, the ECAL invocations will be logged.
+    /// If false, the ECAL invocations error.
+    pub enabled: bool,
+    /// Logs collected from ECAL invocations.
+    pub logs: Arc<Mutex<Vec<LogEntry>>>,
```
xgreenx: Why do we need `Arc<Mutex` if you have access to `mut` state via `vm.ecal_state_mut()`?

Dentosal: Because we don't have access to the VM when using `checked_tx.check_predicates`. We could have a separate type for non-predicate running that doesn't require this.

xgreenx: Okay, I see. But we still need to store logs per each predicate separately. So when `EcalLogCollector` is destroyed, only then we need to dump logs to a some shared mutex.

Dentosal: That's not really doable. We store all predicate logs in the same array, and identify them by `predicate_idx` field of  `LogEntry`.

tests/tests/tx/log_syscall.rs
Dentosal: We generally do not have tests for conflicting flags, and if we're to add those, this is the wrong place to do so. If we're going to add these, I'd like to test for other combinations too, like setting `utxo_validation` to `false`. But these are checked on binary startup, so I don't think automated testing does much good here.

tests/tests/tx/log_syscall.rs
Dentosal: Makes sense. I renamed the tests in df53196c09.

xgreenx (dismissed):

xgreenx (approved):

Issue #1867, opened 2024-04-26
Add executor test to show that Mint can't lie about gas price


MitchTurner:
![image](https://github.com/FuelLabs/fuel-core/assets/11023047/5523bb05-156b-4154-b113-e084ed061fe9)

The fact that we are micro-managing the order of these functions hoping it's correct is not right. We should have a test to show the behavior we want.

MitchTurner:
I don't think this makes any sense. There is no way for us to produce a failing tests for either the validate or the produce cases. `validate` uses the provided mint to choose a value of `gas_price` and `execute` creates the `mint` using whatever `gas_price` you provide.<br><br>Instead I'm going to add a test simply showing that production adds a mint and call it good.

PR #3061, opened 2025-07-25
Add test showing mint added

MitchTurner:
## Linked Issues/PRs
closes https://github.com/FuelLabs/fuel-core/issues/1867

## Description
The issue says we should have a test showing that incorrect mint values won't get validated, but that's not how the code works--expecially from the external apis. There is no way for us to produce a failing tests for either the validate or the produce cases. validate uses the provided mint to choose a value of gas_price and execute creates the mint using whatever gas_price you provide.

Instead, this just adds a test showing that a mint is added when you call the external api `produce_without_commit_with_source`. There is already a test for `validate` (technically `validate_and_commit` which is a bastardization of `validate` only used in tests).

The test coverage for block production and validation is very bad right now, as captured in this issue:
https://linear.app/fuel-network/issue/CLNT-1645/beef-up-execution-tests

We should be very clear about what external apis each test is for, instead many of them are for abstract concepts.

### Before requesting review


xgreenx (approved):

xgreenx (commented):
I think the issue is to add tests that show that usage of the incorrect value in the Mint transaction will cause an error for cases when `ContracId` is zero and non-zero

xgreenx:
You can create a bad block manually with the wrong gas price and try to validate it after

MitchTurner:
> You can create a bad block manually with the wrong gas price and try to validate it after

Sure. But validate uses the gas price on the mint to check. So if we change the gas price it will just validate with the new one. 

We could have the txs be invalid because they didn't provide enough to cover the gas maybe?

xgreenx:
Yes, we need to build a case where the charged amount doesn't match the value provided in the `Mint` transaction

Dentosal (approved):

rymnc (approved):

Issue #3066, opened 2025-07-28
chore(compression): make compression independent from state rewind


rymnc:
currently, compression depends on the state rewind feature and if we're compressing blocks that are beyond the range the node does not progress and gets stuck.

> we should rely on the events from the execution during compression. In this case we will be able to sync any blocks

PR #3071, opened 2025-07-31
Use inputs/outputs instead of events to compress blocks

MitchTurner:
## Linked Issues/PRs

closes https://github.com/FuelLabs/fuel-core/issues/3066

## Description

The `UninitializedCompressionService` relies on the the validation of old blocks. The problem is that if the block was committed too long ago, we can't reproduce the state before it was committed. This will cause the compression service to break if we are trying to sync old blocks.

This removes the need to get the events from execution and instead gets the correct information from inputs and outputs of the block in storage.

This code also improves the performance by sorting all the UTxOs used in each block before looking up the tx_pointer for each tx. So it's O(N) instead of O(N^2).

### Note:
I didn't add any new tests since this should have parity with the old behavior. It might be nice to add a test for the case where we are behind state rewind?

### Before requesting review

### Followup Issues
The work done in this PR wasn't meant to add any behavior, so we didn't add any tests. But it clarified some places we were missing coverage, so I've created a followup issue: https://github.com/FuelLabs/fuel-core/issues/3073

### DIFF
--- a/crates/fuel-core/src/service/adapters/compression_adapters.rs
+++ b/crates/fuel-core/src/service/adapters/compression_adapters.rs
@@ -21,7 +21,10 @@ use fuel_core_compression_service::{
     },
 };
 use fuel_core_storage::transactional::HistoricalView;
-use fuel_core_types::services::block_importer::SharedImportResult;
+use fuel_core_types::{
+    fuel_types::ChainId,
+    services::block_importer::SharedImportResult,
+};
 
 use super::import_result_provider::{
     self,
@@ -69,11 +72,12 @@ impl block_source::BlockSource for CompressionBlockImporterAdapter {
 impl configuration::CompressionConfigProvider
     for crate::service::config::DaCompressionConfig
 {
-    fn config(&self) -> config::CompressionConfig {
+    fn config(&self, chain_id: ChainId) -> config::CompressionConfig {
         config::CompressionConfig::new(
             self.retention_duration,
             self.starting_height,
             self.metrics,
+            chain_id,
         )
     }
 }
--- a/crates/fuel-core/src/service/adapters/compression_adapters.rs
+++ b/crates/fuel-core/src/service/adapters/compression_adapters.rs
@@ -1,3 +1,7 @@
+use super::import_result_provider::{
+    self,
+    ImportResultProvider,
+};
 use crate::{
     database::{
         Database,
@@ -22,13 +26,9 @@ use fuel_core_compression_service::{
 };
 use fuel_core_storage::transactional::HistoricalView;
 use fuel_core_types::{
+    blockchain::block::Block,
     fuel_types::ChainId,
     services::block_importer::SharedImportResult,
-};
-
-use super::import_result_provider::{
-    self,
-    ImportResultProvider,
 };
 
 /// Provides the necessary functionality for accessing latest and historical block data.
@@ -63,9 +63,13 @@ impl block_source::BlockSource for CompressionBlockImporterAdapter {
         self.block_importer.events_shared_result()
     }
 
-    fn get_block(&self, height: BlockAt) -> anyhow::Result<SharedImportResult> {
-        self.import_result_provider_adapter
-            .result_at_height(height.into())
+    fn get_block(&self, height: BlockAt) -> anyhow::Result<Block> {
+        let block = &self
+            .import_result_provider_adapter
+            .result_at_height(height.into())?
+            .sealed_block
+            .entity;
+        Ok(block.clone())
     }
 }
 
--- a/crates/fuel-core/src/service/adapters/compression_adapters.rs
+++ b/crates/fuel-core/src/service/adapters/compression_adapters.rs
@@ -1,6 +1,5 @@
 use super::import_result_provider::{
     self,
-    ImportResultProvider,
 };
 use crate::{
     database::{
@@ -24,7 +23,10 @@ use fuel_core_compression_service::{
         configuration,
     },
 };
-use fuel_core_storage::transactional::HistoricalView;
+use fuel_core_storage::transactional::{
+    AtomicView,
+    HistoricalView,
+};
 use fuel_core_types::{
     blockchain::block::Block,
     fuel_types::ChainId,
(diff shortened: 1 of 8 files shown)


crates/services/compression/src/temporal_registry.rs
```diff
@@ -53,7 +62,48 @@ pub struct CompressionStorageWrapper<'a, CS> {
...
+        block: &'a Block,
+        chain_id: ChainId,
+    ) -> anyhow::Result<Self> {
+        let mut tx_pointers = HashMap::new();
+        for (tx_index, tx) in block.transactions().iter().enumerate() {
+            let tx_index = u16::try_from(tx_index)
+                .map_err(|_| anyhow::anyhow!("Transaction index exceeds u16 limit"))?;
+            let tx_id = tx.id(&chain_id);
+            for input in tx.inputs().iter() {
+                if let fuel_core_types::fuel_tx::Input::CoinPredicate(coin) = input {
```
rymnc: a `match` might be cleaner here :)

crates/services/compression/src/service.rs
```diff
@@ -689,6 +684,7 @@ mod tests {
     async fn compression_service__syncs_from_scratch_when_database_is_empty() {
         // given: we start the compression service, with a canonical height provider of height 5
         let block_count = 10;
+        let chain_id = ChainId::new(123);
```
rymnc: maybe we can just use the default chain id in the tests?

crates/fuel-core/src/service/adapters/compression_adapters.rs
```diff
@@ -60,20 +63,25 @@ impl block_source::BlockSource for CompressionBlockImporterAdapter {
...
-    fn get_block(&self, height: BlockAt) -> anyhow::Result<SharedImportResult> {
-        self.import_result_provider_adapter
-            .result_at_height(height.into())
+    fn get_block(&self, height: BlockAt) -> anyhow::Result<Block> {
+        let block = &self
+            .import_result_provider_adapter
+            .result_at_height(height.into())?
+            .sealed_block
+            .entity;
+        Ok(block.clone())
```
rymnc: do we need this clone here? curious 

perhaps we can clone the `SharedImportResult` first and then return the entity from that because it's wrapped in an Arc. maybe wrong here

crates/services/compression/src/temporal_registry.rs
```diff
@@ -53,7 +62,48 @@ pub struct CompressionStorageWrapper<'a, CS> {
...
+impl<'a, CS> CompressionContext<'a, CS> {
+    /// Creates a new compression context
+    pub fn create_from_block(
+        storage_tx: &'a mut StorageTransaction<CS>,
+        block: &'a Block,
+        chain_id: ChainId,
+    ) -> anyhow::Result<Self> {
+        let mut tx_pointers = HashMap::new();
+        for (tx_index, tx) in block.transactions().iter().enumerate() {
+            let tx_index = u16::try_from(tx_index)
```
rymnc: is this used?

crates/fuel-core/src/service/adapters/compression_adapters.rs
xgreenx: `import_result_provider_adapter` uses state rewind under the hood=DDD

We need just to get the block from the database, not from the `import_result_provider_adapter`

crates/services/compression/src/temporal_registry.rs
```diff
@@ -53,7 +62,48 @@ pub struct CompressionStorageWrapper<'a, CS> {
...
+                    let tx_pointer = coin.tx_pointer;
+                    tx_pointers.insert(utxo_id, tx_pointer);
+                }
+            }
+            for (index, output) in tx.outputs().iter().enumerate() {
+                let index = u16::try_from(index)
+                    .map_err(|_| anyhow::anyhow!("Output index exceeds u16 limit"))?;
+                if let fuel_core_types::fuel_tx::Output::Coin { .. } = output {
+                    let utxo_id = UtxoId::new(tx_id, index);
+                    let tx_pointer = TxPointer::new(*block.header().height(), tx_index);
```
xgreenx: You can create it at the begin of the loop, then it will be easier to see where `tx_index` from @rymnc 's comment is used=D

crates/services/compression/src/temporal_registry.rs
```diff
@@ -53,7 +62,48 @@ pub struct CompressionStorageWrapper<'a, CS> {
...
+                } else if let fuel_core_types::fuel_tx::Input::CoinSigned(coin) = input {
+                    let utxo_id = coin.utxo_id;
+                    let tx_pointer = coin.tx_pointer;
+                    tx_pointers.insert(utxo_id, tx_pointer);
+                }
+            }
+            for (index, output) in tx.outputs().iter().enumerate() {
+                let index = u16::try_from(index)
+                    .map_err(|_| anyhow::anyhow!("Output index exceeds u16 limit"))?;
+                if let fuel_core_types::fuel_tx::Output::Coin { .. } = output {
```
xgreenx: I think `Variable` and `Change` output must be used as well, because it is possible that next transactions will use these coins.

crates/services/compression/src/temporal_registry.rs
xgreenx: ++

crates/services/compression/src/temporal_registry.rs
MitchTurner: Yes. But not until the output section, so I'll move it.

crates/fuel-core/src/service/adapters/compression_adapters.rs
MitchTurner: Changed it to `OnChain` db, so its not an issue now.

crates/services/compression/src/temporal_registry.rs
MitchTurner: We should have coverage for these, then.

MitchTurner: Added followup issue: https://github.com/FuelLabs/fuel-core/issues/3073

PR #3085, opened 2025-09-01
Block Aggregator Service Scaffolding

MitchTurner:
## Linked Issues/PRs

Starts https://github.com/FuelLabs/fuel-core/issues/3086

Follow-up Issues:
https://github.com/FuelLabs/fuel-core/issues/3087
https://github.com/FuelLabs/fuel-core/issues/3088
https://github.com/FuelLabs/fuel-core/issues/3089
https://github.com/FuelLabs/fuel-core/issues/3090
https://github.com/FuelLabs/fuel-core/issues/3091

## Description

This PR adds the core logic of the block aggregator service. The domain includes:
1. Receiving the block from some abstracted source. This will be implemented later as a combination of block importer and access to the a on-chain db, where it will sync any missing blocks before subscribing to the importer values
2. Storing the blocks as raw bytes (? maybe we can abstract the serialization format). This will be implemented as a normal Fuel DB
3. An API abstraction with queries for ranges of blocks as well as info on the current height of the aggregator's db. This could be implemented as some protobuf server or something else.

We don't want this service to block the initialization of other services. This means that syncing will happen in the `run` function (rather than in some `UninitializedTask`). As the syncing progresses, the users can query the current height with the `BlockAggregatorQuery::GetCurrentHeight`.

The response type for a block range query is a assoc type because we are still up in the air about how it will be handled. There is a kinda of example type added, but it's not used anywhere yet, including the tests:
```rs
pub enum BlockRangeResponse {
    Literal(BoxStream<Block>),
    Remote(String),
}
```

## Checklist

### Before requesting review

### DIFF
--- a/crates/services/block_aggregator_api/Cargo.toml
+++ b/crates/services/block_aggregator_api/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2024"
 fuel-core-services = { workspace = true}
 tokio = "1.45.1"
 anyhow = "1.0.98"
+futures-util = "0.3.31"
 
 [dev-dependencies]
 fuel-core-services = { workspace = true, features = ["test-helpers"] }
--- a/crates/services/block_aggregator_api/Cargo.toml
+++ b/crates/services/block_aggregator_api/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "block_aggregator_api"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+fuel-core-services = { workspace = true}
+tokio = "1.45.1"
+anyhow = "1.0.98"
+
+[dev-dependencies]
+fuel-core-services = { workspace = true, features = ["test-helpers"] }
--- a/crates/services/block_aggregator_api/Cargo.toml
+++ b/crates/services/block_aggregator_api/Cargo.toml
@@ -9,6 +9,7 @@ fuel-core-services = { workspace = true }
 rand = { workspace = true }
 tokio = { workspace = true }
 tracing = { workspace = true }
+bytes = { workspace = true }
 
 [dev-dependencies]
 fuel-core-services = { workspace = true, features = ["test-helpers"] }
--- a/crates/services/block_aggregator_api/Cargo.toml
+++ b/crates/services/block_aggregator_api/Cargo.toml
@@ -4,12 +4,11 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
-anyhow = "1.0.98"
+anyhow = { workspace = true }
 fuel-core-services = { workspace = true }
-futures-util = "0.3.31"
-rand = "0.9.2"
-tokio = "1.45.1"
-tracing = "0.1.41"
+rand = { workspace = true }
+tokio = { workspace = true }
+tracing = { workspace = true }
 
 [dev-dependencies]
 fuel-core-services = { workspace = true, features = ["test-helpers"] }
--- a/crates/services/block_aggregator_api/Cargo.toml
+++ b/crates/services/block_aggregator_api/Cargo.toml
@@ -4,12 +4,12 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
-fuel-core-services = { workspace = true}
-tokio = "1.45.1"
 anyhow = "1.0.98"
+fuel-core-services = { workspace = true }
 futures-util = "0.3.31"
-tracing = "0.1.41"
 rand = "0.9.2"
+tokio = "1.45.1"
+tracing = "0.1.41"
 
 [dev-dependencies]
 fuel-core-services = { workspace = true, features = ["test-helpers"] }
--- a/crates/services/block_aggregator_api/Cargo.toml
+++ b/crates/services/block_aggregator_api/Cargo.toml
@@ -9,6 +9,7 @@ tokio = "1.45.1"
 anyhow = "1.0.98"
 futures-util = "0.3.31"
 tracing = "0.1.41"
+rand = "0.9.2"
 
 [dev-dependencies]
 fuel-core-services = { workspace = true, features = ["test-helpers"] }
--- a/crates/services/block_aggregator_api/Cargo.toml
+++ b/crates/services/block_aggregator_api/Cargo.toml
@@ -5,13 +5,13 @@ edition = "2024"
 
 [dependencies]
 anyhow = { workspace = true }
+bytes = { workspace = true }
 fuel-core-services = { workspace = true }
 rand = { workspace = true }
 tokio = { workspace = true }
 tracing = { workspace = true }
-bytes = { workspace = true }
 
 [dev-dependencies]
 fuel-core-services = { workspace = true, features = ["test-helpers"] }
-tracing-subscriber = { workspace = true }
 futures = { workspace = true }
+tracing-subscriber = { workspace = true }
--- a/crates/services/block_aggregator_api/Cargo.toml
+++ b/crates/services/block_aggregator_api/Cargo.toml
@@ -8,6 +8,7 @@ fuel-core-services = { workspace = true}
 tokio = "1.45.1"
 anyhow = "1.0.98"
 futures-util = "0.3.31"
+tracing = "0.1.41"
 
 [dev-dependencies]
 fuel-core-services = { workspace = true, features = ["test-helpers"] }
--- a/crates/services/block_aggregator_api/Cargo.toml
+++ b/crates/services/block_aggregator_api/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2024"
 anyhow = { workspace = true }
 bytes = { workspace = true }
 fuel-core-services = { workspace = true }
+fuel-core-types = { workspace = true, features = ["std"] }
 rand = { workspace = true }
 tokio = { workspace = true }
 tracing = { workspace = true }
--- a/crates/services/block_aggregator_api/Cargo.toml
+++ b/crates/services/block_aggregator_api/Cargo.toml
@@ -13,3 +13,4 @@ tracing = { workspace = true }
 [dev-dependencies]
 fuel-core-services = { workspace = true, features = ["test-helpers"] }
 tracing-subscriber = { workspace = true }
+futures = { workspace = true }
--- a/crates/services/block_aggregator_api/Cargo.toml
+++ b/crates/services/block_aggregator_api/Cargo.toml
@@ -13,3 +13,4 @@ rand = "0.9.2"
 
 [dev-dependencies]
 fuel-core-services = { workspace = true, features = ["test-helpers"] }
+tracing-subscriber = { workspace = true }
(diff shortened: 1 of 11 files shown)


crates/services/block_aggregator_api/src/blocks.rs
```diff
@@ -0,0 +1,38 @@
+use crate::result::Result;
+
+pub trait BlockSource: Send + Sync {
+    fn next_block(&mut self) -> impl Future<Output = Result<(u64, Block)>> + Send;
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
```
rymnc, line 16: probably don't want `Clone` on this type, maybe not `Debug` either

crates/services/block_aggregator_api/src/blocks.rs
```diff
@@ -0,0 +1,38 @@
...
+pub struct Block {
+    bytes: Vec<u8>,
```
rymnc: you should use `bytes::Bytes` here or an `Arc<[u8]>` because as i understand it this wont be mutated, just read from once it is created.

if you use either of these types instead of Vec<u8> then the `Clone` attribute can stay

crates/services/block_aggregator_api/src/db.rs
```diff
@@ -0,0 +1,20 @@
+use crate::{
+    blocks::Block,
+    result::Result,
+};
+use fuel_core_services::stream::BoxStream;
+
+pub trait BlockAggregatorDB: Send + Sync {
```
rymnc, line 8: i don't think we make use of async for storage anyway so better to not have these functions return an `impl Storage`

crates/services/block_aggregator_api/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+impl<Api, DB, Blocks> RunnableTask for BlockAggregator<Api, DB, Blocks>
+where
+    Api: BlockAggregatorApi,
+    DB: BlockAggregatorDB,
+    Blocks: BlockSource,
+{
+    async fn run(&mut self, watcher: &mut StateWatcher) -> TaskNextAction {
+        tracing::debug!("BlockAggregator running");
+        tokio::select! {
+            query_res = self.query.await_query() => self.handle_query(query_res).await,
```
rymnc, line 51: i think this tokio::select! should be reordered in the following manner -

```
biased;

watcher

next_block

query
```

this way if a query comes in that potentially references the latest block we don't return an error

crates/services/block_aggregator_api/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+            block_res = self.block_source.next_block() => self.handle_block(block_res).await,
+            _ = watcher.while_started() => self.stop(),
+        }
+    }
+
+    async fn shutdown(self) -> anyhow::Result<()> {
+        Ok(())
```
rymnc, line 61: dont we want to drain the `block_source` here? while it makes shutting down slower, it speeds up startup later

crates/services/block_aggregator_api/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+where
+    Api: BlockAggregatorApi,
+    DB: BlockAggregatorDB,
+    Blocks: BlockSource,
+{
+    pub fn new(query: Api, database: DB, _block_source: Blocks) -> Self {
+        Self {
+            query,
+            database,
+            block_source: _block_source,
```
rymnc:
```suggestion
    pub fn new(query: Api, database: DB, block_source: Blocks) -> Self {
        Self {
            query,
            database,
            block_source,
```

crates/services/block_aggregator_api/src/tests.rs
```diff
@@ -0,0 +1,198 @@
...
+    let (source, _block_sender) = FakeBlockSource::new();
+
+    let mut srv = BlockAggregator::new(api, db, source);
+    let mut watcher = StateWatcher::started();
+    let (query, response) = BlockAggregatorQuery::get_block_range(2, 3);
+
+    // when
+    tokio::spawn(async move {
+        let _ = srv.run(&mut watcher).await;
+    });
```
rymnc: shouldn't this be in a loop? why does it need to be spawned separately?

crates/services/block_aggregator_api/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+anyhow = "1.0.98"
+fuel-core-services = { workspace = true }
+futures-util = "0.3.31"
+rand = "0.9.2"
+tokio = "1.45.1"
+tracing = "0.1.41"
```
rymnc:
```suggestion
anyhow = { workspace = true }
fuel-core-services = { workspace = true }
rand = { workspace = true }
tokio = { workspace = true }
tracing = { workspace = true }
```

pls use `futures::stream::iter`, and move the `futures` to dev deps as it is not used other than the tests

crates/services/block_aggregator_api/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+
+impl<Api, DB, Blocks> RunnableTask for BlockAggregator<Api, DB, Blocks>
+where
+    Api: BlockAggregatorApi,
+    DB: BlockAggregatorDB,
+    Blocks: BlockSource,
+{
+    async fn run(&mut self, watcher: &mut StateWatcher) -> TaskNextAction {
+        tracing::debug!("BlockAggregator running");
+        tokio::select! {
```
rymnc, line 50: i think based on the call we had on Monday, we want to insert blocks into storage in any order i.e if we have a block from the onchain db / importer, they are treated and processed in parallel

if we want this behaviour, then you have to spawn 2 tasks, one which handles the next block from the importer, and one which handles the next block from the onchain db, and then use tokio::select! on the join handle

i.e something like this -

```rs
let block_from_importer = tokio::spawn(get_next_block_from_importer);
let block_from_onchain_db = tokio::spawn(get_next_block_from_onchain_db);


tokio::select! {
    biased;

    _ = watcher.while_started() => self.stop(),

    res = block_from_importer => { handle_block(res) }
    
    res = block_from_onchain_db => { handle_block(res) }
}
```

might not be necessary at the moment, but if we plan to have this service process blocks async (i.e from the tip of the chain as well as from genesis)

this should allow us to have improved behaviour in terms of how many blocks at the tip we have. however, a downside of this approach is that this is non-deterministic and dependent on the behaviour of the adapters.

with the previous approach, if the on_chain_db block source is more aggressive while sending blocks rather than the importer (which will be the case since the importer is only sending one block every second), we will spend more time processing older blocks rather than new ones that might have come in (which are more relevant)

feel free to correct my understanding here

MitchTurner: Yeah. I've thought about this a bit. The `BlockSource` adapter will be able to handle all of this.

Then the `BlockAggregatorDB` adapter will be able to take them out of order while still just tracking the height for the `get_current_height`. So, we can make a decision what that behavior is. I was thinking it would return the heighest continual.

I went back and forth on what type `next_block` should return. Considering it has to do the serialization locally, I think just `impl Future<Output=(u64, Block)> works. The alternative would be some stream of values or borrowed values. We can improve performance more by buffering values into `store_block` to avoid blocking, etc.

crates/services/block_aggregator_api/src/tests.rs
MitchTurner: > shouldn't this be in a loop?

It doesn't need to be a loop because we are just testing a single iteration of `run`. The fact that `run` is run in a loop in the wild is a the consumer's business.

> why does it need to be spawned separately?

You're right. No need. Fixed.

crates/services/block_aggregator_api/src/lib.rs
MitchTurner: Hmm. I think it all depends on how we design the adapters. They don't need to error, for example, just include an incomplete list.

I'm open to adding `bias`, but it's actually been best practice to avoid it. So I prefer leaving it out unless we really want to enforce some order.

crates/services/block_aggregator_api/src/db.rs
MitchTurner: > better to not have these functions return an impl Storage

Do you mean `impl Future`?

`Storage` might not, but I like to design my ports pretending I don't know what the impl will be.

crates/services/block_aggregator_api/src/blocks.rs
MitchTurner: Moved `Debug` to be `#[cfg(test)]`

MitchTurner: Scratch that. 
Just using `derive` for all because `Bytes` impls `Debug`. I want it for `tracing::debug`, but if `Debug` is too expensive we can do something different on the `tracing::debug`.

Doesn't look too bad though: https://docs.rs/bytes/latest/src/bytes/fmt/debug.rs.html#39

crates/services/block_aggregator_api/src/lib.rs
xgreenx: > i think based on the call we had on Monday, we want to insert blocks into storage in any order i.e if we have a block from the onchain db / importer, they are treated and processed in parallel

I am fine with any implementation. Main idea is that `into_task` shouldn't block `fuel-core`. So if it is done inside of the `run` function, then it works for me=)

I am fine if first we process all blocks from the `on-chain` database first, and only subscribe/start process from the importer after. The main point here, that we shouldn't have a huge queue of blocks in the importer stream(because it will be a lot of data, plus it can block block importer from work because of back pressure).

@MitchTurner I think we shouldn't put this logic on the adapter level, I think we need to architect it from the service itself.

Otherwise, we move a lot of logic outside of the service, overloading the `fuel-core` crate with logic. If we can implement and test the main parts inside the service crate, we should do it there. An adapter should be, in most cases, very primitive and do only basic functionality.

crates/services/block_aggregator_api/src/db.rs
rymnc: yeah, `impl Future`, this can be removed if at the end of this integration we dont use async for the BlockAggregatorDB

crates/services/block_aggregator_api/src/lib.rs
MitchTurner: The followup prs allows both to be syncing at the same time.

crates/services/block_aggregator_api/src/lib.rs
MitchTurner: Sure. That might be needed, but kinda dependent on the implementation. I can add the method to the trait.

MitchTurner: Addressed in https://github.com/FuelLabs/fuel-core/pull/3085/commits/81194d4132a65b1dd2dd1349cabe3fc61d84366e

PR #3092, opened 2025-09-04
Block aggregator db adapter

MitchTurner:
## Linked Issues/PRs

https://github.com/FuelLabs/fuel-core/issues/3088

## Description

implements
```rs
/// The definition of the block aggregator database.
pub trait BlockAggregatorDB: Send + Sync {
    /// The type used to report a range of blocks
    type BlockRange;

    /// Stores a block with the given ID
    fn store_block(
        &mut self,
        height: BlockHeight,
        block: Block,
    ) -> impl Future<Output = Result<()>> + Send;

    /// Retrieves a range of blocks from the database
    fn get_block_range(
        &self,
        first: BlockHeight,
        last: BlockHeight,
    ) -> impl Future<Output = Result<Self::BlockRange>> + Send;

    /// Retrieves the current height of the aggregated blocks If there is a break in the blocks,
    /// i.e. the blocks are being aggregated out of order, return the height of the last
    /// contiguous block
    fn get_current_height(&self) -> impl Future<Output = Result<BlockHeight>> + Send;
}
```

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/block_aggregator_api/src/db/storage_db.rs
```diff
@@ -0,0 +1,190 @@
...
+            highest_contiguous_block,
+            orphaned_heights,
+            storage,
+        }
+    }
+
+    fn update_highest_contiguous_block(&mut self, height: BlockHeight) {
+        if height == self.next_height() {
+            self.highest_contiguous_block = height;
+            while let Some(next_height) = self.orphaned_heights.iter().next().cloned() {
```
rymnc: is there a way we can do this without cloning?

MitchTurner: Very probably. Copilot was pretty eager with that specific function and I just glanced over it before writing the test. Forgot to check if it did anything more subtly stupid.
I'll reevaluate in the morning.

rymnc (commented):
i feel like this is overly complex, why do we care about contiguous blocks?

MitchTurner:
> i feel like this is overly complex, why do we care about contiguous blocks?

The blocks can be stored out of order, but we want the user to be able to see the highest block that's synced contiguously from the beginning. The new logic tracks and updates any gaps in the stored blocks. To do this, we need to track which blocks have been added and "roll up" any continuous blocks once the gaps are filled.

crates/services/block_aggregator_api/src/db/storage_db.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    async fn store_block(&mut self, height: BlockHeight, block: Block) -> Result<()> {
+        self.update_highest_contiguous_block(height);
+        let mut tx = self.storage.write_transaction();
+        tx.storage_as_mut::<Blocks>()
+            .insert(&height, &block)
+            .map_err(|e| Error::DB(anyhow!(e)))?;
+        tx.commit().map_err(|e| Error::DB(anyhow!(e)))?;
+        Ok(())
+    }
```
MitchTurner, line 118: I don't know what the overhead of creating a new storage tx and committing it is, but it might be more performant to only commit when we call `get_block_range` otherwise just hold onto some storage tx. Not gonna over-optimize now.

crates/services/block_aggregator_api/src/db/storage_db/table.rs
```diff
@@ -0,0 +1,64 @@
...
+
+impl Mappable for Blocks {
+    type Key = Self::OwnedKey;
+    type OwnedKey = BlockHeight;
+    type Value = Self::OwnedValue;
+    type OwnedValue = Block;
+}
+
+impl TableWithBlueprint for Blocks {
+    type Blueprint = Plain<Postcard, Postcard>;
```
rymnc, line 58:
```suggestion
    type Blueprint = Plain<Primitive<4>, Postcard>;
```
should be fine here

crates/services/block_aggregator_api/src/db/storage_db.rs
```diff
@@ -0,0 +1,198 @@
...
+                    } else {
+                        break;
+                    }
+                }
+            }
+            Ordering::Greater => {
+                self.orphaned_heights.insert(height);
+            }
+            Ordering::Less => {
+                // ignore duplicate or old block
```
rymnc: warning would be nice here, i dont think we expect this behaviour

crates/services/block_aggregator_api/src/db/storage_db.rs
rymnc: its fine, we do it in compression service when we want to write a new block too

crates/services/block_aggregator_api/src/blocks.rs
```diff
@@ -13,7 +13,7 @@ pub trait BlockSource: Send + Sync {
     fn drain(&mut self) -> impl Future<Output = Result<()>> + Send;
 }
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub struct Block {
     bytes: Bytes,
```
rymnc, line 18: perhaps this struct should be versioned 

```rs
pub struct BlockV1 {
    protobuf_bytes: Bytes
}

pub struct BlockV2 {
    cid: Bytes
}

pub enum Block {
    V1(BlockV1),
    V2(BlockV2),
    // etc
}

impl Block {
    fn as_bytes(&self) -> &Bytes {
        match self {todo!()}
    }
}

```

this way we can model the different data sources as we discussed in the call a couple weeks ago

crates/services/block_aggregator_api/src/result.rs
```diff
@@ -1,6 +1,8 @@
 #[derive(Debug)]
 pub enum Error {
-    ApiError,
-    BlockSourceError,
+    Api,
+    BlockSource,
+    DB(anyhow::Error),
```
rymnc, line 5:
```suggestion
#[derive(thiserror::Error, Debug)]
pub enum Error {
    #[error("api error: `{0}`")]
    Api(String), // or whatever error here
    #[error("block source error: `{0}`")]
    BlockSource(anyhow::Error),
    DB(anyhow::Error),
```

another nice pattern is if you decide to have specific errors for the api, blocksource and db, it can be written this way -

```suggestion
#[derive(thiserror::Error, Debug)]
pub enum Error {
    Api(#[from] ApiError), 
    BlockSource(#[from] BlockSourceError),
    DB(#[from] DatabaseError),
```


might be over-engineering at this point though

crates/services/block_aggregator_api/src/db/storage_db.rs
```diff
@@ -0,0 +1,198 @@
...
+            "Polling next block from storage stream, next height: {:?}",
+            self.next
+        );
+        let this = self.get_mut();
+        if let Some(height) = this.next {
+            let tx = this.inner.read_transaction();
+            let next_block = tx
+                .storage_as_ref::<Blocks>()
+                .get(&height)
+                .map_err(|e| Error::DB(anyhow!(e)));
```
rymnc, line 175: i think a new pattern we have been following in the codebase is to have an extension trait on the storage, which wraps around the whole `storage_as_ref().get().map_err()` bit and makes it something like -


```suggestion
            let next_block = tx.get_block(&height)?;
```

crates/services/block_aggregator_api/src/db/storage_db.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+impl<S> Stream for StorageStream<S>
+where
+    S: Unpin + ReadTransaction + std::fmt::Debug,
+    for<'a> StorageTransaction<&'a S>: StorageInspect<Blocks, Error = StorageError>,
+{
+    type Item = Block;
+
+    fn poll_next(
```
xgreenx, line 161: I think performance of this solution is slower than if we used stream directly.

Maybe we want to reuse `StorageIterator` from `o2`

image


But for that we need to do some benchmarks to see which implementation is faster =)

So it is follow-up task, not for now.

crates/services/block_aggregator_api/src/db/storage_db.rs
```diff
@@ -0,0 +1,198 @@
...
+};
+use table::Blocks;
+
+pub mod table;
+#[cfg(test)]
+mod tests;
+
+pub struct StorageDB<S> {
+    highest_contiguous_block: BlockHeight,
+    orphaned_heights: BTreeSet<BlockHeight>,
```
xgreenx, line 48: How do you plan to use this field?

I think it will be safer to just insert blocks into the database at any height from the importer stream. And keep track of the actual height(which can be less than the importer height), but which you use to sync blocks on startup.

crates/services/block_aggregator_api/src/db/storage_db.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+impl<S, T> BlockAggregatorDB for StorageDB<S>
+where
+    S: Modifiable + std::fmt::Debug,
+    S: KeyValueInspect<Column = Column>,
+    for<'b> StorageTransaction<&'b mut S>: StorageMutate<Blocks, Error = StorageError>,
+    S: AtomicView<LatestView = T>,
+    T: Unpin + Send + Sync + KeyValueInspect<Column = Column> + 'static + std::fmt::Debug,
+    StorageTransaction<T>: AtomicView + StorageInspect<Blocks, Error = StorageError>,
```
rymnc, line 106: does StorageTransaction<T> implement AtomicView? (from the adapters perspective)

crates/services/block_aggregator_api/src/db/storage_db.rs
```diff
@@ -0,0 +1,198 @@
...
+{
+    type BlockRange = BlockRangeResponse;
+
+    async fn store_block(&mut self, height: BlockHeight, block: Block) -> Result<()> {
```
rymnc, line 110: i would prefer if the storage operations were not asynchronous - if we plan to support async backend for the storage i don't think it should be this way. this was a mistake with the Compression traits too, they shouldn't have been async

crates/services/block_aggregator_api/src/db/storage_db/table.rs
rymnc: i dont think postcard is necessary for the value here either, we can have it be `Raw` and just prefix the bytes with the enum and store that in the storage.

crates/services/block_aggregator_api/src/db/storage_db.rs
MitchTurner: Yeah. In general, I don't like to think too hard about performance when implementing code. Obviously, if there is something that is going to make performance really bad, or prevent us from optimizing, then we should discuss that, but it sounds like nothing about the interface is preventing better performance.

crates/services/block_aggregator_api/src/db/storage_db.rs
MitchTurner: The DB isn't aware of which blocks are from the importer and which are from the sync.

I have introduced an enum for
```
#[derive(Debug)]
pub enum BlockSourceEvent {
    NewBlock(BlockHeight, Block),
    OldBlock(BlockHeight, Block),
}
```
in the [next PR](https://github.com/FuelLabs/fuel-core/pull/3096) which could inform this code. It's a domain type, so we can take a dep on it.

But for now, the `orphaned_heights` allows us to track gaps in the blocks:
```
1,2,3,4 .... 8,9,10 ... 15,16,17
```
The `highest_contiguous_block` is 4. But if we add 5,6,7, it will be 10 (not 7).

crates/services/block_aggregator_api/src/db/storage_db/table.rs
MitchTurner: Cool. I had no opinion, just copied some other code. 

`Raw` makes sense if it's just `Bytes` under the hood.

crates/services/block_aggregator_api/src/db/storage_db.rs
MitchTurner: Shouldn't ever happen, but I'll add.

crates/services/block_aggregator_api/src/blocks.rs
MitchTurner: I've captured that requirement with assoc types.

The concrete type is:
```
pub enum BlockRangeResponse {
    Literal(BoxStream<Block>),
    Remote(String),
}
```
and the assoc type is:
```
pub trait BlockAggregatorDB: Send + Sync {
    /// The type used to report a range of blocks
    type BlockRange;
 
    . . .
```

```
pub trait BlockAggregatorApi: Send + Sync {
    /// The type of the block range response.
    type BlockRangeResponse;
```

But now that you bring it up, that doesn't necessarily solve it on the user side. If they're expecting a specific type it doesn't matter if its an assoc type in Rust.

crates/services/block_aggregator_api/src/result.rs
MitchTurner: I've added `thiserror` in [this PR](https://github.com/FuelLabs/fuel-core/pull/3097/files)

Just waited for the the code to demand it before including it, à la TDD.

crates/services/block_aggregator_api/src/db/storage_db.rs
MitchTurner: Might be cleaner. Not against it. This is an adapter though so I don't mind exposing implementation details.

crates/services/block_aggregator_api/src/db/storage_db.rs
MitchTurner: I've added the constraint so we can take the `latest_view` and give ownership to the stream:
```
    async fn get_block_range(
        &self,
        first: BlockHeight,
        last: BlockHeight,
    ) -> Result<BlockRangeResponse> {
        let latest_view = self
            .storage
            .latest_view()
            .map_err(|e| Error::DB(anyhow!(e)))?;
        let stream = StorageStream::new(latest_view, first, last);
        Ok(BlockRangeResponse::Literal(Box::pin(stream)))
    }
```

crates/services/block_aggregator_api/src/db/storage_db.rs
MitchTurner: I can see a YAGNI case fore removing `async` here. As I mentioned before, I don't like prescribing anything about the implementation in the port, but at the same time adding `async` to all port methods is kinda overkill.

crates/services/block_aggregator_api/src/db/storage_db.rs
xgreenx: You don't know about gaps after restart, so I am not sure how do you want to use this code=D

PR #3096, opened 2025-09-09
Add new block subscription to inner service

MitchTurner:
## Description

Addresses comment in https://github.com/FuelLabs/fuel-core/pull/3085
> BTW, we need an option for people to subscribe to blocks for the latest blocks. So you also need to have a stream from the block importer, where you will serialize the block into a protobuf, store it into the database, and broadcast to all listeners.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/block_aggregator_api/src/blocks/importer_and_onchain_source.rs
```diff
@@ -0,0 +1 @@
+
```
rymnc: ?

MitchTurner: Yeah. I was using this PR for another purpose and pivoted during implementation. Removed.

crates/services/block_aggregator_api/src/api.rs
```diff
@@ -65,4 +75,12 @@ impl<T> BlockAggregatorQuery<T> {
         let query = Self::GetCurrentHeight { response: sender };
         (query, receiver)
     }
+
+    pub fn new_block_subscription()
+    -> (Self, tokio::sync::mpsc::Receiver<(BlockHeight, Block)>) {
+        const ARBITRARY_CHANNEL_SIZE: usize = 10;
```
rymnc, line 76: this should probably be larger, because it is `mpsc`, and if the consumer is slow, it will block the producer

crates/services/block_aggregator_api/src/blocks.rs
```diff
@@ -6,13 +6,18 @@ use std::fmt::Debug;
 /// Source from which blocks can be gathered for aggregation
 pub trait BlockSource: Send + Sync {
     /// Asynchronously fetch the next block and its height
-    fn next_block(&mut self)
-    -> impl Future<Output = Result<(BlockHeight, Block)>> + Send;
+    fn next_block(&mut self) -> impl Future<Output = Result<BlockSourceEvent>> + Send;
```
rymnc, line 9: i think this should be boxed, its a rather large type and i think it communicates intention better to someone reading the code

crates/services/block_aggregator_api/src/lib.rs
```diff
@@ -36,12 +33,13 @@ pub struct BlockAggregator<Api, DB, Blocks> {
     query: Api,
     database: DB,
     block_source: Blocks,
+    new_block_subscriptions: Vec<tokio::sync::mpsc::Sender<(BlockHeight, Block)>>,
 }
```
rymnc:
```suggestion
    new_block_subscriptions: Vec<tokio::sync::mpsc::Sender<NewBlock>>,
}

pub struct NewBlock {
    height: BlockHeight,
    block: Block,
}
```

crates/services/block_aggregator_api/src/tests.rs
```diff
@@ -209,3 +213,70 @@ async fn run__get_current_height__returns_expected_height() {
...
+    source_sender.send(event).await.unwrap();
+    let _ = srv.run(&mut watcher).await;
+
+    // then
+    tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
+    let (actual_height, actual_block) =
+        tokio::time::timeout(tokio::time::Duration::from_secs(1), response.recv())
+            .await
+            .unwrap()
+            .unwrap();
```
rymnc: instead of the sleep, we can turn this into a loop

```suggestion
    let (actual_height, actual_block) = tokio::time::timeout(
    tokio::time::Duration::from_secs(1),
    async {
        loop {
            if let Ok(result) = response.try_recv() {
                return result;
            }
            tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
        }
    }
).await.unwrap().unwrap();
```

crates/services/block_aggregator_api/src/tests.rs
```diff
@@ -209,3 +213,70 @@ async fn run__get_current_height__returns_expected_height() {
...
+    let event = BlockSourceEvent::OldBlock(height, block);
+    source_sender.send(event).await.unwrap();
+    let _ = srv.run(&mut watcher).await;
+
+    // then
+    tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
+    let res =
+        tokio::time::timeout(tokio::time::Duration::from_millis(100), response.recv())
+            .await;
+    assert!(res.is_err(), "should have timed out");
```
rymnc, line 276: same as https://github.com/FuelLabs/fuel-core/pull/3096/files#r2349332502

crates/services/block_aggregator_api/src/block_aggregator.rs
```diff
@@ -0,0 +1,142 @@
...
+        res: crate::result::Result<BlockSourceEvent>,
+    ) -> TaskNextAction {
+        tracing::debug!("Handling block: {res:?}");
+        let event = try_or_stop!(res, |e| {
+            tracing::error!("Error receiving block from source: {e:?}");
+        });
+        let (id, block) = match event {
+            BlockSourceEvent::NewBlock(id, block) => {
+                self.new_block_subscriptions.retain_mut(|sub| {
+                    let send_res = sub.try_send((id, block.clone()));
```
rymnc: probably want to put the `Block` in an arc if we have to clone it every time

crates/services/block_aggregator_api/src/block_aggregator.rs
```diff
@@ -0,0 +1,142 @@
...
+                    match send_res {
+                        Ok(_) => true,
+                        Err(tokio::sync::mpsc::error::TrySendError::Full(_)) => {
+                            tracing::error!("Error sending new block to source due to full channel: {id:?}");
```
rymnc:
```suggestion
                            tracing::error!("Error sending new block to sink due to full channel: {id:?}");
```

crates/services/block_aggregator_api/src/blocks.rs
MitchTurner: `BlockSourceEvent`? How big is it? It's just `Bytes` size, essentially. Isn't `Bytes` just a pointer like a `Box`? IDK how big the `vtable` is...
```
pub struct Bytes {
    ptr: *const u8,
    len: usize,
    // inlined "trait object"
    data: AtomicPtr<()>,
    vtable: &'static Vtable,
}
```

crates/services/block_aggregator_api/src/api.rs
MitchTurner: Yes. These constructors are just used in tests for now. I'll add a `#[cfg(test)]` to imply the consumer is responsible for defining the channel size.

crates/services/block_aggregator_api/src/tests.rs
MitchTurner: I don't think I like this as much. It's a lot noisier. What advantage are we getting over just the `timeout`? (the `timeout` already adds a bit of visual noise).

crates/services/block_aggregator_api/src/blocks.rs
rymnc: hmm i see

crates/services/block_aggregator_api/src/tests.rs
rymnc: prevents the tests being flaky because of the sleep before it. it also allows the test to complete faster

crates/services/block_aggregator_api/src/block_aggregator.rs
```diff
@@ -0,0 +1,142 @@
...
+                            tracing::error!("Error sending new block to subscriber due to full channel: {id:?}");
+                            true
+                        },
+                        Err(tokio::sync::mpsc::error::TrySendError::Closed(_)) => {
+                            tracing::debug!("Dropping block subscription due to closed channel");
+                            false
+                        },
+                    }
+                });
+                // do more stuff
```
xgreenx: Do we need this comment?

crates/services/block_aggregator_api/src/lib.rs
```diff
@@ -36,12 +33,28 @@ pub struct BlockAggregator<Api, DB, Blocks> {
...
+    height: BlockHeight,
+    block: Block,
+}
+
+impl NewBlock {
+    pub fn new(height: BlockHeight, block: Block) -> Self {
+        Self { height, block }
+    }
+
+    pub fn decompose(self) -> (BlockHeight, Block) {
```
xgreenx: `into` or `into_inner` will match our code on other places=D

xgreenx (approved):

crates/services/block_aggregator_api/src/tests.rs
MitchTurner: Done. Added a helper for it.

PR #3097, opened 2025-09-10
Block source for aggregator

MitchTurner:
## Linked Issues/PRs

https://github.com/FuelLabs/fuel-core/issues/3087

## Description

This adapter holds two services of its own:
- Importer Task
- Sync Task

The importer task is responsible for getting new blocks as they are produced. This will run in perpetuity.

The sync task will just find any blocks that have already been produced and won't be streamed through the importer. This will run on startup, but should hopefully finish very quickly. We decided not to have the sync task be blocking, like we might do on more critical systems, instead it runs once all the services are started to not block anyone else.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+
+impl<Serializer, DB> RunnableTask for SyncTask<Serializer, DB>
+where
+    Serializer: BlockSerializer + Send + Sync,
+    DB: Send + Sync + 'static,
+    DB: StorageInspect<FuelBlocks> + Send + 'static,
+    DB: StorageInspect<Transactions> + Send + 'static,
+    <DB as StorageInspect<FuelBlocks>>::Error: std::fmt::Debug + Send,
```
rymnc: can we move this all out to a new trait?
```rs
pub trait SyncTaskDb: Send + Sync + StorageInspect<FuelBlocks> + StorageInspect<Transactions> + <Self as StorageInspect<FuelBlocks>>::Error: std::fmt::Debug {}


impl<S> SyncTaskDb for S where ... {}
```

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
```diff
@@ -0,0 +1,146 @@
...
+                                tracing::debug!("tx id not found in db: {:?}", tx_id);
+                                todo!()
+                            }
+                            Err(_) => {
+                                tracing::debug!("error while finding tx: {:?}", tx_id);
+                                todo!()
+                            }
+                        }
+                    }
+                    let block = <fuel_core_types::blockchain::block::Block<TxId> as Clone>::clone(&compressed_block).uncompress(txs);
```
rymnc: why do we need this clone?

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
```diff
@@ -0,0 +1,146 @@
...
+#[async_trait::async_trait]
+impl<Serializer, DB> RunnableService for SyncTask<Serializer, DB>
+where
+    Serializer: BlockSerializer + Send + Sync + 'static,
+    DB: Send + Sync + 'static,
+    DB: StorageInspect<FuelBlocks> + Send + 'static,
+    DB: StorageInspect<Transactions> + Send + 'static,
+    <DB as StorageInspect<FuelBlocks>>::Error: std::fmt::Debug + Send,
+{
+    const NAME: &'static str = "BlockSourceInnerService";
```
rymnc: probably should be a different name than the ImporterTask

crates/services/block_aggregator_api/src/blocks/importer_and_db_source.rs
```diff
@@ -0,0 +1,111 @@
...
+    <DB as StorageInspect<FuelBlocks>>::Error: std::fmt::Debug + Send,
+{
+    pub fn new(
+        importer: BoxStream<SharedImportResult>,
+        serializer: Serializer,
+        database: DB,
+        db_starting_height: BlockHeight,
+        db_ending_height: BlockHeight,
+    ) -> Self {
+        const ARB_CHANNEL_SIZE: usize = 100;
```
rymnc, line 71: what happens to the import task if the service consuming the `BlockSource` is slow?

crates/services/block_aggregator_api/Cargo.toml
```diff
@@ -5,21 +5,25 @@ edition = "2024"
 
 [dependencies]
 anyhow = { workspace = true }
+async-trait = { workspace = true }
 bytes = { workspace = true, features = ["serde"] }
 enum-iterator = { workspace = true }
 fuel-core-services = { workspace = true }
 fuel-core-storage = { workspace = true, features = ["std"] }
-fuel-core-types = { workspace = true, features = ["std"] }
+fuel-core-types = { workspace = true, features = ["std", "test-helpers"] }
```
rymnc: do we need `test-helpers` in the dependencies instead of dev-deps?

crates/services/block_aggregator_api/src/blocks/importer_and_db_source.rs
MitchTurner: I want it to wait on the send and not import more blocks:
```
                let res = self.block_return_sender.send(event).await;
```

In theory, this could cause problems upstream with the importer as well...

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
MitchTurner: Because it's a `Cow`. I was having a hard time getting the compiler to accept just `into_owned()` (even though it's just a `clone` under the hood). But this worked and I moved on to something else.

crates/services/block_aggregator_api/src/blocks/importer_and_db_source.rs
MitchTurner: Considering imports come 1/second and the consumer isn't `bias;`, it shouldn't ever be a problem. But definitely need to be careful not to mess it up.

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
MitchTurner: I'm opposed to this pattern in general. It creates misdirection when reading the code. Yes it's redundant and kinda noisy to have all the trait deps, but I feel like combined traits make things harder to understand.

Keeping requirements close to where it's used > Dry

Not to mention, stacking the requirements into a single trait makes it hard to tell if you are over constraining some generics.

It's essentially an ISP violation ([SOLID](https://en.wikipedia.org/wiki/SOLID)).

crates/services/block_aggregator_api/src/blocks/importer_and_db_source.rs
```diff
@@ -0,0 +1,131 @@
...
+    DB: StorageInspect<FuelBlocks, Error = E>,
+    DB: StorageInspect<Transactions, Error = E>,
+    E: std::fmt::Debug + Send + Sync,
+{
+    async fn next_block(&mut self) -> Result<BlockSourceEvent> {
+        tracing::debug!("awaiting next block");
+        // self.receiver
+        //     .recv()
+        //     .await
+        //     .ok_or(Error::BlockSource(anyhow!("Block source channel closed")))
```
xgreenx, line 114:
```suggestion
```

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
```diff
@@ -0,0 +1,171 @@
...
+    fn get_block(
+        &self,
+        height: &BlockHeight,
+    ) -> Result<Option<fuel_core_types::blockchain::block::Block>, E> {
+        let maybe_block = StorageInspect::<FuelBlocks>::get(&self.db, height)?;
+        if let Some(block) = maybe_block {
+            let tx_ids = block.transactions();
+            let txs = self.get_txs(tx_ids)?;
+            let block =
+                <fuel_core_types::blockchain::block::Block<TxId> as Clone>::clone(&block)
```
xgreenx, line 80: I think you can just do `into_owned`, under the hood it is already `Owned` type

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
```diff
@@ -0,0 +1,171 @@
...
+};
+use tokio::sync::mpsc::Sender;
+
+pub struct SyncTask<Serializer, DB> {
+    serializer: Serializer,
+    block_return_sender: Sender<BlockSourceEvent>,
+    db: DB,
+    next_height: BlockHeight,
+    maybe_stop_height: Option<BlockHeight>,
+    new_ending_height: tokio::sync::oneshot::Receiver<BlockHeight>,
```
xgreenx, line 36: Why is it a `oneshot` channel? You only can read it once. And it is not clear to me how do you want to use it=)

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    async fn maybe_update_stop_height(&mut self) {
+        if let Ok(last_height) = self.new_ending_height.try_recv() {
+            tracing::info!("updating last height to {}", last_height);
+            self.maybe_stop_height = Some(last_height);
+        }
+    }
+
+    fn get_block(
```
xgreenx, line 71: I would like to emphasize that we also require receipts from this service. We should be able to return a block with transactions and receipts.

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
xgreenx: Is the main idea that the importer service will send the first height only once, and after that, the sync service will know when to stop?

Maybe it is better to pass on the chain height on creation of the service to avoid this complicated link between services?

xgreenx (approved):

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
MitchTurner: I agree. It's complicated. But I want to avoid some race condition of telling it the height, but then missing one height from the importer.

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
MitchTurner: Yes. I haven't taken that into account yet. Shouldn't be hard to add later. I'll create an issue.

MitchTurner: https://github.com/FuelLabs/fuel-core/issues/3108

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
MitchTurner: Yes. I had tried that before and the compiler couldn't resolve the type for some reason. This was the first thing that worked.

Just tried on the newer branch and it worked, so I don't know what changed or what was wrong before.

PR #3099, opened 2025-09-15
Use hash-based balance_root and state_root fields

Dentosal:
## Description

@xgreenx writes on [slack](https://fuellabs.slack.com/archives/C01KVGRDZHB/p1755856188397269?thread_ts=1749203886.570819&cid=C01KVGRDZHB):

> Storing storage slots in the transaction makes the blob really, really big; it can even be bigger than the allowance for the block, but we will only know about its size after the execution. So having a malleable list of storage slots is a bad idea, not taking into account that it is not compressible and will make p2p worse.
Also, on the fraud-proving side, in the case of the Validium solution, these storage slots are useless, because we need to prove in one go the latest values for storage slots(which requires a global merkle root).  In the case of V2 fraud proof solution(where we prove the execution only of the transaction), it is more useful because it allows executing the transaction in one go(assuming storage slots are correct). But it is needed only in the case that a block producer used incorrect values during the execution of the transaction.
We can get almost the same properties by going another route:
> 1. Use a hash of all storage slots/balances(hash over sorted list of (Key, Option<Value>)) and use balance_root and state_root fields in the input and output.
> 2. For input of the contract, we will use all storage slots/balances accessed during the execution, even if the value for them is None in the storage. In this case, we can be sure that the block producer listed all the required information to execute the transaction.
> 3. Output will contain a hash of the same storage slots/balances after the execution.
> 4. Regular blocks and transactions that we use to sync with the network(and on UI) will not be affected. But anyone will be able to request additional metadata for a transaction, like storage slots before and after the execution(we will store them in the off-chain database).
> 5. For the validium case, we will just create a global merkle root of the storage and balances for all contracts. It will be much faster, because it is just one bulk update into SMT(a bulk update of 100_000 updates is faster than 100_000 single updates in 50-100 times).
> 6. In the case of V2 fraud proofs, the case with incorrect initial values is already a 2-round game; we will just add an additional round, where the block producer should reveal initial values(it can be done on-chain or off-chain). If we don't want to have an additional step here, then we can post this data along with transactions to DA.
In this case, we don't change the layout of the transaction(we use all the fields that we have right now; the tx_pointer field is useless in this context, but it is fine). We don't affect the network. This information will be accessible via GraphQL.
Implementation, then, is very straightforward and fast. The only problem that we need to solve is how to post the old state of the network(old storage slots and balances before we started to use this approach with storage slots). But we have this issue regardless of the approach we use. So it is a separate topic.

## Checklist

### Before requesting review

### DIFF
--- a/crates/fuel-core/src/executor.rs
+++ b/crates/fuel-core/src/executor.rs
@@ -1689,10 +1689,9 @@ mod tests {
 
         // Ensure all txs succeeded.
         assert!(
-            tx_status.iter().all(|s| (matches!(
-                s.result,
-                TransactionExecutionResult::Success { .. }
-            )))
+            tx_status
+                .iter()
+                .all(|s| matches!(s.result, TransactionExecutionResult::Success { .. }))
         );
 
         // Assert the balance and state roots are same before and after execution.
@@ -2141,10 +2140,9 @@ mod tests {
             block, tx_status, ..
         } = executor.produce_and_commit(block).unwrap();
         assert!(
-            tx_status.iter().all(|s| (matches!(
-                s.result,
-                TransactionExecutionResult::Success { .. }
-            )))
+            tx_status
+                .iter()
+                .all(|s| matches!(s.result, TransactionExecutionResult::Success { .. }))
         );
 
         let executed_tx = block.transactions()[1].as_script().unwrap();
@@ -2504,10 +2502,9 @@ mod tests {
             block, tx_status, ..
         } = executor.produce_and_commit(block).unwrap();
         assert!(
-            tx_status.iter().all(|s| (matches!(
-                s.result,
-                TransactionExecutionResult::Success { .. }
-            )))
+            tx_status
+                .iter()
+                .all(|s| matches!(s.result, TransactionExecutionResult::Success { .. }))
         );
 
         // Check resulting roots
--- a/crates/storage/src/iter/changes_iterator.rs
+++ b/crates/storage/src/iter/changes_iterator.rs
@@ -201,7 +201,7 @@ where
         prefix: Option<&[u8]>,
         start: Option<&[u8]>,
         direction: IterDirection,
-    ) -> BoxedIter<KVWriteItem> {
+    ) -> BoxedIter<'_, KVWriteItem> {
         match self.changes {
             StorageChanges::Changes(changes) => {
                 if let Some(tree) = changes.get(&column.id()) {
@@ -239,7 +239,7 @@ where
     pub fn iter_all_writes<M>(
         &self,
         direction: Option<IterDirection>,
-    ) -> BoxedIter<StorageResult<(M::OwnedKey, Option<M::OwnedValue>)>>
+    ) -> BoxedIter<'_, StorageResult<(M::OwnedKey, Option<M::OwnedValue>)>>
     where
         M: Mappable,
         M: TableWithBlueprint<Column = Column>,


crates/services/executor/src/executor.rs
```diff
@@ -1590,7 +1596,9 @@ where
     where
         T: KeyValueInspect<Column = Column>,
     {
-        let mut sub_block_db_commit = storage_tx
+        let mut storage_tx_record = StorageAccessRecorder::new(&mut *storage_tx);
```
xgreenx: We should implement all methods for the `KeyValueInspect`, if we want to make it work optimally. Otherwise execution will suffer a lot, especially for contracts byte code

image

crates/services/executor/src/storage_access_recorder.rs
```diff
@@ -0,0 +1,56 @@
...
+    }
+}
+
+impl<S> KeyValueInspect for StorageAccessRecorder<S>
+where
+    S: KeyValueInspect,
+{
+    type Column = S::Column;
+
+    fn get(&self, key: &[u8], column: Self::Column) -> StorageResult<Option<Value>> {
```
xgreenx, line 199: I think you can define a recorder just dedicated to balances and storage slots. In this case you need to store less data.

Also, if you are recording on `StorageInspect` level instead of `KeyValueInspect`(so you create write transaction first and wrap `StorageTransaction` into `StorageAccessRecorder`), you don't need to spend resources for serialization and deserialization.

Dentosal: Like this 486612d?

Dentosal: Or like this 95c98de982687853fc7fd55785f1c6dc860a9de0?

crates/services/executor/src/executor.rs
Dentosal: Done in 95c98de982687853fc7fd55785f1c6dc860a9de0

crates/services/executor/src/executor.rs
```diff
@@ -1419,7 +1444,11 @@ where
                 )?;
             }
 
-            self.compute_inputs(core::slice::from_mut(&mut input), storage_tx)?;
+            self.compute_inputs(
+                core::slice::from_mut(&mut input),
+                storage_tx,
+                &Default::default(),
```
xgreenx: Coinbase is a contract, while right now we use very simple contract that doesn't touch the storage(but touches balances), in the future it can be more complex.

We need to record and update values correctly for `Mint` transaction as well

crates/services/executor/src/executor.rs
```diff
@@ -1608,7 +1639,11 @@ where
         )
         .map_err(|e| format!("{e}"))
         .map_err(ExecutorError::CoinbaseCannotIncreaseBalance)?;
-        sub_block_db_commit.commit()?;
+
+        let (recorder, changes) = sub_block_db_commit.into_inner();
+        let record = core::mem::take(&mut *recorder.record.lock()).finalize(&changes);
```
xgreenx: Is it possible to hide this functionality inside of the `Recorder`?

crates/services/executor/src/executor.rs
```diff
@@ -1949,17 +1992,25 @@ where
...
         Self::update_input_used_gas(predicate_gas_used, tx_id, &mut tx)?;
 
+        let (
+            StorageAccessRecorder {
+                storage: storage_tx_recovered,
+                record,
+            },
+            changes,
+        ) = sub_block_db_commit.into_inner();
+        let record = record.lock().clone().finalize(&changes);
```
xgreenx: The same here, it would be nice if `Mutex` is hidden inside of the `StorageAccessRecorder`

crates/services/executor/src/executor.rs
```diff
@@ -1949,17 +1992,25 @@ where
...
+
         // We always need to update inputs with storage state before execution,
         // because VM zeroes malleable fields during the execution.
-        self.compute_inputs(tx.inputs_mut(), storage_tx)?;
+        self.compute_inputs(tx.inputs_mut(), storage_tx_recovered, &record)?;
+        self.update_tx_outputs(tx_id, &mut tx, &record, storage_tx, &changes)?;
```
xgreenx: If the transaction is reverted, we shouldn't update outputs with incorrect information. We should use values before execution.

crates/services/executor/src/executor.rs
```diff
@@ -2235,9 +2301,71 @@ where
...
+                let empty = ContractAccesses::default();
+                let for_contract = record.get(contract_id).unwrap_or(&empty);
+
+                let empty = BTreeMap::default();
+                let asset_changes = changes
+                    .get(&Column::ContractsAssets.as_u32())
+                    .unwrap_or(&empty);
+                let state_changes = changes
+                    .get(&Column::ContractsState.as_u32())
+                    .unwrap_or(&empty);
```
xgreenx: You don't need to work with `Changes`, instead, you can work with `ChangesIterator`. In this case you don't need manually do decoding and work with columns, you can just specify the table which you want to iterate over.

crates/services/executor/src/executor.rs
```diff
@@ -2235,9 +2301,71 @@ where
...
+                                    WriteOperation::Remove => Ok(0),
+                                }
+                            })
+                        {
+                            return changed_value;
+                        }
+                        db.storage::<ContractsAssets>()
+                            .get(&ContractsAssetKey::new(contract_id, key))
+                            .map(|v| v.map_or(0, |c| c.into_owned()))
+                            .map_err(|err| ExecutorError::StorageError(err.to_string()))
```
xgreenx: It is a massive closure(the same below). I think we should calculate hashes before we enter this function.

You can have a separate function that calculates hashes before execution and uses `storage_transaction_before_execution` and `record` as an input. And function that calculates hashes after the execution and gets `Changes` as an input.

crates/services/executor/src/executor.rs
Dentosal: Done in 3ecc453b63

crates/services/executor/src/executor.rs
Dentosal: Done in 3ecc453b63

crates/services/executor/src/executor.rs
Dentosal: Done in 1850e33c5d. I was under the impression we wouldn't store them anyway for reverted txs?

Dentosal:
> We also should think about the way how we will return "touched" storage slots and balances to the user. It should be accessible via GraphQL/RPC, so we need to define the way hot to store this information off-chain.

What's the reason for returning this?

crates/services/executor/src/executor.rs
Dentosal: Having two separate functions will not work. Computing the inputs requires knowing what keys have changed. That said, we can do all this up-front, and maybe that'll be a bit more clear. I moved the whole collection  logic to `ReadsPerContract::finalize` in 3e6ef07d7a.

crates/services/executor/src/executor.rs
Dentosal: Fixed while doing 3e6ef07d7a

crates/services/executor/src/executor.rs
Dentosal: It won't work as the `ChangesIterator` skips removed keys (instead of returning the values wrapped in an `Option<_>`). Do you think it would make sense to add new methods to `ChangesIterator` to allow this use case?

Dentosal: Done in f48fbeded2. Not sure if that's actually worth `4 files changed, 386 insertions(+), 37 deletions(-)`.

rymnc:
> What's the reason for returning this? The information is already available using historical execution storageReadReplay, but that relies on re-executing the tx. Is the idea to do this without re-execution? Do we need to store only accessed/changed keys, or the values too?

@Dentosal i think the reasoning for this is for fraud proving when we try to get this information from the fuel-defender binary ~ we could do historical execution but it could be slower, yeah?

crates/services/executor/Cargo.toml
```diff
@@ -32,6 +32,7 @@ fuel-core-types = { workspace = true, default-features = false, features = [
 ] }
 parking_lot = { workspace = true }
 serde = { workspace = true }
+sha2 = "0.10"
```
rymnc:
```suggestion
sha2 = { version = "0.10", default-features = false }
```

and then add the std feature above on line 15 :)

crates/storage/src/merkle/sparse.rs
```diff
@@ -114,17 +115,11 @@ where
     type OwnedValue = <Table as Mappable>::OwnedValue;
 }
 
-type KeyCodec<Table, TC> =
-    <<Table as TableWithBlueprint>::Blueprint as BlueprintInspect<
-        Table,
-        DummyStorage<MerkleizedColumn<TC>>,
-    >>::KeyCodec;
+type KeyCodec<Table> =
```
rymnc, line 118: thank you for creating `BlueprintCodec` :)

crates/storage/src/iter.rs
```diff
@@ -340,6 +395,228 @@ pub trait IteratorOverTable {
...
+        start: Option<&M::Key>,
+        direction: Option<IterDirection>,
+    ) -> BoxedIter<super::Result<(M::OwnedKey, Option<M::OwnedValue>)>>
+    where
+        P: AsRef<[u8]>,
+    {
+        #[allow(clippy::redundant_closure)]
+        // false positive: https://github.com/rust-lang/rust-clippy/issues/14215
+        let encoder = start
+            .map(|start| <M::Blueprint as BlueprintCodec<M>>::KeyCodec::encode(start));
```
rymnc: seeing as you are using this pattern a lot, would it be meaningful to add -

(simplified snippet, must be missing a trait here or there)
```rs
pub type KeyCodecOf<M, C> = <<M as TableWithBlueprint<Column = C>> as BlueprintCodec<M>::KeyCodec;
```

and then in this code:
```suggestion
            .map(|start| KeyCodecOf::<M, Column>::encode(start));
```

and elsewhere ? im ok with the current implementation, it has less indirection while reviewing the code

rymnc: same for decode, of course

Dentosal: I don't feel like this is used enough to justify the alias.

crates/services/executor/src/executor.rs
```diff
@@ -1608,7 +1621,16 @@ where
...
+
+        let (recorder, changes) = sub_block_db_commit.into_inner();
+        let (state_before, state_after) = recorder
+            .get_reads()
+            .finalize(&storage_tx, &changes)
+            .map_err(|err| ExecutorError::StorageError(err.to_string()))?;
+
+        self.compute_inputs(core::slice::from_mut(input), storage_tx, &state_before)?;
+
+        storage_tx.commit_changes(changes.clone())?;
```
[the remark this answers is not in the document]
Dentosal: None of this makes any sense

xgreenx (approved):

PR #3101, opened 2025-09-19
Integrate Block Aggregator

MitchTurner:
## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/block_aggregator_api/src/lib.rs
```diff
@@ -74,3 +147,27 @@ where
...
+    DB: BlockAggregatorDB<BlockRangeResponse = BlockRange>,
+    Blocks: BlockSource,
+    BlockRange: Send,
+{
+    const NAME: &'static str = "BlockAggregatorService";
+    type SharedData = ();
+    type Task = Self;
+    type TaskParams = ();
+
+    fn shared_data(&self) -> Self::SharedData {}
```
rymnc, line 180: lets add the port here so its easier to test against it

crates/services/block_aggregator_api/src/api/protobuf_adapter.rs
```diff
@@ -182,4 +194,10 @@ impl BlockAggregatorApi for ProtobufAPI {
...
 }
 
-pub struct ProtobufClient;
+impl Drop for ProtobufAPI {
+    fn drop(&mut self) {
+        if let Some(shutdown_sender) = self.shutdown_sender.take() {
+            let _ = shutdown_sender.send(());
+        }
+    }
+}
```
MitchTurner, line 215: This is better than nothing, but if it isn't dropped properly it will be orphaned still. Might be fine.

crates/fuel-core/src/service/config.rs
```diff
@@ -102,6 +109,12 @@ pub struct Config {
     pub memory_pool_size: usize,
 }
 
+#[cfg(feature = "test-helpers")]
+pub fn free_local_addr() -> SocketAddr {
+    let listener = TcpListener::bind("[::1]:0").unwrap();
```
rymnc, line 114: ipv6?

crates/fuel-core/Cargo.toml
```diff
@@ -64,6 +65,7 @@ clap = { workspace = true, features = ["derive"] }
 cosmrs = { version = "0.21", optional = true }
 derive_more = { version = "0.99" }
 enum-iterator = { workspace = true }
+fuel-block-aggregator-api = { workspace = true }
```
rymnc: dont we want to make this an optional dependency for now?

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/serializer_adapter.rs
```diff
@@ -0,0 +1,23 @@
...
+};
+
+use anyhow::anyhow;
+use fuel_core_types::blockchain::block::Block as FuelBlock;
+use postcard::to_allocvec;
+
+#[derive(Clone)]
+pub struct SerializerAdapter;
+
+impl BlockSerializer for SerializerAdapter {
```
rymnc, line 124: do i understand correctly that this is the lifecycle of the block with all the serde:

```
Block => postcard::to_allocvec => protobuf::serialize
```
?

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
```diff
@@ -134,6 +135,7 @@ where
             self.next_height = BlockHeight::from((*next_height).saturating_add(1));
         } else {
             tracing::warn!("no block found at height {:?}, retrying", next_height);
+            tokio::time::sleep(Duration::from_millis(10)).await;
```
rymnc: hmm, is there a way we can throttle how many times run is called without a sleep here for more determinism? i find 10ms to be very arbitrary here

crates/services/importer/src/importer.rs
```diff
@@ -519,6 +520,8 @@ where
                 }
             }
         }
+        let a = 100;
+        let _ = a;
```
rymnc: ?

tests/tests/rpc.rs
```diff
@@ -0,0 +1,174 @@
...
+    let actual_bytes = if let Some(Payload::Literal(block)) = rpc_client
+        .get_block_range(request)
+        .await
+        .unwrap()
+        .into_inner()
+        .next()
+        .await
+        .unwrap()
+        .unwrap()
+        .payload
```
rymnc: :O

tests/tests/rpc.rs
```diff
@@ -0,0 +1,174 @@
...
+    {
+        block.data
+    } else {
+        panic!("expected literal block payload");
+    };
+    let actual_block: Block<Transaction> = postcard::from_bytes(&actual_bytes).unwrap();
```
rymnc: i think we should add serde logic to the client itself, that would be helpful so people dont have to pull in postcard as a dependency to deserialize blocks on their own. also... postcard cant be deserialized in the browser via a native js library so how are we working with those constraints?

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/serializer_adapter.rs
MitchTurner: This is outdated.

In upcoming PRs, this will be Block -> ProtoBlock -> Client (Rust or TS) Block.

Blocks will be serialized using Protobuf's built-in serialization instead of postcard when saved to the DB.

tests/tests/rpc.rs
MitchTurner: This code is outdated now.

In the later PRs we are checking the protobuf response.

benches/benches/block_target_gas.rs
```diff
@@ -401,6 +401,7 @@ fn service_with_many_contracts(
                 Default::default(),
                 Default::default(),
                 Default::default(),
+                Default::default(),
```
xgreenx, line 404: If we add a new database, we should include it in all places where we already interact with databases, like `check_version`, `rollback_to`, and so on(check where do we use databa liek `relayer`)

crates/fuel-core/src/combined_database.rs
```diff
@@ -326,6 +342,10 @@ impl CombinedDatabase {
         &self.compression
     }
 
+    pub fn block_aggregation(&self) -> &Database<BlockAggregatorDatabase> {
```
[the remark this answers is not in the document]
xgreenx: Cursor is right=D

crates/fuel-core/src/database.rs
```diff
@@ -441,6 +442,12 @@ impl Modifiable for Database<GasPriceDatabase> {
     }
 }
 
+impl Modifiable for Database<BlockAggregatorDatabase> {
+    fn commit_changes(&mut self, changes: Changes) -> StorageResult<()> {
+        commit_changes_with_height_update(self, changes, |_iter| Ok(Vec::new()))
```
xgreenx, line 447: I would like to mention the reason why this database doesn't force a monotonic increase in height.

xgreenx: Also, becuase we don't force monotonic height increase, we can't actually re-use rollback feature from historicat database. But we still need the ability to reset the state of the block aggregator to height X. So you can add another function, which we can call in `rollback_to`.

crates/fuel-core/src/service/config.rs
xgreenx: Yeha, let's try to use ip4, if it is possible

crates/services/block_aggregator_api/src/api/protobuf_adapter.rs
```diff
@@ -153,15 +154,26 @@ impl ProtobufAPI {
...
         let server = Server::new(query_sender);
         let addr = url.parse().unwrap();
+        let (shutdown_sender, shutdown_receiver) = tokio::sync::oneshot::channel::<()>();
         let _server_task_handle = tokio::spawn(async move {
-            tonic::transport::Server::builder()
-                .add_service(block_aggregator_server::BlockAggregatorServer::new(server))
-                .serve(addr)
-                .await
-                .unwrap();
+            let service = tonic::transport::Server::builder()
```
xgreenx, line 170: I know that `ServiceRunner` here is an overkill, but we have GraphQL that looks the same, and it was fiiiiine=D Maybe for consistency plus some logging we could reuse it here as well. It's just an internal service for your main service

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
```diff
@@ -98,6 +99,12 @@ where
...
     }
+
+    // For now just have arbitrary 10 ms sleep to avoid busy looping.
+    // This could be more complicated with increasing backoff times, etc.
+    async fn go_to_sleep_before_continuing(&self) {
+        tokio::time::sleep(Duration::from_millis(10)).await;
+    }
 }
 
 impl<Serializer, DB, E> RunnableTask for SyncTask<Serializer, DB>
```
xgreenx: [nit] https://github.com/FuelLabs/fuel-rust-indexer/blob/main/crates/receipts_manager/service.rs#L584

We can replace the whole service with stream of joined events(old + new)

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
```diff
@@ -134,6 +141,7 @@ where
             self.next_height = BlockHeight::from((*next_height).saturating_add(1));
         } else {
             tracing::warn!("no block found at height {:?}, retrying", next_height);
+            self.go_to_sleep_before_continuing().await;
```
xgreenx, line 145: It would be nice to investigate why was it added and maybe remove it

tests/test-helpers/Cargo.toml
```diff
@@ -38,3 +41,4 @@ reqwest = { workspace = true }
 serde_json = { workspace = true }
 tokio = { workspace = true }
 tempfile = { workspace = true }
+async-trait = {workspace = true}
```
xgreenx, line 44:
```suggestion
async-trait = { workspace = true }
```

tests/test-helpers/src/client_ext.rs
```diff
@@ -0,0 +1,64 @@
...
+            BlockByHeightArgsFields,
+            Consensus,
+            Header,
+        },
+        schema,
+        tx::OpaqueTransaction,
+    },
+};
+
+#[derive(cynic::QueryFragment, Debug)]
```
xgreenx, line 18: I think we can move fetching of the full blocks to be a part of default functionality provided by `FuelClient` along with old functionality

Issue #2175, opened 2024-09-06
Handle to big relayer log page size


xgreenx:
## Overview

Some Ethereum RPC provides usually don't allow more than `10_000` events per response. Because of that, setting a big value for `--relayer-log-page-size` may lead to a stuck synchronization with L1.

We need to teach the Relayer service to adjust the page size. 

## Definition of done
When the RPC provider returns an error that the response is too big, we need to decrease the page size automatically.

If we decreased the page size, and for some period of time we didn't have any error, we should try to increase it.

The increased page size shouldn't be more than `--relayer-log-page-size`.

pwalski:
I tried to set `https://sepolia.infura.io/v3/..` RPC server as a relayer together with few servers from https://chainlist.org/chain/11155111. 
`relayer-log-page-size` was set to `1_000_000` in order to collect error responses.

Returned errors had no common messages nor a common error codes:
`code: -32701, message: "exceed maximum block range: 50000"`
`code: -32005, message: "query returned more than 10000 results. Try with this block range [0x5AA536, 0x5BB3FF]."`
`code: -32602, message: "query exceeds max block range 100000"`

No idea how many different combinations of error codes and messages may indicate exceeding a block range limit.
Even in case of infura code `-32005` could indicate exceeding daily request count quota (https://docs.infura.io/api/networks/ethereum/how-to/avoid-rate-limiting) instead of exceeding a block range limit.

Disclaimer: I have no prior experience with eth RPC servers and I wanted to just implement some requested block range exponential backoff based on limit errors. I will not continue because the task requires a decision which cases of error responses should be handled and how to distinguish them from e.g. exceeding request count quota.

PR #3104, opened 2025-09-26
Add dynamic adjustment for relayer log page size.

its-saeed:
## Linked Issues/PRs
Fixes: #2175 

## Description
- Introduced dynamic adjustment logic for --relayer-log-page-size to prevent sync failures caused by oversized Ethereum RPC responses.
- Page size now grows or shrinks based on recent success/failure patterns.

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/services/relayer/src/service.rs
+++ b/crates/services/relayer/src/service.rs
@@ -117,6 +117,9 @@ pub struct Task<P, D> {
     shutdown: StateWatcher,
     /// Retry on error
     retry_on_error: bool,
+    current_page_size: u64,
+    max_page_size: u64,
+    successful_persisted_logs_counter: u64,
 }
 
 impl<P, D> NotInitializedTask<P, D>
@@ -168,6 +171,20 @@ where
         &mut self,
         eth_sync_gap: &state::EthSyncGap,
     ) -> anyhow::Result<()> {
+        const PAGE_GROW_FACTOR_NUM: u64 = 125;
+        const PAGE_GROW_FACTOR_DEN: u64 = 100;
+        const PAGE_SHRINK_FACTOR: u64 = 2;
+        const GROW_THRESHOLD: u64 = 50;
+
+        if self.successful_persisted_logs_counter > GROW_THRESHOLD
+            && self.current_page_size < self.max_page_size
+        {
+            let grown = self.current_page_size.saturating_mul(PAGE_GROW_FACTOR_NUM)
+                / PAGE_GROW_FACTOR_DEN;
+            self.current_page_size = grown.min(self.max_page_size);
+            self.successful_persisted_logs_counter = 0;
+        }
+
         let logs = download_logs(
             eth_sync_gap,
             self.config.eth_v2_listening_contracts.clone(),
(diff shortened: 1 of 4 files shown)


crates/services/relayer/src/service.rs
```diff
@@ -168,15 +213,44 @@ where
         &mut self,
         eth_sync_gap: &state::EthSyncGap,
     ) -> anyhow::Result<()> {
+        let mut successful_rpc_calls: u64 = 0;
+        let mut rpc_error: bool = false;
+
+        println!(
```
xgreenx:
```suggestion
        tracing::info!(
```

crates/services/relayer/src/service.rs
```diff
@@ -168,15 +213,44 @@ where
...
+                rpc_error = true;
+                println!("RPC error encountered");
+            }
+        });
+
         let logs = logs.take_until(self.shutdown.while_started());
+        let result = write_logs(&mut self.database, logs).await;
+
+        println!("Total successful RPC calls: {}", successful_rpc_calls);
+        println!("Had RPC error: {}", rpc_error);
```
xgreenx: I think we an remove it, or you can replace it with `tracing::debug`

crates/services/relayer/src/service.rs
xgreenx: And we need to do something with other `println`

crates/services/relayer/src/service.rs
```diff
@@ -168,15 +213,44 @@ where
...
 
-        write_logs(&mut self.database, logs).await
+        self.adaptive_page_sizer
+            .update(successful_rpc_calls, rpc_error);
```
xgreenx: We only will update the page size at the end of long running function. In the case if we need to sync 200k blocks, we want to increase the page size more often.

xgreenx (commented):
Also, it would be nice, if we checked the size of the response from Ethereal RPC. If the output amount is more than `X`(where `X` is configurable via CLI, and by default is `10_000`), we also decrease the page size.

crates/services/relayer/src/service.rs
```diff
@@ -102,6 +102,50 @@ pub struct NotInitializedTask<P, D> {
...
+            self.successful_rpc_calls = 0;
+            self.current = (self.current / PAGE_SHRINK_FACTOR).max(1);
+        } else {
+            self.successful_rpc_calls = self
+                .successful_rpc_calls
+                .saturating_add(successful_rpc_calls);
+            if self.successful_rpc_calls >= self.grow_threshold && self.current < self.max
+            {
+                let grown = self.current.saturating_mul(PAGE_GROW_FACTOR_NUM)
+                    / PAGE_GROW_FACTOR_DEN;
```
xgreenx: If the `current == 1`, you never will increase it. Maybe we should handle this edge case

crates/services/relayer/src/service.rs
its-saeed: Ah, my bad! I did them to better understand the code.

crates/services/relayer/src/service.rs
its-saeed: Good catch! Even 2 :)!

crates/services/relayer/src/service/get_logs.rs
```diff
@@ -57,16 +58,14 @@ where
...
-                        let page = page.reduce();
-
                         // Get the logs and return the reduced page.
                         eth_node
                             .get_logs(&filter)
                             .await
                             .map_err(|err| {
                                 let ProviderError::JsonRpcClientError(err) = err else {
-                                    return err
+                                    page_sizer.update(0, true);
```
rymnc:
```suggestion
                                    // we optimistically reduce the page size for any rpc error, to reduce the work to check if it is actually because of "too many logs"
                                    page_sizer.update(0, true);
```

crates/services/relayer/src/service/test.rs
```diff
@@ -208,3 +211,147 @@ async fn update_sync__changes_latest_eth_state(
...
+        .download_logs(&eth_state.needs_to_sync_eth().unwrap())
+        .await;
+
+    // Assert
+    assert!(result.is_ok());
+    assert_eq!(relayer.page_sizer.page_size(), 6); // 5 * 125 / 100 = 6.25 → 6
+}
+
+#[tokio::test]
+async fn relayer_respects_max_page_size_limit() {
```
rymnc, line 265: please use the given/when/then pattern here

PR #3105, opened 2025-09-30
Migrate relayer from ethers-rs to alloy.

### LINKED ISSUE #1332: TOB-FUEL-37: Ethabi dependency is no longer maintained
## Description
The ethabi dependency is [no longer maintained](https://github.com/rust-ethereum/ethabi/issues/301#issuecomment-1651389173). More details on 4. of September 2023.

Because the Fuel system does not have server components which parse user provided ABI specifications, this does not pose a security risk to Fuel.

## Recommendations

Short term, upgrade ethabi as soon as a fixed version is released.
Long term, consider switching to an alternative library like [ethers-rs](https://github.com/gakonst/ethers-rs). Note that the current version of ethers-rs is also vulnerable to the same bug as ethabi, so make sure to update the library if fixes are released.

### LINKED ISSUE #2427: RUSTSEC-2024-0384: `instant` is unmaintained
> `instant` is unmaintained

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Status              | unmaintained                |
| Package             | `instant`                      |
| Version             | `0.1.13`                   |
| URL                 | []() |
| Date                | 2024-09-01                         |

This crate is no longer maintained, and the author recommends using the maintained [`web-time`] crate instead.

[`web-time`]: https://crates.io/crates/web-time

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2024-0384.html) for additional details.

### LINKED ISSUE #2826: RUSTSEC-2025-0010: Versions of *ring* prior to 0.17 are unmaintained.
> Versions of *ring* prior to 0.17 are unmaintained.

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Status              | unmaintained                |
| Package             | `ring`                      |
| Version             | `0.16.20`                   |
| URL                 | [https://github.com/briansmith/ring/discussions/2450](https://github.com/briansmith/ring/discussions/2450) |
| Date                | 2025-03-05                         |

*ring* 0.16.20 was released over 4 years ago and isn&#39;t maintained, tested, etc.

Additionally, the project&#39;s general policy is to only patch the latest release,
which is 0.17.12 now. It will be difficult for anybody to backport future fixes
to versions earlier than 0.17.10 due to license changes.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2025-0010.html) for additional details.

### LINKED ISSUE #3077: chore(relayer): use alloy instead of ethers-rs
Closes https://github.com/FuelLabs/fuel-core/issues/1332
Closes https://github.com/FuelLabs/fuel-core/issues/2427
Closes https://github.com/FuelLabs/fuel-core/issues/2826

we make use of the `ethers` crate: https://github.com/FuelLabs/fuel-core/blob/68d263ce84cff5e5076251921b20e10a9b25cc2a/crates/services/relayer/Cargo.toml#L29-L36

which has been deprecated for a while now. we should switch to using `alloy`: https://crates.io/crates/alloy, which is more actively maintained. 

there are some missing features in `alloy` that we might have to implement ourselves, like the QuorumProvider from the `ethers` crate, but that can be addressed in a follow up PR.

the scope of this issue is just to remove the `ethers` dependency, switch it with `alloy` and retain the same api on fuel-core

its-saeed:
## Linked Issues/PRs
Closes #3077
Closes #1332
Closes #2427
Closes #2826 

## Description
Replace deprecated ethers dependencies in `fuel-core/crates/services/relayer` with the actively maintained alloy crate.

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/provider/src/test_helpers/provider.rs
+++ b/crates/provider/src/test_helpers/provider.rs
@@ -177,9 +177,7 @@ impl Provider for MockProvider {
         unreachable!()
     }
 
-    fn get_block_number(
-        &self,
-    ) -> ProviderCall<alloy_rpc_client::NoParams, U64, BlockNumber> {
+    fn get_block_number(&self) -> ProviderCall<NoParams, U64, BlockNumber> {
         let this = self;
         this.before_event(TriggerType::GetBlockNumber);
         let r = self.update_data(|data| data.best_block.number());
@@ -315,7 +313,7 @@ impl Default for MockProvider {
     }
 }
 
-impl fmt::Debug for MockProvider {
+impl Debug for MockProvider {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("MockProvider")
             .field("data", &self.data)
--- a/crates/services/relayer/Cargo.toml
+++ b/crates/services/relayer/Cargo.toml
@@ -11,29 +11,22 @@ rust-version = { workspace = true }
 description = "Fuel Relayer"
 
 [features]
-test-helpers = [
-    "parking_lot",
-    "fuel-core-types/test-helpers",
-    "alloy-consensus",
-    "alloy-rpc-client/reqwest",
-]
+test-helpers = ["fuel-core-provider/test-helpers", "fuel-core-types/test-helpers"]
 
 [dependencies]
-alloy-consensus = { workspace = true, optional = true }
 alloy-primitives = { workspace = true }
 alloy-provider = { workspace = true }
-alloy-rpc-client = { workspace = true, optional = true }
 alloy-rpc-types-eth = { workspace = true }
 alloy-sol-types = { workspace = true }
 anyhow = { workspace = true }
 async-trait = { workspace = true }
 enum-iterator = { workspace = true }
+fuel-core-provider = { workspace = true }
 fuel-core-services = { workspace = true }
 fuel-core-storage = { workspace = true, features = ["std"] }
 fuel-core-types = { workspace = true, features = ["std"] }
 futures = { workspace = true }
 once_cell = { workspace = true }
-parking_lot = { workspace = true, optional = true }
 strum = { workspace = true }
 strum_macros = { workspace = true }
 tokio = { workspace = true, features = ["macros"] }
@@ -41,6 +34,7 @@ tracing = { workspace = true }
 url = { workspace = true }
 
 [dev-dependencies]
+fuel-core-provider = { path = "../../provider" }
 fuel-core-relayer = { path = "", features = ["test-helpers"] }
 fuel-core-services = { path = "../../services", features = ["test-helpers"] }
 fuel-core-storage = { path = "../../storage", features = ["test-helpers"] }
--- a/crates/services/relayer/src/service.rs
+++ b/crates/services/relayer/src/service.rs
@@ -10,11 +10,7 @@ use crate::{
     ports::RelayerDb,
     service::state::EthLocal,
 };
-use alloy_provider::{
-    Provider,
-    RootProvider,
-    network::Ethereum,
-};
+use alloy_provider::Provider;
 use alloy_rpc_types_eth::{
     BlockId,
     BlockNumberOrTag,
@@ -24,6 +20,10 @@ use alloy_rpc_types_eth::{
 };
 use async_trait::async_trait;
 use core::time::Duration;
+use fuel_core_provider::{
+    Quorum,
+    QuorumProvider,
+};
 use fuel_core_services::{
     RunnableService,
     RunnableTask,
@@ -37,7 +37,6 @@ use fuel_core_types::{
 };
 use futures::StreamExt;
 use tokio::sync::watch;
-use url::Url;
 
 mod get_logs;
 mod run;
@@ -74,7 +73,7 @@ type Synced = watch::Receiver<SyncState>;
 type NotifySynced = watch::Sender<SyncState>;
 
 /// The alias of runnable relayer service.
-pub type Service<D> = CustomizableService<FuelEthProvider, D>;
+pub type Service<D> = CustomizableService<QuorumProvider, D>;
 type CustomizableService<P, D> = ServiceRunner<NotInitializedTask<P, D>>;
 
 /// The shared state of the relayer task.
@@ -82,24 +81,6 @@ type CustomizableService<P, D> = ServiceRunner<NotInitializedTask<P, D>>;
 pub struct SharedState {
     /// Receives signals when the relayer reaches consistency with the DA layer.
     synced: Synced,
-}
-
-pub struct FuelEthProvider {
-    provider: RootProvider<Ethereum>,
-}
-
-impl FuelEthProvider {
-    fn new(url: Url) -> Self {
-        Self {
-            provider: RootProvider::<Ethereum>::new_http(url),
-        }
-    }
-}
-
-impl Provider for FuelEthProvider {
-    fn root(&self) -> &RootProvider<Ethereum> {
-        &self.provider
-    }
 }
 
 /// Not initialized version of the [`Task`].
@@ -478,23 +459,13 @@ pub fn new_service<D>(database: D, config: Config) -> anyhow::Result<Service<D>>
 where
     D: RelayerDb + 'static,
 {
-    let url = config
-        .relayer
-        .as_ref()
-        .ok_or_else(|| {
-            anyhow::anyhow!(
-                "Tried to start Relayer without setting an eth_client in the config"
-            )
-        })?
-        .first()
-        .ok_or_else(|| {
-            anyhow::anyhow!(
-                "Tried to start Relayer without setting an eth_client in the config"
-            )
-        })?
-        .clone();
+    let urls = config.relayer.clone().ok_or_else(|| {
+        anyhow::anyhow!(
+            "Tried to start Relayer without setting an eth_client in the config"
+        )
+    })?;
 
-    let eth_node = FuelEthProvider::new(url);
+    let eth_node = QuorumProvider::new(Quorum::Majority, urls);
     let retry_on_error = true;
     Ok(new_service_internal(
         eth_node,
--- a/crates/services/relayer/src/service/get_logs.rs
+++ b/crates/services/relayer/src/service/get_logs.rs
@@ -1,8 +1,11 @@
 use super::*;
 use alloy_primitives::Address;
-use alloy_provider::transport::{
-    TransportError,
-    TransportErrorKind,
+use alloy_provider::{
+    Provider,
+    transport::{
+        TransportError,
+        TransportErrorKind,
+    },
 };
 use fuel_core_types::{
     entities::RelayedTransaction,
--- a/crates/services/relayer/src/service/get_logs/test.rs
+++ b/crates/services/relayer/src/service/get_logs/test.rs
@@ -6,19 +6,17 @@ use crate::{
         Transaction,
     },
     service::state::EthSyncGap,
-    test_helpers::{
-        page_sizer::IdentityPageSizer,
-        provider::{
-            MockProvider,
-            TriggerType,
-        },
-    },
+    test_helpers::page_sizer::IdentityPageSizer,
 };
 use alloy_primitives::{
     IntoLogData,
     U256,
 };
 use alloy_provider::transport::TransportError;
+use fuel_core_provider::test_helpers::provider::{
+    MockProvider,
+    TriggerType,
+};
 use std::{
     ops::RangeInclusive,
     sync::atomic::{
--- a/crates/services/relayer/src/service/test.rs
+++ b/crates/services/relayer/src/service/test.rs
@@ -5,8 +5,11 @@ use crate::{
     ports::RelayerDb,
     service::NotInitializedTask,
 };
+use alloy_primitives::LogData;
+use fuel_core_provider::test_helpers::provider::MockProvider;
 use fuel_core_services::RunnableService;
 use futures::TryStreamExt;
+use std::num::NonZeroU8;
 use test_case::test_case;
 
 use super::*;
@@ -18,14 +21,15 @@ use crate::{
         download_logs,
         state,
     },
-    test_helpers::{
-        page_sizer::IdentityPageSizer,
-        provider::MockProvider,
-    },
+    test_helpers::page_sizer::IdentityPageSizer,
 };
 use alloy_provider::{
     ProviderBuilder,
     mock::Asserter,
+    transport::{
+        TransportError,
+        TransportErrorKind,
+    },
 };
 use alloy_rpc_types_eth::Log;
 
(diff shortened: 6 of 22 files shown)


crates/services/relayer/src/test_helpers/provider.rs
```diff
@@ -0,0 +1,260 @@
...
+    // fn get_transaction_by_hash(&self, hash: TxHash) -> ProviderCall<(TxHash,), Option<<Ethereum as alloy_provider::Network>::TransactionResponse>> {
+    //     let txn: Transaction<TxEnvelope> = Transaction {
+    //         block_number: Some(self.update_data(|data| data.best_block.number())),
+    //         inner:
+    //         block_hash: Some(hash),
+    //         ..Default::default()
+    //     };
+    //     self.asserter.push_success(&txn);
+    //     self.inner.get_transaction_by_hash(hash)
+    // }
```
rymnc:
```suggestion
```

crates/services/relayer/src/service.rs
```diff
@@ -370,18 +367,23 @@ pub fn new_service<D>(database: D, config: Config) -> anyhow::Result<Service<D>>
...
             anyhow::anyhow!(
                 "Tried to start Relayer without setting an eth_client in the config"
             )
         })?
-        .into_iter()
-        .map(|url| WeightedProvider::new(Http::new(url)));
+        .clone();
 
-    let eth_node = Provider::new(QuorumProvider::new(Quorum::Majority, urls));
+    let eth_node = RootProvider::<Ethereum>::new_http(url);
```
rymnc: for now can you create a newtype around RootProvider called `FuelEthProvider` or similar, implement `Provider` for it i.e delegates all fn calls to the internal `RootProvider`? later when we implement `QuorumProvider` then we can just modify `FuelEthProvider` instead of changing `RootProvider` in other places? im ok with current impl, just a thought

crates/services/relayer/Cargo.toml
```diff
@@ -12,39 +12,28 @@ description = "Fuel Relayer"
...
-ethers-providers = { version = "2", default-features = false, features = [
-  "ws",
-  "rustls",
-] }
+alloy-primitives = { version = "1.4.0", default-features = false }
+alloy-rpc-types-eth = "1.0.36"
+alloy-sol-types = "1.4.0"
+alloy-rpc-client = "1.0.37"
+alloy-consensus = "1.0.37"
+alloy-provider = "1.0.37"
```
rymnc: do you want to disable default-features for any of these other crates?

crates/services/relayer/Cargo.toml
```diff
@@ -12,39 +12,28 @@ description = "Fuel Relayer"
...
 fuel-core-services = { workspace = true }
 fuel-core-storage = { workspace = true, features = ["std"] }
 fuel-core-types = { workspace = true, features = ["std"] }
 futures = { workspace = true }
 once_cell = { workspace = true }
-parking_lot = { workspace = true, optional = true }
-serde = { workspace = true, optional = true }
-serde_json = { workspace = true, optional = true }
+ parking_lot = { workspace = true, optional = true }
```
rymnc:
```suggestion
parking_lot = { workspace = true, optional = true }
```

crates/services/relayer/src/test_helpers/provider.rs
```diff
@@ -0,0 +1,260 @@
...
+            _logs_batch_index: 0,
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct MockProvider {
+    inner: Box<DynProvider>,
+    asserter: Asserter,
+    data: Arc<parking_lot::Mutex<InnerState>>,
```
rymnc:
```suggestion
    data: Arc<Mutex<InnerState>>,
```

rymnc: is there any particular reason why a parking_lot::Mutex is being used here?

crates/services/relayer/src/service.rs
its-saeed: Thanks! Good idea!

crates/services/relayer/src/service.rs
```diff
@@ -121,6 +135,7 @@ pub struct Task<P, D> {
 
 impl<P, D> NotInitializedTask<P, D>
 where
+    P: Provider + 'static,
     D: RelayerDb + 'static,
```
MitchTurner, line 202: It looks like `RelayerDb` is a proper port, in that we own the trait and impl some arbitrary type for it.

It looks like `Provider` is a foreign trait. Is it possible for us to invert this dependency? i.e. we should have our own `Provider` trait and at the bare minimum, we should be able to
```
impl<T: alloy_provider::Provider> FuelProvider for T {

}
```
This will also require either
1. we define our own block types
2. we expose assoc types on the port for e.g. block_id, etc. which can resolve at integration

its-saeed: Thanks for this great suggestion. Will address it in #3113

crates/services/relayer/src/test_helpers/provider.rs
its-saeed: The previous code has been ported to use alloy-rs. :)

crates/services/relayer/src/service/get_logs.rs
```diff
@@ -44,7 +45,7 @@ where
                             .from_block(page.oldest())
                             .to_block(page.latest())
                             .address(ValueOrArray::Array(contracts))
-                            .topic0(ValueOrArray::Array(vec![
+                            .event_signature(ValueOrArray::Array(vec![
```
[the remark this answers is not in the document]
its-saeed: `topic0` is deprecated:
```rust
    /// Sets topic0 (the event name for non-anonymous events)
    #[must_use]
    #[deprecated(note = "use `event_signature` instead")]
    pub fn topic0<T: Into<Topic>>(mut self, topic: T) -> Self {
        self.topics[0] = topic.into();
        self
    }
```

and `event_signature` has the same implementation:

```rust
    /// Sets event_signature(topic0) (the event name for non-anonymous events)
    #[must_use]
    pub fn event_signature<T: Into<Topic>>(mut self, topic: T) -> Self {
        self.topics[0] = topic.into();
        self
    }
```

.changes/changed/3105.md
```diff
@@ -0,0 +1 @@
+Migrated relayer service and related tests from ethers-rs to alloy-rs.
```
rymnc, line 1: you probably want to add to the breaking changes that we don't expect multiple rpc urls anymore and will just use the first one. while technically there are no changes on the cli side, it is a breaking change in expected behaviour so we have to clarify this. if we land the `QuorumProvider` before a release is created then the breaking change file in `.changes/` can be removed in that pr

crates/services/relayer/src/service/get_logs/test.rs
```diff
@@ -44,19 +45,28 @@ fn messages(
...
+            data: log_data.clone(),
+        },
+        block_hash: None,
+        block_number: Some(block_number),
+        block_timestamp: None,
+        transaction_hash: None,
+        transaction_index: None,
+        log_index: Some(log_index),
+        removed: false,
+    }
```
rymnc, line 67: i like that this is more explicit now. thank you

crates/services/relayer/src/service/get_logs/test.rs
```diff
@@ -73,25 +83,32 @@ fn transactions(
...
 
 fn transaction(nonce: u64, block_number: u64, contract_address: u32, index: u64) -> Log {
-    let transaction = TransactionFilter {
-        nonce: U256::from_dec_str(nonce.to_string().as_str())
-            .expect("Should convert into U256"),
-        ..Default::default()
+    let transaction = Transaction {
+        nonce: U256::from(nonce),
+        max_gas: Default::default(),
+        canonically_serialized_tx: Default::default(),
```
rymnc, line 87: `Transaction` doesn't implement Default?

crates/services/relayer/src/test_helpers/provider.rs
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    pub fn update_data<R>(&self, delta: impl FnOnce(&mut MockData) -> R) -> R {
+        self.data.lock().update(delta)
+    }
+
+    /// Set a callback before an event.
+    pub fn set_before_event(
+        &self,
+        f: impl for<'a> FnMut(&mut MockData, TriggerType<'a>) + Send + Sync + 'static,
```
rymnc: wondering if it will look cleaner with a generic, but seems ok for now

crates/provider/src/test_helpers/provider.rs
```diff
@@ -0,0 +1,347 @@
...
+        &self,
+        f: impl FnMut(&mut MockData) + Send + Sync + 'static,
+    ) {
+        self.data.lock().override_fn = Some(Box::new(f));
+    }
+}
+
+async fn simulate_network_latency() {
+    tokio::task::yield_now().await;
+}
```
rymnc, line 172: does this actually "simulate" network latency?

crates/services/relayer/src/test_helpers/provider.rs
```diff
@@ -0,0 +1,347 @@
...
+            _logs_batch_index: 0,
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct MockProvider {
+    inner: Box<DynProvider>,
+    asserter: Asserter,
+    data: Arc<parking_lot::Mutex<InnerState>>,
```
rymnc:
```suggestion
    data: Arc<Mutex<InnerState>>,
```

parking_lot::Mutex already imported above

crates/services/relayer/src/log.rs
```diff
@@ -100,23 +97,17 @@ impl TryFrom<&Log> for EthEventLog {
 
 fn parse_message_to_event(log: &Log) -> anyhow::Result<EthEventLog> {
     // event has 3 indexed fields, so it should have 4 topics
-    if log.topics.len() != 4 {
+    if log.topics().len() != 4 {
         return Err(anyhow!("Malformed topics for Message"))
```
rymnc: lets include the expected and found topic length

```suggestion
        return Err(anyhow!("Malformed topics for Message. expected: 4, found: {}", log.topics().len()))
```

rymnc: even better if the magic number `4` is colocated with the sol! macro call

crates/services/relayer/src/log.rs
```diff
@@ -129,42 +120,34 @@ fn parse_message_to_event(log: &Log) -> anyhow::Result<EthEventLog> {
...
-                .as_u64(),
+                .ok_or(anyhow!("Log missing block height"))?,
         ),
     }))
 }
 
 fn parse_forced_tx_to_event(log: &Log) -> anyhow::Result<EthEventLog> {
     // event has one indexed field, so there are 2 topics
-    if log.topics.len() != 2 {
+    if log.topics().len() != 2 {
```
rymnc: same as https://github.com/FuelLabs/fuel-core/pull/3105/files#r2425630780

crates/services/relayer/src/log.rs
```diff
@@ -129,42 +120,34 @@ fn parse_message_to_event(log: &Log) -> anyhow::Result<EthEventLog> {
...
-        serialized_transaction: serialized_transaction.to_vec(),
+        serialized_transaction,
         // Safety: logs without block numbers are rejected by
         // FinalizationQueue::append_eth_log before the conversion to EthEventLog happens.
         // If block_number is none, that means the log is pending.
         da_height: DaBlockHeight::from(
             log.block_number
-                .ok_or(anyhow!("Log missing block height"))?
-                .as_u64(),
+                .ok_or(anyhow!("Log missing block height"))?,
```
rymnc: i think we should do this before line 134, because we waste cycles on running `decode_log` if the block_number is empty anyway we return an error

crates/services/relayer/Cargo.toml
```diff
@@ -12,39 +12,30 @@ description = "Fuel Relayer"
...
     "fuel-core-types/test-helpers",
+    "alloy-consensus",
+    "alloy-rpc-client/reqwest",
 ]
 
 [dependencies]
+alloy-consensus = { version = "1.0.37", default-features = false, optional = true }
+alloy-primitives = { version = "1.4.0", default-features = false }
+alloy-provider = { version = "1.0.37", default-features = false, features = ["reqwest", "reqwest-rustls-tls"] }
+alloy-rpc-client = { version = "1.0.37", default-features = false, features = ["reqwest"], optional = true }
```
rymnc: thanks for this. i think moving forward whenever we add deps we should do it this way. turn off default features and only use what we want. this is how its done in the block-committer

crates/services/relayer/Cargo.toml
```diff
@@ -12,39 +12,30 @@ description = "Fuel Relayer"
...
+alloy-rpc-types-eth = "1.0.36"
+alloy-sol-types = "1.4.0"
```
rymnc: https://github.com/FuelLabs/fuel-core/pull/3105/files#r2425641204 ref this pls

crates/services/relayer/src/service/test.rs
```diff
@@ -97,67 +67,115 @@ async fn quorum_agrees_on_logs() {
...
-    match provider_error {
-        Err(ProviderError::CustomError(e)) => {
-            assert!(e.contains("eth provider failed to get logs: NoQuorumReached"));
-        }
-        _ => {
-            panic!("Expected a JsonRpcClientError")
-        }
-    }
-}
+// FIXME: Enable this test when we have a quorum provider
```
MitchTurner: ARe these attached to specific issues? In general, our rule on TODOs in code is they need to have a github issue attached to them.

crates/services/relayer/src/test_helpers/provider.rs
MitchTurner: Yeah. I think a generic would be cleaner. But just a nit.

crates/provider/src/test_helpers/provider.rs
its-saeed: Hmm, having a delay would cause some tests to fail. I added this function just to make it clear why we did `yield_now`.

crates/services/relayer/src/service/get_logs/test.rs
its-saeed: no

rymnc (dismissed):

crates/services/relayer/src/service.rs
```diff
@@ -253,7 +268,11 @@ where
     ) -> anyhow::Result<()> {
         let logs = download_logs(
             eth_sync_gap,
-            self.config.eth_v2_listening_contracts.clone(),
+            self.config
+                .eth_v2_listening_contracts
+                .iter()
+                .map(|addr| alloy_primitives::Address::from_slice(addr.as_slice()))
+                .collect(),
```
xgreenx: It would be nice to avoid additional iteration on each download=) We can calculate it once during creation and reuse it. I know that difference in performance almost zero, just a better practice

PR #3106, opened 2025-09-30
Add remote block cache

MitchTurner:
## Checklist

### Before requesting review

- [ ] Someone else?


crates/services/block_aggregator_api/src/db/remote_cache.rs
```diff
@@ -0,0 +1,231 @@
...
+            client,
+            local_persisted,
+            sync_from,
+            highest_new_height: None,
+            orphaned_new_height: None,
+            synced: false,
+        }
+    }
+
+    fn url_for_block(base: &str, key: &str) -> String {
```
MitchTurner: I don't know if this is good enough. What will the URL look like and can we simply append the key to the end? 

(also might want to use `url` lib to make it less fragile)

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -461,8 +468,57 @@ pub fn init_sub_services(
...
+        let sync_from = block_aggregator_config.sync_from.unwrap_or_default();
+        let sync_from_height;
+        let db_adapter = if let Some((
+            aws_access_key_id,
+            aws_secrete_access_key,
+            aws_region,
+            aws_bucket,
+            url_base,
+            aws_endpoint_url,
+        )) = get_env_vars()
```
mchristopher: The db adapter needs to be explicit from the command line (or an explicit env var) instead of set this way. In production, most of those values will not get passed in (for example, we do not have access to `AWS_ACCESS_KEY_ID` or `AWS_SECRET_ACCESS_KEY` in prod).

I would also expect this component to fail if I asked for S3 as the backend and the configuration was incorrect, so basing the setup on random env vars isn't a good precedent.

crates/services/block_aggregator_api/src/db/storage_or_remote_db.rs
```diff
@@ -0,0 +1,168 @@
...
+        aws_endpoint_url: Option<String>,
+        sync_from: BlockHeight,
+    ) -> Self {
+        let client = {
+            let mut builder = aws_sdk_s3::config::Builder::new();
+            if let Some(aws_endpoint_url) = aws_endpoint_url {
+                builder.set_endpoint_url(Some(aws_endpoint_url.clone()));
+            }
+
+            let config = builder
```
mchristopher: This needs to use the DefaultCredentialsChain instead of manually specifying a Credential provider. That ensures that we're staying in-line with how the AWS SDK loads credentials, and that the adapter will work in a test env as well as in production.

See https://github.com/FuelLabs/sway/blob/fb10b2fcb185f1a14e79efdade9faf007f40d70d/forc-plugins/forc-client/src/util/aws.rs#L27-L28 for an example (or elsewhere in the Fuel org).

crates/services/block_aggregator_api/src/db/storage_or_remote_db.rs
```diff
@@ -0,0 +1,168 @@
...
+                .force_path_style(true)
+                .region(Region::new(Cow::Owned(aws_region.to_string())))
```
mchristopher: You should not need to manually specify a region here, I believe that can be auto-discovered using the default chain.

Instead of manually putting a region here, I think you can add `.cross_region_access_enabled(true)` which will automatically handle region changes.

crates/services/block_aggregator_api/proto/api.proto
```diff
@@ -5,14 +5,21 @@ package blockaggregator;
...
-  uint32 height = 1;
+  optional uint32 height = 1;
 }
 
 message BlockRangeRequest {
   uint32 start = 1;
   uint32 end = 2;
 }
 
+message RemoteBlockRangeResponse {
```
mchristopher: This should be `RemoteBlockResponse` and not `RemoteBlockRangeResponse`, right?

crates/services/block_aggregator_api/proto/api.proto
```diff
@@ -5,14 +5,21 @@ package blockaggregator;
...
+  string region = 1;
+  string bucket = 2;
+  string key = 3;
+  string url = 4;
```
mchristopher: What is the goal of having `url` here? The `url` value can be derived from the `region`, `bucket` and `key` values.

Also, I think we should add another layer to this, since `RemoteBlockRangeResponse` should be built to support multiple remote protocols from the first release. With this version, we've combined HTTP (with the `url` field) and S3 (with the `region`, `bucket` and `key` fields) into a single response type, which seems bad.

I'd rather separate this out into a `RemoteS3BlockResponse` and a `RemoteHTTPBlockResponse` message, with a `oneof` attribute. That ensures we can cleanly support multiple protocols in the future, and keeps implementation details specific to each message type.

crates/services/block_aggregator_api/src/db/remote_cache.rs
mchristopher: I'm not sure why we'd need this for S3. It makes sense for pure HTTP, but not for S3.

crates/fuel-core/src/service/sub_services.rs
MitchTurner: You're right. It should be more explicit, and I acknowledge that in the Description above. 

Isn't it standard to set secrets using env-vars though. So maybe at least that should be left as an env-var?

mchristopher: It is standard, but we don't actually use env vars for IAM credentials in production. With Kubernetes, there is a special file that is mounted inside the Docker container and a separate environment variable is set pointing to the path of that file.

That process is entirely automated and handled by the AWS SDK, as long as we're using the DefaultCredentialChain when initializing the AWS SDK. That is the reason for my comment - it's not valid to assume a certain set of AWS-specific environment variables will give you the correct data needed to determine if the config is correct. The only way to actually know if the config is right is to use the AWS SDKs to verify it.

MitchTurner: Hmm. Sounds slick... but also I don't know how that will interface with `localstack`. Might need a way to override the credentials.

crates/services/block_aggregator_api/proto/api.proto
```diff
@@ -5,14 +5,21 @@ package blockaggregator;
 message BlockHeightRequest {}
 
 message BlockHeightResponse {
-  uint32 height = 1;
+  optional uint32 height = 1;
```
mchristopher: Just realized - all the block height values should be uint64 to match the graphQL schemas.

MitchTurner: Block heights are ` u32` under the hood:
```rust
key!(BlockHeight, u32);
```
They can always be up-cast.

crates/fuel-core/src/service/sub_services.rs
mchristopher: The only thing you should need to do to support localstack is to allow some way to override the default endpoint used by the SDKs. The remainder of the env vars will be picked up and handled automatically by the AWS SDK.

See https://github.com/FuelLabs/fuel-block-committer/blob/44e35c25123899dcbcb6cd3a8582db4c675029da/packages/adapters/signers/src/kms_utils.rs#L8-L10 for how the block committer handles this.

For the sake of generic S3 support, we should support custom endpoints as a first-class option (instead of just in a test build), since things like Google Cloud Storage or Cloudflare R2 provide S3-compatible APIs which are easy for us to support if we allow users to override the endpoint.

crates/fuel-core/src/combined_database.rs
```diff
@@ -71,15 +75,17 @@ impl CombinedDatabase {
         relayer: Database<Relayer>,
         gas_price: Database<GasPriceDatabase>,
         compression: Database<CompressionDatabase>,
-        block_aggregation: Database<BlockAggregatorDatabase>,
+        block_aggregation_storage: Database<BlockAggregatorDatabaseStorage>,
+        block_aggregation_s3: Database<BlockAggregatorDatabaseS3>,
```
xgreenx: It should be enough for us to have just one database instance per service. In the case if we switched from one type of storing to another one, we can throw and error saying that the user should clean-up the old database.

crates/fuel-core/src/combined_database.rs
```diff
@@ -61,7 +75,10 @@ pub struct CombinedDatabase {
     relayer: Database<Relayer>,
     gas_price: Database<GasPriceDatabase>,
     compression: Database<CompressionDatabase>,
-    block_aggregation: Database<BlockAggregatorDatabase>,
+    #[cfg(feature = "rpc")]
+    block_aggregation_storage: Database<BlockAggregatorDatabaseStorage>,
+    #[cfg(feature = "rpc")]
+    block_aggregation_s3: Database<BlockAggregatorDatabaseS3>,
```
xgreenx: We should ahve only one instance of database per service. If we change the type of storage after a restart, it is fine to throw an error that the storage must be cleaned.

crates/fuel-core/src/database.rs
```diff
@@ -442,12 +453,48 @@ impl Modifiable for Database<GasPriceDatabase> {
...
+impl Database<BlockAggregatorDatabaseS3> {
+    pub fn rollback_to(&mut self, block_height: BlockHeight) -> StorageResult<()> {
+        let mut tx = self.write_transaction();
+        tx.storage_as_mut::<LatestBlock>()
+            .insert(&(), &block_height)
+            .map_err(|e: StorageError| anyhow!(e))?;
+        tx.commit().map_err(|e: StorageError| anyhow!(e))?;
+        Ok(())
+    }
+}
```
xgreenx: It still would be nice to clean up all remaining data, just to follow the idea that we have the exact state as before the rollback.

crates/fuel-core/src/service/config.rs
```diff
@@ -172,6 +177,8 @@ impl Config {
         #[cfg(feature = "rpc")]
         let rpc_config = fuel_core_block_aggregator_api::integration::Config {
             addr: free_local_addr(),
+            sync_from: Some(BlockHeight::from(0)),
+            storage_method: StorageMethod::Local,
         };
```
xgreenx: By default, we want to not use RPC(so it should be `None`). Only some specific sentries will be responsible for posting data to S3

crates/services/block_aggregator_api/src/lib.rs
```diff
@@ -56,43 +63,63 @@ pub mod integration {
...
+    #[derive(Clone, Debug, Default)]
+    pub enum StorageMethod {
+        #[default]
+        Local,
+        S3 {
+            bucket: String,
+            endpoint_url: Option<String>,
+            requester_pays: bool,
+        },
     }
```
xgreenx: So in the mainnet we want to have al lsentries use S3 to return data back to the user, but only on node in the network should post data to S3 bucket. So it would be nice to have something like `S3Read` `S3Write` here.

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -461,8 +472,44 @@ pub fn init_sub_services(
     #[cfg(feature = "rpc")]
     let block_aggregator_rpc = {
```
xgreenx: Can we move this initialization somewhere please?=D `init_sub_services` already si huge, if we put logic here it will be impossible to read in the future. Just some wrapper or function that do all of that inside and return final values

crates/services/block_aggregator_api/build.rs
```diff
@@ -1,7 +0,0 @@
-fn main() -> Result<(), Box<dyn std::error::Error>> {
```
xgreenx, line 1: Can you try to remove `install protobuf deps`

.github/workflows/ci.yml
```diff
@@ -194,12 +192,47 @@ jobs:
...
+      RUSTFLAGS: -D warnings
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.RUST_VERSION }}
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v3
```
xgreenx: Maybe we can remove it now

crates/services/block_aggregator_api/src/api/protobuf_adapter.rs
```diff
@@ -47,34 +62,36 @@ impl Server {
...
+                height: height.map(|inner| *inner),
             })),
             Err(e) => Err(tonic::Status::internal(format!(
                 "Failed to receive height: {}",
                 e
             ))),
         }
     }
-    type GetBlockRangeStream = ReceiverStream<Result<ProtoBlockResponse, Status>>;
+    // type GetBlockRangeStream = ReceiverStream<Result<ProtoBlockResponse, Status>>;
```
xgreenx:
```suggestion
```

crates/services/block_aggregator_api/src/api/protobuf_adapter.rs
```diff
@@ -152,42 +185,74 @@ impl BlockAggregator for Server {
...
+    ) -> anyhow::Result<Self::Task> {
+        Ok(self)
+    }
+}
+
+impl RunnableTask for ServerTask {
+    async fn run(&mut self, watcher: &mut StateWatcher) -> TaskNextAction {
+        let server = Server::new(self.query_sender.clone());
+        let router = tonic::transport::Server::builder()
+            .add_service(ProtoBlockAggregatorServer::new(server));
```
xgreenx: It would be nice to do that inside of `into_task`=)

crates/services/block_aggregator_api/src/db/remote_cache.rs
```diff
@@ -0,0 +1,269 @@
...
+        let body = ByteStream::from(zipped);
+        let req = self
+            .client()
+            .await?
+            .put_object()
+            .bucket(&self.aws_bucket)
+            .key(&key)
+            .body(body)
+            .content_encoding("gzip")
+            .content_type("application/octet-stream");
```
mchristopher: This should probably be `application/grpc-web` instead of generic octet-stream.

mchristopher: Or possibly `application/protobuf`

crates/fuel-core/src/service.rs
```diff
@@ -550,7 +551,7 @@ mod tests {
             service.start_and_await().await.unwrap();
             sleep(Duration::from_secs(1));
             for service in service.sub_services() {
-                assert_eq!(service.state(), State::Started);
+                assert_eq!(service.state(), State::Started,);
```
xgreenx:
```suggestion
                assert_eq!(service.state(), State::Started);
```

crates/fuel-core/src/service/sub_services.rs
```diff
@@ -553,3 +574,103 @@ pub fn init_sub_services(
...
+        BlockAggregator<
+            ProtobufAPI,
+            StorageOrRemoteDB<Database<BlockAggregatorDatabase>>,
+            ImporterAndDbSource<SerializerAdapter, Database<OnChain>, ReceiptSource>,
+            ProtoBlock,
+        >,
+    >,
+> {
+    let block_aggregator_config = config.rpc_config.clone();
+    let sync_from = block_aggregator_config.sync_from.unwrap_or_default();
```
xgreenx: It would be nice to use `genesis_block.header().height()`. Assuming that gensis block can start from height `X`.

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/sync_service.rs
```diff
@@ -100,51 +120,55 @@ where
...
-            && self.next_height >= last_height
-        {
+        if self.next_height >= self.stop_height {
             tracing::info!(
-                "reached end height {}, putting task into hibernation",
-                last_height
+                "reached stop height {}, putting task into hibernation",
+                self.stop_height
             );
             futures::future::pending().await
```
xgreenx: I think we should wait here for stop signal isntead `futures::future::pending().await`

crates/services/block_aggregator_api/src/blocks/importer_and_db_source.rs
```diff
@@ -37,79 +42,80 @@ pub mod serializer_adapter;
...
             importer_task: importer_runner,
             sync_task: sync_runner,
             receiver,
-            _error_marker: std::marker::PhantomData,
         }
     }
 }
 
-impl<Serializer, DB, E> BlockSource for ImporterAndDbSource<Serializer, DB, E>
+impl<Serializer, DB, Receipts> BlockSource
```
xgreenx: We want to add something like:

```rust
        self.sync_task.stop_and_await().await?;
        self.importer_task.stop_and_await().await?;
```

To the `drain` function to shutdown services.

crates/services/block_aggregator_api/src/blocks/importer_and_db_source.rs
```diff
@@ -37,79 +42,80 @@ pub mod serializer_adapter;
...
     Serializer: BlockSerializer + Send + Sync + 'static,
     <Serializer as BlockSerializer>::Block: Send + Sync + 'static,
-    DB: Send + Sync,
-    DB: StorageInspect<FuelBlocks, Error = E>,
-    DB: StorageInspect<Transactions, Error = E>,
-    E: std::fmt::Debug + Send + Sync,
+    DB: Send + Sync + 'static,
+    DB: StorageInspect<FuelBlocks, Error = StorageError>,
+    DB: StorageInspect<Transactions, Error = StorageError>,
+    Receipts: TxReceipts,
```
xgreenx: Looks like we also want to see that our child services are still active(`sync_taask.while_started()`). Because if something goes wrong inside of them, they will stop, and parent tasks will not notice it.

crates/services/block_aggregator_api/src/db/remote_cache.rs
```diff
@@ -0,0 +1,269 @@
...
+                    tracing::info!("Updating latest block to {:?}", new_height);
+                    self.synced = true;
+                    self.highest_new_height = Some(new_height);
+                    self.orphaned_new_height = None;
+                    let mut tx = self.local_persisted.write_transaction();
+                    tx.storage_as_mut::<LatestBlock>()
+                        .insert(&(), &new_height)
+                        .map_err(|e| Error::DB(anyhow!(e)))?;
+                    tx.commit().map_err(|e| Error::DB(anyhow!(e)))?;
+                } else if self.orphaned_new_height.is_none() {
```
xgreenx:
```suggestion
                } else if new_height == self.sync_from || self.height_is_next_height(new_height)? {
                    tracing::info!("Updating latest block to {:?}", new_height);
                    self.synced = true;
                    self.highest_new_height = Some(new_height);
                    self.orphaned_new_height = None;
                    let mut tx = self.local_persisted.write_transaction();
                    tx.storage_as_mut::<LatestBlock>()
                        .insert(&(), &new_height)
                        .map_err(|e| Error::DB(anyhow!(e)))?;
                    tx.commit().map_err(|e| Error::DB(anyhow!(e)))?;
                } else if self.orphaned_new_height.is_none() {
```

crates/services/block_aggregator_api/src/db/remote_cache.rs
```diff
@@ -0,0 +1,269 @@
...
+                    tracing::info!("Marking block as orphaned: {:?}", new_height);
+                    self.orphaned_new_height = Some(new_height);
+                }
+            }
+            BlockSourceEvent::OldBlock(height, _) => {
+                tracing::debug!("Old block: {:?}", height);
+                tracing::info!("Old block: {:?}", height);
```
xgreenx:
```suggestion
```

crates/services/block_aggregator_api/src/block_range_response.rs
```diff
@@ -1,24 +1,31 @@
...
 pub type BoxStream<T> = core::pin::Pin<Box<dyn Stream<Item = T> + Send + 'static>>;
 
 /// The response to a block range query, either as a literal stream of blocks or as a remote URL
 pub enum BlockRangeResponse {
     /// A literal stream of blocks
-    Literal(BoxStream<ProtoBlock>),
+    Literal(BoxStream<(BlockHeight, ProtoBlock)>),
     /// A remote URL where the blocks can be fetched
-    Remote(String),
+    S3(BoxStream<(BlockHeight, RemoteS3Response)>),
```
xgreenx, line 15: I think in the case of S3 we can use `Vec<>`, because we know al lvalues in advacne and they are small and fast to create

crates/services/block_aggregator_api/src/db/remote_cache.rs
```diff
@@ -0,0 +1,269 @@
...
+            local_persisted,
+            sync_from,
+            highest_new_height: None,
+            orphaned_new_height: None,
+            synced: false,
+        }
+    }
+
+    async fn client(&mut self) -> crate::result::Result<&Client> {
+        self.init_client().await;
```
xgreenx: It would be nice to move this to `into_task` function somehow

crates/services/block_aggregator_api/src/db/storage_db.rs
```diff
@@ -110,15 +81,61 @@ where
...
+                    tx.commit().map_err(|e| Error::DB(anyhow!(e)))?;
+                } else if self.height_is_next_height(new_height)? {
+                    let mut tx = self.storage.write_transaction();
+                    self.synced = true;
+                    self.highest_new_height = Some(new_height);
+                    tx.storage_as_mut::<LatestBlock>()
+                        .insert(&(), &new_height)
+                        .map_err(|e| Error::DB(anyhow!(e)))?;
+                    tx.commit().map_err(|e| Error::DB(anyhow!(e)))?;
+                } else if self.orphaned_new_height.is_none() {
```
xgreenx:
```suggestion
                } else if new_height == self.sync_from || self.height_is_next_height(new_height)? {
                    let mut tx = self.storage.write_transaction();
                    self.synced = true;
                    self.highest_new_height = Some(new_height);
                    tx.storage_as_mut::<LatestBlock>()
                        .insert(&(), &new_height)
                        .map_err(|e| Error::DB(anyhow!(e)))?;
                    tx.commit().map_err(|e| Error::DB(anyhow!(e)))?;
                } else if self.orphaned_new_height.is_none() {
```

crates/services/block_aggregator_api/src/db/table.rs
```diff
@@ -0,0 +1,146 @@
...
+    Eq,
+    enum_iterator::Sequence,
+    Hash,
+    num_enum::TryFromPrimitive,
+)]
+pub enum Column {
+    Metadata = 0,
+    Blocks = 1,
+    LatestBlock = 2,
+    Mode = 3,
```
xgreenx: I think you can include `Mode` part into `LatestBlock`. In this case we don't need to have many small tables. Something like:

```rust
pub enum StorageHeight {
    Local(BlockHeight),
    S3(BlockHeight),
}
```

tests/tests/rpc.rs
```diff
@@ -1,27 +1,67 @@
...
+    Some((aws_id, aws_secret, aws_region))
+}
 
 #[tokio::test(flavor = "multi_thread")]
-async fn get_block_range__can_get_serialized_block_from_rpc() {
+async fn get_block_range__can_get_serialized_block_from_rpc__literal() {
+    if env_vars_are_set() {
+        tracing::info!("Skipping test: AWS credentials are set");
+        return;
+    }
```
xgreenx: Let's do the same what we did for `aws-kms` tests:

```
    // This test is only enabled if the environment variable is set
    let Some(kms_arn) = option_env!("FUEL_CORE_TEST_AWS_KMS_ARN") else {
        return;
    };
```

But with new env variable like `FUEL_CORE_TEST_S3_ARN`

tests/tests/rpc.rs
```diff
@@ -1,27 +1,67 @@
...
+    let aws_secret = std::env::var("AWS_SECRET_ACCESS_KEY").ok()?;
+    let aws_region = std::env::var("AWS_REGION").ok()?;
+    Some((aws_id, aws_secret, aws_region))
+}
 
 #[tokio::test(flavor = "multi_thread")]
-async fn get_block_range__can_get_serialized_block_from_rpc() {
+async fn get_block_range__can_get_serialized_block_from_rpc__literal() {
+    if env_vars_are_set() {
+        tracing::info!("Skipping test: AWS credentials are set");
```
xgreenx: We shouldn't skip the test in the case of `literal`

tests/tests/rpc.rs
```diff
@@ -59,23 +173,38 @@ async fn get_block_range__can_get_serialized_block_from_rpc() {
...
-    };
     // then
-    assert_eq!(expected_header.height.0, actual_height);
+    let key = block_height_to_key(&expected_height);
+    let expected = ProtoRemoteBlockResponse {
+        location: Some(Location::S3(RemoteS3Bucket {
+            bucket: "test-bucket".to_string(),
+            key,
+            requester_pays: false,
+            endpoint: Some("http://127.0.0.1:4566".to_string()),
```
xgreenx: You can put string into a constant and reuse in all tests(plus add a comment that it is Localstack address)

tests/tests/rpc.rs
```diff
@@ -1,27 +1,67 @@
 #![allow(non_snake_case)]
```
xgreenx, line 1: Since we require single-threaded execution for these tests, let's have something similar to `#[cfg(not(feature = "only-p2p"))]` but for S3, where we only run single-threaded tests for S3 without any other tests, to not increase CI time.

crates/services/block_aggregator_api/src/block_range_response.rs
MitchTurner: I'm not against this. WE need to update the protobuf types to return a vec instead of a stream. So I'm going to wait on this.

crates/services/block_aggregator_api/src/db/remote_cache.rs
MitchTurner: Added `UninitializedTask` for this.

crates/fuel-core/src/combined_database.rs
```diff
@@ -452,6 +496,25 @@ impl CombinedDatabase {
...
+                    block_aggregation_storage_height,
+                    target_block_height,
+                );
+
+                if !block_aggregation_storage_rolled_back {
+                    self.block_aggregation_storage_mut()
+                        .rollback_to(target_block_height)?;
+                }
+            }
+
```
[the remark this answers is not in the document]
xgreenx: Agree with cursor!

crates/fuel-core/src/database.rs
```diff
@@ -442,12 +457,46 @@ impl Modifiable for Database<GasPriceDatabase> {
...
+        let new = match mode {
+            None => None,
+            Some(Mode::Local(_)) => Some(Mode::new_local(block_height)),
+            Some(Mode::S3(_)) => Some(Mode::new_s3(block_height)),
+        };
+        if let Some(Mode::Local(_)) = mode {
+            let remove_heights = tx
+                .iter_all_keys::<Blocks>(Some(IterDirection::Reverse))
+                .flatten()
+                .take_while(|height| height <= &block_height)
```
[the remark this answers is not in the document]
xgreenx: And cursor is right again!

crates/services/block_aggregator_api/src/api/protobuf_adapter.rs
```diff
@@ -152,41 +188,97 @@ impl BlockAggregator for Server {
...
+    async fn shutdown(self) -> anyhow::Result<()> {
+        Ok(())
+    }
+}
+
+impl ProtobufAPI {
+    pub fn new(url: String) -> Result<Self> {
+        let (query_sender, query_receiver) = tokio::sync::mpsc::channel::<
+            BlockAggregatorQuery<BlockRangeResponse, ProtoBlock>,
+        >(100);
```
xgreenx: Looks like `100` should be customizable via CLI, becuase it defiend how many concurrent user we can support via RPC

crates/services/block_aggregator_api/src/api/protobuf_adapter.rs
```diff
@@ -47,34 +66,35 @@ impl Server {
...
+    type GetBlockRangeStream = BoxStream<Result<ProtoBlockResponse, Status>>;
 
     async fn get_block_range(
         &self,
         request: tonic::Request<ProtoBlockRangeRequest>,
     ) -> Result<tonic::Response<Self::GetBlockRangeStream>, tonic::Status> {
-        const ARB_LITERAL_BLOCK_BUFFER_SIZE: usize = 100;
         let req = request.into_inner();
         let (response, receiver) = tokio::sync::oneshot::channel();
         let query = BlockAggregatorQuery::GetBlockRange {
```
xgreenx: Looks like each query will go to main `run` loop, so all read queries will become sequential, instead of parallel. Looks like you could pass database/s3 configuraiton here, and directly fetch the data fro mthe soruce, isntead of sending a query to the service, which will fetch it from the source for you.

crates/services/block_aggregator_api/src/api/protobuf_adapter.rs
```diff
@@ -137,8 +172,9 @@ impl BlockAggregator for Server {
 
         let (task_sender, task_receiver) = tokio::sync::mpsc::channel(ARB_CHANNEL_SIZE);
         tokio::spawn(async move {
```
xgreenx: We can remove tokio spwan now, I guess =D

crates/services/block_aggregator_api/src/api/protobuf_adapter.rs
```diff
@@ -152,41 +188,97 @@ impl BlockAggregator for Server {
...
+        let router_res = self.get_router();
+        let router = try_or_stop!(router_res, |e| tracing::error!(
+            "Failed to get router, has not been started: {:?}",
+            e
+        ));
+        tokio::select! {
+                res = router.serve(self.addr) => {
                     if let Err(e) = res {
                         tracing::error!("BlockAggregator tonic server error: {}", e);
+                        TaskNextAction::ErrorContinue(anyhow!(e))
```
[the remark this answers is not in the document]
xgreenx: Agree with cursor!=D

crates/services/block_aggregator_api/src/db/remote_cache.rs
```diff
@@ -0,0 +1,238 @@
...
+        }
+        Ok(())
+    }
+
+    async fn get_block_range(
+        &self,
+        first: BlockHeight,
+        last: BlockHeight,
+    ) -> crate::result::Result<Self::BlockRangeResponse> {
+        // TODO: Check if it exists
```
xgreenx: Let's return an error in case we are not synced up to the requested height. But we don't need to check existance on S3, it will be too slow

MitchTurner: I've implemented this. It revealed that if we error, we aren't sending an error to the user, just handling the error in the service. We need to do both. I'll work on adding this and a test showing that it gets included.

PR #3116, opened 2025-10-13
Add better proto conversion coverage

MitchTurner:
## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/services/block_aggregator_api/src/blocks/importer_and_db_source/serializer_adapter.rs
+++ b/crates/services/block_aggregator_api/src/blocks/importer_and_db_source/serializer_adapter.rs
@@ -1489,6 +1489,8 @@ pub fn proto_header_to_empty_consensus_header(
     }
 }
 
+// TODO: Add coverage for V2 Block stuff
+//   https://github.com/FuelLabs/fuel-core/issues/3139
 #[cfg(not(feature = "fault-proving"))]
 #[allow(non_snake_case)]
 #[cfg(test)]


crates/services/block_aggregator_api/proto/api.proto
```diff
@@ -98,7 +98,7 @@ message V1Header {
   bytes event_inbox_root = 8;
   bytes prev_root = 9;
   uint32 height = 10;
-  bytes time = 11;
+  uint64 time = 11;
```
rymnc, line 101: tai64 fits in this? interesting

crates/types/src/test_helpers.rs
```diff
@@ -36,3 +71,603 @@ pub fn create_contract<R: Rng>(
...
+prop_compose! {
+    fn arb_msg_id()(inner in any::<[u8; 32]>()) -> MessageId {
+        MessageId::new(inner)
+    }
+}
+
+#[allow(unused)]
+fn arb_inputs() -> impl Strategy<Value = Vec<Input>> {
+    prop::collection::vec(arb_input_any(), 0..10)
+}
```
rymnc, line 238: ?

crates/services/block_aggregator_api/proto/api.proto
Dentosal: The name kind of implies it

crates/services/block_aggregator_api/proto/api.proto
```diff
@@ -195,6 +185,59 @@ message ScriptTransaction {
...
+  bytes salt = 2;
+  repeated StorageSlot storage_slots = 3;
+  Policies policies = 4;
+  repeated Input inputs = 5;
+  repeated Output outputs = 6;
+  repeated bytes witnesses = 7;
+  CreateMetadata metadata = 8;
+}
+
+message MintTransaction {
```
Dentosal, line 199: Doens't mint have inputs and outputs too?

crates/services/block_aggregator_api/proto/api.proto
```diff
@@ -195,6 +185,59 @@ message ScriptTransaction {
...
 }
 
+message CreateTransaction {
+  uint32 bytecode_witness_index = 1;
+  bytes salt = 2;
+  repeated StorageSlot storage_slots = 3;
+  Policies policies = 4;
+  repeated Input inputs = 5;
+  repeated Output outputs = 6;
+  repeated bytes witnesses = 7;
```
Dentosal, line 195: I think all txs have these, maybe we could move them to the top-level `Transaction`?

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/serializer_adapter.rs
```diff
@@ -3,43 +3,119 @@ use crate::protobuf_types::V2Header as ProtoV2Header;
...
+        policies::{
+            Policies as FuelPolicies,
+            PoliciesBits,
+            PolicyType,
+        },
     },
+    tai64,
 };
 
 #[derive(Clone)]
```
xgreenx, line 121: We should work with reference to everywhere in this file. And we should also avoid usage of `unswap_or_default`. All fields are always set, so it is strange why it can be `None`

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/serializer_adapter.rs
```diff
@@ -163,36 +249,1276 @@ fn proto_tx_from_tx(tx: FuelTransaction) -> ProtoTransaction {
...
+    }
+    if policies.is_set(PolicyType::MaxFee) {
+        values[3] = policies.get(PolicyType::MaxFee).unwrap_or_default();
+    }
+    if policies.is_set(PolicyType::Expiration) {
+        values[4] = policies.get(PolicyType::Expiration).unwrap_or_default();
+    }
+    if policies.is_set(PolicyType::Owner) {
+        values[5] = policies.get(PolicyType::Owner).unwrap_or_default();
+    }
```
xgreenx, line 632:
```rust
if let Some(value) = policies.get(PolicyType::Owner) {
    values[5] = value;
}
```

xgreenx, line 632:
```rust
if let Some(value) = policies.get(PolicyType::Owner) {
    values[5] = value;
}
```

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/serializer_adapter.rs
```diff
@@ -163,36 +249,1276 @@ fn proto_tx_from_tx(tx: FuelTransaction) -> ProtoTransaction {
...
-        PolicyType::MaxFee,
-        PolicyType::Expiration,
-    ];
-
-    let values = POLICY_ORDER
-        .iter()
-        .map(|policy_type| policies.get(*policy_type).unwrap_or_default())
-        .collect();
-
+    let mut values = [0u64; 6];
```
xgreenx, line 612: Why values for policies in an empty array? If the policy is not set, we shouldn't include them. If no policy is set, then it will be empty vector.

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/serializer_adapter.rs
```diff
@@ -163,36 +249,1276 @@ fn proto_tx_from_tx(tx: FuelTransaction) -> ProtoTransaction {
...
+        Input::MessageCoinPredicate(message) => ProtoInput {
+            variant: Some(ProtoInputVariant::MessageCoinPredicate(
+                ProtoMessageCoinPredicateInput {
+                    sender: message.sender.as_ref().to_vec(),
+                    recipient: message.recipient.as_ref().to_vec(),
+                    amount: message.amount,
+                    nonce: message.nonce.as_ref().to_vec(),
+                    witness_index: 0,
+                    predicate_gas_used: message.predicate_gas_used,
+                    data: Vec::new(),
```
xgreenx, line 458: If we use Rust definition to define variants inside of enums, then we can remove useless fields like `data` for `MessageCoinPredicate` and `witness_index` for `MessageCoinPredicate`

crates/services/block_aggregator_api/src/blocks/importer_and_db_source/serializer_adapter.rs
```diff
@@ -3,43 +3,119 @@ use crate::protobuf_types::V2Header as ProtoV2Header;
 use crate::{
```
xgreenx, line 3: I think instead of creating many procedure fucntions, we could implement `TryFrom` and `From` for types, plus, we could split them into its own folders by transactions and some common folder.

crates/types/src/blockchain/header.rs
```diff
@@ -79,6 +79,17 @@ impl BlockHeader {
...
+    /// Get the application portion of the header.
+    pub fn application_v1(
+        &self,
+    ) -> Option<&ApplicationHeader<GeneratedApplicationFieldsV1>> {
+        match self {
+            BlockHeader::V1(header) => Some(header.application()),
+            #[cfg(feature = "fault-proving")]
+            BlockHeader::V2(_header) => None,
+        }
+    }
```
xgreenx, line 91: We don't need to expose the application header. You can find an example on how to create a header here: https://github.com/FuelLabs/fuel-rust-indexer/blob/main/crates/receipts_manager/adapters/graphql_event_adapter.rs#L250

xgreenx: And the same for the block, you don't need manually calculate it, you can do that from the `Block::new`.


image

crates/types/src/test_helpers.rs
```diff
@@ -36,3 +71,603 @@ pub fn create_contract<R: Rng>(
...
+        fuel_block.header_mut().set_transaction_root(tx_root);
+        fuel_block.header_mut().set_message_outbox_root(msg_root);
+        fuel_block.header_mut().set_event_inbox_root(event_root);
+
+        // Consensus
+        // TODO: Include V2 Application with V2 Header
+        let application_hash = fuel_block.header().application_v1().unwrap().hash();
+        consensus_header.generated.application_hash = application_hash;
+        fuel_block.header_mut().set_consensus_header(consensus_header);
+        (fuel_block, msg_ids, event_root)
```
xgreenx, line 671: `Block::new` should be enough for you to create a header and a block. You don't need to implement all the logic by yourself.

You can clean up getters and setters which you've added, after you udpated code here=)

crates/types/src/test_helpers.rs
MitchTurner: Is the question about the `unused`? I think it's a bug in the compiler/linter. I can try to remove it again.

crates/services/block_aggregator_api/proto/api.proto
MitchTurner: Maybe. You're right. We could maybe do a chargeable tx pattern. I'm impartial. I don't think chargeable tx is the best abstraction, but it does dry up the code a bit.

crates/services/block_aggregator_api/proto/api.proto
MitchTurner: No. I doesn't:
 ```
 pub struct Mint {
    /// The location of the transaction in the block.
    #[cfg_attr(feature = "da-compression", compress(skip))]
    pub(crate) tx_pointer: TxPointer,
    /// The `Input::Contract` that assets are minted to.
    pub(crate) input_contract: input::contract::Contract,
    /// The `Output::Contract` that assets are being minted to.
    pub(crate) output_contract: output::contract::Contract,
    /// The amount of funds minted.
    pub(crate) mint_amount: Word,
    /// The asset IDs corresponding to the minted amount.
    pub(crate) mint_asset_id: AssetId,
    /// Gas Price used for current block
    pub(crate) gas_price: Word,
    #[serde(skip)]
    #[educe(PartialEq(ignore))]
    #[educe(Hash(ignore))]
    #[canonical(skip)]
    #[cfg_attr(feature = "da-compression", compress(skip))]
    pub(crate) metadata: Option<MintMetadata>,
}
```

PR #3127, opened 2025-10-21
Add multiple RPCs support to fuel-core-client.

its-saeed:
Please go to the `Preview` tab and select the appropriate sub-template:

* [Classic PR](?expand=1&template=default.md)
* [Bump version](?expand=1&template=bump_version.md)

### DIFF
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -800,7 +800,6 @@ impl FuelClient {
             io::Error::other("Failed to get status from the submission")
         })??;
 
-        println!("{status:?}");
         Ok(status)
     }
 


crates/client/src/reqwest_ext.rs
```diff
@@ -141,8 +141,10 @@ impl<ResponseData, Errors> CynicReqwestBuilder<ResponseData, Errors> {
     }
 }
 
-impl<ResponseData: serde::de::DeserializeOwned, Errors: serde::de::DeserializeOwned>
-    std::future::IntoFuture for CynicReqwestBuilder<ResponseData, Errors>
+impl<
+    ResponseData: serde::de::DeserializeOwned + Send + 'static,
+    Errors: serde::de::DeserializeOwned + Send + 'static,
+> std::future::IntoFuture for CynicReqwestBuilder<ResponseData, Errors>
```
rymnc: can you move the bounds into the next line ? 
```suggestion
impl<ResponseData, Errors> std::future::IntoFuture for CynicReqwestBuilder<ResponseData, Errors> 
where
    ResponseData: serde::de::DeserializeOwned + Send + 'static,
    Errors: serde::de::DeserializeOwned + Send + 'static
{
    /// rest of your code
}
```

crates/client/src/transport.rs
```diff
@@ -0,0 +1,293 @@
...
+            match self
+                .internal_query(query, url.clone(), required_block_height)
+                .await
+            {
+                Ok(response_data) => {
+                    return Ok(response_data);
+                }
+                Err(err) => last_err = Some(err),
+            }
+        }
```
rymnc: depending on how we plan to use this, i think it might be nicer to actually race the futures

crates/client/src/transport.rs
```diff
@@ -0,0 +1,293 @@
...
+            match self
+                .internal_subscribe(query, url.clone(), required_block_height)
+                .await
+            {
+                Ok(response_data) => return Ok(response_data),
+                Err(err) => last_err = Some(err),
+            }
+        }
+        Err(last_err.unwrap())
+    }
```
rymnc, line 176: i think we need to make this function a little more useful - here just the subscription creation is made fault tolerant, but what about event stream itself? 

this needs some additional work to make it work in such a way that multiple subscriptions are made and we dedup/multiplex the returned items from the event streams into a single durable event stream

rymnc: cc: @xgreenx for thoughts on this

rymnc (commented):
in general working around all the clones would be nice

crates/client/src/transport.rs
its-saeed: Good idea. Maybe [select_ok](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.19/futures/future/fn.select_ok.html) does the job nicely.

its-saeed: @xgreenx requested the previous implementation and a default endpoint that will be updated if the previous one fails.

crates/client/src/transport.rs
its-saeed: @xgreenx wanted the current implementation + default index logic similar to `query`

rymnc: how would this help for o2? only in the case that it tries re-subscription but durability of a single subscription is not guaranteed in this case

its-saeed: Yesterday's meeting summary:
 
He pointed out that subscribing to different endpoints, for example, for transaction updates, can yield divergent streams depending on the source, making it impossible to reconcile them reliably. Therefore, if a subscription attempt fails, it's our responsibility(FailOverTransport) to retry with the next available endpoint. However, once a stream is successfully established, any errors that occur within that stream are the end-user's responsibility to handle.

He also wanted to implement a default provider mechanism: if the current default fails to respond to a query or establish a subscription, we should automatically attempt the next available provider until one succeeds. Once a successful provider is found, it should be promoted as the new default for future requests.

rymnc (approved):

crates/client/src/lib.rs
```diff
@@ -5,6 +5,7 @@
 pub mod client;
 pub mod reqwest_ext;
 pub mod schema;
+mod transport;
```
xgreenx:
```suggestion
pub mod transport;
```

crates/client/src/client.rs
```diff
@@ -932,6 +800,7 @@ impl FuelClient {
             io::Error::other("Failed to get status from the submission")
         })??;
 
+        println!("{status:?}");
```
xgreenx:
```suggestion
```

PR #3147, opened 2025-11-15
Improve S3 Configuration

MitchTurner:
## Linked Issues/PRs

Update to use the protobuf definitions here:
https://github.com/FuelLabs/fuel-core-protobuf/pull/2

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/.github/actions/slack-notify-template/action.yml
+++ b/.github/actions/slack-notify-template/action.yml
@@ -1,0 +1,25 @@
+name: Notify Slack on Failure
+description: Sends notification to Slack if job fails
+
+inputs:
+  github_token:
+    description: Github Token Secret
+    required: true
+  slack_webhook:
+    description: Slack webhook URL
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Notify if Job Fails
+      uses: ravsamhq/notify-slack-action@v2
+      with:
+        status: ${{ job.status }}
+        token: ${{ inputs.github_token }}
+        notification_title: '{workflow} has {status_message}'
+        message_format: '{emoji} *{workflow}* {status_message} in <{repo_url}|{repo}> : <{run_url}|View Run Results>'
+        footer: ''
+        notify_when: failure
+      env:
+        SLACK_WEBHOOK_URL: ${{ inputs.slack_webhook }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
         run: cargo install cargo-sort
       - name: Run Cargo.toml sort check
         run: cargo sort -w --check
-      - uses: FuelLabs/.github/.github/actions/slack-notify-template@master
+      - uses: ./.github/actions/slack-notify-template
         if: always() && github.ref == 'refs/heads/master'
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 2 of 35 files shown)


crates/services/block_aggregator_api/src/db/storage_or_remote_db.rs
```diff
@@ -46,40 +47,35 @@ impl<R, S> StorageOrRemoteDB<R, S> {
...
+            if let Some(aws_endpoint_url) = &aws_endpoint_url {
                 builder.set_endpoint_url(Some(aws_endpoint_url.clone()));
             }
-
+            // TODO: This is a little gross spinning up the runtime just to get credentials.
+            //   If this takes a long time maybe we should move this to the service or something.
+            let rt_handle = tokio::runtime::Handle::current();
+            let credentials = tokio::task::block_in_place(|| {
+                rt_handle.block_on(DefaultCredentialsChain::builder().build())
+            });
```
MitchTurner: Gross. The `DefaultCredentialsChain` builder only provides an async `build` method. We could move it into the service somewhere which is async.

crates/services/block_aggregator_api/src/db/storage_or_remote_db.rs
```diff
@@ -46,40 +47,35 @@ impl<R, S> StorageOrRemoteDB<R, S> {
...
-                    aws_id,
-                    aws_secret,
-                    None,
-                    None,
-                    "block-aggregator",
-                ))
+                .credentials_provider(credentials)
                 .behavior_version_latest()
+                // TODO: Move to config
+                .region(Region::new(Cow::Owned("us-east-1".to_string())))
```
MitchTurner: Need to resolve this before merging. The `cross_region_access_enabled` method that exists for the Java sdk doesn't seem to exist for Rust.

crates/services/block_aggregator_api/src/db/remote_cache.rs
mchristopher: Something that would be better for S3 due to how data is distributed behind the scenes (to ensure we don't have hot partitions) is to separate values into "directories".

For this, it would be best to split the 8 hex characters into 4 parts, with the `/` character between them. Something like this:

```rust
let raw = format!("{:08x}", height).as_bytes();
format!(
    "{}/{}/{}/{}",
    &raw[0..2],
    &raw[2..4],
    &raw[4..6],
    &raw[6..8],
)
```

mchristopher, line 131: Can we gzip this data before storing it & then add a `.content_encoding("gzip")` to the put call?

MitchTurner: Haha. I was thinking that originally. Just focused on different things. Or even each hex char. 256/folder isn't too bad though.

mchristopher: Yeah, 256 is generally considered just fine for load balancing.

MitchTurner: ended up being:
```rust
pub fn block_height_to_key(height: &BlockHeight) -> String {
    let raw: [u8; 4] = height.to_bytes();
    format!(
        "{:02}/{:02}/{:02}/{:02}",
        &raw[0], &raw[1], &raw[2], &raw[3]
    )
}
```

and confirmed in localstack:
```
? awslocal s3api list-objects \
  --bucket test-bucket
{
    "Contents": [
        {
            "Key": "00/00/00/00",
            . . .
        },
        {
            "Key": "00/00/00/01",
            . . .
        }
    ],
    "RequestCharged": null
}
```

MitchTurner: Done

PR #3155, opened 2025-11-20
Add integration tests for preconfirmation transaction indices to verify correct indexing.

its-saeed:
Add integration tests for preconfirmation transaction indices to verify correct indexing.


.gitignore
```diff
@@ -4,6 +4,7 @@
 *.tfstate*
 *.terraform.lock.hcl*
 .vscode
+.zed
```
Voxelot, line 7: 🥇

PR #3164, opened 2025-12-09
Splitted read storage from write storage

xgreenx:
- Splitted read storage from write storage
- Instead of using channels, now we query data directly from the storage adapter
- Removed cloning of blocks by suing `Arc`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -100,7 +100,7 @@ flate2 = "1.1.5"
 fuel-core = { version = "0.47.1", path = "./crates/fuel-core", default-features = false }
 fuel-core-bin = { version = "0.47.1", path = "./bin/fuel-core" }
 # Workspace members
-fuel-core-block-aggregator-api = { version = "0.47.1", path = "crates/services/block_aggregator_api" }
+fuel-core-block-aggregator-api = { version = "0.47.1", path = "./crates/services/block_aggregator_api" }
 fuel-core-chain-config = { version = "0.47.1", path = "./crates/chain-config", default-features = false }
 fuel-core-client = { version = "0.47.1", path = "./crates/client" }
 fuel-core-compression = { version = "0.47.1", path = "./crates/compression" }
--- a/ci_checks.sh
+++ b/ci_checks.sh
@@ -27,7 +27,7 @@ cargo check -p fuel-core-client --target wasm32-unknown-unknown --no-default-fea
 cargo check -p fuel-core-chain-config --target wasm32-unknown-unknown --no-default-features &&
 cargo check -p fuel-core-executor --target wasm32-unknown-unknown --no-default-features --features alloc &&
 cargo check -p fuel-core --no-default-features &&
-cargo test check --features rpc --no-default-features &&
+cargo check --features rpc --no-default-features --tests &&
 cargo make check --all-features --locked &&
 cargo make check --locked &&
 OVERRIDE_CHAIN_CONFIGS=true cargo test --test integration_tests local_node &&
--- a/crates/client/Cargo.toml
+++ b/crates/client/Cargo.toml
@@ -42,7 +42,7 @@ cynic = { workspace = true }
 derive_more = { workspace = true }
 eventsource-client = { version = "0.13.0", optional = true }
 flate2 = { workspace = true, optional = true }
-fuel-core-block-aggregator-api = { path = "../services/block_aggregator_api", optional = true }
+fuel-core-block-aggregator-api = { workspace = true, optional = true }
 fuel-core-types = { workspace = true, features = ["alloc", "serde"] }
 futures = { workspace = true, optional = true }
 hex = { workspace = true }
--- a/crates/services/block_aggregator_api/src/db/table.rs
+++ b/crates/services/block_aggregator_api/src/db/table.rs
@@ -1,18 +1,11 @@
 use fuel_core_storage::{
     Mappable,
     blueprint::plain::Plain,
-    codec::{
-        Decode,
-        Encode,
-    },
     kv_store::StorageColumn,
     structured_storage::TableWithBlueprint,
 };
 use fuel_core_types::fuel_types::BlockHeight;
-use std::{
-    borrow::Cow,
-    sync::Arc,
-};
+use std::sync::Arc;
 
 #[repr(u32)]
 #[derive(
@@ -62,7 +55,7 @@ impl Mappable for Blocks {
 }
 
 impl TableWithBlueprint for Blocks {
-    type Blueprint = Plain<Primitive<4>, Postcard>;
+    type Blueprint = Plain<Primitive<4>, Raw>;
     type Column = Column;
 
     fn column() -> Self::Column {
@@ -113,34 +106,5 @@ impl TableWithBlueprint for LatestBlock {
 use fuel_core_storage::codec::{
     postcard::Postcard,
     primitive::Primitive,
+    raw::Raw,
 };
-use prost::Message;
-
-pub struct ProtoBufCodec;
-
-impl<T> Encode<T> for ProtoBufCodec
-where
-    T: Sized + Message,
-{
-    type Encoder<'a>
-        = Cow<'a, [u8]>
-    where
-        T: 'a;
-
-    fn encode(value: &T) -> Self::Encoder<'_> {
-        let mut buffer = Vec::new();
-        value.encode(&mut buffer).expect(
-            "It should be impossible to fail unless serialization is not implemented, which is not true for our types.",
-        );
-        buffer.into()
-    }
-}
-
-impl<T> Decode<T> for ProtoBufCodec
-where
-    T: Message + Default,
-{
-    fn decode(bytes: &[u8]) -> anyhow::Result<T> {
-        Ok(T::decode(bytes)?)
-    }
-}


crates/fuel-core/src/p2p_test_helpers.rs
```diff
@@ -93,7 +93,7 @@ pub struct CustomizeConfig {
     max_discovery_peers_connected: Option<u32>,
     subscribe_to_transactions: Option<bool>,
     #[cfg(feature = "rpc")]
-    rpc_config: Option<fuel_core_block_aggregator_api::integration::Config>,
+    rpc_config: Option<fuel_core_block_aggregator_api::service::Config>,
```
MitchTurner, line 96: We don't like `integration` over `service` :P ?

crates/services/block_aggregator_api/src/api/protobuf_adapter.rs
```diff
@@ -49,34 +45,39 @@ use tonic::{
 #[cfg(test)]
 mod tests;
 
-pub struct Server {
-    query_sender:
-        tokio::sync::mpsc::Sender<BlockAggregatorQuery<BlockRangeResponse, ProtoBlock>>,
+#[cfg_attr(test, mockall::automock)]
+pub trait BlocksAggregatorApi: Send + Sync + 'static {
```
MitchTurner, line 50: In reality, we could just depend on the BlocksProvider trait here and directly on the broadcast::Sender that we have in the SharedStorage. This would parallel what we are doing in the aggregator service already, since it takes a dep on BlocksStorage and broadcast::Sender.

I think that's the right pattern.

crates/services/block_aggregator_api/src/api/protobuf_adapter.rs
```diff
@@ -153,55 +144,53 @@ impl BlockAggregator for Server {
...
     }
 }
 
-pub struct ProtobufAPI {
-    _server_service: ServiceRunner<ServerTask>,
-    query_receiver:
-        tokio::sync::mpsc::Receiver<BlockAggregatorQuery<BlockRangeResponse, ProtoBlock>>,
+pub struct UninitializedTask<B> {
+    addr: SocketAddr,
+    api: B,
```
MitchTurner: See above comment. Maybe rename something like `block_aggregator`

crates/services/block_aggregator_api/src/api/protobuf_adapter.rs
```diff
@@ -211,87 +200,51 @@ impl RunnableService for ServerTask {
...
-
-    fn get_router(&mut self) -> anyhow::Result<Router> {
-        self.router
-            .take()
-            .ok_or_else(|| anyhow!("Router has not been initialized yet"))
+        let task = Task {
+            addr: self.addr,
+            router: Some(router),
+        };
+        Ok(task)
```
MitchTurner: Do we really need an `UninitializedTask` just for this? What's the point of `into_task` for anyway haha?

crates/services/block_aggregator_api/src/api/protobuf_adapter.rs
```diff
@@ -211,87 +200,51 @@ impl RunnableService for ServerTask {
...
-            query_receiver,
-        };
-        Ok(api)
-    }
-}
-
-impl BlockAggregatorApi for ProtobufAPI {
-    type BlockRangeResponse = BlockRangeResponse;
-    type Block = ProtoBlock;
+pub type APIService<B> = ServiceRunner<UninitializedTask<B>>;
```
MitchTurner: Nit: I'm not a fan of aliases like this, as it adds misdirection. If I publically saw an `APIService`, I wouldn't assume that it was in fact the `ServiceRunner`.

crates/services/block_aggregator_api/src/api/protobuf_adapter/tests.rs
```diff
@@ -35,76 +32,52 @@ use futures::{
...
-    // given
-    let path = free_local_addr();
-    let mut api = ProtobufAPI::new(path.to_string(), 100).unwrap();
-    tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+    let socket = free_local_addr();
+    let mut api = MockBlocksAggregatorApi::default();
+
+    // Given
+    api.expect_get_current_height()
+        .times(1)
```
MitchTurner, line 52: This is why I don't like `mockall`. We don't care how many times it is called--it's over-constrained.

i.e. it's like we have an extra `// then` that isn't in the `// then` section.

crates/services/block_aggregator_api/src/api/protobuf_adapter/tests.rs
```diff
@@ -35,76 +32,52 @@ use futures::{
...
-    } else {
-        panic!("expected GetCurrentHeight query");
-    }
-    let res = handle.await.unwrap();
+
+    // When
+    let result = client.get_synced_block_height(BlockHeightRequest {}).await;
+
+    // Then
+    let value = result.expect("could not get height");
```
MitchTurner, line 68: Nit:
```suggestion
    let value = client.get_synced_block_height(BlockHeightRequest {}).await.unwrap();

    // Then
```

MitchTurner: Nit. We do similar unwraps below with separate llines for the unwraps.

crates/services/block_aggregator_api/src/api/protobuf_adapter/tests.rs
```diff
@@ -35,76 +32,52 @@ use futures::{
...
 
     // assert client received expected value
```
MitchTurner, line 70: We can get rid of these comments now. Was kinda a confusing flow before but now we just use a mock.

crates/services/block_aggregator_api/src/service.rs
```diff
@@ -0,0 +1,367 @@
...
+    },
+    // Assumes another node is publishing blocks to S3 bucket, but relaying details
+    S3NoPublish {
+        bucket: String,
+        endpoint_url: Option<String>,
+        requester_pays: bool,
+    },
+}
+
+pub struct SharedState<S>
```
MitchTurner, line 100: It's an anti-pattern to group the blocks_broadcast and the storage together like this. They aren't related except as an adapter for the proto api. So you could move this to the adapter code and rename it `BlockAggregatorAdapter` or whatever.

There shouldn't be a SharedState type, we can just construct the adapter with the rest of the service.

crates/services/block_aggregator_api/src/service.rs
```diff
@@ -0,0 +1,367 @@
...
+    pub fn new_block_subscription(
+        &self,
+    ) -> impl Stream<Item = anyhow::Result<(BlockHeight, Arc<S::Block>)>> + 'static {
+        let receiver = self.blocks_broadcast.subscribe();
+        tokio_stream::wrappers::BroadcastStream::new(receiver)
+            .map(|result| result.map_err(|e| anyhow::anyhow!("Broadcast error: {:?}", e)))
+    }
+}
+
+impl<S> BlocksAggregatorApi for SharedState<S>
```
MitchTurner, line 168: We shouldn't be iimplementing a foreign port like this here. See above comment on moving renaming and moving SharedState.

crates/services/block_aggregator_api/src/task.rs
```diff
@@ -0,0 +1,108 @@
...
+    try_or_stop,
+};
+use std::{
+    fmt::Debug,
+    sync::Arc,
+};
+
+/// The Block Aggregator service, which aggregates blocks from a source and stores them in a database
+/// Queries can be made to the service to retrieve data from the `DB`
+pub struct Task<S1, S2, Blocks>
```
MitchTurner, line 30: Why S1 and S2? Can we use S and P? or Storage and PRovider?

crates/services/block_aggregator_api/src/db.rs
```diff
@@ -10,29 +10,33 @@ pub mod storage_db;
...
+    ) -> Result<Self::BlockRangeResponse>;
 
     /// Retrieves the current height of the aggregated blocks If there is a break in the blocks,
     /// i.e. the blocks are being aggregated out of order, return the height of the last
     /// contiguous block
-    fn get_current_height(
-        &self,
-    ) -> impl Future<Output = Result<Option<BlockHeight>>> + Send;
+    fn get_current_height(&self) -> Result<Option<BlockHeight>>;
+}
```
MitchTurner, line 26: This can move the the protobuf adapter code and replace `BlocksAggregatorAPI` port, like I mentioned above.

crates/services/block_aggregator_api/src/db/storage_db.rs
```diff
@@ -133,8 +133,19 @@ where
         }
         Ok(())
     }
+}
+
+impl<S> BlocksProvider for StorageBlocksProvider<S>
```
MitchTurner, line 103: If we move this port to the protobuf adapter code, this impl should go with it.

crates/services/block_aggregator_api/src/db/storage_or_remote_db.rs
```diff
@@ -108,30 +111,41 @@ where
...
     }
+}
 
-    async fn get_block_range(
+impl<S> BlocksProvider for StorageOrRemoteBlocksProvider<S>
+where
+    S: 'static,
+    S: KeyValueInspect<Column = Column>,
+    S: AtomicView,
+    S::LatestView: Unpin + Send + Sync + KeyValueInspect<Column = Column> + 'static,
```
MitchTurner, line 124: IDK how you remember which traits to use to keep it this clean :P

PR #3166, opened 2025-12-10
Unified new and old block in one stream to simplify handling of these blocks

xgreenx:


crates/fuel-core/src/database.rs
```diff
@@ -462,38 +449,12 @@ impl Modifiable for Database<BlockAggregatorDatabase> {
...
-            tx.storage_as_mut::<LatestBlock>().insert(&(), &new)?;
-            tx.commit().map_err(|e: StorageError| anyhow!(e))?;
-        }
-        Ok(())
+        commit_changes_with_height_update(self, changes, |iter| {
+            iter.iter_all_keys::<fuel_core_block_aggregator_api::db::table::Blocks>(Some(
+                IterDirection::Reverse,
+            ))
+            .try_collect()
+        })
```
MitchTurner, line 457: Was it not being added before?

crates/services/block_aggregator_api/src/blocks/old_block_source.rs
```diff
@@ -1,143 +1,190 @@
...
+        receipts: &[Vec<FuelReceipt>],
     ) -> Result<Self::Block>;
 }
 
-/// A block source that combines an importer and a database sync task.
-/// Old blocks will be synced from a target database and new blocks will be received from
-/// the importer
-pub struct ImporterAndDbSource<Serializer, DB, Receipts>
+pub trait TxReceipts: Send + Sync + 'static {
+    fn get_receipts(&self, tx_id: &TxId) -> Result<Vec<Receipt>>;
```
MitchTurner, line 41: Why remove the `async`?

crates/services/block_aggregator_api/src/blocks/importer_and_db_source.rs
```diff
@@ -1,143 +1,190 @@
...
+}
+
+pub struct OldBlocksSource<Serializer, DB, Receipts> {
+    serializer: Arc<Serializer>,
+    db: Arc<DB>,
+    receipts: Arc<Receipts>,
+}
+
+impl<Serializer, DB, Receipts> OldBlocksSource<Serializer, DB, Receipts> {
+    pub fn new(serializer: Arc<Serializer>, db: DB, receipts: Receipts) -> Self {
```
MitchTurner: Why is just `Serializer` wrapped as a param?

crates/services/block_aggregator_api/src/blocks/importer_and_db_source.rs
```diff
@@ -1,143 +1,190 @@
...
+
+impl<Serializer, DB, Receipts> OldBlocksSource<Serializer, DB, Receipts>
 where
-    Serializer: BlockSerializer + Send + Sync + 'static,
-    <Serializer as BlockSerializer>::Block: Send + Sync + 'static,
     DB: Send + Sync + 'static,
     DB: StorageInspect<FuelBlocks, Error = StorageError>,
     DB: StorageInspect<Transactions, Error = StorageError>,
     Receipts: TxReceipts,
+    Serializer: BlockSerializer,
```
MitchTurner: Note: I think we should rename `BlockSerializer` before we merge all of this (unless we go back to storing blocks as bytes). It's not a serializer, more a converter haha.

crates/services/block_aggregator_api/src/task.rs
```diff
@@ -26,60 +33,117 @@ use std::{
...
+                }
+                Err(_) => true,
+            };
+
+            async move { take }
+        });
+
+        let receiver = self.shared_state.blocks_broadcast.subscribe();
+        let life_stream =
+            tokio_stream::wrappers::BroadcastStream::new(receiver).map(|result| {
```
MitchTurner, line 113: So the idea is that the we send to all the subscribers, and also to itself? It's a little confusing reading the `run` function as it's not clear what the `continuous_stream_of_events` is (As opposed to just calling the port abstraction).

 But in terms of complexity, it is better. 
 
 Can we move the stream construct into the `impl BlockSource` somehow?

MitchTurner: Thinking about this more, I'd like to keep the complexity inside the domain. It's better for testing.

But maybe we can rename `continuous_stream_of_events` to `old_and_new_block_stream`? Just be more explicit.

MitchTurner: Also
```
let _ = self
    .shared_state
    .blocks_broadcast
    .send((height, Arc::new(block)));
```
 Is not clear is sending to that stream too. How can we communicate it better in the `run` function that that's where the blocks are coming from?

tests/tests/dos.rs
```diff
@@ -445,68 +442,6 @@ async fn concurrency_limit_0_allows_unthrottled_queries() {
     assert_eq!(result.status(), 200);
 }
 
-#[tokio::test(flavor = "multi_thread", worker_threads = 8)]
-async fn concurrency_limit_1_prevents_concurrent_queries() {
```
MitchTurner, line 449: lol. Why we removing this random test?

crates/services/block_aggregator_api/src/blocks/old_block_source.rs
```diff
@@ -1,143 +1,190 @@
 use crate::{
-    blocks::{
-        BlockSource,
-        BlockSourceEvent,
-        importer_and_db_source::importer_service::ImporterTask,
-    },
+    blocks::BlockSource,
```
MitchTurner, line 2: We can rename `importer_and_db_source.rs` to e.g. `old_block_source.rs`

tests/tests/dos.rs
xgreenx: Because it fails quite often and test itself doesn't actually test what it was indeed to test

crates/services/block_aggregator_api/src/blocks/importer_and_db_source.rs
xgreenx: Because we cloned it before, other are not cloned

crates/services/block_aggregator_api/src/blocks/old_block_source.rs
xgreenx: Because we don't need to make code harder beforehand. When we need this function to be `async`, we will make it to be `async`=)

crates/fuel-core/src/database.rs
xgreenx: It was not possible to do before, because we were committing blocks in random order. Now we commit them only sequentially

PR #3167, opened 2025-12-11
Use nested receipts and blocks as bytes

MitchTurner:
## Description

Adds use of these new protobuf types:
https://github.com/FuelLabs/fuel-core-protobuf/pull/6

Specifically Changing `Vec<Receipt>` -> `Vec<Vec<Receipt>>` and adding a bytes type for Block.

### DIFF
--- a/crates/services/block_aggregator_api/src/blocks/old_block_source/convertor_adapter.rs
+++ b/crates/services/block_aggregator_api/src/blocks/old_block_source/convertor_adapter.rs
@@ -93,7 +93,7 @@ mod tests {
               // when
               let receipts = vec![receipts];
               let bytes = convertor.convert_block(&block, &receipts).unwrap();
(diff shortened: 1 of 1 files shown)


crates/services/block_aggregator_api/src/blocks/old_block_source/convertor_adapter.rs
```diff
@@ -8,22 +15,18 @@ use crate::{
...
 use fuel_core_types::{
     blockchain::block::Block as FuelBlock,
     fuel_tx::Receipt as FuelReceipt,
 };
+use prost::Message;
+use std::sync::Arc;
 
 #[derive(Clone)]
-pub struct ConvertorAdapter;
+pub struct ProtobufBlockConverter;
```
MitchTurner, line 26: We discussed the possibility of adding a large `Vec` here to use as a buffer that doesn't need to be reallocated each time.

Idk what the actual performance changes would be. But might be good.

PR #3168, opened 2025-12-16
add proto and s3 to client

MitchTurner:
## Description

This is adding support for accessing the new block aggregation RPC and subsequent S3 calls using the Fuel Client.

### DIFF
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -286,28 +286,22 @@ impl FromStr for FuelClient {
     type Err = anyhow::Error;
 
     fn from_str(str: &str) -> Result<Self, Self::Err> {
-        let mut raw_url = str.to_string();
-        if !raw_url.starts_with("http") {
-            raw_url = format!("http://{raw_url}");
-        }
-
-        let mut url = reqwest::Url::parse(&raw_url)
-            .map_err(anyhow::Error::msg)
-            .with_context(|| format!("Invalid fuel-core URL: {str}"))?;
-        url.set_path("/v1/graphql");
-
-        Ok(Self {
-            transport: FailoverTransport::new(vec![url])?,
-            require_height: ConsistencyPolicy::Auto {
-                height: Arc::new(Mutex::new(None)),
-            },
-            chain_state_info: Default::default(),
-            #[cfg(feature = "rpc")]
-            rpc_client: None,
-            #[cfg(feature = "rpc")]
-            aws_client: None,
-        })
+        Self::with_urls(vec![str_to_url(str)?])
     }
+}
+
+fn str_to_url(str: &str) -> anyhow::Result<Url> {
+    let mut raw_url = str.to_string();
+    if !raw_url.starts_with("http") {
+        raw_url = format!("http://{raw_url}");
+    }
+
+    let mut url = reqwest::Url::parse(&raw_url)
+        .map_err(anyhow::Error::msg)
+        .with_context(|| format!("Invalid fuel-core URL: {str}"))?;
+    url.set_path("/v1/graphql");
+
+    Ok(url)
 }
 
 impl<S> From<S> for FuelClient
@@ -340,10 +334,13 @@ impl FuelClient {
 
     #[cfg(feature = "rpc")]
     pub async fn new_with_rpc<G: AsRef<str>, R: AsRef<str>>(
-        graph_ql_url: G,
+        graph_ql_urls: impl Iterator<Item = G>,
         rpc_url: R,
     ) -> anyhow::Result<Self> {
-        let mut client = Self::new(graph_ql_url)?;
+        let urls = graph_ql_urls
+            .map(|str| str_to_url(str.as_ref()))
+            .try_collect()?;
+        let mut client = Self::with_urls(urls)?;
         let mut raw_rpc_url = <R as AsRef<str>>::as_ref(&rpc_url).to_string();
         if !raw_rpc_url.starts_with("http") {
             raw_rpc_url = format!("http://{raw_rpc_url}");
@@ -1686,6 +1683,7 @@ impl FuelClient {
             .clone()
             .ok_or(io::Error::other("RPC client not initialized"))
     }
+
     pub async fn get_block_range(
         &self,
         start: BlockHeight,
@@ -1736,13 +1734,11 @@ impl FuelClient {
             Payload::Bytes(bytes) => {
                 let proto_block =
                     ProtoBlock::decode(bytes.as_slice()).map_err(io::Error::other)?;
-                fuel_block_from_protobuf(proto_block, &[], Bytes32::default()).map_err(
-                    |e| {
-                        io::Error::other(format!(
-                            "Failed to convert RPC block to internal block: {e:?}"
-                        ))
-                    },
-                )
+                fuel_block_from_protobuf(proto_block).map_err(|e| {
+                    io::Error::other(format!(
+                        "Failed to convert RPC block to internal block: {e:?}"
+                    ))
+                })
             }
             Payload::Remote(remote) => {
                 let RemoteBlockResponse { location } = remote;
@@ -1767,16 +1763,12 @@ impl FuelClient {
                             ProtoBlock::decode(block_bytes.as_slice()).map_err(|e| {
                                 io::Error::other(format!("Failed to decode block: {e}"))
                             })?;
-                        let (block, receipts) = fuel_block_from_protobuf(
-                            block,
-                            &[],
-                            Bytes32::default(),
-                        )
-                        .map_err(|e| {
-                            io::Error::other(format!(
-                                "Failed to convert RPC block to internal block: {e:?}"
-                            ))
-                        })?;
+                        let (block, receipts) =
+                            fuel_block_from_protobuf(block).map_err(|e| {
+                                io::Error::other(format!(
+                                    "Failed to convert RPC block to internal block: {e:?}"
+                                ))
+                            })?;
                         Ok((block, receipts))
                     }
                     _ => Err(io::Error::other("Remote blocks are not supported yet")),
--- a/crates/fuel-core/src/service.rs
+++ b/crates/fuel-core/src/service.rs
@@ -45,6 +45,7 @@ use fuel_core_types::{
     fuel_types::BlockHeight,
 };
 
+use self::adapters::BlockImporterAdapter;
 use crate::{
     combined_database::{
         CombinedDatabase,
@@ -60,7 +61,10 @@ use crate::{
     },
 };
 
-use self::adapters::BlockImporterAdapter;
+#[cfg(feature = "rpc")]
+use crate::database::database_description::block_aggregator::BlockAggregatorDatabase;
+#[cfg(feature = "rpc")]
+use fuel_core_block_aggregator_api::db::storage_or_remote_db::StorageOrRemoteBlocksProvider;
 
 pub mod adapters;
 pub mod config;
@@ -98,6 +102,13 @@ pub struct SharedState {
     pub compression: Option<fuel_core_compression_service::service::SharedData>,
     /// The gas price service shared data.
     pub gas_price_service: fuel_core_gas_price_service::v1::service::SharedData,
+    #[cfg(feature = "rpc")]
+    /// The block aggregator RPC shared data.
+    pub block_aggregator_rpc: Option<
+        fuel_core_block_aggregator_api::service::SharedState<
+            StorageOrRemoteBlocksProvider<Database<BlockAggregatorDatabase>>,
+        >,
+    >,
 }
 
 pub struct FuelService {
@@ -171,10 +182,9 @@ impl FuelService {
         #[cfg(feature = "rpc")]
         let rpc_address = runner
             .shared
-            .config
-            .rpc_config
+            .block_aggregator_rpc
             .clone()
-            .map(|config| config.addr);
+            .map(|state| state.bound_address);
 
         Ok(FuelService {
             sub_services,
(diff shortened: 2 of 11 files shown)


crates/client/src/client.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unused_imports)]
```
rymnc: ?

crates/client/src/client.rs
```diff
@@ -246,6 +284,8 @@ pub struct FuelClient {
     transport: FailoverTransport,
     require_height: ConsistencyPolicy,
     chain_state_info: ChainStateInfo,
+    #[cfg(feature = "rpc")]
+    rpc_client: Option<ProtoBlockAggregatorClient<Channel>>,
```
rymnc, line 280: probably a good chance to create a `FuelClientBuilder` now, all the constructors are duped code right now

crates/client/src/client.rs
```diff
@@ -1620,3 +1725,183 @@ impl FuelClient {
             .transpose()?)
     }
 }
+
+#[cfg(feature = "rpc")]
+impl FuelClient {
+    fn rpc_client(&self) -> io::Result<ProtoBlockAggregatorClient<Channel>> {
```
rymnc, line 1681: io::Result ?

crates/client/src/client.rs
```diff
@@ -1620,3 +1725,183 @@ impl FuelClient {
...
+        let config = builder.force_path_style(true).build();
+        Client::from_conf(config)
+    }
+
+    fn unzip_bytes(bytes: &[u8]) -> Vec<u8> {
+        let mut decoder = GzDecoder::new(bytes);
+        let mut output = Vec::new();
+        decoder.read_to_end(&mut output).unwrap();
+        output
+    }
```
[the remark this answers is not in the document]
rymnc: cursor is right here

crates/client/src/client.rs
```diff
@@ -1620,3 +1725,183 @@ impl FuelClient {
...
+        let builder = aws_sdk_s3::config::Builder::from(&sdk_config);
+        let config = builder.force_path_style(true).build();
+        Client::from_conf(config)
+    }
+
+    fn unzip_bytes(bytes: &[u8]) -> Vec<u8> {
+        let mut decoder = GzDecoder::new(bytes);
+        let mut output = Vec::new();
+        decoder.read_to_end(&mut output).unwrap();
+        output
```
rymnc: not guaranteed that we used GzEncoder to push to s3, i think this part needs some reworking to make it a bit more stricter. you can check the header to see if it can be gzip decompressed ~ unwrap probably not a good idea here

crates/client/src/client.rs
```diff
@@ -300,13 +342,76 @@ impl FuelClient {
...
+    ) -> anyhow::Result<Self> {
+        let mut raw_url = <G as AsRef<str>>::as_ref(&graph_ql_url).to_string();
+        if !raw_url.starts_with("http") {
+            raw_url = format!("http://{raw_url}");
+        }
+
+        let mut url = reqwest::Url::parse(&raw_url)
+            .map_err(anyhow::Error::msg)
+            .with_context(|| format!("Invalid Fuel GraphQL URL: {raw_url}"))?;
+        url.set_path("/v1/graphql");
```
mchristopher: This is not a valid assumption, other RPC providers like Quicknode and Ankr add prefixes onto the URL and don't support accessing the graphql endpoint from the root.

The URLs for both RPC & GraphQL should be explicitly passed in here, I don't think there is a way we can auto-configure this.

crates/client/src/client.rs
```diff
@@ -1620,3 +1725,183 @@ impl FuelClient {
...
+                match location {
+                    Some(Location::S3(s3)) => {
+                        let RemoteS3Bucket {
+                            bucket,
+                            key,
+                            endpoint,
+                            ..
+                        } = s3;
+                        let s3_url = endpoint
+                            .unwrap_or_else(|| "https://s3.amazonaws.com".to_string());
```
mchristopher: You should not pass in manual endpoints if one is not provided. If something is not provided, let the S3 SDK figure it out based on the bucket name.

crates/client/src/client.rs
```diff
@@ -1620,3 +1725,183 @@ impl FuelClient {
...
+                io::Error::other(format!("Failed to get object from S3: {e:?}"))
+            })?
+            .into_bytes();
+        Ok(bytes)
+    }
+
+    async fn aws_client(url: &str) -> Client {
+        let credentials = DefaultCredentialsChain::builder().build().await;
+        let _aws_region =
+            std::env::var("AWS_REGION").expect("AWS_REGION env var must be set");
```
mchristopher: This is not a valid assumption on the client side. Even when it's required, the AWS SDK itself will throw an error when it doesn't have the right data.

crates/client/src/client.rs
```diff
@@ -1620,3 +1725,183 @@ impl FuelClient {
...
+        let sdk_config = aws_config::defaults(BehaviorVersion::latest())
+            .credentials_provider(credentials)
+            .endpoint_url(url)
+            .load()
+            .await;
+        let builder = aws_sdk_s3::config::Builder::from(&sdk_config);
+        let config = builder.force_path_style(true).build();
+        Client::from_conf(config)
+    }
```
mchristopher, line 1823: Should create a global client that gets reused after the first init. Need to verify, but I'm pretty sure everything is thread safe in the AWS SDK, so cross-thread access shouldn't be a problem

crates/client/src/client.rs
MitchTurner: Thanks. Nice catch. I just moved a bunch of stuff over from the test code but didn't fix all of the differences.

crates/client/src/client.rs
rymnc: tbh i don't think it's worth adding these dependencies if we can fetch using http

rymnc: tbh i don't think it's worth adding these dependencies if we can fetch using http

crates/client/src/client.rs
MitchTurner: I know. It's what all the other responses from the client give. I'm not sure why.

crates/client/src/client.rs
MitchTurner: I don't think it's quite there. I'm a big fan of builders, but essentially it's just 1 optional param.

I've deduped the code (and one of them I didn't end up using so I removed): https://github.com/FuelLabs/fuel-core/pull/3168/commits/5681c3ec50f009a367b2938c1cb107ba1f35652d

crates/client/src/client.rs
MitchTurner: We are making 2 assumptions here:
- the bytes are the serialized ProtoBlock
- the bytes have been gzipped

Would you prefer the protobuf message to include `compression` and `serialization` enums to allow expansion of this in the future? Or we could add more clarfication to the `Location` type:
```
/// Nested message and enum types in `RemoteBlockResponse`.
pub mod remote_block_response {
    #[derive(serde::Serialize, serde::Deserialize)]
    #[allow(clippy::large_enum_variant)]
    #[derive(Clone, PartialEq, ::prost::Oneof)]
    pub enum Location {
        #[prost(message, tag = "1")]
        Http(super::RemoteHttpEndpoint),
        #[prost(message, tag = "2")]
        S3(super::RemoteS3Bucket),
    }
}
```
Maybe change `S3` to `S3ProtoGZip` lol. Then we could just add new variants if they ever happen.

I understand the concern, but I'm also _OK_ with the current design.

crates/client/src/client.rs
MitchTurner: This is the existing behavior. I'm not changing anything to do with the GraphQL server in this code. I think it would be a breaking change to change this here.

I did have a couple bugs in the way the rpc url was being set though, and I've fixed those.

crates/client/src/client.rs
MitchTurner: https://github.com/FuelLabs/fuel-core/pull/3168/commits/09cdf7dc69e04634036583a1b5263936bd299925

crates/client/src/client.rs
MitchTurner: https://github.com/FuelLabs/fuel-core/pull/3168/commits/49e6e244c905c998e8984ae3127590842c33287d

crates/client/src/client.rs
```diff
@@ -1620,3 +1672,183 @@ impl FuelClient {
...
+                }
+            }
+        }
+    }
+    async fn get_block_from_s3_bucket(
+        url: &str,
+        bucket: &str,
+        key: &str,
+    ) -> io::Result<prost::bytes::Bytes> {
+        let client = Self::aws_client(url).await;
```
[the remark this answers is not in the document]
MitchTurner: We rely on the response from the rpc server to know the url. We would have to make some assumptions about the url in order to construct a client earlier.

This isn't performant though. So maybe we could only construct a new client if a (new?) url is provided.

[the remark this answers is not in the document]
MitchTurner: We could create a default client (with not url specified) that can be reused, but we handle each response in a stream and we'd have to clone the shared client into each thread of that stream:
```rust
            .then(|res| async move {
                let resp =
                    res.map_err(|e| io::Error::other(format!("RPC error: {:?}", e)))?;
                Self::convert_block_response(resp).await
            });
```

 We could wrap it in a `Arc` and clone and move it into each thread, but idk if we are getting much at that point. I don't think the aws client is particularly expensive to construct.

crates/client/src/client.rs
MitchTurner: _Copying from my convo with Cursor below:_

We handle each response in a stream and we'd have to clone the shared client into each thread of that stream:
```rust
            .then(|res| async move {
                let resp =
                    res.map_err(|e| io::Error::other(format!("RPC error: {:?}", e)))?;
                Self::convert_block_response(resp).await
            });
```
We could wrap it in a Arc and clone and move it into each thread, but idk if we are getting much at that point. I don't think the aws client is particularly expensive to construct.

MitchTurner: > tbh i don't think it's worth adding these dependencies if we can fetch using http

It sounds like we have a whole stack of AWS credential management setup and we will be reusing that with this new code. That's more complicated than just simple HTTP requests.

crates/client/src/client.rs
mchristopher: I think we should parse the HTTP response headers and check if the `Content-Encoding` header has been set. Then, if it's set to gzip, we ungzip it, otherwise we just assume it's not encoded.

This is similar to how reqwest works - https://github.com/seanmonstar/reqwest/blob/master/src/async_impl/client.rs#L1062

crates/client/src/client.rs
mchristopher: We need to include a basic S3 client (not just HTTP) since we want to support requestor pays and potentially other authenticated calls. We don't need a ton of logic, but we do need basic AWS SDK support for authenticated calls.

MitchTurner: I added a shared client. It looks like it's the intended use of the aws client, as it has an `Arc` inside and is `Clone`.

crates/fuel-core/src/service.rs
```diff
@@ -165,11 +168,21 @@ impl FuelService {
         );
         let bound_address = runner.shared.graph_ql.bound_address;
 
+        #[cfg(feature = "rpc")]
+        let rpc_address = runner
+            .shared
+            .config
+            .rpc_config
+            .clone()
+            .map(|config| config.addr);
```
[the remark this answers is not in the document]
MitchTurner: That's fine. We have two addresses here.

crates/client/src/client.rs
```diff
@@ -1620,3 +1678,199 @@ impl FuelClient {
...
+        url: Option<&String>,
+        bucket: &str,
+        key: &str,
+    ) -> io::Result<prost::bytes::Bytes> {
+        let client = if let Some(inner) = url {
+            Self::new_aws_client(Some(inner)).await
+        } else {
+            s3_client.ok_or(io::Error::other("No AWS client configured"))?
+        };
+        tracing::info!("getting block from bucket: {} with key {}", bucket, key);
```
mchristopher: I'd put this at debug - it's going to be very noisy if it's info

crates/client/src/client.rs
```diff
@@ -1620,3 +1678,199 @@ impl FuelClient {
...
+        bucket: &str,
+        key: &str,
+    ) -> io::Result<prost::bytes::Bytes> {
+        let client = if let Some(inner) = url {
+            Self::new_aws_client(Some(inner)).await
+        } else {
+            s3_client.ok_or(io::Error::other("No AWS client configured"))?
+        };
+        tracing::debug!("getting block from bucket: {} with key {}", bucket, key);
+        let req = client.get_object().bucket(bucket).key(key);
```
mchristopher: Need to add a check for requester pays here & add support. Something like:

```
let req = if requester_pays {
    req.request_payer(RequestPayer::Requester)
}
```

crates/client/src/client.rs
```diff
@@ -1620,3 +1675,201 @@ impl FuelClient {
...
+        s3_client: Option<AWSClient>,
+        url: Option<&String>,
+        bucket: &str,
+        key: &str,
+        requester_pays: bool,
+    ) -> io::Result<prost::bytes::Bytes> {
+        use aws_sdk_s3::types::RequestPayer;
+
+        let client = if let Some(inner) = url {
+            Self::new_aws_client(Some(inner)).await
```
xgreenx, line 1789: Response from `fuel-core` always contains the URL, so it means we will create an AWS client on each response, which is `async`, and will add an additional latency for the user. We should try to have the map of AWS clients which we can reuse(assuming we will have the same URL in most caes, it shouldn't eat a lot of memory). In this case we will initialzie it only once, and user will have lwoer latency.

It can be done in another PR.

xgreenx (approved):

PR #3172, opened 2026-01-05
fix: validate txpool transactions against next block height

### LINKED ISSUE #3169: When TxPool performs validation of the transactions it should use next block height
When TxPool performs validation of the transactions, it should use the next block height(last_height + 1)

its-saeed:
## Linked Issues/PRs
Closes #3169

## Description
When the TxPool validates transactions, it should use the next block height (`current_height + 1`) rather than the current block height. This is because transactions submitted to the pool will be executed in the next block, not the current one.

This fix is particularly important for transaction expiration validation. Previously, a transaction with an expiration height equal to the current block height would be incorrectly accepted, only to fail when actually executed in the next block (where it would already be expired).

## Checklist

### Before requesting review

### DIFF
--- a/crates/services/txpool_v2/src/service.rs
+++ b/crates/services/txpool_v2/src/service.rs
@@ -471,10 +471,7 @@ where
             let current_height = current_height_reader.read();
             // Transactions will be executed in the next block, so we validate
             // against the next block height.
-            let next_block_height = current_height.succ().expect(
-                "The block height should be less than `BlockHeight::MAX` \
-                during transaction validation",
-            );
+            let next_block_height = current_height.succ().unwrap_or(current_height);
 
             // TODO: This should be removed if the checked transactions
             //  can work with Arc in it


crates/services/txpool_v2/src/tests/tests_service.rs
```diff
@@ -618,3 +621,67 @@ async fn pending_pool__returns_error_after_timeout_for_transaction_that_spends_u
 
     service.stop_and_await().await.unwrap();
 }
+
+#[tokio::test]
+async fn insert__tx_with_expiration_at_next_block_height_is_accepted() {
+    // A transaction with expiration = current_height + 1 should be accepted
+    // because it will be executed in block current_height + 1, before it expires.
```
MitchTurner: nit: I'm of the opinion that comments should be kept to a minimum. I think your test name and variable names are plenty documentation :)

crates/services/txpool_v2/src/tests/tests_service.rs
```diff
@@ -618,3 +621,67 @@ async fn pending_pool__returns_error_after_timeout_for_transaction_that_spends_u
...
+        result.unwrap_err()
+    );
+
+    service.stop_and_await().await.unwrap();
+}
+
+#[tokio::test]
+async fn insert__tx_with_expiration_at_current_block_height_is_rejected() {
+    // A transaction with expiration = current_height should be rejected
+    // because it will be executed in block current_height + 1, which is after expiration.
```
MitchTurner: nit: Same here

crates/services/txpool_v2/src/tests/tests_service.rs
```diff
@@ -618,3 +621,67 @@ async fn pending_pool__returns_error_after_timeout_for_transaction_that_spends_u
 
     service.stop_and_await().await.unwrap();
 }
+
+#[tokio::test]
+async fn insert__tx_with_expiration_at_next_block_height_is_accepted() {
```
MitchTurner: nit: I think these two tests could be a little more general.

This PR is to address the off-by-one bug we were seeing. We could consider having prop tests that show that any tx with an expiration <= current block is rejected, and any tx with expiration > current block is accepted. That's the _actual_ behavior we are interested in :)

Could convert these tests to use `proptest` and test the whole space of possibilities

MitchTurner:
> In general this is probably the right course of action - but I'm concerned about the case where we have open block production w/ preconfs and the tx lands within the block production window of the current block and is executed in the same block.

I understood the bug as being that the tx pool was treating the "open block" height as being the block that was just added to the chain, not the one being built.

Voxelot:
> > In general this is probably the right course of action - but I'm concerned about the case where we have open block production w/ preconfs and the tx lands within the block production window of the current block and is executed in the same block.
> 
> I understood the bug as being that the tx pool was treating the "open block" height as being the block that was just added to the chain, not the one being built.

that makes sense - thanks for the clarification

MitchTurner (approved):
🔥

Issue #3171, opened 2026-01-01
Create AWS per URL only once


xgreenx:
Response from `fuel-core` always contains the URL, so it means we will create an AWS client on each response, which is `async`, and will add an additional latency for the user. We should try to have the map of AWS clients which we can reuse(assuming we will have the same URL in most caes, it shouldn't eat a lot of memory). In this case we will initialzie it only once, and user will have lwoer latency.

It can be done in another PR.

_Originally posted by @xgreenx in https://github.com/FuelLabs/fuel-core/pull/3168#discussion_r2656572975_

PR #3173, opened 2026-01-05
cache aws clients for all urls

MitchTurner:
## Linked Issues/PRs

Closes https://github.com/FuelLabs/fuel-core/issues/3171

## Description

We were assuming that _most_ rpc remote responses would not include a url for the s3 bucket. This was a wrong assumption. This change allows improved performance for when the responses include a URL.

### DIFF
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -283,7 +283,7 @@ pub struct FuelClient {
     #[cfg(feature = "rpc")]
     rpc_client: Option<ProtoBlockAggregatorClient<Channel>>,
     #[cfg(feature = "rpc")]
-    aws_client: Option<AWSClientManager>,
+    aws_client: AWSClientManager,
 }
 
 #[cfg(feature = "rpc")]
@@ -379,8 +379,7 @@ impl FuelClient {
         }
         let rpc_client = ProtoBlockAggregatorClient::connect(raw_rpc_url).await?;
         client.rpc_client = Some(rpc_client);
-        let aws_client_manager = AWSClientManager::new();
-        client.aws_client = Some(aws_client_manager);
+        client.aws_client = AWSClientManager::new();
         Ok(client)
     }
 
@@ -394,7 +393,7 @@ impl FuelClient {
             #[cfg(feature = "rpc")]
             rpc_client: None,
             #[cfg(feature = "rpc")]
-            aws_client: None,
+            aws_client: AWSClientManager::new(),
         })
     }
 }
@@ -1754,7 +1753,7 @@ impl FuelClient {
 
     async fn convert_block_response(
         resp: BlockResponse,
-        s3_client: Option<AWSClientManager>,
+        s3_client: AWSClientManager,
     ) -> io::Result<(fuel_core_types::blockchain::block::Block, Vec<Vec<Receipt>>)> {
         let payload = resp
             .payload
@@ -1783,18 +1782,15 @@ impl FuelClient {
                             endpoint,
                             requester_pays,
                         } = s3;
-                        let zipped_bytes = if let Some(inner_s3_client) = s3_client {
-                            Self::get_block_from_s3_bucket(
-                                inner_s3_client,
-                                &endpoint,
-                                &bucket,
-                                &key,
-                                requester_pays,
-                            )
-                            .await
-                        } else {
-                            Err(io::Error::other("S3 client not initialized"))
-                        }?;
+                        let zipped_bytes = Self::get_block_from_s3_bucket(
+                            s3_client,
+                            &endpoint,
+                            &bucket,
+                            &key,
+                            requester_pays,
+                        )
+                        .await?;
+
                         let block_bytes = Self::unzip_bytes(&zipped_bytes)?;
                         let block =
                             ProtoBlock::decode(block_bytes.as_slice()).map_err(|e| {


crates/client/src/client.rs
```diff
@@ -347,7 +385,9 @@ impl FuelClient {
         }
         let rpc_client = ProtoBlockAggregatorClient::connect(raw_rpc_url).await?;
         client.rpc_client = Some(rpc_client);
-        client.aws_client = Some(Self::new_aws_client(None).await);
+        let default = Self::new_aws_client(None).await;
+        let aws_client_manager = AWSClientManager::new(default);
+        client.aws_client = Some(aws_client_manager);
```
xgreenx: I think `AWSClientManager` should always exist, but it can be empty(without `default = Self::new_aws_client(None).await`). It is hard for the user to know upfront the actual endpoint before he receives first response.

crates/client/src/client.rs
```diff
@@ -1777,18 +1817,18 @@ impl FuelClient {
         }
     }
     async fn get_block_from_s3_bucket(
-        s3_client: Option<AWSClient>,
+        s3_client: Option<AWSClientManager>,
```
xgreenx:
```suggestion
        s3_client: AWSClientManager,
```

crates/client/src/client.rs
```diff
@@ -279,7 +283,41 @@ pub struct FuelClient {
...
+#[cfg(feature = "rpc")]
+impl AWSClientManager {
+    pub fn new(default: AWSClient) -> Self {
+        Self {
+            default,
+            specific: Arc::new(RwLock::new(HashMap::new())),
+        }
+    }
+
+    pub async fn get_client(&self, url: Option<&String>) -> Result<AWSClient, io::Error> {
```
xgreenx: I don't think that `url` ever should be `None` here. The core upper should return an error or fetch the URL on its own, but the code here already should operate with actual endpoint

MitchTurner: The thing is that the aws client doesn't expect any url. It knows how to resolve the url on its own. The point of specifying the url is to allow an override.

This is how a lot of the aws client is working, it uses a complex dependency chain to construct the request that is hidden from the user.

I don't want to require a url.

crates/client/src/client.rs
MitchTurner: The problem is with the old `with_urls` constructor for `FuelClient`. It is `sync`, but the `AWSClient` constructor is `async`

xgreenx: I am trying to say that you don't need `let default = Self::new_aws_client(None).await;` at all. You just create empty `AWSClientManager`. When someone will query data from RPC, you will create it inside of the `get_client` function

crates/client/src/client.rs
xgreenx: But you still pass the URL here:

image

And you know that you will set it from `fuel-core`

MitchTurner:
```rs
pub enum StorageMethod {
    ...
    S3 {
        ...
        #[clap(long = "endpoint_url", env)]
        endpoint_url: Option<String>,
        ...
    },
    ...
}
```
It's `Option` in `fuel-core` too :)

And if we leave this `None`, then what do we set the client as?

xgreenx: Why is it `Option` of `fuel-core` side? We don't know what is the configuration in the case of S3?

On the client I was expecting to return an error in the case if url is not set. How can we fetch data without known from where?

MitchTurner: I'm just following @mchristopher's guidance. He says that buckets are unique for all of S3 and the s3 client knows how to resolve it even with no url set.

xgreenx: Okay, if @mchristopher wants us to use `Option`, we can use it. But still, we don't need to create `default` AWS client during creation of the AWSClientManager. We can do it later, when user gives as an URL. So `map` can look like `HashMap<Option<String>, AWSClient>`

MitchTurner: Done

crates/client/src/client.rs
```diff
@@ -279,7 +283,35 @@ pub struct FuelClient {
     #[cfg(feature = "rpc")]
     rpc_client: Option<ProtoBlockAggregatorClient<Channel>>,
     #[cfg(feature = "rpc")]
-    aws_client: Option<AWSClient>,
+    aws_client: Option<AWSClientManager>,
```
xgreenx:
```suggestion
    aws_client: AWSClientManager,
```

I think we can use it now like this

MitchTurner: Woops. Missed that. I'm afk. I'll fix it when I'm home.

crates/client/src/client.rs
```diff
@@ -279,7 +283,35 @@ pub struct FuelClient {
...
+            return Ok(existing);
+        }
+        let client = FuelClient::new_aws_client(url).await;
+        let mut guard = self.specific.write().await;
+        let client = guard
+            .entry(url.clone())
+            .or_insert_with(|| client.clone())
+            .clone();
+        Ok(client)
+    }
```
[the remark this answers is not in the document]
MitchTurner: Yeah. This will only happen near the beginning of the first stream to use that url. All subsequent scenarios it will be read fine. I don't think this is a problem.

PR #3174, opened 2026-01-06
Fix dry_run to get the consensus parameter version of the old block

MitchTurner:
## Linked Issues/PRs

closes https://github.com/FuelLabs/fuel-core/issues/3175

## Description

The `dry_run` was always getting the current consensus parameter version, instead of the version for the specified height. This was resulting in undefined behavior.

## Checklist

### Before requesting review

### DIFF
--- a/.changes/fixed/3172.md
+++ b/.changes/fixed/3172.md
@@ -1,0 +1,1 @@
+Validate TxPool transactions against next block height instead of current block height.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4907,6 +4907,7 @@ dependencies = [
  "num-rational",
  "parking_lot",
  "petgraph",
+ "proptest",
  "rand 0.8.5",
  "tokio",
  "tokio-stream",
--- a/crates/services/txpool_v2/Cargo.toml
+++ b/crates/services/txpool_v2/Cargo.toml
@@ -36,6 +36,7 @@ fuel-core-storage = { workspace = true, features = ["std", "test-helpers"] }
 fuel-core-trace = { path = "../../trace" }
 fuel-core-txpool = { path = ".", features = ["test-helpers"] }
 mockall = { workspace = true }
+proptest = { workspace = true }
 rand = { workspace = true }
 tokio = { workspace = true, features = ["sync", "test-util"] }
 tokio-stream = { workspace = true }
--- a/crates/services/txpool_v2/src/service.rs
+++ b/crates/services/txpool_v2/src/service.rs
@@ -469,6 +469,9 @@ where
 
         let insert_transaction_thread_pool_op = move || {
             let current_height = current_height_reader.read();
+            // Transactions will be executed in the next block, so we validate
+            // against the next block height.
+            let next_block_height = current_height.succ().unwrap_or(current_height);
 
             // TODO: This should be removed if the checked transactions
             //  can work with Arc in it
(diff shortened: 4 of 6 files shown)


crates/services/producer/src/block_producer.rs
```diff
@@ -214,24 +226,29 @@ where
...
+            return Err(Error::CannotProduceGenesisBlock.into());
+        }
+        if height != next_height {
+            return Err(Error::MustProduceBlockWithExpectedHeight {
+                requested_height: height,
+                expected_height: next_height,
             }
-            .into())
+            .into());
         }
```
MitchTurner, line 245: I moved these checks here because we don't need the `height` param anymore.

However `produce_and_execute_block_transactions` is `pub` and I didn't know if we wanted this a breaking change.

We have the option of removing it completely since we can just rely on `latest_height`

crates/services/producer/src/block_producer/tests.rs
```diff
@@ -100,7 +100,7 @@ mod produce_and_execute_block_txpool {
         assert!(
             matches!(
                 err.downcast_ref::<Error>(),
-                Some(Error::BlockHeightShouldBeHigherThanPrevious { .. })
+                Some(Error::CannotProduceGenesisBlock)
```
rymnc, line 103: why did this behaviour change?

MitchTurner: I don't think the old error was the correct error to use. Take a look at the test name:
```
cant_produce_at_genesis_height
```

That's not a clear error to get (not that it matters since we should never see this error)

crates/services/producer/src/block_producer.rs
```diff
@@ -522,11 +541,14 @@ where
...
-        height: BlockHeight,
         block_time: Tai64,
         view: &ViewProvider::LatestView,
     ) -> anyhow::Result<PartialBlockHeader> {
-        let previous_block_info = self.previous_block_info(height, view)?;
+        let previous_block_info = self.latest_block_info(view)?;
+        let height = previous_block_info
+            .height
+            .succ()
+            .ok_or(Error::NoGenesisBlock)?;
```
xgreenx, line 551: Wrong error

crates/services/producer/src/block_producer.rs
```diff
@@ -82,6 +84,16 @@ pub enum Error {
         best: DaBlockHeight,
         previous_block: DaBlockHeight,
     },
+    #[display(fmt = "Attempting to produce genesis block")]
```
Dentosal, line 87:
```suggestion
    #[display(fmt = "Couldn't produce genesis block")]
```

crates/services/producer/src/block_producer.rs
```diff
@@ -214,24 +226,29 @@ where
...
         let view = self.view_provider.latest_view()?;
 
-        let header = self
-            .new_header_with_new_da_height(height, block_time, &view)
-            .await?;
-
         let latest_height = view.latest_height().ok_or(Error::NoGenesisBlock)?;
+        let next_height = latest_height
+            .succ()
+            .expect("Should always be able to increment the latest height");
```
xgreenx, line 234: We shouldn't panic in the production code

crates/services/producer/src/block_producer.rs
```diff
@@ -350,19 +367,23 @@ where
...
-        });
-
         let simulated_time = time.unwrap_or_else(|| {
             view.get_block(&latest_height)
                 .map(|block| block.header().time())
                 .unwrap_or(Tai64::UNIX_EPOCH)
         });
 
-        let header = self.new_header(simulated_height, simulated_time, &view)?;
+        let next_height = latest_height
```
xgreenx, line 376: We shouldn't panic in the production code

PR #3181, opened 2026-01-13
Add expensive op extension

MitchTurner:
## Description

This is a protection against DoS by spamming expensive requests that can block the threadpool. We will be migrating to a better, paginated request pattern, but that will be a breaking change.

The plan for now is add this change as a stop-gap, depricate this endpoint and move users to the new pattern.

## Checklist

### Before requesting review

### DIFF
--- a/crates/fuel-core/src/graphql_api/extensions/expensive_op_guard.rs
+++ b/crates/fuel-core/src/graphql_api/extensions/expensive_op_guard.rs
@@ -103,17 +103,16 @@ impl Extension for ExpensiveOpGuard {
             self.semaphore.available_permits(),
         );
 
-        match is_expensive {
-            true => {
-                self.expensive_execution(
-                    ctx,
-                    operation_name,
-                    next,
-                    expensive_root_field_count,
-                )
-                .await
-            }
-            false => next.run(ctx, operation_name).await,
+        if is_expensive {
+            self.expensive_execution(
+                ctx,
+                operation_name,
+                next,
+                expensive_root_field_count,
+            )
+            .await
+        } else {
+            next.run(ctx, operation_name).await
         }
     }
 }
@@ -146,7 +145,7 @@ impl ExpensiveOpGuard {
         let out = tokio::time::timeout(self.timeout, fut).await;
         tracing::warn!(
             "finished executing in {:?}ns, success: {:?}",
-            starting_time.elapsed().as_nanos(),
+            starting_time.elapsed(),
             out.is_ok(),
         );
 


crates/fuel-core/src/graphql_api/extensions/expensive_op_guard.rs
```diff
@@ -0,0 +1,112 @@
...
+#[async_trait::async_trait]
+impl Extension for ExpensiveOpGuard {
+    async fn execute(
+        &self,
+        ctx: &ExtensionContext<'_>,
+        operation_name: Option<&str>,
+        next: NextExecute<'_>,
+    ) -> Response {
+        let op = operation_name.clone().unwrap_or_default();
+        let is_expensive = self.expensive_op_names.iter().any(|name| op == name);
```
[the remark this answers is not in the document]
MitchTurner: Is that true if the server doesn't know about those operation names?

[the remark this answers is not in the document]
Voxelot: Yes I think cursor is right here, operation name is an optional user assigned alias for the query: https://graphql.org/learn/queries/#operation-type-and-name

Might have better luck using governer or similar directly on the async fn that runs the expensive op.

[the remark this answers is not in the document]
MitchTurner: How can we inspect the actual query sent, rather than the name assigned by the client? Because clearly we have the query someplace. Maybe the `resolve` method on the `trait Extension`?

It takes `ResolveInfo` with `QueryPathNode`
```
/// Parameters for `Extension::resolve_field_start`
pub struct ResolveInfo<'a> {
    /// Current path node, You can go through the entire path.
    pub path_node: &'a QueryPathNode<'a>,

    /// Parent type
    pub parent_type: &'a str,

    /// Current return type, is qualified name.
    pub return_type: &'a str,

    /// Current field name
    pub name: &'a str,

    /// Current field alias
    pub alias: Option<&'a str>,

    /// If `true` means the current field is for introspection.
    pub is_for_introspection: bool,

    /// Current field
    pub field: &'a Field,
}
```

Voxelot (dismissed):

bin/fuel-core/src/cli/run/graphql.rs
```diff
@@ -105,6 +105,12 @@ pub struct GraphQLArgs {
 
     #[clap(flatten)]
     pub costs: QueryCosts,
+
+    #[clap(long = "concurrent-full-block-requests", default_value = "10", env)]
+    pub concurrent_full_block_requests: usize,
+
+    #[clap(long = "full-block-request-timeout", default_value = "3s", env)]
+    pub full_block_request_timeout: humantime::Duration,
```
Dentosal, line 113: Might be nice if this was optional for local envs, but just setting a high limit should be fine

crates/fuel-core/src/graphql_api/extensions/expensive_op_guard.rs
```diff
@@ -0,0 +1,224 @@
...
+            "Executing operation: {:?}, expensive root field count: {:?}, expected root fields: {:?}, expensive: {:?}, timeout: {:?}, semaphore_size: {:?}",
+            operation_name,
+            expensive_root_field_count,
+            self.expensive_root_field_names,
+            is_expensive,
+            self.timeout,
+            self.semaphore.available_permits(),
+        );
+
+        match is_expensive {
```
Dentosal: nit: why not `if`?

crates/fuel-core/src/graphql_api/extensions/expensive_op_guard.rs
```diff
@@ -0,0 +1,224 @@
...
+            }
+        };
+
+        // Time bound (avoid request pile-ups)
+        let fut = next.run(ctx, operation_name);
+        let starting_time = tokio::time::Instant::now();
+        let out = tokio::time::timeout(self.timeout, fut).await;
+        tracing::warn!(
+            "finished executing in {:?}ns, success: {:?}",
+            starting_time.elapsed().as_nanos(),
```
Dentosal: we could probably just use the `Debug` impl of `Duration` here, it gives nice units by default

Dentosal (dismissed):

bin/fuel-core/src/cli/run/graphql.rs
MitchTurner: I looked into this, it makes the code a bit more complicated. Still worth considering, but not prioritizing now.

PR #3203, opened 2026-02-13
PoA Leader Lock for Block Production

MitchTurner:
## Description

The goal with this change is to allow multiple authority nodes to be live without causing contention on who should produce the next block.

Introduced the `LeaderLeasePort`:
```rs
#[async_trait::async_trait]
pub trait LeaderLeasePort: Send + Sync {
    async fn can_produce_block(&self, height: BlockHeight) -> anyhow::Result<bool>;
    async fn release_lease(&self) -> anyhow::Result<()>;
}
```

Where each block producer can check if the lease is available before producing a block. On shutdown, it will always release the lease as well.

The implementation is `RedisLeaderLeaseAdapter`, where we use a lease on a Redis server. The server can have its own ttl for the lease so if a block producer dies before releasing the lease, it can still be grabbed by the other producers :)

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/chain-config/src/lib.rs
+++ b/crates/chain-config/src/lib.rs
@@ -1,6 +1,6 @@
 #![deny(clippy::cast_possible_truncation)]
 #![deny(clippy::arithmetic_side_effects)]
-#![cfg_attr(not(test), deny(unused_crate_dependencies))]
+#![deny(unused_crate_dependencies)]
 #![deny(warnings)]
 
 extern crate alloc;
--- a/crates/fuel-core/src/service/adapters/consensus_module/poa.rs
+++ b/crates/fuel-core/src/service/adapters/consensus_module/poa.rs
@@ -456,21 +456,10 @@ impl Drop for RedisLeaderLeaseAdapter {
             return;
         }
 
-        let runtime = tokio::runtime::Builder::new_current_thread()
-            .enable_all()
-            .build();
-        match runtime {
-            Ok(runtime) => {
-                if runtime.block_on(release_future).is_err() {
-                    error!("Failed to release leader lease: timeout");
-                }
-            }
-            Err(err) => {
-                error!(
-                    "Failed to create runtime for leader lease release: {:?}",
-                    err
-                );
-            }
+        let result = futures::executor::block_on(release_future);
+
+        if let Err(err) = result {
+            tracing::error!("Failed to release leader lease: {err}");
         }
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4768,6 +4768,7 @@ dependencies = [
  "anyhow",
  "derive_more 0.99.20",
  "enum-iterator",
+ "fuel-core-storage",
  "fuel-core-types 0.47.1",
  "fuel-vm 0.65.0",
  "impl-tools",
--- a/crates/storage/Cargo.toml
+++ b/crates/storage/Cargo.toml
@@ -21,7 +21,7 @@ version = { workspace = true }
 default = ["std"]
 alloc = ["fuel-vm-private/alloc", "fuel-core-types/alloc"]
 std = ["alloc", "fuel-vm-private/std", "fuel-core-types/std"]
-test-helpers = ["std", "dep:mockall", "dep:rand"]
+test-helpers = ["dep:mockall", "dep:rand"]
 fault-proving = ["fuel-core-types/fault-proving"]
 
 [dependencies]
@@ -45,6 +45,7 @@ strum = { workspace = true }
 strum_macros = { workspace = true }
 
 [dev-dependencies]
+fuel-core-storage = { path = ".", features = ["test-helpers"] }
 fuel-core-types = { workspace = true, default-features = false, features = [
   "serde",
   "random",


bin/fuel-core/src/cli/run/consensus.rs
```diff
@@ -84,6 +96,51 @@ struct Open {
...
+        } = value;
+        if enabled {
+            Some(RedisLeaderLockConfig {
+                redis_url: redis_url
+                    .expect("`redis_url` is required when `enabled` is true"),
+                lease_key: lease_key
+                    .expect("`lease_key` is required when `enabled` is true"),
+                lease_ttl: lease_ttl
+                    .expect("`lease_ttl` is required when `enabled` is true")
+                    .into(),
```
xgreenx:
```suggestion
                redis_url: redis_url?,
                lease_key: lease_key?,
                lease_ttl: lease_ttl?.into(),
```

bin/fuel-core/src/cli/run/consensus.rs
```diff
@@ -15,6 +19,14 @@ pub struct PoATriggerArgs {
     interval: Interval,
     #[clap(flatten)]
     open: Open,
+    #[clap(flatten)]
+    leader_lock: LeaderLock,
+}
+
+impl PoATriggerArgs {
+    pub fn leader_lock(self) -> Option<RedisLeaderLockConfig> {
```
xgreenx:
```suggestion
    pub fn leader_lock(&self) -> Option<RedisLeaderLockConfig> {
```

bin/fuel-core/src/cli/run/consensus.rs
```diff
@@ -15,6 +19,14 @@ pub struct PoATriggerArgs {
...
+        self.leader_lock.into()
```
xgreenx: Can we move code from the `From` to here please?

crates/fuel-core/Cargo.toml
```diff
@@ -95,6 +95,7 @@ num_cpus = { version = "1.16.0", optional = true }
 parking_lot = { workspace = true }
 postcard = { workspace = true }
 rand = { workspace = true }
+redis = { version = "0.27", features = ["tokio-comp"] }
```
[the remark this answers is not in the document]
xgreenx: Agree with cursor, let's make it optional

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -562,6 +569,32 @@ where
...
+        match self.can_produce_next_block().await {
+            Ok(true) => self.handle_normal_block_production(deadline).await,
+            Ok(false) => {
+                tokio::time::sleep(Duration::from_millis(10)).await;
+                TaskNextAction::Continue
+            }
+            Err(err) => {
+                tokio::time::sleep(Duration::from_millis(10)).await;
+                TaskNextAction::ErrorContinue(err)
+            }
```
xgreenx: Why do we need to wait for 10 milliseconds? Looks like something is wrong with this logic. We shouldn't call `handle_triggered_block_production` before next production has come

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -675,20 +709,25 @@ where
...
-                self.handle_normal_block_production(deadline).await
+                self.handle_triggered_block_production(deadline).await
             }
         }
     }
 
     async fn shutdown(self) -> anyhow::Result<()> {
         tracing::info!("PoA MainTask shutting down");
+        if let Some(leader_lease_port) = &self.leader_lease_port
+            && let Err(err) = leader_lease_port.release_lease().await
```
xgreenx, line 717: I think we also should implement `Drop` trait to ensure that it will be called

tests/tests/lib.rs
```diff
@@ -40,6 +40,10 @@ mod graphql_extensions;
 mod health;
 #[cfg(not(feature = "only-p2p"))]
 mod helpers;
+#[cfg(feature = "leader_lock")]
+#[cfg(not(feature = "not_leader_lock"))]
```
xgreenx, line 44: Do we really need these features? Why is it a problem mto run theses tests? Looks like they should finish fast enough

tests/tests/leader_lock.rs
```diff
@@ -0,0 +1,401 @@
...
+
+impl RedisTestServer {
+    fn spawn() -> Self {
+        let socket =
+            TcpListener::bind(SocketAddrV4::new(std::net::Ipv4Addr::LOCALHOST, 0))
+                .expect("Should bind an ephemeral port");
+        let port = socket.local_addr().expect("Should get local addr").port();
+        drop(socket);
+
+        let child = Command::new("redis-server")
```
xgreenx: Do we need to have it installed locally in order to run it?

MitchTurner: yes. That's why I've added a separate integ test in ci for it.

crates/services/consensus_module/poa/src/service.rs
MitchTurner: But how does the producer know it's _not_ the producer for the current block? And if another producer dies/is shut down mid-production, should we not start producing immediately?

Do you not think there should be any loop?

crates/services/consensus_module/poa/src/service.rs
MitchTurner: It's `async` so I don't know what that would look like. I can investigate if there are sync alternatives to all the parts.

tests/tests/lib.rs
MitchTurner: If we want it to run in `--all-features`, then we need to install `redis-server` in that ci image.

crates/fuel-core/Cargo.toml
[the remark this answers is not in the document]
MitchTurner: It's configured at runtime. You want me to add a feature as well?

mchristopher:
Do we want to implement Redlock per https://redis.io/docs/latest/develop/clients/patterns/distributed-locks/? That's a more robust locking system which relies on multiple redis instances (instead of a single instance, which is how this code would handle it).

.changes/breaking/3204.md
```diff
@@ -0,0 +1 @@
+bump rustc from 1.90.0 to 1.93.1
```
xgreenx: We will create a minor release 0.47.2, so we shouldn't increase the Rust version

bin/fuel-core/src/cli/run.rs
```diff
@@ -463,6 +463,7 @@ impl Command {
         #[cfg(feature = "rpc")]
         let rpc_config = rpc_args.map(|args| args.into_config());
 
+        let leader_lock = poa_trigger.clone().leader_lock()?;
```
xgreenx:
```suggestion
        let leader_lock = poa_trigger.leader_lock()?;
```

crates/fuel-core/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 #![deny(clippy::arithmetic_side_effects)]
 #![deny(clippy::cast_possible_truncation)]
-#![deny(unused_crate_dependencies)]
```
xgreenx, line 3: Why do we need to change it? Looks like you only added `redis`, which you use in the main code.

If you don't use in this crate, then it should be in `dev-deps`

crates/services/consensus_module/poa/src/service.rs
xgreenx: I think in the case of `Ok(false)`, we should just wait for the deadline to pass. It is fine if we skip one block in time, but the loop will not be hot.

In the case of `Err(err)`, I am fine with sleeping, but not for 10 milliseconds. Maybe for 100ms. Or maybe it should be configurable

crates/services/consensus_module/poa/src/service.rs
xgreenx: You still can do `now_or_never()` which will start request to the server. Or you can have a loop that will do that. Or you can use `futures::task::spawn`. Many options to run it form the `drop`.

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -36,21 +37,366 @@ use fuel_core_types::{
...
+                false
+            }
+        }
+    }
+
+    fn quorum_reached(&self, success_count: usize) -> bool {
+        success_count >= self.quorum
+    }
+
+    fn calculate_remaining_validity_millis(&self, elapsed_millis: u64) -> u64 {
```
xgreenx, line 268: Looks like with the current implementation, we need to have ttls higher than production time(more than one second). Because we only renew the lease when we try to produce next block

MitchTurner: Yes. So that's a risk on configuration, but I'm planning on at least 2s.

crates/chain-config/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 #![deny(clippy::cast_possible_truncation)]
 #![deny(clippy::arithmetic_side_effects)]
-#![cfg_attr(not(test), deny(unused_crate_dependencies))]
```
MitchTurner, line 3: I thought I removed this... Was there another location?

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -456,21 +456,10 @@ impl Drop for RedisLeaderLeaseAdapter {
...
-                    error!("Failed to release leader lease: timeout");
-                }
-            }
-            Err(err) => {
-                error!(
-                    "Failed to create runtime for leader lease release: {:?}",
-                    err
-                );
-            }
+        let result = futures::executor::block_on(release_future);
```
MitchTurner, line 459: hmmmm I had this before and it was panicking.

xgreenx: Now because you have `tokio::current_thread` it fixed the panic for tokio runtime

PR #3244, opened 2026-04-01
docs: fix stale Ignition checkout example in README

ming1523:
## Summary
- fix the stale `git checkout` example in the Ignition node instructions
- align the example with the current Fuel Ignition version table in the README
- fix the section title grammar while touching the same block

## Validation
- verified the README now points to `v0.47.1`, which matches the Fuel Ignition version listed in the table above

This is a docs-only change. The `no changelog` label seems appropriate if maintainers agree.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ rustup target add wasm32-unknown-unknown
 ```
 
 
-## Running a Ignition node
+## Running an Ignition node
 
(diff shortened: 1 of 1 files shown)


xgreenx:
Thank you for your PR. You are right that we need to actualzie readme in many palces, so instead of many small PRs, we will just have one with all actualization

Issue #975, opened 2023-02-02
Unify defaults for config


freesig:
The cli run:
https://github.com/FuelLabs/fuel-core/blob/3a88bececec821bf8ebbc33d001b84f34968a191/bin/fuel-core/src/cli/run.rs#L62
has a set of defaults that duplicate and are sometimes different to the actual defaults for the real config. We should unify these values so there is a single source of truth.

MitchTurner:
For example.
Both `Config` and `fuel_core_txpool::Config` have a `utxo_validation` field. But `Config` also contains the `fuel_core_txpool::Config`! WE should just have a single config that can be broken into the sub-configs.

PR #3246, opened 2026-04-01
refactor: share GraphQL default configuration values

ming1523:
## Summary
- add shared GraphQL default constants in `graphql_api.rs`
- reuse those defaults from the GraphQL CLI args
- align the local test-node GraphQL config with the same shared values
- add a CLI parsing test to keep the shared defaults in sync

## Why
This is a focused follow-up for `#975`. The GraphQL defaults were duplicated across the CLI layer and the local-node config helper, and a few values had already drifted apart. This change narrows the fix to the GraphQL surface first so the default source is easier to maintain and test.

## Validation
- added a test that checks the parsed CLI defaults against the shared GraphQL defaults
- manually verified the local-node GraphQL config now uses the same shared constants

Ref #975

This is a small refactor/docs-of-code style change. The `no changelog` label seems appropriate if maintainers agree.


ming1523:
Follow-up update: I narrowed the scope of this PR to avoid changing the `local_node` GraphQL defaults in `service/config.rs`.

The branch now keeps the test-helper/local-node values unchanged and only centralizes the shared GraphQL defaults used by the CLI surface, along with the CLI sync test. That should remove the behavioral risk called out in the bot summary while keeping the default-source cleanup focused.

ming1523:
Follow-up note on the earlier Bugbot review:

The two timeout/body-limit concerns were reported against an earlier diff that temporarily changed `crates/fuel-core/src/service/config.rs`.

The current head no longer modifies `service/config.rs`. The local-node/test-helper GraphQL defaults remain unchanged there, including:
- `request_body_bytes_limit = 16 * 1024 * 1024`
- `api_request_timeout = 60s`

At this point the PR is scoped back down to sharing the CLI-facing GraphQL defaults and keeping the CLI default parsing test in sync with those shared constants.

xgreenx:
Thank you for your PR, but it is not clear what the benefits of having values in constants vs CLI. I am closing PR for now, but if you think there is a good reason, we can discuss it.

PR #3247, opened 2026-04-01
test: add service-level txpool integrity checks

ming1523:
## Summary
- add a test-only `SharedState::assert_integrity` helper for txpool service tests
- route the helper through a test-only pool read request handled by the pool worker
- use the helper in service tests, including an empty-pool case

## Why
This is a focused follow-up for `#2581`. The txpool already has integrity assertions in lower-level test helpers, but service-level tests did not have a direct way to trigger the same consistency check through the service surface. This adds that missing hook without changing runtime behavior.

## Validation
- the new helper is gated behind `#[cfg(test)]`
- service-level tests now exercise integrity checks through the shared service interface

Ref #2581

This is a test-only change. The `no changelog` label seems appropriate if maintainers agree.

### DIFF
--- a/crates/services/txpool_v2/src/pool_worker.rs
+++ b/crates/services/txpool_v2/src/pool_worker.rs
@@ -19,6 +19,8 @@ use std::{
     sync::Arc,
     time::SystemTime,
 };
+#[cfg(test)]
+use std::collections::HashSet;
 use tokio::{
     sync::{
         broadcast,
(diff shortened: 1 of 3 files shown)


xgreenx:
Thank you for your PR, but it is not clear what the benefits of doing it. I am closing PR for now, but if you think there is a good reason, we can discuss it.

Issue #1662, opened 2024-02-12
Create users friendly logs


xgreenx:
Deliverables:

- [ ] Add some standards to the Engineering Style Guide
- [ ] Logs for relayer block logs are not clearly from L1. We should also highlight L2 blocks that way.
- [ ] Logs for P2P connections
- [ ] Logs for Gas Price Service/Updater code

MitchTurner:
This could be a whole feature, depending on how we choose to go about it.

MitchTurner:
One major place is the current state of P2P, like if we aren't connected, it would be good to make sure it is clear to the user.

The point of this is to make it easier to solve problems more quickly, especially internally.

MitchTurner:
One thing we can do is have everyone on the team spin up a node locally and catalog any issues you have with that experience. Dog food the process.

MitchTurner:
We can also make the consumption of our logs more thorough, e.g. take advantage of our structured logging.

MitchTurner:
I'm not a fan of this issue because it's not actionable. We need to break this down into specific tasks.

PR #3249, opened 2026-04-01
chore: improve gas price service log context

ming1523:
## Summary
- replace generic gas price service debug logs with more contextual messages
- include L2 block height / gas price information when processing L2 blocks
- include DA bundle ids, L2 block ranges, and DA cost fields when buffering and applying DA block costs
- log recorded height and next gas price updates with explicit fields

## Why
This is a small follow-up for `#1662`. The gas price service already emits logs, but several of them are very generic (`Received ...`, `Updating ...`) or dump full debug payloads. This change keeps the same behavior while making the log stream more useful for operators and developers.

## Validation
- log-only change in `crates/services/gas_price_service/src/v1/service.rs`
- no behavior changes to gas price calculation or persistence logic

Ref #1662

This change is operational/logging oriented. The `no changelog` label seems appropriate if maintainers agree.

### DIFF
--- a/crates/services/gas_price_service/src/v1/service.rs
+++ b/crates/services/gas_price_service/src/v1/service.rs
@@ -184,11 +184,28 @@ where
         self.sync_notifier
             .send(crate::sync_state::SyncState::NotSynced)
             .ok();
-        tracing::debug!("Received L2 block result: {:?}", l2_block_res);
+        match &l2_block_res {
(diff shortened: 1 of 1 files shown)


xgreenx:
Thank you for your PR, but these debug logs are not used too often, and I think we will even remove them, to reduce noise

PR #3251, opened 2026-04-01
Improve Leader Lock Code

MitchTurner:
## Linked Issues/PRs

Followup to feedback in https://github.com/FuelLabs/fuel-core/pull/3241

## Description

- [x] Do we need to update the timestamp of the block as well and re-calculate Instant? Liek we do in the case of SynState with the network ([8081c81](https://github.com/FuelLabs/fuel-core/pull/3251/commits/8081c81d70e7267e6a3fb7141acb06ce16930e17))
- [x] The same here, I think we should update Isntant accordingly to ensure that next block will be produced correctly ([8081c81](https://github.com/FuelLabs/fuel-core/pull/3251/commits/8081c81d70e7267e6a3fb7141acb06ce16930e17))
- [ ] Have you tried to build a table for these metrics to ensure that they work?
- [x] I think we coudl be more descriptive what was implemented=D Plus, we have a lot of docs related files, so we also can mention what information do we have there and how reader can consume its content ([eedec99](https://github.com/FuelLabs/fuel-core/pull/3251/commits/eedec99e88d005b1c6015a5ed0e425ed20d0f9d8),  [85cb842](https://github.com/FuelLabs/fuel-core/pull/3251/commits/85cb8429b39020773c883201876ea1f4c8b6df06))
- [x] Do we handle 0 properly?
- [ ] This await can be really long, in the case when we have multiple redis nodes, huge blocks. Especially if we need to load multiple blocks. Maybe we should consider fetching just IDs, and for each ID, we can get them from Redis by ID.
- [x] This is a really long function, with a lot of logic. Are you sure that it is not possible to write in a smaller way with suing less iteratators?=) Maybe we could use some data structure to get properties which we are looking for. ([9734a39](https://github.com/FuelLabs/fuel-core/pull/3251/commits/9734a39d3ed664f564b7a1006b72f91c025ae52a))
- [ ] This is a very long adapter code; usually, we try to keep them smaller. Maybe we should consider factoring this adapter into smaller components. Or at least extract pure redis interactions into its own module. Because we have a lot of simialrt code in order to reach quorume, and I suppose it could be hidden behind the traits

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/fuel-core/src/service/adapters/consensus_module/poa.rs
+++ b/crates/fuel-core/src/service/adapters/consensus_module/poa.rs
@@ -131,23 +131,36 @@ impl Clone for RedisNode {
 /// identify the true block producer of any given block. This is done by iterating over each epochs
 /// and finding the highest epoch with a quorum of blocks produced.
 pub struct RedisLeaderLeaseAdapter {
-    /// list of all redis nodes to connect to
+    /// Redis nodes that participate in leader election and block reconciliation
     redis_nodes: Vec<RedisNode>,
-    /// number of nodes for which a lock needs to be held to consider the producer legitimate
+    /// Minimum number of Redis nodes that must agree for ownership or writes to be accepted
     quorum: usize,
+    /// Redis key that stores the current lease owner id with a TTL
     lease_key: String,
+    /// Redis key that stores the monotonically increasing fencing token
     epoch_key: String,
+    /// Redis stream key that stores published blocks for reconciliation
     block_stream_key: String,
-    lease_owner_token: String,
-    drop_release_guard: std::sync::Arc<()>,
+    /// Unique id for this adapter instance, written into the lease key when it acquires leadership
+    lease_owner_id: String,
+    /// Locally cached fencing token currently used for block writes.
     current_epoch_token: std::sync::Arc<std::sync::Mutex<Option<u64>>>,
+    /// Lease TTL applied to the Redis lease key and renewed on successful writes.
     lease_ttl_millis: u64,
+    /// Safety margin subtracted from the nominal TTL when evaluating remaining lease validity.
     lease_drift_millis: u64,
+    /// Timeout used for individual Redis node operations.
     node_timeout: Duration,
-    retry_delay_millis: u64,
-    max_retry_delay_offset_millis: u64,
-    max_attempts: usize,
+    /// Approximate upper bound for the Redis block stream and per-round reconciliation reads.
     stream_max_len: u32,
+    /// Base delay before retrying lease acquisition after a failed attempt.
+    retry_delay_millis: u64,
+    /// Random jitter added to the retry delay to reduce election collisions.
+    max_retry_delay_offset_millis: u64,
+    /// Maximum number of lease acquisition attempts per election round.
+    max_attempts: usize,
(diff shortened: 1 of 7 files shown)


MitchTurner:
> Have you tried to build a table for these metrics to ensure that they work?

@Voxelot added these. I think he can give more context.

MitchTurner:
> Do we handle 0 properly?

Yes, should be safe:
```rs
   fn calculate_quorum(redis_nodes_len: usize, quorum_disruption_budget: u32) -> usize {
        let majority = redis_nodes_len
            .checked_div(2)
            .unwrap_or(0)
            .saturating_add(1);
        let disruption_budget = usize::try_from(quorum_disruption_budget).unwrap_or(0);
        majority
            .saturating_add(disruption_budget)
            .min(redis_nodes_len)
    }
```

MitchTurner:
>  This await can be really long, in the case when we have multiple redis nodes, huge blocks. Especially if we need to load multiple blocks. Maybe we should consider fetching just IDs, and for each ID, we can get them from Redis by ID.

The idea was we have pagination to minimize this time... but might not be enough as block size gets bigger?

MitchTurner:
> I think we coudl be more descriptive what was implemented=D Plus, we have a lot of docs related files, so we also can mention what information do we have there and how reader can consume its content 

Added documentation to the adapter here: [eedec99](https://github.com/FuelLabs/fuel-core/pull/3251/commits/eedec99e88d005b1c6015a5ed0e425ed20d0f9d8)

Happy to add more.

Voxelot:
> > Have you tried to build a table for these metrics to ensure that they work?
> 
> @Voxelot added these. I think he can give more context.

I can confirm using these metrics when debugging the failover behavior in devnet and that they worked. I'm not sure what is meant by building a table? Like writing to rocksdb as well as metrics?

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -118,44 +118,69 @@ impl Clone for RedisNode {
...
+    /// Ensures only the last live clone attempts lease release during drop.
+    drop_release_guard: std::sync::Arc<()>,
 }
 
 impl Clone for RedisLeaderLeaseAdapter {
     fn clone(&self) -> Self {
         Self {
             redis_nodes: self.redis_nodes.clone(),
             quorum: self.quorum,
-            quorum_disruption_budget: self.quorum_disruption_budget,
```
Voxelot, line 145: i think we might be using this setting to ensure we have more than n/2 + 1 threshold in some situations. Will it still be possible to require `n/2 + m` without this quorum disruption budget setting?

Voxelot:
> > This await can be really long, in the case when we have multiple redis nodes, huge blocks. Especially if we need to load multiple blocks. Maybe we should consider fetching just IDs, and for each ID, we can get them from Redis by ID.
> 
> The idea was we have pagination to minimize this time... but might not be enough as block size gets bigger?

This is partially mitigated by using local block height as a cursor. However, if there are many conflicting blocks in redis, agree that it could be potentially lighter weight to only fetch and compare block ids first to determine the canonical path, before fetching the actual data. In that case we could also limit the data fetch to be from just one redis node, instead of fetching multiple copies of the data from all of them at once.

PR #3252, opened 2026-04-02
docs: remove stale Kubernetes manifest reference

ming1523:
## Summary
- remove the stale `deployment/fuel-core.yml` example from the root README
- clarify that the repository currently ships container build assets under `deployment/`, but not a checked-in Kubernetes manifest
- point readers to `deployment/README.md` and the Fuel node operator docs instead

## Why
The current README still tells readers to run `kubectl create -f deployment/fuel-core.yml`, but that file is not present in the repository. Replacing the stale command with pointers to the current deployment assets and operator docs makes the entrypoint more accurate for contributors and operators.

## Validation
- verified that `deployment/fuel-core.yml` does not exist in the current tree
- verified that `deployment/README.md` exists and is the repository-local deployment entrypoint
- docs-only change, no runtime behavior changed

This is a docs-only change. A no-changelog path seems appropriate if maintainers agree.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,13 +218,14 @@ docker build -t fuel-core . -f deployment/Dockerfile
 
 # Delete Docker Image
 docker image rm fuel-core
-
-# Create Kubernetes Volume, Deployment & Service
-kubectl create -f deployment/fuel-core.yml
-
-# Delete Kubernetes Volume, Deployment & Service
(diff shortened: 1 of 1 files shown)


xgreenx:
Thank you for your PR, we will actualize readme file in a separte PR just dedicated for that, since we have many outdated inforamtion in there

PR #3255, opened 2026-04-03
feat: add support for http endpoints and dynamic decompression

mchristopher:
## Description

This PR adds code support for the existing protobuf HTTP-based remote RPC block accessor. This allows the use of CDNs or other HTTP-based URLs, while still publishing blocks to S3. It also cleans up decompression support in the client, in order to more correctly support HTTP standards.

## Checklist

### Before requesting review

- [ ] Someone else?


crates/client/Cargo.toml
```diff
@@ -52,6 +52,8 @@ hyper-rustls = { version = "0.24", features = ["webpki-tokio"], optional = true
 itertools = { workspace = true }
 postcard = { workspace = true, optional = true }
 prost = { workspace = true, optional = true }
+# Inherits workspace reqwest; do not enable gzip/brotli/deflate/zstd here—`FuelClient` uses a
+# `Client::builder().no_*()` for remote block objects and manual `Content-Encoding` handling.
```
MitchTurner: so if we enable these features it will break?

crates/client/src/client.rs
```diff
@@ -1929,17 +1993,129 @@ impl FuelClient {
...
+    };
+    use std::io::{
+        Cursor,
+        Read,
+    };
+
+    /// Decodes the object body using a `Content-Encoding` value from HTTP headers or S3 object
+    /// metadata. Missing or empty means identity (raw protobuf bytes). Codings are listed in
+    /// application order; this reverses to decode from the outermost coding inward (RFC 9110).
+    pub fn decode_body_by_content_encoding(
```
MitchTurner: Is this a public function. I see it's `pub`, but like does the user call this or just use the client?

You've added tests for this method, but it's generally considered best practice to only test the public interfaces in case internal implementation changes.

crates/client/Cargo.toml
mchristopher: Ah, good point. Claude added this comment & I thought I removed it before I committed, but I guess it slipped through.

No - there is no risk here. The code handles this case already, so it's a non-issue.

crates/client/src/client.rs
mchristopher: Fair, moved it back into the main impl so it's not exposed as a module (the mod required a pub function).

MitchTurner: Is this method called on the response though? I'm curious what the use case is for this method on the client? Like, wouldn't it just get decoded as part of the response?

mchristopher: Yes, it's called with the response. It's so that the response from the server is treated like a normal HTTP response, and uses the HTTP response headers to decode the payload instead of assuming the payload is in a particular format. This brings the client in-line with HTTP standards, and gives us more flexibility on the server side in the future.

xgreenx (approved):

xgreenx (dismissed):

xgreenx (dismissed):

crates/client/src/client.rs
```diff
@@ -2069,6 +2204,42 @@ mod tests {
         );
     }
 
+    #[cfg(feature = "rpc")]
+    mod remote_block_object_decode_tests {
```
MitchTurner: I still think these tests should be done with the external interfaces. Specifically in integration tests using `localstack`

MitchTurner: We can create a tech-debt item for this... but I'd more comfortable in the feature if we are actually testing against something real...

MitchTurner (dismissed):

MitchTurner (approved):

PR #3264, opened 2026-04-14
Rollback unsuccessful preconfs in the mempool

### LINKED ISSUE #3098: Rollback unsuccessful preconfs in the mempool
When the mempool receives a pre-confirmation, it optimistically marks the inputs as "spent" in it's local cache. 

Currently there is no mechanism to rollback failed preconfs in the mempool, apart from a full restart of the node. This means when a preconfirmation doesn't finalize, those inputs remain "stuck", and users are prevented from using those inputs until the mempool is reset.

This has been observed nearly anytime the block producer is restarted, we find there are stuck utxos in the RPC mempools.

1. We need to investigate why restarting the block producer causes so many dropped transactions from a pre-conf perspective. Ideally if there is graceful shutdown, the in-flight blocks / transactions get finalized before the process exits.
2. We need a mechanism to automatically clear these cached utxos in the mempool without manual intervention (i.e. rebooting the mempool), either via TTL or some other reconciliation process that detects if a transaction did not finalize. Ie. any cached preconf data older than 1 block ago should be discarded.

Dentosal:
Closes #3098.

## Problem

When a block producer sends preconfirmation updates, sentry nodes optimistically treat the included transactions as committed, removing them from the mempool and marking their inputs as spent. If the producer crashes and re-produces a block at the same height without those transactions, the mempool is left in a stale state: inputs stay marked as spent and outputs linger in `extracted_outputs`, preventing re-submission of rolled-back transactions and causing dependents to reference non-existent UTXOs.

## Solution

This PR makes preconfirmed transactions tentative until the canonical block at their height is imported. On import, preconfirmed txs present in the block are confirmed and their tracking is cleared; those absent are rolled back by restoring inputs, purging dependents, and emitting `SqueezedOut` notifications. It also adds integration tests: re-insertion after rollback, dependent eviction, normal confirmation, and stale-height cleanup.

## Checklist

### Before requesting review

### DIFF
--- a/crates/services/txpool_v2/src/collision_manager/basic.rs
+++ b/crates/services/txpool_v2/src/collision_manager/basic.rs
@@ -81,6 +81,7 @@ impl<StorageIndex> BasicCollisionManager<StorageIndex> {
         self.messages_spenders.is_empty()
             && self.coins_spenders.is_empty()
             && self.contracts_creators.is_empty()
+            && self.contract_users.is_empty()
             && self.blobs_users.is_empty()
     }
 
@@ -93,6 +94,7 @@ impl<StorageIndex> BasicCollisionManager<StorageIndex> {
         let mut message_spenders = HashMap::new();
         let mut coins_spenders = BTreeMap::new();
         let mut contracts_creators = HashMap::new();
+        let mut contract_users: HashMap<ContractId, Vec<TxId>> = HashMap::new();
         let mut blobs_users = HashMap::new();
         for tx in expected_txs {
             if let PoolTransaction::Blob(checked_tx, _) = tx.deref() {
@@ -115,7 +117,12 @@ impl<StorageIndex> BasicCollisionManager<StorageIndex> {
                     }) => {
                         message_spenders.insert(*nonce, tx.id());
                     }
-                    Input::Contract { .. } => {}
+                    Input::Contract(ContractInput { contract_id, .. }) => {
+                        contract_users
+                            .entry(*contract_id)
+                            .or_default()
+                            .push(tx.id());
+                    }
                 }
             }
             for output in tx.outputs() {
@@ -156,6 +163,26 @@ impl<StorageIndex> BasicCollisionManager<StorageIndex> {
             contracts_creators.is_empty(),
             "Some contract creators are missing from the collision manager: {:?}",
             contracts_creators
+        );
+        for (contract_id, users) in &self.contract_users {
+            let expected = contract_users.remove(contract_id).unwrap_or_else(|| panic!(
+                "A contract ({}) user list is present on the collision manager that shouldn't be there.",
+                contract_id
+            ));
+            let mut actual_sorted = users.clone();
+            actual_sorted.sort();
+            let mut expected_sorted = expected;
+            expected_sorted.sort();
+            assert_eq!(
+                actual_sorted, expected_sorted,
+                "contract_users mismatch for contract {}",
+                contract_id
+            );
+        }
+        assert!(
+            contract_users.is_empty(),
+            "Some contract users are missing from the collision manager: {:?}",
+            contract_users
         );
     }
 }
--- a/crates/services/txpool_v2/src/collision_manager/basic.rs
+++ b/crates/services/txpool_v2/src/collision_manager/basic.rs
@@ -21,6 +21,7 @@ use fuel_core_types::{
                 CoinPredicate,
                 CoinSigned,
             },
+            contract::Contract as ContractInput,
             message::{
                 MessageCoinPredicate,
                 MessageCoinSigned,
@@ -57,6 +58,9 @@ pub struct BasicCollisionManager<StorageIndex> {
     coins_spenders: BTreeMap<UtxoId, StorageIndex>,
     /// Contract -> Transaction that currently create the contract
     contracts_creators: HashMap<ContractId, StorageIndex>,
+    /// Contract -> Transactions (by TxId) that currently use the contract as an input.
+    /// Symmetric to `contracts_creators`; used to evict dependents during rollback.
+    contract_users: HashMap<ContractId, Vec<TxId>>,
     /// Blob -> Transaction that currently create the blob
     blobs_users: HashMap<BlobId, StorageIndex>,
 }
@@ -67,6 +71,7 @@ impl<StorageIndex> BasicCollisionManager<StorageIndex> {
             messages_spenders: HashMap::new(),
             coins_spenders: BTreeMap::new(),
             contracts_creators: HashMap::new(),
+            contract_users: HashMap::new(),
             blobs_users: HashMap::new(),
         }
     }
@@ -174,6 +179,17 @@ where
             .collect()
     }
 
+    fn get_contract_users(&self, contract_id: &ContractId) -> Vec<TxId> {
+        self.contract_users
+            .get(contract_id)
+            .cloned()
+            .unwrap_or_default()
+    }
+
+    fn contract_created_in_pool(&self, contract_id: &ContractId) -> bool {
+        self.contracts_creators.contains_key(contract_id)
+    }
+
     fn find_collisions(
         &self,
         transaction: &PoolTransaction,
@@ -248,6 +264,7 @@ where
             let blob_id = checked_tx.transaction().blob_id();
             self.blobs_users.insert(*blob_id, storage_id);
         }
+        let tx_id = store_entry.transaction.id();
         for input in store_entry.transaction.inputs() {
             match input {
                 Input::CoinSigned(CoinSigned { utxo_id, .. })
@@ -262,7 +279,12 @@ where
                     // insert message
                     self.messages_spenders.insert(*nonce, storage_id);
                 }
-                _ => {}
+                Input::Contract(ContractInput { contract_id, .. }) => {
+                    self.contract_users
+                        .entry(*contract_id)
+                        .or_default()
+                        .push(tx_id);
+                }
             }
         }
         for output in store_entry.transaction.outputs().iter() {
@@ -284,6 +306,7 @@ where
             let blob_id = checked_tx.transaction().blob_id();
             self.blobs_users.remove(blob_id);
         }
+        let tx_id = transaction.id();
         for input in transaction.inputs() {
             match input {
                 Input::CoinSigned(CoinSigned { utxo_id, .. })
@@ -298,7 +321,14 @@ where
                     // remove message
                     self.messages_spenders.remove(nonce);
                 }
-                _ => {}
+                Input::Contract(ContractInput { contract_id, .. }) => {
+                    if let Some(users) = self.contract_users.get_mut(contract_id) {
+                        users.retain(|id| id != &tx_id);
+                        if users.is_empty() {
+                            self.contract_users.remove(contract_id);
+                        }
+                    }
+                }
             }
         }
         for output in transaction.outputs().iter() {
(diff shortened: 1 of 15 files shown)


crates/services/txpool_v2/src/spent_inputs.rs
```diff
@@ -27,13 +27,17 @@ pub struct SpentInputs {
     /// transaction spent it. Later, this information can be used to unspent
     /// or fully spend the input.
     spender_of_inputs: HashMap<TxId, Vec<InputKey>>,
+    /// Inputs permanently spent during preconfirmation processing, saved so
+    /// they can be rolled back if the preconfirmation turns out to be stale.
+    tentative_spent: HashMap<TxId, Vec<InputKey>>,
```
MitchTurner, line 32: Is there any chance that his will grow indefinitely?

Seems fine to me. It's not good to test internal values anyway, but worth considering if this is a "leak" in any way.

MitchTurner (commented):
The logic looks good to me. 

I've noticed that our code has a lot more comments than it has in the past. I assume this is due to agents we are using. I'm of the opinion that we include an `AGENTS.md` file in our repos to ensure that we follow some standards. I don't mind the comments in the domain code--although I don't like a ton--I definitely don't like so many in the tests though. 

nit: the test are using given/when/then kinda, but aren't following the `send__when_x_then_y_happens` pattern which in conjunction with removing the comments would make them easier to read.

Dentosal:
> I've noticed that our code has a lot more comments than it has in the past. I assume this is due to agents we are using. I'm of the opinion that we include an `AGENTS.md` file in our repos to ensure that we follow some standards.

Sounds reasonable.

> I don't mind the comments in the domain code--although I don't like a ton--I definitely don't like so many in the tests though.

I personally like having more comments rather than less. IMO our codebases are very under-commented when it comes to reasons why things look like they do, and this makes it rather it hard to follow i.e. what a field is used for. This is especially nice in tests which are most of the time written once and rarely touched again.

> nit: the test are using given/when/then kinda, but aren't following the `send__when_x_then_y_happens` pattern which in conjunction with removing the comments would make them easier to read.

I much prefer the style where the human readable text is in a comment block and the test name is mostly used as a shorthand identifier.

crates/services/txpool_v2/src/spent_inputs.rs
Dentosal: I don't think so, unless blocks are not getting imported at all. These are always cleared for when the associated block height gets imported.

MitchTurner:
> I much prefer the style where the human readable text is in a comment block and the test name is mostly used as a shorthand identifier.

In it's best form, the `send__` style of tests end up being a series of small tests that you can quickly check the behavior/coverage. So if your tests all start looking like each other, and are only 10-20 lines long, you can quickly understand the scope of all your tests. i.e. If I'm checking coverage of my code, having 10 short tests is much less cognitive overhead, where the differences are encapsulated in the name of one variable and the  `expect` in the `//then` block, rather than having to read each test anew.

That's the vision.

Voxelot:
I think there is still a rollback gap for spent-input cleanup in the producer-local extracted-first path.

The happy paths seem fine:

1. Sentry / non-extracted path
- T is still in txpool when the preconfirmation arrives
- `process_preconfirmed_committed_transaction(T)` finds T in `tx_id_to_storage_id`
- it records rollback state with `record_tentative_spend(T, inputs)`
- if canon later omits T, `unspend_preconfirmed(T)` restores those inputs from `tentative_spent`

2. Local producer / extracted path where T becomes canonical
- T is extracted first, so `maybe_spend_inputs(T, inputs)` marks T’s inputs as spent in `spent_inputs` and stores the input list in `spender_of_inputs[T]`
- later the local preconfirmation arrives and `spend_inputs_by_tx_id(T)` drains `spender_of_inputs[T]`
- if the canonical block includes T, this is fine: the inputs are supposed to remain spent, and no rollback is needed

The unhappy path is local producer + rollback:

1. T is extracted for production
2. `maybe_spend_inputs(T, inputs)` does two things:
   - marks T’s inputs as spent in `spent_inputs`
   - stores the same input list in `spender_of_inputs[T]`
3. T is removed from pool storage/selection
4. local preconfirmation for T arrives
5. `process_preconfirmed_committed_transaction(T)` calls `spend_inputs_by_tx_id(T)`
6. `spend_inputs_by_tx_id(T)` drains `spender_of_inputs[T]`, but it does NOT clear those inputs from `spent_inputs`:
   - it removes the bookkeeping entry
   - it leaves/reapplies the live spent markers in `spent_inputs`
7. because T was already extracted, it is no longer in `tx_id_to_storage_id`, so the branch that calls `record_tentative_spend(T, inputs)` is skipped
8. later the canonical block omits T
9. `rollback_preconfirmed_transaction(T)` calls `unspend_preconfirmed(T)`
10. `unspend_preconfirmed(T)` only clears:
   - `InputKey::Tx(T)`
   - the per-input keys saved in `tentative_spent[T]`
11. but `tentative_spent[T]` was never populated in this extracted-first path
12. so T’s coin/message input keys remain present in `spent_inputs`

That is the bug: the original inputs remain marked spent in `spent_inputs`, but the metadata needed to clear them was lost when `spender_of_inputs[T]` was drained and never copied into `tentative_spent[T]`. **This means these inputs will be silently rejected by the authority node (ie p2p message rejection) until they expire from the LRU cache, blocking the user from retrying their transaction!**

Suggested fix direction:
- when a preconfirmation arrives for an already-extracted tx, preserve the original input keys in rollbackable state before `spend_inputs_by_tx_id(T)` drains `spender_of_inputs[T]`
- in other words, the extracted / `maybe_spent` state needs to transition into `tentative_spent`, even when the tx is no longer present in `tx_id_to_storage_id`

I think this needs a regression test covering:
- producer extracts T
- producer locally processes preconfirmation for T
- canonical block at that height omits T
- T’s original inputs are no longer present in `spent_inputs`

Voxelot:
Another potential rollback gap to consider is transactions that were admitted only because a preconfirmed tx temporarily created a contract.

The coin-dependent rollback path looks covered, but I do not see equivalent cleanup for contract-created dependents. The happy path is straightforward: if a preconfirmed tx `P` creates contract `C`, and a dependent tx `D` using `Input::Contract(C, ...)` is admitted while that preconfirmation is live, then everything is fine if the canonical block later includes `P`, because `C` really exists.

The concerning case is when `P` is preconfirmed, txpool records that temporary contract existence in `extracted_outputs`, and `D` is admitted because validation accepts `extracted_outputs.contract_exists(...)`, but the canonical block at that height later omits `P`. Rollback clears the extracted outputs for `P`, but `rollback_preconfirmed_transaction(P)` appears to remove dependents only through `get_coins_spenders(P)`. I do not see a symmetric cleanup path for txs that were admitted only because the temporary contract existed.

So this looks like a separate rollback edge case: the temporary contract creation can be removed, but a dependent tx that was only valid because of that temporary contract may still remain in the pool.

It would probably be worth either adding symmetric contract-dependent tracking alongside `get_coins_spenders(...)`, or deriving those dependents from the dependency graph during rollback. I think this also needs a regression test covering a preconfirmed tx that creates a contract, a dependent tx inserted via `Input::Contract(...)`, and then rollback when the canonical block omits the creator.

Voxelot:
A third thing worth considering is stale preconfirmations that arrive after the node has already imported a newer canonical block.

This feels less like a pure cleanup/rollback issue and more like an admission check: I do not see `process_preconfirmed_transaction(...)` rejecting a preconfirmation whose `tx_pointer.block_height()` is already at or below the node’s current canonical height.

That means if the node has already imported canonical height `H`, and a delayed preconfirmation later arrives for some height `<= H`, txpool can still apply it immediately:
- mark the tx as committed from txpool’s perspective
- insert its `resolved_outputs` into `extracted_outputs`
- admit dependents against those outputs
- mark inputs spent

This is only really a problem if the late preconfirmation is for a tx that did not actually end up in the canonical block at the height referenced by its `tx_pointer`. In that case, the preconfirmation is now stale, but txpool will still temporarily treat it as live. That can block the tx’s original inputs from being usable, and can also temporarily admit dependents against outputs that never became canonical.

This PR will eventually reconcile that state on a later block import, so it is not necessarily permanent corruption. But it does create a real stale-acceptance window where txpool can temporarily treat old, non-canonical preconfirmed state as live until the next block import arrives.

So I think there is a real correctness risk here, not just a defensive tidy-up. If stale preconfirmations can arrive late, they should probably be ignored before mutating txpool state whenever `preconf_height <= current_canonical_height`.

I think this needs a regression test covering:
- node imports canonical block at height `H`
- delayed preconfirmation later arrives for height `<= H`
- the referenced tx is not in the canonical block at that height
- txpool ignores the preconfirmation and does not mutate `spent_inputs` / `extracted_outputs`

crates/services/txpool_v2/src/collision_manager/basic.rs
```diff
@@ -67,6 +71,7 @@ impl<StorageIndex> BasicCollisionManager<StorageIndex> {
             messages_spenders: HashMap::new(),
             coins_spenders: BTreeMap::new(),
             contracts_creators: HashMap::new(),
+            contract_users: HashMap::new(),
```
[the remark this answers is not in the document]
Dentosal: Fixed in 84b0c4ab36

Dentosal:
@Voxelot Addressed all these in 01410e0. I mostly followed your suggested fixes. I added tests for all of them too.

Voxelot (approved):
fixes look good, not sure why ci isn't happy though. will approve again if needed once xi is passing.

crates/services/txpool_v2/src/pool.rs
```diff
@@ -685,6 +731,87 @@ where
...
+        // Capture contracts created by this preconfirmation BEFORE clearing
+        // extracted_outputs, so we can evict any pool txs that were admitted
+        // only because of the now-stale temporary contract existence.
+        let created_contracts: Vec<_> =
+            self.extracted_outputs.contracts_created_by(&tx_id).to_vec();
+
+        // Remove preconfirmed outputs so dependents can't use them.
+        self.extracted_outputs.new_skipped_transaction(&tx_id);
+        // Allow the transaction itself to be re-submitted.
+        self.spent_inputs.unspend_preconfirmed(tx_id);
```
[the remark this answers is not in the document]
Voxelot: @Dentosal I don't think this is a real issue in practice, as we only use utxo_validation = false as a developer/debug mode.

crates/services/txpool_v2/src/collision_manager/basic.rs
```diff
@@ -298,7 +348,14 @@ where
...
                 }
-                _ => {}
+                Input::Contract(ContractInput { contract_id, .. }) => {
+                    if let Some(users) = self.contract_users.get_mut(contract_id) {
+                        users.retain(|id| id != &tx_id);
+                        if users.is_empty() {
+                            self.contract_users.remove(contract_id);
+                        }
+                    }
+                }
```
xgreenx, line 358: I think removing transaction IDs here, you will not have access to them during `rollback_preconfirmed_transaction`. Because we remove it when a transaction is included in the block during production, while you call `rollback_preconfirmed_transaction` after block is done, so you will not have access to them. It only should affect authority, I guess, so maybe it is fine. I think sentries should still be able to work with it. But maybe we want to move this clean up int o`process_block`

Dentosal: Follow-up: https://github.com/FuelLabs/fuel-core/pull/3268

PR #3278, opened 2026-04-22
fix(poa): prevent block-production hang when one Redis node goes half-alive

Voxelot:
## Summary

- Bumps `redis` crate from `1` → `1.2`. Upstream `connect()` was rewritten to apply read/write socket timeouts to the post-connect handshake pipeline (`CLIENT SETINFO LIB-NAME` / `LIB-VER`); on the previously-deployed `0.27` line, `get_connection_with_timeout` only bounded the TCP connect step, so a half-alive peer hung the call indefinitely.
- Replaces `std::thread::scope` in `publish_block_on_all_nodes` with detached `std::thread::spawn` workers reporting into an mpsc channel, returning as soon as `Written` quorum is reached. Stragglers are abandoned; their later `tx.send` fails silently because the receiver is dropped. This is class-of-failure protection against any future single-node hang that survives per-syscall timeouts.

## Background

On 2026-04-22 mainnet halted block production at height 51649390 for ~22 minutes when ElastiCache node entered recovery and accepted TCP but stopped responding to commands. The root cause was the interaction between the two issues above:

1. The buggy handshake hung one `scope.spawn` thread inside `redis::Client::get_connection_with_timeout`.
2. `std::thread::scope` waited for every spawned thread to join before returning, so the publish wedged forever.
3. POA's block-production loop was awaiting `commit_result`, whose oneshot never fired. The importer guard semaphore stayed held — observed externally as a flood of `commit is already in the progress: no permits available` errors logged by the sync service.

The redis upgrade closes the specific upstream bug. The thread-scope replacement closes the class of failure independently of the redis client.

## Test plan

- [x] `cargo test -p fuel-core --lib --features leader_lock service::adapters::consensus_module::poa::tests::` → 27 existing publish / lease / repair tests pass.
- [x] `cargo test -p fuel-core --lib service::adapters::consensus_module::hang_repro_test` → new regression test passes. It spawns a TCP listener that accepts and never responds, calls `redis::Client::get_connection_with_timeout(1s)` and asserts it returns within 5 s. Against `redis 0.27` the call would hang indefinitely; against `redis 1.2` it returns `Err` in ~2 s.
- [x] `cargo +nightly-2025-09-28 fmt -- --check` clean.
- [x] `cargo sort -w --check` clean.

## Notes for cherry-pick

This change is intended to be cherry-picked onto a `release/0.47.x-hotfix` branch from `v0.47.4`. The patch applies cleanly there because the `Improve redis publish performance (#3272)` commit is already in `v0.47.4`.

## Follow-ups (not in this PR)

- Move the publish ownership from the `Importer` trait into POA itself so the publish runs natively on the main async runtime (no rayon, no `block_on`, no `std::thread::spawn`). Tracked separately as the architectural cleanup. The bridge approach in this PR is intentionally minimal so it's safe to backport.


crates/fuel-core/src/service/adapters/consensus_module/hang_repro_test.rs
```diff
@@ -0,0 +1,135 @@
...
+/// returns within a bounded time when the peer is half-alive.
+///
+/// On `redis 0.27.x` this call hangs indefinitely; the test would never
+/// return a result and would deadlock or be killed by the test timeout.
+/// On `redis >= 1.2` it returns an `Err` within roughly `2 * node_timeout`
+/// (one timeout per pipeline command in the handshake).
+///
+/// This is the regression guard for the redis-crate upgrade.
+#[test]
+fn redis_get_connection_with_timeout_is_bounded_against_halflive_peer() {
```
MitchTurner: This is a test on an internal component... this doesn't repro the error we saw in production right? Is there a way we can repro the error we saw specifically and then show that these changes fixed it?

MitchTurner: I see your comments in Slack:
> I did a root cause analysis and repro'd the issue in local integration testing.

and 

> I was able to reproduce this block production freeze scenario by creating a dummy mock redis server that accepts connections quickly and stalls the subsequent handshake using the out of date redis sdk. The newer SDK correctly times out.

These just weren't kept permanently in the code, I take.

MitchTurner (approved):
lgtm

Voxelot:
## Empirical proof that the adapter-level test catches the production hang

Running the exact new adapter test (`publish_produced_block__returns_within_bound_when_one_node_is_half_alive`) against the **deployed v0.47.4 code** (redis 0.27 + `std::thread::scope`-based `publish_block_on_all_nodes`) reproduces the mainnet wedge deterministically. I ported the test into a worktree checked out at the `v0.47.4` tag without modifying any production code; only the test module was edited.

### On v0.47.4 (redis 0.27 + scope-based publish) — pre-fix

**Library-level (`redis_get_connection_with_timeout__is_bounded_against_halflive_peer`):**

```
thread '... redis_get_connection_with_timeout__is_bounded_against_halflive_peer' panicked:
REGRESSION: hang did not return within 5s
test ... FAILED
test result: FAILED. 0 passed; 1 failed;
```

Confirms `redis::Client::get_connection_with_timeout(1s)` hangs against a half-alive TCP peer (the `CLIENT SETINFO` pipeline runs without a read timeout on 0.27).

**Adapter-level (`publish_produced_block__returns_within_bound_when_one_node_is_half_alive`):**

```
thread '... publish_produced_block__returns_within_bound_when_one_node_is_half_alive' panicked:
publish must return within deadline — pre-fix it hangs forever: Elapsed(())
test ... has been running for over 60 seconds
```

The `tokio::time::timeout(5s, ...)` wrapper fires and the test panics at the deadline, but the `spawn_blocking` thread is still stuck inside `scope.join` (waiting for the per-node thread that's hung in `get_connection_with_timeout`). The test binary cannot exit cleanly and the runner reports "has been running for over 60 seconds". **This is exactly the production wedge**: the importer's rayon worker is blocked on an un-cancellable sync call, which is why `commit_result` in POA never returned at 11:01 UTC on 2026-04-22.

### On this PR (redis 1.2 + detached threads + quorum short-circuit) — post-fix

Both tests pass in well under the 5-second deadline:

```
test service::adapters::consensus_module::poa::tests::redis_get_connection_with_timeout__is_bounded_against_halflive_peer ... ok
test service::adapters::consensus_module::poa::tests::publish_produced_block__returns_within_bound_when_one_node_is_half_alive ... ok
```

The adapter test returns in ~0.15 s: the 3 healthy redis-server nodes ack `Written` quickly, the channel drain hits quorum (3 of 4), the loop breaks, and the 4th half-alive thread is abandoned.

### How to reproduce locally

```bash
# Post-fix (this PR) — both pass
git checkout fix/poa-redis-hang
cargo test -p fuel-core --lib --features leader_lock \
  service::adapters::consensus_module::poa::tests -- --test-threads=1

# Pre-fix (deployed v0.47.4) — both fail, adapter test wedges the runner
git checkout v0.47.4
# [port the two test bodies + spawn_halflive_redis_listener helper into mod tests]
cargo test -p fuel-core --lib --features leader_lock \
  publish_produced_block__returns_within_bound_when_one_node_is_half_alive
# → panics at the 5s deadline; process hangs until killed
```

MitchTurner:
ran it locally on the old branch and it got hung. Passes with update.

PR #3280, opened 2026-04-23
refactor(poa): move redis publish into PoA, native-async with cached connections

Voxelot:
## Summary

- Moves `BlockReconciliationWritePort` out of `fuel_core_importer::ports` and into `fuel_core_poa::ports`, making it async. The importer no longer knows anything about redis; PoA owns the publish path end to end.
- Rewrites `RedisLeaderLeaseAdapter::publish_produced_block` (and `publish_block_on_all_nodes`, `repair_sub_quorum_block`) as native async: `FuturesUnordered` fan-out, `tokio::time::timeout(node_timeout, ...)` per node, short-circuit on quorum by dropping the remaining futures. No rayon bridge, no `std::thread::spawn` leak.
- Reuses the existing per-node `redis::aio::MultiplexedConnection` cache for publishes so we stop opening a fresh TCP connection per block.

## Context

Follow-up to #3278 (the 2026-04-22 mainnet hang hotfix). That PR kept the importer-owned publish + rayon bridge and used detached sync threads with an mpsc channel for the quorum short-circuit. Discussed with the original authors — the importer never had a reason to own publish, that was an artifact of the initial scaffolding — and confirmed with the team that moving it to PoA is the intended direction. This PR does that cleanup and goes native-async at the same time.

## Behavior-equivalence

- Same publish-then-commit ordering: `reconciliation_port.publish_produced_block(...).await?` is called in `MainTask::produce_block` (and `produce_predefined_block`) immediately before `block_importer.commit_result(...).await?`. If publish fails to reach quorum, the commit is skipped — identical invariant to the old `Source::Local` gate in `_commit_result`.
- Only locally-produced blocks are published. P2P-synced blocks go only through the importer (which no longer touches redis at all).
- Quorum math, fencing behavior, and `write_block.lua` contract are unchanged.
- Cached `MultiplexedConnection`s are cleared on transport errors / timeouts, same pattern already used by `read_latest_stream_entry_on_node` and the lease-ops paths.

## Test plan

- [x] 29 adapter tests under `service::adapters::consensus_module::poa::tests` pass with `--features leader_lock`, including the 2026-04-22 hang-repro (`publish_produced_block__returns_within_bound_when_one_node_is_half_alive`) — the `spawn_blocking` wrapper is gone and the publish returns cleanly in the normal timeout budget.
- [x] 51 service tests in `fuel-core-poa` pass (new trait bound required a `BlockReconciliationWritePort` impl on `FakeReconciliationPort`).
- [x] `cargo +nightly-2025-09-28 fmt --check` clean.
- [x] `cargo sort -w --check` clean.

## Notes

- Draft until #3278 has shipped to production and soaked; this is not intended to be cherry-picked to 0.47.x.
- The hang-repro test now exercises the full async publish path without `spawn_blocking`. It still asserts a 5s wall-clock deadline; with the native-async fan-out + drop-cancel, a single half-alive node resolves the call in one timeout interval instead of contributing a leaked OS thread.


crates/services/consensus_module/poa/src/service.rs
```diff
@@ -836,7 +849,7 @@ where
     P: P2pPort,
     C: GetTime,
     RS: WaitForReadySignal,
-    RP: BlockReconciliationReadPort + 'static,
+    RP: BlockReconciliationReadPort + BlockReconciliationWritePort + 'static,
```
MitchTurner: nit: do these need to be separate traits?

Voxelot:
Addressed in 5c2002f4e0: collapsed `BlockReconciliationReadPort` + `BlockReconciliationWritePort` into a single `BlockReconciliationPort`. Trait bounds, imports, and impls all simplify; no behavior change. Adapter tests (29) and PoA service tests (51) pass; `cargo clippy --all-targets --all-features` clean.

Voxelot:
Addressed in bf285299c4: `drop_release_guard` is now `Option<Arc<()>>`. Logical clones carry `Some` (existing Drop semantics preserved). Spawned-task clones in `publish_block_on_all_nodes` set it to `None` so they don't extend the logical adapter lifetime — the owning adapter's Drop now releases the lease immediately even with in-flight background publish tasks. Added a regression test (`drop__after_publish_releases_lease_promptly_despite_inflight_background_tasks`) that publishes against 3 healthy + 1 half-alive nodes, drops the adapter, and asserts release happens within a window under `node_timeout` while the half-alive node's background task is still running. 30 adapter tests pass; clippy clean.

Voxelot:
Added `poa_outstanding_publish_tasks` gauge in b5a04f94b6 — tracks background per-node publish tasks via an RAII guard that inc's at spawn time and dec's on closure exit (covers normal completion, panics, mid-await cancellation). Regression test `publish_produced_block__outstanding_tasks_metric_drains_to_baseline` verifies the gauge climbs above baseline immediately after quorum short-circuit and drains back down once background tasks time out. Useful as an alertable signal in production: sustained non-zero would indicate per-call timeouts aren't holding.

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -937,170 +945,155 @@ impl RedisLeaderLeaseAdapter {
...
+            anyhow::Result<WriteBlockResult>,
+        )>();
+        for idx in 0..n {
+            // Clone the adapter for the spawned task, but null out the
+            // drop guard so this clone doesn't pin the logical adapter
+            // lifetime. Without this, a lingering background task after
+            // quorum short-circuit would keep `Arc::strong_count(...) > 1`
+            // and the owning adapter's `Drop` would skip lease release.
+            let mut adapter = self.clone();
+            adapter.drop_release_guard = None;
```
MitchTurner: I think this would be better as a helper method on `self`.

``` 
self.clone_non_reference_counted()
```
or something like that.

BUT... do we even need this reference counter at all anymore? Wasn't it because we were sharing it between the two services as separate port impls?

MitchTurner: I think we still clone it in `sub_services.rs`, but that's just residual.

MitchTurner: Ended up creating a new `RedisPublishChild` for handling the dropping for a given adapter. We don't need the `drop_release_guard` anymore.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -417,6 +417,14 @@ where
             consensus: seal,
         };
 
+        // Publish the block to the leader-lock backend BEFORE committing
+        // it locally. If publish fails to reach quorum, do not commit —
+        // this preserves the invariant "no block exists in the local DB
+        // unless it has been published to quorum by its producer".
+        self.reconciliation_port
+            .publish_produced_block(&block)
+            .await?;
```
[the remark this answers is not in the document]
MitchTurner: That was the old behavior... it just published inside of `commit_result`

Voxelot:
@Voxelot Pushed `6efaceab` addressing the read-path concern (slow node holding connection arc-mutex blocking other ops):

**Refactored to FuturesUnordered (cancel-on-quorum, no spawn):**
- `has_lease_owner_quorum` — drops the stream on quorum-of-owners; the still-pending half-alive read is cancelled cooperatively. The expand-to-non-owners write extension is still a detached background task.
- `should_reconcile_from_stream` — short-circuits on the first node reporting backlog. The "no-backlog" path waits for all responses (otherwise empty-stream nodes can race past an orphan-holding minority and we'd miss a sub-quorum repair cycle), but each read is bounded by `node_timeout`.

**Kept the spawn pattern:**
- `acquire_lease_if_free` — writes the lease, so background tasks usefully extend coverage to slow peers (same durability rationale as the publish path).

**Removed:**
- `OutstandingReadTaskGuard` + `poa_outstanding_read_tasks` gauge — cancelled futures are immediate, not long-lived, so the gauge wasn't meaningful.

**Added regression tests** modeled on `publish_produced_block__returns_within_bound_when_one_node_is_half_alive`:
- `acquire_lease_if_free__returns_within_bound_when_one_node_is_half_alive`
- `has_lease_owner_quorum__returns_within_bound_when_one_node_is_half_alive`
- `unreconciled_blocks__returns_within_bound_when_one_node_is_half_alive_with_backlog`
- `unreconciled_blocks__returns_within_bound_when_one_node_is_half_alive_no_backlog`

All 35 leader_lock tests pass under `--test-threads=1`. Workspace clippy + fmt clean.

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -136,7 +137,10 @@ pub struct RedisLeaderLeaseAdapter {
     stream_max_len: u32,
 }
 
-struct RedisPublishChild {
+/// Per-node operation context. Carried by spawned tokio tasks so they
+/// don't need a clone of the full adapter — covers publish, lease-owner
+/// check, lease promotion, and stream reads.
+struct RedisNodeOp {
```
MitchTurner: not sure what `Op` stands for

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -414,42 +358,72 @@ impl RedisLeaderLeaseAdapter {
...
+        Ok(false)
+    }
+
+    /// Best-effort: try to acquire the lease on nodes we don't already
+    /// own (or haven't yet confirmed as owners). If a newly-acquired
+    /// node reports a higher epoch (election-storm drift), adopt it so
+    /// `write_block.lua` uses a consistent epoch across owned nodes.
+    /// Runs as a fire-and-forget background task so the calling
+    /// `has_lease_owner_quorum` can short-circuit on quorum.
+    fn spawn_expand_to_non_owners(&self, confirmed_owner: &[bool]) {
```
MitchTurner: Another issue I have with agents. This name is nonsense to me.

Why not just name it `try_to_acquire_the_lease_on_nodes_we_don't_already_own` and remove the comment? Or keep the comment for the other details.

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -587,65 +593,38 @@ impl RedisLeaderLeaseAdapter {
...
+        while let Some(result) = reads.next().await {
             match result {
-                Ok(Some((latest_height, _latest_stream_id))) => {
-                    successful_reads = successful_reads.saturating_add(1);
+                Ok(Some((latest_height, _))) => {
                     if latest_height >= next_height {
-                        nodes_indicating_backlog =
-                            nodes_indicating_backlog.saturating_add(1);
+                        drop(reads);
+                        return Ok(true);
```
MitchTurner: This returns `Ok(true)` on the first one that is `>=`? But what if we don't have quorum why would we trust the first? Maybe I'm missing something?

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -937,170 +939,149 @@ impl RedisLeaderLeaseAdapter {
...
             let tx = tx.clone();
+            let block = block.clone();
             let block_data = block_data.clone();
-            let block_stream_key = block_stream_key.clone();
-            let epoch_key = epoch_key.clone();
-            let lease_key = lease_key.clone();
-            let lease_owner_token = lease_owner_token.clone();
-            std::thread::spawn(move || {
-                let result = Self::invoke_write_block_script(
-                    &client,
```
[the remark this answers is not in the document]
Voxelot: Obsolete — `drop_release_guard` was removed entirely in 34531de9a8 ("use explicit types instead of implicit Arc to track if the adapter should be dropped"). The publish path no longer clones a release-guard Arc into spawned tasks.

crates/services/consensus_module/poa/src/service.rs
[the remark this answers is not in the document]
Voxelot: Acknowledged — this ordering was an explicit decision after talking with the original authors. The publish-inside-importer was an artifact and not load-bearing for atomicity: followers reconcile from Redis on next sync, so a published-but-uncommitted block converges naturally on local restart/sync. Keeping the publish in PoA so the importer remains a pure local-state component.

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -448,32 +435,63 @@ impl RedisLeaderLeaseAdapter {
                     *epoch = Some(max_new);
                 }
             }
-        }
-
-        Ok(true)
+        });
```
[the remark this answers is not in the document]
Voxelot: Fixed in 580c67836b — `spawn_expand_to_non_owners` now uses `futures::future::join_all` so all non-owner promotions run in parallel, bounded by 1 × node_timeout total.

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -448,32 +441,63 @@ impl RedisLeaderLeaseAdapter {
...
+                if total_responded >= n {
+                    break;
+                }
+            }
+            // Drop the receiver so the still-running tasks' sends are no-ops
+            // (they'll continue to attempt the promotion against their nodes
+            // for write-coverage extension, bounded by node_timeout).
+            drop(rx);
+
+            if acquired {
```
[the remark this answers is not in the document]
Voxelot: Fixed in 478556e7a4 — each spawned `promote_leader` task now max-CASes its returned token into `current_epoch_token` directly via `fold_epoch_max` (in addition to sending on the channel for the parent's quorum tally). Late tokens are no longer discarded when the receiver is dropped; the shared cell converges on the global max within node_timeout. The parent's post-acquire epoch update also uses the same helper so it can never lower a value a late task already raised. Two regression tests added: `acquire_lease_if_free__folds_late_higher_epoch_into_current_epoch_token` (pre-drifts node C and asserts current_epoch_token reaches the higher value) and `fold_epoch_max__monotonic_max_under_concurrent_writers` (unit-level monotonicity guard).

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -448,32 +462,69 @@ impl RedisLeaderLeaseAdapter {
...
     async fn acquire_lease_if_free(&self) -> anyhow::Result<bool> {
         let promotion_start = std::time::Instant::now();
+        let n = self.redis_nodes.len();
         for attempt_index in 0..self.max_attempts {
             let start = std::time::Instant::now();
-            let promoted_nodes = futures::future::join_all(
-                self.redis_nodes
-                    .iter()
-                    .map(|redis_node| self.promote_leader_on_node(redis_node)),
-            )
```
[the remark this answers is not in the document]
Voxelot: Fixed in 59231ec98b — replaced the inline max-CAS in `spawn_expand_to_non_owners` with per-token `fold_epoch_max` calls so the `leader_epoch` gauge moves in lock-step with the cell. Dropped the tracing::debug log since the metric already exposes the transition.

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
```diff
@@ -574,65 +605,38 @@ impl RedisLeaderLeaseAdapter {
...
+        while let Some(result) = reads.next().await {
             match result {
-                Ok(Some((latest_height, _latest_stream_id))) => {
-                    successful_reads = successful_reads.saturating_add(1);
+                Ok(Some((latest_height, _))) => {
                     if latest_height >= next_height {
-                        nodes_indicating_backlog =
-                            nodes_indicating_backlog.saturating_add(1);
+                        drop(reads);
+                        return Ok(true);
```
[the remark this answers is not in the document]
Voxelot: Fixed in 2d2e64acaf — backlog short-circuit now requires both a positive signal AND quorum of successful responses. Empty-stream responses count toward the quorum tally, so a single backlog signal isn't blocked behind the slowest node — just behind quorum-1 more responses, which is the same wait the caller would do downstream anyway. Degraded-quorum now Errs at should_reconcile_from_stream as before, instead of doing a wasted read_stream_entries round.

MitchTurner (dismissed):

crates/fuel-core/src/service/adapters/consensus_module/poa.rs
Voxelot: Renamed to `PerNodeAdapter` (factory `per_node_adapter`) in 1891e64510 — it's the adapter scoped to one redis node, and the methods on it (`publish_block_on_node`, `check_lease_owner`, `promote_leader`, `read_latest_stream_entry`) are the same operations `RedisLeaderLeaseAdapter` would issue, just narrowed to a single node so they can be moved into a spawned task.

gondoi (dismissed):

PR #3289, opened 2026-05-04
bugfix: improve shutdown and stop PoA sooner

MitchTurner:
## Description

Because the services were being being shutdown in the same order they were started, the PoA service can be held up by earlier services. This means that our producers could be nominally "shut down" but the PoA service still runs a while. The old leader would still hold on to its lock, but it wouldn't be broadcasting blocks anymore. 

The solution here is to shutdown all of the services in parallel, this way none of them is blocking the others.

The other part of the PR is taking care of the GraphQL which was the culprit for PoA never being shut down. We've added a timeout to force it to shutdown if it doesn't shut down gracefully.

## Checklist

### Before requesting review

- [ ] Someone else?

### DIFF
--- a/crates/fuel-core/src/service.rs
+++ b/crates/fuel-core/src/service.rs
@@ -531,7 +531,8 @@ impl RunnableTask for Task {
     }
 
     async fn shutdown(self) -> anyhow::Result<()> {
-        for service in self.services.iter() {
+        // Stop services in reverse order so PoA, which is inserted last, is stopped first.
+        for service in self.services.iter().rev() {
             let result = service.stop_and_await().await;
 
             if let Err(err) = result {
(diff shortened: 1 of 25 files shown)


crates/fuel-core/src/service.rs
```diff
@@ -531,21 +531,48 @@ impl RunnableTask for Task {
     }
 
     async fn shutdown(self) -> anyhow::Result<()> {
-        for service in self.services.iter() {
-            let result = service.stop_and_await().await;
+        let total_services = self.services.len();
+        futures::future::join_all((1_usize..).zip(self.services.iter()).map(
+            |(service_num, service)| {
+                stop_sub_service(service_num, total_services, service.as_ref())
```
MitchTurner, line 537: I'd prefer to use the actual service name here, but I don't want to touch to many pieces of code in a bugfix like this. This doesn't have access to `RunnableService::NAME`

xgreenx (approved):

PR #3320, opened 2026-08-04
feat: add /v1/ready readiness endpoint

gondoi:
## Linked Issues/PRs

None — not yet tracked by a GitHub issue.

## Description
Adds a new `GET /v1/ready` HTTP endpoint, separate from the existing `/v1/health` (which stays a dumb, process-only liveness check). `/v1/ready` reports `200`/`503` plus a JSON breakdown of readiness conditions:

- `services_started` — all sub-services have finished `start_and_await()` (new `is_ready()` on the existing `ReadySignal`, backed by an `Arc<AtomicBool>` alongside the existing `Notify`)
- `poa_enabled` / `synced` — whether PoA is running and, if so, whether it's p2p-synced (`SyncState::Synced`). `poa::sync` is now `pub` so `SyncState` is nameable outside the `poa` crate; `SharedState` gains a `sync_state()` accessor sourced from the existing internal `watch::channel` (no new channel introduced)

Overall readiness is `services_started && (poa disabled || synced)`. Leader-lock status is intentionally excluded — readiness isn't leadership, and gating on it would pull healthy followers out of Service Endpoints.

Motivation: `/v1/health` is a hardcoded `{"up":true}` with no DB/sync/PoA check, so it reports ready well before the node actually is — this is intended as the target for a Kubernetes startup/readiness probe.

Tests added:
- `ReadySignal::is_ready()` transitions false → true across `send_ready_signal()`
- `PoAAdapter::sync_state()` returns `None` when PoA is disabled
- Integration test hitting the real `/v1/ready` route on a running node, asserting `200` and all fields `true` once started

Changelog: `.changes/added/3320.md` (PR number may need correcting once assigned).

## Checklist
- [x] Breaking changes are clearly marked as such in the PR description and changelog — N/A, not a breaking change (same category as prior additive param growth to this same `new_service` fn, e.g. #3074, #2859)
- [ ] [The specification](https://github.com/FuelLabs/fuel-specs/) matches the implemented behavior (link update PR if changes are needed) — N/A, operational HTTP endpoint, not part of the GraphQL/consensus spec surface

### Before requesting review

Not applicable — no Rust SDK, Sway compiler, or platform documentation impact.


gondoi:
Pushed a fix (`6c7d7c8`) for a real deadlock in the `synced` computation, found while validating this endpoint on our devnet: `fuelcore-authority-0` never reached `Ready` despite being fully caught up to chain tip.

**Root cause:** `synced` relied on PoA's own `SyncState`, which only transitions `NotSynced -> Synced` after a quiet window (`--time-until-synced`) with no new blocks — every incoming block, including locally-produced ones, restarts that timer (`sync.rs`, `SufficientPeers` branch). `Trigger::Open { period }` (`--poa-open-period`) produces a block unconditionally every `period`, regardless of tx volume. On any chain where `period < time_until_synced`, that quiet window can structurally never occur, so the transition never fires for any node — leader or follower.

Checked our fleet: every active PoA producer (devnet, testnet, our sole Tokyo mainnet producer) runs `--poa-open-period 1s` against `--time-until-synced 2s`. So this wasn't a one-off devnet misconfig; `/v1/ready`'s `synced` bit was essentially unreachable except by a lucky startup race, everywhere.

**Fix:** added an independent signal — this node's locally imported height compared against the highest height any connected peer has reported via heartbeat (`is_height_caught_up_to_peers` / `height_caught_up_to_peers` in `api_service.rs`). Either signal (the existing quiet-window `SyncState`, or the new height comparison) can satisfy the shared `has_synced_once` latch. `PoA`'s `SyncState` itself is untouched — its original leader-promotion semantics are unaffected.

Plumbing note: this needed `combined_read_database` and `p2p_service` available to the new `/v1/ready` handler after they're moved into the GraphQL schema's `.data(...)` calls, so I wrapped each in `Arc` and split off a clone before the move. Only two resolver call sites referenced the pre-`Arc` types (`node_info.rs`); both updated to look up the `Arc`-wrapped type — no other resolver code changes, since `Arc<T>: Deref<Target = T>` keeps every existing method call working unchanged.

Added 3 unit tests for the new pure comparison logic (`is_height_caught_up_to_peers`). `cargo test`/`clippy` clean locally (`-D warnings`).

gondoi:
Good catch — thanks. Pushed the fix (\`96f2912\`).

My earlier grep only matched the turbofish form (\`data_unchecked::<ReadDatabase>()\`) and missed the more common let-binding-annotation form (\`let x: &ReadDatabase = ctx.data_unchecked();\`) used everywhere else in this codebase. Missed 4 call sites, all now updated to \`&Arc<ReadDatabase>\`:

- \`schema.rs\` — \`ReadViewProvider::read_view()\`'s mutation/subscription branch
- \`graphql_api/extensions/required_fuel_block_height.rs\` — runs on every request; also the place that populates the \`ReadView\` the Query-path branch of \`ReadViewProvider\` and other resolvers read back out
- \`schema/storage.rs\` — \`contract_slot_values\`, \`contract_balance_values\`

This time verified against the actual \`fuel-core-tests\` e2e crate, not just unit tests scoped to \`graphql_api::api_service::\`. Ran the integration tests covering exactly these paths: \`required_fuel_block_height_extension\`, \`storage\`, \`tx::txn_status_subscription\` — 22/22 pass.

\`clippy -D warnings\` still clean.

xgreenx:
Thanks for the PR! The goal is right, but I don't like how readiness is assembled on the HTTP side: fetching `SyncState` out of the PoA crate, latching it in the handler (`latch_synced` + `has_synced_once`), and adding a separate peer-height fallback (`height_caught_up_to_peers`) that pulls `Arc<P2pService>` and `Arc<ReadDatabase>` into the router (with the collateral `Arc` churn across `schema.rs`, `node_info.rs`, `storage.rs`, `required_fuel_block_height.rs`). That's three sources of truth for one fact, and the latch lives in the API service, so restarting the API resets it.

Instead, the PoA service should own this state and simply expose one boolean. When the node is fully synced with the network — and therefore ready to produce — PoA switches the flag in its `SharedState`, and everyone (including `/v1/ready`) just reads it:

```
ready = services_started && (poa disabled || poa.is_ready())
```

On the definition of "synced": we don't need to wait for the `--time-until-synced` quiet window at all. It is enough to know that our height gap with the network is 1 or less (local height >= max peer heartbeat height - 1). The quiet-window TTL is exactly why `Synced` can structurally never fire on chains producing blocks faster than the window — which is the problem your peer-height fallback works around from the outside. Let's fix the definition inside the sync task instead of adding a parallel heuristic in an HTTP handler; that also matters beyond probes, since `ensure_synced` gates production takeover too.

With that change this PR simplifies a lot:
- no need to make `poa::sync` pub or expose `sync_state()`;
- `latch_synced`, `has_synced_once`, and `height_caught_up_to_peers` disappear;
- the `Arc<ReadDatabase>`/`Arc<P2pService>` plumbing and the schema-file changes revert;
- `new_service` only needs the readiness handle, not the `PoAAdapter` as well.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -370,6 +383,23 @@ where
...
+        let synced_header = match &*self.sync_task_handle.shared.borrow() {
+            SyncState::Synced(block_header) => Some(Arc::clone(block_header)),
+            SyncState::NotSynced => None,
+        };
+        if let Some(block_header) = synced_header {
+            self.update_last_block_values(&block_header);
+        }
+        if let Ok(Some(db_header)) = self.block_importer.latest_block_header() {
+            self.update_last_block_values(&Arc::new(db_header));
+        }
```
xgreenx: You always use the latest block height for important; what is the point of using `synced_header` at all?

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -35,78 +37,91 @@ pub enum SyncState {
...
 
+/// Drop peer heartbeats older than this when computing the network tip.
+/// ~2× default p2p `heartbeat_max_avg_interval` (20s) so a departed peer's
+/// height cannot pin us `NotSynced` forever.
+const PEER_HEIGHT_TTL: Duration = Duration::from_secs(40);
+/// Period < TTL so a silent tip is pruned soon after expiry, not only after
+/// a full extra TTL of waiting for the next tick.
+const PEER_HEIGHT_PRUNE_INTERVAL: Duration = Duration::from_secs(20);
+
 pub struct SyncTask {
```
xgreenx, line 81: Looks like we duplicate too much logic from fuel-sync. I think instead of duplicating it here, we should ask the syncer what the state is and whether we need to sync or not. Or keep the old logic where we used a timeout. What was the problem with the timeout logic? If we just need to know whether we are good to take over production if there is a difference between us and the network block height(from reserved peers), then we can just ask p2p what is the network block height and compare it to our and decide. We don't need to do that management here

gondoi:
The remaining red check was `cargo-audit`, not nextest. `RUSTSEC-2026-0258` (`h2` unbounded empty DATA frames) was issued 2026-08-18, after this PR's last green run — same class of repo-wide CI hygiene as #3323, not `/v1/ready` behavior.

Pushed `4c52f2a430`:
- bump `h2` 0.4.13 → 0.4.16 (patched)
- ignore the advisory for `h2` 0.3.27 (`hyper` 0.14 / `reqwest` 0.11 / `tonic` 0.11 have no 0.3 patch)

Local `cargo audit` exits 0. Nextest/clippy were already green on the previous SHA.

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -175,12 +199,21 @@ where
         let (last_height, last_timestamp, last_block_created) =
             Self::extract_block_info(clock.now(), last_block);
 
+        let block_importer = Arc::new(block_importer);
+        let local_height = {
+            let importer = Arc::clone(&block_importer);
+            Arc::new(move || importer.latest_block_height().ok().flatten())
+                as Arc<dyn Fn() -> Option<BlockHeight> + Send + Sync>
+        };
```
xgreenx: We shouldn't put `importer.latest_block_height()` into a closure. If we want to share state to answer what height it is, we can pass the importer in there. But it is strange in virst place who will ask about the height PoA service?

crates/services/consensus_module/poa/src/service.rs
```diff
@@ -521,22 +572,19 @@ where
...
+            }
             tokio::select! {
                 biased;
                 result = watcher.while_started() => {
                     return Some(result.map(|state| state.started()).into())
                 }
-                _ = sync_state.changed() => {}
+                _ = sleep(Duration::from_millis(100)) => {}
             }
         }
```
xgreenx: It is a hack, we shouldn't use a loop and sleep. Why the old way was not working for you? You just listen for `sync_state.changed()` to change, and once sync state know that we are synced(in your case height difference meat, then we can inform the PoA service that production is turned on).

crates/services/p2p/src/service.rs
```diff
@@ -410,6 +423,11 @@ impl Broadcast for SharedState {
         Ok(())
     }
 
+    fn peer_disconnected_broadcast(&self, peer_id: FuelPeerId) -> anyhow::Result<()> {
+        self.peer_disconnected_broadcast.send(peer_id)?;
+        Ok(())
+    }
+
```
xgreenx: We can remove functions which we don't need

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -49,6 +50,25 @@ impl SyncState {
     }
 }
 
+/// Live Ready check: Height Gap = reserved-peer network height − local DB height.
+/// Isolated/dev (`min_connected_reserved_peers == 0`) skips the comparison.
+pub(crate) fn height_gap_is_ready(
```
xgreenx, line 59: You don't use this method... I guess the change in this file is not finished. 

The code should use timeout unit lsynced and `height_gap_is_ready`. Once `height_gap_is_ready` returns true, we wait for time unit lsynced, but if this time is zero, then you can move to sycned state right away. 

In this case old CLI will be not ignored and with proper configuration we can achieve the goal what we want

crates/services/consensus_module/poa/src/service.rs
gondoi: Addressed in `311f3b5948`.

Agreed — dropped the dual `synced_header` / DB path. Production tip refresh is now `sync_last_height_from_db()` only (latest DB header for height + timestamp).

crates/services/consensus_module/poa/src/sync.rs
gondoi: Addressed in `311f3b5948`.

Removed the peer-height map / TTL / disconnect-stream management from SyncTask. Ready now snapshots P2P via `reserved_peer_network_height()` and compares against local DB height (`height_gap_is_ready`), combined with the existing `--time-until-synced` debounce as you suggested in the follow-up.

crates/services/consensus_module/poa/src/service.rs
gondoi: Addressed in `311f3b5948`.

Removed the `Arc<dyn Fn() -> Option<BlockHeight>>` closure from `SharedState`. SyncTask now holds `Arc<dyn BlockImporter>` and reads height at check time. `SharedState::is_ready()` only reads the `SyncState` watch channel — SyncTask owns the height-gap logic.

crates/services/consensus_module/poa/src/service.rs
gondoi: Addressed in `311f3b5948`.

Restored `ensure_synced` to wait on `sync_state.changed()` (no poll loop / sleep). SyncTask flips `SyncState` when the height gap + debounce are satisfied, which unblocks production the same way as before.

crates/services/p2p/src/service.rs
gondoi: Addressed in `311f3b5948`.

Removed `peer_disconnected_broadcast` / `subscribe_peer_disconnected` — unused after switching to live P2P reserved-peer height snapshots.

crates/services/consensus_module/poa/src/sync.rs
gondoi: Addressed in `311f3b5948`.

`height_gap_is_ready` is now the gate inside SyncTask: once it returns true we wait `--time-until-synced` (immediate when zero) before moving to `Synced`. CLI flag is active again (not ignored). A 10s recheck keeps the live gap accurate for heartbeat-only entry and lag exit.

gondoi:
@xgreenx Following up on your Aug-12 design note and the Aug-18/19 line comments — addressed in `311f3b5948`:

- PoA owns Ready: SyncTask publishes `SyncState`; `/v1/ready` and `SharedState::is_ready()` only read that flag (no HTTP-side latch / peer-height fallback / Arc\<ReadDatabase\> plumbing).
- Synced = live reserved-peer height gap ≤ `--max-sync-height-diff`, then `--time-until-synced` debounce (immediate when zero).
- No fuel-sync duplication in PoA — P2P snapshot via `reserved_peer_network_height()`.
- `ensure_synced` back on `sync_state.changed()`; unused `peer_disconnected_broadcast` removed.

Replied + resolved on each of your review threads. Happy to iterate if anything still looks off.

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -210,7 +325,7 @@ impl RunnableTask for SyncTask {
                             // we considered to be synced but we're obviously not!
                             if *has_sufficient_peers {
                                 self.inner_state = InnerSyncState::SufficientPeers(block_info.block_header);
-                                self.restart_timer();
+                                self.on_sufficient_peers_activity().await;
```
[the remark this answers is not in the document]
gondoi: Fixed in `206ae1d4ee`.

Publish `NotSynced` **before** `on_sufficient_peers_activity()` so a `time_until_synced=0` re-entry cannot leave the watch channel on `NotSynced` while `inner_state` is already `Synced`. Added `sync_task_network_block_keeps_inner_and_published_state_consistent`.

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -220,17 +335,20 @@ impl RunnableTask for SyncTask {
...
+                self.advance_to_synced_if_ready().await;
+                self.recompute_if_synced_lagging().await;
+                TaskNextAction::Continue
+            }
+            _ = self.synced_recheck.tick() => {
+                // Live P2P snapshot (Green): heartbeats can satisfy the gap while we
+                // sit in SufficientPeers with --time-until-synced 0 and no timer arm.
+                self.on_sufficient_peers_activity().await;
+                self.advance_to_synced_if_ready().await;
+                self.recompute_if_synced_lagging().await;
```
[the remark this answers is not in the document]
gondoi: Fixed in `206ae1d4ee`.

Recheck arm only calls `on_sufficient_peers_activity()` (arms `--time-until-synced` debounce; does not call `advance_to_synced_if_ready`). Tick advances only when `debounce_armed`, so a stale `Interval` from startup cannot skip the CLI. Added `sync_task_recheck_respects_time_until_synced_debounce`.

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -95,18 +142,97 @@ impl SyncTask {
...
+        }
+        if !self.height_gap_ready().await {
+            self.debounce_armed = false;
+            return;
+        }
+        if self.time_until_synced == Duration::ZERO {
+            self.advance_to_synced_if_ready().await;
+        } else {
+            self.restart_timer();
+            self.debounce_armed = true;
```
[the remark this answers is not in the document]
gondoi: Fixed in b432aa2eb0: `on_sufficient_peers_activity` only calls `restart_timer()` when `!debounce_armed`, so the 10s recheck no longer resets an in-flight `--time-until-synced` window. Covered by `sync_task_recheck_does_not_restart_armed_debounce`.

crates/services/consensus_module/poa/src/sync.rs
```diff
@@ -95,18 +142,97 @@ impl SyncTask {
...
+
+    async fn advance_to_synced_if_ready(&mut self) {
+        let InnerSyncState::SufficientPeers(block_header) = self.inner_state.clone()
+        else {
+            return;
+        };
+        if !self.height_gap_ready().await {
+            return;
+        }
+        self.debounce_armed = false;
```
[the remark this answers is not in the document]
gondoi: Fixed in b432aa2eb0: `advance_to_synced_if_ready` clears `debounce_armed` when the height gap fails, so a later timer tick cannot enter Synced without a fresh debounce. Covered by `sync_task_failed_debounce_tick_disarms`.

Voxelot (approved):

xgreenx (approved):

Issue #1340, opened 2023-08-31
Peer Report Configuration


MitchTurner:
As a follow-up to https://github.com/FuelLabs/fuel-core/issues/943 we might want to provide a way for the client operator to modify the `PeerReportConfig` that is passed to the `P2PAdapter`.

It should be configurable via CLI and maybe it worth to have a separate `--p2p-reputation` config.

MitchTurner:
This applies to a number of locations marked with TODOs and this issue number, including heartbeat peer rep, etc.

MitchTurner:
This is probably a good thing to have, but no one is asking for it.

On the other hand, any values we are using inside peer review are arbitrary, so having some control of them is probably important.

PR #3325, opened 2026-08-11
fix(p2p): penalize peers that fail our requests, and prefer reserved peers for block fetching

xgreenx:
Based on `v0.48.2`, targeting `release/v0.48.2` (which points at the pure `v0.48.2` tag).

## The production problem

Fuel mainnet's public p2p sentries run 10-45 blocks behind the block producer.

A **single** misbehaving external peer is responsible for \~100% of the block-header fetch failures. The failures are transport level, not application level:

* `P2P(Timeout)`
* `P2P(ConnectionClosed)`
* `P2P(Io(Custom { kind: Other, error: "Hit the end of buffer, expected more data" }))`

Every one of them aborts the whole import range and costs a full `--request-timeout` (20s by default) of stalled sync. Measured on the sentries: p50 16.6-21.5s behind the producer, with 72-75% of wall clock spent inside a stalled range.

Two things in the code let one peer do that indefinitely.

### 1\. The failure never reaches the peer's reputation

In `crates/services/sync/src/import.rs`, `get_sealed_block_headers` throws the error away:

```rust
p2p.get_sealed_block_headers(range)
    .await
    .trace_err("Failed to get headers")
    .ok()
```

`get_headers_batch` then returns `Batch::new(None, range, vec![])`, and `report_peer` is defined as

```rust
fn report_peer<P>(p2p: &Arc<P>, peer_id: Option<PeerId>, reason: PeerReportReason) {
    if let Some(peer_id) = peer_id { ... }
}
```

so with `peer_id: None` it silently does nothing. The sync service structurally cannot report a transport failure, because by the time the error surfaces there the peer id is gone.

The peer id *is* known, in `crates/services/p2p/src/p2p_service.rs`, in the `request_response::Event::OutboundFailure { peer, error, request_id }` handler — which until now only `tracing::error!`d it and forwarded the error to the response channel.

### 2\. Peer selection ignores reputation and does not prefer reserved peers

`PeerManager::get_peer_id_with_height` chained both peer sets and picked uniformly at random over the union:

```rust
self.non_reserved_connected_peers
    .iter()
    .chain(self.reserved_connected_peers.iter())
    .filter(|(_, peer_info)| peer_info.heartbeat_data.block_height >= Some(*height))
    .map(|(peer_id, _)| *peer_id)
    .choose(&mut range)
```

(with an `// TODO: Optimize the selection of the peer.` on top of it). The bad peer advertises the highest height, so it always passes the height filter and keeps being redrawn — and the operator's own reserved nodes get no preference over a random peer from the open network.

### (A) Penalize peers for transport-level request failures

`FuelP2PService::handle_request_response_event` now reports the peer from the `OutboundFailure` arm, where the peer id is still available.

Only the failures the remote peer is actually responsible for are penalized:

| `OutboundFailure` | Penalized | Why |
| -- | -- | -- |
| `Timeout` | yes | It accepted the request and never answered. This is the failure that costs 20s of stalled sync. |
| `ConnectionClosed` | yes | It went away in the middle of the exchange. |
| `Io(_)` | yes | It wrote a response we cannot read (the truncated-frame errors above). |
| `UnsupportedProtocols` | **no** | A version/protocol mismatch is not misbehaviour, and we already disconnect from those peers to look for one that speaks our protocol. Penalizing here would double-punish and could permanently ban a peer for running a different release. |
| `DialFailure` | **no** | The dial may have failed because of *our* connectivity or NAT; the peer never even received the request. |

The match is exhaustive (`OutboundFailure` is not `#[non_exhaustive]`), so a libp2p bump forces an explicit decision on any new variant rather than silently defaulting to "punish".

**Reserved peers are never penalized.** This already held incidentally — `PeerManager::update_app_score` only looks up `non_reserved_connected_peers`, and a report against a reserved peer falls into the `log_missing_peer` arm. That is now stated in a doc comment on `update_app_score` and pinned by `reputation_reports_never_affect_reserved_peers`, so it cannot be lost in a refactor.

Reports are tagged `p2p_request_response` rather than reusing the heartbeat's `p2p`, so the two kinds of p2p penalty can be told apart in production logs.

### (B) Prefer reserved peers when choosing who to fetch a block height from

`get_peer_id_with_height` now picks uniformly at random among the **reserved** peers that satisfy the height filter, and only falls back to a uniform random choice among the qualifying **non-reserved** peers when no reserved peer holds the requested height. The height-filter semantics are unchanged; the shared filter moved into a small `peers_with_height` helper.

## Penalty magnitude, and why it is `-48`

This is the part that needed the most care, because the reputation decay makes the obvious values wrong in both directions.

The constraints, all from the existing code:

* `MIN_APP_SCORE = -50.0` — crossing it calls `punisher.ban_peer`, which is `libp2p::allow_block_list`. There is no `unblock_peer` anywhere in the repo, so **a ban is permanent for the process lifetime**.
* `DECAY_APP_SCORE = 0.9`, applied to every non-reserved peer **once per second** (`REPUTATION_DECAY_INTERVAL_IN_SECONDS = 1`). The score half-life is `ln(0.5)/ln(0.9) = 6.6s`.
* A peer that keeps timing out can only be reported once per `--request-timeout`, i.e. once per 20s by default.

**Why small penalties are a no-op.** 20 seconds of decay retain only `0.9^20 = 12.2%` of the score. A peer penalized by `p` every 20s converges to

```
p / (1 - 0.9^20) = p / 0.878 = 1.14 * p
```

so it can never reach -50 unless `|p| > 50 / 1.14 = 43.9`. Reusing the heartbeat's `-5.` would settle at `-5.7` forever — it would ship a change that looks like a fix and punishes nobody. (This is also why the existing `-5.` heartbeat penalties never ban anything.)

**Why large penalties are too harsh.** Reusing `missing_block_headers: -100.` bans on the *first* failure, permanently, for one transient network blip. `-50` exactly is nearly as bad: a second failure at *any* later point still crosses the threshold, because `0.9^n` only underflows to zero after \~1.8 hours.

That leaves a narrow feasible band, `43.9 < |p| <= 50`. `-48` sits inside it:

* **One failure**: score `-48`. No ban (`-48 < -50` is false). Decays to `-4.7` after 22s and back above `-0.5` after 44s — a peer with an isolated blip is left alone.
* **Two failures 20s apart** (the observed production cadence): `48 * (1 + 0.9^20) = 53.8` → banned.
* **Ban window**: `48 * (1 + 0.9^dt) > 50` holds for `dt < 30.2s`. So it is "two failures within \~30 seconds", which covers the 20s timeout cadence with \~50% margin (and the faster `ConnectionClosed`/`Io` failures comfortably), while two failures more than 30s apart never ban on their own.

`44 * ...` would also technically work but leaves only 0.85s of margin over the 20s cadence — a little scheduling jitter and the change becomes a no-op again. `-48` is the value that is unambiguously on the right side of both constraints.

The arithmetic is written out on the constant, and both ends are pinned by tests (`single_request_failure_does_not_ban_the_peer`, `repeated_request_failures_ban_the_peer_despite_the_decay`) so that anyone who later tunes the number sees immediately which invariant they broke.

### Note on `PeerReportReason`

I did not add a variant to the sync service's `PeerReportReason` / `PeerReportConfig`. That enum describes *application-level* misbehaviour observed by the sync service and is mapped to a score by the fuel-core adapter; a transport failure is detected inside the p2p swarm loop, which the sync service never sees and which has no access to that adapter. A dedicated constant plus a dedicated reporting-service tag in the p2p crate gives the same independent tunability, and matches how `HeartbeatPeerReputationConfig`'s penalties already live in the p2p crate. Happy to plumb it through `PeerReportConfig` instead (and out to a CLI flag) if you would rather have it configurable — see also FuelLabs/fuel-core#1340.

## Risks

* **False-positive bans.** The sync pipeline runs up to `block_stream_buffer_size` (10) concurrent range requests, each selecting its own peer, so a single 20s network blip can produce two failures against the same peer and ban it. Mitigations: `DialFailure`/`UnsupportedProtocols` are excluded; reserved peers are exempt; the ban is in-memory and cleared on restart; and after (B) a mistakenly-banned *non-reserved* peer barely matters for block fetching anyway. The alternative — the status quo — is a peer that stalls sync forever.
* **The ban is permanent for the process lifetime.** That is pre-existing behaviour (`allow_block_list` with no `unblock_peer`), but this change makes it reachable from a new code path, so it is now easier to hit. A follow-up giving banned peers a way back (time-based unban, or letting the score decay them out of the block list) would make this safer.
* **(B) concentrates block-fetch load on reserved peers.** For a sentry with a small reserved set this shifts header/transaction requests onto the operator's own nodes, which is the intent, but it is a real change in traffic distribution. Nodes with no reserved peers configured are unaffected — the fallback is exactly the old behaviour restricted to non-reserved peers.
* **Reduced peer diversity for block fetching**, which slightly increases exposure if a reserved peer itself serves bad data. Reserved peers are operator-chosen and already trusted elsewhere (they are exempt from gossip rejection and from reputation bans), so this is consistent with the existing trust model.

## Tests

New in `crates/services/p2p/src/peer_manager.rs`:

* `reserved_peer_is_preferred_when_it_holds_the_requested_height` — 2 reserved + 5 non-reserved peers, all at a sufficient height; 100 draws all return a reserved peer.
* `non_reserved_peer_is_used_when_no_reserved_peer_holds_the_requested_height` — reserved peers lag; the fallback returns non-reserved peers, and `None` when nobody holds the height.
* `reputation_reports_never_affect_reserved_peers` — a ban-worthy report leaves the reserved peer at `DEFAULT_APP_SCORE` and unbanned, while the non-reserved peer is scored and banned.
* `single_request_failure_does_not_ban_the_peer` / `repeated_request_failures_ban_the_peer_despite_the_decay` — the two ends of the penalty-magnitude reasoning above.

New/updated in `crates/services/p2p/src/p2p_service/tests.rs`:

* `only_the_peer_at_fault_is_penalized_for_an_outbound_failure` — the per-variant table.
* `req_res_outbound_timeout_works` now also asserts that the peer which timed out ends up with a negative score (it stayed at `0.0` before this change).

Results:

```
cargo test -p fuel-core-p2p --features test-helpers
    test result: ok. 63 passed; 0 failed; 2 ignored

cargo clippy -p fuel-core-p2p --all-targets --all-features   # RUSTFLAGS="-D warnings"
cargo check -p fuel-core --all-targets
cargo +nightly-2025-09-28 fmt --all -- --check
```

all clean.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

### DIFF
--- a/.changes/fixed/3325.md
+++ b/.changes/fixed/3325.md
@@ -1,0 +1,1 @@
+Penalize the reputation of non-reserved peers that fail our request/response requests at the transport level, and prefer reserved peers when selecting the peer to fetch a block height from, so that a single misbehaving peer can no longer stall block synchronization indefinitely.
--- a/crates/services/p2p/src/p2p_service.rs
+++ b/crates/services/p2p/src/p2p_service.rs
@@ -93,9 +93,56 @@ mod tests;
 /// Maximum amount of peer's addresses that we are ready to store per peer
 const MAX_IDENTIFY_ADDRESSES: usize = 10;
 
+/// The name reported along the request/response reputation penalties, so that
+/// they can be told apart from the heartbeat ones in the logs.
+const REQUEST_RESPONSE_REPORTING_SERVICE: &str = "p2p_request_response";
+
+/// Reputation penalty applied to a non-reserved peer every time an outbound
+/// request to it fails at the transport level.
+///
+/// The magnitude is dictated by the reputation decay: the score of every
+/// non-reserved peer is multiplied by `DECAY_APP_SCORE`(0.9) once per second,
+/// and the peer is banned as soon as its score drops below
+/// `MIN_APP_SCORE`(-50). A failing peer can only be reported once per
+/// `set_request_timeout`(20 seconds by default), and 20 seconds of decay keep
+/// only `0.9^20 = 12.2%` of the previous score, so the score of a peer failing
+/// at that rate converges to `penalty / (1 - 0.9^20) = 1.14 * penalty`. Any
+/// penalty weaker than `-50 / 1.14 = -43.9` can therefore never ban a peer, no
+/// matter for how long it keeps failing; it would be a no-op. At the same time,
+/// the penalty must stay above `MIN_APP_SCORE`, otherwise a single transient
+/// failure permanently bans an otherwise healthy peer.
+///
+/// `-48` sits inside that narrow window. One failure leaves the peer at `-48`,
+/// which is not a ban and decays back above `-0.5` in 44 seconds, while two
+/// failures less than 30 seconds apart - `48 * (1 + 0.9^30) = 50.04` - do ban
+/// it. That covers the 20 seconds timeout cadence with margin, while tolerating
+/// isolated blips.
+pub(crate) const REQUEST_FAILURE_PENALTY: AppScore = -48.;
+
 impl Punisher for Swarm<FuelBehaviour> {
     fn ban_peer(&mut self, peer_id: PeerId) {
         self.behaviour_mut().block_peer(peer_id)
+    }
+}
+
+/// Whether the remote peer is at fault for the outbound `error`, and so should
+/// pay `REQUEST_FAILURE_PENALTY` for it.
+///
+/// Only the failures the remote peer is responsible for are penalized: it
+/// accepted the request and never answered it(`Timeout`), it went away in the
+/// middle of the exchange(`ConnectionClosed`), or it wrote a response we can't
+/// read(`Io`). Those are the failures that stall the block import.
+fn is_peer_at_fault(error: &OutboundFailure) -> bool {
+    match error {
+        OutboundFailure::Timeout
+        | OutboundFailure::ConnectionClosed
+        | OutboundFailure::Io(_) => true,
+        // A protocol mismatch is not a misbehaviour, and we already disconnect
+        // from those peers to look for another one that supports the protocol.
+        OutboundFailure::UnsupportedProtocols => false,
+        // The dial may have failed because of our own connectivity, and the
+        // remote peer never even received the request.
+        OutboundFailure::DialFailure => false,
     }
 }
 
@@ -779,6 +826,18 @@ impl FuelP2PService {
                     request_id,
                     error
                 );
+
+                // A peer that doesn't answer our requests stalls the block import
+                // for a full request timeout, so it has to pay for it with its
+                // reputation. Reserved peers are never affected by this, since
+                // `PeerManager::update_app_score` only touches non-reserved peers.
+                if is_peer_at_fault(&error) {
+                    self.report_peer(
+                        peer,
+                        REQUEST_FAILURE_PENALTY,
+                        REQUEST_RESPONSE_REPORTING_SERVICE,
+                    );
+                }
 
                 if let Some(channel) = self.outbound_requests_table.remove(&request_id) {
                     match channel {
--- a/crates/services/p2p/src/p2p_service.rs
+++ b/crates/services/p2p/src/p2p_service.rs
@@ -128,15 +128,28 @@ impl Punisher for Swarm<FuelBehaviour> {
 /// Whether the remote peer is at fault for the outbound `error`, and so should
 /// pay `REQUEST_FAILURE_PENALTY` for it.
 ///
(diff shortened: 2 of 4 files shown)


gondoi (approved):

xgreenx:
Merged into master

